PR #385, opened 2021-09-22
Refactor end-to-end tests

krzysz00:
We've got a lot of copy-pasted `mlir-rocm-runner` calls running around, and also a bunch of tests that are the same thing except with different layouts/directions/datatypes. I've tried to put a dent into these issues with both `mlir-miopen-e2e-validate.sh` (runs all datatypes/layouts/directions for some arguments) and `run-conv-integration-test.sh` (which just abstracts out the mlir-rocm-runner call).

Some of the other e2e tests might also be refactorable, but I couldn't tell if they were at a glance, so I left them as is.

PR #396, opened 2021-09-30
[DON'T SQUASH] LLVM update 9/21/21

sjw36:

PR #401, opened 2021-10-05
[HOTFIX] Teach e2e tests to run single threaded.

whchung:
Empirical e2e testing show GNU parallel with unbounded threads would fail.
Reduce the job count to 1 to make everything work.

Nightly CI run is executed via LLVM lit, and LLVM lit is already using all CPU threads. And with PR #385, GNU parallel is again trying to submit jobs on all CPU threads, potentially ending up with over-subscription and we could bump into some ROCm linux kernel driver-related issues where multiple CPU process trying to submit jobs to GPU.

### DIFF
--- a/mlir/test/mlir-miopen-e2e-validate.sh.in
+++ b/mlir/test/mlir-miopen-e2e-validate.sh.in
@@ -18,7 +18,7 @@ if [[ "$1" == "--test-only-ops" ]]; then
 fi
 
 declare -i exit_status
-parallel run-conv-integration-test.sh -p=false "$@" \
+parallel -j 1 run-conv-integration-test.sh -p=false "$@" \
     @MLIR_XDLOPS@  @MLIR_POPULATE_VALIDATION@ @MLIR_RANDOM_DATA@ \
     --operation {1} --in_layout {2} --fil_layout {3} --out_layout {4} -t {5} -c \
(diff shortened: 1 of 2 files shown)


krzysz00:
@whchung You'll also want to hit https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/miopen-dialect/mlir/utils/jenkins/miopen-tests/miopen_validate.sh with the hotfix

My thoughts are that the cold fix is to both enable multi-GPU in mlir-rocm-runner and then to have -j $NGPU, but that's a future us problem.

krzysz00 (approved):

jerryyin (approved):

whchung:
The PR in its current form fixes 3 out of 5 failing unit tests in nightly CI, still 2 failing. Look into them.

Those 2 failing configs are passing if executed individually.

My current idea is to reduce LLVM lit workers, and possibly ditch the dependency to `GNU parallel`.

whchung:
Close this PR as #405 supersedes.

whchung:
@sjw36 You might want to cherry pick commits in this PR (specifically d6bd4c1) to #396 . We want `GNU parallel` to run with `-j 1` to suppress CPU threads from spawning unbounded. I believe this should address some issues we see running tests from MIOpen side.

PR #402, opened 2021-10-05
[DON'T SQUASH] Device switch support

krzysz00:
[MLIR][GPU] Add support for selecting the GPU on multi-GPU systems.

This allows GPU dialect users to ensure that the GPU they think they're using is the one that's receiving the runtime mgpu* calls, and allows integration tests to be spread across multiple GPUs on systems that have them.

Add option to mlir-miopen-driver to add device selection in host code

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/GPU/GPUOps.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/GPU/GPUOps.td
@@ -526,6 +526,18 @@ def GPU_LaunchOp : GPU_Op<"launch">,
   let hasCanonicalizer = 1;
 }
 
+def GPU_SetDefaultDeviceOp : GPU_Op<"set_default_device",
+                                    [MemoryEffects<[MemWrite]>]>,
+    Arguments<(ins I32Attr:$devIndex)> {
+  let summary = "Set default GPU for operations after this by index";
+  let description = [{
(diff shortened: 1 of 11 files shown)


jerryyin (approved):
Implementation is very good, thanks for your work! Mind adding a [DO NOT SQUASH] on this one?

Now the follow-up question is sort of: Once we done this, will we register ourselves to replicate the entire runtime functions? Because if a binary will need to "get available gpus" before being able to set a particular device. Also, it looks like now all unit test have the burden of choosing a device number now.

krzysz00:
@jerryyin 
> will we register ourselves to replicate the entire runtime functions? 
What's meant here?

And, unless there's something I don't know about in OpenCL, the runtime can take an integer and make it the default device or error without having a preceeding call to "how many GPUs do we have here?", as the library will error out on invalid indices. Also, if the option isn't provided, there are no set_default_device calls emitted and it's assumed that the client does that before calling our code.

As to the question of picking a device, I see a few options
1. Add a flag to mlir-miopen-driver, something like `--pick-device` that tries to pick which GPU to run on with some scheme that'll be close to uniform (for example, `(31 * getpid() + 19) % num_gpus` (I may be misquoting a hash function there)
2. If there's support for it, using which lit worker we're executing on as an index for the device
3. ???

PR #407, opened 2021-10-07
Adding chip applicability check to filter off unsupported architectures

jerryyin:
In this PR, I focus mostly on refactoring:
 - Added the new applicability filter on chip architectures.
 - Did a fair amount of refactoring to decouple applicability and parsing.
   - Future work, use a dedicated component instead of conv2d generator to judge if the config is valid
 - Fix `mlir-miopen-driver` such that it does rely on the command line options to initialize in all scenarios
 - Fix `mlir-miopen-lib` such that it doesn't try to parse in the lowering stages, and scoped parsing to only handle creation stage.


mlir/include/mlir/Dialect/MIOpen/Generator/Conv2dGenerator.h
```diff
@@ -77,9 +80,6 @@ class Conv2dGenerator {
 
   void flipXdlops();
 
-  static Optional<miopen::ConvOpType> getOpForName(const StringRef name);
```
krzysz00, line 80: Thanks for catching the old method declarations

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
```diff
@@ -98,7 +101,7 @@ LogicalResult hasDimensions(const llvm::StringMap<int64_t> &map,
 
 } // namespace
 
-LogicalResult Conv2dGenerator::isValidConvolution() const {
+LogicalResult Conv2dGenerator::isValidDimension() const {
```
krzysz00: Might be better to rename to something like hasValidProblem() ?

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
```diff
@@ -197,6 +200,21 @@ LogicalResult Conv2dGenerator::isValidConvolution() const {
...
+  // We support xdlops iff it is a gfx908 chip, fail otherwise
+  if (config.xdlops && config.chip != "gfx908")
+    return failure();
+
+  // We support up to gfx908 for nonxdlops algorithm
+  // For example, both gfx90a and gfx1030 are unsupported now
+  unsigned int chipHexNumber = 0;
+  sscanf(config.chip.c_str(), "gfx%x", &chipHexNumber);
+  if (chipHexNumber > 0x908)
+    return failure();
```
krzysz00, line 227: Since we're here, might as well reject chips older than 0x900, which'll also catch a sscanf() failure?

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
```diff
@@ -280,65 +298,77 @@ LogicalResult Conv2dGenerator::parseConvConfig(const char *arguments) {
...
+      setting = argMap[key];
+    }
+  };
+
+  std::string arch;
+  strToStr("arch", arch);
+  IsaNameParser parser(arch);
+  auto status =
+      parser.parseIsaName(config.chip, config.triple, config.features);
+  if (status.failed()) {
```
krzysz00: Maybe this could be an `if(failed(parser.parseIsaName(...)))`?

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
```diff
@@ -280,65 +298,77 @@ LogicalResult Conv2dGenerator::parseConvConfig(const char *arguments) {
...
+      argMap["out_layout"], std::string("NGCHW"), std::string("ngkhw"));
+
+  // Determine tensor dimensions.
+  status = parseConvDims(strToLong("batchsize"), strToLong("groupsize"),
                          strToLong("in_channels"), strToLong("in_h"),
                          strToLong("in_w"), strToLong("out_channels"),
                          strToLong("out_h"), strToLong("out_w"),
                          strToLong("fil_w"), strToLong("fil_h"));
+
+  if (status.failed()) {
```
krzysz00, line 380: On the other hand, having a `status =` makes sense here.

mlir/test/mlir-miopen-lib/applicability/chip.mlir
```diff
@@ -0,0 +1,9 @@
+// RUN: mlir-miopen-lib-test --args " --x2 1 --operation conv2d --arch amdgcn-amd-amdhsa:gfx900 --num_cu 64 --in_type fp32 --fil_type fp32 --out_type fp32 --fil_layout GNCHW --in_layout NGCHW --out_layout NGCHW --batchsize 64 --in_channels 1024 --out_channels 1024 --in_h 14 --in_w 14 --out_h 14 --out_w 14 --fil_h 1 --fil_w 1 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h 0 --padding_w 0 --kernel_name conv2d_nchw_kcyx_nkhw --groupsize 1" --option tuningparams; test $? -eq 0 || echo "expected failure" | FileCheck %s --check-prefix=GFX900_NO_XDLOPS
```
krzysz00: I think the "expected failure" case can be handled with `not` , that is `// RUN: not mlir-miopen-lib-test --args ...`

That'll let us get rid of the FileCheck calls and the echo

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
```diff
@@ -2588,30 +2589,54 @@ int main(int argc, char **argv) {
...
 
+  // TODO: Extract isValid check to be its own component
+  auto status = conv2dGenerator.isValidDimension();
+  if (failed(status)) {
+    llvm::errs() << "Invalid convolution dimension\n";
+    exit(1);
+  }
+
+  status = conv2dGenerator.isValidChip();
+  if (failed(status)) {
```
krzysz00: I think this can be moved into the if, same for the other validity check. That is, `if(failed(conv2dGenerator.isValidChip())` as the test.

krzysz00 (commented):
I've mainly got nitpicks, overall this all makes sense.

(I'm willing to approve even if you don't like my feedback)

jerryyin:
> (I'm willing to approve even if you don't like my feedback)

Pleas wait till I addressed all feedback to give the approve. I indeed like mostly all the feedbacks you give. Thanks for catching these!

krzysz00:
> I indeed like mostly all the feedbacks you give. Thanks for catching these!
Thank you!

My comment was more an indication that I didn't see anything _incorrect_ in here, just the sort of thing where people might disagree, and so "no, I don't like that change" wasn't going to be a barrier to an approve.

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
jerryyin: I kept the old naming so that it isn't too generic. `hasValidProblem()` is almost as generic as the top level function `isApplicable()`

mlir/test/mlir-miopen-lib/applicability/chip.mlir
jerryyin: Thanks for letting me know of it, very convenient!

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
jerryyin: Agreed, addressed in 07e2608

mlir/test/mlir-miopen-lib/applicability/chip.mlir
```diff
@@ -0,0 +1,10 @@
+// Note: The below are unsupported case and guarantee that a non zero return code is populated
+
+// Xdlops support gfx908 only
+// RUN: ! mlir-miopen-lib-test --args " --x2 1 --operation conv2d --arch amdgcn-amd-amdhsa:gfx900 --num_cu 64 --in_type fp32 --fil_type fp32 --out_type fp32 --fil_layout GNCHW --in_layout NGCHW --out_layout NGCHW --batchsize 64 --in_channels 1024 --out_channels 1024 --in_h 14 --in_w 14 --out_h 14 --out_w 14 --fil_h 1 --fil_w 1 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h 0 --padding_w 0 --kernel_name conv2d_nchw_kcyx_nkhw --groupsize 1" --option tuningparams
```
krzysz00, line 4: I didn't know you can write `not` as `!` in lit tests

Cool

mlir/include/mlir/Dialect/MIOpen/Generator/Conv2dGenerator.h
```diff
@@ -127,6 +126,8 @@ class Conv2dGenerator {
     return permutation;
   }
   int getBwdDataKernelCount() const;
+  LogicalResult isValidDimension() const;
```
krzysz00: My one complaint is that it feels like the name should be `hasValidDimension()` (and also `hasValidChip()` below) since those make the code more grammatical for me. That is, "if is applicable" makes sense (and its out of context "if generator is applicable", but not "if generator is valid dimension", which I'd prefer to be "if generator has valid dimension"

krzysz00 (approved):
LGTM, aside from some minor nitpicking about names you can address or not

mlir/test/mlir-miopen-lib/applicability/chip.mlir
jerryyin: I thought ! is the only way for logical not, [here](https://bash.cyberciti.biz/guide/Logical_Not_!).

krzysz00: The llvm testsuite provides `not`, which probably has the side effect of being less platform-dependent https://llvm.org/docs/TestingGuide.html#other-features

mlir/include/mlir/Dialect/MIOpen/Generator/Conv2dGenerator.h
jerryyin: Fair enough. I had the naming almost with the mindset of calling those functions from another component.

I feel like conv2d generator component does much more than it should have and it should be:
 - conv2d config/descriptor - the data structure that carry information to describe a conv2d
 - conv2d parser - that only parse conv2d configs to a descriptor
 - conv2d checker - that validate the parsed result
 - conv2d generator - that does only generation

Whereas now we sort of coupled everything together in one class.

mlir/test/mlir-miopen-lib/applicability/chip.mlir
jerryyin: Ha, didn't know that

mlir/include/mlir/Dialect/MIOpen/Generator/Conv2dGenerator.h
krzysz00: Yeah, agreed

It doesn't help that we don't carry that descriptor around much - it's mainly used to create the conv2d op

PR #409, opened 2021-10-08
Replace all uses of vector tuple ops with variadic arguments

krzysz00:
This commit replaces all cases where we passed a tuple argument to
MIOpen ops, such as threadwise_load or blockwise_store with a variadic
argument. This change allows us to not need to maintain vector.tuple
or vector.tuple_get , which have been removed with upstream.

This PR has the following consequences:

- The print formats of {blockwise,threadwise}_{load,store} have been
changed. Variadic arguments are separated by the rest of the list by
some delimiter other than a comma so that they can be parsed
unambigously. This also makes the printed representations more
human-readable and brings them closer in line with
memref.load/memref.store.
- In threadwise_store, the operand attribute in the coordinate
transformations is no longer 1, but the number of values passed in to
be stored. (For example, `miopen.threadwise_store %1, %2 -> %3[%4]` will
have an operand attribute of 2.) This only affects threadwise_store.
- Calls to computeSliceLengths() now pass nullptr as the type for
irrelevant arguments.

These are the changed discussed in https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/306 and would close that issue.

(and so @whchung )

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
+++ b/mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
@@ -108,8 +108,9 @@ inline bool overrideLoadStoreHack(const DictionaryAttr &transformSpec) {
 // Utility function to determine the type to be loaded
 //===----------------------------------------------------------------------===//
 template <typename T>
-inline std::tuple<Type, TupleType, int, int, int>
+inline std::tuple<Type, int, int, int>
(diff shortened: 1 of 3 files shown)


whchung (commented):
In general I agree with this approach. It removes the now-obsolete vector.tuple usage. Let's hold it after #396 lands.

sjw36 (approved):
LGTM.

whchung:
jenkins: retest this please.

PR #413, opened 2021-10-21
Add gemmSizes checks to quickly reject an invalid configuration

yiqian1:
Check gemm sizes for backward data convolutions in miirCreateHandle() so that a configuration with invalid gemm sizes will be rejected without going through the lowering pass. The motivation was to regain the performance loss of the function IsApplicable() in MIOpen's MLIR solvers as described in [this ticket](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1062).

This patch regains most (meaning >90%) of the performance of IsApplicable(). For example, for an invalid configuration, the original IsApplication() took 12us, the current IsApplication() takes 3700us, and with the changes in this PR, IsApplication() would take 180us. Although it’s much faster than the current IsApplication(), it’s still 15x slower than the original one. This is due to the overhead of passing the configuration to MLIR library and extracting necessary data for analysis.

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/Generator/Conv2dGenerator.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Generator/Conv2dGenerator.h
@@ -14,6 +14,7 @@
 #define MLIR_DIALECT_MIOPEN_CONV2DGENERATOR_H_
 
 #include "mlir/Dialect/MIOpen/MIOpenOps.h"
+#include "mlir/Dialect/MIOpen/utility/BackwardDataValidation.h"
 #include "mlir/IR/BuiltinOps.h"
 #include "mlir/Support/LogicalResult.h"
 
@@ -113,6 +114,9 @@ public:
   }
   LogicalResult isApplicable() const;
 
+  int64_t getBwdDataNumberOfGemm() const;
+  std::tuple<int64_t, int64_t, int64_t> getBwdDataGemmSizes(int64_t) const;
+
 private:
   template <typename Vector>
   std::vector<int64_t> layoutPermutation(const Vector &src,
@@ -128,6 +132,7 @@ private:
   int getBwdDataKernelCount() const;
   LogicalResult hasValidDimension() const;
   LogicalResult hasValidChip() const;
+  LogicalResult hasValidGemmSizes() const;
 
   // Generator config
   Config config;
--- a/mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
+++ b/mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
@@ -126,6 +126,11 @@ LogicalResult Conv2dGenerator::isApplicable() const {
     return failure();
   }
 
+  if (config.operation == miopen::Conv2DBwdDataOpType &&
+      failed(hasValidGemmSizes())) {
+    return failure();
+  }
+
   return success();
 }
 
@@ -258,6 +263,25 @@ LogicalResult Conv2dGenerator::hasValidChip() const {
 
   if ((chipHexNumber > 0x908) || (chipHexNumber < 0x900))
     return failure();
+
+  return success();
+}
+
+LogicalResult Conv2dGenerator::hasValidGemmSizes() const {
+  const int64_t waveSize = 64;
+  if (config.operation != miopen::Conv2DBwdDataOpType)
+    return success();
+
+  int64_t gemmM, gemmN, gemmK;
+  for (int64_t gemmId = 0; gemmId < getBwdDataNumberOfGemm(); ++gemmId) {
+    std::tie(gemmM, gemmN, gemmK) = getBwdDataGemmSizes(gemmId);
+    if (config.xdlops) {
+      if (failed(isValidGridGemmXdlops(gemmM, gemmN, gemmK, waveSize)))
+        return failure();
+    } else if (failed(isValidGemmNonXdlops(gemmM, gemmN, gemmK))) {
+      return failure();
+    }
+  }
 
   return success();
 }
@@ -622,4 +646,42 @@ LogicalResult Conv2dGenerator::genConvModule(ModuleOp &module,
   block->push_back(returnOp);
 
   return success();
+}
+
+int64_t Conv2dGenerator::getBwdDataNumberOfGemm() const {
+  auto gcdStrideDilationH =
+      math_util::gcd(config.strideHeight, config.dilationHeight);
+  auto gcdStrideDilationW =
+      math_util::gcd(config.strideWidth, config.dilationWidth);
+
+  auto yTilda = config.strideHeight / gcdStrideDilationH;
+  auto xTilda = config.strideWidth / gcdStrideDilationW;
+  return yTilda * xTilda;
+}
+
+std::tuple<int64_t, int64_t, int64_t>
+Conv2dGenerator::getBwdDataGemmSizes(int64_t gemmId) const {
+  auto inDim = canonicalizeDims(config.inputDimension, config.inputLayout);
+  auto filDim = canonicalizeDims(config.filterDimension, config.filterLayout);
+  auto outDim = canonicalizeDims(config.outputDimension, config.outputLayout);
+
+  int64_t y = filDim["y"];
+  int64_t x = filDim["x"];
+  int64_t ho = outDim["h"];
+  int64_t wo = outDim["w"];
+  int64_t hi = inDim["hi"];
+  int64_t wi = inDim["wi"];
+  int64_t c = inDim["c"];
+  int64_t n = outDim["n"];
+  int64_t k = outDim["k"];
+  int64_t strideH = config.strideHeight;
+  int64_t strideW = config.strideWidth;
+  int64_t dilationH = config.dilationHeight;
+  int64_t dilationW = config.dilationWidth;
+  int64_t leftPadH = config.paddingHeightLeft;
+  int64_t leftPadW = config.paddingWidthLeft;
+
+  return calculateBwdDataGemmSizes(y, x, c, n, k, ho, wo, hi, wi, strideH,
+                                   strideW, dilationH, dilationW, leftPadH,
+                                   leftPadW, gemmId);
 }
--- a/mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
+++ b/mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
@@ -94,7 +94,6 @@ LogicalResult MIOpenEnabled(const Conv2dGenerator::Config& conf) {
   bool noBF16 = conf.dataTypeStr != "bf16";
   return LogicalResult::success(layoutSupported && noBF16);
 }
-
 } // namespace
 
 typedef void *MiirHandle;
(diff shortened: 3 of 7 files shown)


mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
```diff
@@ -590,3 +590,71 @@ LogicalResult Conv2dGenerator::genConvModule(ModuleOp &module,
...
+}
+
+std::tuple<int64_t, int64_t, int64_t>
+Conv2dGenerator::getBwdDataGemmSizes(int64_t gemmId) const {
+  auto inDim = canonicalizeDims(config.inputDimension, config.inputLayout);
+  auto filDim = canonicalizeDims(config.filterDimension, config.filterLayout);
+  auto outDim = canonicalizeDims(config.outputDimension, config.outputLayout);
+
+  int64_t y, x, ho, wo, hi, wi;
+  y = x = ho = wo = hi = wi = 0;
```
krzysz00: I'd do these as `int y = filDim["y"];` and so on

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
```diff
@@ -590,3 +590,71 @@ LogicalResult Conv2dGenerator::genConvModule(ModuleOp &module,
...
+  y = filDim["y"];
+  x = filDim["x"];
+  ho = outDim["h"];
+  wo = outDim["w"];
+  hi = inDim["hi"];
+  wi = inDim["wi"];
+  auto strideH = config.strideHeight;
```
krzysz00: I'd argue against `auto` here, but that's a personal opinion

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
```diff
@@ -95,6 +95,45 @@ LogicalResult MIOpenEnabled(const Conv2dGenerator::Config& conf) {
   return LogicalResult::success(layoutSupported && noBF16);
 }
 
+LogicalResult validGemmSizes(const Conv2dGenerator &convGen) {
+  const int64_t waveSize = 64;
```
krzysz00: `constexpr`, perhaps?

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
```diff
@@ -95,6 +95,45 @@ LogicalResult MIOpenEnabled(const Conv2dGenerator::Config& conf) {
...
+
+  // obtain GemmSize
+  int64_t gemmM, gemmN, gemmK;
+  std::tie(gemmM, gemmN, gemmK) = convGen.getBwdDataGemmSizes(0);
+  if (config.xdlops) {
+    // unsupported xdlops-gemm
+    if (gemmM % 16 != 0 && gemmN % 64 != 0)
+      return LogicalResult::failure();
+
+    if (!((gemmM * gemmN) % 256 == 0 && (gemmK * gemmM) % waveSize == 0 &&
```
krzysz00: These are some really strict conditions. Where do they come from?

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
```diff
@@ -95,6 +95,45 @@ LogicalResult MIOpenEnabled(const Conv2dGenerator::Config& conf) {
...
+          (gemmK * gemmN) % waveSize == 0 && gemmN % 16 == 0 &&
+          gemmM % 4 == 0 && gemmK % 4 == 0))
+      return LogicalResult::failure();
+  } else {
+    if (gemmM % 32 != 0 || gemmN % 32 != 0 || gemmK % 4 != 0)
+      return LogicalResult::failure();
+  }
+
+  for (int64_t gemmId = 1; gemmId < convGen.getBwdDataNumberOfGemm();
```
krzysz00: I'm suspicious of a for loop that starts from 1. Is this an off-by-1?

mlir/include/mlir/Dialect/MIOpen/Generator/Conv2dGenerator.h
```diff
@@ -113,6 +113,9 @@ class Conv2dGenerator {
   }
   LogicalResult isApplicable() const;
 
+  int64_t getBwdDataNumberOfGemm() const;
+  std::tuple<int64_t, int64_t, int64_t> getBwdDataGemmSizes(int64_t) const;
```
krzysz00, line 118: Name the parameter

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
yiqian1: The same test in the lowering pass: https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/miopen-dialect/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h#L873

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
yiqian1: The above large if-statement checks gemmM and gemmN (they're the same for all gemmId) and gemmK for gemmId=0. This for loop checks gemmK for gemmId>=1.

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
```diff
@@ -112,14 +151,18 @@ extern "C" MiirHandle miirCreateHandle(const char *arguments) {
...
-  OpBuilder builder(&(handle->getContext()));
-
   const auto &config = conv2dGenerator.getConfig();
   if (failed(MIOpenEnabled(config))) {
     return nullptr;
   }
 
+  if (failed(validGemmSizes(conv2dGenerator))) {
+    return nullptr;
+  }
```
jerryyin: Move `validGemmSizes()` into `conv2dGenerator::isApplicable()`

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
```diff
@@ -265,12 +308,13 @@ extern "C" const char *miirGenIgemmCflags(MiirHandle mlirHandle) {
...
 extern "C" MiirStatus miirLowerTuningParams(MiirHandle mlirHandle) {
   const std::lock_guard<std::mutex> lock(mutex);
-  miirLazyInit();
 
   MiirHandle_s *handle = static_cast<MiirHandle_s *>(mlirHandle);
   if (handle == nullptr)
     return MIIR_INVALID_PARAM;
 
+  miirLazyInit();
+
```
jerryyin, line 273: Do the same for `miirLowerBin()`

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
```diff
@@ -590,3 +590,71 @@ LogicalResult Conv2dGenerator::genConvModule(ModuleOp &module,
 
   return success();
 }
+
+int64_t Conv2dGenerator::getBwdDataNumberOfGemm() const {
```
jerryyin, line 651: We already have `getBwdDataKernelCount()`, what's the difference between the two?

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
```diff
@@ -590,3 +590,71 @@ LogicalResult Conv2dGenerator::genConvModule(ModuleOp &module,
...
+  auto gcdStrideDilationW =
+      math_util::gcd(config.strideWidth, config.dilationWidth);
+
+  auto yTilda = config.strideHeight / gcdStrideDilationH;
+  auto xTilda = config.strideWidth / gcdStrideDilationW;
+  return yTilda * xTilda;
+}
+
+std::tuple<int64_t, int64_t, int64_t>
+Conv2dGenerator::getBwdDataGemmSizes(int64_t gemmId) const {
```
jerryyin, line 663: I'm confused:

- Does this apply only to backward data? 
- What's special about bwd?
- Why do we want to do gemm size calculation for bwd? Since I think this will be done in deriving the tuning parameter anyways?

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
jerryyin: Then we should share code between that component instead of copy/paste. What if the condition changed in one location and we forget to update the other?

yiqian1: I see this PR as a temporary workaround. We'll remove these checks when padding kernels are done. So I wanted to limit the changes within IsApplicable()-related code and not touch other code. For the same reason, I put validGemmSizes() in the mlir-miopen-lib.cpp scope, not conv2dGenerator::isApplicable(), as I don't want to affect other callers to conv2dGenerator::isApplicable().

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
yiqian1: getBwdDataKernelCount() returns the number of kernels with gemmK>0. 
I think I can call getBwdDataNumberOfGemm() inside getBwdDataKernelCount().

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
yiqian1: Yes, because my previous changes to MIOpen only affected bwd. https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1016

That PR moved the gemmSizes checks into MLIR's deriving-tuning-parameters pass and resulted in huge performance degradation of MIOpen's IsApplicable(). [(this ticket)](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1062). I'm now moving the gemmSizes checks early so that we don't need to go through the lowering pass to find an in-applicable config.

We could do this for other directions, but that's not related to the problem I'm trying to solve. And again, we'll eventually remove these checks, so I don't feel we need to do this for fwd and wrw.

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
jerryyin: > I see this PR as a temporary workaround.

I'm not sure we are on the same page here. This only stand if and only if you are based on a number of assumptions, including:
 - Will padding kernel come to production soon enough
 - How badly hurt we are by the applicability and to what extend is it tolerable by MIOpen

But you are right in terms that this was originally aimed at addressing bwd applicability check performance only and we have the next release to address other issues.

>  as I don't want to affect other callers

You won't affect other callers as long as there is check to scope the check to bwd only.

----

Please start a documentation ticket that document the things you have done to speed up bwd. Also, mention the padding kernel in that ticket. The goal of that ticket is to find out how soon padding kernel support is complete and what we want to do with those "temporary" fixes - Either fix them to make them permanent or revert them. 

Although they are temporary in nature, several changes need to happen before this PR is mergable:
 - Eliminate the duplication between those applicability checks and the checks in tuning parameter derive stage. Move them to common header/compilation unit if necessary. If we don't do this work now, very soon no one will dare to touch or remove those those. This will be a different story if both component share the same implementation and we have comments saying that putting those checks beforehand is only a short-term hack.
 - Move the checks you have for this PR to be part of `isApplicable()` check: isApplicable check is used in more than one place, and across our stack. It is dangerous if we get different behavior based on different applicability checks. Move them into a centralized location so at least we get similar behavior across our stack.

jerryyin (changes requested):
I'd block this PR until review comments are addressed for now.

yiqian1:
Close this PR as we need a better understanding on the applicability checks of MIOpen solvers.

PR #414, opened 2021-10-21
Install headers and include targets for the shared library build

jungpark-mlir:
- Exports header files
- Don't exclude all external/ when BUILD_FAT_LIBMLIRMIOPEN=OFF

Related to https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/348
but not necessary because C-API is already included in libMLIRMIOpen.a

Example build tested with both static and shared library


mlir/CMakeLists.txt
```diff
@@ -34,3 +34,46 @@ add_subdirectory(lib)
 add_subdirectory(lib/CAPI)
 add_subdirectory(test)
 add_subdirectory(tools)
+
+if (NOT LLVM_INSTALL_TOOLCHAIN_ONLY)
+  install(DIRECTORY include/mlir include/mlir-c
```
jerryyin: Does all those files below needed when build with `-DBUILD_FAT_LIBMLIRMIOPEN`?

cmake/llvm-project.cmake
```diff
@@ -53,4 +53,8 @@ list(APPEND CMAKE_EXE_LINKER_FLAGS
   " -Wl,-rpath -Wl,${CMAKE_CURRENT_BINARY_DIR}/external/llvm-project/llvm/lib"
 )
 
-add_subdirectory("${LLVM_PROJ_SRC}/llvm" "external/llvm-project/llvm" EXCLUDE_FROM_ALL)
+if( BUILD_FAT_LIBMLIRMIOPEN )
+  add_subdirectory("${LLVM_PROJ_SRC}/llvm" "external/llvm-project/llvm" EXCLUDE_FROM_ALL)
```
jerryyin: This excludes from any llvm files being installed?

jerryyin (commented):
Please provide a snapshot of exactly what get installed to user system in the below scenarios:
- `-DBUILD_FAT_LIBMLIRMIOPEN=On` only
- Any configuration required for CAPI inference usage

mlir/CMakeLists.txt
jungpark-mlir: Yes but we didn't install the header files which is also needed by the external users.

cmake/llvm-project.cmake
jungpark-mlir: Yes, I believe so.

mlir/CMakeLists.txt
jerryyin: > but we didn't install the header files ...

Who consume those header files? MIGraphX?

jungpark-mlir: Yes, MIGraphX is one of the user and I suppose all the users of MLIRMIOpen and MLIRMIGraphX as a library need the headers.

jerryyin: It's good to confirm, but MIOpen doesn't need any additional header than Miir.h, so you may have to guard those under additional cmake variable to make sure no redundant files get installed in MIOpen integration case.

jungpark-mlir: I see, that's a good point. Thanks for the clarification.

jungpark-mlir:
I've added following build option which exports all mlir libraries and header files.
-DMLIR_EXPORT_ALL=On

jerryyin:
Let me clarify the use case, is it right that:
 - The original static fat lib build still use the -DBUILD_FAT_LIBMLIRMIOPEN=On only
 - The new MIGraphX required option use the combined: `-DBUILD_FAT_LIBMLIRMIOPEN=On` and `-DMLIR_EXPORT_ALL=On`

Please correct me if I'm wrong. Also, please address my previous comment and provide a snapshot of what files will get installed.

jungpark-mlir:
- The original static fat lib build still use the -DBUILD_FAT_LIBMLIRMIOPEN=On only : Correct
- The new MIGraphX required option use the combined: -DBUILD_FAT_LIBMLIRMIOPEN=On and -DMLIR_EXPORT_ALL=On
: No, MLIR_EXPORT_ALL is separate option. MIGraphX requires -DMLIR_EXPORT_ALL=On but can use either static fat libmlirmiopen or dynamically link necessary mlir libraries.

Let me try to get the list of files with -DMLIR_EXPORT_ALL=On but the list is huge, I'll attach a file.

No additional configuration required by MIGraphX for CAPI, it's common MLIR interface. And currently only using it for generating an IR and create a module.

jungpark-mlir:
lib.txt and h.txt for -DMLIR_EXPORT_ALL=On
off.txt for -DMLIR_EXPORT_ALL=Off
(fat library is off)
[export.zip](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/files/7455021/export.zip)

jerryyin (approved):
LGTM

@sjw36 Please also review, such that you can review the installation in MIGraphX aspect. I can only tell that the installation does not introduce redundant files from the fat static library from MIOpen perspective. Therefore giving the approval.

jungpark-mlir:
Discussed with @jerryyin and agreed to expose build target instead of each flag.
Now, user (MIGraphX) can use -DBUILD_MIXR_TARGET=On, which builds static fat MIOpen including MLIR and exports headers.

PR #415, opened 2021-10-21
Update Dockerfile to ROCm 4.3

krzysz00:
@okakarpa Are all the CI nodes updated to the 4.3 kernel driver?

### DIFF
--- a/mlir/utils/jenkins/Dockerfile.rocm
+++ b/mlir/utils/jenkins/Dockerfile.rocm
@@ -2,10 +2,10 @@
 FROM ubuntu:bionic
 MAINTAINER Zhuoran Yin <zhuoran.yin@amd.com>
 
-ARG ROCM_DEB_REPO=http://repo.radeon.com/rocm/apt/4.1.1/
+ARG ROCM_DEB_REPO=http://repo.radeon.com/rocm/apt/4.3.1/
 ARG ROCM_BUILD_NAME=xenial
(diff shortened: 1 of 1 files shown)


okakarpa:
> @okakarpa Are all the CI nodes updated to the 4.3 kernel driver?

The base rocm on private ci [http://rocmhead.amd.com:8080/] all of the nodes are at 4.3.1 except prj47-rack-48(gfx906)

For Public-CI[http://ml-ci.amd.com:21096/] most of the nodes Base rocm are at rocm-4.1

whchung:
Applying `skip-ci` label for now to prevent it being accidentally merged. We want some dry runs on both private / public CI nodes before taking this.

whchung:
@okakarpa in other projects have we tried run 4.3 docker instances on 4.1 batemetals?

jerryyin:
@krzysz00 I assume you did reasonable amount of testing to make sure no regression happens?

jerryyin (approved):
Okay let's use CI to simplify our lives... If we find any regression in adopting this, we can dial back the ROCm release from jenkinsfile.

krzysz00:
If anything, the 4.1 tagged image will still be around, so nothing should break from this PR - we'll just be creating 4.3 images going forward.

okakarpa:
> @okakarpa in other projects have we tried run 4.3 docker instances on 4.1 batemetals?

Miopen did test that way for a while

PR #417, opened 2021-10-21
Somehow this file made it through, though now it causes clang-format …

pcf000:
…failures.  Fix it.


pcf000:
A separate pull request because mine is entirely external and this file is internal.

krzysz00 (approved):
Not sure how that happened, maybe this was the few weeks we weren't doing the clang-format checks

PR #418, opened 2021-10-22
Add resnext101 benchmark configs

jerryyin:
This PR is used for regression test.

1st commit add resnext101 configs according to [here](https://confluence.amd.com/pages/viewpage.action?spaceKey=MLSE&title=Convolution+Models+Performance+Optimization#ConvolutionModelsPerformanceOptimization-Resnext101_32x4d,PytorchNVexample,FP16BS256).

Once test is complete, I will revert 2nd commit cfcb8b3 and merge just so that we also keep the catalog of configs in the first commit.


jerryyin:
We have a full pass in tuning/testing against all resnext101 configs, [here](http://rocmhead.amd.com:8080/blue/organizations/jenkins/mlir/detail/PR-418/10/pipeline).

Reverting d4321cf now

krzysz00 (approved):
I don't see any issue adding this file so that resnext benchmarks/testing will be easier

PR #419, opened 2021-10-26
MLIR#370: Enhance MIIR C-API to take a module built by MLIR C-API

sjw36:
* added miirCreateHandleWithModule
* enhanced mlir-miopen-lib.cpp to add extra passes when created with module
* still needs MIXR passes

See https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/370

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/Pipeline.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Pipeline.h
@@ -21,43 +21,14 @@
 namespace mlir {
 namespace miopen {
 
-struct PipelineConfig {
-  bool highLevelInput = false;
-  bool tuningTest = false;
-  bool runBackend = false;
-  struct PerfConfig {
-    std::string config;
-    int64_t gridSize = 0;
-    int64_t blockSize = 0;
-  } perf;
-  struct TargetConfig {
-    std::string triple;
-    std::string chip;
-    std::string features;
-  } target;
-};
+// Compilation pipeline from MIXR/TOSA to MIOpen
+void addHighLevelPipeline(PassManager &pm);
 
-struct TuningPipeline : PipelineConfig {
-  TuningPipeline(const std::string &perfConfig, int64_t gridSize = 0,
-                 int64_t blockSize = 0)
-      : PipelineConfig{false, true, false, {perfConfig, gridSize, blockSize}} {}
-};
+// Compilation pipeline from MIOpen to LLVM
+void addPipeline(PassManager &pm, const std::string &perfConfig="", bool applicability=false, bool highLevel=false);
 
-struct DriverPipeline : PipelineConfig {
-  DriverPipeline(const std::string &perfConfig)
-      : PipelineConfig{false, false, false, {perfConfig}} {}
-};
-
-template <bool highLevel = false> struct KernelPipeline : PipelineConfig {
-  KernelPipeline(const std::string &triple, const std::string &chip,
-                 const std::string &features,
-                 const std::string &perfConfig = "")
-      : PipelineConfig{
-            highLevel, false, true, {perfConfig}, {triple, chip, features}} {}
-};
-
-// Consolidate MIOpen pipeline creation
-void addPipeline(mlir::PassManager &pm, const PipelineConfig &pc);
+// Compilation pipeline from LLVM to Binary
+void addBackendPipeline(PassManager &pm, const std::string &triple, const std::string &chip, const std::string &features, int32_t optLevel=3);
 
 } // namespace miopen
 } // namespace mlir
--- a/mlir/include/mlir/Dialect/MIOpen/Pipeline.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Pipeline.h
@@ -1,0 +1,65 @@
+//===- Pipeline.h - MIOpen pipeline creation ---------------===//
+//
+// Part of the MLIR Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file declares the miopen::addPipeline routine to consolidate the
+// MIOpen pipeline creation.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef MLIR_DIALECT_MIOPEN_PIPELINE_H
+#define MLIR_DIALECT_MIOPEN_PIPELINE_H
+
+#include <mlir/Pass/PassManager.h>
+
+#include <string>
+
+namespace mlir {
+namespace miopen {
+
+struct PipelineConfig {
+  bool highLevelInput = false;
+  bool tuningTest = false;
+  bool runBackend = false;
+  struct PerfConfig {
+    const std::string &config = "";
+    int64_t gridSize = 0;
+    int64_t blockSize = 0;
+  } perf;
+  struct TargetConfig {
+    const std::string &triple = "";
+    const std::string &chip = "";
+    const std::string &features = "";
+  } target;
+};
+
+struct TuningPipeline : PipelineConfig {
+  TuningPipeline(const std::string &perfConfig, int64_t gridSize = 0,
+                 int64_t blockSize = 0)
+      : PipelineConfig{false, true, false, {perfConfig, gridSize, blockSize}} {}
+};
+
+struct DriverPipeline : PipelineConfig {
+  DriverPipeline(const std::string &perfConfig)
+      : PipelineConfig{false, false, false, {perfConfig}} {}
+};
+
+template <bool highLevel = false> struct KernelPipeline : PipelineConfig {
+  KernelPipeline(const std::string &triple, const std::string &chip,
+                 const std::string &features,
+                 const std::string &perfConfig = "")
+      : PipelineConfig{
+            highLevel, false, true, {perfConfig}, {triple, chip, features}} {}
+};
+
+// Consolidate MIOpen pipeline creation
+void addPipeline(mlir::PassManager &pm, const PipelineConfig &pc);
+
+} // namespace miopen
+} // namespace mlir
+
+#endif // MLIR_DIALECT_MIOPEN_PIPELINE_H
--- a/mlir/include/mlir/Dialect/MIOpen/Pipeline.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Pipeline.h
@@ -26,14 +26,14 @@ struct PipelineConfig {
   bool tuningTest = false;
   bool runBackend = false;
   struct PerfConfig {
-    const std::string &config = "";
+    std::string config;
     int64_t gridSize = 0;
     int64_t blockSize = 0;
   } perf;
   struct TargetConfig {
-    const std::string &triple = "";
-    const std::string &chip = "";
-    const std::string &features = "";
+    std::string triple;
+    std::string chip;
+    std::string features;
   } target;
 };
 
--- a/mlir/include/mlir/Dialect/MIOpen/Pipeline.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Pipeline.h
@@ -25,10 +25,13 @@ namespace miopen {
 void addHighLevelPipeline(PassManager &pm);
 
 // Compilation pipeline from MIOpen to LLVM
-void addPipeline(PassManager &pm, const std::string &perfConfig="", bool applicability=false, bool highLevel=false);
+void addPipeline(PassManager &pm, const std::string &perfConfig = "",
+                 bool applicability = false, bool highLevel = false);
 
 // Compilation pipeline from LLVM to Binary
-void addBackendPipeline(PassManager &pm, const std::string &triple, const std::string &chip, const std::string &features, int32_t optLevel=3);
+void addBackendPipeline(PassManager &pm, const std::string &triple,
+                        const std::string &chip, const std::string &features,
+                        int32_t optLevel = 3);
 
 } // namespace miopen
 } // namespace mlir
--- a/mlir/lib/Dialect/MIOpen/Pipeline/Pipeline.cpp
+++ b/mlir/lib/Dialect/MIOpen/Pipeline/Pipeline.cpp
@@ -1,7 +1,6 @@
-//===- Pipeline.cpp - Create MIOpen compilation pipeline
-//------------------===//
+//===- Pipeline.cpp - Create MIOpen compilation pipeline ---------------===//
 //
-// Copyright 2020 The MLIR Authors.
+// Copyright 2021 The MLIR Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -16,9 +15,11 @@
 // limitations under the License.
 // =============================================================================
 //
-// This
+// This interface adds the MIOpen compilation pipeline for various flows but
+// keeping a unified ordering of the pipeline.
 //
 //===----------------------------------------------------------------------===//
+
 
 #include "mlir/Dialect/MIOpen/Pipeline.h"
 
@@ -36,31 +37,29 @@
 
 using namespace mlir;
 
-//===- miopen::createPipeline -----------------------------------===//
-//===- Consolidate the MIOpen Pipeline here ---------------------===//
+//===- Consolidate the MIOpen Pipelines here ---------------------===//
 
-void miopen::addPipeline(PassManager &pm, const PipelineConfig &pc) {
-  if (pc.highLevelInput) {
-    // passes for TOSA and bufferization
-    pm.addPass(tosa::createTosaToMIOpenPass());
-    pm.addPass(tosa::createTosaToLinalgOnTensors());
-    pm.addPass(createLinalgElementwiseOpFusionPass());
-    pm.addPass(createLinalgBufferizePass());
-    pm.addPass(createFuncBufferizePass());
-    pm.addPass(createBufferResultsToOutParamsPass());
-    pm.addPass(createFinalizingBufferizePass());
-    pm.addPass(miopen::createMIOpenCopyOptPass());
-  }
+void miopen::addHighLevelPipeline(PassManager &pm) {
+  // passes for TOSA and bufferization
+  pm.addPass(tosa::createTosaToMIOpenPass());
+  pm.addPass(tosa::createTosaToLinalgOnTensors());
+  pm.addPass(createLinalgElementwiseOpFusionPass());
+  pm.addPass(createLinalgBufferizePass());
+  pm.addPass(createFuncBufferizePass());
+  pm.addPass(createBufferResultsToOutParamsPass());
+  pm.addPass(createFinalizingBufferizePass());
+  pm.addPass(miopen::createMIOpenCopyOptPass());
+}
 
+void miopen::addPipeline(PassManager &pm, const std::string &perfConfig, bool applicability, bool highLevel) {
   // Passes for lowering MIOpen dialect.
-  pm.addPass(miopen::createAffixTuningParametersPass(
-      pc.perf.gridSize, pc.perf.blockSize, pc.perf.config));
+  pm.addPass(miopen::createAffixTuningParametersPass(0, 0, perfConfig));
   pm.addPass(miopen::createLowerMIOpenOpsStep1Pass());
   pm.addPass(miopen::createAffineTransformPass());
   pm.addPass(miopen::createLowerMIOpenOpsStep2Pass());
 
-  if (!pc.tuningTest) {
-    if (pc.highLevelInput) {
+  if (!applicability) {
+    if (highLevel) {
       pm.addPass(miopen::createMIOpenLinalgAlignPass());
       pm.addPass(createConvertLinalgToAffineLoopsPass());
     }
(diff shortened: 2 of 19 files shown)


sjw36:
@jungpark-mlir this example show a method for generating the func type string from the actual type.

mlir/test/CAPI/tosa_miir.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+  return moduleOp;
+}
+
+static int constructAndTraverseIr(MlirContext ctx) {
+  MlirLocation location1 = mlirLocationUnknownGet(ctx);
+  MlirModule moduleOp1 = makeAndDumpMIXR(ctx, location1);
+
+  const char *arch = "amdgcn-amd-amdhsa:gfx908";
+  MiirHandle handle = miirCreateHandleWithModule(moduleOp1, arch);
```
jerryyin: Need to explicitly destroy the handle

mlir/test/CAPI/tosa_miir.cpp
```diff
@@ -0,0 +1,234 @@
...
+
```
jerryyin: There is also a context inside the handle, I think you need to consolidate the handle into one otherwise it gets confusing.

mlir/test/CAPI/tosa_miir.cpp
```diff
@@ -0,0 +1,234 @@
...
+  int status = miirLowerBin(handle);
```
jerryyin: Could you add test to verify that lowering to tuning params API also work with the "miirCreateHandleWithModule()" API?

mlir/test/CAPI/tosa_miir.cpp
```diff
@@ -0,0 +1,234 @@
...
+  if (status != MIIR_SUCCESS) {
+    printf("miirLowerBin FAILED!\n");
+    return status;
+  }
+
+  mlirModuleDestroy(moduleOp1);
```
jerryyin, line 240: If module get destroyed with the handle, we'll likely not need this. BTW, if that's the case, we should disable this API in case it used in mix-and-match manner.

mlir/test/CAPI/tosa_miir.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+  // CHECK: PASSED!
+  printf("PASSED!\n");
+  return 0;
+}
+
+int main() {
+  MlirContext ctx = mlirContextCreate();
+  mlirRegisterAllDialects(ctx);
+  mlirContextSetAllowUnregisteredDialects(ctx, true /*allow*/);
```
jerryyin: What does this do and why is it necessary?

mlir/test/CAPI/tosa_miir.cpp
```diff
@@ -0,0 +1,234 @@
...
+  //     min_fp = -1.000000e+00 : f32, min_int = -1 : i64} :
+  //     (tensor<1x64x56x56xf32>) -> tensor<1x64x56x56xf32> return %1 :
+  //     tensor<1x64x56x56xf32>
+  //   }
+  // }
+
+  return moduleOp;
+}
+
+static int constructAndTraverseIr(MlirContext ctx) {
```
jerryyin: Can client use a return type like LogicalResult?

mlir/test/CAPI/tosa_miir.cpp
sjw36: Does not appear to be necessary.

mlir/test/CAPI/tosa_miir.cpp
krzysz00: I'd probably argue for `bool` and documenting which direction is which

krzysz00: (or the status enum)

mlir/test/CAPI/tosa_miir.cpp
sjw36: I will make it use the module context.

mlir/test/CAPI/tosa_miir.cpp
sjw36: Since that is not needed for the november deliverable I filed a ticket: https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/373

sjw36:
All review comments have been addressed. Please re-review.

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
```diff
@@ -128,12 +133,50 @@ extern "C" MiirHandle miirCreateHandle(const char *arguments) {
...
+  MiirHandle_s *handle = new MiirHandle_s(m);
+
+  // Conv2dGenerator conv2dGenerator; //(chip, triple, features);
+
+  // walk module and test conv2d
+  // const auto &config = conv2dGenerator.getConfig();
+  // if (failed(MIOpenEnabled(config))) {
+  //   return nullptr;
+  // }
+
```
jerryyin: Remove those comments

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
```diff
@@ -128,12 +133,50 @@ extern "C" MiirHandle miirCreateHandle(const char *arguments) {
...
+  handle->chip = chip;
+  handle->features = features;
+  // handle->perfConfig = config.perfConfig;
+
+  // count kernels from module funcs
+  handle->kernelCount = 0;
+  m.walk([&](FuncOp func) {
+    if (func->hasAttr("kernel"))
+      handle->kernelCount++;
+  });
```
jerryyin: Why do we have to manually count the kernel? It looks like whatever number of kernel you get from the beginning may not be the real number here because there's a likelyhood of populating more: Is there any way you can reuse the `Conv2dGenerator::getKernelCount()`?

mlir/test/CAPI/tosa_miir.cpp
```diff
@@ -0,0 +1,233 @@
...
+  return moduleOp;
+}
+
+static MiirStatus constructAndTraverseIr(MlirContext ctx) {
+  MlirLocation location1 = mlirLocationUnknownGet(ctx);
+  MlirModule moduleOp1 = makeAndDumpMIXR(ctx, location1);
+
+  const char *arch = "amdgcn-amd-amdhsa:gfx908";
+  MiirHandle handle = miirCreateHandleWithModule(moduleOp1, arch);
+
```
jerryyin: It still seems to exists two independent piece of context:
 - The argument of `MlirContext ctx`
 - The `static MLIRContext context` within the handle

mlir/test/CAPI/tosa_miir.cpp
jerryyin: Then I'm pretty sure `miirGetExecutionDims()` is necessary to validate otherwise how does the client know about how the populated kernel can be invoked.

mlir/test/CAPI/tosa_miir.cpp
jerryyin: Also, instead of creating a new module in `makeAndDumpMIXR` why not use the existing module inside the handle? In that sense, you should create the handle first then grab everything out of the handle.

mlir/test/CAPI/tosa_miir.cpp
jerryyin: I think this needs to be addressed: There needs a top-down approach to make sure all the APIs we provide to the client doesn't go into conflict with each other

mlir/tools/mlir-miopen-lib/Miir.h
```diff
@@ -47,6 +49,14 @@ struct StridedMemRef5D {
  */
 extern "C" MiirHandle miirCreateHandle(const char *options);
 
+/*! @brief Create the MLIR handle with an MLIR module
+ *  @param module  An MLIR module built with the MLIR C-API
+ *  @param arch    The architectural designation
+ *  @return        MLIR handle
+ */
+extern "C" MiirHandle miirCreateHandleWithModule(MlirModule module,
+                                                 const char *arch);
```
jerryyin: From the look of it, we perhaps need a different API that is able to give the reference of module owned by the handle, rather than creating a handling using a new module.

Actually I think we do have the `getModule()` function as public available member, so we should use that instead.

mlir/test/CAPI/tosa_miir.cpp
```diff
@@ -0,0 +1,234 @@
...
+  // Set func op
+  MlirOperationState funcState = mlirOperationStateGet(
+      mlirStringRefCreateFromCString("builtin.func"), location);
+  mlirOperationStateAddAttributes(&funcState, 3, funcAttrs);
+  mlirOperationStateAddOwnedRegions(&funcState, 1, &funcBodyRegion);
+  MlirOperation func = mlirOperationCreate(&funcState);
+  mlirBlockInsertOwnedOperation(moduleBody, 0, func);
+
+  //-------------- conv0 = tosa.conv2d
```
jerryyin, line 95: I feel like each of the below shall have some sort of "generator" component to make them re-usable. What if in another test we'd want a different sequence/combination of `tosa.conv2d`, `tosa.relu`

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
sjw36: It is assumed the module is fully formed by this point. The only way to be sure would be to run this test in miirGetKernelCount.

mlir/test/CAPI/tosa_miir.cpp
sjw36: The static MLIRContext is now only used to create the module in the MiirHandle_s constructor. All subsequent references use the context associated with the module.

sjw36: I prefer only the constructor takes the module, like passing ownership. In this fashion it follows the same paradigm of miirCreateHandle where the module gets created and populated and is static from there on. If we had an API where we could give the user a module reference at any time, then it opens the door for them to change even the standard flow.

mlir/tools/mlir-miopen-lib/Miir.h
sjw36: the MiirHandle_s class is opaque from the header file, so that API is not visible. See comments above about exposing the module.

mlir/test/CAPI/tosa_miir.cpp
sjw36: That is basically what Paul is creating in MIGraphX. He walks his internal IR and generates the MLIR accordingly (using the standard MLIR C-API).

sjw36:
Comment from @jerryyin :
`Then I'm pretty sure miirGetExecutionDims() is necessary to validate otherwise how does the client know about how the populated kernel can be invoked.`

I cannot reply above..

But this is a good point and I will add that API test and make sure it works.

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
jerryyin: But since we already use that API to get kernel count, why invent a new way to do it? When the assumption is not right, it is confusing to return a wrong value. e.g, at that scenario, `miirGetKernelCount` will return this "possibly wrong" kernel count from the handle.

mlir/test/CAPI/tosa_miir.cpp
jerryyin: >  then it opens the door for them to change even the standard flow

What benefit does it bring to give user that flexibility? I see this as a violation of the assumption why we create handle in the first place. We should address this by:
 - Either get rid of context/module inside of the handle, making them completely orthogonal
 - Or, do not open the second door to do the same thing

mlir/test/CAPI/tosa_miir.cpp
jerryyin: Sounds good.

mlir/tools/mlir-miopen-lib/Miir.h
jerryyin: That's true, but we can create that API to return the module associated with the handle instead. I think now the question boil to API design and we need to reach at an agreement on the way how client can use them consistently.

sjw36:
I guess this is confusing because we are using this API for 2 different flows. 

In the original flow the kernels are generated based on configuration settings.

In the new flow, a user creates his own module and simply uses the C-API to run the compilation pipeline and get a kernel blob.

This second approach could simply call the MLIR C-API's to do the same. That way the MIIR C-API remains a kernel generator flow with no potential for entanglement.

jerryyin:
> I guess this is confusing because we are using this API for 2 different flows.

Exactly, that's the part I'm worrying about: When another developer come in and see all the available APIs, he will naturally use the combination of the two workflows. However, it is going to be terrible that he later finds out only a subset of API from original flow can be used together with the new flow. In summary, I have no doubt that things can "work" now but this story can happen to you and me when we each further developed the original/new API it becomes more confusing what exactly can/should be used with another.

> This second approach could simply call the MLIR C-API's to do the same. 

Agreed, I'd prefer the second approach to not install/include the original API if possible. However, if there are functionality that they could share, they should. Under the hood, some of the common functionalities such as the lowering pipeline can come from the same compilaiton unit (with some refactoring). We should extract most shared functionality of `mlir-miopen-lib.cpp` into a third party implementation so it can either be pulled into the original or the new flow.

sjw36:
@jerryyin see new update which removes changes to MIIR.

mlir/lib/Dialect/MIOpen/Pipeline/Pipeline.cpp
```diff
@@ -0,0 +1,87 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// =============================================================================
+//
+// This
+//
+//===----------------------------------------------------------------------===//
```
jerryyin: Formatting looks wired. Also 2020 -> 2021

mlir/include/mlir/Dialect/MIOpen/Pipeline.h
```diff
@@ -0,0 +1,65 @@
...
+
+struct PipelineConfig {
+  bool highLevelInput = false;
+  bool tuningTest = false;
+  bool runBackend = false;
+  struct PerfConfig {
+    std::string config;
+    int64_t gridSize = 0;
+    int64_t blockSize = 0;
+  } perf;
```
jerryyin: I'm not convinced they should belong to pipeline.

For example, if we have two convolutions yet we only have one config/gridSize/blockSize, are we guaranteed that both convolutions can use same of those parameters?

mlir/include/mlir/Dialect/MIOpen/Pipeline.h
```diff
@@ -0,0 +1,65 @@
...
+  struct TargetConfig {
+    std::string triple;
+    std::string chip;
+    std::string features;
+  } target;
+};
+
+struct TuningPipeline : PipelineConfig {
+  TuningPipeline(const std::string &perfConfig, int64_t gridSize = 0,
+                 int64_t blockSize = 0)
```
jerryyin: Same here, they shouldn't belong to a pipeline's constructor.

mlir/include/mlir/Dialect/MIOpen/Pipeline.h
```diff
@@ -0,0 +1,65 @@
...
+      : PipelineConfig{false, true, false, {perfConfig, gridSize, blockSize}} {}
+};
+
+struct DriverPipeline : PipelineConfig {
+  DriverPipeline(const std::string &perfConfig)
+      : PipelineConfig{false, false, false, {perfConfig}} {}
+};
+
+template <bool highLevel = false> struct KernelPipeline : PipelineConfig {
```
jerryyin: Why does a pipeline inherits from the config? I feel like the relationship should be composition, not inheritance.

mlir/lib/Dialect/MIOpen/Pipeline/Pipeline.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include "mlir/InitAllDialects.h"
+#include "llvm/Support/TargetSelect.h"
+
+using namespace mlir;
+
+//===- miopen::createPipeline -----------------------------------===//
+//===- Consolidate the MIOpen Pipeline here ---------------------===//
+
+void miopen::addPipeline(PassManager &pm, const PipelineConfig &pc) {
+  if (pc.highLevelInput) {
```
jerryyin: Prefer to be more specific in parameter naming, `tosaInput` is better than `highLevelInput`

mlir/lib/Dialect/MIOpen/Pipeline/Pipeline.cpp
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  // Passes for lowering MIOpen dialect.
+  pm.addPass(miopen::createAffixTuningParametersPass(
+      pc.perf.gridSize, pc.perf.blockSize, pc.perf.config));
+  pm.addPass(miopen::createLowerMIOpenOpsStep1Pass());
+  pm.addPass(miopen::createAffineTransformPass());
+  pm.addPass(miopen::createLowerMIOpenOpsStep2Pass());
+
+  if (!pc.tuningTest) {
```
jerryyin: I'd prefer `pc.applicability`

mlir/lib/Dialect/MIOpen/Pipeline/Pipeline.cpp
```diff
@@ -0,0 +1,87 @@
...
+    pm.addPass(miopen::createLowerMIOpenOpsStep5Pass());
+    pm.addPass(createLowerMIOpenOpsToGPUPass());
+
+    // Passes for lowering linalg dialect.
+    pm.addPass(createConvertLinalgToAffineLoopsPass());
+    pm.addPass(createLowerAffinePass());
+    pm.addPass(createLowerToCFGPass());
+
+    // Passes for lowering ROCDL dialect
+    if (pc.runBackend) {
```
jerryyin: Which pipeline requires runBackend == true?

mlir/test/CAPI/tosa_miir.cpp
```diff
@@ -0,0 +1,257 @@
...
+  MlirContext ctx = mlirContextCreate();
+  mlirRegisterAllDialects(ctx);
+  if (!constructAndTraverseIr(ctx)) {
+    printf("FAILED!\n");
+    return 1;
+  }
+
+  mlirContextDestroy(ctx);
+
+  return 0;
```
jerryyin, line 260: Thanks it looks more consistent now.

mlir/include/mlir/Dialect/MIOpen/Pipeline.h
sjw36: I can drop those because they are only used explicitly by mlir-miopen-driver and it only runs mlir::miopen::createAffixTuningParametersPass and does not use the pipeline.

mlir/lib/Dialect/MIOpen/Pipeline/Pipeline.cpp
sjw36: I will make an enum since it will soon include MIXR as well.

mlir/lib/Dialect/MIOpen/Pipeline/Pipeline.cpp
sjw36: MIGraphX.

mlir/include/mlir/Dialect/MIOpen/Pipeline.h
sjw36: It is the kind of PipelineConfig, see usage. They must derive from PipelineConfig to be passed into addPipeline.

jerryyin: Okay then could you rename it as `KernelPipelineConfig`.

Also, why use the template with the default bool? If you are using this as a thin wrapper you can expose it in the constructor.

mlir/include/mlir/Dialect/MIOpen/Pipeline.h
jerryyin: That sounds good. Then later we should make them part of the conv2d op.

mlir/lib/Dialect/MIOpen/Pipeline/Pipeline.cpp
sjw36: pretty much copied from LowerMIOpenOps.cpp.. will fix the copyright

mlir/lib/Dialect/MIOpen/Pipeline/Pipeline.cpp
sjw36: Also MIIR.

mlir/test/CAPI/tosa_miir.cpp
```diff
@@ -0,0 +1,261 @@
+//===- tosa_miir.cpp - Simple test of C and MIIR APIs ---------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM
+// Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+/* RUN: mlir-tosa-miir-test 2>&1 | FileCheck %s
```
jerryyin, line 10: Interesting way to trigger a test in cpp file.

jerryyin (approved):
Thanks for the work, LGTM

mlir/include/mlir/Dialect/MIOpen/Pipeline.h
```diff
@@ -0,0 +1,36 @@
...
+namespace miopen {
+
+// Compilation pipeline from MIXR/TOSA to MIOpen
+void addHighLevelPipeline(PassManager &pm);
+
+// Compilation pipeline from MIOpen to LLVM
+void addPipeline(PassManager &pm, const std::string &perfConfig="", bool applicability=false, bool highLevel=false);
+
+// Compilation pipeline from LLVM to Binary
+void addBackendPipeline(PassManager &pm, const std::string &triple, const std::string &chip, const std::string &features, int32_t optLevel=3);
```
jerryyin: I like those much better than the previous version to use another config to control the flow of pipelines. Thanks

sjw36: As long as a you call in the right order... 🥇

PR #343, opened 2021-08-17
first version of backward data padding kernel

asleepzzz:
backward data padding kernel with workaround,
non xdlops still have issues when nchw stride 2 pad_h,pad_w 1
I'll disable it

next step of this PR:
1.disable non xdlops nchw stride 2 pad_h,pad_w 1
2.add unit test

at meanwhile, jack and Chao will help me solving the limitations  of padding kernel

PR #420, opened 2021-10-26
Skip 2 E2E tests on non-XDLOPS path.

whchung:
Regression on non-XDLOPS path caused by PR #343 .

Move tests which must run in XDLOPS path into a separate directory. Disable them unless we run E2E tests on a XDLOPS configuration.


whchung:
Notice I didn't replace `-x2` with `%xdlops` variable in those tests, because when @asleepzzz developed #343, he deliberately separated non-XDLOPS e2e tests from XDLOPS e2e tests. And I try to continue honor that because we have experienced quite a few issues on the compiler backend in backward padding kernels, and non-XDLOPS and XDLOPS path have their own specific compiler-backend-related issues.

sjw36 (approved):
LGTM

PR #422, opened 2021-10-27
[AMDGPU] Disable d16 loads/stores to high register halves on gfx90a

krzysz00:
Testing reveals that, on gfx90a, *_load_short_d16_hi does note
preserve the low-order bits of the destination register, even with
SramECC disabled. To prevent the generation of semantically incorrect
code, disable D16PreservesUnesedBits on this target.

(A more specific test may be needed to account for future gfx90*
targets, but checking for gfx90a instructions was what was in the
existing code)

Internal note: This patch causes the PR E2E tests to pass on Lockhart, resolving https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/363, and I'll be sending it upstream shortly.

### DIFF
--- a/external/llvm-project/llvm/lib/Target/AMDGPU/AMDGPU.td
+++ b/external/llvm-project/llvm/lib/Target/AMDGPU/AMDGPU.td
@@ -1343,7 +1343,8 @@ def HasPackedD16VMem : Predicate<"!Subtarget->hasUnpackedD16VMem()">,
 
 def D16PreservesUnusedBits :
   Predicate<"Subtarget->d16PreservesUnusedBits()">,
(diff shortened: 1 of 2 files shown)


krzysz00:
https://reviews.llvm.org/D112635 is the upstream version

krzysz00:
(and since I'm getting review upstream, drafting this)

whchung:
@krzysz00 I guess this PR is no longer need? I just saw you closed D112635 upstream.

krzysz00:
Yeah, we have a better workaround in the form of a kernel parameter and there's a fix being worked on by the driver team, closing.

PR #423, opened 2021-10-28
[DO NOT SQUASH] KPACK logic.

whchung:
KPACK logic.

Divide GemmK dimension into GemmK and GemmKPack dimension to allow wider LDS -> VGPR data transfer.

Features:
- supports fp16 / fp32.
- supports NCHW / NHWC.
- could be tuned via `--perf_config`.
- forward and backward convolutions for Resnet50 configs are all validated.

Limitations:
- XDLOPS path only. The logic does support non-XDLOPS path but is not enabled yet.
- forward convolution only. does NOT support backward convolutions yet.
- does NOT support padding kernels along K/M/N directions.


whchung:
We can NOT trust the green check mark just yet as I disabled quite a few tests on commit 629e309. The next steps for me is to address all those disabled tests in that commit.

whchung:
Attaching skip-ci label for now. And convert it to draft.

Will undo all these after I fix some more e2e unit tests.

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -103,9 +103,9 @@ def MIOpen_TransformOp :
 
 def MIOpen_GridwiseGemmOp :
     MIOpen_Op<"gridwise_gemm">,
-    Arguments<(ins MemRefRankOf<[F32, F16, I16], [3]>:$filter,
-                   MemRefRankOf<[F32, F16, I16], [3]>:$input,
-                   MemRefRankOf<[F32, F16, I16], [3]>:$output)> {
+    Arguments<(ins MemRefRankOf<[F32, F16, I16], [3, 4]>:$filter,
+                   MemRefRankOf<[F32, F16, I16], [3, 4]>:$input,
```
krzysz00: I take it we'll be replacing this with the variadic arguments?

mlir/test/mlir-miopen-driver/auto_e2e/Resnet50/Resnet50_nchw_f16_bwd.mlir
```diff
@@ -1,3 +1,6 @@
+// XFAIL: *
+// FIXME, ALL CONFIGS IN THIS TEST FILE NEED TO BE REVIEWED.
```
krzysz00: Reviewed in what sense?

mlir/test/mlir-miopen-driver/auto_e2e/padding_kernel_all_fwd.mlir
```diff
@@ -3,7 +3,7 @@
...
 // CHECK_RESNET50_CONFIG1: [1]
 
-// RUN: mlir-miopen-driver --operation=conv2d -t f16 -p=false  -fil_layout=gkcyx -in_layout=ngchw -out_layout=ngkhw -batchsize=20 -groupsize=1 -in_channels=3 -out_channels=6 -in_h=32 -in_w=32 -fil_h=7 -fil_w=7 --dilation_h=1 --dilation_w=1 --padding_h=3 --padding_w=3 --conv_stride_h=2 --conv_stride_w=2 %pv %random_data %xdlops -c | mlir-rocm-runner --shared-libs=%rocm_wrapper_library_dir/librocm-runtime-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CHECK_RESNET50_CONFIG2
+// NOT: mlir-miopen-driver --operation=conv2d -t f16 -p=false  -fil_layout=gkcyx -in_layout=ngchw -out_layout=ngkhw -batchsize=20 -groupsize=1 -in_channels=3 -out_channels=6 -in_h=32 -in_w=32 -fil_h=7 -fil_w=7 --dilation_h=1 --dilation_w=1 --padding_h=3 --padding_w=3 --conv_stride_h=2 --conv_stride_w=2 %pv %random_data %xdlops -c | mlir-rocm-runner --shared-libs=%rocm_wrapper_library_dir/librocm-runtime-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CHECK_RESNET50_CONFIG2
```
krzysz00: Why're we flipping tests off?

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -770,15 +776,30 @@ inline void emitNaiveTensorCopyLogic(
...
-      srcUpperIndices.push_back(b.create<IndexCastOp>(
-          loc, b.create<AddIOp>(loc, zeroConstantOp, sourceCoord[0]),
-          b.getIndexType()));
-      srcUpperIndices.push_back(b.create<IndexCastOp>(
-          loc, b.create<AddIOp>(loc, ivo_i32, sourceCoord[1]),
-          b.getIndexType()));
-      srcUpperIndices.push_back(b.create<IndexCastOp>(
-          loc, b.create<AddIOp>(loc, ivi_i32, sourceCoord[2]),
-          b.getIndexType()));
+      if (sourceCoord.size() == 3) {
```
krzysz00: This is all supersceded by #425, sorry about the merge conflicts

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -1767,19 +1810,19 @@ struct Conv2DRewritePattern : public OpRewritePattern<T> {
...
           b.getNamedAttr("upper_layer_names", b.getArrayAttr({b.getStringAttr(
-                                                  arg0TargetLayoutName0)}))};
+                                                  arg0TargetLayoutName[0])}))};
 
       llvm::SmallVector<NamedAttribute, 3> targetGemm1DimAttr{
           b.getNamedAttr("upper_layer_dimensions",
                          b.getArrayAttr({b.getI32IntegerAttr(1)})),
           b.getNamedAttr("upper_layer_names", b.getArrayAttr({b.getStringAttr(
-                                                  arg0TargetLayoutName1)}))};
+                                                  arg0TargetLayoutName[1])}))};
```
krzysz00: Nice little bit of cleanup here!

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -1829,9 +1870,9 @@ struct Conv2DRewritePattern : public OpRewritePattern<T> {
     // set upper_layer_layout attribute.
     transformedFilterAttrs.push_back(b.getNamedAttr(
         "upper_layer_layout",
-        b.getArrayAttr({b.getStringAttr(arg0TargetLayoutName0),
-                        b.getStringAttr(arg0TargetLayoutName1),
-                        b.getStringAttr(arg0TargetLayoutName2)})));
+        b.getArrayAttr({b.getStringAttr(arg0TargetLayoutName[0]),
```
krzysz00: Could this instead be `b.getStrArrayAttr()`?

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -2906,16 +3045,113 @@ struct Conv2DRewritePattern : public OpRewritePattern<T> {
       // input padding end
     }
 
+    // KPack for input tensor.
```
krzysz00, line 1879: Note to our future selves: there might be refactoring possibilities here.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -5079,18 +5331,38 @@ void affixBlockwiseCopyAttributes(
...
   bop->setAttr("block_size", gop->getAttr("block_size"));
 
-  bop->setAttr("source_dim_access_order", b.getArrayAttr({
+  if (KPack > 1) {
+    bop->setAttr("source_dim_access_order", b.getArrayAttr({
+                                                b.getI32IntegerAttr(0),
+                                                b.getI32IntegerAttr(1),
+                                                b.getI32IntegerAttr(2),
+                                                b.getI32IntegerAttr(3),
+                                            }));
```
krzysz00: `b.getI32ArrayAttr({0, 1, 2 , 3})`

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -5390,85 +5672,152 @@ struct GridwiseGemmRewritePattern : public OpRewritePattern<miopen::GridwiseGemm
...
+            GemmABlockCopyThreadSliceLengths_GemmK;
+      }
       break;
     case GemmMorN:
-      GemmABlockCopyThreadSliceLengths_GemmM = matrix_a_source_data_per_read;
-      GemmABlockCopyThreadSliceLengths_GemmK =
-          GemmABlockCopyNumberDataPerThread /
-          GemmABlockCopyThreadSliceLengths_GemmM;
+      // TBD: FIXME. Review logic here.
+      if (KPack > 1) {
```
krzysz00, line 2522: Something doesn't feel right about the logic here. Can you talk me through what these variables do, how they're computed, and how they're affected by kpack?

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -5484,28 +5833,110 @@ struct GridwiseGemmRewritePattern : public OpRewritePattern<miopen::GridwiseGemm
...
+
+    Value GemmABlockCopyDestCoord_Z_i32;
+    if (KPack > 1) {
+      GemmABlockCopyDestCoord_Z_i32 = b.create<AddIOp>(
+          loc, zeroConstantI32Op, GemmAThreadDataIdBegin_Z_i32);
+    }
+    GemmABlockCopyDestCoord_Y_i32 =
         b.create<AddIOp>(loc, zeroConstantI32Op, GemmAThreadDataIdBegin_Y_i32);
-    auto GemmABlockCopyDestCoord_X_i32 =
+    GemmABlockCopyDestCoord_X_i32 =
```
krzysz00: Can we get rid of these +0 ops? They'll make the code clearer and might enable more optimizations (see #440).

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -5613,9 +6143,15 @@ struct GridwiseGemmRewritePattern : public OpRewritePattern<miopen::GridwiseGemm
         b.create<miopen::GpuAllocOp>(loc, threadCRegisterMemRefType);
 
     // Determine vector / scalar load type for Matrix A / B.
-    SmallVector<uint64_t, 3> blockwiseCopyABounds = {
-        1, GemmABlockCopyThreadSliceLengths_GemmK,
-        GemmABlockCopyThreadSliceLengths_GemmM};
+    SmallVector<uint64_t, 3> blockwiseCopyABounds;
```
krzysz00: Let's update this to 4 to prevent heap allocations

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -5634,17 +6172,23 @@ struct GridwiseGemmRewritePattern : public OpRewritePattern<miopen::GridwiseGemm
...
     // llvm::errs() << "vector load dim: " << blockwiseAVectorDim << "\n";
     // llvm::errs() << "element type: " << blockwiseLoadAType << "\n";
     // llvm::errs() << "load size: " << blockwiseLoadAVectorLength << "\n";
     // llvm::errs() << "store size: " << blockwiseStoreAVectorLength << "\n";
 
-    SmallVector<uint64_t, 3> blockwiseCopyBBounds = {
-        1, GemmBBlockCopyThreadSliceLengths_GemmK,
-        GemmBBlockCopyThreadSliceLengths_GemmN};
-    TypeRange blockwiseLoadBType;
+    SmallVector<uint64_t, 3> blockwiseCopyBBounds;
```
krzysz00: Same size update here

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -5676,58 +6222,126 @@ struct GridwiseGemmRewritePattern : public OpRewritePattern<miopen::GridwiseGemm
...
-    auto blockwiseCopyVectorCoordType =
-        VectorType::get({3}, b.getIntegerType(32));
+    Type blockwiseCopyVectorCoordType;
+    if (KPack > 1) {
+      blockwiseCopyVectorCoordType = VectorType::get({4}, b.getIntegerType(32));
+    } else {
+      blockwiseCopyVectorCoordType = VectorType::get({3}, b.getIntegerType(32));
+    }
 
     // Matrix A: {0, 0, m_block_data_on_global}, {0, 0, 0}
```
krzysz00: Let's update this comment for kpack

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -5849,19 +6463,28 @@ struct GridwiseGemmRewritePattern : public OpRewritePattern<miopen::GridwiseGemm
...
 
     // Blockwise copy from global (generic tensor) to register (naive tensor).
-    Value blockwiseCopyASrcVectorUpdated = lb.create<miopen::MovePosV2Op>(
-        loc, blockwiseCopyVectorCoordType, loopOp.getRegionIterArgs()[0],
-        ValueRange{zeroConstantI32Op, KPerBlockConstantI32Op,
-                   zeroConstantI32Op});
+
+    Value blockwiseCopyASrcVectorUpdated;
+    if (KPack > 1) {
+      blockwiseCopyASrcVectorUpdated = lb.create<miopen::MovePosV2Op>(
```
krzysz00: This entire section of the change'll get clobbered by #425.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -6447,79 +7055,152 @@ struct GridwiseGemmV2RewritePattern
 
     // Compute ThreadSliceLengths for Matrix A.
     uint64_t GemmABlockCopyNumberDataPerThread =
-        MPerBlock * KPerBlock / BlockSize;
+        MPerBlock * KPerBlock * KPack / BlockSize;
```
krzysz00, line 3525: Let's put some parentheses around the multiplication to make sure we don't get messed up by `KPack / BlockSize` ending up as part of the computation (for all I know, it might)

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -6447,79 +7055,152 @@ struct GridwiseGemmV2RewritePattern
...
+            GemmBBlockCopyThreadSliceLengths_GemmK;
+      }
       break;
     case GemmMorN:
-      GemmBBlockCopyThreadSliceLengths_GemmN = matrix_b_source_data_per_read;
-      GemmBBlockCopyThreadSliceLengths_GemmK =
-          GemmBBlockCopyNumberDataPerThread /
-          GemmBBlockCopyThreadSliceLengths_GemmN;
+      // TBD: FIXME. Review logic here.
+      if (KPack > 1) {
```
krzysz00, line 3624: As a general thought, would anything break (especially given #440) if, instead of all these `if KPack > 1` checks, we just always had a KPack dimension that was sometimes fixed to 1?

krzysz00: (and yeah, I know that applies to #441's approach to the output tensor as well)

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -6565,28 +7327,108 @@ struct GridwiseGemmV2RewritePattern
...
+    Value GemmBThreadDataIdBegin_Y;
+    Value GemmBThreadDataIdBegin_X;
+
+    Value GemmBBlockCopyThreadClusterId_Z;
+    Value GemmBThreadDataIdBegin_Z;
+
+    if (KPack > 1) {
+      GemmBBlockCopyThreadClusterId_Z = b.create<SignedDivIOp>(
+          loc,
+          b.create<SignedDivIOp>(loc, tid,
```
krzysz00: Signed division? Neither of these can be negative, let's go unsigned instead

krzysz00: Heck, that goes for all the rest of these computations.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -6565,28 +7327,108 @@ struct GridwiseGemmV2RewritePattern
...
-    auto GemmBBlockCopySourceCoord_X_i32 = b.create<AddIOp>(
-        loc, n_block_data_on_global_i32, GemmBThreadDataIdBegin_X_i32);
-    auto GemmBBlockCopyDestCoord_Y_i32 =
+    Value GemmBBlockCopySourceCoord_Y_i32;
+    Value GemmBBlockCopySourceCoord_X_i32;
+
+    Value GemmBBlockCopySourceCoord_Z_i32;
+    if (KPack > 1) {
+      GemmBBlockCopySourceCoord_Z_i32 = b.create<AddIOp>(
+          loc, zeroConstantI32Op, GemmBThreadDataIdBegin_Z_i32);
```
krzysz00: Same as above with adding 0

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -8232,45 +9260,97 @@ struct ThreadwiseGemmRewritePattern
...
+      else
+        mul = b.create<MulFOp>(loc, dataType, gemmAKM, gemmBKN);
+      SmallVector<Value, 3> memIndicesMN;
+      extractForInductionVars({loopG, loopM, loopN}, &memIndicesMN);
+      auto gemmCMN = b.create<AffineLoadOp>(loc, gemmC, memIndicesMN);
+
+      Value add;
+      if (dataType.isa<IntegerType>())
+        add = b.create<AddIOp>(loc, dataType, mul, gemmCMN);
+      else
```
krzysz00: Braces on all the if statements, please? It makes adding debugging info easier, and I'm tempted to ask we add a lint forbidding this sort of thing going forward.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -8416,6 +9496,7 @@ struct ThreadwiseCopyRewritePattern
       ArrayAttr layeredSourceTransformMetadata;
       ArrayAttr layeredDestTransformMetadata;
 
+      legacyLoad = true;
```
krzysz00: Please explain this

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -9883,6 +10974,11 @@ struct XdlopsGemmV2RewritePattern
                   loc, ilnkb.create<MulIOp>(loc, ilnkiv, NConstantOp), laneId),
               ilnkb.create<MulIOp>(loc, NPerXdlopsConstantOp, olniv)));
 
+      if (KPack > 1)
```
krzysz00, line 6622: Why's this multiplication called for?

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -9911,48 +11015,133 @@ struct XdlopsGemmV2RewritePattern
...
+        Value offset =
+            innerLoopb.create<MulIOp>(loc, innerLoopiv, KBaseConstantOp);
+        if (bufferAElementType.isa<VectorType>()) {
+          // bufferA/BElement loaded on LDS are vectors.
+          // argA/B to be supplied to MFMA XDLOPS are also vectors.
+          assert(bufferAElementType.isa<VectorType>());
+          assert(bufferBElementType.isa<VectorType>());
+
+          argA = innerLoopb.create<SplatOp>(loc, zeroOp, argType);
+          argB = innerLoopb.create<SplatOp>(loc, zeroOp, argType);
```
krzysz00, line 6731: We have zero constant creation functions in #425 (and there's a small fix to them in #441 I might want to backport related to bfloats) - FYI for your merge/rebase

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -10010,8 +11199,12 @@ struct XdlopsGemmV2RewritePattern
       else
         sourceOffsetA.push_back(sourceOffsetBeforeTransformA);
 
-      auto valueA = lklb.create<memref::LoadOp>(loc, dataType, op.matrixA(),
-                                        ValueRange{sourceOffsetA});
+      Value valueA;
+      if (KPack > 1) {
+        valueA = emitLoadLogic(lklb, loc, op.matrixA().getType().template cast<MemRefType>(), argType, false, {}, op.matrixA(), ValueRange{sourceOffsetA});
```
krzysz00: General comment: is it possible to statically compute the operations emitLoadLogic will lead to and apply those transformations to sourceOffsetA?

krzysz00 (commented):
Ok, so, this PR is pretty noisy due to how copy-pasted our lowering code is, so it's hard to give high-level feedback, but I've left some style notes, a few questions, and notes on things that'll get clearer (or at least messed up by) #425  (which really had rough timing).

The logic you marked with a TBD confused me as well - it felt like the divisions were in the wrong order, so more words about it would be nice.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
whchung: This logic has been eliminated in the latest logic.

whchung:
jenkins: retest this please.

whchung:
jenkins: retest this please.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
whchung: done

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
whchung: the computation depends on `waveOffsetA`, which is derived from the current workitem ID, so it's not statically determined.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
whchung: With KPack, we are moving along LDS with stride in the multiples of KPack.

whchung:
jenkins: retest this please.

whchung:
Addded @yiqian1 as a reviewer. It has been identified with @jerryyin and I that existing KPACK logic would introduce butterfly effect that the FASTEST kernel coming from the tuning parameter search space currently used by MIOpen MLIR solver would result in INCORRECT kernel. On the other hand, the kernel is working fine if using hand-picked heuristic tuning parameter currently employed in the branch.

Before the root cause is yet known, disabling tuning for now.

yiqian1:
> Addded @yiqian1 as a reviewer. It has been identified with @jerryyin and I that existing KPACK logic would introduce butterfly effect that the FASTEST kernel coming from the tuning parameter search space currently used by MIOpen MLIR solver would result in INCORRECT kernel. On the other hand, the kernel is working fine if using hand-picked heuristic tuning parameter currently employed in the branch.
> 
> Before the root cause is yet known, disabling tuning for now.

I'll disable the weekly CI once this PR is merged. It doesn't make sense to run weeklies with tuning disabled.

krzysz00:
@whchung I'd prefer that we debug why KPACK doesn't work with all tuning parameters or to add a flag (like my `enableOutSwizzles`) that disables KPACK when the tuning parameters cannot accomodate KPACK

external/llvm-project/mlir/.clang-format
```diff
@@ -1,2 +1,2 @@
 BasedOnStyle: LLVM
-AlwaysBreakTemplateDeclarations: Yes
\ No newline at end of file
+AlwaysBreakTemplateDeclarations: true
```
krzysz00: Why'd we make this change?

whchung:
@krzysz00 The bug is in several different layers beyond the scope of MIOpen dialect so it's a bit nasty. I'm still in the process to figure out where in the parameter search space triggered the issues.

- vector dialect lowering : complete legit `vector.transfer_read` op ended up crashing at lowering to LLVM dialect. This is what causing weekly CI to break.
- backend : complete legit tuning parameters ending up with incorrect kernel. I forced setting KPACK to be all 1 from MIOpen and started tuning. And even with that we get incorrect kernels, and unfortunately that incorrect one happens to be the fastest one.

krzysz00:
@whchung Yowch. Good luck and let me know if you need a sounding board!

whchung:
Now if I could get rid of the vector -> llvm lowering failure we could have a passing weekly CI...

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -5265,7 +5517,7 @@ struct GridwiseGemmRewritePattern : public OpRewritePattern<miopen::GridwiseGemm
         max_lds_align *
         math_util::integer_divide_ceil<int64_t>(MPerBlock, max_lds_align);
     a_block_space = math_util::integer_least_multiple(KPerBlock * AlignedMPerBlock,
-                                                 max_lds_align);
+                                                 max_lds_align) * KPack;
```
asroy, line 2302: This is the C++ logic for A/B tensor in LDS with [KPerBlock, MPerBlock, KPack] layout

https://github.com/ROCmSoftwarePlatform/composable_kernel/blob/a4f24233e51854c4b5cb7d75637fa0f235f78f8e/composable_kernel/include/tensor_operation/gridwise_gemm_xdlops_v2r3.hpp#L409-L439

asroy, line 2302: This is the C++ logic for A/B tensor in LDS with [KPerBlock, MPerBlock, KPack] layout

https://github.com/ROCmSoftwarePlatform/composable_kernel/blob/a4f24233e51854c4b5cb7d75637fa0f235f78f8e/composable_kernel/include/tensor_operation/gridwise_gemm_xdlops_v2r3.hpp#L409-L439

whchung:
> @krzysz00 The bug is in several different layers beyond the scope of MIOpen dialect so it's a bit nasty. I'm still in the process to figure out where in the parameter search space triggered the issues.
> 
> * vector dialect lowering : complete legit `vector.transfer_read` op ended up crashing at lowering to LLVM dialect. This is what causing weekly CI to break.

Turns out we may be able to get rid of `vector.transfer_read` completely. Testing this assumption with CI.

> * backend : complete legit tuning parameters ending up with incorrect kernel. I forced setting KPACK to be all 1 from MIOpen and started tuning. And even with that we get incorrect kernels, and unfortunately that incorrect one happens to be the fastest one.

Will address this after we can confirm it's possible to get rid of all all `vector.transfer_read` ops. Also this PR has been merged with the latest `miopen-dialect` which has @pcf000 's upstream merge.

mlir/utils/jenkins/miopen-tests/miopen_validate.sh
```diff
@@ -136,6 +136,7 @@ function clean_miopen_caches() {
 function setup_environment() {
     export MIOPEN_FIND_MODE=1
     export MIOPEN_DRIVER_USE_GPU_REFERENCE=1
+    export MIOPEN_COMPILE_PARALLEL_LEVEL=1
```
jerryyin: This needs to be reverted for final merge: MIOPEN_COMPILE_PARALLEL_LEVEL=1 is a debugging flag (that alters tuning behavior) and should not be used in production.

whchung: This was added in 91e318b as a measure to get rid of errors encountered in the tuning process. Will keep an eye on it.

whchung:
jenkins: retest this please.

whchung:
Per discussion with @jerryyin , identified a temporarily measure to skip tuning for backward data and backward weight for fp16 data type. Prior to the latest LLVM upgrade, they could be tuned, but now the fastest kernel after tuning would yield incorrect results.

Currently:

- fp32 : fwd / bwd / wrw could all be tuned and yield correct results.
- fp16 : fwd could be tuned and yield correct results.
- fp16 bwd / wrw tuning has been disabled, and rely on the built-in handpicked tuning parameter set in MLIR MIOpen dialect.

krzysz00:
@whchung Can we get a performance comparison on the kernels that have had their tuning disabled to show that kpack being enabled leads to better performance than not having kpack and allowing for tuning?

I'm not comfortable merging this code without, at the very least, data showing that disabling tuning for broad classes of kernels is worth it.

whchung:
> @whchung Can we get a performance comparison on the kernels that have had their tuning disabled to show that kpack being enabled leads to better performance than not having kpack and allowing for tuning?
> 
> I'm not comfortable merging this code without, at the very least, data showing that disabling tuning for broad classes of kernels is worth it.

It is yet known which part caused the fastest kernel in fp16 backward (data / weight) convolution become incorrect : tuning search space / dialect implementation / LLVM core / LLVM backend.

The latest commit is a step to try obtain a full execution of nightly + weekly CI so we can get performance comparisons.

mlir/utils/jenkins/miopen-tests/miopen_validate.sh
whchung: Per discussion with @jerryyin , another temporarily measure is to modify the semantic of `miirLowerTuningParams`, which is used by `IsApplicable` in MIOpen MLIR solvers to determine if a tuning parameter is valid, so that `miirLowerTuningParams` would behave exactly as `miirLowerBin` for the whole lowering pipeline.

The rationale why `MIOPEN_COMPILE_PARALLEL_LEVEL=1` was introduced was because it was identified that:

1) without the environment variable, the tuning process would crash upon receiving a failure in getting a kernel binary blob from `miirLowerbin`. 
2) `miirLowerTuningParams` was invoked prior to `miirLowerBin` to help determine whether a tuning parameter is valid. It does NOT run through the whole lowering pipeline.
3) Should anything bad happens in the later part of the lowering pipeline, `miirLowerTuningParams` couldn't catch that.
4) MIOpen, as a client of `miirLowerTuningParams`, upon receiving the a success return value from the function, would assume `miirLowerBin` would definitely yield a working kernel, which is NOT always the case. Something bad might happen in the later part of the lowering pipeline. It could be in MIOpen dialect, could be in LLVM core, could be in LLVM AMDGPU backend.
5) The exception handling behavior inside MIOpen when `MIOPEN_COMPILE_PARALLEL_LEVEL=1` is NOT set is UNDEFINED. And currently it would just crash the tuning process.

krzysz00:
@whchung You don't need a full nightly/weekly run for performance comparisons - you can manually call the benchmarking scripts on 112 to get results for just the kernels of interests and generate the plots.

Also, re the paralellism level, wouldn't it make more sense to,  in `miirLowerBin` and its friends, have a thing that'll catch exceptions (and _maybe_ signals, though we could just compile with assertions off for release and be under the assumption that we won't segfault) and return failure if an exception occurs? I'm pretty sure that's a typical pattern when wrapping C++ for C?

whchung:
> @whchung You don't need a full nightly/weekly run for performance comparisons - you can manually call the benchmarking scripts on 112 to get results for just the kernels of interests and generate the plots.

Thanks. Will do. Also I still want a full successful CI run to help validate the PR.

> Also, re the paralellism level, wouldn't it make more sense to, in `miirLowerBin` and its friends, have a thing that'll catch exceptions (and _maybe_ signals, though we could just compile with assertions off for release and be under the assumption that we won't segfault) and return failure if an exception occurs? I'm pretty sure that's a typical pattern when wrapping C++ for C?

The crash is not happening in MLIR, but inside MIOpen, which we have less control. In MLIR side within `miirLowerBin` we already properly catch the failures and return the result to MIOpen.

krzysz00:
@whchung Ok, so, to make sure I understand correctly: if compilation fails, either in our code or in LLVM, `miirLowerBin()` will correctly return a failure result to MIOpen and will not crash the process. However, MIOpen assumes that `miirLowerBin()` doesn't fail, and then crashes when it encounters that failure result.

If that's a good summary, then I think we need to some combination of
1. Send a PR to MIOpen to have them check for miirLowerBin() failures and not ship anything that could cause such failures until after they pull that change in
2. Identify ways to stop ourselves failing after we've accepted a set of tuning parameters - this is hard and requires being really careful about what the tuning parameters are and how to fall back to slower algorithms (see, for example, my swizzle code, which has this massive list of preconditions before you can start using swizzles and falls back when they're not available)

whchung:
> @whchung Ok, so, to make sure I understand correctly: if compilation fails, either in our code or in LLVM, `miirLowerBin()` will correctly return a failure result to MIOpen and will not crash the process. However, MIOpen assumes that `miirLowerBin()` doesn't fail, and then crashes when it encounters that failure result.
> 
> If that's a good summary, then I think we need to some combination of
> 
> 1. Send a PR to MIOpen to have them check for miirLowerBin() failures and not ship anything that could cause such failures until after they pull that change in
> 2. Identify ways to stop ourselves failing after we've accepted a set of tuning parameters - this is hard and requires being really careful about what the tuning parameters are and how to fall back to slower algorithms (see, for example, my swizzle code, which has this massive list of preconditions before you can start using swizzles and falls back when they're not available)

@jerryyin What's your take on 1? You have more experiences working on MIOpen and should know better about the impact of this proposed change.

jerryyin:
> The crash is not happening in MLIR, but inside MIOpen, which we have less control. In MLIR side within miirLowerBin we already properly catch the failures and return the result to MIOpen.

I'm not fully convinced that it is MIOpen that crashed due to MLIR's failure. To accurately determine which component does crash I need to rerun the attach a debugger to make sure. However, I do think that having 20 threads all or majority of which all failing the MLIR compilation would push MIOpen to its extreme in handling a condition it is not prepared for. MIOpen assumes that once something is marked as applicable, it must not fail.

> Send a PR to MIOpen to have them check for miirLowerBin() failures and not ship anything that could cause such failures until after they pull that change in

That's on us because MIOpen get the applicability check result from our API `miirLowerTuningParams`. So this basically reduce to that we do the full compilation in the applicability check. This is fine only as a temporary measure but shall not persist as our permanent solution. We still have to figure out exactly what caused these issues.

krzysz00:
@jerryyin To clarify, the general contract of MIOpen is that, if a solver declares that it is applicable for a particular configuration (problem size + set of tuning parameters) it must always successfully generate a kernel for that configuration, and that, if it somehow fails to generate the kernel, things will explode and it's a bug in the solver, right? Or, equivalently, is `isApplicable(config) == true` implying `bin = generateKernel(config); [bin is a runnable binary]`  a core assumption of MIOpen that they can't/won't change?

If so, then I claim we must not merge KPack until we can refine the applicability checks so that kernel generation below the applicability checks never crashes.

If that isn't a core contract of theirs, then we should
a) If we're the crash source, add a better crash-catching shell
b) If MIOpen's the crash source, improve their error handling

jerryyin:
> if it somehow fails to generate the kernel, things will explode and it's a bug in the solver, right? 

Correct, it is clearly communicated as it is the core assumption MIOpen's architecture relied on. It is written in their spec document. I can have Artem to send over that if necessary.

> Or, equivalently, is isApplicable(config) == true implying bin = generateKernel(config); [bin is a runnable binary] a core assumption of MIOpen that they can't/won't change?

Also correct, changing this requires some architectural modifications in MIOpen, unless we have a good reason to make MIOpen to tolerate otherwise. But such request will need a much broader discussion includes why the applicability check contract cannot work for us.

krzysz00:
@jerryyin Makes sense, that's what I thought, thanks for the info!

mlir/utils/jenkins/miopen-tests/miopen_validate.sh
whchung: Commit 41b0a81 incorporate this temporary measure.

whchung: I have just reverted the commit. Empirical testing shows the crash would still occur. And indeed it's inside MIOpen, albeit in a different place.

whchung: I'll keep `MIOPEN_COMPILE_PARALLEL_LEVEL=1` in the test script for now.

After getting a deeper understanding of the other issue (tuning for fp16 backward data / weight) then revisit this topic.

whchung:
jenkins: retest this please.

whchung:
The workaround introduced in commit 0003add9 could get us through the "Tune MLIR kernels" stage in build 68.

Let me run the performance metric report and compare with the mainline.

Also it's needed to understand why the fastest kernel in fp16 backward data/weight result in incorrect kernels, and decide:
- whether to alter the tuning search space.
- identify/fix potential bugs inside dialect implementation.
- identify/avoid/fix potential bugs inside LLVM (core or AMDGPU backend).

mlir/utils/jenkins/miopen-tests/miopen_validate.sh
jerryyin: Thanks for confirming. As you want to proceed with merging and revisit those topics anyways I think you can just disable fp16 tuning in general as a temporary measure (Either in bash script or in Jenkinsfile). Otherwise we will be under trial-and-error mode for the code path untested (and probably shouldn't be there). Also, if you decide to disable fp16 tuning, please do file ticket to document all issues in independent tickets. This way we have a way to track those issues. 

The concern I had with all the temporary measures that got to merge with this PR is that it may another implicit deadline: the next ROCm release. On the other hand, if you keep the kpack in a separate branch then it will not impact anything in the mainline. Compared to the two, I actually think keeping it in the separate branch will give us more flexibility as it is unknown how much effort is needed to root cause all those issues.

whchung: In the latest implementation, this hack is no longer necessary.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -1647,13 +1661,61 @@ template <typename T> struct Conv2DRewritePattern : public OpRewritePattern<T> {
...
+                                           "gemmK",
+                                           {gemmKLength / KPack, KPack});
+        }
+        TransformMapAttr kpackGemmFilterTransformAttr =
+            kpackGemmFilterTransform.get();
+        gemmFilterKPack = b.create<miopen::TransformOp>(
+            loc, gemmFilterPad, kpackGemmFilterTransformAttr);
+      }
+    }
+
```
krzysz00, line 1701: I think this logic can be simplified a lot.
You can, at the very least, go `kpackGemmFilterTransform.passThrough(padGemmFilterTransform.endName(0))` and so on, and otherwise use `endName()` to not have to branch on whether we padded K or not.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -1827,6 +1894,55 @@ template <typename T> struct Conv2DRewritePattern : public OpRewritePattern<T> {
...
+                                          "gemmK",
+                                          {gemmKLength / KPack, KPack});
+        }
+        TransformMapAttr kpackGemmInputTransformAttr =
+            kpackGemmInputTransform.get();
+        gemmInputKPack = b.create<miopen::TransformOp>(
+            loc, gemmInputPad, kpackGemmInputTransformAttr);
+      }
+    }
+
```
krzysz00, line 1906: Same notes re there being ways to avoid this if nest

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
krzysz00, line 1701: I think there's a much simpler way to do this, using `endName` on `padGemmFilterTransform` (which might just be `gemmFilterTransform`). That way, you can grab the relevant dimension names, whether they're "gemmXPad" or "gemmX" and don't have to have this pile of branches that doesn't add much.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
krzysz00, line 1906: Same comment as wrt filter.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -2390,85 +2525,152 @@ struct GridwiseGemmRewritePattern : public OpRewritePattern<miopen::GridwiseGemm
...
+        GemmBBlockCopyThreadSliceLengths_GemmK =
+            KPack / matrix_b_source_data_per_read;
+        GemmBBlockCopyThreadSliceLengths_GemmN =
+            GemmBBlockCopyNumberDataPerThread / KPack;
+      } else {
+        GemmBBlockCopyThreadSliceLengths_GemmK = matrix_b_source_data_per_read;
+        GemmBBlockCopyThreadSliceLengths_GemmN =
+            GemmBBlockCopyNumberDataPerThread /
+            GemmBBlockCopyThreadSliceLengths_GemmK;
+      }
```
krzysz00, line 2589: As a general question ... what would happen if, instead of having all this `if (kPack > 1)` stuff, you just always added a kPack dimension that was frequently length 1? Would we expect that to lead to worse ISA?

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -5787,17 +6615,23 @@ struct XdlopsGemmV2RewritePattern
 
       auto destOffsetA = ilmkb.create<AddIOp>(loc, ilmkiv, kOffsetA);
 
-      auto valueA =
-          ilmkb.create<memref::LoadOp>(loc, dataType, op.matrixA(), sourceOffsetA);
+      Value valueA;
+      if (KPack > 1) {
+        valueA = emitLoadLogic(
+            ilmkb, loc, op.matrixA().getType().template cast<MemRefType>(),
```
krzysz00, line 6586: Couldn't this always be emitLoadLogic and it'll drop down to the memref.load in the scalar case?

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
```diff
@@ -911,6 +928,35 @@ class PopulateParamsXDL : public PopulateParamsBase {
...
+    }
+
+    // XXX FIXME: Ignore KReduction XDLOPS path for forward convolution now.
+    // These M/NPerBlock combinations will result in lowering errors at tuning.
+    if (ctx.getOpType() == miopen::ConvOpType::Fwd) {
+      if ((param.gemmMPerBlock == 16 || param.gemmMPerBlock == 32 ||
+           param.gemmMPerBlock == 64) &&
+          (param.gemmNPerBlock == 16 || param.gemmNPerBlock == 32 ||
+           param.gemmNPerBlock == 64)) {
+        return failure();
```
krzysz00, line 949: Are these perhaps victims of https://ontrack-internal.amd.com/browse/SWDEV-318900

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
```diff
@@ -57,11 +57,122 @@ void AffixTuningParameters::affixTuningParametersImpl(T &op) {
...
+  // Get shape of filter tensor.
+  auto filterType = op.filter().getType().template cast<MemRefType>();
+  auto filterShape = filterType.getShape();
+
+  // Get shape of input tensor.
+  auto inputType = op.input().getType().template cast<MemRefType>();
+  auto inputShape = inputType.getShape();
+
+  // Get shape of output tensor.
+  auto outputType = op.output().getType().template cast<MemRefType>();
```
krzysz00, line 76: With how many times we've duplicated this bit in our codebase ... let's have a future PR where we pull it out to some common utility?

krzysz00 (commented):
(I'm hoping refreshing mid-review didn't leave duplicate comments)

So, overall, if this works, it looks good.

One thing I'd like to see cleaned up a bit is how the KPack transform is created, I left comments above on that.

The other main observation, which I'd say doesn't block the PR but would make a plausible followup ticket - unless we want to do it now ... is that we've got all this branching on whether we have KPack. Maybe the right thing to do would be to always have the KPack dimension but set it to length 1 most of the time.

external/llvm-project/mlir/.clang-format
whchung: Some places in LLVM upstream use `Yes`, and some use `true`. I seem to bump into some issues locally awhile so I introduced this line. Will remove it.

krzysz00: We can probably keep it - interestingly, upstream doesn't seem to set this value at all?
I just was trying to understand what the motivation here was.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
whchung: We don't know that yet. It could be addressed in a future PR to simplify the logic.

krzysz00: Fair enough, future PR it is - this wasn't a critical complaint, just an observation. Do we want a ticket for the investigation?

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
whchung: `endName()` API is not used anywhere in the implementation so I didn't use it. Now it's used in the latest logic.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
whchung: the logic has been simplified.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
whchung: I guess it's possible. But I don't want to change this until a full passing CI. I'd do it in a later PR to further clean this up.

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
whchung: defintely.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
whchung: logic here has been simplified.

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
whchung: this deserves further investigation. Currently it doesn't appear to be the case.

krzysz00:
Overall, this all looks good to me - or at least it looks like we have some cleanup items for later but nothing egregious.

I'd poke someone else to check things over once more, but I'm willing to approve if you're sick of waiting.

krzysz00:
(I mean, the fact that we're going to town with disabling certain xdlops cases is concerning, but that could be a future-us issue.)

whchung:
> Overall, this all looks good to me - or at least it looks like we have some cleanup items for later but nothing egregious.
> 
> I'd poke someone else to check things over once more, but I'm willing to approve if you're sick of waiting.

Let's wait until we have a complete passing CI before actually doing so.
I have less confidence on how I use the APIs in `BottomUpCTBuilder` and `TransformMapAttr`.
Other than addressing code refactoring changes in PR #469, we'll also have to adopt changes introduced in PR #461 here.

krzysz00:
I can also take the time tomorrow to run this on MI-200 with my "hack around SWDEV-316900" patch applied to get that angle as well.

whchung:
@asroy / @jerryyin / @krzysz00 now finally we have an all-passing CI run.

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
```diff
@@ -911,6 +928,35 @@ class PopulateParamsXDL : public PopulateParamsBase {
     if ((param.gemmNPerBlock % param.gemmNPerWave) != 0)
       return failure();
 
+    if ((param.gemmKPerBlock % param.gemmKPack) != 0)
+      return failure();
+
+    // Reject too wide KPACK values for fp32/fp16/bf16 types.
+    auto dataType = ctx.getDataType();
+    if (dataType.isF32() && param.gemmKPack >= 8) {
```
krzysz00, line 936: Shouldn't this be `> 4`?

mlir/lib/Dialect/MIOpen/Transforms/LowerMIOpenOps.cpp
```diff
@@ -221,7 +221,6 @@ LogicalResult backwardWeightAtomicAdd(miopen::Conv2DBwdWeightOp op,
     outputNames.push_back(outputAttr.getValue());
 
     if (filterAttr.getValue() == "g") {
-      g = filterShape[i];
```
krzysz00, line 224: Why leave the empty if statement?

mlir/test/Dialect/MIOpen/lowering_blockwise_gemm_v2.mlir
```diff
@@ -1,18 +1,24 @@
-// XFAIL: *
 // RUN: miopen-opt -miopen-lowering-step3 %s | FileCheck %s
 
-func @miopen_blockwise_gemm_v2_two_results(%matrixA : memref<12288xf32, 3>, %matrixB : memref<12288xf32, 3>) -> (vector<32xf32>, vector<32xf32>) {
+#map = affine_map<(d0) -> (d0 + 512)>
+
+func @miopen_blockwise_gemm_v2_two_results(%matrixA : memref<512xf32, 3>, %matrixB : memref<512xf32, #map, 3>, %bufferA : memref<2xvector<2xf32>, 5>, %bufferB : memref<2xvector<2xf32>, 5>) -> (vector<32xf32>, vector<32xf32>) {
```
krzysz00, line 5: Re this test: I don't think this'll properly exercise the function we care about. To match how `blockwise_gemm_v2` gets invoked, you'd want to pass in a `Slice` transform_map with `#map` as its backing affine map

mlir/test/Dialect/MIOpen/lowering_xdlops_gemm_v2.mlir
```diff
@@ -1,18 +1,25 @@
...
+    k = 2 : i32,
+    kpack = 2 : i32,
+    m = 128 : i32,
+    m_per_wave = 64 : i32,
+    m_waves = 2 : i32,
+    n = 128 : i32,
+    n_per_wave = 64 : i32,
+    n_waves = 2 : i32,
+    transforms = [[], []]
+  } : memref<512xf32, 3>, memref<512xf32, #map, 3>, index, index, memref<2xvector<2xf32>, 5>, memref<2xvector<2xf32>, 5>, vector<32xf32>, vector<32xf32> -> vector<32xf32>, vector<32xf32>
```
krzysz00, line 23: Same comment re raw affine maps on the memref

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -1650,13 +1665,40 @@ template <typename T> struct Conv2DRewritePattern : public OpRewritePattern<T> {
...
+      Value source = (isFilterPad) ? gemmFilterPad : gemmFilter;
+
+      BottomUpCTBuilder kpackGemmFilterTransform =
+          BottomUpCTBuilder::above(sourceTransform, sourceTransformAttr);
+      // Passthrough gemmG (dim 0) and gemmM (dim 2).
+      kpackGemmFilterTransform.passThrough(sourceTransform.endName(0));
+      kpackGemmFilterTransform.passThrough(sourceTransform.endName(2));
+      // Use Unmerge to split gemmK (dim 1) into gemmK and gemmKPack, place
+      // gemmKPack at dim 3.
+      int64_t gemmKLength = sourceTransform.endSize(sourceTransform.endName(1));
```
krzysz00: Nit: We could probably do `kName = sourceTransform.endName(1);` before this bit

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -1824,12 +1870,40 @@ template <typename T> struct Conv2DRewritePattern : public OpRewritePattern<T> {
...
+      Value source = (isInputPad) ? gemmInputPad : gemmInput;
+
+      BottomUpCTBuilder kpackGemmInputTransform =
+          BottomUpCTBuilder::above(sourceTransform, sourceTransformAttr);
+      // Passthrough gemmG (dim 0) and gemmN (dim 2).
+      kpackGemmInputTransform.passThrough(sourceTransform.endName(0));
+      kpackGemmInputTransform.passThrough(sourceTransform.endName(2));
+      // Use Unmerge to split gemmK (dim 1) into gemmK and gemmKPack, place
+      // gemmKPack at dim 3.
+      int64_t gemmKLength = sourceTransform.endSize(sourceTransform.endName(1));
```
krzysz00: Did I not add an `endSize()` that takes an index` making this `sourceTransform.endSize(1);`? If not ... whoops.

krzysz00 (approved):
Overall, this looks like a quite solid first pass at kPack support.

I've left some minor nitpicks, but none of those are serious enough that I'd make them block merging this.

@whchung, since this PR has a lot of "TODO"/"add support for" sections, please create tickets for them. For example,
- non-xdlops kpack
- Adding kPack to backwards ops where possible (wrw seems like a solid candidate, backward data is rather agressively scalar in its memory reads)
- Potentially tightening up some of the logic so there's less weird copy-paste

Thanks for suffering through all my comments and for putting in all the work needed to get these massive speedups!

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
whchung: Will fix that in a separate cleanup PR.

mlir/lib/Dialect/MIOpen/Transforms/LowerMIOpenOps.cpp
whchung: Will fix that in a separate cleanup PR.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
whchung: Will fix that in a separate cleanup PR.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
whchung: There is indeed an `endSize(uint32_t dim)` interface. I was just use all names here. Will fix that in a separate cleanup PR.

mlir/test/Dialect/MIOpen/lowering_blockwise_gemm_v2.mlir
whchung: The test was marked XFAIL in `miopen-dialect` for some time. It was originally created to test if a `blockwise_gemm_v2` op could be lowered to `xdlops_gemm_v2` successfully. We can improve the test, and also `lowering_xdlops_gemm_v2.mlir` to match more closely to how we really use these ops. Will do that in a separate cleanup PR.

mlir/test/Dialect/MIOpen/lowering_xdlops_gemm_v2.mlir
whchung: Will do that in a separate cleanup PR.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
asroy: a_block_space = math_util::integer_least_multiple(KPerBlock * MPerBlock, max_lds_align) * KPack

same for b matrix

mlir/test/Dialect/MIOpen/lowering_blockwise_gemm_v2.mlir
whchung: This test is actually already reflecting how we use `blockwise_gemm_v2`. The second memref does carry an affine map which is the result of `Slice`.

mlir/test/Dialect/MIOpen/lowering_xdlops_gemm_v2.mlir
whchung: Same as comment above.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
whchung: @asroy . PR #490 .

PR #424, opened 2021-10-28
Misc tweaks.

whchung:
- Remove some unused variables.
- Use `Type` instead of `TypeRange` as other similar places all employ `Type`.


krzysz00 (approved):
LGTM!

PR #425, opened 2021-10-29
Clean up our emitted IR (and our code) in several ways

krzysz00:
1. Switch from scalar constants + splat to vector float constants

2. Change blockwise_{load,store,copy} to take 3 scalar arguments instead of a vector of 3 arguments. This reduces the amount of extractElement/insertElement we have running around.

2b. As a consequence of the above, simplify the blockwise gemm loop so it doesn't pass around non-constant values as iteration arguments (the old code had a bunch of cases where we were adding 0)

3. Since it's now unused, remove mov_pos_v2

4. Remove several unused variables and lambda captures


whchung:
This PR looks very promising and get rid of many legacy stuffs derived from C++-based implementation. I'll temporarily hold it until I sort things out a bit more in #423, and modify #423 so it adopts the logic behind this PR.

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -178,21 +178,6 @@ def MIOpen_FillOp:
   }];
 }
 
-def MIOpen_MovePosV2Op:
```
whchung, line 181: Nice job getting rid of this op.

krzysz00:
@whchung Wouldn't it make sense to merge this and then merge it into #423?

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -236,8 +221,8 @@ def MIOpen_BlockwiseCopyOp:
     MIOpen_Op<"blockwise_copy">,
     Arguments<(ins AnyMemRef:$source,
                    AnyMemRef:$dest,
-                   VectorOfLengthAndType<[3], [I32]>:$sourceCoordVector,
-                   VectorOfLengthAndType<[3], [I32]>:$destCoordVector,
+                   I32:$sourceCoord0, I32:$sourceCoord1, I32:$sourceCoord2,
```
whchung: On the other hand, I don't like this approach. In #423 I just changed `blockwise_copy` so it supports either 3D or 4D coordinates, depending on whether we split K dimension or not.

We might need to consider making these operands variadic.

whchung:
> @whchung Wouldn't it make sense to merge this and then merge it into #423?

#423 happens to address similar issues, although in a different fashion. So I don't want to merge this right away until I figure out how to cope with your proposed changes into #423.

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
krzysz00: .. Ah. In which case, I'll go and make it variadic. I went with 3 constants since the existing code always had 3 things.

Question: do the source and destination always have the same number of  coordinates?

whchung: Yes, the dimensions between source and destination should match.

krzysz00:
Makes total sense. Do let me know if tehre's anything about this that's incompatible with #423 in a serious way so I can adapt (like the variadic thing)

whchung:
> Makes total sense. Do let me know if tehre's anything about this that's incompatible with #423 in a serious way so I can adapt (like the variadic thing)

Right now the biggest hurdle is just making `blockwise_copy` to take `VectorOfLengthAndType<[3], [I32]>` to 3 `I32`. So should you make it be variadic I could follow suit in #423.

krzysz00:
@whchung I've swapped over the blockwise ops to take variadic arguments

whchung:
The PR is getting ready. I'll work with @zjing14 to address some dialect-level issues I identified in #423. And then apply techniques introduced in this PR in #423. And we'll merge this PR first. And after that merge #423 once other compiler backend-level issues are sorted out (or worked around). To save CI resources I'm applying `skip-ci` label for now.

whchung:
jenkins: retest this please.

whchung (approved):
LGTM. Waiting for CI to report test results.

PR #426, opened 2021-10-29
Hoist loop invariant computations out of loops in xdlops lowering

krzysz00:
I've identified several expressions that are re-computed in inner loops that don't depend on the loop variable in those loops. In principle, a loop invariant hoisting pass should get rid of these but, since we're not going through anything like a C compiler, I'd rather not rely on one being run.


whchung (approved):
LGTM!

PR #427, opened 2021-10-29
Use std::call_once to get rid of compile-time warning.

whchung:
This change is potentially big so I'm asking you guys to help review it.

### DIFF
--- a/mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
+++ b/mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
@@ -34,10 +34,11 @@ struct MiirHandle_s {
       return registry;
     };
     static MLIRContext context(getRegistry());
-    static bool once = []() {
+    static std::once_flag once;
+    std::call_once(once, []() {
       context.loadDialect<miopen::MIOpenDialect, StandardOpsDialect>();
       return true;
-    }();
+    });
     return context;
   }
   MiirHandle_s() {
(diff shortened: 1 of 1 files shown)


sjw36 (approved):
Nice!

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
```diff
@@ -34,10 +34,11 @@ struct MiirHandle_s {
       return registry;
     };
     static MLIRContext context(getRegistry());
-    static bool once = []() {
+    static std::once_flag once;
+    std::call_once(once, []() {
       context.loadDialect<miopen::MIOpenDialect, StandardOpsDialect>();
       return true;
```
jerryyin: Looks like the return is useless now so better get rid of it.

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
```diff
@@ -62,7 +63,8 @@ struct MiirHandle_s {
 // condition of one thread doing intialization and another doing
 // lowering.
 bool miirLazyInit() {
-  static const bool once = []() {
+  static bool once;
+  std::call_once(once, []() {
```
jerryyin: Same here:
 - Use std::once_flag
 - get rid of return

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
whchung: done

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
whchung: done

jerryyin (approved):
LGTM

PR #429, opened 2021-11-01
Do MIOpen validation tests and MLIR kernels tuning weekly

yiqian1:
- Add a "weekly" parameter to Jenkinsfile. The combination of "staticLib" and "weekly" does:

  Validate MIOpen Resnet50 with tuning and archive the user perfDb.
  Run MIopen tests without tuning.

- Reduce the matrix size of the "Test MIOpen config" to 2x2

- Add a "--tune-all" command line option to miopen_validate.sh which does tuning on all Resnet50 configs.

- Set Xdlops based on GPU type (via rocm_agent_enumerator), thus we can separate MIOpen tests on private CI and public CI.

A successful weekly CI is:
http://ml-ci.amd.com:21096/blue/organizations/jenkins/MLIR%2Ftest_weeklyCI/detail/test_weeklyCI/9/pipeline/276/

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -243,7 +243,10 @@ pipeline {
                         stage("Build MIOpen with libMLIRMIOpen") {
                             when { anyOf {
                                     equals expected: true, actual: params.nightly;
-                                    equals expected: true, actual: params.weekly;
+                                    allOf {
+                                        equals expected: true, actual: params.weekly;
+                                        equals expected: true, actual: params.staticLib;
+                                    }
                             }}
                             steps {
                                 dir('MIOpen') {
@@ -271,7 +274,10 @@ pipeline {
                         stage("Stash MIOpen with libMLIRMIOpen") {
                             when { anyOf {
                                     equals expected: true, actual: params.nightly;
-                                    equals expected: true, actual: params.weekly;
+                                    allOf {
+                                        equals expected: true, actual: params.weekly;
+                                        equals expected: true, actual: params.staticLib;
+                                    }
                             }}
                             steps {
                                 stash name:"MIOpen-MLIR-kernels",\
@@ -284,6 +290,7 @@ pipeline {
                         stage("Test MIOpen config with tuning") {
                             when {
                                     equals expected: true, actual: params.weekly;
+                                    equals expected: true, actual: params.staticLib;
                             }
                             environment {
                                 // Make libMIOpen.so accessible to the test driver
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -8,11 +8,7 @@ void buildProject(String target, String cmakeOpts) {
         cmakeArgs: "-DMLIR_MIOPEN_DRIVER_ENABLED=1 $cmakeOpts"
 }
 
-void getAndBuildMIOpen(String prefixOpt, String cmakeOpts) {
-    git branch: 'develop', poll: false,\
-        url: 'https://github.com/ROCmSoftwarePlatform/MIOpen.git'
-    cmake arguments: "-P install_deps.cmake --minimum ${prefixOpt}",\
-        installation: "InSearchPath"
+void buildMIOpen(String cmakeOpts) {
     sh '[ ! -d build ] || rm -rf build'
     cmakeBuild generator: 'Unix Makefiles',\
         buildDir: 'build',\
@@ -23,6 +19,14 @@ void getAndBuildMIOpen(String prefixOpt, String cmakeOpts) {
                      ${cmakeOpts}
                      """
     sh 'cd build; make -j $(nproc) MIOpenDriver'
+}
+
+void getAndBuildMIOpen(String prefixOpt, String cmakeOpts) {
+    git branch: 'develop', poll: false,\
+        url: 'https://github.com/ROCmSoftwarePlatform/MIOpen.git'
+    cmake arguments: "-P install_deps.cmake --minimum ${prefixOpt}",\
+        installation: "InSearchPath"
+    buildMIOpen(cmakeOpts)
 }
 
 void showEnv() {
@@ -53,8 +57,8 @@ void preMergeCheck() {
     sh 'python3 ./mlir/utils/jenkins/static-checks/premerge-checks.py'
 }
 
-void testMIOpenDriver(String dtype, String xdlops, String layout, String direction, String tuning) {
-    echo "Config is (${dtype}, ${layout}, xdlops=${xdlops}, dir=${direction}, tuning=${tuning})"
+void testMIOpenDriver(String dtype, String layout, String direction, String tuning) {
+    echo "Config is (${dtype}, ${layout}, dir=${direction}, tuning=${tuning})"
     sh 'ldconfig'
     dir('MIOpen/build/') {
         sh """
@@ -62,7 +66,6 @@ void testMIOpenDriver(String dtype, String xdlops, String layout, String directi
         --layout ${layout} \
         --direction ${direction} \
         --dtype ${dtype} \
-        ${xdlops == '1' ? '--xdlops' : '--no-xdlops'} \
         ${tuning == '1' ? '--tuning' : '--no-tuning'} \
         < ${WORKSPACE}/mlir/utils/jenkins/miopen-tests/resnet50-miopen-configs"""
     }
@@ -97,6 +100,8 @@ pipeline {
                      description: 'Run extra nightly-only tests')
         booleanParam(name: 'canXdlops', defaultValue: params.canXdlops == false ? false : true,
                      description: 'Can this CI instance use xdlops (no for public server)')
+        booleanParam(name: 'weekly', defaultValue: params.weekly == false ? false : true,
+                     description: 'Run weekly-only jobs')
 
         // Each below control whether to run a individual stage from parallel run
         // They default to true but deverloper can toggle them for debugging purpose
@@ -236,13 +241,10 @@ pipeline {
                             }
                         }
                         stage("Build MIOpen with libMLIRMIOpen") {
-                            when { allOf {
-                                equals expected: true, actual: params.nightly;
-                                anyOf {
-                                    equals expected: true, actual: params.perfTest;
-                                    equals expected: true, actual: params.canXdlops;
-                                } }
-                            }
+                            when { anyOf {
+                                    equals expected: true, actual: params.nightly;
+                                    equals expected: true, actual: params.weekly;
+                            }}
                             steps {
                                 dir('MIOpen') {
                                 getAndBuildMIOpen('--prefix ./MIOpenDeps', '''-DMIOPEN_USE_MLIR=On
@@ -267,12 +269,10 @@ pipeline {
                             }
                         }
                         stage("Stash MIOpen with libMLIRMIOpen") {
-                            when { allOf {
-                                equals expected: true, actual: params.nightly;
-                                anyOf {
-                                equals expected: true, actual: params.perfTest;
-                                equals expected: true, actual: params.canXdlops;
-                            } } }
+                            when { anyOf {
+                                    equals expected: true, actual: params.nightly;
+                                    equals expected: true, actual: params.weekly;
+                            }}
                             steps {
                                 stash name:"MIOpen-MLIR-kernels",\
                                       includes: """
@@ -281,6 +281,29 @@ pipeline {
                                       """
                             }
                         }
+                        stage("Test MIOpen config with tuning") {
+                            when {
+                                    equals expected: true, actual: params.weekly;
+                            }
+                            environment {
+                                // Make libMIOpen.so accessible to the test driver
+                                LD_LIBRARY_PATH="${WORKSPACE}/MIOpen/build/lib:$LD_LIBRARY_PATH:"
+                            }
+                            steps {
+                                sh 'ldconfig'
+                                dir('MIOpen/build/') {
+                                    sh 'date --utc +%Y-%m-%d > tuning-date'
+                                    sh """
+                                    bash ${WORKSPACE}/mlir/utils/jenkins/miopen-tests/miopen_validate.sh --tune-all\
+                                       < ${WORKSPACE}/mlir/utils/jenkins/miopen-tests/resnet50-miopen-configs"""
+                                    sh 'ls -l MIOpenUserDB'
+                                }
+                                // Save user database  for nightly jobs
+                                archiveArtifacts artifacts: 'MIOpen/build/MIOpenUserDB/**, MIOpen/build/tuning-date',\
+                                                 onlyIfSuccessful: true
+                            }
+                        }
+ 
                     }
                     post {
                         always {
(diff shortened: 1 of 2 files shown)


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -236,13 +241,10 @@ pipeline {
...
-                                equals expected: true, actual: params.nightly;
-                                anyOf {
-                                    equals expected: true, actual: params.perfTest;
-                                    equals expected: true, actual: params.canXdlops;
-                                } }
-                            }
+                            when { anyOf {
+                                    equals expected: true, actual: params.nightly;
+                                    equals expected: true, actual: params.weekly;
+                            }}
```
krzysz00, line 250: Let's keep the control constants in here

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -267,12 +269,10 @@ pipeline {
...
-                            when { allOf {
-                                equals expected: true, actual: params.nightly;
-                                anyOf {
-                                equals expected: true, actual: params.perfTest;
-                                equals expected: true, actual: params.canXdlops;
-                            } } }
+                            when { anyOf {
+                                    equals expected: true, actual: params.nightly;
+                                    equals expected: true, actual: params.weekly;
+                            }}
```
krzysz00, line 281: Same deal - let's make sure we can run individual parts of nightly

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -281,6 +281,29 @@ pipeline {
...
+                                sh 'ldconfig'
+                                dir('MIOpen/build/') {
+                                    sh 'date --utc +%Y-%m-%d > tuning-date'
+                                    sh """
+                                    bash ${WORKSPACE}/mlir/utils/jenkins/miopen-tests/miopen_validate.sh --tune-all\
+                                       < ${WORKSPACE}/mlir/utils/jenkins/miopen-tests/resnet50-miopen-configs"""
+                                    sh 'ls -l MIOpenUserDB'
+                                }
+                                // Save user database  for nightly jobs
+                                archiveArtifacts artifacts: 'MIOpen/build/MIOpenUserDB/**, MIOpen/build/tuning-date',\
```
krzysz00, line 309: 1) This only runs weekly, right?

2) Have you tested to make sure this won't conflict with performance report archiving?

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -354,27 +377,35 @@ pipeline {
...
                 stage("Copy MIOpen with MLIR kernels") {
                     steps {
                         sh 'rm -rf MIOpen'
                         unstash name: "MIOpen-MLIR-kernels"
+                        //copyArtifacts filter: 'MIOpen/build/MIOpenUserDB,MIOpen/build/tuning-date',\
+                        //                       optional: true,\
+                        //                       flatten: true,\
+                        //                       projectName: "/${JOB_NAME}",\
+                        //                       selector: lastSuccessful(),\
+                        //                       target: 'MIOpen/build/'
```
krzysz00, line 413: Why's this commented out?

mlir/utils/jenkins/miopen-tests/miopen_validate.sh
```diff
@@ -78,8 +94,8 @@ function construct_fixed_args() {
        exit 2
     fi
     case "$DTYPE" in
-        fp16 ) FIXED_CONFIG_ARGS+=("convfp16") ;;
-        fp32 ) FIXED_CONFIG_ARGS+=("conv") ;;
+        fp16 ) FIXED_CONFIG_ARGS=("convfp16") ;;
```
krzysz00, line 111: Why this change?

mlir/utils/jenkins/miopen-tests/miopen_validate.sh
```diff
@@ -125,8 +136,12 @@ function setup_environment() {
         4) MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvMlirIgemmWrW ;;
         *) echo "$0: Unsupported direction flag $DIRECTION"; exit 2
     esac
-    if [[ $XDLOPS == 1 ]]; then
+    local xdlops
+    xdlops=$(/opt/rocm/bin/rocm_agent_enumerator |grep gfx908)
```
krzysz00: Is this really the right approach? It feels like we'd want to have xdlops be something we can pass in instead of relying on autodetection.

That is, I'm curious about the logic here.

mlir/utils/jenkins/miopen-tests/miopen_validate.sh
```diff
@@ -138,8 +153,6 @@ function run_tests() {
         exit 3
     fi
 
-    clean_miopen_caches
```
krzysz00, line 141: ?

mlir/utils/jenkins/miopen-tests/miopen_validate.sh
```diff
@@ -160,12 +173,35 @@ function run_tests() {
...
+            done
+        done
+    done
+}
+
+function main() {
+    clean_miopen_caches
+    parse_options "$@"
+    get_configs
+    if [[ $TUNEALL == 1 ]]; then
```
krzysz00: The tuneall machinery is so you can have the Jenkinsfile run all the tuning on one node?

krzysz00 (commented):
I'm not entirely understanding how tuning will interact with the rest of nightly CI - ex. with the performance reports, especially since the important bit seems to be commented out. Can you clarify?

mlir/utils/jenkins/Jenkinsfile
yiqian1: > I'm not entirely understanding how tuning will interact with the rest of nightly CI - ex. with the performance reports, especially since the important bit seems to be commented out. Can you clarify?

The tuning results will be used by the performance reports in nightly CI. This will be in another PR.

mlir/utils/jenkins/Jenkinsfile
yiqian1: Yes. This runs weekly. I haven't tested this with performance reports yet, but I don't think they will conflict, as they are different CI jobs.

mlir/utils/jenkins/Jenkinsfile
yiqian1: This will be in another PR.

mlir/utils/jenkins/miopen-tests/miopen_validate.sh
yiqian1: To reset FIXED_CONFIG_ARGS every time construct_fixed_args() is called. For the tune-all case, construct_fixed_args() is called multiple times.

mlir/utils/jenkins/miopen-tests/miopen_validate.sh
yiqian1: The autodetection allows us to run xdlops tests on private CI and non-xdlops on public CI. That's what the performance test is doing. Thus we can test on more CI nodes and have a more balanced work on private and public CI.

mlir/utils/jenkins/miopen-tests/miopen_validate.sh
yiqian1: Moved clean_miopen_caches to main().

mlir/utils/jenkins/miopen-tests/miopen_validate.sh
yiqian1: Yes. I want the tuning to generate one user perfDb which I can use for performance report. If the tuning is run in parallel, there will be multiple perfDbs at different nodes, making them hard to be used.

mlir/utils/jenkins/Jenkinsfile
yiqian1: What do you mean?
With this change, the stage can be run for nightly+staticLib (MIOpen tests), nightly+perfTest, or weekly+staticLib. If canXdlops=true, it is on a private CI node; otherwise, on a public CI node.

mlir/utils/jenkins/Jenkinsfile
krzysz00: Re the control constants, what I mean is we still want the `perfTest` variable so that we can run just performance tests, or something analogous to it.

mlir/utils/jenkins/Jenkinsfile
krzysz00: Yeah, that's true, but currently `weekly` duplicates a lot of `nightly` and so we'll be getting things like weekly performance regressions.

mlir/utils/jenkins/miopen-tests/miopen_validate.sh
krzysz00: Makes sense

mlir/utils/jenkins/miopen-tests/miopen_validate.sh
krzysz00: Good point, and this makes sense as a change.

krzysz00: Though it might be nice to retain the ability to override `xdlops` for manual testing

mlir/utils/jenkins/miopen-tests/miopen_validate.sh
```diff
@@ -59,13 +63,25 @@ function parse_options() {
     done
 
     # Check required args
-    [[ $got_layout == 1 && $got_direction == 1 && $got_dtype == 1 ]] || usage
+    [[ $got_layout == 1 && $got_direction == 1 && $got_dtype == 1 ]] || [[ $TUNEALL == 1 ]] || usage
```
krzysz00: I think you can have only one call to `[[`, but that's a matter of taste

mlir/utils/jenkins/Jenkinsfile
krzysz00: I mean that we don't have this stage restricted to (perfTest || staticLib) && (nightly || weekly)

krzysz00 (commented):
I've got a few minor nitpicks but otherwise I'm happy with this

mlir/utils/jenkins/Jenkinsfile
yiqian1: Changed it to weekly && staticLib.

jerryyin (dismissed):
Based on the existing [run](http://ml-ci.amd.com:21096/blue/organizations/jenkins/MLIR%2Ftest_weeklyCI/detail/test_weeklyCI/9/pipeline/276), my high level feedbacks are:

- In weekly CI, MIOpen resnet 50 configs shall test with both tuning on and off, yet you only have tuning off
- In nightly CI, we need at least 2 configs of all alternatives to make sure its minimal test coverage

yiqian1:
> Based on the existing [run](http://ml-ci.amd.com:21096/blue/organizations/jenkins/MLIR%2Ftest_weeklyCI/detail/test_weeklyCI/9/pipeline/276), my high level feedbacks are:
> 
> * In weekly CI, MIOpen resnet 50 configs shall test with both tuning on and off, yet you only have tuning off

Tuning is tested in the "Test MIOpen config with tuning" stage.

> * In nightly CI, we need at least 2 configs of all alternatives to make sure its minimal test coverage

This will be in another PR.

jerryyin:
> Tuning is tested in the "Test MIOpen config with tuning" stage.

That stage alone is costing more than 3 hours, is that the best way to do it?

Tuning on private CI will take even longer time. That's why we don't want nightly CI to do it. I'm not too worried about the time as it is run on a weekly base. This stage includes tuning for NCHW and NHWC. I can make them in parallel. That's probably the only thing I can do.

> This will be in another PR.

May I ask what the ETA for the other PR: In the mean time we'd lose coverage if anything break MIOpen integration (and will only know until weekly tests). Therefore desirably the next PR needs to be out as soon as possible.

I want to make it out this week. The only thing that is not covered by the current nightly CI is tuning for NHWC. The performance test most likely will fail if any tuning for NCHW fails.

mlir/utils/jenkins/miopen-tests/miopen_validate.sh
yiqian1: Good idea. I think I will have it in the next PR.

krzysz00: Could you add it here?

The way I'd do this is to have the xdlops autodetection in options parsing - if the user doesn't pass in `--xdlops` or `--no-xdlops`, call the agent enumerator and detect a setting.

mlir/utils/jenkins/miopen-tests/miopen_validate.sh
```diff
@@ -59,13 +66,36 @@ function parse_options() {
...
+        case "$DTYPE" in
+            fp16 | fp32 ) ;;
+            * ) echo "$0: Invalid dtype $DTYPE"; usage ;;
+        esac
+    fi
+
+    # Detect XDLOPS if not specified
+    local xdlops
+    if [[ $XDLOPS == -1 ]]; then
+      xdlops=$(/opt/rocm/bin/rocm_agent_enumerator |grep gfx908)
```
krzysz00, line 82: My brain is saying `grep -q` and checking the exit code in the if statement

That is dropping `xdlops` and going
```
if /opt/rocm/bin/rocm_agent_enumerator | grep -q gfx908; then
...
```

but that's a really tiny nitpick

mlir/utils/jenkins/Jenkinsfile
krzysz00: Ok, just to make sure, this test needs to run in every nightly, but only in weeklies where we're doing static library tests?

`nightly || (weekly && staticLib)` doesn't feel right.

And, as a followup, why is this now `staticLib` and not `perfTest`?

krzysz00 (commented):
One nitpick, one actual question, otherwise looks fine

mlir/utils/jenkins/Jenkinsfile
yiqian1: > Ok, just to make sure, this test needs to run in every nightly, but only in weeklies where we're doing static library tests?
> 
This is true for now. Eventually this test is only run in weeklies. Nightlies will only run two selected MIOpen configs.

> `nightly || (weekly && staticLib)` doesn't feel right.
> 

When I'm done with the new nightly CI, this will become (weekly && staticLib)

> And, as a followup, why is this now `staticLib` and not `perfTest`?

Because it's not part of performance tests.

krzysz00: Ah. In which case, so as to not break the selective CI setup, can you make this `(nightly || weekly) && staticLib`

yiqian1: Sorry. I said wrong. This test is run (nightly&&perfTest) || (nightly&&staticLib) || (weekly&&staticLib). The condition (perfTest || staticLIb) has been specified for the top-level stage "Static Library". Since perfTest is only done in nightlies, I disallow it for weeklies.

krzysz00: Ah. That makes sense, thanks for clarifying!

krzysz00 (approved):
LGTM, overall

PR #431, opened 2021-11-02
Computer more of xdlops addressing at compile time.

krzysz00:
A lot of the code in the xdlops writeback pass depends only on the loop iteration, which means we can perform the computations on those values at compile time, reducing the amount of code we generate.


whchung (approved):
Nice! Thanks!

PR #434, opened 2021-11-04
[DO NOT MERGE] Revert "Hoist loop invariant computations out of loops in xdlops lowering"

jerryyin:
For regression testing.


jerryyin:
Does not change expected CI result, closing now.

PR #437, opened 2021-11-08
Compute more of xdlops addressing at compile time. (#431 redo)

krzysz00:
A lot of the code in the xdlops writeback pass depends only on the loop iteration, which means we can perform the computations on those values at compile time, reducing the amount of code we generate.

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
+++ b/mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
@@ -6996,22 +6996,10 @@ struct GridwiseGemmV2RewritePattern
     // llvm::errs() << "iterationsPerVectorC: " << iterationsPerVectorC << "\n";
     // llvm::errs() << "vectorCoffset: " << vectorCoffset << "\n";
 
-    auto MPerXdlopsConstantOp = b.create<ConstantIndexOp>(loc, MPerXdlops);
-    auto NPerXdlopsConstantOp = b.create<ConstantIndexOp>(loc, NPerXdlops);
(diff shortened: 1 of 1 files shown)


whchung (approved):

krzysz00:
From re-running the failing tuning configs locally, I think I fixed the bug (I used the wrong constant in one spot, so there was genuinely a bug in the PR), but I think we'll want to give it a CI run just to be safe.

Now I just want to know why this didn't get caught before tuning.

krzysz00:
Update: the tuning bug was triggered by non-default tuning configs.

```
./bin/mlir-miopen-driver "--conv-config=--x2 1 --operation conv2d --kernel_id 0 --num_cu 120 --arch amdgcn-amd-amdhsa:gfx908:sramecc+:xnack- --groupsize 1 --fil_layout GNCHW --fil_type fp16 --in_layout NGCHW --out_layout NGCHW --in_type fp16 --out_type fp16 --batchsize 256 --in_channels 128 --out_channels 128 --in_h 28 --in_w 28 --out_h 28 --out_w 28 --fil_h 3 --fil_w 3 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h 1 --padding_w 1 --kernel_name mlir_gen_igemm_conv2d_v4r4_fwd_xdlops0 --perf_config 128,256,8,64,128,8,1,1" -c -pv
```
runs successfully with the commit that fixes the typo and fails before it.

( @jerryyin  for the status update )

jerryyin:
@krzysz00 Nice! Thanks for confirming. This means there's testing coverage perspective we are good and the right regression catches the issue as it should be.

PR #438, opened 2021-11-08
Set the default value of 'weekly'

yiqian1:


jerryyin:
Abstain. I'm not as familiar in the proper syntax in making trigger selection sticky. @krzysz00 knows much better than me.

krzysz00 (approved):
LGTM - I'm reasonably confident that this does "if weekly is on, set weekly to on, otherwise set it to false"

PR #440, opened 2021-11-10
Simplify index diff map computation

krzysz00:
1. Constant fold more of the computation if possible.
This both simplifies the generated IR and enables further
optimizations, such as

2. Refactor carry application in Merge/Unfold operations.
Instead of a bunch of if statements, use remainders and integer
divisions to perform the carry logic. While in more complex cases,
such as the padding-1 kernel in the PR E2E tests, this ultimately
optimizes to the same ISA (at least, looking at the store logic),
it has the advantage of being correct when the overflow in a merge in
more than 1. In addition, this patch removed a lot of store address
computation logic from the stores in

-fil_layout=kyxc -in_layout=nhwc -out_layout=nhwk -batchsize=1 -in_channels=64 -in_h=2 -in_w=2 -out_channels=64 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=0 --padding_w=0 -t f32 --operation conv2d -x2 -pv -c

where the generated ISA now has a bunch of buffer_store_dword
instructions with constant offsets, which demonstrates the potential
value of this change.

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
+++ b/mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
@@ -54,6 +54,7 @@
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Support/ErrorHandling.h"
 #include "llvm/Support/MathExtras.h"
+#include "llvm/Support/raw_ostream.h"
 #include <algorithm>
 #include <iterator>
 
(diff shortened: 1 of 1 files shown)


asroy (approved):
The constant propagating part LGTM

For the change to "merge" logic, I now understand that: for some reason compiler was not lowering the original MLIR implementation to good ISA, and this PR also tries to address it. I think if the modified logic make the ISA better, that also LGTM. However, the original implementation, if compiled properly, may give better ISA than the modified version. So @krzysz00 if you want to get to the bottom of the original issue, I can help with that. But I'm going to prove this PR as it is.

cc @whchung

krzysz00:
From the example command above, I'm attaching the changes to the generated LLVM and ISA. Observe that, after this PR, there are `buffer_store_dwordx4` instructions that weren't present in the pre-#440 ISA (this is all near the bottom of the kernel).  Also, observe that the final address computations are like
```
        v_add_u32_e32 v2, v18, v0
        v_accvgpr_read_b32 v1, a30
        v_lshl_add_u32 v2, v2, 2, v4
        v_add_u32_e32 v0, v19, v0
        buffer_store_dword v1, v2, s[8:11], 0 offen offset:4095
```
in the pre-440 ISA, while they're 
```
        v_accvgpr_read_b32 v0, a31
        v_add_u32_e32 v1, 0x8000006b, v16
        s_nop 0
        buffer_store_dword v0, v1, s[8:11], 0 offen
```
showing that the out-of-bounds checks are being implemented more efficiently.

(Heck, notice how many of the thread-conditional things involve constants now)

Files at: [440-files.zip](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/files/7558068/440-files.zip)

PR #441, opened 2021-11-10
Swizzles part 2

krzysz00:
Enable vector stores in matrix C store back to memory when using xdlops.

Use swizzles to transpose multiplication results in order to enable this vector store in the NKHW case.

TODO items:

- Applying similar transformations to non-xdlops
- Changing the backward weight atomic add code to not move dimensions around in order to enable vector writes in those kernels

(This is currently skip-ci due to dependencies on #437 and #425 but is ready for review)

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
+++ b/mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
@@ -6920,6 +6920,9 @@ struct GridwiseGemmV2RewritePattern
 
     auto gemmCVectorizedMatrixDim =
         op->getAttrOfType<IntegerAttr>("matrix_c_source_vector_read_dim");
+    int64_t matrixCDataPerCopy =
+        op->getAttrOfType<IntegerAttr>("matrix_c_data_per_copy").getInt();
+
     constexpr int64_t swizzleGroup = 4;
     // Ensure that the prerequisites are met
     // - The N dimension of the output will be stored vectorized
@@ -6927,8 +6930,11 @@ struct GridwiseGemmV2RewritePattern
     //    so transpose is well defined
     // - None of the larger dimensions of interest have overhangs that lead to
     //    incomplete transposes
+    // - The writes will vectorize: if we're not getting vectorization
+    //    due to HW % swizzleGroup != 0, then there's no point
     bool enableOutSwizzles =
         gemmCVectorizedMatrixDim.getInt() == gemmCDimN &&
+        (matrixCDataPerCopy >= swizzleGroup) &&
         (M2 == swizzleGroup && (m % swizzleGroup == 0) &&
          (n % swizzleGroup == 0) && (MPerWave % swizzleGroup == 0) &&
          (NPerWave % swizzleGroup == 0));
--- a/mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
+++ b/mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
@@ -16,14 +16,15 @@
 #include "mlir/Conversion/AffineToStandard/AffineToStandard.h"
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/GPU/GPUDialect.h"
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
 #include "mlir/Dialect/MIOpen/AffineMapHelper.h"
 #include "mlir/Dialect/MIOpen/MIOpenOps.h"
 #include "mlir/Dialect/MIOpen/Passes.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/SCF/SCF.h"
-#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
 #include "mlir/Dialect/StandardOps/IR/Ops.h"
 #include "mlir/Dialect/Vector/VectorOps.h"
+#include "mlir/IR/AffineExpr.h"
 #include "mlir/IR/AffineMap.h"
 #include "mlir/IR/Attributes.h"
 #include "mlir/IR/Builders.h"
@@ -481,7 +482,8 @@ inline void emitStoreLogic(
     MemRefType destType, Type typeToStore, bool toEmitOOBStoreCheckLogic,
     const SmallVector<unsigned, 8> &oobStoreCheckDims, const Value &dest,
     const SmallVector<Value, 8> &destLowerIndices, const Value &value,
-    InMemoryDataOperation memoryOp = InMemoryDataOperation::Set) {
+    InMemoryDataOperation memoryOp = InMemoryDataOperation::Set,
+    bool print = false) {
   auto emitStoreInstruction = [&b, &loc](
                                   const Value &value, MemRefType destType,
                                   Type typeToStore, const Value &dest,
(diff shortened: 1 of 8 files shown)


krzysz00:
@whchung 

So, since I checked, performance is mostly flat after this PR - some really minor improvements to fp16 wrw, pretty much no change (and _maybe_ a slight increase in overhead) in forward kernels. There's also some funky plunge in one backward data kernel, but this PR doesn't touch those so ????

My hypotheses
- AMD hardware batches up memory writes more agressively than expected, especially at the end  of kernels, so swizzles may not be buying us much in the NCHW case at the end of the kernel
- fp16 performance is lower than I'd like due to the fact that we swizzle the fp32 values and then convert them. I have a plan for this but it's a followup PR
- There's actually overhead caused by the `select` workaround

In any case, this PR also vectorizes NHWC writes, which is likely to help with things.

All numbers taken on 112 yesterday evening.
[swizzles-performance-impacts-nchw.pdf](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/files/7564336/swizzles-performance-impacts-nchw.pdf)

krzysz00:
If folks find time to review this before I get back, I'd appreciate it, but no pressure

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -5090,18 +5119,37 @@ static void affixThreadwiseCopyV2Attributes(miopen::ThreadwiseCopyV2Op top,
...
+  } else {
+    top->setAttr("dim_access_order", b.getI32ArrayAttr({0, 1, 2, 3, 4}));
+  }
+
+  // Account for split m/n dimension
+  bool vectorStoreOverride = false;
+  int64_t vectorGemmDim =
+      gop->getAttrOfType<IntegerAttr>("matrix_c_source_vector_read_dim")
+          .getInt();
+  if (vectorGemmDim == 1) {
```
whchung: some comments might be needed here to explain the rationale.
or perhaps introduce an enumeration to hide these numbers?

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -6986,110 +7037,246 @@ struct GridwiseGemmV2RewritePattern
...
         loc, laneId_xdlops_gemm, num_threads_blk_ConstantOp);
 
-    // emit vector swizzles
-
-    auto *ctx = op.getContext();
-
+    auto gemmCVectorizedMatrixDim =
+        op->getAttrOfType<IntegerAttr>("matrix_c_source_vector_read_dim");
+    constexpr int64_t swizzleGroup = 4;
+    bool enableOutSwizzles =
```
whchung, line 6935: the heuristic might better be commented.

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
```diff
@@ -415,6 +425,64 @@ class PopulateParamsBase {
...
+    int64_t dataPerThread =
+        (params->gemmMPerBlock * params->gemmNPerBlock) / blockSize;
+    if (!(dataPerThread > 0)) {
+      return failure();
+    }
+
+    // TODO: Allow vectorization group size of 2
+    int64_t vectorizationSize = 4;
+    if (ConvOpType::Conv2DBwdDataOpType == op) {
+      vectorizationSize = 1;
```
whchung, line 454: comment might be neede to state we are disabling vector stores for backward data convolution.

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
```diff
@@ -415,6 +425,64 @@ class PopulateParamsBase {
...
+    } else {
+      out.dataPerCopy = 1;
+    }
+
+    auto &dimIndexVal = ctx.dimIndexVal;
+    // Find dimensions in which the copy will take place
+    switch (op) {
+    case mlir::miopen::Conv2DOpType:
+      if (dimIndexVal["ko"].first == 4) {
+        out.gemmVectorDim = 1;
```
whchung: Due to PR #423, gemm dimensions may not be fixed as 3 (GemmK / GemmM / GemmN), but might be 4 (GemmK / GemmM / GemmN / GemmKPack), would it be a better idea to use an enumeration?

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -6986,110 +7037,246 @@ struct GridwiseGemmV2RewritePattern
...
-         getAffineDimExpr(3, ctx)},
-        op.getContext());
+    auto *ctx = op.getContext();
 
+    Value newOutputTransformOp;
+    SmallVector<Value, 4> vectors;
+    vectors.reserve(tailResults.size());
+    AffineMap tensorToXdlopsVectorMap;
+    if (enableOutSwizzles) {
+      // The swizzle operation doesn't fundamentally affect the mapping
```
whchung, line 6949: For the attribute maps here, perhaps it'd be a good idea to have a meeting with @asroy / @zjing14 together so we review the logic together?

whchung (commented):
Just did the initial overview of the PR to grasp the basic idea.

I think we need an in-depth discussion with @asroy / @zjing14 to summarize your findings.

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
krzysz00: Not a bad idea, can you schedule?

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
krzysz00: I don't think that's related - these are GemmG x GemmM x GemmN dimensions in the output, which isn't affected by KPACK. We could introduce an enum for clarity, but then we'd have some inconsistency about when we are/aren't using it.

krzysz00:
Note: even though Nothing Has Changed (tm), the config
```
--x2 1 --operation conv2d --kernel_id 0 --num_cu 120 --arch amdgcn-amd-amdhsa:gfx908:sramecc+:xnack- --groupsize 1 --fil_layout GNCHW --fil_type fp16 --in_layout NGCHW --out_layout NGCHW --in_type fp16 --out_type fp16 --batchsize 256 --in_channels 256 --out_channels 256 --in_h 14 --in_w 14 --out_h 14 --out_w 14 --fil_h 3 --fil_w 3 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h 1 --padding_w 1 --kernel_name mlir_gen_igemm_conv2d_v4r4_fwd_xdlops0 --perf_config 128,128,8,64,64,4,1,1
```
is broken here but working in the main miopen-dialect branch

krzysz00:
Update: I think I found the issue (I was, for example, trying to vectorize writes to a 128 x 128 x 1 x 1 output tensor over the gemmN dimension)

krzysz00:
(@whchung In case we're running into similar)

The configuration
```
--x2 1 --operation conv2d --kernel_id 0 --num_cu 120 --arch amdgcn-amd-amdhsa:gfx908:sramecc+:xnack- --groupsize 1 --fil_layout GNCHW --fil_type fp16 --in_layout NGCHW --out_layout NGCHW --in_type fp16 --out_type fp16 --batchsize 256 --in_channels 256 --out_channels 256 --in_h 14 --in_w 14 --out_h 14 --out_w 14 --fil_h 1 --fil_w 1 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h 0 --padding_w 0 --kernel_name mlir_gen_igemm_conv2d_v4r4_fwd_xdlops0 --perf_config 128,128,8,64,64,8,1,1
``` 
fails when the GPU kernel is compiled with -O1 (or -O3) but passes with -O0.

(There's a similar failure on the somewhat minimized:
```
--x2 1 --operation conv2d --kernel_id 0 --num_cu 120 --arch amdgcn-amd-amdhsa:gfx908:sramecc+:xnack- --groupsize 1 --fil_layout GNCHW --fil_type fp16 --in_layout NGCHW --out_layout NGCHW --in_type fp16 --out_type fp16 --batchsize 32 --in_channels 64 --out_channels 128 --in_h 12 --in_w 12 --out_h 12 --out_w 12 --fil_h 1 --fil_w 1 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h 0 --padding_w 0 --kernel_name mlir_gen_igemm_conv2d_v4r4_fwd_xdlops0 --perf_config 128,128,8,64,64,8,1,1
```
)

This PR will probably now be twiddling its thumbs for a hot minute because of this, though I'd still like review

krzysz00:
@asroy @whchung I've updated this PR to point at the upstream updates branch to work around a compiler bug - it still needs review.

Now I just need to work out how to run CI on this...

whchung (approved):

krzysz00:
Note: this PR is currently running nightly tests.

krzysz00:
Weeklies passed, merging

PR #442, opened 2021-11-11
Remove extra parentheses from MIOpenDriver.py

yiqian1:


krzysz00 (approved):
Looks like a straightforward bugfix to me, approved

PR #443, opened 2021-11-11
Restore the --gpu-to-hsaco option because it's used in integration te…

pcf000:
…sts.

Update the tests to derive the chip in the lit config and specify it as
sub-option, replacing the old use of the default.

### DIFF
--- a/external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp
+++ b/external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp
@@ -508,10 +508,8 @@ void mlir::registerGpuSerializeToHsacoPass() {
         LLVMInitializeAMDGPUTargetInfo();
         LLVMInitializeAMDGPUTargetMC();
 
-        // Known-bad values for constructor arguments since the instance of the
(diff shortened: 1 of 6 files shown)


pcf000:
@krzysz00 , this is what I sent you the other day (which message I now can't find).  I hadn't made a pull request.

krzysz00 (approved):
LGTM, but please make sure the changes are mirrored in https://reviews.llvm.org/D112668 or send me the diff if they aren't. Also, I'd double-check we're setting `config.rocm_path` in the external lit.cfg.py.in - that's a change I had to make upstream.

pcf000:
Date: Thursday, November 11, 2021  11:21PM CST
    From: Krzysztof Drewniak ***@***.***>

    LGTM, but please make sure the changes are mirrored in
    https://reviews.llvm.org/D112668 or send me the diff if they aren't.
    Also, I'd double-check we're setting config.rocm_path in the
    external lit.cfg.py.in - that's a change I had to make upstream.

Yes to both:  all changes are in the upstream review, and
config.rocm_path is set in both mlir/test/lit.site.cfg.py.in and
external/llvm-project/mlir/test/lit.site.cfg.py.in.

wcy123 (approved):

PR #444, opened 2021-11-11
Add a full-model Resnet50 test

pcf000:
A python script uses Tensorflow's Keras to make a model with a single fixed
image, then uses recent TF enhancements to generate TOSA MLIR to stdout,
whence it is piped to mlir-opt and mlir-cpu-runner. to verify the result.

Requires tf-nightly >= 2.8.0-dev20211102, which in turn requires python
3.7.  Adds those items to the rocm/mlir Dockerfile.  Marks the test as
unsupported if they aren't available.


mlir/test/e2e/Inputs/resnet50-test.py
```diff
@@ -0,0 +1,84 @@
...
+import numpy as np
+import argparse
+import sys
+from contextlib import redirect_stdout
+from tensorflow.keras.applications.resnet50 import preprocess_input, decode_predictions
+from tensorflow.keras.preprocessing import image
+from tensorflow.keras.applications.resnet50 import ResNet50
+from tensorflow.python.framework.convert_to_constants import convert_variables_to_constants_v2
+
+# wget https://upload.wikimedia.org/wikipedia/commons/thumb/3/31/Red_Smooth_Saluki.jpg/590px-Red_Smooth_Saluki.jpg
```
pcf000, line 13: Where the image file I'm using came from.  It's included in the commit.

mlir/test/e2e/lit.local.cfg
```diff
@@ -0,0 +1,7 @@
+import importlib
+import shutil
+
+# Test file current forces python3.7, because it needs >=3.7.
+# Test requires tensorflow (which should be a recent tf-nightly).
+if not shutil.which('python3.7') or not importlib.util.find_spec('tensorflow'):
```
krzysz00: Let's not have a hard dependency on 3.7 - check the python version this configuration is being run in instead

mlir/test/e2e/resnet50-tosa-partition.mlir
```diff
@@ -0,0 +1,13 @@
+// RUN:   python3.7 %S/Inputs/resnet50-test.py %S/Inputs/590px-Red_Smooth_Saluki.jpg \
+// RUN: | mlir-opt --tosa-partition-pipeline --tosa-to-linalg-on-tensors --tosa-to-standard \
```
krzysz00: Sanity check: is this `mlir-opt` or `miopen-opt` (that is, have we been adding the tosa partitioning stuff to our directory or to external/)?

mlir/utils/jenkins/Dockerfile.rocm
```diff
@@ -105,3 +105,12 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 RUN git clone https://github.com/RadeonOpenCompute/rocminfo.git && \
     cp rocminfo/rocm_agent_enumerator /opt/rocm/bin && \
     rm -rf rocminfo
+
+# The resnet50-tosa-partition.mlir test requires python3.7 (it needs a recent
+# tensorflow 2.8.0 or later, which requires >=3.7, and test itself hard-codes
+# 3.7).
+RUN apt-add-repository ppa:deadsnakes/ppa && apt-get update \
+   && apt-get install -y --no-install-recommends \
+        python3.7 python3.7-venv \
```
krzysz00: If we're doing this, might as well go to 3.9?

mlir/test/e2e/resnet50-tosa-partition.mlir
krzysz00: More significantly, per @whchung , generating test inputs at test-time is generally considered not good in LLVM - see https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/miopen-dialect/mlir/utils/generateInWarpTransposeTest.py and https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/miopen-dialect/mlir/test/Dialect/MIOpen/integration/in_warp_transpose/in_warp_transpose.mlir

Could this test be reworked to generate the MLIR offline and then to run that through the pipeline during testing? This has the advantage of letting us test the partitioning pass as well as the full end-to-end setup.

krzysz00 (changes requested):
Thanks for putting in the work on this!

I'm requesting changes because I don't like how we're checking for Python version, but there might be some more significant issues with how this is structured that other folks can comment on.

mlir/utils/jenkins/Dockerfile.rocm
```diff
@@ -106,11 +106,11 @@ RUN git clone https://github.com/RadeonOpenCompute/rocminfo.git && \
...
+# The resnet50-tosa-partition.mlir test requires python >= 3.7 (it needs
+# a recent tensorflow 2.8.0 or later).
 RUN apt-add-repository ppa:deadsnakes/ppa && apt-get update \
-   && apt-get install -y --no-install-recommends \
-        python3.7 python3.7-venv \
-   && python3.7 -m pip install --upgrade pip \
-   && python3.7 -m pip install pillow tf-nightly
+    && apt-get install -y --no-install-recommends \
+           python3.7 python3.7-venv \
+    && rm -f /usr/bin/python3 && ln -s /usr/bin/python3.7 /usr/bin/python3 \
```
krzysz00: `update-alternatives`, please?

pcf000:
This guy's been hanging around unresolved forever.  Closing because the world has changed around it, and we'll likely have something analogous soon, but with torch-mlir and less trouble.

PR #445, opened 2021-11-12
MLIR#380: support bias in tosa-to-miopen

sjw36:
* optimize out zero bias

https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/380

### DIFF
--- a/mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
+++ b/mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
@@ -63,10 +63,10 @@ public:
   matchAndRewrite(tosa::Conv2DOp op, ArrayRef<Value> operands,
                   ConversionPatternRewriter &rewriter) const final {
     auto loc = op->getLoc();
+    auto context = op->getContext();
     auto input_t = operands[0];
     auto filter_t = operands[1];
-    // TODO(sjw): add bias op linalg.
-    // auto bias_t = operands[2];
+    auto bias_mr = operands[2];
     auto results = op->getResults();
 
     // attach kernel attr to parent function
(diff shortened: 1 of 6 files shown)


jungpark-mlir (approved):
Looks good to me

mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
```diff
@@ -170,7 +171,44 @@ class ConvConverter final : public OpConversionPattern<tosa::Conv2DOp> {
...
 
-    rewriter.replaceOp(op, output_mr);
+    // test for zero bias, and ignore
+    auto bias_t = op.getOperand(2);
+    bool zero_bias = false;
+    if (auto cst = bias_t.getDefiningOp<ConstantOp>()) {
+      auto val = cst.getValue().cast<ElementsAttr>();
+      zero_bias = true;
+      for (auto ii = val.value_begin<APFloat>();
+           zero_bias && ii != val.value_end<APFloat>(); ++ii)
```
krzysz00: Question: what about bfloats (which are integers, not floats)?

krzysz00 (commented):
Just one question about the code

mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
jungpark-mlir: Are you sure they are treated as int? Bfloats are float with different number of mantissa and exponent, I suppose they should be float. Just not sure how far are they supported.
I think at least bf16 support is in (as float).

On the other hand, integer support may need totally different compilation path for its quantization.

krzysz00: In the MIOpen parts of our codebase, half-floats are stored as `f16`, while BFloats are stored as `int16`, and so a BFloat would be stored as an IntegerAttr and not a FloatAttr (though APFloat could be used to get a float out of it). 

I don't know how TOSA handles this, but I wanted to raise the issue.

jungpark-mlir: Interesting point, I didn't even know it's treated as short in miopen. I presume it's an (old) workaround when bfloat is not supported? I had a quick search and it looks both llvm and mlir have bf16 type now and better assume Tosa handles bf16 as float.
If not, we need to fix tosa.

And for miopen, if needed (especially if we want to support bf16 type for tosa->miopen path), for the short term; we could lower bf16 as int16 for miopen. Or we can fix it to use bf16 type.
All assuming there is no hidden problem using bf16 in mlir. We'd better check with @whchung

jungpark-mlir: @whchung 
Is there any special reason bf16 is treated as int16 in miopen?
I presume it's an early implementation and we can move to use bf16 type later but maybe I'm wrong.

PR #446, opened 2021-11-18
Add MIGraphX code object operation and pipelines

jungpark-mlir:


mlir/lib/Conversion/GPUToMIGraphX/GPUToMIGraphX.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+using namespace mlir;
+
+namespace {
+
+class FuncToCOBJPattern : public OpConversionPattern<CallOp> {
+  using OpConversionPattern<CallOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(CallOp op, ArrayRef<Value> operands,
```
sjw36, line 32: Does this assume all calls are to kernels?

mlir/lib/Conversion/GPUToMIGraphX/GPUToMIGraphX.cpp
```diff
@@ -0,0 +1,122 @@
...
+    SmallVector<Value, 8> cobjArgs;
+    mrOperands.push_back(resultAlloc);
+
+    SmallVector<IntegerAttr, 5> globalSizeAttr;
+    SmallVector<IntegerAttr, 5> localSizeAttr;
+    SymbolRefAttr kernelRefAttr;
+
+    auto fusedFuncOp =
+      op->getParentOfType<ModuleOp>().lookupSymbol<FuncOp>(fnAttr.getValue());
+
```
sjw36, line 46: Also check the kernelFunc is found and may have already been processed by other calls.

sjw36 (changes requested):

mlir/lib/Conversion/GPUToMIGraphX/GPUToMIGraphX.cpp
sjw36: Nevermind. This does not change the kernelFunc just the calls so it is correct.

sjw36 (approved):
Looks good.

PR #448, opened 2021-11-19
Add reboot in post build action

jerryyin:
This will help get rid of zombie process after mlir execution in each parallel stage.

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -92,6 +92,10 @@ void postProcessPerfRes() {
     archiveArtifacts artifacts: 'build/mlir_*.csv,build/perf-run-date', onlyIfSuccessful: true
 }
 
+def reboot(){
+    build job: 'reboot-slaves', propagate: false , parameters: [string(name: 'server', value: "${env.NODE_NAME}"),]
+}
+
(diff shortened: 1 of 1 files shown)


yiqian1:
Normally how long does it take to reboot? 3-5min?

jerryyin:
@yiqian1 It won't matter because all the reboot are at the end of the parallel step, so has no impact on execution of the stages themselves. The benefit it bring is that when jenkins pick up the machine to execute the next job, it guaranteed to have no zombie process.

okakarpa:
> Normally how long does it take to reboot? 3-5min?

The node should be back online in 5 mins unless there is an issue with hardware

yiqian1:
> 
Have you tested the change?

The reboot will kill all processes including the current jenkins process. Will the current jenkins process notify next step that "this step is done" before it gets killed?

jerryyin:
> The reboot will kill all processes including the current jenkins process. Will the current jenkins process notify next step that "this step is done" before it gets killed?

I'm not sure I get the question: The reboot only happens when the stage chain ends. What do you expect jenkins to do when the stage chain ends? There's no next step when it needs to reboot on the particular jenkins slave.

okakarpa (approved):

PR #452, opened 2021-11-30
Make the rocm_agent_enumerator test more precise.

pcf000:


jerryyin (commented):
LGTM, thank you!

Do you mind to also add the `/opt/rocm/bin/rocm_agent_enumerator` to the environment stage in jenkinsfile? Just so that we can see what the actual output is for the node.

mlir/utils/jenkins/miopen-tests/miopen_validate.sh
```diff
@@ -177,7 +175,7 @@ function run_tests() {
     rm "$TMPFILE"
     echo -n "$DTYPE $LAYOUT XDLOPS=$XDLOPS DIR=$DIRECTION: "
     if [[ $exit_status != 0 ]]; then
-        echo "$exit_status tests failed"
+        echo "$exit_status tests failed on $(hostname)"
```
jerryyin: I think we should revert this change given that it will not give additional meaningful information. Instead, https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/453 exposed the hostname earlier in the pipeline.

jerryyin (approved):

PR #453, opened 2021-11-30
Echo jenkins slave hostname in environment step

jerryyin:
This should expose the right hostname instead of previous docker container hash

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -37,7 +37,7 @@ void showEnv() {
 }
 
 String dockerArgs() {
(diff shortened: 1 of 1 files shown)


jerryyin:
With the above change we can see the node name earlier in the first row of the environment stage:

image

sjw36:
Can you also print the IP address.

jerryyin:
@sjw36 Most of them are qts machines so ip address is likely to be the same and it is port number that matters. I suggest not to expose the information here in jenkins but use a internal excel to track `hostname - ip - port mapping`.

yiqian1 (approved):

PR #455, opened 2021-12-07
Run docker container with a non-root user

yiqian1:
Remove user-specified "--user" and "-u 0" parameters from the docker command. Instead use the default -u added by Jenkins.

Set $HOME in Jenkinsfile

Use deleteDir() to clean up the workspace

Avoid using MIOpen-automatically-built libMLIRMIOpen

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -170,7 +170,7 @@ pipeline {
                      }
                      post {
                         always {
-                            deleteDir()
+                            cleanWs()
                         }
                     }
                 }
@@ -214,7 +214,7 @@ pipeline {
                     }
                     post {
                         always {
-                            deleteDir()
+                            cleanWs()
                         }
                     }
                 }
@@ -249,7 +249,7 @@ pipeline {
                                 checkout scm
                                 buildProject('libMLIRMIOpen',
                                              '-DBUILD_FAT_LIBMLIRMIOPEN=ON')
-                                cmake arguments: "--install . --component libMLIRMIOpen --prefix ${WORKSPACE}/libMLIRMIOpen",\
+                                cmake arguments: "--install . --component libMLIRMIOpen --prefix ${WORKSPACE}/MIOpenDeps",\
                                 installation: 'InSearchPath', workingDir: 'build'
                             }
                         }
@@ -266,13 +266,11 @@ pipeline {
                                 git branch: 'develop', poll: false,\
                                     url: 'https://github.com/ROCmSoftwarePlatform/MIOpen.git'
                                 sh 'sed -i "/ROCmSoftwarePlatform\\/llvm-project-mlir/d" ./requirements.txt'
-                                cmake arguments: "-P install_deps.cmake --minimum --prefix ${WORKSPACE}/MIOpen/MIOpenDeps",\
+                                cmake arguments: "-P install_deps.cmake --minimum --prefix ${WORKSPACE}/MIOpenDeps",\
                                     installation: "InSearchPath"
-                                sh "cp ${WORKSPACE}/libMLIRMIOpen/lib/libMLIRMIOpen.a ${WORKSPACE}/MIOpen/MIOpenDeps/lib"
-                                sh "cp ${WORKSPACE}/libMLIRMIOpen/include/Miir.h ${WORKSPACE}/MIOpen/MIOpenDeps/include"
                                 buildMIOpen('''-DMIOPEN_USE_MLIR=On
                                         -DMIOPEN_BACKEND=HIP
-                                        -DCMAKE_PREFIX_PATH="${WORKSPACE}/MIOpen/MIOpenDeps"
+                                        -DCMAKE_PREFIX_PATH="${WORKSPACE}/MIOpenDeps"
                                         -DMIOPEN_USER_DB_PATH=${WORKSPACE}/MIOpen/build/MIOpenUserDB
                                         -DMIOPEN_TEST_FLAGS=\'--verbose --disable-verification-cache\'
                                         ''')
@@ -342,7 +340,7 @@ pipeline {
                     }
                     post {
                         always {
-                            deleteDir()
+                            cleanWs()
                         }
                     }
                 }
@@ -404,7 +402,7 @@ pipeline {
                     }
                     post {
                         always {
-                            deleteDir()
+                            cleanWs()
                         }
                     }
                 }
@@ -480,7 +478,7 @@ pipeline {
                         }
                         post {
                             always {
-                                deleteDir()
+                                cleanWs()
                             }
                         }
                     }
@@ -556,7 +554,7 @@ pipeline {
             }
             post {
                 always {
-                    deleteDir()
+                    cleanWs()
                 }
              }
         }
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -35,10 +35,11 @@ void showEnv() {
     sh 'ulimit -a'
     sh '/opt/rocm/bin/rocm-smi'
     sh '/opt/rocm/bin/rocm_agent_enumerator'
+    sh 'id'
 }
 
 String dockerArgs() {
-    return '--user "$(id -u):$(id -g)" --device=/dev/kfd --device=/dev/dri --group-add video -u 0'
+    return "--device=/dev/kfd --device=/dev/dri --group-add video -v /etc/passwd:/etc/passwd:ro -v /etc/group:/etc/group:ro"
 }
 
 String dockerImage() {
@@ -60,7 +61,6 @@ void preMergeCheck() {
 
 void testMIOpenDriver(String dtype, String layout, String direction, String tuning) {
     echo "Config is (${dtype}, ${layout}, dir=${direction}, tuning=${tuning})"
-    sh 'ldconfig'
     dir('MIOpen/build/') {
         sh """
         bash ${WORKSPACE}/mlir/utils/jenkins/miopen-tests/miopen_validate.sh \
@@ -136,6 +136,9 @@ pipeline {
                             alwaysPull true
                         }
                     }
+                    environment {
+                        HOME="${WORKSPACE}"
+                    }
                     stages {
                         stage('Environment') {
                             steps {
@@ -167,7 +170,7 @@ pipeline {
                      }
                      post {
                         always {
-                            cleanWs()
+                            deleteDir()
                         }
                     }
                 }
@@ -186,6 +189,9 @@ pipeline {
                             label "${params.canXdlops ? 'gfx908' : 'rocm' }"
                             alwaysPull true
                         }
+                    }
+                    environment {
+                        HOME="${WORKSPACE}"
                     }
                     stages {
                         stage('Environment') {
@@ -208,7 +214,7 @@ pipeline {
                     }
                     post {
                         always {
-                            cleanWs()
+                            deleteDir()
                         }
                     }
                 }
@@ -230,6 +236,7 @@ pipeline {
                     }
                     environment {
                         PATH="/opt/rocm/llvm/bin:$PATH"
+                        HOME="${WORKSPACE}"
                     }
                     stages {
                         stage('Environment') {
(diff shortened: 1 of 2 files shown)


yiqian1:
Don't merge until Omkar verifies that deleteDir() does the job and userID is constant across the node.

krzysz00:
@yiqian1 I don't know if we need to rely on consistent user IDs - more on that Jenkins will unpack archived things to the correct user ID for the node being unpacked to.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -35,10 +35,11 @@ void showEnv() {
     sh 'ulimit -a'
     sh '/opt/rocm/bin/rocm-smi'
     sh '/opt/rocm/bin/rocm_agent_enumerator'
+    sh 'id'
 }
 
 String dockerArgs() {
-    return '--user "$(id -u):$(id -g)" --device=/dev/kfd --device=/dev/dri --group-add video -u 0'
+    return "--device=/dev/kfd --device=/dev/dri --group-add video -v /etc/passwd:/etc/passwd:ro -v /etc/group:/etc/group:ro"
```
krzysz00, line 42: I'm not sure we need to pass in /etc/passwd and /etc/group - what happens if we don't do that?

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -187,6 +190,9 @@ pipeline {
                             alwaysPull true
                         }
                     }
+                    environment {
+                        HOME="${WORKSPACE}"
```
krzysz00, line 194: Why?

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -208,7 +214,7 @@ pipeline {
                     }
                     post {
                         always {
-                            cleanWs()
+                            deleteDir()
```
krzysz00: Come to think of it, wouldn't it make sense to not delete the whole directory but just the build?

krzysz00 (commented):
There's a few changes here that don't feel necessary - why did you need to make them?

mlir/utils/jenkins/Jenkinsfile
yiqian1: I got something like this:
"groups: cannot find name for group ID"
"I have no name!"

mlir/utils/jenkins/Jenkinsfile
yiqian1: Our dockerfile sets HOME to /root.

mlir/utils/jenkins/Jenkinsfile
yiqian1: Why? We've finished all tasks. It's OK to delete the whole directory and leave.

mlir/utils/jenkins/Jenkinsfile
krzysz00: I think the better fix here is to not do that in the Dockerfile, or to unset HOME altogether. Do we _need_ a value of $HOME in this build?

mlir/utils/jenkins/Jenkinsfile
krzysz00: What command does this come from?

mlir/utils/jenkins/Jenkinsfile
krzysz00: On the other hand, that means we'd need to re-clone everything, no?

yiqian1:
> 

I'm sure if that's necessary either.

yiqian1:
> I don't know if we need to rely on consistent user IDs - more on that Jenkins will unpack archived things to the correct user ID for the node being unpacked to.

I'm not sure either.

mlir/utils/jenkins/Jenkinsfile
okakarpa: it's better to clear the workspace folder most of the times after certain number of  builds the workspace folder gets corrupted somehow causing the build to fail, Besides for mlir the job does shallow clone so the cloning is faster

mlir/utils/jenkins/Jenkinsfile
jerryyin: Agreed, on a second look I think `ENV HOME /root/` was a mistake I made (in combination with `-u 0`). I'm not sure why I explicitly set the root back then but it does seem to be a more robust fix to get rid of it al-together in the dockerfile.

mlir/utils/jenkins/Jenkinsfile
jerryyin: When we validated that `deleteDir()` can work, can we fall back to `cleanWs()` @okakarpa? I think it is still good for this task to go async as long as it still does its job.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -256,7 +263,14 @@ pipeline {
...
                             steps {
                                 dir('MIOpen') {
-                                getAndBuildMIOpen('--prefix ./MIOpenDeps', '''-DMIOPEN_USE_MLIR=On
+                                git branch: 'develop', poll: false,\
+                                    url: 'https://github.com/ROCmSoftwarePlatform/MIOpen.git'
+                                sh 'sed -i "/ROCmSoftwarePlatform\\/llvm-project-mlir/d" ./requirements.txt'
+                                cmake arguments: "-P install_deps.cmake --minimum --prefix ${WORKSPACE}/MIOpen/MIOpenDeps",\
+                                    installation: "InSearchPath"
+                                sh "cp ${WORKSPACE}/libMLIRMIOpen/lib/libMLIRMIOpen.a ${WORKSPACE}/MIOpen/MIOpenDeps/lib"
+                                sh "cp ${WORKSPACE}/libMLIRMIOpen/include/Miir.h ${WORKSPACE}/MIOpen/MIOpenDeps/include"
```
jerryyin: No, we should let `igemm build` stage to handle the installation not to manually do it like this.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -242,7 +249,7 @@ pipeline {
                                 checkout scm
                                 buildProject('libMLIRMIOpen',
                                              '-DBUILD_FAT_LIBMLIRMIOPEN=ON')
-                                cmake arguments: '--install . --component libMLIRMIOpen --prefix /opt/rocm',\
+                                cmake arguments: "--install . --component libMLIRMIOpen --prefix ${WORKSPACE}/libMLIRMIOpen",\
```
jerryyin: I prefer `/usr/local` as prefix over a local directory. This way, you wouldn't have to make another manual installation and MIOpen should be able to find this library in the right location.

yiqian1: /usr/local is owned by root.

mlir/utils/jenkins/Jenkinsfile
yiqian1: MIOpen needs to access $HOME when installing dependences (cget and sqlite). We can remove HOME from the Dockerfile, but we still need to set HOME here, since the default HOME for a jenkins user would be /home/jenkins (owned by root), while we want /home/jenkins/$JOBNAME, which is $WORKSPACE

mlir/utils/jenkins/Jenkinsfile
yiqian1: From the docker command without -v

mlir/utils/jenkins/Jenkinsfile
yiqian1: Alternatively I think I can install everything at ~/usr/local to avoid manual installation.

mlir/utils/jenkins/Jenkinsfile
okakarpa: Most likely cleanWs() should work as well since we are using the same uid as jenkins for deleting workspace 
once the PR is updated i can review the node workspace again

mlir/utils/jenkins/Jenkinsfile
jerryyin: Sounds good, I'm good as long as we don't need another manual installation.

krzysz00: @jerryyin What do you mean by "manual installation"?
I think having MIOpen go look for the libmlirmiopen.a file from MLIR's build directory is fine, as is it copying it into where it would be if the dependency was installed normally.

jerryyin: @krzysz00 I was referring to [this comment](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/455#discussion_r764936400) as the manual installation, as the practice there was to use another two bash script to move the installed targets again. This is brittle and will break if we altered the cmake installation logic.

jerryyin (commented):
LGTM, consider my review as approval. Please make sure Krzysztof and Omkar are also good so we are ready to go.

mlir/utils/jenkins/Jenkinsfile
krzysz00: Makes sense, I can see why we'd want to bind mount in this info

mlir/utils/jenkins/Jenkinsfile
krzysz00: I take it there isn't a way to do this in the Docker launch?

krzysz00 (approved):
LGTM, though I'd wait for omkar

mlir/utils/jenkins/Jenkinsfile
jerryyin: @krzysz00 Doing a google search show there is a better way to do it [here](https://stackoverflow.com/questions/30494050/how-do-i-pass-environment-variables-to-docker-containers). Seems like `-e` parameter is the way to go.

yiqian1: Not sure if '-e' would work. Another way is to add a parameter using ARG to the dockerfile, but then we need to rebuild docker every time when we start a docker container.

okakarpa:
> LGTM, consider my review as approval. Please make sure Krzysztof and Omkar are also good so we are ready to go.

The workspace is getting deleted with no trace for this PR

mlir/utils/jenkins/Jenkinsfile
krzysz00: @yiqian1 Can you try adding `-e HOME=${WORKSPACE}` (using double quotes for the Docker args) instead of these environment lines?

yiqian1: I've tried using '-e'. Neither double quotes nor single quotes worked. I also hard-coded the value of HOME, but that didn't change the default value of HOME in the docker container (i.e. '-e' didn't pass the HOME value into the docker container). Hmm, looks like it's not very straightforward of how '-e' works.

krzysz00: Just to check, did you try `-e HOME=${WORKSPACE}`?

yiqian1: Yes. You can see what I've tried [here](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/commits/testHome/mlir/utils/jenkins/Jenkinsfile)

krzysz00: I'm surprised https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/commit/af6bb211be89e14244eeeb8824b6ff152f4972ff#diff-16af82a543140fc9970902655eb9815da207d857b849d253ab13c4c0f344aa6a didn't work - what went wrong with it?

yiqian1: http://rocmhead.amd.com:8080/job/testHome/2/console
groovy.lang.MissingPropertyException: No such property: WORKSPACE for class: groovy.lang.Binding

jerryyin: Maybe you need to first define HOME as an environment variable before setting it. So maybe test it by locally adding a line of `ENV HOME /home`, and then reset it in invocation time, for example.

krzysz00: Nah, that's a Jenkins-side error.

One way to clean things up would be to make dockerArgs() take a `String homeDir` .parameter and pass in `"${WORKSPACE}"` there so that we're not having Jenkins set environment variables that could be reasonably delegated to the Docker container startup.

yiqian1: No luck with adding a dockerArgs() parameter. Got the same error:
groovy.lang.MissingPropertyException: No such property: WORKSPACE for class: groovy.lang.Binding

krzysz00: What I meant was that instead of substituting in WORKSPACE, to add a parameter to the function, substitute in that parameter, and then generally set that parameter to WORKSPACE (or even something interesting like /tmp)

yiqian1: That was exactly what I did in [this commit](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/commit/2225ab54e3a65096b40c0b42cdb148294f5799aa#diff-16af82a543140fc9970902655eb9815da207d857b849d253ab13c4c0f344aa6a)

And the jenkins job failed http://rocmhead.amd.com:8080/job/testHome/9/console

PR #412, opened 2021-10-18
#152 (Partition TOSA at conv2d) and #162 (Element-wise attribute for TOSA ops)

pcf000:
Look for tosa::Conv2D ops followed and/or preceded by element-wise ops, and pull them out into separate functions with a "kernel" attribute for later fusion.  Requires the invention of an element-wise attribute.

Also update some integration  tests for the external/llvm-project update.

PR #458, opened 2021-12-16
Make the merging of identical kernels part of the outlining process,

pcf000:
rather than a separate pass after the partition pass.

That makes the separate pass, and the follow-up symbolDCE pass,
unnecessary, and thus makes the partition-pipeline pass also
unnecessary.  It uses some of the logic of mergeIdenticalBlocks() but
different code, since it isn't manipulating Blocks.

Change --tosa-partition from function-level to module-level pass,
because it creates FuncOps.  There were possibilities for race
conditions there.  In conjunction, move candidates list from module- to
function-level to avoid some trouble with shared nodes.

The one catch is that we no longer catch identical kernel functions that
originate in different primary functions, since we have only a
function-level view instead of a module-level view when merging.


pcf000:
Improvements to the code in pull-request #412.

sjw36 (approved):
Looks good!

PR #459, opened 2021-12-16
Use index almost everywhere

krzysz00:
Now that vector supports indexing by index, there's no good reason to have i32 values in our lowering pipeline. The one exception, for now, is that the GPU OOB hack still requires i32 indices to work.

Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/384


whchung (approved):
Nice work!

whchung:
We need to monitor the performance impacts of this change though. By default index type of 64 bits wide, but we should have already set `index-bitwidth=32`.

krzysz00:
@whchung  Given that we have index-bitwidth=32 in all our lowering pipelines (or at least I'm pretty sure we do) this shouldn't have any performance impact?

PR #461, opened 2021-12-21
MLIR MIOpen test generator - miopen-gen

sjw36:
* extracted from mlir-miopen-driver
* allows for user defined (as input MLIR) kernel test generation
* fixed MIOpenToGPU, now converts calls to gpu.launch_func automatically
* cleanup Conv2dGenerator kernel name generation
* changed rocm-runner lib mcpuMemset5DHalfRandFloat to generate small numbers when not random to avoid overflow
* updated all mlir-miopen-driver tests

* still needs to run nightly and weekly

See: https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/368


mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
```diff
@@ -422,9 +425,9 @@ short randomIntegerValue(short min, short max) {
 }
 
 float randomFloatValue(short min, short max) {
-  float minAsF = static_cast<float>(min);
+  auto minAsF = static_cast<float>(min);
   if (min == max)
-    return minAsF;
+    return minAsF * 0.1f;
```
pcf000: ??  randomFloatValue(8, 9) returns a value between 8 and 9, but randomFloatValue(8, 8) returns 0.8?

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -0,0 +1,1752 @@
...
+    else
+      filterLayout.setValue("g" + filterLayoutValue);
+  }
+  if (outputLayoutValue.size() == 4) {
+    if (outputLayoutValue == "nkhw")
+      outputLayout.setValue("ngkhw");
+    else if (outputLayoutValue == "nhwk")
+      outputLayout.setValue("nhwgk");
+    else
+      outputLayout.setValue("g" + outputLayoutValue);
```
pcf000, line 375: Should this be '"n" + ...'?

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -0,0 +1,1752 @@
...
+  }
+
+  if (inputLayoutValue.size() == 4) {
+    if (inputLayoutValue == "nchw")
+      inputLayout.setValue("ngchw");
+    else if (inputLayoutValue == "nhwc")
+      inputLayout.setValue("nhwgc");
+    else
+      inputLayout.setValue("g" + inputLayoutValue);
```
pcf000, line 384: Also "n"?

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
sjw36: Yes, this is only generated for randomFloatValue(1, 1) -> 0.1.  This helps keep fp16 values from overflow in large convolutions.

mlir/tools/miopen-gen/miopen-gen.cpp
sjw36: No, this is adding the 'g'roup dimension when not present.

mlir/include/mlir/Dialect/MIOpen/Generator/Conv2dGenerator.h
```diff
@@ -55,10 +55,9 @@ class Conv2dGenerator {
   };
 
   Conv2dGenerator(const std::string &chip = "", const std::string &triple = "",
-                  const std::string &features = "",
-                  const std::string &perfConfig = "", int num_cu = 0,
+                  const std::string &features = "", int num_cu = 0,
                   bool xdlops = false,
-                  const miopen::ConvOpType operation = miopen::Conv2DOpType,
+                  const miopen::ConvOpType operation = miopen::NoOpType,
```
krzysz00: FYI, this enum is about to become a TableGen'd thing - and maybe this makes more sense as `const Optional<miopen::ConvOpType> operation = llvm::None`?

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.h
```diff
@@ -36,7 +36,12 @@ namespace mlir {
 
 namespace miopen {
 
-enum ConvOpType { Conv2DOpType, Conv2DBwdDataOpType, Conv2DBwdWeightOpType };
+enum ConvOpType {
+  Conv2DOpType,
+  Conv2DBwdDataOpType,
+  Conv2DBwdWeightOpType,
+  NoOpType
```
krzysz00: Why `NoOpType`? This'll add a bunch of weird switch/case in lowering?

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
```diff
@@ -178,19 +178,34 @@ void LowerMIOpenOpsToGPUPass::runOnOperation() {
...
 
-  auto processGpuKernelFunc = [&](FuncOp &theFunc,
-                                  OpBuilder &b) -> gpu::GPUFuncOp {
+  auto processGpuKernelFunc = [&](gpu::GPUModuleOp &gpuMod,
+                                  FuncOp &theFunc) -> gpu::GPUFuncOp {
+    // Set up the symbol table for the GPU ModuleOp.
+    SymbolTable gpuModuleSymbolTable(gpuMod);
+    // Reset builder insertion point to the beginning of the GPU module,
+    // as it would be modified inside the lambda.
+    OpBuilder b(gpuMod.getContext());
```
krzysz00, line 184: Might this also be a case for the scoped insertion point APIs? If not, you should probably pass the listener from the other builder as the second argument to the constructor.

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
```diff
@@ -213,6 +228,34 @@ void LowerMIOpenOpsToGPUPass::runOnOperation() {
...
     b.create<BranchOp>(loc, clonedFuncEntry);
 
+    // convert all calls to gpu.launch_func
+    SmallVector<CallOp, 4> calls;
+    op.walk([&](CallOp call) {
+      if (auto callable = call.getCallableForCallee()) {
+        if (FlatSymbolRefAttr symRef = callable.dyn_cast<SymbolRefAttr>()
+                                           .dyn_cast<FlatSymbolRefAttr>()) {
+          if (symRef.getValue() == theFunc.getName()) {
+            OpBuilder b(call);
```
krzysz00, line 235: I still argue we don't want to be creating new builders when we can temporarily modify the insertion point of the existing one.

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
```diff
@@ -213,6 +228,34 @@ void LowerMIOpenOpsToGPUPass::runOnOperation() {
...
+            auto blockVal = b.create<ConstantIndexOp>(loc, blockSize);
+            auto cst1 = b.create<ConstantIndexOp>(loc, 1);
+            auto dynamicSharedMemSize =
+                b.create<ConstantIntOp>(loc, 0, b.getI32Type());
+            gpu::KernelDim3 gridDims{gridVal, cst1, cst1};
+            gpu::KernelDim3 blockDims{blockVal, cst1, cst1};
+            b.create<gpu::LaunchFuncOp>(loc, gpuFunc, gridDims, blockDims,
+                                        dynamicSharedMemSize,
+                                        call.getArgOperands());
+            calls.push_back(call);
```
krzysz00, line 246: Hold on a minute - what's stopping this code from running on functions that go on the GPU but aren't kernels? How's that handled?

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
```diff
@@ -26,17 +26,17 @@ using namespace mlir;
...
+    const miopen::ConvOpType operation, const std::string &dataTypeStr,
+    int dilationHeight, int dilationWidth, int strideHeight, int strideWidth,
+    int paddingHeightLeft, int paddingHeightRight, int paddingWidthLeft,
+    int paddingWidthRight, const std::string &filterLayout,
+    const std::string &inputLayout, const std::string &outputLayout,
+    const std::string &kernelBaseName)
     : config{chip,
              triple,
              features,
-             perfConfig,
```
krzysz00, line 40: We still need perfConfig for tuning?

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
```diff
@@ -267,6 +270,8 @@ int Conv2dGenerator::getKernelCount() const {
     return 1;
   }
   switch (config.operation) {
+  case miopen::NoOpType:
+    break;
```
krzysz00: breaking into an unreachable doesn't feel right to me

mlir/lib/Dialect/MIOpen/MIOpenOps.cpp
```diff
@@ -46,11 +46,13 @@ Optional<ConvOpType> getConvOpTypeForName(const StringRef name) {
   if (name == "conv2d_bwd_weight") {
     return Conv2DBwdWeightOpType;
   }
-  return llvm::None;
+  return NoOpType;
```
krzysz00: This won't work in the future - this function has been removed and will be tablegen'd

krzysz00: If we want to explicitly have a noop operation in our system, that's another story, but that's not what I'm seeing so far

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
```diff
@@ -422,9 +427,9 @@ short randomIntegerValue(short min, short max) {
 }
 
 float randomFloatValue(short min, short max) {
-  float minAsF = static_cast<float>(min);
+  auto minAsF = static_cast<float>(min);
   if (min == max)
-    return minAsF;
+    return minAsF * 0.1f; // avoid inf
```
krzysz00, line 432: ... I don't like it. Wouldn't it be better to change the min and max in the cases that have this issue?

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
```diff
@@ -673,25 +678,22 @@ extern "C" void mcpuMemset5DHalfRandInt(
...
   for (unsigned i = 0; i < size0; ++i)
     for (unsigned j = 0; j < size1; ++j)
       for (unsigned k = 0; k < size2; ++k)
         for (unsigned l = 0; l < size3; ++l)
           for (unsigned m = 0; m < size4; ++m) {
-            value = randomIntegerValue(min, max);
+            float value = randomFloatValue(minf, maxf);
             aligned[i * stride0 + j * stride1 + k * stride2 + l * stride3 +
-                    m * stride4] = float_to_fp16(value, basetable, shifttable);
+                    m * stride4] = float_to_fp16(value);
```
krzysz00, line 696: For example, you could put the multiplication by 0.1 here

mlir/tools/miopen-gen/miopen-gen.cpp
krzysz00: Does that dimension always go on the beginning? I could've sworn I've seen cases where the `g` goes after `n` and so on

SJW: for some reason I cannot reply to this comment directly, but here is the answer.
- the dimension can be added anywhere, and we just add it with size=1.

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -0,0 +1,1758 @@
...
+      FuncOp::create(loc, StringRef(funcName), gpuWrapperFuncType);
+  module.push_back(gpuWrapperFunc);
+
+  // Emit gpu convolution logic.
+  auto block = gpuWrapperFunc.addEntryBlock();
+  b.setInsertionPoint(block, block->begin());
+
+  // Emit device selection
+  if (deviceNum.getNumOccurrences() > 0)
+    b.create<gpu::SetDefaultDeviceOp>(
```
krzysz00, line 659: Note to future us (not for this PR but): add an option that'll query the environment for how many GPUs there are and pick one in some reasonably fair way so that we can spread the load out over multiple GPUs

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -0,0 +1,1758 @@
...
+      } else if (elemFType.isBF16()) {
+        apVal.convert(llvm::APFloat::BFloat(), APFloat::rmNearestTiesToEven,
+                      &ignored);
+      }
+      return loopB.create<arith::ConstantFloatOp>(loc, apVal, elemFType);
+    };
+
+    // <test> = <cpu> - <gpu>
+    auto subfOp = loopB.create<arith::SubFOp>(loc, cpuLoadVal, gpuLoadVal);
+    auto divfOp = loopB.create<arith::DivFOp>(loc, subfOp, cpuLoadVal);
```
krzysz00, line 1377: Hang on a second, I don't think this is the right way to compare floats? Or at least something feels fishy about this bit.

Can we adapt https://stackoverflow.com/questions/4915462/how-should-i-do-floating-point-comparison#answer-32334103 instead?

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -0,0 +1,1758 @@
...
+  std::map<short, mlir::Value> i16vals;
+  auto getI16Val = [&](short v) {
+    if (i16vals.find(v) == i16vals.end()) {
+      auto i16Type = b.getIntegerType(16);
+      i16vals.emplace(v, b.create<arith::ConstantIntOp>(loc, v, i16Type));
+    }
+    return i16vals[v];
+  };
+
+  std::map<int32_t, mlir::Value> i32vals;
```
krzysz00: How many values in these maps? What's the range on the keys? Would one of the llvm map structures be more reasonable here?

krzysz00: Yeah, this feels like a `llvm::SmallDenseMap` candidate, at the very least

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -0,0 +1,1758 @@
...
+  OpBuilder builder(&context);
+  ModuleOp module;
+
+  verifyLayout();
+  correctParameters();
+  populateDefaults();
+
+  Conv2dGenerator conv2dGenerator;
+
+  std::list<KernelIF> kernels;
```
krzysz00: llvm::SmallVector, surely?

krzysz00 (changes requested):
Overall, nice refactor!

However,
1. Adding a noop arm to the ConvOpType enum is something I strongly disapprove of. This is why `Optional` exists.
2. I think something's off about the float comparison logic you wrote, which explains some of the issues and the need to pick weirdly large epsilons
3. The function to pick random floats shouldn't have a correction to shrink the values in that function. If anything, that means we need to change the default arguments to that function, possibly with respect to the data type.

But nitpicking about things that'll be surprising to future people aside, thanks for making this change!

mlir/include/mlir/Dialect/MIOpen/Generator/Conv2dGenerator.h
sjw36: Good, will give that a try.

krzysz00: Come to think it, under what circumstances will you call this generator without knowing which convolution you want - and wouldn't they instead justify erroring out beforehand instead?

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
sjw36: This only applies to Calls to theFunc which was converted above to gpu.func.

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
sjw36: Prefer not to affect external builders. This is lightweight and sets insertion point.

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
sjw36: Not necessary for func insertion.

krzysz00: How so? And what's not necessary?

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
krzysz00: https://mlir.llvm.org/doxygen/classmlir_1_1OpBuilder_1_1InsertionGuard.html is presumably an object for a reason

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
krzysz00: Oh, yeah, there it is - it was hiding outside the diff.

mlir/tools/miopen-gen/miopen-gen.cpp
sjw36: Basically the same except I use golden CPU value instead of adding both. Epsilon varies greatly from f16 to f32.

Ideally I would like the CPU computed with f16 for apples to apples.

mlir/include/mlir/Dialect/MIOpen/Generator/Conv2dGenerator.h
sjw36: It unfortunately carries some values (chip, triple, features, xdlops, etc) that are needed in host generation even when passing a user defined kernel, which may have an miopen.conv2d.

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
sjw36: It is passed in directly to the miopen::addPipeline in mlir-miopen-driver.

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
sjw36: should probably assert..

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
sjw36: yes, that can be generated in the host generator.

mlir/include/mlir/Dialect/MIOpen/Generator/Conv2dGenerator.h
krzysz00: Ok, so, in which case, either walk the kernel to find the ConvOpType or use an explicit Optional<> - I don't like adding what's effectively a null value to ConvOpType

mlir/tools/miopen-gen/miopen-gen.cpp
krzysz00: It looks a lot different, is what I'm noting. I might need to stare at the output to reassure myself, but, well, yeah

And true re eps - I can see that it went down to 1e-1 or 1e-2 and feel comfortable with that - since the epsilons I generally have on my fingers are 1e-5/1e-6 for singles and 1e-9/1e-10 for doubles, so that all progresses nicely by analogy.

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
sjw36: Lambda is meant to be a Functional Programming mechanism. Prefer to be additive only.

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
sjw36: Func is in global scope so it can be added anywhere in the parent module.

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
krzysz00: Fair enough. At the very least we should pull the listener through since it presumably does _something_.

And I don't think these lambdas are meant to be stateless, necessarily.

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
krzysz00: Yeah, true. I'd at least see an argument for picking a concrete spot (beginning or end) here, even though it doesn't matter.

But that's orthogonal-ish to whether to create a new builder or reuse one, which is fundamentally your call here.

mlir/tools/miopen-gen/miopen-gen.cpp
sjw36: Pretty insignificant, but ok.

mlir/tools/miopen-gen/miopen-gen.cpp
sjw36: Also insignificant. Done

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
sjw36: unfortunately the interface is 'short, short' which is why I had to put it here... very ugly. I will file a ticket to fix later.

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
krzysz00: But what if I have a perfConfig I want to test - maybe it's the result of a tuning failure.

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
krzysz00: Fair enough - and this at least applies just to fp16 it sounds like. I think we can live with it

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
sjw36: Actually will apply to f32 as well but just means smaller numbers.

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
sjw36: you can still pass it to mlir-miopen-driver, which is where it is used.

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
sjw36: actually applies to f32 as well but just means smaller numbers.

mlir/tools/miopen-gen/miopen-gen.cpp
krzysz00: Ok, yeah, I've stared at it, seems fine.

krzysz00 (commented):
Ok, so, my bigger complaints are addressed.

One question I have: a convConfig , as spit out by, for example, the tuning process, has both information about the problem size - used here, along with a perfConfig. If I want to test a convConfig, will I have to manually break it apart or pass it to _both_ miopen-gen and mlir-miopen-driver? I'm not sure I like either of those options, and I'd like to confirm that you can pass in a `convConfig` string and have that carry through to generating a kernel with the correct tuning parameters.

sjw36:
Check out this test for perfConfig usage changes: mlir/test/mlir-miopen-lib/perf_config.mlir

It can be passed through convConfig as in mlir-miopen-lib-test but generally for testing it should be passed through mlir-miopen-driver.

@jerryyin please also review this change.

sjw36:
jenkins: please retest

sjw36:
Finally have a green on full nightly testing. @krzysz00  please approve asap.

krzysz00 (approved):
I'm approving because this is in an overall good state and since it's both helpful and blocking a lot of stuff related to testing.

*However,* I'd like to see a followup PR in the very near future that addresses getting the perf config from `miopen-gen` to `mlir-miopen-driver` so that passing in a `--conv-config` argument to the test generator actually works as expected when debugging a tuning issue.

sjw36:
> I'm approving because this is in an overall good state and since it's both helpful and blocking a lot of stuff related to testing.
> 
> _However,_ I'd like to see a followup PR in the very near future that addresses getting the perf config from `miopen-gen` to `mlir-miopen-driver` so that passing in a `--conv-config` argument to the test generator actually works as expected when debugging a tuning issue.

@krzysz00 please file a ticket.

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
```diff
@@ -673,25 +678,22 @@ extern "C" void mcpuMemset5DHalfRandInt(
...
+  float minf = min;
+  float maxf = max;
+
   for (unsigned i = 0; i < size0; ++i)
     for (unsigned j = 0; j < size1; ++j)
       for (unsigned k = 0; k < size2; ++k)
         for (unsigned l = 0; l < size3; ++l)
           for (unsigned m = 0; m < size4; ++m) {
-            value = randomIntegerValue(min, max);
+            float value = randomFloatValue(minf, maxf);
```
whchung, line 694: This is INCORRECT. The signature of the function demands an integer.

krzysz00: Could this explain the need for really high tolerances as well?

(and if for some reason a float value is needed, cast after rand()?)

whchung: I wasn't aware of the context of high tolerances.

If a float value is needed, `@mcpuMemset5DHalfRandFloat` should be used instead of `@mcpuMemset5DHalfRandInt`

pcf000: randomFloatValue is the one with the 0.1 factor when minf and maxf are equal.  I bet that's making the difference, more than integer-vs-float.

PR #462, opened 2021-12-31
Use a custom attribute to store coordinate transforms and other refactors

krzysz00:
[This PR is not yet done, the tests need to be updated for the new syntax - although we do have a pass of all the nightly e2e tests. However, I'm sending it out as a draft for general feedback / nitpicking about names and to request we merge this before kpack if I can finish it up next week.]

Highlights:
- Each components of a coordinate transformation is now stored in a `TransformAttr`
- A complete set of transformations is packaged up into a `TransformsAttr` *along with the corresponding affine map*, thus making the affix-affine-transform pass unnecessary
- Information about GEMM padding (mExtra, nExtra, kExtra, and the backwards data padding kernel info) is now packaged into a `PaddingInfoAttr` which is an attribute of gemm operations or loads/stores
- Information about which lower dimensions to check for OOB stores is now stored in boolean attrs on loads and stores (and, intermediately, on GEMM)
- Load/store/copy ops that cannot interface with global memory are no longer capable of holding transformation attributes to reflect the fact that their inputs and outputs are naive tensors that cannot be transformed
- Several enums have been moved to the tablegen system. This includes the backwards data padding information, which is now a bitmap enum instead of an enumerated set of cases (changing emitStoreLogic to take the new enum is one of the things I haven't done yet)
- Backward data lowering and the v4r4 atomic add kernel have been moved into LowerMIOpenOps.cpp instead of headers. As a consequence, the `utility/` directory has been reorganized
- The (gridwise) GEMM ops do not carry attributes from the original convolution such as if padding was applied or what the strides were. This information generally wasn't used (except with the backward data padding kernel, but that's been moved up) and shouldn't have been present. (This allows gridwise_gemm to be used as a GEMM for C =A^T * B)
- `miopen.subview` has been deleted - it can be replaced with calls to `miopen.transform` and its lowering was full of ugly hacks
- We can now write unit tests

(Description to be edited if I run into anything else while wrapping this up)


whchung:
+ @asroy @zjing14 . This PR is a good direction to move the dialect and its C++ interface to a more declarative fashion.

krzysz00:
@sjw36 - I had to make a minor change to your linalg tiling alignment pass - could you take a look at that?

whchung:
@krzysz00 I'll give this PR another around of view before approving it.

krzysz00:
@whchung, no problem - it'll be good to have someone have actually looked at this in detail before it goes in

sjw36 (commented):
Looks good, nice cleanup!

mlir/lib/Dialect/MIOpen/MIOpenDialect.cpp
```diff
@@ -0,0 +1,496 @@
...
+    llvm::function_ref<mlir::InFlightDiagnostic()> emitError,
+    mlir::MLIRContext *context, TransformType type, ArrayRef<int64_t> params,
+    ArrayRef<StringRef> upperNames, ArrayRef<uint32_t> upperDims,
+    ArrayRef<StringRef> lowerNames, ArrayRef<uint32_t> lowerDims) {
+  return TransformAttr::getChecked(emitError, context, type, params, upperNames,
+                                   upperDims, lowerNames, lowerDims);
+}
+
+//===---------------------------------------------------------
+// TransformsAttr
```
whchung: nit: TransformMapAttr?

mlir/include/mlir/Dialect/MIOpen/MIOpen.h
```diff
@@ -0,0 +1,275 @@
...
+  llvm::SmallVector<int64_t, 8> startShape;
+
+  llvm::StringMap<uint32_t> endIndices;
+  llvm::SmallMapVector<uint32_t, SmallString<8>, 8> endNames;
+  llvm::SmallVector<int64_t, 8> endShape;
+
+  bool frozen = false;
+};
+
+class TopDownCTBuilder : public CoordTransformsBuilder {
```
whchung, line 184: When to use `TopDownCTBuilder` and when to use `BottomUpCTBuilder` shall be described.

mlir/lib/Dialect/MIOpen/Transforms/LowerMIOpenOps.cpp
```diff
@@ -107,13 +107,896 @@ const ArgumentFields Conv2DRewritePattern<miopen::Conv2DBwdWeightOp>::fields = {
...
+    gop->setAttr("k_per_thread", convOp->getAttr("k_per_thread"));
+    gop->setAttr("m_level0_cluster", convOp->getAttr("m_level0_cluster"));
+    gop->setAttr("m_level1_cluster", convOp->getAttr("m_level1_cluster"));
+    gop->setAttr("n_level0_cluster", convOp->getAttr("n_level0_cluster"));
+    gop->setAttr("n_level1_cluster", convOp->getAttr("n_level1_cluster"));
+  }
+}
+
+/// Lowerings for particular convolution algorithms (TODO, new file?)
+LogicalResult backwardWeightAtomicAdd(miopen::Conv2DBwdWeightOp op,
```
whchung, line 158: Can we safely assume the logic of the removed file for backward weight atomic kernel is located here?

mlir/lib/Dialect/MIOpen/Transforms/LowerMIOpenOps.cpp
```diff
@@ -107,13 +107,896 @@ const ArgumentFields Conv2DRewritePattern<miopen::Conv2DBwdWeightOp>::fields = {
...
+                    "shouldn't have been tried without them");
+  }
+
+  // Finally, erase the original Conv2D op.
+  op.erase();
+
+  return success();
+}
+
+LogicalResult backwardData(miopen::Conv2DBwdDataOp op, PatternRewriter &b) {
```
whchung, line 422: Can we safely assume the logic of the removed file for backward data padding kernel is located here?

mlir/unittests/Dialect/MIOpen/CoordTransformBuilderTests.cpp
```diff
@@ -0,0 +1,220 @@
...
+#include "gtest/internal/gtest-internal.h"
+
+using namespace mlir;
+using namespace mlir::miopen;
+
+//===----------------------------------------------------------------------===//
+// Test Fixture
+//===----------------------------------------------------------------------===//
+
+class CTBuilderTest : public ::testing::Test {
```
whchung, line 24: It's nice to see addition in `unittests` directory. Could you share some experiences / rationales between `test` and `unittests`?

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
```diff
@@ -1061,24 +973,22 @@ computeIndexDiffMap(OpBuilder &b, Location loc,
...
+
+    if (transformation == TransformType::Embed) {
       assert(e.size() == p.size());
       assert(q.size() == 1);
       Value lowerDiff = zeroConstantOp;
       for (unsigned iter = 0; iter < e.size(); ++iter) {
-        int64_t coefficient = e[iter].template cast<IntegerAttr>().getInt();
-        int64_t upperDim = p[iter].template cast<IntegerAttr>().getInt();
+        int64_t coefficient = e[iter];
+        uint32_t upperDim = p[iter];
```
whchung, line 991: It seems we enforce the rule to use `uint32_t` over `int64_t` / `unsigned` to index dimensions, correct?

whchung (commented):
I did a first iteration of checks. In general, very splendid cleanup! Getting rid of manually constructing attributes in the lowering logic really makes the code more comprehensible.

I think a documentation on when to use BottomUpCTBuilder, and when to use TopDownCTBuilder would be necessary.

In the next iteration I'll study the lowering logic of newly added classes: TransformAttr et al.

mlir/unittests/Dialect/MIOpen/CoordTransformBuilderTests.cpp
krzysz00: Well, the idea of the unit tests was to let me directly test functionality that wasn't necessarily exposed in the input/output to a pass, namely the coordinate transform builders themselves and how they generate affine maps. (Sure, we can check for a given convolution that we get the correct `transform_map`, but  I wanted to test the builders themselves before I started refactoring lowering).

mlir/include/mlir/Dialect/MIOpen/LowerMIOpenOps.h
krzysz00: That's what I've been trying to do where I could, as it made sense to me.

mlir/lib/Dialect/MIOpen/Transforms/LowerMIOpenOps.cpp
krzysz00: Yep

mlir/lib/Dialect/MIOpen/Transforms/LowerMIOpenOps.cpp
krzysz00: Yep

mlir/include/mlir/Dialect/MIOpen/MIOpen.h
```diff
@@ -0,0 +1,290 @@
...
+private:
+  llvm::StringMap<uint32_t> startIndices;
+  llvm::SmallVector<SmallString<8>, 8> startNames;
+  llvm::SmallVector<int64_t, 8> startShape;
+
+  llvm::StringMap<uint32_t> endIndices;
+  llvm::SmallMapVector<uint32_t, SmallString<8>, 8> endNames;
+  llvm::SmallVector<int64_t, 8> endShape;
+
+  bool frozen = false;
```
whchung, line 173: dead code?

whchung (approved):
Nice job done!

One thing I can think of worth amending is to document the syntax of `TransforMapAttr` and `Transform` in the TableGen definition, so we could consider upstream it later on.

mlir/include/mlir/Dialect/MIOpen/MIOpen.h
krzysz00: It shouldn't be flagging as such, and we do set `frozen=true;` in `get()`

krzysz00:
@whchung I'll add that documentation (of TransformAttr, at least, TransformMapAttr has an assemblyFormat) and then merge.

PR #465, opened 2022-01-06
Store a release build

yiqian1:
Build a specific release branch and store the bin, lib and include directories. We can add other stuffs later if needed, but these are the minimal set for MLIR to used as a standalone tool.


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -559,5 +560,51 @@ pipeline {
...
+                          -DBUILD_FAT_LIBMLIRMIOPEN=ON
+                          -DCMAKE_BUILD_TYPE=Release
+                          -DCMAKE_EXPORT_COMPILE_COMMANDS=1
+                        """)
+                        cmake arguments: "--install . --component libMLIRMIOpen --prefix ${WORKSPACE}/build/include",\
+                                installation: 'InSearchPath', workingDir: 'build'
+                        sh """#!/bin/bash
+                            temp=$params.releaseBranch
+                            release=\${temp////-}
+                            tar cvzf mlir-\$release.tar.gz ./build/bin ./build/lib ./build/include\
```
krzysz00: To clarify - we're building a binary release? Wouldn't it make more sense to use something like `CPack` here so that, on top of getting a tarball, we could also do something like make a `.deb` automatically if at some point we wanted to?

krzysz00 (commented):
Looks good, just had a thought

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -559,5 +560,51 @@ pipeline {
...
+                    args dockerArgs()
+                    label "${params.canXdlops ? 'gfx908' : 'rocm' }"
+                    alwaysPull true
+                }
+            }
+            stages{
+                stage ('Build and Archive a Release Branch') {
+                    steps {
+                        showEnv()
+                        sh "git checkout ${params.releaseBranch}"
```
krzysz00: There might be a way to do this with Jenkins - maybe in combination with skipDefaultCheckout?

mlir/utils/jenkins/Jenkinsfile
krzysz00: I'm not sold on the directory structure of this tarball - I'd want it to be `miir-x.y.z/{bin,include,lib,external}` and with a README and the like.

I'd also want to confirm that we don't end up with any concerning absolute paths in the resulting binaries.

krzysz00 (commented):
I've got a couple concerns here - do we want to talk  about this?

mlir/utils/jenkins/Jenkinsfile
yiqian1: Good point. Let me try it. It's not clear yet how we are going to use the stored package, but we'll figure it out later.

mlir/utils/jenkins/Jenkinsfile
yiqian1: It sounds doable. I'll give it a try too.

yiqian1:
Confirmed that this created an archive with relative paths.

krzysz00 (commented):
This all looks pretty good, but I have another weirdly fundamental question - why's this in CI? Wouldn't it make sense to have a `make_miir_release.sh` sitting around for when we need it?

Also, let's stick the README and the like into the tarballs

yiqian1:
> This all looks pretty good, but I have another weirdly fundamental question - why's this in CI? Wouldn't it make sense to have a `make_miir_release.sh` sitting around for when we need it?
> 
> Also, let's stick the README and the like into the tarballs

This is in CI because I don't want to maintain a script in each release branch.

krzysz00:
Wouldn't it make more sense for there to be a general `release_miir.sh` that picks up a version number from our CMake files etc etc.? (Heck, that's probably where CPack comes in). If that lived on our develop branch, it'd get pulled onto subsequent releases as part of the branching process and wouldn't clutter our CI with tasks that really aren't continuous or integration.

Also, in files I just remembered we should include - `LICENSE`

yiqian1:
> Wouldn't it make more sense for there to be a general `release_miir.sh` that picks up a version number from our CMake files etc etc.? (Heck, that's probably where CPack comes in). If that lived on our develop branch, it'd get pulled onto subsequent releases as part of the branching process and wouldn't clutter our CI with tasks that really aren't continuous or integration.
> 
Not sure how exactly the version number in CMAKE or CPack would work, but a general script that does what this PR does is easy to implement. The problem is that the existing release branches wouldn't have this script. We've decided only bug fixes can be merged into a release branch. So how should we handle existing release branches? Another questions is how do we want to maintain the script in release branches? Do we want to update it whenever the script in our develop branch changes?

> Also, in files I just remembered we should include - `LICENSE`

What should be included in LICENSE? Can I simply make a copy of mlir/LICENSE.txt?

krzysz00:
Re existing release branches: What I was thinking is that it'd be fine if we didn't have the script on the branch for the 5.1 release - if it comes to it, we can `git checkout miopen-dialect -- mlir/utils/make_release.sh` (IIRC) to run it there.

The release-making script would live on the development branch, and one part of doing a pre-release branch off would be to run and test the script (and do things like increment the version number *on the dev branch, before making the release branch*). Then, after all the bug fixes had gone in and it was release time, we'd check out the release branch and run the script.

(Depending on how we expect people to use these releases, there might be one commit on the branch that bumps the number from `X.Y-rc1` to `X.Y` but that's not critical to the idea.

Basically, I don't think making a release tarball should live in CI.

yiqian1:
> Re existing release branches: What I was thinking is that it'd be fine if we didn't have the script on the branch for the 5.1 release - if it comes to it, we can `git checkout miopen-dialect -- mlir/utils/make_release.sh` (IIRC) to run it there.
> 
> The release-making script would live on the development branch, and one part of doing a pre-release branch off would be to run and test the script (and do things like increment the version number _on the dev branch, before making the release branch_). Then, after all the bug fixes had gone in and it was release time, we'd check out the release branch and run the script.
> 
> (Depending on how we expect people to use these releases, there might be one commit on the branch that bumps the number from `X.Y-rc1` to `X.Y` but that's not critical to the idea.
> 

We definitely need a thorough plan for future releases. These are great thoughts. I will create a ticket to record the discussions here.

> Basically, I don't think making a release tarball should live in CI.

We need to store release tarballs on Jenkins servers, at least for now. So I feel it's OK to make a release tarball in CI until we figure out other ways to store a release build.

krzysz00:
Why do we need to store release tarballs on Jenkins?

Also, even though we just spent a bunch of time wrangling things into a single Jenkinsfile ... if this has to be a CI pipeline it should be its own file, I think, so it doesn't clutter up the output of all the PR/nightly/weekly builds

krzysz00:
... if nothing else, we can store our releases on _Github_

yiqian1:
> ... if nothing else, we can store our releases on _Github_

Yeah, Github was another option. I chose Jenkins simply because it can be easily integrated into our existing infrastructure. I used to have a separate Jenkinsfile for doing this, but Jerry prefers to have a single file for everything. I don't have a strong opinion. I actually prefer to do this in its own file.

krzysz00:
@jerryyin Can you provide some context please? (Or should this discussion be a meeting?)

jerryyin:
@krzysz00 Yes we can extend this into a discussion if necessary. Here are my two cents:
- Making a debian release package: 

Since we are not a independently release and distributed product, we should hold off from this. If we are to make release package, I want to make sure the requirements are clarified (because it seems to fall into QA or devops responsibility to maintain them.

- Create release in Jenkinsfile or bash script

I don't see a problem in either approach. I agree that if we are to add it to jenkinsfile it better be triggered by a different jenkins job, not by our nightly or weekly CI, but only triggered 1 time per release. Desirably, in that post release jenkins job (whether it is using jenkinsfile or bash under the hood), it needs to allow user to type in the release branch (and then hit to button to store the tarball).

- Store release in github

I argue we should do both: Once the release jenkinsfile finished building, manually download the tarball and upload it to github. This way the tarball exists both in jenkins and github.

mlir/utils/jenkins/Jenkinsfile.release
```diff
@@ -0,0 +1,96 @@
...
+                    environment {
+                        HOME="${WORKSPACE}"
+                        branch=sh(returnStdout:true, script:"""#!/bin/bash
+                            temp=$params.releaseBranch
+                            release=\${temp////-}
+                            echo \$release""").trim()
+                    }
+                    steps {
+                        showEnv()
+                        sh "git checkout ${params.releaseBranch}"
```
krzysz00, line 67: Could this be handled with `skipDefaultCheckout` and the Jenkins git plugin instead of `sh`?

mlir/utils/jenkins/Jenkinsfile.release
```diff
@@ -0,0 +1,96 @@
...
+                        sh "tar czf mlir-source-${branch}.tar.gz \$(ls)"
```
krzysz00: Would it be possible to have the tarball unpack into a directory. That is, instead of having the structure be `external/...`, `mlir/...`/, ..., we'd have `miir-${branch}/external/...` `miir-source-${branch}/mlir/...` and so on?

Tarballs that surprise-expand into the current directory are rather annoying, mostly because they're surprising.

krzysz00 (commented):
Overall LGTM, just one change I'd like to see and one comment that can go either way.

If you don't want to change anything, let me know and I'll approve

mlir/utils/jenkins/Jenkinsfile.release
yiqian1: I think the easiest way is temporarily move everything under a directory then create a tarball.

mlir/utils/jenkins/Jenkinsfile.release
yiqian1: I don't think so. We need to run Jenkinsfile in miopen-dialect not the one in the release branch.

mlir/utils/jenkins/Jenkinsfile.release
krzysz00: Looks like there's an option to tar for doing things like this https://serverfault.com/questions/86478/tar-how-to-create-a-tar-file-with-arbitrary-leading-directories-w-o-cding-to - I'd look at the second answer with `--transform`

Having tested out, something like this'll do 
```
tar --transform 's#^#a-1.0/#' -cvzf a-1.0.tar.gz *
```

also noting that * is probably better-behaved than $(ls)

mlir/utils/jenkins/Jenkinsfile.release
krzysz00: Fair enough, having poked around this does look like the easier solution

krzysz00 (approved):
LGTM, thanks so much for your work on this!

PR #466, opened 2022-01-07
Cherry-picking upstream commit 267bb194f3cedd88db0e7a5ea86169d669948e…

jerryyin:
…74 Remove old tc linalg ods generator

https://github.com/llvm/llvm-project/commit/267bb194f3cedd88db0e7a5ea86169d669948e74


krzysz00:
Do we want to get this on the dev branch first?

krzysz00 (approved):
Approved, but please also submit this change to the dev branch if it's not already there

jerryyin:
> Approved, but please also submit this change to the dev branch if it's not already there

@krzysz00 Thanks, yes the change is already there. This commit happens in September 30 last year and is already incorporated into the https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/456 from Paul.

However, it is still a stretch because I don't have a clue whether this can help the internal ticket [here](https://ontrack-internal.amd.com/browse/SWDEV-317683) without being able to reproduce locally.

krzysz00:
@jerryyin Can you have them test with this branch for a bit?

jerryyin:
@krzysz00 Nope, they don't have bandwidth for validation, especially because the failure happens once in more than ten run. Let me see what I can do further. In the mean time I will let the PR sit until I have better confidence this does fix the issue they have.

jerryyin:
FYI folks I did a second study on the failure in [SWDEV-317683](https://ontrack-internal.amd.com/browse/SWDEV-317683) and tend to seek for merge (backport) of this PR to the release branch.

I did mostly a comparison study of cmake generated artifacts and compared between:
 - The bad build in a previous 5.0 based commit, from QA provided docker
 - A good local build in a previous 5.0 based commit
 - A good local build in the `release/rocm-5.1` branch
 - A good local build in this PR's patch branch

The conclusion is that: 
 - The build artifact stay the same between bad/good builds. It does not seem possible to deduct what went wrong at the build exactly by just looking at the build-time generated files.
 - This patch get rid of the `LinalgNamedStructuredOpsSpec.tc` generation, which together with the failing target, `LinalgNamedStructureOps.yaml` both belong to `LinalgNamedStructureOps`. From the look of it, it makes the `LinalgNamedStructureOps` builds faster and therefore less likely to encounter such race conditions.

Therefore, please review and give approval/leave feedback.

krzysz00:
I think a cherry-pick from upstream is, overall, a reasonable thing to do and I don't expect it'll cause us problems ... assuming that the results of the old generator aren't something we _use_, but since you've tested that it should be fine.

jerryyin:
@mvermeulen @whchung Gentle ping on this backport PR.

mvermeulen (approved):
Approve given context - including potential effect on race condition (as well as already in our dev branch being exercised).

PR #467, opened 2022-01-10
[DO NOT SQUASH] Add a dump asm option

krzysz00:
[MLIR][GPU] Add a dump-isa pass option to GPU serialization.

The dump-isa pass option allows developers to get the compiled ISA for their kernels, or, if they used an option such as --stop-after, to extract the intermediate IR from the partially-completed compilation process. This simplifies the debugging of compilation issues arising from the serialization of GPU code to binary blobs.

(Rationale: needing to go in and add a call to print the ISA out and then take it back out again before committing was getting to be a pain.)

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/GPU/Passes.h
+++ b/external/llvm-project/mlir/include/mlir/Dialect/GPU/Passes.h
@@ -88,6 +88,12 @@ protected:
       *this, "gpu-binary-annotation",
       llvm::cl::desc("Annotation attribute string for GPU binary"),
       llvm::cl::init(getDefaultGpuBinaryAnnotation())};
+  Option<bool> dumpAsm{
+      *this, "dump-asm",
+      llvm::cl::desc("Whether the final generated instructions or intermediate "
+                     "IR (if stopping early) for a kernel should be dumped to "
+                     "the debug stream"),
+      llvm::cl::init(false)};
 };
 } // namespace gpu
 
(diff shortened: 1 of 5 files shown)


krzysz00:
jenkins: retest this please

pcf000 (commented):
I hate the name -- "ISA" is the architecture, not a set of instructions in that architecture -- but I also recognise that the existing code already uses it that way.  Ick.  For Hsaco, it's asm, but I can't tell what it is for CUDA.

Otherwise looks fine, aside from spelling errors in comments, about which I'm regrettably pedantic.  "Simultaneously" and "separate."

external/llvm-project/mlir/include/mlir/Dialect/GPU/Passes.h
```diff
@@ -88,6 +88,11 @@ class SerializeToBlobPass : public OperationPass<gpu::GPUModuleOp> {
       *this, "gpu-binary-annotation",
       llvm::cl::desc("Annotation attribute string for GPU binary"),
       llvm::cl::init(getDefaultGpuBinaryAnnotation())};
+  Option<bool> dumpISA{
+      *this, "dump-isa",
+      llvm::cl::desc("Whether the assempled ISA or intermediate IR for a kernel"
```
pcf000: "assembled"

krzysz00:
@pcf000 I see no reason not to apply your naming nitpick and even agree with it, so changed. Also, thanks for catching the typos, they sounded pretty similar over here

(I also extended the test suite to make sure this option works)

pcf000 (approved):
Still looks good to me.

PR #469, opened 2022-01-17
Remove LowerMIOpenOps.h

krzysz00:
This commit doesn't actually change any part of the code. However, it moves all the rewrite patterns into a few separate files to get rid of the monster header file that must be compiled twice.

The one problem with this PR is that anyone messing with LowerMIOpenOps.h (so #423, I think) will need to migrate their changes to the correct file(s) once the merge button gets hit.


mlir/lib/Dialect/MIOpen/Transforms/LowerMIOpenOps.cpp
```diff
@@ -15,51 +15,32 @@
 // limitations under the License.
 // =============================================================================
 //
-// This pass converts miopen.conv2d into miopen.transform and
-// miopen.gridwise_gemm.
+// These passes convert the MIOpen threadwise ops into constructs from the
+// rest of MLIR so that they can be lowered to the GPU and LLVM dialects.
```
jerryyin: Maybe `ThreadwiseGemmToStd.cpp` to be consistent with the rest of the file names? `LowerMIOpenOps.cpp` seems too generic compared to what it does now.

mlir/include/mlir/Conversion/MIOpenPasses.td
```diff
@@ -38,24 +38,6 @@ def ConvertMIOpenToGPU : Pass<"convert-miopen-to-gpu", "ModuleOp"> {
   ];
 }
 
-//===----------------------------------------------------------------------===//
-// MIOpenToGPU within an existing GPUModule
-//===----------------------------------------------------------------------===//
-
-def ConvertMIOpenWithinGPUModule : Pass<"test-miopen-lowering-gpu-module", "gpu::GPUModuleOp"> {
-  let summary = "Lower the operations from the MIOpen dialect within an "
```
jerryyin, line 46: This pass doesn't have a consumer in our lowering pipeline, which is why you removed it?

jerryyin (approved):
LGTM, you made the refactor that I want to make for a long time. Good job!

Please make sure @whchung also agree and approve. It may make more sense for his change to merge before this PR.

mlir/lib/Dialect/MIOpen/Transforms/LowerMIOpenOps.cpp
krzysz00: ... good point, I can do that.

mlir/include/mlir/Conversion/MIOpenPasses.td
krzysz00: This pass was the main reason we still had a `LowerMIOpenOps.h`, isn't consumed anywhere but some old test code, and was very much not being maintained to keep up with pipeline changes

jerryyin: Sounds good.

whchung:
can we revert this commit?

whchung:
this PR is nullifying my efforts in PR #423 once again. I have just recovered from illness and fixed the merge conflict with upstream and this PR was merged without my consent.

krzysz00:
@whchung I can revert, but this PR is, at its core, a reorganizational one - are the merge conflicts particularly difficult to handle? That is, would it be faster/better to file the revert, wait for tests to go through, etc etc. or to move the kpack logic into the relevant c++ files?

whchung:
@krzysz00 I have NOT reviewed the PR yet, therefore I don't know the scope of the changes introduced in this PR. The merge commit was a direct violation to the review comment posted by @jerryyin .

krzysz00:
Fair enough, my bad

I've filed the revert, though I'd like to see this go through instead of bouncing around between reverted and unreverted.

whchung:
We can re-introduce changes in this PR, after at least I get a full cycle of CI tests in PR #423, and after I get better understand about how you split the logic from one header file to several source files.

#423 is almost usable, and was already used in the latest iteration of weekly CI already to give us a better perfDB geared toward KPACK. And this PR poses another unnecessary hurdle to the merger of it in my opinion.

krzysz00:
Agreed - I wasn't sure how close KPack was to going in.

PR #471, opened 2022-01-17
Fix remaining MI-200 failures with broad workaround

krzysz00:
This change disables the instruction scheduler in the compiler backend, which """fixes""" the remaining fp16 xdlops issues we're having. See https://ontrack-internal.amd.com/browse/SWDEV-318900 for more info and **don't merge this**.

### DIFF
--- a/external/llvm-project/llvm/lib/Target/AMDGPU/GCNSubtarget.h
+++ b/external/llvm-project/llvm/lib/Target/AMDGPU/GCNSubtarget.h
@@ -750,7 +750,10 @@ public:
   Align getStackAlignment() const { return Align(16); }
 
   bool enableMachineScheduler() const override {
-    return true;
+    // XXX(kdrewnia): Don't merge this initially, and don't merge without
+    // discussion of cleaner hammers. This will get past the MI-200 test
+    // failures on fp16 xdlops at the cost of performance everywhere
+    return false;
   }
 
   bool useAA() const override;


pcf000:
I can attest that it fixes the three failing tests that I saw on lockhart6.  It also roughly doubles or triples the testing time when doing check-mlir-miopen with nightly-style options, from two minutes to five (but I haven't measured the performance tests which are probably a bigger deal).

jungpark-mlir:
Nice debugging @krzysz00 and looks very interesting (weird) restriction. 
Assuming current experiment in the ticket goes well, I'd ask them if we can have a new variant of the MFMA intrinsics explicitly declaring as MAC (src==dst) while restrict the existing MFMA as MAD (src != dst).
Actually, this is not a necessary proposal because it's quite possible for backend compiler to have such an analysis but not sure how easy for them to implement it. It's rather easier for us to specify MAC or MAD from user side, we can suggest this option if they struggle to find their time for the further work.

krzysz00:
@jungpark-mlir I'm not sure we have the knowledge to make that call intelligently (and that's already something like the workaround they're looking at)

jungpark-mlir:
I mean reusing the destination vector to feed back to the mfma instruction from our code,  I suppose we already do it in the loop. Backend compiler doesn't simply reuse the same vector, instead the destination vector is determined by register allocator.
Such analysis (MAD or MAC) in the backend is not trivial but should be much easier if the input program gives hint by using a specific intrinsic. (I expect they also don't like new intrinsic either though)

Let's see what they get back.

jungpark-mlir:
Nevermind, HasOnUse simply did the magic in the backend. Hope everything works well with the fix.

PR #399, opened 2021-10-03
Refactor Jenkinsfile

yiqian1:
- Refactor MIOpen-build stages that are needed by both performance tests and MIOpen validations

- Make E2E tests with fixed data and random data in parallel to reduce the execution time.

- Increase the timeout limit for tuning MLIR kernels.

Test CIs:
nightly+sharedLib+staticLib+perfTest
http://ml-ci.amd.com:21096/blue/organizations/jenkins/MLIR%2Frefactor-Jenkinsfile/detail/refactor-Jenkinsfile/24/pipeline

nightly + perfTest
http://ml-ci.amd.com:21096/blue/organizations/jenkins/MLIR%2Frefactor-Jenkinsfile/detail/refactor-Jenkinsfile/20/pipeline

nightly+staticLib
http://ml-ci.amd.com:21096/blue/organizations/jenkins/MLIR%2Frefactor-Jenkinsfile/detail/refactor-Jenkinsfile/19/pipeline

nightly+sharedLib
http://ml-ci.amd.com:21096/blue/organizations/jenkins/MLIR%2Frefactor-Jenkinsfile/detail/refactor-Jenkinsfile/18/pipeline

PR #472, opened 2022-01-19
Include constants, in any operand position, in frontOps.

pcf000:
We already follow first-operand element-wise-op chains back from the conv2d that is the basis for the partition.  This commit adds a special case to include constant operands in any position.


pcf000:
Private issue #399.

sjw36:
@pcf000 this can be closed right?

pcf000:
Yes.  I was just waiting till your #500 went through.

pcf000:
Absorbed into pull request #500 .

PR #473, opened 2022-01-19
Disable fixed-E2E-tests in weeklies

yiqian1:
The fixed-E2E-tests should be run in nightlies or PR.


krzysz00 (approved):
LGTM, especially since nightlies run on the weekly days anyway.

krzysz00 (approved):
LGTM, especially since I've now checked that nothing actually depends on the results of that previous build and nightlies run on the weekly days.

PR #475, opened 2022-01-20
Dump asm as in upstream

krzysz00:
This is a commit to harmonize the assembly dumping infrastructure with how the upstream folks wanted it rewritten.

See below for new commit message there:

[MLIR][GPU] Add debug output to enable dumping GPU assembly

- Set the DEBUG_TYPE of SerializeToBlob to serialize-to-blob
- Add debug output to print the assembly or PTX for GPU modules before
  they are assembled and linked

Note that, as SerializeToBlob is a superclass of SerializeToCubin and
SerializeToHsaco, --debug-only=serialize-to-blom will dump the
intermediate compiler result for both of these passes.

In addition, if LLVM options such as --stop-after are used to control
the GPU kernel compilation process, the debug output will contain the
appropriate intermediate IR.

Differential Revision: https://reviews.llvm.org/D117519

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/GPU/Passes.h
+++ b/external/llvm-project/mlir/include/mlir/Dialect/GPU/Passes.h
@@ -88,12 +88,6 @@ protected:
       *this, "gpu-binary-annotation",
       llvm::cl::desc("Annotation attribute string for GPU binary"),
       llvm::cl::init(getDefaultGpuBinaryAnnotation())};
-  Option<bool> dumpAsm{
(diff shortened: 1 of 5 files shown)


pcf000:
Looks good to me.  It's basically replacing your first version with LLVM_DEBUG and a special setting of DEBUG_TYPE.

pcf000 (approved):
Looks good to me. It's basically replacing your first version with LLVM_DEBUG and a special setting of DEBUG_TYPE.

PR #404, opened 2021-10-06
Refactor jenkinsfile

okakarpa:
Refactor MIOpen-build stages that are needed by both performance tests and MIOpen validations

Make E2E tests with fixed data and random data in parallel to reduce the execution time.

Increase the timeout limit for tuning MLIR kernels.

Test CIs:
nightly+sharedLib+staticLib+perfTest
http://ml-ci.amd.com:21096/blue/organizations/jenkins/MLIR%2Frefactor-Jenkinsfile/detail/refactor-Jenkinsfile/24/pipeline

nightly + perfTest
http://ml-ci.amd.com:21096/blue/organizations/jenkins/MLIR%2Frefactor-Jenkinsfile/detail/refactor-Jenkinsfile/20/pipeline

nightly+staticLib
http://ml-ci.amd.com:21096/blue/organizations/jenkins/MLIR%2Frefactor-Jenkinsfile/detail/refactor-Jenkinsfile/19/pipeline

nightly+sharedLib
http://ml-ci.amd.com:21096/blue/organizations/jenkins/MLIR%2Frefactor-Jenkinsfile/detail/refactor-Jenkinsfile/18/pipeline

This is the same PR as #399. I had to close the old one and open a new PR in order to make the status checks work for the new stages.

PR #478, opened 2022-01-21
Remove hardwired twist from wrw atomic add

krzysz00:
The way the WRW atomic add kernel, every tensor had its layout forced to GNCHW before further transformations were performed. I theorized this would have a negative impact on NHWC atomic add performance - the related ticket is [https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/404](#404). Earlier meetings with MIOpen clarified that they don't do this.

Doing the same to backward data will be a separate PR so as to keep changes at least somewhat atomic where possible.

As to performance results, I ran test 7 (the one that hits this kernel and has padding) from the PR test through the benchmarking script and got the following numbers:

| | NCHW | NHWC |
| ------------ | ------ | ------ |
| Pre-change | 27.50 | 21.83 |
| Post-change | 27.53 | **26.60** |

That is a 21.8% performance increase with little effect on the NCHW performance.

The change that isn't happening in this PR (again for atomicity) is sorting out the ordering of G and N0/KBlock - ideally, if we had a GN... layout, we'd have the gemmG dimension be G x {kBlock/N0}, but for an NG... layout, we'd go with gemmG = {N0 x G}, so that steps in the gemmG dimension aren't twisted out of order. (since kBlock is something we declare into existence, all we have to make sure is that we match the ordering the input and output tensors use). I suspect that'd give us a very minor performance win, if any, but it's worth noting what hasn't happened.


mlir/lib/Dialect/MIOpen/Transforms/LowerMIOpenOps.cpp
```diff
@@ -292,15 +295,22 @@ LogicalResult backwardWeightAtomicAdd(miopen::Conv2DBwdWeightOp op,
 
   // Transform input tensor
   {
-    // Force the layout of the input tensor to GNCHW, pad H and W, and split n
-    // into n0 and n1 where n0 has size kBlocks and n1 is what's left
+    // Pad H and W and split N into  n0 and n1 where n0 has size kBlocks and n1
```
whchung, line 301: This approach looks promising. We certainly don't need to force GNCHW here.

mlir/test/Dialect/MIOpen/lowering_wrw_atomic_transform.mlir
```diff
@@ -7,8 +7,8 @@ module  {
...
-// CHECK: #transform_map{{[0-9]+}} = #miopen.transform_map<affine_map<(d0, d1, d2) -> (d0 floordiv 2, d0 mod 2, d1 floordiv 81, d2, (d1 mod 81) floordiv 9, d1 mod 9)> by[#miopen.transform<Merge{1, 2} ["gemmG"] at [0] -> ["go", "n0"] at [0, 1]>, #miopen.transform<Merge{16, 9, 9} ["gemmK"] at [1] -> ["n1", "ho", "wo"] at [2, 4, 5]>, #miopen.transform<PassThrough ["gemmM"] at [2] -> ["ko"] at [3]>] bounds = [2, 1296, 32] -> [1, 2, 16, 32, 9, 9]>
+// CHECK: #transform_map{{[0-9]+}} = #miopen.transform_map<affine_map<(d0, d1, d2, d3, d4, d5) -> (d0 * 16 + d1, d2, d3, d4 - 2, d5 - 2)> by[#miopen.transform<PassThrough ["gi"] at [2] -> ["gi"] at [1]>, #miopen.transform<Unmerge{2, 16} ["n0", "n1"] at [0, 1] -> ["ni"] at [0]>, #miopen.transform<PassThrough ["ci"] at [3] -> ["ci"] at [2]>, #miopen.transform<Pad{2, 2, 2, 2} ["hipad", "wipad"] at [4, 5] -> ["hi", "wi"] at [3, 4]>] bounds = [2, 16, 1, 32, 11, 11] -> [32, 1, 32, 7, 7]>
```
whchung, line 10: the unit test here only ues NCHW, can we add another NHWC case, so perhaps it can be observed from the `#transform_map` attributes that a simpler transformation is derived?

whchung:
@ltqin , please provide your review comments for this PR. This change could potentially improve performance of MIOpen C++ kernels as well.

krzysz00:
To my understanding, this change already exists in MIOpen - the relayout forcing was an artifact of porting the kernel to MLIR naively

mlir/test/Dialect/MIOpen/lowering_wrw_atomic_transform.mlir
krzysz00: Done

whchung (approved):
LGTM.

@ltqin. Please do help provide your feedbacks for this PR. Thanks.

@krzysz00, I do not know if @ltqin is available to review the PR. You may merge it even without his approval.

ltqin:
@krzysz00  I'm sorry, this project has changed a lot from the original one. so I can't rewiew this PR on time. I can only provide some ideas: 1 After split K, if  the gridsize is a multiple of 120(for example gridsize = 360 or 720)  , the efficiency will be higher.  2 The efficiency of nhwc should be faster than nchw, so I am a little confused about the result of this pr.

krzysz00:
@ltqin The situation, from what I can see, is that the way the WRW atomic add kernel was ported to MLIR added a transformation that made all the tensors look GNCHW/GKCYX/... before the GEMM was created. This prevented vectorization opportunities, since, supposing we'd started with a GKYXC filter layout - which was forced to GKCYX, the GEMM results would map `(g, m, n) -> (g / KBlock, m, n / YX, (n % YX) / X, n % X) -> (g, k, (n % YX) / X, n % X, N / YX)`, which means that a step by 1 in *any* of the GEMM dimensions doesn't correspond to a step by 1 in the tensor memory.

This PR keeps the native layout before going to GEMM, as in MIOpen (I think), so the map for the filter tensor, for example, is `(g, m, n) -> (g / KBlock, m, n / XC, (n % XC) / C, n % C)` which can be vectorized.

PR #479, opened 2022-01-21
Make the build order of miopen-gen be deterministic.

whchung:
Since PR #461, miopen-gen has become a required tool. Make it be built
explicitly in "Build MLIR" stage to prevent non-determinisim.


whchung:
In the hindsight, #461 should have been checked with not only nightly CI but also weekly CI due to its sheer scope.

yiqian1 (approved):

krzysz00 (approved):

PR #480, opened 2022-01-21
Stop forcing NCHW in backward data kernels

krzysz00:
This is a followup to #478 that applies the same untwisting to backward data and introduces some useful abstractions for coordinate transformations that use embed().

Unlike in the WRW case, performance seems mostly flat, probably because backward data is already extremely scalar on its way in and out of memory.

(On that note, I should go make sure that WRW atomic add is getting better vectorization parameters on NHWC later)


whchung (approved):
LGTM. Thanks for cleaning this up!

PR #379, opened 2021-09-16
Reorder compilation process so `-miopen-affix-params` becomes the first pass.

whchung:
Reorder the lowering pipeline to `-miopen-affix-params` could become the very first pass.

This PR aims to fulfill https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/337

PR #481, opened 2022-01-24
[DO NOT SQUASH] Make the BF16 conversion ops elementwise

krzysz00:
- Abstract out the TLike interfaces in OpBase.td (this'll get submitted upstream)
- Make both miopen.data_convert and gpu.bf_convert take vector arguments
- Push the vector arguments through to LLVM so so the optimizer has opportunities to be clever and so I don't have to add the elementwise op unrolling pass to GPUToROCDL

This is a minor PR that cleans up the annoyance living at [issue #379](https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/379) and makes the data op conversion code more uniform.


whchung (approved):
LGTM

PR #485, opened 2022-01-26
Adding i8 suport from gpu to rocdl conversion

jerryyin:
This PR implements https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/411. It looks like ROCDL->LLVM conversion is already done so the only thing necessary left is GPU->ROCDL conversion.

### Background: all i8 xdlops

```llvm
llvm.amdgcn.mfma.i32.32x32x4i8
llvm.amdgcn.mfma.i32.16x16x4i8
llvm.amdgcn.mfma.i32.4x4x4i8
llvm.amdgcn.mfma.i32.32x32x8i8
llvm.amdgcn.mfma.i32.16x16x16i8
```

Among the only useful ones are the last two with reduction (Others may be deprecated in future hardware, thanks for tips from @zjing14)
```llvm
llvm.amdgcn.mfma.i32.32x32x8i8
llvm.amdgcn.mfma.i32.16x16x16i8
```

I worked in the following components in this PR:
 - GPU ops tablegen: Added i8 support to gpu.mfma op
 - LowerGpuOpsToROCDLOps: Added i8 support to two mfma instructions
 - Unit tests including:
   - gpu to rocdl lowering
   - gpu.mfma parsing tests

Note: all components touched are in `./external` repo and are upstream-able.

### DIFF
--- a/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
+++ b/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
@@ -875,6 +875,17 @@ struct MFMAOpLowering : ConvertToLLVMPattern {
             op, adaptor.destC().getType(),
             ValueRange{adaptor.sourceA(), adaptor.sourceB(), adaptor.destC(),
                        immValues[0], immValues[1], immValues[2]});
+    } else if (mfmaInstr.endswith("i8")) {
+      if (mfmaInstr == "mfma_i32_32x32x8i8")
+        rewriter.replaceOpWithNewOp<ROCDL::mfma_i32_32x32x8i8>(
+            op, adaptor.destC().getType(),
+            ValueRange{adaptor.sourceA(), adaptor.sourceB(), adaptor.destC(),
+                       immValues[0], immValues[1], immValues[2]});
+      else if (mfmaInstr == "mfma_i32_16x16x16i8")
+        rewriter.replaceOpWithNewOp<ROCDL::mfma_i32_16x16x16i8>(
+            op, adaptor.destC().getType(),
+            ValueRange{adaptor.sourceA(), adaptor.sourceB(), adaptor.destC(),
+                       immValues[0], immValues[1], immValues[2]});
     }
 
     return success();
--- a/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
+++ b/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
@@ -886,6 +886,9 @@ struct MFMAOpLowering : ConvertToLLVMPattern {
             op, adaptor.destC().getType(),
             ValueRange{adaptor.sourceA(), adaptor.sourceB(), adaptor.destC(),
                        immValues[0], immValues[1], immValues[2]});
+      else {
+        return failure();
+      }
     }
 
     return success();
--- a/external/llvm-project/mlir/test/Conversion/GPUToROCDL/gpu-to-rocdl.mlir
+++ b/external/llvm-project/mlir/test/Conversion/GPUToROCDL/gpu-to-rocdl.mlir
@@ -1,7 +1,7 @@
 // RUN: mlir-opt %s -convert-gpu-to-rocdl -split-input-file | FileCheck %s
 // RUN: mlir-opt %s -convert-gpu-to-rocdl='index-bitwidth=32' -split-input-file | FileCheck --check-prefix=CHECK32 %s
 
-gpu.module @test_module {
+gpu.module @test_module_index_op {
   // CHECK-LABEL: func @gpu_index_ops()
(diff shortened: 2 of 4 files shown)


external/llvm-project/mlir/test/Conversion/GPUToROCDL/gpu-to-rocdl.mlir
```diff
@@ -277,7 +277,7 @@ gpu.module @test_module {
 
 // -----
 
-gpu.module @test_module {
+gpu.module @test_module_gpu_sqrt {
```
jerryyin: TODO rename to `test_gpu_module_gpu_rsqrt`

external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
```diff
@@ -875,6 +875,17 @@ struct MFMAOpLowering : ConvertToLLVMPattern {
...
             ValueRange{adaptor.sourceA(), adaptor.sourceB(), adaptor.destC(),
                        immValues[0], immValues[1], immValues[2]});
+    } else if (mfmaInstr.endswith("i8")) {
+      if (mfmaInstr == "mfma_i32_32x32x8i8")
+        rewriter.replaceOpWithNewOp<ROCDL::mfma_i32_32x32x8i8>(
+            op, adaptor.destC().getType(),
+            ValueRange{adaptor.sourceA(), adaptor.sourceB(), adaptor.destC(),
+                       immValues[0], immValues[1], immValues[2]});
+      else if (mfmaInstr == "mfma_i32_16x16x16i8")
+        rewriter.replaceOpWithNewOp<ROCDL::mfma_i32_16x16x16i8>(
```
krzysz00, line 885: We sure those're the only ones we'll need? Not all our kernels have k reductions?

external/llvm-project/mlir/test/Dialect/GPU/ops.mlir
```diff
@@ -251,6 +251,26 @@ module attributes {gpu.container_module} {
 
       gpu.return
     }
+
+    // CHECK-LABEL: gpu.func @mfma_i8_4xi32
+    // CHECK:       gpu.mfma(%{{.*}}, %{{.*}}, %{{.*}}) : i32, vector<4xi32> 
+    // CHECK-NEXT:  gpu.mfma(%{{.*}}, %{{.*}}, %{{.*}}) : i32, vector<4xi32>
+    gpu.func @mfma_i8_4xi32(%a : i32, %b : i32, %c : vector<4xi32>) {
```
krzysz00, line 258: Does it make sense to test for failure to lower? I suppose it does, so that we'll know to update the tests if we expose those intrinsics

krzysz00 (approved):
Aside from not being sure this is all the intrinsics we'll need, LGTM

As an aside, do we know the packing order of the `i8` in those `i32` arguments?

external/llvm-project/mlir/test/Dialect/GPU/ops.mlir
jerryyin: I'm not sure there's a good way to test "unexposed lowering patterns" because then the unit test run result will be non-zero and it will be misunderstood as a real failure.

Here's what I see when I added the non-reduction version of unit test:

> within split at /root/llvm-project-mlir/external/llvm-project/mlir/test/Conversion/GPUToROCDL/gpu-to-rocdl.mlir:390 offset :17:21: error: failed to legalize operation 'gpu.mfma' that was explicitly marked illegal
    %result16xi32 = gpu.mfma(%arg_i32, %arg_i32, %arg_vi32x32) {instr = "mfma_i32_32x32x4i8"} : i32, vector<32xi32>
Failed Tests (1):
  MLIR :: Conversion/GPUToROCDL/gpu-to-rocdl.mlir

I think we are good in terms that `LowerGpuToROCDLOps.cpp` will immediately return a failure in lowering if this is a pattern not explicitly handled. If we ended up need to handle other type of i8 xdlops, it will be apparent at that point.

external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
jerryyin: > We sure those're the only ones we'll need? Not all our kernels have k reductions?

No I'm not sure but it should be easy enough to add new ones if necessary. Right now I just expose the absolutely minimal. In case we figure later that more needs to be done, I/someone will notice that.

jerryyin:
> As an aside, do we know the packing order of the i8 in those i32 arguments?

Good question, I don't know. I'm assuming since this is just matrix multiplication it should ordered through the smallest axis (Such that for A (m x k) it is packed through k and B(k x n) it is packed through n. Could you help confirm my understanding @zjing14 ?

krzysz00:
@jerryyin I'll note that our matrix multiply is C = A^T B - A is a K x M buffer in its access pattern

jerryyin:
@krzysz00 I have clarified with @zjing14, my assumptions are wrong. Both A and B's data are packed through K dimension.

PR #491, opened 2022-01-28
[DO NOT SQUASH] Use the BF16 type throughout our lowerings and convert to I16 before going to LLVM

krzysz00:
Instead of knowing that i16 just happens to mean bf16, we now use bf16 all the way through the GPU dialect (which also allows us to use truncf/extf). Then, during conversion to LLVM, we override the type conversion logic to map `bf16` to `i16`. After that, we add a software implementation for bf16 truncation and a translation from addf/mulf to add/mul.

This lets us be more clear about our types during lowering and removes a bunch of custom ops, at the cost of complicating the GPUToROCDL pass somewhat.

Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/395 

(there's also fixes to `check-mlir` here that'll be upstreamed once the build is unbroken)


whchung (commented):
Did the first iteration of quick look. Looks fine. Will see what happens in CI.

sjw36:
Looking real good!

krzysz00:
@whchung @sjw36 Looks like CI is fine - then again, we don't actually test any of the bf16 kernels end to end?

whchung (approved):

PR #493, opened 2022-01-31
Remove LowerMIOpenOps.h, take 2

krzysz00:
(I put this on top of the bf16 patch to dodge one set of merge conflicts)

Since the lowering code is now in a somewhat calmer state, I think it's a good time to push through the LowerMIOpenOps.h removal before we get moving on int8. @jerryyin and @whchung to confirm?


whchung (approved):
Now it's a good timing to get rid of the header file!

whchung:
@jerryyin Need your consent to get this done.

mlir/lib/Dialect/MIOpen/utilities.cpp
```diff
@@ -0,0 +1,66 @@
+//===- utilities.cpp - MIOpen utility functions ---------------------===//
```
jerryyin: Nit: I think we can move this inside `GridwiseGemmToBLockwise.cpp` considering that this has been only used there.

This catch my attention because of the file names as `utilities.cpp` or `common.h` sounds like a bad idea. I'd also be fine with you keep this compilation as-is but rename the header/cpp.

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
```diff
@@ -21,6 +20,8 @@
 #include <sstream>
 #include <string>
 
+using namespace mlir;
+
```
jerryyin, line 24: Nit: Why adding this namespace? Looks like symbols below are scoped with the namespace under mlir/llvm already.

mlir/lib/Dialect/MIOpen/CMakeLists.txt
```diff
@@ -7,9 +7,10 @@ if (NOT DEFINED ROCM_PATH)
   endif()
 endif()
 
-file(GLOB globbed *.c *.cpp)
 add_mlir_dialect_library(MLIRMIOpenOps
-  ${globbed}
+  CoordTransformBuilder.cpp
+  MIOpenDialect.cpp
+  utilities.cpp
```
jerryyin: Good job!

jerryyin (approved):
Some nitpicks but LGTM

mlir/lib/Dialect/MIOpen/utilities.cpp
krzysz00: I could probably pull this up a level into `Dialect/MIOpen/utilities.h`

This particular change came out of the test pattern refactor, where I wanted `createConstantFloatOp()` in `miopen-gen`. Since I was restructuring everything, I thought I'd stick it here.

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
krzysz00: Not all the symbols were so prefixed, and things broke without the `using namespace mlir;` that was in `LowerMIOpenOps.h`

mlir/lib/Dialect/MIOpen/utilities.cpp
jerryyin: Sounds good, when do do it there, I'd prefer to more specific name such as `createConstantOp.cpp/h`. Unless you plan to push more generic enough utilities in there.

krzysz00: (also, the functions are used in a few places in lowering)

My general sense is that having a file into which we can chuck utility routines that don't obviously belong anywhere else yet is a good call?

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
jerryyin: Okay gotcha, I think the reason why it wasn't an issue before is probably due to the mlir namespace spilled from the the header into other compilation units. Looks like your refactor corrected that.

mlir/lib/Dialect/MIOpen/utilities.cpp
jerryyin: Yes, avoiding duplication is always a good thing. I'm trying to be defensive on a slightly different topic: If a component doesn't have a specific enough purpose it can evolve into something completely unexpected (I guess LowerMIOpenOps.h is one such example). To prevent such things from happening giving it a specific enough purpose/naming might ring a bell to developer/reviewers.

krzysz00: Hm ... yeah, something like `constantUtils.cpp` or `builderUtils.cpp` might be a good bit of renaming - it leaves the file generic enough that it might collect some other widely-used patterns, and it'd at least match things like `loweringUtils`

PR #494, opened 2022-01-31
Adding i8 suport from miopen to gpu conversion

jerryyin:
This PR adds support from miopen to gpu conversion, and make sure that gpu.mfma accepts argument like i32.

Note: the vector<4xi8> to i32 conversion will behave like a typecast and not incur any additional overhead.

### Example

For such MLIR prototype:
```mlir
module {
  func @mfma_i8_4xi32(%a : vector<4xi8>, %b : vector<4xi8>, %result : memref<16xi32>) attributes {kernel = 0 : i32} {
    %vec = arith.constant dense<0> : vector<16xi32>
    %d0 = miopen.mfma_v2(%a, %b, %vec) { instr = "mfma_i32_32x32x8i8", imm = [0, 0, 0]}: vector<4xi8>, vector<16xi32>

    // Make sure that compiler don't optimize the above away
    %c0 = constant 0 : index
    %v0 = vector.extractelement %d0[%c0 : index] : vector<16xi32>
    memref.store %v0, %result[%c0] : memref<16xi32>
    return }
}
```

The IR for conversion:
```mlir 
      %1 = llvm.mlir.constant(0 : i32) : i32                                                                                                                                                                                                                                       
      %2 = llvm.mlir.constant(1 : index) : i64                                                                                                                                                                                                                                     
      %3 = llvm.mlir.constant(8 : i32) : i32
      %4 = llvm.mlir.constant(2 : index) : i64
      %5 = llvm.mlir.constant(16 : i32) : i32
      %6 = llvm.mlir.constant(3 : index) : i64
      %7 = llvm.mlir.constant(24 : i32) : i32
      %8 = llvm.mlir.constant(0 : index) : i64
      llvm.br ^bb1
    ^bb1:  // pred: ^bb0
      %9 = llvm.extractelement %arg0[%8 : i64] : vector<4xi8>
      %10 = llvm.zext %9 : i8 to i32
      %11 = llvm.shl %10, %1  : i32
      %12 = llvm.extractelement %arg0[%2 : i64] : vector<4xi8>
      %13 = llvm.zext %12 : i8 to i32
      %14 = llvm.shl %13, %3  : i32
      %15 = llvm.or %11, %14  : i32
      ...
      %42 = rocdl.mfma.i32.32x32x8i8 %23, %38, %0, %39, %40, %41 : (i32, i32, vector<16xi32>, i32, i32, i32) -> vector<16xi32>
```


The ISA would be:
```assembly
mfma_i8_4xi32:                                                                                                                                                                                                                                                       
        s_load_dwordx2 s[0:1], s[4:5], 0x0
        s_load_dwordx2 s[2:3], s[4:5], 0x10
        v_accvgpr_write_b32 a0, 0
        v_accvgpr_write_b32 a1, 0
        v_accvgpr_write_b32 a2, 0
        s_waitcnt lgkmcnt(0)
        v_mov_b32_e32 v0, s0
        v_accvgpr_write_b32 a3, 0
        ...
        v_accvgpr_write_b32 a15, 0
        v_mov_b32_e32 v1, s1
        s_nop 1
        v_mfma_i32_32x32x8i8 a[0:15], v0, v1, a[0:15]
```

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
+++ b/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
@@ -481,10 +481,10 @@ def MIOpen_ThreadwiseGemmOp:
 def MIOpen_MFMAV2Op:
     MIOpen_Op<"mfma_v2", [AllTypesMatch<["sourceA", "sourceB"]>,
       AllTypesMatch<["destC", "destD"]>]>,
-    Arguments<(ins AnyTypeOf<[F32, VectorOfLengthAndType<[4], [F16]>, VectorOfLengthAndType<[2], [BF16]>]>: $sourceA,
-                   AnyTypeOf<[F32, VectorOfLengthAndType<[4], [F16]>, VectorOfLengthAndType<[2], [BF16]>]>: $sourceB,
-                   VectorOfRankAndType<[1], [F32, F16]>: $destC)>,
-    Results<(outs VectorOfRankAndType<[1], [F32, F16]>: $destD)> {
+    Arguments<(ins AnyTypeOf<[I32, F32, VectorOfLengthAndType<[4], [F16]>, VectorOfLengthAndType<[2], [BF16]>]>: $sourceA,
+                   AnyTypeOf<[I32, F32, VectorOfLengthAndType<[4], [F16]>, VectorOfLengthAndType<[2], [BF16]>]>: $sourceB,
+                   VectorOfRankAndType<[1], [I32, F32, F16]>: $destC)>,
+    Results<(outs VectorOfRankAndType<[1], [I32, F32, F16]>: $destD)> {
   let summary = "XDLOPS MFMA V2";
   let description = [{
     The `miopen.mfma_v2` op is an abstraction of XDLOPS.
--- a/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
+++ b/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
@@ -481,8 +481,8 @@ def MIOpen_ThreadwiseGemmOp:
 def MIOpen_MFMAV2Op:
     MIOpen_Op<"mfma_v2", [AllTypesMatch<["sourceA", "sourceB"]>,
       AllTypesMatch<["destC", "destD"]>]>,
-    Arguments<(ins AnyTypeOf<[I32, F32, VectorOfLengthAndType<[4], [F16]>, VectorOfLengthAndType<[2], [BF16]>]>: $sourceA,
-                   AnyTypeOf<[I32, F32, VectorOfLengthAndType<[4], [F16]>, VectorOfLengthAndType<[2], [BF16]>]>: $sourceB,
+    Arguments<(ins AnyTypeOf<[F32, VectorOfLengthAndType<[4], [F16]>, VectorOfLengthAndType<[2], [BF16]>, VectorOfLengthAndType<[4], [I8]>]>: $sourceA,
+                   AnyTypeOf<[F32, VectorOfLengthAndType<[4], [F16]>, VectorOfLengthAndType<[2], [BF16]>, VectorOfLengthAndType<[4], [I8]>]>: $sourceB,
                    VectorOfRankAndType<[1], [I32, F32, F16]>: $destC)>,
     Results<(outs VectorOfRankAndType<[1], [I32, F32, F16]>: $destD)> {
   let summary = "XDLOPS MFMA V2";
--- a/mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
+++ b/mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
@@ -48,6 +48,17 @@ struct XdlopsCodeSelection {
   int64_t cycles;
   int64_t k_base;
 
+  static void dumpUnsupported(const mlir::Type &dataType, int64_t MPerWave,
+                              int64_t NPerWave) {
+    llvm::errs() << "Unsupported case:\n";
+    llvm::errs() << "MPerWave: " << MPerWave << "\n";
+    llvm::errs() << "NPerWave: " << NPerWave << "\n";
+    llvm::errs() << "dataType: ";
+    dataType.dump();
+    llvm::errs() << "\n";
+    llvm_unreachable("Can't meaningfully continue xdlop generation");
+  }
+
   static XdlopsCodeSelection get(mlir::Type &dataType, int64_t MPerWave,
                                  int64_t NPerWave, OpBuilder &b) {
     // Determine which XDLOPS be used.
@@ -178,14 +189,7 @@ struct XdlopsCodeSelection {
         imms.push_back({0, 0, 0});
         argType = b.getF32Type();
       } else {
-        llvm::errs() << "Unsupported case:\n";
-        // llvm::errs() << "M, N, K:" << M << " " << N << " " << K << "\n";
-        llvm::errs() << "MPerWave: " << MPerWave << "\n";
-        llvm::errs() << "NPerWave: " << NPerWave << "\n";
-        llvm::errs() << "dataType: ";
-        dataType.dump();
-        llvm::errs() << "\n";
-        llvm_unreachable("Can't meaningfully continue xdlop generation");
+        dumpUnsupported(dataType, MPerWave, NPerWave);
       }
     } else if (dataType == b.getF16Type()) {
       if (MPerWave == 128 && NPerWave == 64) {
@@ -307,14 +311,7 @@ struct XdlopsCodeSelection {
         imms.push_back({0, 0, 0});
         argType = VectorType::get({4}, b.getF16Type());
       } else {
-        llvm::errs() << "Unsupported case:\n";
-        // llvm::errs() << "M, N, K:" << M << " " << N << " " << K << "\n";
-        llvm::errs() << "MPerWave: " << MPerWave << "\n";
-        llvm::errs() << "NPerWave: " << NPerWave << "\n";
-        llvm::errs() << "dataType: ";
-        dataType.dump();
-        llvm::errs() << "\n";
-        llvm_unreachable("Can't meaningfully continue xdlop generation");
+        dumpUnsupported(dataType, MPerWave, NPerWave);
       }
     } else if (dataType == b.getBF16Type()) {
       if (MPerWave == 128 && NPerWave == 64) {
@@ -436,14 +433,31 @@ struct XdlopsCodeSelection {
         imms.push_back({0, 0, 0});
         argType = VectorType::get({2}, b.getBF16Type());
       } else {
-        llvm::errs() << "Unsupported case:\n";
-        // llvm::errs() << "M, N, K:" << M << " " << N << " " << K << "\n";
-        llvm::errs() << "MPerWave: " << MPerWave << "\n";
-        llvm::errs() << "NPerWave: " << NPerWave << "\n";
-        llvm::errs() << "dataType: ";
-        dataType.dump();
-        llvm::errs() << "\n";
-        llvm_unreachable("Can't meaningfully continue xdlop generation");
+        dumpUnsupported(dataType, MPerWave, NPerWave);
+      }
+    } else if (dataType == b.getIntegerType(8)) {
+      if (MPerWave == 32 && NPerWave == 32) {
+        mfmaInstr = "mfma_i32_32x32x8i8";
+        MPerXdlops = 32;
+        NPerXdlops = 32;
+        MRepeats = 1;
+        NRepeats = 1;
+        vectorType = VectorType::get({16}, b.getI32Type());
+        vectorNumber = 1;
+        imms.push_back({0, 0, 0});
+        argType = b.getIntegerType(32);
+      } else if (MPerWave == 16 && NPerWave == 16) {
+        mfmaInstr = "mfma_i32_16x16x16i8";
+        MPerXdlops = 16;
+        NPerXdlops = 16;
+        MRepeats = 1;
+        NRepeats = 1;
+        vectorType = VectorType::get({4}, b.getI32Type());
+        vectorNumber = 1;
+        imms.push_back({0, 0, 0});
+        argType = b.getIntegerType(32);
+      } else {
+        dumpUnsupported(dataType, MPerWave, NPerWave);
       }
     } else {
       llvm_unreachable("XDLOPs for unsupported data types should be rejected");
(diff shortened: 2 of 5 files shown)


krzysz00:
Random question: do the xdlops here take an i32, or do they take an i8 that happens to live in an i32?

jerryyin:
@krzysz00 The input i32 are packed i8s such that each i32 is a by 4 packed i8 in k direction. The output i32 is real i32s storing mfma results.

krzysz00:
Good to know - that makes it a good thing we've got kpack.

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -481,10 +481,10 @@ def MIOpen_ThreadwiseGemmOp:
 def MIOpen_MFMAV2Op:
     MIOpen_Op<"mfma_v2", [AllTypesMatch<["sourceA", "sourceB"]>,
       AllTypesMatch<["destC", "destD"]>]>,
-    Arguments<(ins AnyTypeOf<[F32, VectorOfLengthAndType<[4], [F16]>, VectorOfLengthAndType<[2], [I16]>]>: $sourceA,
-                   AnyTypeOf<[F32, VectorOfLengthAndType<[4], [F16]>, VectorOfLengthAndType<[2], [I16]>]>: $sourceB,
-                   VectorOfRankAndType<[1], [F32, F16]>: $destC)>,
-    Results<(outs VectorOfRankAndType<[1], [F32, F16]>: $destD)> {
+    Arguments<(ins AnyTypeOf<[I32, F32, VectorOfLengthAndType<[4], [F16]>, VectorOfLengthAndType<[2], [I16]>]>: $sourceA,
+                   AnyTypeOf<[I32, F32, VectorOfLengthAndType<[4], [F16]>, VectorOfLengthAndType<[2], [I16]>]>: $sourceB,
```
krzysz00: Do we instead want to call this `vector<4xi8>` since that's what it is? `bitcast` exists on the way down?

Or will that transition live elsewhere in the pipeline?

krzysz00 (commented):
I'm going to wait for someone else to go y/n on this, but it seems to me like we want to encode the fact that the input `i32` are actually `i8 ++ i8 ++ i8 ++ i8` at this high in the lowering level? At the very least we need kPack to trigger on these values.

krzysz00:
(As a side note, I think we'll want to add `miopen.pack` / `miopen.unpack` (and then upstream them) that do things like `%5 = miopen.pack %1, %2, %3, %4 : i8, i8, i8, i8 -> i32`)

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
jerryyin: From my limited knowledge in those passes, I can see miopen.mfa->gpu.mfa->xdlops are mostly a smooth 1:1 transition, so it doesn't look like the I32 <-> vector<4xi8> transition happens in those last couple of passes. 

@whchung Which pass do you think should we do the packing?

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -436,14 +433,31 @@ struct XdlopsCodeSelection {
...
+      } else if (MPerWave == 16 && NPerWave == 16) {
+        mfmaInstr = "mfma_i32_16x16x16i8";
+        MPerXdlops = 16;
+        NPerXdlops = 16;
+        MRepeats = 1;
+        NRepeats = 1;
+        vectorType = VectorType::get({4}, b.getI32Type());
+        vectorNumber = 1;
+        imms.push_back({0, 0, 0});
+        argType = b.getIntegerType(32);
```
jerryyin, line 458: @zjing14 I want to make sure you agree with those.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -663,6 +677,34 @@ struct XdlopsCodeSelection {
...
+      num_threads_blk = 16;
+      wave_size = 64;
+      num_input_blks = wave_size / num_threads_blk;
+      num_output_blks = 1;
+      num_regs_xdlops = num_regs_blk * num_output_blks;
+      m = 16;
+      n = 16;
+      k = 16;
+      cycles = 32;
+      k_base = 4;
```
jerryyin, line 707: @zjing14 I want to make sure you agree with those.

jerryyin:
> (As a side note, I think we'll want to add miopen.pack / miopen.unpack (and then upstream them)

@krzysz00 I believe they already exists somewhere, I just don't know. fp16, for example, also requires some packing.

krzysz00:
@jerryyin I don't think we quite have that operation. We do have some stuff where our in-register and in-LDS memrefs have a vector type at their base, which might be what we want to adapt. I'm not sure that guarantees the in-register packing behavior we need though.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -436,14 +433,31 @@ struct XdlopsCodeSelection {
...
+      if (MPerWave == 32 && NPerWave == 32) {
+        mfmaInstr = "mfma_i32_32x32x8i8";
+        MPerXdlops = 32;
+        NPerXdlops = 32;
+        MRepeats = 1;
+        NRepeats = 1;
+        vectorType = VectorType::get({16}, b.getI32Type());
+        vectorNumber = 1;
+        imms.push_back({0, 0, 0});
+        argType = b.getIntegerType(8);
```
krzysz00: Is `argType` here a vector of i8 or just a single i8?

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
```diff
@@ -125,12 +125,53 @@ struct MIIdRewritePattern : public OpRewritePattern<Tmi> {
...
+      Value sourceA =
+          b.create<arith::ConstantIntOp>(loc, 0, b.getIntegerType(32));
+      Value sourceB =
+          b.create<arith::ConstantIntOp>(loc, 0, b.getIntegerType(32));
+      vector4i8Toi32(op.sourceA(), sourceA, b);
+      vector4i8Toi32(op.sourceB(), sourceB, b);
+      gpuMfmaOp = b.create<gpu::MFMAOp>(op.getLoc(), op.getType(), sourceA,
+                                        sourceB, op.destC());
+    } else {
+      gpuMfmaOp = b.create<gpu::MFMAOp>(op.getLoc(), op.getType(), op.sourceA(),
```
krzysz00: It might be better to always have `Value sourceA = op.sourceA();` then to have the if statement to update `sourceA` and `sourceB` for `i8` vectors - that wey we don't repeat the creation of the MFMA Op

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
```diff
@@ -125,12 +125,53 @@ struct MIIdRewritePattern : public OpRewritePattern<Tmi> {
 struct MIMFMARewritePattern : public OpRewritePattern<miopen::MFMAV2Op> {
   using OpRewritePattern<miopen::MFMAV2Op>::OpRewritePattern;
 
+  // The below implements the concatenation of 4 i8 to an i32:
+  // a[0] | (a[1] << 8) | (a[2] << 16) | (a[3] << 24)
+  void vector4i8Toi32(const Value &vec4i8, Value &i32vals,
```
krzysz00: Can this be a [wrapped in a] more general utility that goes from `vector<4Nxi8>` to `vector<Nxi32>` with the special case for N=1? This function feels quite fragile.

krzysz00 (commented):
I've got minor nitpicks and some questions, but I like where this is headed.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -436,14 +433,31 @@ struct XdlopsCodeSelection {
...
+        mfmaInstr = "mfma_i32_16x16x16i8";
+        MPerXdlops = 16;
+        NPerXdlops = 16;
+        MRepeats = 1;
+        NRepeats = 1;
+        vectorType = VectorType::get({4}, b.getI32Type());
+        vectorNumber = 1;
+        imms.push_back({0, 0, 0});
+        argType = b.getIntegerType(8);
+      } else {
```
jerryyin, line 459: @zjing14 Please help review, thanks!

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
jerryyin: Nice catch, this should be vector<4xi8> lookin at how it is defined for other types. Will amend.

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
jerryyin: Sounds reasonable, will do

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
jerryyin: Are you talking about adding N as a argument? 

I feel like instead I can also parse the type of the incoming `vec4i8`, make sure it is 1d shape and dimension % 4 == 0, then use the dimension as-is. Hopefully it makes it more robust and generalized.

krzysz00: I meant parsing the shape, yeah
That way, if I give the function a vector<8xi8>, I'll get a vector<2xi32> back, if I pass in vector<4xi8> I get an i32, and so on

(and if I send in a vector<3xi8> I get errors)

jerryyin: Hmm I'm not sure there's a need for that because we don't have any xdlops related test cast that will treat vector<8xi8> as vector<2xi32> (it feels wrong since if they are already using vector types, why not use the right ones?). Besides, the vector to vector conversion feels like it really needs to sit in vector dialect, or at least serve as an extension of an existing op rather than in our lowering pattern. 

Right now I'd prefer an explicit assert on size equal to 4 and implement this only when it turn out to be necessary.

krzysz00: Makes sense. Asserting that we're specifically converting a vector<4xi8> to an i32 is a perfectly solid way to handle things for now and we can add more generality when we need it.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
jerryyin: Addressed in d1a2fab

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
jerryyin: Addressed in d1a2fab

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -436,14 +433,31 @@ struct XdlopsCodeSelection {
...
+        mfmaInstr = "mfma_i32_16x16x16i8";
+        MPerXdlops = 16;
+        NPerXdlops = 16;
+        MRepeats = 1;
+        NRepeats = 1;
+        vectorType = VectorType::get({4}, b.getI32Type());
+        vectorNumber = 1;
+        imms.push_back({0, 0, 0});
+        argType = VectorType::get({4}, b.getIntegerType(8));
+      } else {
```
jerryyin, line 459: @zjing14 Please help review, thanks

zjing14: LGTM

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
zjing14: LGTM

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
```diff
@@ -125,12 +125,71 @@ struct MIIdRewritePattern : public OpRewritePattern<Tmi> {
 struct MIMFMARewritePattern : public OpRewritePattern<miopen::MFMAV2Op> {
   using OpRewritePattern<miopen::MFMAV2Op>::OpRewritePattern;
 
+  // The below implements the concatenation of vector<4xi8> to an i32:
+  // a[0] | (a[1] << 8) | (a[2] << 16) | (a[3] << 24)
+  LogicalResult vector4i8Toi32(const Value &vec4i8, Value &i32vals,
+                               PatternRewriter &b) const {
+    if (!vec4i8.getType().isa<VectorType>())
+      return failure();
+    auto vecType = vec4i8.getType().cast<VectorType>();
```
krzysz00, line 134: Nit: this can be `auto vecType = vec4i8.getType().dyn_cast<VectorType()` followed by `if (!vecType) return failure();`

krzysz00 (approved):
LGTM, one non-critical style nit

mlir/lib/Conversion/MIOpenToGPU/MIOpenToGPU.cpp
jerryyin: Sounds reasonable, saves an additional check! I'd skip the fix tho this time but remember to use it next time

krzysz00: Yeah - the LLVM style guide had a mention of `isa<>() ... cast<>` being an anti-pattern that should usually be `dyn_cast<>`

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -436,14 +433,31 @@ struct XdlopsCodeSelection {
...
+        vectorNumber = 1;
+        imms.push_back({0, 0, 0});
+        argType = VectorType::get({4}, b.getIntegerType(8));
+      } else if (MPerWave == 16 && NPerWave == 16) {
+        mfmaInstr = "mfma_i32_16x16x16i8";
+        MPerXdlops = 16;
+        NPerXdlops = 16;
+        MRepeats = 1;
+        NRepeats = 1;
+        vectorType = VectorType::get({4}, b.getI32Type());
```
krzysz00, line 455: Hold on a minute, shouldn't this be a vector 16xi8?

krzysz00 (changes requested):
Found an issue at the last minute

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
jerryyin: Hah. I believe vector type is where return type is defined; argType are where argument type are defined.
 - For 16x16x16i8 xdlops, the return type is a vector of 4xi32. 
 - For 32x32x8i8 xdlops, the return type is a vector of 16xi32

krzysz00 (approved):
Right, I somehow blanked on the nearby argType, my bad

jerryyin:
@krzysz00 No worries, and thanks for keeping an extra eye on it.

PR #495, opened 2022-02-01
MLIR#406: make perf_config an attribute on miopen.conv2d*

sjw36:
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/406


krzysz00 (approved):
LGTM, might want to wait for @jerryyin to take a look at the MIOpen parts

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
```diff
@@ -606,6 +606,12 @@ LogicalResult Conv2dGenerator::genConvModule(ModuleOp &module, int kernel_id,
...
 
+  // perf_config
+  if (!config.perfConfig.empty()) {
+    attributes.push_back(builder.getNamedAttr(
+        "perf_config", builder.getStringAttr(config.perfConfig)));
+  }
+
   // ignore tuning.
   if (ignoreTuning) {
     attributes.push_back(
```
jerryyin: Could you help get rid of this `ignore_tuning` attribute. Now that you made `perfConfig` to be a independent attribute, this `ignore_tuning` doesn't seem necessary anymore.

Instead, you can make it much cleaner by such that: when `ignoreTuning == true`, assign an empty string to the perfConfig associated with the conv2d op. This way at the op creation time I can make the perf_config initialized correctly (instead of the hack that correct the attribute later)

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
```diff
@@ -168,6 +166,11 @@ void AffixTuningParameters::affixTuningParametersImpl(T &op) {
...
   };
 
+  std::string perfConfig;
+  if (auto perfConfigAttr =
+          op->template getAttrOfType<StringAttr>("perf_config")) {
+    perfConfig = perfConfigAttr.getValue().str();
+  }
   auto xdlopsV2Attr = op->template getAttrOfType<BoolAttr>("xdlopsV2");
   auto ignoreTuningAttr = op->template getAttrOfType<BoolAttr>("ignore_tuning");
   bool ignoreTuning =
```
jerryyin: You can get rid of this, as well as line 188.

jerryyin (commented):
Looks good overall thanks for the work. Has one minor nit, once you did the last cleanup I'm ready to give approval.

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
sjw36: should be cleaned up now.

PR #497, opened 2022-02-03
Add a script that automatically generates E2E tests

yiqian1:
This work implements [this issue](https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/356).

I've used this script to generate Resnet50 test cases and run the test cases successfully.


mlir/utils/generateE2ETest.py
```diff
@@ -0,0 +1,138 @@
...
+        options=[]
+        for value in item[key2]:
+            if item["name"] in prefixes:
+                opt = prefixes[item["name"]] + ' ' + value
+            else:
+                opt = value
+            options.append(opt)
+        options_list.append(options)
+
+    combinations=[]
```
krzysz00, line 48: Random question - could we want this to be a `set()` 

Heck, you can probably do `combinations = frozenset(itertools.product(*options_list))`

krzysz00: And IIRC set/dict have a stable iteration order these days

mlir/utils/generateE2ETest.py
```diff
@@ -0,0 +1,138 @@
...
+                f.write("//This file is automatically generated. Do not edit!\n\n")
+                exclusions = []
+                if "exclude" in test:
+                    exclusions = generate_option_list(axis_prefixes, test, "exclude", "values")
+                for opt in combinations:
+                    opt_idx = opt_idx + 1
+                    config = prefix + ' ' +test["config"]
+                    skip = False;
+                    for exclude in exclusions:
+                        if (all (x in opt for x in exclude)):
```
krzysz00, line 124: If we went set, this'd just be a subset check, but is fine as it is

krzysz00 (approved):
Good work! I'm assuming this is tested etc etc.

I have one minor nit - options/excludes could be sets instead of lists, but that's not blocking

PR #498, opened 2022-02-03
Fix invalid out of bound vector access.

whchung:
Issue: https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/423

### DIFF
--- a/mlir/test/Dialect/MIOpen/lowering_affix_params_negative.mlir
+++ b/mlir/test/Dialect/MIOpen/lowering_affix_params_negative.mlir
@@ -1,6 +1,10 @@
-// XFAIL: *
-// REQUIRES: miopen-driver
-// RUN: miopen-gen -p -x2 -t f16 --perf_config "128,128,2,64,64,2,1,1" | mlir-miopen-driver -miopen-affix-params
-// RUN: miopen-gen -p -x2 -t f16 --perf_config "128,128,2,64,64,16,1,1" | mlir-miopen-driver -miopen-affix-params
-
 // Negative test to deliberately pass incorrect tuning parameters.
+
+// RUN: miopen-gen -p -x2 -t f16 --perf_config "128,128,2,64,64,2,1,1" | miopen-opt -miopen-affix-params -verify-diagnostics
+// expected-error{{Incorrect KPACK tuning parameter: 2}}
+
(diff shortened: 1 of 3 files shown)


mlir/test/Dialect/MIOpen/lowering_affix_params_negative.mlir
```diff
@@ -0,0 +1,6 @@
+// XFAIL: *
+// REQUIRES: miopen-driver
+// RUN: miopen-gen -p -x2 -t f16 --perf_config "128,128,2,64,64,2,1,1" | mlir-miopen-driver -miopen-affix-params
+// RUN: miopen-gen -p -x2 -t f16 --perf_config "128,128,2,64,64,16,1,1" | mlir-miopen-driver -miopen-affix-params
```
krzysz00: Shouldn't this be in the `mlir-miopen-driver` tests?

krzysz00 (commented):
LGTM but you'll want to move the test, I think

mlir/test/Dialect/MIOpen/lowering_affix_params_negative.mlir
whchung: I'll move the test.
Wondering what's a good guideline to distinguish between `Dialect/MIOpen` and `mlir-miopen-driver` tests?

krzysz00: On the other hand, this could go in Dialect/MIOpen

I think the distinction is whether you call `mlopen-gen/mlir-miopen-driver` or `miopen-opt`

krzysz00: If you wanted this is a Dialect/MIOpen test, I'd pass in just the conv op with the perf_config attribute set

whchung: Let's keep this test under `Dialect/MIOpen` as I feel it's more related to the lowering process, not the driver tool itself.

I also revised the test so it doesn't depend on `mlir-miopen-driver`.

krzysz00 (approved):
The above nitpicking about test location has no business blocking this PR, but is demonstrating a wider issue.

IMO, especially now with Yi's script, we should adjust the test layout so we have a clean separation between things that require a GPU and things that don't.

PR #499, opened 2022-02-04
MLIR#425: enhanced miopen-gen to generate test code for specified func

sjw36:
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/425


sjw36:
@yiqian1 I just added a test that generates a conv2d on cpu using linalg and prints the results. It looks correct. Once I add the cloning code for kernels we should be able to test this against a gpu kernel as well.

sjw36:
For different layouts, we can insert transpose operations around the conv and Jungwook is adding code to fuse those into the miopen.conv2d.

yiqian1 (approved):

yiqian1:
> @yiqian1 I just added a test that generates a conv2d on cpu using linalg and prints the results. It looks correct. Once I add the cloning code for kernels we should be able to test this against a gpu kernel as well.

Nice. Thank you.

PR #500, opened 2022-02-07
[DO NOT SQUASH] MLIR#407: Clone kernels into __miopen module for kernel compilation

sjw36:
* Only run kernel pipeline on __miopen module in mlir-miopen-driver (-kernel-pipeline)
    * Specify host-pipeline in mlir-miopen-driver
    * Extract kernels from __miopen module, apply to host kernel func (-host-pipeline=execmodel)
    * Reordered mlir-rocm-runner passes for host vs kernel pipelines

This is phase 1 of creating an Execution Model representation that can be run in a heterogeneous rocm-runner. In phase 2, kernels can be generated for multiple target architectures and the rocm-runner could run a model heterogeneously (maybe phase 3).

    https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/407


mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
```diff
@@ -55,86 +56,168 @@ static cl::opt<std::string> outputFilename("o", cl::desc("Output filename"),
...
+    int optLevel = gpuOpt.getValue();
+    if (optLevel < 0 || optLevel > 3) {
+      llvm::errs() << "Invalid GPU optimization level: " << optLevel << "\n";
+      return failure();
     }
+    bool systemOverride = false;
+    if (tripleName.empty() && targetChip.empty() && features.empty()) {
+      systemOverride = true;
+    }
+    BackendUtils utils(tripleName, targetChip, features, systemOverride);
```
jerryyin: This has to stay in the `mlir-rocm-runner`. Instead, you need to force user to give all those values from command line.

If not, this will just compile silently use whatever machine you are on and give the binary with its arch to populate in the graph.

mlir/lib/Dialect/MIOpen/Transforms/ApplyImpl.cpp
```diff
@@ -0,0 +1,108 @@
...
+using namespace mlir;
+
+namespace {
+
+struct MIOpenApplyImplPass
+    : public MIOpenApplyImplPassBase<MIOpenApplyImplPass> {
+
+  void runOnOperation() override {
+    ModuleOp mod = getOperation();
+    llvm::StringRef modName = "__miopen";
```
jerryyin: `__miopen` needs to be declared/defined in a common header

mlir/lib/Dialect/MIOpen/Transforms/ApplyImpl.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+      SmallVector<gpu::GPUModuleOp, 8> gpuMods;
+      miopenMod->walk([&](gpu::GPUModuleOp gpuMod) {
+        // auto loc = gpuMod->getLoc();
+        auto binaryAttr = gpuMod->getAttrOfType<StringAttr>(
+            gpu::getDefaultGpuBinaryAnnotation());
+        if (!binaryAttr) {
+          gpuMod.emitOpError() << "missing gpu.binary attribute";
+          return;
+        }
```
jerryyin, line 57: Shall we emit an error here? It looks like something we should tolerate (so far as I can remember from your design: If it doesn't already have a binary, it should be still populated and allow JIT to pick it up and decide later)

mlir/lib/Dialect/MIOpen/Transforms/ApplyImpl.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+        gpuMods.push_back(gpuMod);
+
+        SmallString<128> binaryCstName(
+            SymbolTable::getSymbolName(gpuMod).getValue());
+        binaryCstName.append("_gpubin_cst");
+        // make global constant in TopModule
+        auto binaryCst = b.create<ConstantOp>(loc, binaryAttr);
+        binaryCst->setAttr(SymbolTable::getSymbolAttrName(),
+                           b.getStringAttr(binaryCstName));
```
jerryyin: Are we guaranteed that this binary name is unique? Could we be populating more than 1 kernel which happen to populate the same constant name?

mlir/lib/Dialect/MIOpen/Transforms/CloneKernels.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include "llvm/ADT/SmallVector.h"
+
+using namespace mlir;
+
+namespace {
+
+struct MIOpenCloneKernelsPass
+    : public MIOpenCloneKernelsPassBase<MIOpenCloneKernelsPass> {
+  void runOnOperation() override {
+    ModuleOp mod = getOperation();
```
jerryyin, line 37: How many times will we be running this pass? If there are more than 1 high level op, will we traverse this multiple times each time creating a different gpu module?

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
```diff
@@ -55,86 +56,168 @@ static cl::opt<std::string> outputFilename("o", cl::desc("Output filename"),
...
+                                "partition,highlevel,execmodel"),
+                 cl::init(""));
+
+static cl::opt<bool> legacyMiopenPipeline("c", cl::Hidden, cl::init(false),
+                                          cl::Optional,
+                                          cl::cb<void, bool>([](bool v) {
+                                            if (v) {
+                                              kernelPipeline.setValue("gpu");
+                                            }
+                                          }));
```
jerryyin, line 79: Why the need for this option?

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
```diff
@@ -55,86 +56,168 @@ static cl::opt<std::string> outputFilename("o", cl::desc("Output filename"),
...
-    } else if (pipeline == "rocdl") {
-      // Set up the lowering pipeline which goes down to ROCDL dialect.
-      miopen::addPipeline(pm);
-      pm.addPass(createLowerGpuOpsToROCDLOpsPass(/*indexBitWidth=*/32));
+  if (kernelPipelineSet.contains("binary")) {
+    // Set up the lowering pipeline which goes down to ROCDL dialect.
+    int optLevel = gpuOpt.getValue();
+    if (optLevel < 0 || optLevel > 3) {
+      llvm::errs() << "Invalid GPU optimization level: " << optLevel << "\n";
+      return failure();
```
jerryyin: This is shared between the rocm runner, you need a common header/compilation unit. Otherwise if we ever place to change it, there's more than one place to remember to update.

jerryyin (dismissed):

sjw36:
Resulting XMIR from fusion/tosa-to-miopen-relu.xmir.mlir should look like this:

```
#map0 = affine_map<(d0, d1, d2, d3) -> (d0 * 73984 + d1 * 2176 + d2 * 64 + d3 + 2240)>
#map1 = affine_map<(d0, d1, d2, d3) -> (d3, d0, d1, d2)>
#map2 = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>
module  {
  func private @resnet50_outlined_part_0(%arg0: memref<1x32x32x64xf32>, %arg1: memref<64x3x3x64xf32>, %arg2: memref<1x32x32x64xf32>) attributes {kernel, targets = [{arch = "gfx908", binary = @resnet50_outlined_part_0_module_gpubin_cst, block_size = 64 : i32, grid_size = 16 : i32, type = "gpu"}]} {
    %cst_1 = arith.constant 6.000000e+00 : f32
    %cst_2 = arith.constant 0.000000e+00 : f32
    %0 = memref.alloc() : memref<1x34x34x64xf32>
    linalg.fill(%cst_2, %0) : f32, memref<1x34x34x64xf32>
    %1 = memref.alloc() : memref<1x34x34x64xf32>
    linalg.copy(%0, %1) : memref<1x34x34x64xf32>, memref<1x34x34x64xf32>
    %2 = memref.subview %1[0, 1, 1, 0] [1, 32, 32, 64] [1, 1, 1, 1] : memref<1x34x34x64xf32> to memref<1x32x32x64xf32, #map0>
    linalg.copy(%arg0, %2) : memref<1x32x32x64xf32>, memref<1x32x32x64xf32, #map0>
    %3 = memref.alloc() : memref<3x3x64x64xf32>
    linalg.generic {indexing_maps = [#map1, #map2], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%arg1 : memref<64x3x3x64xf32>) outs(%3 : memref<3x3x64x64xf32>) {
    ^bb0(%arg3: f32, %arg4: f32):  // no predecessors
      linalg.yield %arg3 : f32
    }
    %4 = memref.alloc() : memref<1x32x32x64xf32>
    linalg.fill(%cst_2, %4) : f32, memref<1x32x32x64xf32>
    %5 = memref.alloc() : memref<1x32x32x64xf32>
    linalg.copy(%4, %5) : memref<1x32x32x64xf32>, memref<1x32x32x64xf32>
    linalg.conv_2d_nhwc_hwcf {dilations = dense<1> : tensor<2xi64>, strides = dense<1> : tensor<2xi64>} ins(%1, %3 : memref<1x34x34x64xf32>, memref<3x3x64x64xf32>) outs(%5 : memref<1x32x32x64xf32>)
    linalg.generic {indexing_maps = [#map2, #map2], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%5 : memref<1x32x32x64xf32>) outs(%arg2 : memref<1x32x32x64xf32>) {
    ^bb0(%arg3: f32, %arg4: f32):  // no predecessors
      %6 = arith.addf %cst_2, %arg3 : f32
      %7 = arith.cmpf olt, %6, %cst_2 : f32
      %8 = select %7, %cst_2, %6 : f32
      %9 = arith.cmpf olt, %cst_1, %6 : f32
      %10 = select %9, %cst_1, %8 : f32
      linalg.yield %10 : f32
    }
    return
  }
  func private @resnet50_outlined_part_1(%arg0: memref<1x32x32x64xf32>, %arg1: memref<64x3x3x64xf32>, %arg2: memref<1x32x32x64xf32>, %arg3: memref<1x32x32x64xf32>) attributes {kernel, targets = [{arch = "gfx908", binary = @resnet50_outlined_part_1_module_gpubin_cst, block_size = 64 : i32, grid_size = 16 : i32, type = "gpu"}]} {
    %cst_1 = arith.constant 6.000000e+00 : f32
    %cst_2 = arith.constant 0.000000e+00 : f32
    %0 = memref.alloc() : memref<1x34x34x64xf32>
    linalg.fill(%cst_2, %0) : f32, memref<1x34x34x64xf32>
    %1 = memref.alloc() : memref<1x34x34x64xf32>
    linalg.copy(%0, %1) : memref<1x34x34x64xf32>, memref<1x34x34x64xf32>
    %2 = memref.subview %1[0, 1, 1, 0] [1, 32, 32, 64] [1, 1, 1, 1] : memref<1x34x34x64xf32> to memref<1x32x32x64xf32, #map0>
    linalg.copy(%arg0, %2) : memref<1x32x32x64xf32>, memref<1x32x32x64xf32, #map0>
    %3 = memref.alloc() : memref<3x3x64x64xf32>
    linalg.generic {indexing_maps = [#map1, #map2], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%arg1 : memref<64x3x3x64xf32>) outs(%3 : memref<3x3x64x64xf32>) {
    ^bb0(%arg4: f32, %arg5: f32):  // no predecessors
      linalg.yield %arg4 : f32
    }
    %4 = memref.alloc() : memref<1x32x32x64xf32>
    linalg.fill(%cst_2, %4) : f32, memref<1x32x32x64xf32>
    %5 = memref.alloc() : memref<1x32x32x64xf32>
    linalg.copy(%4, %5) : memref<1x32x32x64xf32>, memref<1x32x32x64xf32>
    linalg.conv_2d_nhwc_hwcf {dilations = dense<1> : tensor<2xi64>, strides = dense<1> : tensor<2xi64>} ins(%1, %3 : memref<1x34x34x64xf32>, memref<3x3x64x64xf32>) outs(%5 : memref<1x32x32x64xf32>)
    linalg.generic {indexing_maps = [#map2, #map2, #map2], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%arg2, %5 : memref<1x32x32x64xf32>, memref<1x32x32x64xf32>) outs(%arg3 : memref<1x32x32x64xf32>) {
    ^bb0(%arg4: f32, %arg5: f32, %arg6: f32):  // no predecessors
      %6 = arith.addf %cst_2, %arg5 : f32
      %7 = arith.cmpf olt, %6, %cst_2 : f32
      %8 = select %7, %cst_2, %6 : f32
      %9 = arith.cmpf olt, %cst_1, %6 : f32
      %10 = select %9, %cst_1, %8 : f32
      %11 = arith.addf %arg4, %10 : f32
      linalg.yield %11 : f32
    }
    return
  }
  func @resnet50(%arg0: memref<1x32x32x64xf32>, %arg1: memref<64x3x3x64xf32>, %arg2: memref<64x3x3x64xf32>, %arg3: memref<1x32x32x64xf32>) {
    %0 = memref.alloc() : memref<1x32x32x64xf32>
    call @resnet50_outlined_part_0(%arg0, %arg1, %0) : (memref<1x32x32x64xf32>, memref<64x3x3x64xf32>, memref<1x32x32x64xf32>) -> ()
    %1 = memref.alloc() : memref<1x32x32x64xf32>
    call @resnet50_outlined_part_1(%0, %arg2, %arg0, %1) : (memref<1x32x32x64xf32>, memref<64x3x3x64xf32>, memref<1x32x32x64xf32>, memref<1x32x32x64xf32>) -> ()
    memref.copy %1, %arg3 : memref<1x32x32x64xf32> to memref<1x32x32x64xf32>
    return
  }
  %cst = constant  {sym_name = "resnet50_outlined_part_0_module_gpubin_cst"} "\7FELF\02\01..." // full binary
  %cst_0 = constant  {sym_name = "resnet50_outlined_part_1_module_gpubin_cst"} "\7FELF\02\01..."
}
```

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
sjw36: This is only needed if you compile to binary. As part of lowering to XMIR we need the binaries to plug into the host code in the target spec (see example below).

I can change this to error if the target chip is not specified.

mlir/lib/Dialect/MIOpen/Transforms/ApplyImpl.cpp
sjw36: Possibly yes we could reference the gpu.func instead of the binary.

mlir/lib/Dialect/MIOpen/Transforms/ApplyImpl.cpp
sjw36: Good call. I will add a check/uniquifier.

mlir/lib/Dialect/MIOpen/Transforms/CloneKernels.cpp
sjw36: This pass only gets called on the ModuleOp so it will only traverse each Module 1 time looking for kernel funcs.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
sjw36: So I didn't have to update all tests... it is like an alias.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
sjw36: It seems a bit to trivial at this point. I did change to use the miopen::addBackendPipeline.

mlir/lib/Dialect/MIOpen/Transforms/ApplyImpl.cpp
sjw36: I will file a ticket to add this capability in a later version. The first target is to generate binaries.

sjw36: See https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/428

mlir/lib/Dialect/MIOpen/Transforms/ApplyImpl.cpp
sjw36: Moved to MIOpenDialect

sjw36:
CI is green. Please complete reviews.

mlir/lib/Dialect/MIOpen/Transforms/CloneKernels.cpp
```diff
@@ -0,0 +1,77 @@
+//===- ClonKernels.cpp ----------------------------------------------------===//
```
pcf000: 'e'

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -321,21 +322,66 @@ void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
...
+      }
+    }
+    return is_zero;
+  }
+
+  void traceInputs(Operation *op, SmallVector<Operation *> &predecessors,
+                   SetVector<Value> &inputNodes) {
+    for (const auto &opnd : op->getOperands()) {
+      Operation *usedOp = opnd.getDefiningOp();
+      if (usedOp && (isConstantZero(usedOp) || isElementwiseOp(usedOp))) {
```
pcf000, line 368: I had a hasOneUse() test in here because we're absorbing the op into the outlined kernel, removing the original, and there's no way for a second use to see it after that.  I notice that you just don't erase nodes that still have uses, but is that sufficient for correct behavior?

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
jerryyin: It is a confusing way to handle it, anyone who does not review this PR may not know what legacy means. Besides, you are not using this as a real option but a convenient way to set another option. Since both are static variables, the sequence which one get initialized may be different from your expectation thus creating surprises.

I don't think it is a huge problem if we need to update all tests, however it will be a serious pollution to this PR's diff. Could you do the following:
  - In this PR, revert the `kernelPipeline` to `loweringTargetDialect`. This allows you to just extend rather than break tests.
  - Once this PR is merged, rename the `loweringTargetDialect` to `kernelPipeline`, and update all tests. Do all this and only this change in the new PR such that it is possible to compose/review cleanly.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -321,21 +322,66 @@ void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
...
+    }
   }
 }
 
 // Inspired by / adapted from TestSCFIfUtilsPass in
 // test/lib/Transforms/TestSCFUtils.cpp.
 class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
 public:
+  bool isConstantZero(Operation *op) {
+    // test for zero
```
pcf000, line 357: Can this be simplified using llvm::all_of(), as in isZeroAttribute() in external/llvm-project/mlir/lib/Dialect/LLVMIR/IR/LLVMDialect.cpp?

mlir/tools/mlir-rocm-runner/mlir-rocm-runner.cpp
```diff
@@ -67,17 +68,19 @@ static LogicalResult runMLIRPasses(ModuleOp m) {
...
-  if (tripleName.empty() && targetChip.empty() && features.empty()) {
-    systemOverride = true;
-  }
-
   int optLevel = gpuOpt.getValue();
   if (optLevel < 0 || optLevel > 3) {
     llvm::errs() << "Invalid GPU optimization level: " << optLevel << "\n";
     return failure();
   }
-  BackendUtils utils(tripleName, targetChip, features, systemOverride);
```
jerryyin, line 80: You will have to keep the old behavior, or at least leave it to another review.

Without `systemOverride` you must guarantee all options passed in are valid, otherwise they can all remain to be empty.

pcf000 (changes requested):

mlir/lib/ExecutionEngine/ROCm/BackendUtils.cpp
```diff
@@ -79,17 +79,16 @@ void getGpuGCNArchName(hipDevice_t device, std::string &gcnArchName) {
...
 BackendUtils::BackendUtils(const std::string &defaultTriple,
                            const std::string &defaultChip,
-                           const std::string &defaultFeatures,
-                           bool systemOverride)
+                           const std::string &defaultFeatures)
     : triple(defaultTriple), chip(defaultChip), features(defaultFeatures) {
-  if (systemOverride) {
+  if (triple.empty()) {
     triple = kTargetTriple;
+  }
```
jerryyin: No, if the client didn't give us a valid triple, we should not replace it with our own but throw them an error. This can be guaranteed with `systemOverrid` but right now we will swap an empty string with ours thus allow the error to propogate.

Same goes for `chip` and `features`. Besides, `configTarget` will not guarantee to give a valid `chip`/`features` combination. @yiqian1 is working on that.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
```diff
@@ -55,86 +56,166 @@ static cl::opt<std::string> outputFilename("o", cl::desc("Output filename"),
...
+                                       cl::value_desc("triple string"),
+                                       cl::init(""));
+
+static cl::opt<std::string> targetChip("target", cl::desc("target chip"),
+                                       cl::value_desc("AMDGPU ISA version"),
+                                       cl::init(""));
+
+static cl::opt<std::string> features("feature", cl::desc("target features"),
+                                     cl::value_desc("AMDGPU target features"),
+                                     cl::init(""));
```
jerryyin, line 97: Though agree that those information has to be there in attempt to compile to binary target, I'm a bit hesitant that we blur the responsibility between the `mlir-miopen-driver` and `mlir-rocm-runner` like this. 

I think we need at least to add a check to make sure: 
 - When kernelPipeline is set to binary, the user must provide value to all these otherwise throw an error
 - When kernelPipeline is not set to binary, the user must not provide value to those because they will not be useful at all

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
```diff
@@ -55,86 +56,166 @@ static cl::opt<std::string> outputFilename("o", cl::desc("Output filename"),
...
-static void populateTuningPipeline(PassManager &pm) {
-  pm.addPass(
-      mlir::miopen::createAffixTuningParametersPass(blockSize, gridSize));
+static LogicalResult
+parsePipeline(StringRef pipeline, llvm::SmallDenseSet<StringRef> &pipelineSet,
+              llvm::SmallDenseSet<StringRef> &pipelineOptions) {
+  size_t start = 0;
+  size_t pos = 0;
+  do {
+    pos = pipeline.find_first_of(",", start);
```
jerryyin: I'm pretty sure you should use `find` instead of `find_first_of`. Although interchangeable at this point, it will immediately evolve into an error once someone change the target into a string.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
```diff
@@ -55,86 +56,166 @@ static cl::opt<std::string> outputFilename("o", cl::desc("Output filename"),
...
+    auto opt = pipeline.substr(start, pos).trim();
+    if (opt.empty()) {
+    } else if (opt == "all" || opt == "full") {
+      pipelineSet = pipelineOptions;
```
jerryyin: Just from kernel pipeline aspect, aren't `full`, `all` and `binary` aliases? Why do we need the additional two aliases of `full` and `all`?

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
```diff
@@ -55,86 +56,166 @@ static cl::opt<std::string> outputFilename("o", cl::desc("Output filename"),
...
+      pipelineSet.insert(opt);
+    } else {
+      SmallString<256> opts;
+      for (auto pipeline : pipelineOptions) {
+        if (opts.size())
+          opts += ", ";
+        opts += pipeline;
+      }
+      llvm::errs() << "Invalid pipeline: " << opt << "\n"
+                   << "   Valid options: " << opts << " or full\n";
```
jerryyin: This check, together with `kernelPipelineOptions` should not be defined here, but as a validation check before `runMLIRPasses` invoked.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
pcf000: StringRef has a split() call, used something like "pipeline.split(optVector, ',')", which could make find_first_of() moot.

mlir/tools/mlir-rocm-runner/mlir-rocm-runner.cpp
sjw36: I just moved the empty check inside the BackendUtils constructor.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
sjw36: Yes, it is sufficient because we clone the op, it is not actually moved. Then later if all uses of the original are deleted it gets removed.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
jerryyin: Superseded by later discussions, marking this as resolved.

mlir/lib/Dialect/MIOpen/Transforms/ApplyImpl.cpp
jerryyin: Sounds good

mlir/lib/Dialect/MIOpen/Transforms/ApplyImpl.cpp
```diff
@@ -0,0 +1,112 @@
...
+        gpuMods.push_back(gpuMod);
+
+        // make global constant in TopModule
+        SmallString<128> binaryCstName(
+            SymbolTable::getSymbolName(gpuMod).getValue());
+        binaryCstName.append("_gpubin_cst");
+        // uniquify if necessary
+        while (symbolTable.lookup(binaryCstName)) {
+          binaryCstName += "0";
+        }
```
jerryyin: Hmmm... Can we do better than this? `_gpubin_cast00000...000` will be a possibility?

mlir/lib/Dialect/MIOpen/Transforms/ApplyImpl.cpp
jerryyin: Superseded by later discussions

mlir/lib/Dialect/MIOpen/Transforms/CloneKernels.cpp
jerryyin: I get it that for this pass you will populate a `__miopen` (aka kernel) module the copy all the kernel into that module. Am I right that after that for each kernel, we will have a copy such that:
 - The kernel in the original module op will be there for possibilities to compile differently
 - The kernel in the `__miopen` module will be compiled all the way down the binary and then referenced from the original module op

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
```diff
@@ -55,86 +56,166 @@ static cl::opt<std::string> outputFilename("o", cl::desc("Output filename"),
...
+      return failure();
+    }
+  }
+
+  // Find kernel module, defaults to top module
+  auto kernelModule =
+      module.lookupSymbol<ModuleOp>(miopen::MIOpenDialect::kKernelModuleName);
+  if (!kernelModule) {
+    kernelModule = module;
+  }
```
jerryyin, line 168: Aren't we guaranteed to have a kernel module at this point? If so I'd prefer this as an assert rather than we default kernel module to module.

mlir/lib/Dialect/MIOpen/Transforms/CloneKernels.cpp
jerryyin: BTW, when the compilation finished and that we reference the kernel as global const, does `mlir-rocm-runner` understand about how to execute a partial compiled graph like that? What if that partial compiled graph does not match with the actual target system, then what do we do? Do we feed the graph back again into the `mlir-miopen-driver`?

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
sjw36: Good call, I will look into that.

mlir/lib/Dialect/MIOpen/Transforms/ApplyImpl.cpp
krzysz00: Yeah, there's code that goes 0,1,2,... in the gpu.printf lowering, might be worth looking at that

mlir/lib/ExecutionEngine/ROCm/BackendUtils.cpp
sjw36: Ok, so I can make the same check for all three are empty. This would be equivalent to previous code in mlir-rocm-runner that just set systemOverride based on same condition. This is the only place that uses BackendUtils, and now also in mlir-miopen-driver.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
sjw36: For 'feature', is empty a valid value? If so, then we cannot say empty is invalid.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
sjw36: Yes, it is an alias for just run the entire pipeline. It is more relevant for the host-pipeline where different pipelines are non-overlapping.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
krzysz00: Yeah, you can have an empty features string

mlir/lib/Dialect/MIOpen/Transforms/ApplyImpl.cpp
sjw36: sure, I will add an incrementing suffix.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
sjw36: It is not guaranteed at this point. The old pipeline and tests still work because we do not require an __miopen module.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
sjw36: Seems like there should be a standard way of doing this but I cannot find a cl::opt way to specify a list of values. This is similar to the opt 'pass-pipeline' but much more simple.

sjw36:
@jerryyin , @pcf000 Are all issues resolved?

pcf000:
Looks good from here.

pcf000 (approved):

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
jerryyin: I think only one option as wildcard should be enough. Also, could you add this to the `hostPipeline::value_desc` comments so user know this is a valid option.

jerryyin:
@sjw36 Please respond to the last 3 unresolved ones. Other than those it looks solid, thanks!

sjw36:
> @sjw36 Please respond to the last 3 unresolved ones. Other than those it looks solid, thanks!

I responded to all issues. Let me know which are unsatisfactory and why.

jerryyin:
@sjw36 Basically all the three that didn't get marked as resolved. I wouldn't call them "unsatisfactory" ones, just seeking for continue the conversation there. For example, the one related with static initialization ordering(`legacyMiopenPipeline`), how are you planning to address it?

https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/500#discussion_r801812773
https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/500#discussion_r801899765
https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/500#discussion_r803013953

mlir/lib/Dialect/MIOpen/Transforms/CloneKernels.cpp
sjw36: In Phase 2, I will update the mlir-rocm-runner to understand this XMIR format but for now (phase 1) it will not work.

jerryyin: Sounds good.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
sjw36: I can do that but I really see option -c as a short-hand or alias anyway, which depends on -target to determine which pipeline gets run.

jerryyin: If you want a short hand that's fine, however I think they way this is implemented will cause OB.

```cpp
if (v) {
  kernelPipeline.setValue("gpu");
}
```

Now `kernelPipeline` is also a static variable. The code chunk above assumes one static variable `kernelPipeline` to be initialized before another static variable `legacyMiopenPipeline`. I believe using it this way caused undefined behavior. What would happen if `kernelPipeline` is not initialized if you insist on using it this way?

sjw36: Actually no that is a lambda that is only called if -c is encountered in cl::ParseCommandLineOptions.

jerryyin: Hmm sounds reasonable, so this is no longer a blocker. I still find this behavior (of using one option to alias another) to be very tricky. Why not initialize `kernelPipeline` directly as "gpu" but use another option to do it? You can put the lambda into `kernelPipeline` as well.

sjw36: Well, you may want to run only the host-pipeline or you may want to run specific passes, so defaults to no kernel passes. 

BTW, -c was an alias already and I also use the technique in miopen-gen for short-hand options like -pv.

jerryyin (approved):

PR #501, opened 2022-02-07
Fix side effects caused by #498.

whchung:
Fix side effects caused by #498 . 2 issues identified:

- For `f16` type, we should still allow `KPACK` be 1. We really just don't want to see `KPACK` be set as 2, which is too narrow for XDLOPS `f16` instructions.
- When `KPACK` is larger than 1, we need to check if `GemmK` dimension can be divided by `KPerBlock * KPACK`.

All 3 failing configs raised in https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/427 are fixed in this PR. This should unblock the CI.

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
@@ -567,15 +567,6 @@ protected:
     return (gemmSize.gemmM / param->gemmMPerBlock) *
            (gemmSize.gemmN / param->gemmNPerBlock) * gemmSize.gemmG;
   }
-
-  mlir::LogicalResult isValidGemm(InitParams *param, GemmSize &gemmSize) {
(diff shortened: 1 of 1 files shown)


mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
```diff
@@ -939,6 +939,7 @@ class PopulateParamsXDL : public PopulateParamsBase {
                    << "\n";
       return failure();
     } else if ((dataType.isF16() || dataType.isBF16()) &&
+               (param.gemmKPack != 1) &&
                ((param.gemmKPack < 4) || (param.gemmKPack > 8))) {
```
jerryyin, line 943: I think if we want to be explicit, we can explicitly reject 2 and allow everything including 1,4,8:

```cpp
// KPACK of 2 is too narrow for XDLOPS f16 instructions
((param.gemmKPack == 2) || (param.gemmKPack > 8))
```

jerryyin (approved):
LGTM, one minor nit

PR #502, opened 2022-02-08
Clean up the use of rocm_agent_enumerator

yiqian1:
Per conversation with Joseph Greathouse, it'd be problematic if rocm_agent_enumerator might be called a huge number of times in parallel during the build process. This patch reverts the use of "rocm_agent_enumerator -name" in BackendUtils.cpp.

### DIFF
--- a/mlir/lib/ExecutionEngine/ROCm/BackendUtils.cpp
+++ b/mlir/lib/ExecutionEngine/ROCm/BackendUtils.cpp
@@ -44,7 +44,7 @@
 
 // If an old rocm_agent_enumerator that has no "-name" option is used, rely on
 // the hip runtime function to provide GPU GCN Arch names.
-#if __USE_ROCM_4_4_OR_OLDER__
+#if __INCLUDE_HIP__
 #pragma GCC diagnostic push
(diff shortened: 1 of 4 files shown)


mlir/lib/ExecutionEngine/ROCm/BackendUtils.cpp
```diff
@@ -168,15 +161,6 @@ void BackendUtils::configTarget(std::string &targetChip,
...
-  for (llvm::line_iterator lines(*gfxArchList); !lines.is_at_end(); ++lines) {
-    // Use the first Arch name found.
-    gcnArchName = lines->str();
-    break;
-  }
-  auto status = IsaNameParser::parseArchName(gcnArchName, targetChip, features);
-  if (status.failed()) {
-    llvm_unreachable("ArchName parsing failed.");
-  }
+  llvm_unreachable("BackendUtils.cpp does not include HIP.");
```
jerryyin, line 164: Maybe more descriptive like: You are not supposed to use `configTarget` when compiling without HIP runtime.

jerryyin (approved):
LGTM

PR #503, opened 2022-02-08
Correct BF16 arithmetic lowering

krzysz00:


whchung:
should we add tests to this PR?

krzysz00:
@whchung I was thinking it might make more sense to add the testing during a followup PR, since what's mainly missing is bf16 e2e tests

whchung:
SGTM.

whchung (approved):
Let's get this PR in and use a follow-up PR to add e2e tests.

PR #504, opened 2022-02-10
[DO NOT SQUASH] Use LLD from ROCm instead of linking our own copy

krzysz00:
Discussion with upstream revealed the lld folks don't want us calling them as a library, and discussion here showed that we can use ROCm's lld without much trouble.

This is an import of the equivalent upstream PR https://reviews.llvm.org/D119463 and a removal of lld from the build process here.

I'd like to land this before the upstream update so that this is consistent and agreed on.


jerryyin (approved):
LGTM

PR #505, opened 2022-02-11
Adding xdlopsgemm and blockwisegemm i8 implementation

jerryyin:
I did the following in this PR:
 - Added i8 definition for xdlops gemm and blockwise gemm op
 - Amended builderUtils to take i8 into account
 - Added unit tests for both amended ops

Sending the i8 PRs in reasonable smaller chunks so that it is easier to review. I'm hoping to get feedback such as to address issues early.

Going forward I plan on to work on following areas:
 - bottom up, next PR: continue to work on i8 conversion of step4 ops including:
   - ThreadwiseCopyV2
   - ThreadwiseStore
   - ThreadwiseLoad
   - InWarpTransposeOp  
 - top down, next PR: Start from a set of tuning parameters and lower from gridwisegemm -> blockwise level ops

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
+++ b/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
@@ -442,8 +442,14 @@ def MIOpen_BlockwiseGemmV2Op:
                    ArgTransforms<2>:$transforms,
                    Index:$waveOffsetA,
                    Index:$waveOffsetB,
-                   MemRefOf<[F32, F16, BF16, I8, VectorOfLengthAndType<[2, 4], [F32]>, VectorOfLengthAndType<[2, 4, 8], [F16, BF16]>, VectorOfLengthAndType<[4], [I8]>]>:$bufferA,
-                   MemRefOf<[F32, F16, BF16, I8, VectorOfLengthAndType<[2, 4], [F32]>, VectorOfLengthAndType<[2, 4, 8], [F16, BF16]>, VectorOfLengthAndType<[4], [I8]>]>:$bufferB,
+                   MemRefOf<[F32, F16, BF16, I8, 
+                             VectorOfLengthAndType<[2, 4], [F32]>, 
+                             VectorOfLengthAndType<[2, 4, 8], [F16, BF16]>, 
+                             VectorOfLengthAndType<[4], [I8]>]>:$bufferA,
+                   MemRefOf<[F32, F16, BF16, I8, 
+                             VectorOfLengthAndType<[2, 4], [F32]>, 
+                             VectorOfLengthAndType<[2, 4, 8], [F16, BF16]>, 
+                             VectorOfLengthAndType<[4], [I8]>]>:$bufferB,
                    Variadic<VectorOfRankAndType<[1], [F32, F16, I32]>>:$vectorCs)>,
     Results<(outs Variadic<VectorOfRankAndType<[1], [F32, F16, I32]>>: $vectorDs)> {
   let summary = "Blockwise GEMM XDLOPS version";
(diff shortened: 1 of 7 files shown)


mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -437,15 +437,15 @@ def MIOpen_BlockwiseGemmOp:
...
+    Arguments<(ins MemRefOf<[F32, F16, BF16, I8]>:$matrixA,
+                   MemRefOf<[F32, F16, BF16, I8]>:$matrixB,
                    ArgTransforms<2>:$transforms,
                    Index:$waveOffsetA,
                    Index:$waveOffsetB,
-                   MemRefOf<[F32, F16, BF16, VectorOfLengthAndType<[2, 4], [F32]>, VectorOfLengthAndType<[2, 4, 8], [F16, BF16]>]>:$bufferA,
-                   MemRefOf<[F32, F16, BF16, VectorOfLengthAndType<[2, 4], [F32]>, VectorOfLengthAndType<[2, 4, 8], [F16, BF16]>]>:$bufferB,
-                   Variadic<VectorOfRankAndType<[1], [F32, F16]>>:$vectorCs)>,
-    Results<(outs Variadic<VectorOfRankAndType<[1], [F32, F16]>>: $vectorDs)> {
+                   MemRefOf<[F32, F16, BF16, I8, VectorOfLengthAndType<[2, 4], [F32]>, VectorOfLengthAndType<[2, 4, 8], [F16, BF16]>, VectorOfLengthAndType<[4], [I8]>]>:$bufferA,
```
krzysz00: Shouldn't those VectorOfLengthAndType<[4], [I8]> be something like [4, 16], or am I misremembering the context?

Also, while you're here, could you break those constraints up onto multiple lines - they've gotten really long.

mlir/lib/Dialect/MIOpen/utility/builderUtils.cpp
```diff
@@ -19,14 +19,22 @@ using mlir::arith::ConstantOp;
...
+  if (auto vecType = type.dyn_cast<VectorType>()) {
+    retValue =
+        b.create<ConstantOp>(loc, SplatElementsAttr::get(vecType, value));
+  } else {
+    retValue = b.create<ConstantOp>(loc, b.getI8IntegerAttr(value), type);
+  }
+
+  return retValue;
+}
+
```
krzysz00, line 35: Why not expose this? We might want it in miopen-gen.cpp later

krzysz00 (commented):
Minor nitpicks, otherwise looks solid to me - thanks for keeping up on this!

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
jerryyin: Sure, will break to make it shorter.

The reason for `VectorOfLengthAndType<[4], [I8]>` definition is that for both xdlops exposed they are only able to execute the argument of `i32`, or equivalent to 4xi8.

In the lowered code, bufferA and bufferB will load temporary result and get directly feed into the xdlops instruction therefore it has to match the length of argument with xdlops.

mlir/lib/Dialect/MIOpen/utility/builderUtils.cpp
jerryyin: Just trying to follow the defensive programming style that we can expose it when we need it. It's going to be a fairly trivial change. By then we can judge how and what to expose as public.

But right now since there's no consumer I'm just going to hide the implementation in this compilation unit only.

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
krzysz00: Noted that both the xdlops just take i32 - for some reason, I thought some of them were wider

mlir/lib/Dialect/MIOpen/utility/builderUtils.cpp
krzysz00: Makes total sense

krzysz00 (approved):
Approved, though if the anti-long-lines changes go in before merge it'd be great

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
jerryyin: Agreed, I think you meant to say that the data preparation time we can take advantage of kpack to better take advantage of the bandwidth, which will translate to more inner loops in xdlops execution.

jerryyin: breaking lines addressed in [700a3a9](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/505/commits/700a3a9c5b5a3522b38d16dfc9cb77cfa984018b)

PR #508, opened 2022-02-14
Enable int8 from step 1 to step 3 lowering

jerryyin:
This PR enables conv op lowering from step 1 to step 3, with a caveats that tuning parameters needs to be re-defined for i8, existing tuning parameters are for testing only. It has been a surprisingly easy PR. After this is merged, the only left is to make sure lowering step 4 work with i8 type.

With this PR we support the below with a manually written conv op with i8 type
```bash
 ./bin/miopen-opt -miopen-affix-params -miopen-lowering -miopen-lowering-step2 -miopen-lowering-step3 ./conv_op.mlir
```
Once @yiqian1 finished her changes on conv2d generator, I can add more tests into mlir sanity on int8 using `miopen-gen`

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
+++ b/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
@@ -35,9 +35,9 @@ class ArgTransforms<int n> : Confined<ArgTransformsAttr, [ArrayCount<n>]>;
 
 def MIOpen_Conv2DOp :
     MIOpen_Op<"conv2d">,
-    Arguments<(ins MemRefRankOf<[F32, F16, BF16], [5]>:$filter,
-                   MemRefRankOf<[F32, F16, BF16], [5]>:$input,
-                   MemRefRankOf<[F32, F16, BF16], [5]>:$output)> {
+    Arguments<(ins MemRefRankOf<[F32, F16, BF16, I8], [5]>:$filter,
+                   MemRefRankOf<[F32, F16, BF16, I8], [5]>:$input,
+                   MemRefRankOf<[F32, F16, BF16, I8], [5]>:$output)> {
   let summary = "2D convolution forward";
   let description = [{
     The `miopen.conv2d` op computes 2D convolution forward.
@@ -152,9 +152,9 @@ def MIOpen_GridwiseGemmOp :
 // gridwise_gemm_v2
 def MIOpen_GridwiseGemmV2Op :
     MIOpen_Op<"gridwise_gemm_v2">,
-    Arguments<(ins MemRefRankOf<[F32, F16, BF16], [3, 4]>:$a,
-                   MemRefRankOf<[F32, F16, BF16], [3, 4]>:$b,
-                   MemRefRankOf<[F32, F16, BF16], [3, 4]>:$c,
+    Arguments<(ins MemRefRankOf<[F32, F16, BF16, I8], [3, 4]>:$a,
+                   MemRefRankOf<[F32, F16, BF16, I8], [3, 4]>:$b,
+                   MemRefRankOf<[F32, F16, BF16, I8], [3, 4]>:$c,
                    BoolArrayAttr:$aOobDims,
                    BoolArrayAttr:$bOobDims,
                    BoolArrayAttr:$cOobDims,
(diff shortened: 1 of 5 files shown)


mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -35,9 +35,9 @@ class ArgTransforms<int n> : Confined<ArgTransformsAttr, [ArrayCount<n>]>;
 
 def MIOpen_Conv2DOp :
     MIOpen_Op<"conv2d">,
-    Arguments<(ins MemRefRankOf<[F32, F16, BF16], [5]>:$filter,
-                   MemRefRankOf<[F32, F16, BF16], [5]>:$input,
-                   MemRefRankOf<[F32, F16, BF16], [5]>:$output)> {
+    Arguments<(ins MemRefRankOf<[F32, F16, BF16, I8], [5]>:$filter,
+                   MemRefRankOf<[F32, F16, BF16, I8], [5]>:$input,
+                   MemRefRankOf<[F32, F16, BF16, I8], [5]>:$output)> {
```
krzysz00: Bug: this is an `i32` output

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -152,9 +152,9 @@ def MIOpen_GridwiseGemmOp :
 // gridwise_gemm_v2
 def MIOpen_GridwiseGemmV2Op :
     MIOpen_Op<"gridwise_gemm_v2">,
-    Arguments<(ins MemRefRankOf<[F32, F16, BF16], [3, 4]>:$a,
-                   MemRefRankOf<[F32, F16, BF16], [3, 4]>:$b,
-                   MemRefRankOf<[F32, F16, BF16], [3, 4]>:$c,
+    Arguments<(ins MemRefRankOf<[F32, F16, BF16, I8], [3, 4]>:$a,
+                   MemRefRankOf<[F32, F16, BF16, I8], [3, 4]>:$b,
+                   MemRefRankOf<[F32, F16, BF16, I8], [3, 4]>:$c,
```
krzysz00: Simalarly, I32 on output

krzysz00 (commented):
I'm not sure about the output types, but this overall seems reasonable

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
jerryyin: Nice catch, thank you!

jerryyin:
@krzysz00 Gentle ping for taking a second look

mlir/lib/Dialect/MIOpen/utility/builderUtils.cpp
```diff
@@ -72,6 +85,8 @@ Value createZeroConstantOp(OpBuilder &b, Location loc, Type type) {
 
   if (elementType == b.getIntegerType(8)) {
     return createConstantI8Op(b, loc, type, elementType, 0);
+  } else if (elementType == b.getIntegerType(32)) {
```
krzysz00: ... hold on, we can use the same trick as with `createConstantFloatOp` to make the code generic over integer types.

krzysz00 (commented):
Looks good to me overall, but there's a refactor of the constant code I'd like to see

mlir/lib/Dialect/MIOpen/utility/builderUtils.cpp
jerryyin: Again, nice catch! I wouldn't know there's a `APInt` exist without you poking at it. Addressed in adb57e3

krzysz00 (approved):

jerryyin:
> ... hold on, is the int8 in the int8 kernel signed or unsigned?

Should be signed. Did you stare at `isIntOrIndex` and wondering if that should be swapped to a `isSignedInteger()`? I can make this update but it probably don't matter now.

krzysz00:
@jerryyin Nah, I was staring at the parameters to APInt, where, IIRC, the boolean is signed vs unsigned. In which case, can we go and make sure that the non-xdlops path is using signed extension instead of unsigned?

jerryyin:
@krzysz00 I touched zero per cent of non-xdlops int8 code. Not sure it's even part of the roadmap. But let me get it documented in a separate ticket. It deserve a reexamine on the entire lowering pipeline.

PR #509, opened 2022-02-15
Move Dockerfile to ROCm 5.0

krzysz00:
Per omkar, MI-200 nodes should have a ROCm version >= 4.5
Since we're unlikely to break anyone else with a longer upgrade, go straight to 5.0.

Note that this PR doesn't actually move us over, it just causes the new Docker image to be built.


jerryyin (approved):
LGTM

PR #510, opened 2022-02-15
MLIR-435: fixed crash in multiple calls to MIOpenToTosaPass.

sjw36:
See ticket: https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/435


jungpark-mlir (approved):
All good

sjw36:
This fixes the crash but does not fix the tests which were broken by the LLVM update. @pcf000 will follow up with a PR to restore the AbstractElementwise trait to Tosa Elementwise ops.

pcf000:
> @pcf000 will follow up with a PR to restore the AbstractElementwise trait to Tosa Elementwise ops.

I can fix two partition-related tests, but tosa-to-miopen-relu.xmir.mlir looks like a binary-blob mismatch and I don't know where that's coming from.  It will stay XFAIL.

jungpark-mlir:
Kicked the PR CI again, previous run failed because of a failure in MIOpenDriver test caused by gpu page fault. Likely to be occurred by external reason.

PR #511, opened 2022-02-15
Move all MIGraphX ops to assemblyFormat

krzysz00:


jungpark-mlir (approved):
Looks good, thanks for cleaning this up.

PR #439, opened 2021-11-09
Update Jenkinsfile to use weekly CI results

yiqian1:
- Nightly CI copies the achieved user perfDb from the most recent successful weekly for performance tests.
- Nightly CI runs MIOpen tests on two selected Resnet50 configs (filter 1x1 and 3x3).

These changes have passed validations:
http://ml-ci.amd.com:21096/job/MLIR/job/test-nightlyCI/13/
http://rocmhead:8080/job/test-nightlyCI/3/

Weekly CI job's name is mlir-weekly.
http://ml-ci.amd.com:21096/job/MLIR/job/mlir-weekly/2/
http://rocmhead:8080/job/mlir-weekly/4/

PR #512, opened 2022-02-15
End-to-end impl for fp16 backward weight convolution using atomic add.

whchung:
Rebased with the tip.

Using atomic adds under the following conditions:
- backward weight convolution
- fp16
- XDLOPS
- do not use padding along Gemm M/N/K dimension

If all conditions are met, an extra workspace is required to hold the result in fp32 data type. An `Miir` interface is added to let clients query how large the workspace should be.

`miopen-gen` is modified to automatically change the kernel signature when such an extra workspace is needed.

Utility functions inside `Conv2dGenerator` are modified to compute if a config requires a workspace. Notice there are hard-coded tuning paremeters in the current implementation which is very not ideal.

Added test cases to check various configs on whether a workspace is needed.

Empirical performance tests show 3.5X performance improvements on supported resnet50 fp16 backward weight convolution configs.

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
+++ b/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
@@ -67,7 +67,8 @@ def MIOpen_Conv2DBwdWeightOp :
     MIOpen_Op<"conv2d_bwd_weight">,
     Arguments<(ins MemRefRankOf<[F32, F16, BF16], [5]>:$filter,
                    MemRefRankOf<[F32, F16, BF16], [5]>:$input,
-                   MemRefRankOf<[F32, F16, BF16], [5]>:$output)> {
+                   MemRefRankOf<[F32, F16, BF16], [5]>:$output,
+                   Variadic<MemRefRankOf<[F32], [5]>>:$workspace)> {
   let summary = "2D convolution backward weight";
   let description = [{
     The `miopen.conv2d_bwd_weight` op computes 2D convolution backward weight.
--- a/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
+++ b/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
@@ -68,7 +68,7 @@ def MIOpen_Conv2DBwdWeightOp :
     Arguments<(ins MemRefRankOf<[F32, F16, BF16], [5]>:$filter,
                    MemRefRankOf<[F32, F16, BF16], [5]>:$input,
                    MemRefRankOf<[F32, F16, BF16], [5]>:$output,
-                   Variadic<MemRefRankOf<[F32], [5]>>:$workspace)> {
+                   Optional<MemRefRankOf<[F32], [5]>>:$workspace)> {
   let summary = "2D convolution backward weight";
   let description = [{
     The `miopen.conv2d_bwd_weight` op computes 2D convolution backward weight.
--- a/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
@@ -1074,4 +1074,61 @@ public:
   }
 };
 
+// The function is used to compute extra padding sizes.
+// For example, if gemmM size is 3 and gemmMPerBlock is 64,
+// we set gemmMExtra be 64 so (gemmM+gemmMExtra)%gemmMPerBlock=0.
+//
+// Returns:
+// - isOriginalKernelSupport : a bool only used in backward data convolution.
+// - needExtraPad : a bool to indicate whether padding kernel is needed.
+// - gemmM/N/KExtra : additional padding required along Gemm M/N/K dimension.
+//                    They would be all be 0 in case needExtraPad is false.
+template <typename T>
+std::tuple<bool, bool, int64_t, int64_t, int64_t>
+calculatePaddingKernelSize(int64_t gemmMSize, int64_t gemmNSize,
+                           int64_t gemmKSize, ConvolutionContext &convContext,
+                           T populateParams) {
+  bool isOriginalKernelSupport = true;
+  bool needExtraPad = false;
+  int64_t gemmMExtra, gemmNExtra, gemmKExtra;
+  gemmMExtra = gemmNExtra = gemmKExtra = 0;
+
+  auto configParams = populateParams.getTuningParameters(convContext);
+  size_t numOfFailedConfigs = 0;
+  for (auto &params : configParams) {
+    if (gemmMSize % params.gemmMPerBlock == 0 &&
+        gemmKSize % params.gemmKPerBlock == 0 &&
+        gemmNSize % params.gemmNPerBlock == 0) {
+      isOriginalKernelSupport = true;
+      break;
+    }
+    isOriginalKernelSupport = false;
+    numOfFailedConfigs++;
+  }
+
+  auto extraParams = populateParams.getUniversalParameters();
+  if (numOfFailedConfigs == configParams.size()) {
+    needExtraPad = true;
+    int64_t gemmMRemain, gemmKRemain, gemmNRemain;
+
+    gemmMRemain = gemmMSize % extraParams.gemmMPerBlock;
+    if (gemmMRemain != 0)
+      gemmMExtra = extraParams.gemmMPerBlock - gemmMRemain;
+
+    gemmNRemain = gemmNSize % extraParams.gemmNPerBlock;
+    if (gemmNRemain != 0)
+      gemmNExtra = extraParams.gemmNPerBlock - gemmNRemain;
+
+    gemmKRemain = gemmKSize % extraParams.gemmKPerBlock;
+    if (gemmKRemain != 0)
+      gemmKExtra = extraParams.gemmKPerBlock - gemmKRemain;
+
+    // llvm::errs() << "gemmMExtra: " << gemmMExtra << "gemmNExtra: " <<
+    // gemmNExtra << "gemmKExtra: " << gemmKExtra << "\n";
+  }
+
+  return std::make_tuple(isOriginalKernelSupport, needExtraPad, gemmMExtra,
+                         gemmNExtra, gemmKExtra);
+}
+
 #endif // MLIR_DIALECT_MIOPEN_GRIDWISE_GEMM_PARAMS_H
--- a/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
@@ -778,7 +778,7 @@ public:
                 DerivedOutParams &gemmCDerivedParam, int64_t &gridSize);
 
   llvm::SmallVector<InitParamsNonXDL, 8>
-  getTuningParameters(ConvolutionContext &ctx) {
+  getTuningParameters(miopen::ConvOpType dir, mlir::Type dataType) {
     return initParameters;
   }
 
@@ -1045,13 +1045,12 @@ public:
                               int64_t &blockSize, int64_t &gridSize);
 
   llvm::SmallVector<InitParamsXDL, 4>
-  getTuningParameters(ConvolutionContext &ctx) {
-    auto dataType = ctx.getDataType();
+  getTuningParameters(miopen::ConvOpType dir, mlir::Type dataType) {
     if (dataType.isInteger(8)) {
       return initParametersForwardI8;
     }
 
-    switch (ctx.getOpType()) {
+    switch (dir) {
     case miopen::ConvOpType::Fwd:
       return initParametersForward;
     case miopen::ConvOpType::BwdData:
@@ -1086,14 +1085,14 @@ public:
 template <typename T>
 std::tuple<bool, bool, int64_t, int64_t, int64_t>
 calculatePaddingKernelSize(int64_t gemmMSize, int64_t gemmNSize,
-                           int64_t gemmKSize, ConvolutionContext &convContext,
-                           T populateParams) {
+                           int64_t gemmKSize, miopen::ConvOpType dir,
+                           mlir::Type dataType, T populateParams) {
   bool isOriginalKernelSupport = true;
   bool needExtraPad = false;
   int64_t gemmMExtra, gemmNExtra, gemmKExtra;
   gemmMExtra = gemmNExtra = gemmKExtra = 0;
 
-  auto configParams = populateParams.getTuningParameters(convContext);
+  auto configParams = populateParams.getTuningParameters(dir, dataType);
   size_t numOfFailedConfigs = 0;
   for (auto &params : configParams) {
     if (gemmMSize % params.gemmMPerBlock == 0 &&
--- a/mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
+++ b/mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
@@ -318,6 +318,83 @@ Type Conv2dGenerator::getDataType(OpBuilder &builder) const {
   return dataType;
 }
 
+bool Conv2dGenerator::hasWorkspace(OpBuilder &builder) const {
+  // Decide if a workspace is needed.
+  // Preconditions:
+  // - data type: fp16
+  // - operation: backward weight conv2d.
+  // - use XDLOPS.
+  // - No need to pad along Gemm M/N/K dimension.
+  bool result = false;
+
+  if (config.operation.hasValue()) {
+    mlir::Type dataType = getDataType(builder);
+    if ((config.operation.getValue() == miopen::ConvOpType::BwdWeight) &&
+        config.xdlops && (dataType == builder.getF16Type())) {
+
+      // Logic to check if we need to pad along GemmM/N/K dimension for backward
+      // weight convolution.
+      // FIXME. Merge with logic used in the tuning process.
+      auto inDim = canonicalizeDims(config.inputDimension, config.inputLayout);
+      auto filDim =
+          canonicalizeDims(config.filterDimension, config.filterLayout);
+      auto outDim =
+          canonicalizeDims(config.outputDimension, config.outputLayout);
+
+      int64_t gemmMSize, gemmNSize, gemmKSize;
+      gemmMSize = filDim["k"];
+      gemmKSize = outDim["n"] * outDim["h"] * outDim["w"];
+      gemmNSize = filDim["c"] * filDim["y"] * filDim["x"];
+
+      // Initial tuning parameters for backward convolution.
+      llvm::SmallVector<llvm::SmallVector<int64_t, 6>, 7>
+          initParametersBackward = {
+              // M/block N/block K/block M/wave N/wave kPack
+              {128, 128, 8, 64, 64, 1}, {128, 128, 16, 64, 64, 1},
+              {8, 64, 8, 8, 64, 1},     {4, 64, 16, 4, 64, 1},
+              {32, 64, 4, 32, 64, 1},   {16, 16, 16, 16, 16, 1},
+              {16, 16, 4, 16, 16, 1},
+          };
+
+      bool needExtraPad = false;
+      size_t numOfFailedConfigs = 0;
+      for (auto &params : initParametersBackward) {
+        if (gemmMSize % params[0] == 0 && gemmNSize % params[1] == 0 &&
+            gemmKSize % params[2] == 0) {
+          break;
+        }
+        numOfFailedConfigs++;
+      }
+
+      if (numOfFailedConfigs == initParametersBackward.size()) {
+        needExtraPad = true;
+      }
+
+      // In case we need extra padding, do not use workspace.
+      result = (needExtraPad == false);
+    }
+  }
+  return result;
+}
+
+int Conv2dGenerator::getWorkspaceSize(ModuleOp &module) const {
+  // Currently onlt in the following condition would a workspace is needed.
+  // - data type: fp16
+  // - operation: backward weight conv2d.
+  // - use XDLOPS.
+  // - No need to pad along Gemm M/N/K dimension.
+  // Workspace size is the same as the filter dimension, with fp32 type.
+  int result = 0;
+  OpBuilder builder(module.getContext());
+  if (hasWorkspace(builder)) {
+    result = std::accumulate(config.filterDimension.begin(),
+                             config.filterDimension.end(), 1,
+                             std::multiplies<int>()) *
+             builder.getF32Type().getWidth() / 8;
+  }
+  return result;
+}
+
 LogicalResult Conv2dGenerator::parseConvConfig(const char *arguments) {
   std::map<std::string, std::string> argMap;
   strToTokens(arguments, argMap);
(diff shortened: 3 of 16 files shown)


mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1318,7 +1314,7 @@ struct GridwiseGemmV2RewritePattern
     auto loc = op.getLoc();
 
     // Obtain data type.
-    auto elementType = op.c().getType().cast<MemRefType>().getElementType();
+    auto elementType = op.b().getType().cast<MemRefType>().getElementType();
```
jerryyin: I happen to need to change this as well because the kernel return arg C type is different with A or B matrix. However, how does the fp16 not using atomic add being able to work before (I'm assuming C type is fp32, so the old code sort of give us the element type of fp32?)

mlir/tools/mlir-miopen-lib/mlir-miopen-lib-test.cpp
```diff
@@ -43,6 +43,8 @@ int main(int argc, char **argv) {
     std::cout << "ExecutionDims - globalSize=" << globalSize
               << ", localSize=" << localSize << std::endl;
 
+  } else if (option.getValue() == "workspace") {
+    std::cout << "Workspace=" << miirGetWorkspaceSize(handle) << std::endl;
```
jerryyin, line 47: Mind adding at least 1 unit test on the library test side with workspace size no equal to zero?

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
```diff
@@ -318,6 +318,83 @@ Type Conv2dGenerator::getDataType(OpBuilder &builder) const {
...
+  bool result = false;
+
+  if (config.operation.hasValue()) {
+    mlir::Type dataType = getDataType(builder);
+    if ((config.operation.getValue() == miopen::ConvOpType::BwdWeight) &&
+        config.xdlops && (dataType == builder.getF16Type())) {
+
+      // Logic to check if we need to pad along GemmM/N/K dimension for backward
+      // weight convolution.
+      // FIXME. Merge with logic used in the tuning process.
```
jerryyin: Why do you need to execute a portion of the tuning logic beforehand? I think you only need to check if all preconditions are met. Then if it ended up to be a config that can't satisfy the below, you can let tuning step to reject it naturally.

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -291,8 +301,9 @@ LogicalResult backwardWeightAtomicAdd(Conv2DBwdWeightOp op,
     addKBlockWrap.passThrough({"k", "c", "y", "x"});
 
     TransformMapAttr addKBlockTransformAttr = addKBlockTransform.get();
-    Value withKBlock =
-        b.create<TransformOp>(loc, op.filter(), addKBlockTransformAttr);
+    Value filterTensorInUse = (hasWorkspace) ? op.workspace()[0] : op.filter();
+    Value withKBlock = b.create<miopen::TransformOp>(loc, filterTensorInUse,
+                                                     addKBlockTransformAttr);
```
jerryyin: Out of curiosity, is this the key logic for using the workspace? I'm asking because I don't see much relevant logic below gridwise lowering.

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -67,7 +67,8 @@ def MIOpen_Conv2DBwdWeightOp :
     MIOpen_Op<"conv2d_bwd_weight">,
     Arguments<(ins MemRefRankOf<[F32, F16, BF16], [5]>:$filter,
                    MemRefRankOf<[F32, F16, BF16], [5]>:$input,
-                   MemRefRankOf<[F32, F16, BF16], [5]>:$output)> {
+                   MemRefRankOf<[F32, F16, BF16], [5]>:$output,
+                   Variadic<MemRefRankOf<[F32], [5]>>:$workspace)> {
```
jerryyin: To be honest, I almost feel like this should be a different high level op.

mlir/tools/mlir-miopen-lib/mlir-miopen-lib-test.cpp
whchung: We do have a test to check non-zero workspace size. Please refer to `workspace.mlir` in this PR.

whchung:
@jerryyin CI failing at `Test selected MIOpen configs` because MIOpen change to allocate workspace has not been submitted. On my local working MIOpen branch, `mlir-workspace`, the implementation is ready. How should we solve this chicken-and-egg problem?

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
whchung: Yes. This is where we specify the true output in a backward weight convolution when we use atomic add.

jerryyin:
@whchung In this same PR, make a change in jenkinsfile and swap the miopen develop branch from our CI with your branch. Then after your MIOpen PR is merged into MIOpen, submit another PR to swap that back to develop branch.

mlir/test/mlir-miopen-lib/workspace.mlir
```diff
@@ -0,0 +1,45 @@
...
+// RUN: mlir-miopen-lib-test --args " --x2 1 --operation conv2d_bwd_weight --arch amdgcn-amd-amdhsa:gfx908:sramecc+:xnack- --num_cu 120 --in_type fp32 --fil_type fp32 --out_type fp32 --fil_layout GNCHW --in_layout NGCHW --out_layout NGCHW --batchsize 64 --in_channels 1024 --out_channels 1024 --in_h 14 --in_w 14 --out_h 14 --out_w 14 --fil_h 1 --fil_w 1 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h 0 --padding_w 0 --kernel_name conv2d_nchw_kcyx_nkhw --groupsize 1" --option workspace | FileCheck %s --check-prefix=WORKSPACE_ZERO
+// fp16+XDLOPS+backward weight. This would require a workspace.
+// RUN: mlir-miopen-lib-test --args " --x2 1 --operation conv2d_bwd_weight --arch amdgcn-amd-amdhsa:gfx908:sramecc+:xnack- --num_cu 120 --in_type fp16 --fil_type fp16 --out_type fp16 --fil_layout GNCHW --in_layout NGCHW --out_layout NGCHW --batchsize 64 --in_channels 1024 --out_channels 1024 --in_h 14 --in_w 14 --out_h 14 --out_w 14 --fil_h 1 --fil_w 1 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h 0 --padding_w 0 --kernel_name conv2d_nchw_kcyx_nkhw --groupsize 1" --option workspace | FileCheck %s --check-prefix=WORKSPACE_NONZERO
```
whchung, line 22: @jerryyin This is the config where we explicitly require a workspace, as it fulfills all the criteria:

- fp16
- XDLOPS
- backward weight
- not padding along Gemm M/N/K

jerryyin: Thanks, I somehow thought this is a test through `miopen-gen`. Looks good.

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
whchung: I'm open to ideas here. This part is one of the worst place I think in this PR.

This is one of the major dilemma I faced earlier this week solving the unit test failures.

One of the precondition: NOT padding along Gemm M/N/K requires understanding what tuning parameters (M/N/K per block) be picked. 

We can understand the Gemm M/N/K size from the convolution configuration, but in `Conv2dGenerator` we don't yet the what the tuning parameters would be.

And structures inside `include/Dialect/MIOpen/Tuning` and `include/Dialect/MIOpen/Generator` are not compatible. So I had to manually extract logic here.

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
whchung: I don't agree. Adding one extra top-level op would incur additional overheads in tools such as `Conv2dGenerator` and `miopen-gen`.

I will try make this `Variadic` be `Optional` on the other hand.

mlir/test/mlir-miopen-lib/workspace.mlir
whchung: `miopen-gen` tests don't require `Miir` interface so I didn't add test cases there.

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
jerryyin: > One of the precondition: NOT padding along Gemm M/N/K requires understanding what tuning parameters (M/N/K per block) be picked.

Just to understand: When this condition does not meet, we will fall back to the conv2d kernel that does not require a workspace?

Another question: If the answer is yes for the previous question, then does the tuning parameter between the kernel with/out the workspace work between one another?

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
krzysz00: Yeah, I vote for Optional here too

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
whchung: on the non-atomic add version, because each thread stores to a different location on the global memref, we do an in-place fp32->fp16 truncation before storing it.

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
whchung: If any of the preconditions is not met:
- backward weight convolution
- fp16
- XDLOPS
- not padding along Gemm M/N/K dimension

we will fallback to the conv2d kernel that does not require a workspace.

The tuning parameters with or without workspace can be used interchangeably.

krzysz00: One question: what breaks when there's gemmM/N/K padding?

Also, re duplicate tuning logic - I think what needs to happen there is to refactor that whole pile of code since we're poking around in there: make some sort of `ConvDefinition` struct that has all the relevant info and then both the generator and tuning code will operate on (or have a) `ConvDefinition` - we can also make `ConvDefinition` convertible to `ConvContext` for serialization purposes, but still.

That'd also allow us to factor out the gemm size computations that have been copied everywhere in `ConvToGemm`

jerryyin: > The tuning parameters with or without workspace can be used interchangeably.

I know that when MIOpen is facing this sort of situation, they will expose both ops. Such that:
 - Both ops are candidate. 
    - One faster op is faster but requires more memory
    - Another slower op is slower but requires less memory
 - The compromise between speed and memory is something that (MIOpen) client can choose in their specific scenarios. 
 - At the tuning time, MIOpen will choose tuning parameter independently for individual ops

I think the question goes back to whether you/we are okay with always choosing the absolutely most high performance kernel over any other constraint? If yes, then this chunk of precondition decision would be totally necessary and there's no other way around a decision like this in conv2d generator.

whchung: > One question: what breaks when there's gemmM/N/K padding?

In `miopen-dialect`, if we deliberately allow padding cases using atomic add kernel, the result would be 0 everywhere on both fp32/fp16 cases. The root cause is not known yet.


> Also, re duplicate tuning logic - I think what needs to happen there is to refactor that whole pile of code since we're poking around in there: make some sort of `ConvDefinition` struct that has all the relevant info and then both the generator and tuning code will operate on (or have a) `ConvDefinition` - we can also make `ConvDefinition` convertible to `ConvContext` for serialization purposes, but still.
> 
> That'd also allow us to factor out the gemm size computations that have been copied everywhere in `ConvToGemm`

Agree an overhaul here is necessary, I see somewhat duplicated structures like `Conv2dGenerator::Config` in the generator directory, and `ConvContext` in the tuning domain.

whchung:
jenkins: retest this please.

whchung:
Observing something new on CI logs in `Test selected MIOpen configs` stage:
```
"hipErrorNoBinaryForGpu: Unable to find code object for all current devices!"
```
I assume it's a node-specific issue. The very same test command run fine on -112.

krzysz00:
@whchung Yeah, node-specific issue. Means the node needs a GPU reset

Also, instead of "retest this please", you can go into Jenkins and call for the project to be built again

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
krzysz00: And the advantage to the overhaul is that we could memoize the gemm size computation and tuning selection, instead of running that valid tuning parameter section loop a bunch of times.

So yeah, I'd rather this not go in without an overhaul to get rid of this copy-paste.

And can we re-test the padding thing, just to see if something changed in the meantime?

whchung: Let me do an assessment on the scope of overhaul so we can prioritize it. I think an overhaul is necessary, but whether it blocks the PR and subsequent ones (MIOpen solver change, use GPU to do fp32->fp16 zeroing / conversion) is up for discussion.

krzysz00: The objection I have is that the temporary copy-paste may become much less temporary because new features will get prioritized over fixing this

whchung: That's why we use tickets to track the laundry list of known todo items.

whchung:
jenkins: retest this please.

whchung:
On CI we already have 2 passing logs but it's not reflected here so I'm triggering once again via jenkins retest command.

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
whchung: https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/439 to keep track of the new requirements identified.

whchung: As an initial step I'm thinking to abstract `calculatePaddingKernelSize` used in several places (`-miopen-affix-params`, various kernel generators, and here) into a utility function first.

This should reduce the scope of additional codes be impacted in the PR, while paving way for a more complete overhaul in #439.

jerryyin (approved):
LGTM. Please keep an eye on weekly (or run through all tuning to make sure things won't break)

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
whchung: Meanwhile, supplying a vector of heuristic tuning parameters here isn't entirely correct. But works for now.

We supply a vector of tuning parameters based on heuristics in `-miopen-affix-params` out of necessity. If a client application doesn't specify one, we must come up one for later stages.

On the other hand, at `conv2dGenerator` stage, which tuning parameter would be picked isn't known beforehand, nor do we have an interface for client applications to pass in the specification of the tuning search space so the generator could try the entire space.

So basically logic here is mostly like:
- Let's assume the client won't set a tuning parameter, so we just steal the vector we'll try in the next stage `-miopen-affix-params` and run the checks.

This logic above is not ideal, but work most of the time, and work for all of the e2e tests we do right now, because for cases which requires padding kernels, one of Gemm M / N / K would either be a odd number, or not be divided by 4.

whchung: So the dilemma we face here is:
1) `conv2dGenerator` is not a tuner (ex: solver within MIOpen), it possess zero knowledge of the tuning search space.
2) `conv2dGenerator` does not know the internals of the tuning process (M/N/KPerBlock, M/NPerWave, KPACK, etc).

But, to determine whether padding kernel is used requires the knowledge of the tuning parameters possibly be used for a particular convolution configuration.

As a result, in the current form of the PR, since `conv2dGenerator` doesn't know the tuning search space (1 above), it has to be amended to possess part of 2. Therefore we have duplicated logic.

jerryyin: @whchung Actually inspired by what you are doing here. Wouldn't it be possible to derive the tuning parameter in constructing the conv op:
 - In constructing the conv op, derive tuning parameters before a conv op is constructed
   - If succeed, the constructed conv op would have tuning parameters, so there's no need to do it as a lowering step
   - If failed, we'd return a empty handle, allowing the failure to happen even earlier
 - It sounds like if all tuning parameters validation are brought earlier, we can (use the info to) decide what kind of conv op to construct

whchung: To remedy duplicated logic, I'd like create an oracle function to judge whether a padding kernel is needed, given the GemmM/N/K size, and the direction of the convolution.

The oracle function would possess the vector of initial tuning parameters, just like what's in use in `GridwiseGemmParams`.

On the other hand, I will NOT consolidate somewhat duplicated structures between generator and tuning modules: `Conv2dGenerator::Config` / `ConvContext` in this PR. Such refactoring looks completely beyond the scope of the this PR and is better handled as a separate ticket.

whchung: @jerryyin You are talking about eliminating `-miopen-affix-params` as an independent pass and run it inside `conv2dGenerator`? I'm not against the idea but it still seems to be beyond the scope of the PR. I'll log it in https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/439

krzysz00: I agree with @jerryyin's proposal, more or less (One interesting thing to consider is whether a convOp can hold the set of valid configurations - or if it could generate it - so that  tuning loops can be told what the valid parameters are for a particular configuration)

whchung:
Applying `skip-ci` label as quite a few refactoring commits are coming in.

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
whchung: Initial version of refactoring is done. `calculatePaddingKernelSize` is made as a function instead a lambda.

whchung:
Implemented the first version of the oracle function `calculatePaddingKernelSize`, the interface is not ideal and can be further amended. I don't quite understand the rationale of the parameter `isOriginalKernelSupport`. It seems to be one of the nuisance caused by backward data kernel.

Remove `skip-ci` and let the CI do some initial checks first. The PR is getting bigger and I'd like to verify it before moving further.

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
```diff
@@ -778,7 +778,7 @@ class PopulateParams : public PopulateParamsBase {
                 DerivedOutParams &gemmCDerivedParam, int64_t &gridSize);
 
   llvm::SmallVector<InitParamsNonXDL, 8>
-  getTuningParameters(ConvolutionContext &ctx) {
+  getTuningParameters(miopen::ConvOpType dir, mlir::Type dataType) {
```
ltqin, line 781: Two parameters are useless?

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -195,10 +195,20 @@ LogicalResult backwardWeightAtomicAdd(Conv2DBwdWeightOp op,
...
+
   // Get shape of filter tensor.
   auto filterType = op.filter().getType().template cast<MemRefType>();
   auto filterShape = filterType.getShape();
 
+  // Determine whether to use workspace.
+  bool hasWorkspace =
+      (filterType.getElementType() == b.getF16Type() && isXdlops);
+  if (hasWorkspace) {
+    assert(op.workspace());
```
ltqin: assert() just run in debug, is there right?

krzysz00:
@whchung I think you need to also edit `Jenkinsfile.downstream` to make CI pass

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
krzysz00: I think it runs always unless explicitly disabled. However, can we stick an error message on the assert? `assert(op.workspace() && "Op has no workspace")`

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -1172,56 +1148,32 @@ template <typename T> struct Conv2DRewritePattern : public OpRewritePattern<T> {
...
-            // gemmNExtra << "gemmKExtra: " << gemmKExtra << "\n";
-          }
-        };
-
     if (!isXdlops) {
       PopulateParams populateParams;
-      calculatePaddingKernelSize(populateParams);
+      std::tie(isOriginalKernelSupport, needExtraPad, gemmMExtra, gemmNExtra,
+               gemmKExtra) =
+          calculatePaddingKernelSize(gemmMSize, gemmNSize, gemmKSize,
```
krzysz00, line 1160: Thanks for pulling these out

krzysz00:
(Update, what you might need to do in Jenkinsfile.downstream is the same fix we have in the main Jenknisfile where you stop the MIOpen build system from building a copy of MLIR with `sed`)

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
whchung: Will do.

whchung:
Studying failures on the non-XDLOPS path.
http://ml-ci.amd.com:21096/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-512/20/pipeline/64

MIOpen `mlir-workspace` branch needs to be amended so workspace is only allocated when truly needed. The current implementation merely assumes as long as the following preconditions are met, a workspace is allocated and used:
- fp16
- XDLOPS
- backward weight

It forgot to take padding kernel into consideration. Need to understand when the invoker gets invoked by a solver, does the solver already understands that a workspace is not needed via the amended `Miir` interface.

krzysz00:
@whchung I don't think that's the problem - the failure is because the non-xdlops path compiles the `miopen-dialect` branch as a dependency instead of using the static library built during CI. I now have a PR that changes this so the non-xdlops path behaves like the xdlops one.

whchung:
> @whchung I don't think that's the problem - the failure is because the non-xdlops path compiles the `miopen-dialect` branch as a dependency instead of using the static library built during CI. I now have a PR that changes this so the non-xdlops path behaves like the xdlops one.

Thanks. Let me take a look at that PR. On the other hand, existing logic I put together in MIOpen `mlir-workspace` branch is indeed flawed, and I'm working on fixing it.

whchung:
Fixed MIOpen `mlir-workspace` branch. Give this PR another go.

PR #507, opened 2022-02-11
[DO NOT SQUASH] February 2022 upstream update

krzysz00:
This is our newest sync to upstream update. Compared to previous ones, this one seems to have been rather well-behaved and took less than a day to put together.

Notable changes include:
- **Watch out** `FunctionPass` is deleted in favor of `OperationPass<FuncOp>` This appears to have been a straightforward search-replace, both here and upstream. However, it does mean that the former `FunctionPass`es will be run on function declarations if any of them end up in scope. Since nothing broke I think we're fine, but I wanted to raise the backwards compatibility issue.
- `br` and friends have moved from `std` to a new `cf` (ControlFlow) dialect, changing the names of passes
- `select` moved to `arith`
- Various ext/trunc/cast ops lost their (loc, value, type) builders in favor of (loc, type, value)
- Due to an initial circular dependency issue, and later because it was cleaner that way, I moved `--mlir-to-rocdlir` (the translation) into our repo as `--gpu-module-to-rocdlir`
  - Future work may have an "extract LLVM in GPU" pass getting sent upstream to replace this
- The Tosa to Linalg conversion now has a TosaToLinalgNamed pass that needs to be run before it to handle osp like `tosa.conv2d -> linalg.conv2d`

PR #513, opened 2022-02-16
Remove overrideLoadStoreHack.

whchung:
The function was used in padding kernels. Now it seems with newer LLVM
the issues observed was no more.

No new E2E test case is added as existing cases would sufficiently
detect the issue.

### DIFF
--- a/mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
+++ b/mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
@@ -58,20 +58,6 @@ struct LowerMIOpenOpsStep5Pass
 constexpr int kTwoGB = 2147483647;
 
 //===----------------------------------------------------------------------===//
-// FIXME. XXX.
-// Force the use of affine maps over index maps in the presence of padding on
-// GEMM during threadwise load/store/copy when the gemm is padded due to bugs in
-// the index diff map implementation (or incompletenesses in it?)
-//===----------------------------------------------------------------------===//
-bool overrideLoadStoreHack(const PaddingInfoAttr paddingInfo, bool original) {
-  if (paddingInfo.getExtraM() > 0 || paddingInfo.getExtraK() > 0 ||
-      paddingInfo.getExtraN() > 0) {
-    return true;
-  }
-  return original;
-}
-
-//===----------------------------------------------------------------------===//
 // Utility function to compute sliceLengths for threadwise_copy and
 // threadwise_copy_v2 to determine the bounds of load/store loops.
 //===----------------------------------------------------------------------===//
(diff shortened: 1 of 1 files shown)


krzysz00 (approved):
Would it make sense to get rid of the non index diff path in threadwise_{load,store,copy} here as well, since this's the only way to set it off?

whchung:
The `legacy_load` / `legacy_store` attributes could still be set, so the logic to switched between index diff map/non-index diff map are still retained.

In https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/205 we can revisit this topic, which is already slated to be done in ROCm 5.2 timeframe.

krzysz00:
That all makes sense to me and I'll be happy sending in this change as-is

krzysz00:
Before merging, could you do me a favor and get a local performance comparison for some of the padding kernels?

whchung:
Performance report on a selected case which would be impacted by this PR:
```
./bin/MIOpenDriver.py -o xxx.csv conv -F 1 -f NCHW -n 20 -c 3 -H 32 -W 32 -k 6 -x 7 -y 7 -p 3 -q 3 -u 1 -v 1 -l 1 -j 1 -g 1
```

3 different metrics:

1) On `miopen-dialect` , after #507 gets in:
```
root@d115d375356c:~/llvm-project-mlir/build# cat nchw-miopen-dialect-fwd.csv
,Direction,DataType,XDLOPS,FilterLayout,InputLayout,OutputLayout,N,C,H,W,K,Y,X,DilationH,DilationW,StrideH,StrideW,PaddingH,PaddingW,TFlops
0,fwd,f32,True,kcyx,nchw,nkhw,20,3,32,32,6,7,7,1,1,1,1,3,3,0.602112
```

2) `remove-load-store-hack` branch (this PR), rebase with `miopen-dialect`.
```
root@d115d375356c:~/llvm-project-mlir/build# cat nchw-index-diff-map-rebased-fwd.csv
,Direction,DataType,XDLOPS,FilterLayout,InputLayout,OutputLayout,N,C,H,W,K,Y,X,DilationH,DilationW,StrideH,StrideW,PaddingH,PaddingW,TFlops
0,fwd,f32,True,kcyx,nchw,nkhw,20,3,32,32,6,7,7,1,1,1,1,3,3,0.6640941176470588
```

3) `remove-load-store-hack` branch (this PR) as is.
```
root@d115d375356c:~/llvm-project-mlir/build# cat nchw-index-diff-map-fwd.csv
,Direction,DataType,XDLOPS,FilterLayout,InputLayout,OutputLayout,N,C,H,W,K,Y,X,DilationH,DilationW,StrideH,StrideW,PaddingH,PaddingW,TFlops
0,fwd,f32,True,kcyx,nchw,nkhw,20,3,32,32,6,7,7,1,1,2,2,3,3,0.16362035544122175
```

In general, very good news. We can deduce that:
- LLVM update (#507) brings overall perf improvement.
- `remove-load-store-hack` (this PR) gets us another 10% boost on those padding kernels

PR #516, opened 2022-02-16
ROCm 5.0 changed the build name

krzysz00:


jerryyin (approved):
Approving, provided the local testing indicate this is the only thing needs to do.

krzysz00:
I've tagged up a local build of this Dockerfile and am making sure that there's not new packages we'd need to install or something so we only have one more of these PRs

krzysz00:
Good news, my container (which doesn't install any extra ROCm) runs our test suite successfully

PR #517, opened 2022-02-16
Move our CI over to ROCm 5.0

krzysz00:


krzysz00:
MIOpen build looks like it might need a separate rocblas-dev install, double-checking that now

jerryyin (approved):

PR #518, opened 2022-02-17
Abstract elementwise restore

pcf000:
Bring back the AbstractElementwise trait, because it indicates ops that are element-wise in the abstract sense, while the existing Elementwise trait indicates element-wise ops whose operands all have identical shapes.  Ie, AbstractElementwise allows for broadcasts, and that's what we wanted for partitioning.

That in itself fixes the tosa-partition.mlir test.  To fix tosa-partition-run.mlir, I removed the --linalg-detensorize pass.  There are enough changes to the bufferize passes that I guess that one only messes things up.

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOpBase.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOpBase.td
@@ -200,4 +200,9 @@ class Tosa_Op<string mnemonic, list<Trait> traits = []> :
     Op<Tosa_Dialect, mnemonic, !listconcat(traits, [TosaOpInterface])> {
 }
 
+//===----------------------------------------------------------------------===//
+// TOSA element-wise ops.  The Elementwise trait requires identical shapes.
+//===----------------------------------------------------------------------===//
+def AbstractElementwise : NativeOpTrait<"tosa::AbstractElementwise">;
+
 #endif // TOSA_OP_BASE
--- a/external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOpTraits.h
+++ b/external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOpTraits.h
@@ -1,0 +1,33 @@
+//===- TosaOpTraits.h - MLIR TOSA operation traits --------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file declares C++ classes for operation traits in the TOSA dialect.
+//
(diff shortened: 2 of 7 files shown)


sjw36 (approved):
Looks good. Is there a ticket to upstream?

krzysz00 (changes requested):
I'd like folks to first identify the change upstream and then to work out if it needs a revert or if we should change our design. For example, ElementWise might now imply broadcst semantics now

sjw36:
> I'd like folks to first identify the change upstream and then to work out if it needs a revert or if we should change our design. For example, ElementWise might now imply broadcst semantics now

It doesn't support broadcast.

krzysz00:
@sjw36 I don't think the TOSA ops are meant to have implicit broadcast semantics? Wouldn't it make more sense to handle that by adding explicit `tosa.expand_shape`?

More generally, having checked the history, it looks like AbstractElementwise is ours, and not something that ever existed upstream ... and that didn't have an upstreaming ticket for it.

Given that history, I'm withdrawing my current objections to letting it back in so that we don't break things.

 *However,* I don't like maintaining something like this, and I'd like you and @pcf000 to please go start a discussion on the MLIR Discourse about alternate approaches and how this should be handled.

krzysz00 (approved):

sjw36:
> @sjw36 I don't think the TOSA ops are meant to have implicit broadcast semantics? Wouldn't it make more sense to handle that by adding explicit `tosa.expand_shape`?
> 
It is explicitly stated in the source code that it supports broadcast and there are tests to make sure it does.

There should be a ticket to upstream in some fashion. @pcf000 please file one.

krzysz00:
@sjw36 `Elementwise` itself does mean broadcast. Its semantics are "this op that's defined over scalars can be expanded to vectors or tensors by going elementwise"

krzysz00:
And so I don't understand why we need this new trait, and I'd like to muse sure we've currently understood everything. Filing a revision onto upstream might get y'all that clarity, and so that might be the right go

sjw36:
> @sjw36 `Elementwise` itself does mean broadcast. Its semantics are "this op that's defined over scalars can be expanded to vectors or tensors by going elementwise"

I would like Elementwise to mean that and maybe that can be the solution but right now they explicitly require the shapes to be the same.

pcf000:
The element-wise trait has evolved.  When I first started, it meant a combination of several things.  Later, it was exploded into four separate traits.  I will go back to the discussions around the evolution and see what that tells me.

pcf000:
A plan, now that I've been thinking about it:

1. Merge this guy, once it passes, to unblock Simon.
2. Make a ticket about making an upstreamable version.
     a. Could be this code.
     b. Could be a proposal to separate element-wise from same-shape.
     c. Could be we do something ourselves to give operands the same shape.

Will require studying the history of the trait and probably discussing the purposes and spec.

krzysz00:
@pcf000 That's just about what I was thinking, though I think the starting point for step 2 would be for Simon and you to go to [the MLIR mailing list-equivalent](https://discourse.llvm.org/c/mlir/31) and write up the question there, including context on the partitioning pass.

PR #521, opened 2022-02-17
Support int8 in the test generator miopen-gen

yiqian1:
Generate int8 E2E test cases.

### DIFF
--- a/mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
+++ b/mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
@@ -158,7 +158,7 @@ LogicalResult Conv2dGenerator::hasValidDimension() const {
                                                   {"fp16", 2},
                                                   {"f16", 2},
                                                   {"bf16", sizeof(uint16_t)},
-                                                  {"i8", 1}};
+                                                  {"i8", sizeof(int8_t)}};
 
   auto checkDimSizes = [](const SmallVector<int64_t, 5> &dims) -> bool {
     return std::all_of(dims.begin(), dims.end(),
--- a/mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
+++ b/mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
@@ -157,7 +157,8 @@ LogicalResult Conv2dGenerator::hasValidDimension() const {
                                                   {"fp32", sizeof(float)},
                                                   {"fp16", 2},
                                                   {"f16", 2},
-                                                  {"bf16", sizeof(uint16_t)}};
+                                                  {"bf16", sizeof(uint16_t)},
+                                                  {"i8", 1}};
 
   auto checkDimSizes = [](const SmallVector<int64_t, 5> &dims) -> bool {
     return std::all_of(dims.begin(), dims.end(),
@@ -315,6 +316,8 @@ Type Conv2dGenerator::getDataType(OpBuilder &builder) const {
     dataType = builder.getF16Type();
   } else if (config.dataTypeStr == "bf16") {
     dataType = builder.getBF16Type();
+  } else if (config.dataTypeStr == "i8") {
+    dataType = builder.getI8Type();
   }
   return dataType;
 }
@@ -574,6 +577,10 @@ LogicalResult Conv2dGenerator::genConvModule(ModuleOp &module, int kernel_id,
     return failure();
   }
 
+  Type outputDataType = dataType;
+  if (dataType.isInteger(8)) {
+    outputDataType = builder.getIntegerType(32);
+  }
   // Construct a new FuncOp.
   auto filterArgType =
       MemRefType::get(ArrayRef<int64_t>(config.filterDimension.begin(),
--- a/mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
+++ b/mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
@@ -153,12 +153,10 @@ LogicalResult Conv2dGenerator::hasValidDimension() const {
     return failure();
   }
 
-  static const llvm::StringMap<size_t> typeWidths{{"f32", sizeof(float)},
-                                                  {"fp32", sizeof(float)},
-                                                  {"fp16", 2},
-                                                  {"f16", 2},
-                                                  {"bf16", sizeof(uint16_t)},
-                                                  {"i8", sizeof(int8_t)}};
+  static const llvm::StringMap<size_t> typeWidths{
+      {"f32", sizeof(float)},     {"fp32", sizeof(float)},
+      {"fp16", sizeof(uint16_t)}, {"f16", sizeof(uint16_t)},
+      {"bf16", sizeof(uint16_t)}, {"i8", sizeof(int8_t)}};
 
   auto checkDimSizes = [](const SmallVector<int64_t, 5> &dims) -> bool {
     return std::all_of(dims.begin(), dims.end(),
@@ -593,7 +591,7 @@ LogicalResult Conv2dGenerator::genConvModule(ModuleOp &module, int kernel_id,
   auto outputArgType =
       MemRefType::get(ArrayRef<int64_t>(config.outputDimension.begin(),
                                         config.outputDimension.end()),
-                      dataType);
+                      outputDataType);
 
   bool hasWorkspace = this->hasWorkspace(builder);
   Type workspaceArgType;
--- a/mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
+++ b/mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
@@ -518,6 +518,10 @@ extern "C" void mcpuPrintF32(float f1, float f2) {
   printf("Values: %f, %f\n", f1, f2);
 }
 
+extern "C" void mcpuPrintInt32(int32_t d1, int32_t d2) {
+  printf("Values: %d, %d\n", d1, d2);
+}
+
 // 2D float memref utility routines.
 
 extern "C" void mcpuMemset2DFloat(float *allocated, float *aligned,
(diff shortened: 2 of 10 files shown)


krzysz00:
This is a note that CI is passing and that Thursday the 17th includes merges of unrelated fix PRs. Therefore, on approval, we can cheat the CI (manually kick off a build that unchecks sharedLib and staticLib)

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
```diff
@@ -156,7 +156,8 @@ LogicalResult Conv2dGenerator::hasValidDimension() const {
                                                   {"fp32", sizeof(float)},
                                                   {"fp16", 2},
                                                   {"f16", 2},
-                                                  {"bf16", sizeof(uint16_t)}};
+                                                  {"bf16", sizeof(uint16_t)},
+                                                  {"i8", 1}};
```
sjw36: should use sizeof for all these even tho it is obvious

yiqian1: Should we use sizeof(uint16_t) for f16?

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
```diff
@@ -314,6 +329,129 @@ extern "C" void mgpuMemCopyInt32(int32_t *sourceAllocated,
             static_cast<hipMemcpyKind>(copyDirection));
 }
 
+extern "C" StridedMemRefType<int8_t, 5>
+mgpuMemAlloc5DInt8(int8_t *allocated, int8_t *aligned, int64_t offset,
+                   int64_t size0, int64_t size1, int64_t size2, int64_t size3,
+                   int64_t size4, int64_t stride0, int64_t stride1,
+                   int64_t stride2, int64_t stride3, int64_t stride4) {
```
sjw36, line 336: This is fine for now but we should probably just have a single alloc func and insert casts from it.

sjw36: @yiqian1 can you file a ticket for that. thx

sjw36 (approved):
Looks good. Maybe some follow-up cleanup tickets.

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
krzysz00: I think y'all might want to look at what's been done upstream here as well - last I checked they've got a lot of these allocation functions templated in interesting ways.

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
```diff
@@ -507,6 +510,10 @@ LogicalResult Conv2dGenerator::genConvModule(ModuleOp &module, int kernel_id,
     return failure();
   }
 
+  Type outputDataType = dataType;
+  if (dataType.isInteger(8)) {
+    outputDataType = builder.getIntegerType(32);
+  }
```
jerryyin, line 581: Irrelevant to where I put the review...

I'd prefer a small end-to-end test (one of all 1, another of all 0s, for example) that exercise i8 convolution and get expected result by just using `miopen-gen`.

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -1356,10 +1380,13 @@ createVerifierFunc(ModuleOp &module, const KernelIF &kernel,
 
   // Lower cpu values to gpu precision
   mlir::FloatType elemFType = floatType;
-  if (!elemType.isF32()) {
+  if (elemType.isIntOrIndex()) { // i8, i32
+    cpuLoadVal = loopB.create<arith::FPToSIOp>(loc, elemType, cpuLoadVal);
+    gpuFPVal = loopB.create<arith::SIToFPOp>(loc, floatType, gpuLoadVal);
```
krzysz00: Wait a minute, I don't think this is it?
Shouldn't we verify int8 convolution exactly?

krzysz00: ... which we are doing below, so now the question is - where'd the floating point come from? Shouldn't we add an integer convolution verifier?

jerryyin (changes requested):
Implementation looks good. I'm putting a blocker for test case. Need at least a couple of small scale e2e test case to make sure i8 testing does what we expect it to do.

mlir/tools/miopen-gen/miopen-gen.cpp
yiqian1: We're verifying int8 convolution. The int8 results are casted to floating point only for printing.

krzysz00: I see `cpuLoadVal` being passed to in integer comparison further down the line though?

krzysz00: Ok, so, I'm confused. If we're verifying an int8 convolution, why is the CPU verifier returning floats?

yiqian1: Oh, I see what you were asking. The CPU implementation is floating point. The result is casted to integer to compare with the gpu convolution. This should be equivalent to an integer convolution. I feel there is no need to do a cpu integer convolution.

krzysz00: Having looked, no, no it's not for output values above 2^24 , per https://stackoverflow.com/questions/3793838/which-is-the-first-integer-that-an-ieee-754-float-is-incapable-of-representing-e

yiqian1: You're right. I thought the floating point was safe because I was thinking the output was also int8.

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
```diff
@@ -1164,15 +1152,59 @@ getSizesAndStrides(int64_t rank1, StridedMemRefType<float, 5> *filter,
...
+
+  auto outputSizes = llvm::ArrayRef<int64_t>(output->sizes, rank3);
+  auto outputStrides = llvm::ArrayRef<int64_t>(output->strides, rank3);
+
+  getSizesAndStrides(filterSizes, filterStrides, inputSizes, inputStrides,
+                     outputSizes, outputStrides, f_layout, i_layout, o_layout,
+                     fSizes, fStrides, iSizes, iStrides, oSizes, oStrides);
+}
+
+static void getSizesAndStridesInt8(
```
jerryyin: `getSizesAndStridesInt8()` and `getSizesAndStridesFloat()` are basically copy except argument type therefore needs to be templatized

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
```diff
@@ -1363,6 +1394,193 @@ mcpuConv2dBwdData(int64_t rank1, void *f_ptr, int64_t rank2, void *i_ptr,
           }
 }
 
+extern "C" void
+mcpuConv2dInt8(int64_t rank1, void *f_ptr, int64_t rank2, void *i_ptr,
```
jerryyin, line 1404: `mcpuConv2dInt8()` and `mcpuConv2d()` can be templatized on accumulator's type.

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
```diff
@@ -1363,6 +1394,193 @@ mcpuConv2dBwdData(int64_t rank1, void *f_ptr, int64_t rank2, void *i_ptr,
...
+                                                   fil_w * filterStrides[4]]);
+                }
+
+            outputAllocated[g * outputStrides[0] + n * outputStrides[1] +
+                            k * outputStrides[2] + out_h * outputStrides[3] +
+                            out_w * outputStrides[4]] = acc;
+          }
+}
+
+extern "C" void mcpuConv2dBwdWeightInt8(
```
jerryyin: I don't think this is necessary as we only support forward conv for i8

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
```diff
@@ -1363,6 +1394,193 @@ mcpuConv2dBwdData(int64_t rank1, void *f_ptr, int64_t rank2, void *i_ptr,
...
+                                                     out_w * outputStrides[4]]);
+                }
+            filterAllocated[g * filterStrides[0] + k * filterStrides[1] +
+                            c * filterStrides[2] + y * filterStrides[3] +
+                            x * filterStrides[4]] = (int8_t)acc;
+          }
+}
+
+extern "C" void
+mcpuConv2dBwdDataInt8(int64_t rank1, void *f_ptr, int64_t rank2, void *i_ptr,
```
jerryyin: I don't think this is necessary as we only support forward conv for i8

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -1462,7 +1480,8 @@ createVerifierFunc(ModuleOp &module, const KernelIF &kernel,
   } else {
     cmpVal = loopB.create<arith::CmpIOp>(loc, arith::CmpIPredicate::ne,
                                          cpuLoadVal, gpuLoadVal);
-    percentDiffVal = loopB.create<arith::SubIOp>(loc, cpuLoadVal, gpuLoadVal);
+    //    percentDiffVal = loopB.create<arith::SubIOp>(loc, cpuLoadVal,
+    //    gpuLoadVal);
```
jerryyin: Remove those?

mlir/tools/miopen-gen/miopen-gen.cpp
jerryyin: Is this addressed? I still see int to float conversion?

jerryyin (approved):
LGTM

yiqian1:
The old code converted integer results to floating point for printing. The new code directly prints integer results.

PR #522, opened 2022-02-17
Simplify XDlopsGemmV2 to use buffer offsets instead of transforms

krzysz00:
The only reason XdlopsGemmV2 participates in the coordinate
transformation machinery is so that it can learn the offset at which
the LDS buffers for A and B start. XdlopsGemmV2 needs these offsets so
that we can allocate one LDS buffer instead of two (note to reviewer:
why?).

It's much simpler to pass in these offsets directly. This also allows
us to improve the quality of the generated code by moving certain
computations outside of the load/store loops.

The effect of this PR on the generated ISA is somewhere between "minor reorderings" and "nothing" and so I don't expect it to have a performance impact. However, it makes it clearer what's going on, prevents people from passing some weird multi-dimensional view of the buffer to `xdlops_gemm_v2`, and I think it looks nicer this way so ... let's?


whchung:
In the original C++ implementation, LDS is allocated as a flat C array and pointer arithmetic are used to partition it into 2 pieces for Matrix A / B. These pieces are then further transformed into multi-dimensional arrays to facilitate semantics asked by threadwise_copy.

That's the reason why in the MLIR implementation, we had `miopen.subview` in the first place. The op is now removed.

Back to the original C++ implementation, when LDS is later needed for preparing data for XDLOPS MFMA instructions, they are accessed as a linear array, without needing the coordinate transforms used for threadwise_copy.

So in general, this PR does make sense and provides a simpler interface to users of `xdlops_gemm_v2`.

krzysz00:
( @whchung ) I've skip-ci'd this until it's approved so we don't burn resources, especially since there's a few small CI-related PRs earlier in my personal pipeline.

krzysz00:
Handled elsewhere, closing

PR #523, opened 2022-02-18
Add NCHW support for convolution from MIGraphX 

jungpark-mlir:
- Add support for C-API pipeline
- Add support for NCHW convolution from MIGraphX : insert transpose
- Add support for NCHW convolution in tosa : fold transpose

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/Pipeline.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Pipeline.h
@@ -31,7 +31,7 @@ void addPipeline(PassManager &pm, bool applicability = false,
 // Compilation pipeline from LLVM to Binary
 void addBackendPipeline(PassManager &pm, const std::string &triple,
                         const std::string &chip, const std::string &features,
-                        int32_t optLevel = 3);
+                        int32_t optLevel = 3, int32_t indexBitWidth = 32);
 
 } // namespace miopen
 } // namespace mlir
--- a/mlir/lib/CAPI/Dialect/MIGraphX.cpp
+++ b/mlir/lib/CAPI/Dialect/MIGraphX.cpp
@@ -12,7 +12,7 @@
 #include "mlir/Dialect/MIGraphX/MIGraphXOps.h"
 #include "mlir/Dialect/MIGraphX/Pipeline.h"
 #include "mlir/Dialect/MIOpen/Pipeline.h"
-#include "mlir/Dialect/MIOpen/MIOpenOps.h"
+#include "mlir/Dialect/MIOpen/MIOpen.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
 #include "mlir/Dialect/GPU/GPUDialect.h"
 #include "mlir/Dialect/GPU/Passes.h"
--- a/mlir/lib/CAPI/Dialect/MIGraphX.cpp
+++ b/mlir/lib/CAPI/Dialect/MIGraphX.cpp
@@ -123,8 +123,8 @@ void mlirMIGraphXAddHighLevelPipeline(MlirPassManager pm) {
 
 MLIR_CAPI_EXPORTED void mlirMIGraphXAddBackendPipeline(MlirPassManager pm,
                                                        const char *chip,
-                                                       const char *triple = "amdgcn-amd-amdhsa",
-                                                       const char *features = "") {
+                                                       const char *triple,
+                                                       const char *features) {
   mlir::registerGpuSerializeToHsacoPass();
   auto passMan = unwrap(pm);
   passMan->setNesting(mlir::PassManager::Nesting::Implicit);
--- a/mlir/lib/CAPI/Dialect/MIGraphX.cpp
+++ b/mlir/lib/CAPI/Dialect/MIGraphX.cpp
@@ -24,61 +24,56 @@
 MLIR_DEFINE_CAPI_DIALECT_REGISTRATION(MIGraphX, migraphx,
                                       mlir::migraphx::MIGraphXDialect)
 
-// Returns the number of operands in the FuncOp and fill information in the
-// passed ptr.
+// Returns the required buffer size if called with null buffer
+// and fill information in the passed ptr when provided.
 MLIR_CAPI_EXPORTED
-void mlirGetKernelInfo(MlirModule module, void *data) {
+void mlirGetKernelInfo(MlirModule module, int *size, void *data) {
   auto mod = unwrap(module);
-  std::vector<int> info;
   int argNum = 0;
   int argIdx = 0;
   llvm::StringRef kernelName;
-  mod.walk([&](mlir::FuncOp f) {
-    auto args = f.getArguments();
-    for (auto arg : args) {
-      argNum++;
-      auto sType = arg.getType().template cast<mlir::ShapedType>();
-      auto rank = sType.getRank();
-      info.push_back(rank);
-      for (int i = 0; i < rank; i++)
-        info.push_back(sType.getDimSize(i));
-      argIdx += rank;
+  if (data == nullptr && size != nullptr) {
+    mod.walk([&](mlir::FuncOp f) {
+      auto args = f.getArguments();
+      for (auto arg : args) {
+        argNum++;
+        auto sType = arg.getType().template cast<mlir::ShapedType>();
+        auto rank = sType.getRank();
+        argIdx += rank;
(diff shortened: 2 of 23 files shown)


mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
```diff
@@ -229,10 +267,161 @@ class ConvConverter final : public OpConversionPattern<tosa::Conv2DOp> {
   }
 };
 
+class TransposeConverter final : public OpConversionPattern<tosa::TransposeOp> {
```
sjw36, line 270: I think the approach should be to backtrack the inputs and outputs of the tosa::Conv2DOp to see if it is written by a transpose. Then attach the transpose input and fix the layout. This would make the TransposeOp dead-code.

sjw36: We discussed offline and this will be sufficient for now. We will generalize transpose folding and fusion in a later PR. @jungpark-mlir please create a ticket and link to this PR.

mlir/include/mlir/Dialect/MIOpen/Pipeline.h
```diff
@@ -31,7 +31,7 @@ void addPipeline(PassManager &pm, bool applicability = false,
 // Compilation pipeline from LLVM to Binary
 void addBackendPipeline(PassManager &pm, const std::string &triple,
                         const std::string &chip, const std::string &features,
-                        int32_t optLevel = 3);
+                        int32_t optLevel = 3, int32_t indexBitWidth = 32);
```
sjw36, line 34: Is this still necessary? Can MIGraphX be improved to support 32-bit index type? 
It is a performance hit with 64-bit.

mlir/lib/CAPI/Dialect/MIGraphX.cpp
```diff
@@ -7,8 +8,134 @@
...
+#include "llvm/Support/TargetSelect.h"
+#include <vector>
 
 MLIR_DEFINE_CAPI_DIALECT_REGISTRATION(MIGraphX, migraphx,
                                       mlir::migraphx::MIGraphXDialect)
+
+// Returns the number of operands in the FuncOp and fill information in the
+// passed ptr.
+MLIR_CAPI_EXPORTED
+void mlirGetKernelInfo(MlirModule module, void *data) {
```
sjw36: See miirGetBuffer API.  Something like this:

void mlirGetKernelInfo(MlirModule module, void *data, int32_t *size);

1. Call once with nullptr for data and an int32_t variable referenced by size. Size is computed and returned in size ptr.
2. Allocate data buffer for returned size.
3. Call again with data and size ptrs.

mlir/lib/CAPI/Dialect/MIGraphX.cpp
```diff
@@ -7,8 +8,134 @@
...
+  for (int i = 0; i < size; i++)
+    argData[i + 1] = info[i];
+  char *nameData = (char *)(argData + size + 1);
+  for (int i = 0; i < kernelName.size(); i++) {
+    nameData[i] = kernelName[i];
+  }
+}
+
+MLIR_CAPI_EXPORTED
+int mlirGetKernelInfoSize(MlirModule module) {
```
sjw36: This is a good alternative but can get out of sync with previous. Combine under the hood or as stated with previous comment.

mlir/lib/CAPI/Dialect/MIGraphX.cpp
```diff
@@ -7,8 +8,134 @@
...
+      argIdx += rank;
+      kernelName = f.getName();
+    }
+  });
+  int size = (1 + argNum + argIdx) * sizeof(int) + kernelName.size();
+  return size;
+}
+
+// Returns block_size and grid_size as int[2]
+MLIR_CAPI_EXPORTED void mlirGetKernelAttrs(MlirModule module, int *attrs) {
```
sjw36: Use int32_t or int64_t, vs int.

mlir/lib/CAPI/Dialect/MIGraphX.cpp
```diff
@@ -7,8 +8,134 @@
...
+  auto mod = unwrap(module);
+  mod.walk([&](mlir::LLVM::LLVMFuncOp llvmFunc) {
+    attrs[0] =
+        llvmFunc->getAttrOfType<mlir::IntegerAttr>("block_size").getInt();
+    attrs[1] = llvmFunc->getAttrOfType<mlir::IntegerAttr>("grid_size").getInt();
+  });
+}
+
+// Returns the size of compiled binary
+MLIR_CAPI_EXPORTED int mlirGetBinarySize(MlirModule module) {
```
sjw36: Similar comment as mlirGetKernelInfo comments.

mlir/lib/CAPI/Dialect/MIGraphX.cpp
```diff
@@ -7,8 +8,134 @@
...
+MLIR_CAPI_EXPORTED
+void mlirMIGraphXAddHighLevelPipeline(MlirPassManager pm) {
+  auto passMan = unwrap(pm);
+  passMan->setNesting(mlir::PassManager::Nesting::Implicit);
+  mlir::migraphx::addHighLevelPipeline(*passMan);
+  mlir::miopen::addHighLevelPipeline(*passMan);
+}
+
+MLIR_CAPI_EXPORTED void mlirMIGraphXAddBackendPipeline(MlirPassManager pm,
+                                                       const char *chip) {
```
sjw36: we will probably want the triple passed in as well.

sjw36: and features.

mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
```diff
@@ -63,20 +81,30 @@ class ConvConverter final : public OpConversionPattern<tosa::Conv2DOp> {
       return Value();
     }
     auto shape = oprType.getShape();
-    SmallVector<int64_t, 5> expShape(shape.begin(), shape.end());
-    expShape.push_back(1);
-    auto newType = MemRefType::get(expShape, oprType.getElementType());
+    SmallVector<int64_t, 5> expShape;
```
sjw36, line 84: I made some similar updates to this routine in PR https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/527 but I will wait for your PR and merge then.

sjw36 (approved):
This approach is good for now but we will generalize support for transpose folding and fusion. Just some minor cleanup.

mlir/include/mlir/Dialect/MIOpen/Pipeline.h
jungpark-mlir: This includes passing strides which might be possible to overflow, so we actually need 64bit.
I would properly investigate the reason for the performance degradation if using 64bit causes any significant issue, it shouldn't that bad.

mlir/lib/CAPI/Dialect/MIGraphX.cpp
```diff
@@ -7,8 +8,126 @@
...
+
+// Returns the required buffer size if called with null buffer
+// and fill information in the passed ptr when provided.
+MLIR_CAPI_EXPORTED
+void mlirGetKernelInfo(MlirModule module, int *size, void *data) {
+  auto mod = unwrap(module);
+  int argNum = 0;
+  int argIdx = 0;
+  llvm::StringRef kernelName;
+  if (data == nullptr && size != nullptr) {
```
sjw36, line 51: Better to share the collection of info and just conditionally assign size and data.

mlir/lib/CAPI/Dialect/MIGraphX.cpp
```diff
@@ -7,8 +8,126 @@
...
+    attrs[1] = llvmFunc->getAttrOfType<mlir::IntegerAttr>("grid_size").getInt();
+  });
+}
+
+// Returns the size of compiled binary if called with null ptr
+// and return the compiled binary when buffer is provided
+MLIR_CAPI_EXPORTED bool mlirGetBinary(MlirModule module, int *size, char *bin) {
+  bool success = false;
+  auto mod = unwrap(module);
+  if (bin == nullptr && size != nullptr) {
```
sjw36: see comment above.

sjw36 (approved):

PR #524, opened 2022-02-18
Adding i32 support to raw buffer store intrinsics

jerryyin:
This is the last op missing for i8 threadwise copy result: Since xdlops populate i32 vector, its lowered result requires raw buffer store on i32 vector types. I added i32 type lowering on raw buffer store from gpu->rocdl->llvm. The existing implementation is robust enough to handle the additional type. I updated tablegen and unit test to make sure this new type will continue be tracked.

With this change I'm able to do end-to-end lowering with:
```bash
./bin/miopen-opt -miopen-affix-params -miopen-lowering -miopen-lowering-step{2,3,4,5} -convert-miopen-to-gpu -convert-gpu-to-rocdl ./conv_op.mlir &> rocdl_wrong.mlir | ./bin/miopen-translate -gpu-module-to-rocdlir |  /opt/rocm/llvm/bin/opt -passes='default<O3>,strip' -S | /opt/rocm/llvm/bin/llc -mcpu=gfx908
```

Next steps:
 - Exercise more configs with different problem sizes
 - Exercise more tuning parameters
 - Add and fix end-to-end test when infrastructure from @yiqian1 is ready

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/GPU/GPUOps.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/GPU/GPUOps.td
@@ -1308,9 +1308,10 @@ def GPU_MubufStoreOp :
 // buffer store
 def GPU_RawbufStoreOp :
     GPU_Op<"raw_buffer_store">,
-    Arguments<(ins AnyTypeOf<[BF16, F16, F32,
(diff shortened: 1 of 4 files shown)


whchung (approved):
LGTM.

changes to ROCDL dialect and its tests shall be upstreamed.

PR #525, opened 2022-02-18
Update public CI to match changes to private CI

krzysz00:
This'll unblock merging PR #512 and improve CI build times.

(It'll also allow us to "cheat" CI if there's a passing run that gets
obliterated by an unrelated merge - I trust everyone not to abuse this)

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile.downstream
+++ b/mlir/utils/jenkins/Jenkinsfile.downstream
@@ -55,7 +55,7 @@ pipeline {
                         docker {
                             image dockerImage()
                             args dockerArgs()
-                            label "gfx906"
+                            label "rocm"
                             alwaysPull true
                         }
                     }
--- a/mlir/utils/jenkins/Jenkinsfile.downstream
+++ b/mlir/utils/jenkins/Jenkinsfile.downstream
@@ -20,14 +20,6 @@ void buildMIOpen(String cmakeOpts) {
     sh 'cd build; make -j $(nproc) MIOpenDriver'
 }
 
-void getAndBuildMIOpen(String prefixOpt, String cmakeOpts) {
-    git branch: 'develop', poll: false,\
-        url: 'https://github.com/ROCmSoftwarePlatform/MIOpen.git'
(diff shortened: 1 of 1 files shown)


mlir/utils/jenkins/Jenkinsfile.downstream
```diff
@@ -45,23 +37,33 @@ String dockerImage() {
...
+                    when {
+                        beforeAgent true;
+                        equals expected: true, actual: params.sharedLib;
+                    }
                     agent {
                         docker {
                             image dockerImage()
                             args dockerArgs()
-                            label "rocm"
+                            label "gfx906"
```
jerryyin: I think only static library does tuning and needs to run on gfx906 only. We can still use gfx900 on shared library tests.

mlir/utils/jenkins/Jenkinsfile.downstream
```diff
@@ -98,16 +104,21 @@ pipeline {
...
                                 installation: 'InSearchPath', workingDir: 'build'
                             }
                         }
                         stage("Build MIOpen with libMLIRMIOpen") {
                             steps {
                                 dir('MIOpen') {
-                                getAndBuildMIOpen('--prefix ./MIOpenDeps', '''-DMIOPEN_USE_MLIR=On
+                                git branch: 'develop', poll: false,\
+                                    url: 'https://github.com/ROCmSoftwarePlatform/MIOpen.git'
+                                sh 'sed -i "/ROCmSoftwarePlatform\\/llvm-project-mlir/d" ./requirements.txt'
```
jerryyin, line 116: Nice catch, after all those times we are testing an old branch. However, it does seem to be a reminder to extract common logic between private/public CI into a common place.

mlir/utils/jenkins/Jenkinsfile.downstream
krzysz00: Ah, I didn't know `rocm` implied a GPU. I'll roll this change back

jerryyin (approved):
LGTM

PR #526, opened 2022-02-21
Run performance tests on gfx906 as user perfDB was created on gfx906

yiqian1:


krzysz00 (approved):
Good catch, approved!

PR #527, opened 2022-02-23
MLIR#450: added conversion from tosa.matmul to miopen.conv2d

sjw36:
* using nhwc and 1x1 filter

https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/450

### DIFF
--- a/mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
+++ b/mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
@@ -215,31 +215,15 @@ public:
     auto bias_mr = operands[2];
     auto resultType = op.getType();
 
-    bool bNCHW = checkNCHW(op);
-    int iExpand = 4, fExpand = 4, oExpand = 4; // kyxc[g], nhwc[g], nhwk[g]
-    if (bNCHW) {
-      iExpand = 1; // k[g]cyx
-      fExpand = 0; // [g]nchw
-      oExpand = 1; // n[g]khw
-    }
-
-    // expand tensors from rank 4 (NHWC) to rank 5 (NHWCG)
-    auto inputExpanded = expandMemRef(op, input_t, iExpand, rw);
-    auto filterExpanded = expandMemRef(op, filter_t, fExpand, rw);
-
     auto outputType =
         getTypeConverter()->convertType(resultType).cast<MemRefType>();
     Value output = rw.create<memref::AllocOp>(loc, outputType);
 
-    auto outputExpanded = expandMemRef(op, output, oExpand, rw);
+    bool isNCHW = checkNCHW(op);
 
-    auto outputType =
-        getTypeConverter()->convertType(resultType).cast<MemRefType>();
-    Value output = rw.create<memref::AllocOp>(loc, outputType);
-
-    const char *filterLayout = "kgyxc";
-    const char *inputLayout = "gnhwc";
-    const char *outputLayout = "nghwk";
+    const char *filterLayout = isNCHW ? "kcyxg" : "kyxcg";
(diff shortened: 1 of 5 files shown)


sjw36:
@jungpark-mlir I know this conflicts with your PR but I will not merge until yours is in. Just want to get feedback started.

mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
```diff
@@ -215,15 +233,52 @@ class ConvConverter final : public OpConversionPattern<tosa::Conv2DOp> {
...
-      output = rewriter.create<tosa::AddOp>(loc, op.getType(),
-                                            ValueRange{conv_output_t, bias_t});
+      auto bias_t = rw.create<bufferization::ToTensorOp>(loc, bias_expand_mr);
+      output = rw.create<tosa::AddOp>(loc, op.getType(),
+                                      ValueRange{conv_output_t, bias_t});
+    }
+    return success();
+  }
+};
+
```
jungpark-mlir, line 265: We can do this with 1) tosa conversion on tensor,
tosa.matmul -> tosa.conv2d.NCHW
:matA(1, m, k) * matB(1, k, n) 
-> conv2d :  
input (matB - N:1, C:k, H:n, W:1), 
filter (matA - N:m, C:k, H:1, W:1)
-> output (matC - N:m, C:1, H:n, W:1)

2) followed by existing tosa.conv to miopen.conv with bufferization (with no change)

jungpark-mlir: We may want to have miopen.matmul because of the batch support, which might be possible via using group as you mentioned. But group dimension goes into the C dim and requires transpose at the end. Maybe not if miopen.conv support ANY layout.

sjw36: NHWC is a better layout since the inner dimension corresponds to the K dimension of GEMM and we can use coalesce loads. But would be good to transpose input to also NHWC.

When we get general transpose support this will be a good optimization.

jungpark-mlir: I think, coalesced load can be enabled only when the original data layout is in MxK and NxK, i.e. matB is transposed. Otherwise HW can't coalesce the vertical vector load.
I just prefer to split compilation on tensors and compilation on buffers but this is still fine.
However in the future, I think we will need low level operation miopen.matmul with transpose option to each matrix, like BLAS GEMM API has (+more layout if needed) .

whchung:
@sjw36 wondering does this PR run end-to-end? if so could we get the initial perf of:
- M=N=K=8K
- M=N=8K, K=16384
on MI-100?

sjw36:
> @sjw36 wondering does this PR run end-to-end? if so could we get the initial perf of:
> 
> * M=N=K=8K
> * M=N=8K, K=16384
>   on MI-100?

It does run e2e so we should be able to easily get those numbers.

jungpark-mlir (approved):

mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
krzysz00: Yeah, allowing for transpose and adding a `miopen.matmul` is a good call

One thing I'll note is that NCHW is faster than NHWC from what I've seen

sjw36: @krzysz00 is that true for 1x1. It is already a matrix.

krzysz00: I'll double-check perf results later

PR #528, opened 2022-02-25
Explicitly turn off MLIR solvers for performance comparison

yiqian1:


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -376,7 +376,7 @@ pipeline {
                            steps {
                                dir('MIOpen') {
                                    getAndBuildMIOpen("--prefix ${WORKSPACE}/MIOpen/MIOpenDeps",
-                                                     '''-DMIOPEN_BACKEND=HIP
+                                                     '''-DMIOPEN_BACKEND=HIP -DMIOPEN_USE_MLIR=Off
```
krzysz00, line 379: I think cmake booleans are usually spelled in all caps, so `OFF` here is good style - or, in the alternative, `0`

krzysz00 (approved):
LGTM mod one style nit that doesn't strictly need to be resolved

whchung:
@yiqian1 Can we add one more column in the performance report to show: MIOpen (with tuned MLIR kernel) / MIOpen (without MLIR)?

yiqian1:
> @yiqian1 Can we add one more column in the performance report to show: MIOpen (with tuned MLIR kernel) / MIOpen (without MLIR)?

Yes. Will do.

PR #529, opened 2022-02-25
Disable ROCm runner build when we build a static library.

whchung:
Supply -DMLIR_ENABLE_ROCM_RUNNER=0 when we build libMLIRMIOpen.
This will eliminate the dependency to a working ROCm installation with a
working ROCm-compatible GPU installed.

This change is recommended by @jerryyin reviewing an MIOpen PR: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1443

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -251,7 +251,7 @@ pipeline {
                             steps {
                                 checkout scm
                                 buildProject('libMLIRMIOpen',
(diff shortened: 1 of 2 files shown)


whchung:
@jerryyin Please help advise if this PR is relevant given: https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/miopen-dialect/CMakeLists.txt#L77-L81

whchung:
Close this PR before I understand the real problem better. It's very nebular to me right now.

PR #531, opened 2022-02-28
[DO NOT SQUASH] Remove static library's build-time dependency on ROCm

krzysz00:
Fixes
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/454

This combines with the external-part commit following to fix this
issue.

This commit also removes redundant copies of code from
external/llvm-project/mlir/CMakeLists.txt and renames all our uses of
MLIR_ROCM_RUNNER_ENABLED with MLIR_ENABLE_ROCM_RUNNER to match
upstream.

---

[[MLIR]AMDGPU] Use correct flags when building SerializeToHsaco](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/commit/36d568a5f00a59c5d04ba5ec8ae83cc21bae58ef)

The SerializeToHsaco pass does not depend on ROCm being available on
the build system - it only requires ROCm to be present at runtime.
However, the CMake file that built it tested for
MLIR_ENABLE_ROCM_RUNNER , which implies that ROCm is currently
available and is used to control building ROCm integration tests.

Referencing MLIR_ENABLE_ROCM_RUNNER instead of
MLIR_ENABLE_ROCM_CONVERSIONS in the SerializeToHsaco build therefore
causes problems for clients who wish to build projects that depend on
this pass on a system without an AMD GPU present.


CMakeLists.txt
```diff
@@ -40,15 +40,6 @@ set(LLVM_ENABLE_TERMINFO OFF CACHE BOOL "")
 # LLVM settings that have an effect on the MLIR dialect
 set(LLVM_TARGETS_TO_BUILD "X86;AMDGPU" CACHE STRING "")
 
-# Build the ROCm conversions and run according tests if the AMDGPU backend
```
whchung, line 43: The logic is placed at `external/llvm-project/mlir/CMakeLists.txt` too so we don't need duplication here.

jerryyin (approved):
LGTM, I've always been confused by the diff between `MLIR_ENABLE_ROCM_RUNNER` and `MLIR_ROCM_RUNNER_ENABLED` too. Thanks for helping to clean things up.

whchung (approved):
LGTM. Thanks for coming up with this fix!

PR #533, opened 2022-03-01
transforming_for

krzysz00:
This is the refactor of how our lowering talks about loads, stores, and copies proposed in [issue #402](https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/402) .

The highlights:
- There's a new looping construct available, `miopen.transforming_for`. The arguments of this loop are
  - One or more iteration domains, which consist of an initial coordinate in an upper coordinate space and a series of `transform_map`s that map the upper space to a lower space
  - A constant array `bounds` that specifies the dimensions of the rectangle in the upper spaces that will be iterated through
- These `transforming_for` loops are lowered to affine loops that compute the lower coordinates either by applying the affine map in the `transform_map`s or by using index diffs - the choice is controlled by an attribute on the loop
- These loops may (and, in every case we have, are) be unrolled after creation, as controlled by the `forceUnroll` attribute. There is a delay between loop creation and unrolling that allows us to run loop invariant code motion before we unroll, which is an improvement (imo) over the previous design of doing the unrolling in C++
- `miopen-lowering-step4` runs common subexpression elimination after loop unrolling
- `miopen-lowering-step5` has been renamed to a more descriptive `miopen-loops-to-cf` and lives in its own file
- There are now shorthand ops, `miopen.insert_slice` and `miopen.extract_slice`. These allow you to get/put a sequence of entries in a vector without having to write out all the `extractelement` + `insertelement` code yourself. They also lower directly to `insertelement`/`extractelement` when trying to take a scalar "slice". These ops are handled by the new `miopen-expand-shorthand` pass, which runs after step 4.
- The global-memory access parts of the former `emitLoadLogic` and `emitStoreLogic` have been moved to the ops `miopen.buffer_store` and `miopen.buffer_load`
- `threadwise_load` and `threadwise_store` have been removed
- `threadwise_copy` and `threadwise_copy_v2` are retained for compatibility with fusion. I'd like to see the fusion process not rely on the presence of `threadwise_copy` after some future work, but that's a whole process that might involve a big rework of how fusion works

Note: this PR incorporates (and so, if merged, superscedes, #522

TODO for me:

- [x] Debug the atomic add test failures
- [x] Remove `emitStoreLogic`, migrating `xdlops_gemm_v2` to have the stores/loads inline (there being no other callers for than function)
- [x] Maybe generate some more `buffer_store` tests with a one-off script


krzysz00:
@whchung, this is now ready for review

krzysz00:
(however, having looked at the ISA, I realized there's a problem with how vector movement to/from LDS is lowering that I'll be fixing)

krzysz00:
The problem noted above has been fixed

(Also, there might be call for using `in_bounds_load` in xdlops_gemm_v2, but the memrefs of vectors made verification not work: that might be a future PR)

whchung:
I'll go over this PR today (3/3/2022) to understand the concepts behind it.

sjw36:
@whchung can you review please. We may have need of this for fixing xdlops fusion and in any case may conflict.

krzysz00:
To provide context, per discussions with @jungpark-mlir , the current semantics of `threadwise_copy_v2` make writing fusion for it a massive pain. The paint points are both the presence of `sourceOffset` (which this PR removes in favor of taking slices) and the lack of a memref -> vector copy operation.

In our meeting yesterday, I suggested to Jungwook that, instead of trying to create a cloned copy operation, he directly emit a `transforming_for` to load the values from the fusion buffer (using the coordinate transformations and starting coordinates from the result tensor and the vector being replicated).

That way, we don't need to define new specialized copy ops and can be clear about what we're doing.

sjw36:
I also need the fix for reading in the TransformMapAttr..

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -444,16 +693,17 @@ def MIOpen_BlockwiseGemmV2Op:
...
                    MemRefOf<[F32, F16, BF16, I8]>:$matrixB,
-                   ArgTransforms<2>:$transforms,
+                   IndexAttr:$ldsBufferOffsetA,
+                   IndexAttr:$ldsBufferOffsetB,
                    Index:$waveOffsetA,
                    Index:$waveOffsetB,
-                   MemRefOf<[F32, F16, BF16, I8, 
-                             VectorOfLengthAndType<[2, 4], [F32]>, 
-                             VectorOfLengthAndType<[2, 4, 8], [F16, BF16]>, 
+                   MemRefOf<[F32, F16, BF16, I8,
```
sjw36, line 721: Looks like alot of whitespace changes.

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -260,74 +256,380 @@ def MIOpen_WorkitemIdOp:
...
+    %i, %j = miopen.compute_transforms [](%c0, %c0)
+    %a0_0_0, %a1_0_0 = miopen.index_diff #transform_map1[0, 0] (%x, %y) -> (%a0, %a1, %a2)
+    %i_0_0 = addi %i, %cst0
+    %j_0_0 = addi %j, %cst0
+    %v_0_0 miopen.buffer_load {...} %place[%a0_0_0, %a1_0_0, %a2_0_0]
+    miopen.buffer_store {...} %v -> %buffer[%i_0_0, %j_0_0]
+    %res_0_0 = addi %c0, %c1 : inedx
+    %a0_0_1, %a1_0_1, %a2_0_1 = miopen.index_diff #transform_map1[0, 1] (%x, %y) -> (%a0, %a1, %a2)
+    ...
+    %res_3_7 = addi %res_3_6, %c1 : index
```
whchung: Several ops in this example don't really exist: `compute_transforms`, `index_diff`. Could this example be revised to depict the proposed semantic?

Also, assuming readers are more familiar with the constructs provided by affine dialect, perhaps it will be better to start from an example without unrolling or index diff maps. And then introduce the effects of unrolling and index diff on top of the naive nested affine loops produced by `transforming_for`.

krzysz00: That's what I get for writing the docs first and then changing names partway through. I'll fix it up.

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -159,7 +158,6 @@ def MIOpen_GridwiseGemmV2Op :
                    BoolArrayAttr:$aOobDims,
                    BoolArrayAttr:$bOobDims,
                    BoolArrayAttr:$cOobDims,
-                   ArgTransforms<3>:$transforms,
                    MIOpen_PaddingInfoAttr:$paddingInfo,
                    InMemoryDataOperation:$storeOperation)> {
```
whchung: Can we consolidate `storeOperation` used in `gridwise_gemm_v2`, and `dataOperation` in `buffer_store` into the same name?

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -260,74 +256,380 @@ def MIOpen_WorkitemIdOp:
...
+    `miopen.buffer_store` stores data to a global buffer, ignoring out of bounds writes.
+
+    The bounds on a write are the dimensions of the memref, though only the
+    dimensions specified in `oobDims` are checked before the write is issued.
+
+    Writes that our out of bounds are ignored.
+
+    This op supports vector writes and uses buffer store intrinsics if needed.
+
+    `paddingInfo` is used to apply backwards data store hacks if they're needed.
```
whchung, line 575: Can we add documentation on the meaning of `dataOperation`, and perhaps find a common name with the attribute `storeOperation` used in `gridwise_gemm_v2`?

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
krzysz00: Yeah. IMO we should have a lint for trailing whitespace so we stop getting stuff like this

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
krzysz00: Got any thoughts on which of the names makes more sense? Or if we should go all-in on the renamings and make it all something like `storeMethod`?

On the other hand, the advantage of the current scheme is that on gridwise_gemm_v2 it's clear the enum is about how the data will be stored, while in `buffer_store` (where we know it's a store) we don't do anything redundant. Maybe `storeOperation` on gemm and `method` on the actual store operation?

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -260,74 +256,380 @@ def MIOpen_WorkitemIdOp:
...
+
+    The buffer must reside in global memory.
```
whchung, line 580: Reading the code it seems the check sits in the verifier method.

Wondering is there a way in TableGen to guarantee that with a Trait? This can be done in the future PRs.

krzysz00: I didn't see a trait for memory space, that does feel like a nice todo item though

mlir/include/mlir/Dialect/MIOpen/utility/loweringUtils.h
```diff
@@ -47,6 +47,11 @@ inline int64_t calculateKBlockNum(int64_t n, int64_t ho, int64_t wo) {
   return gemmKBlocks;
 }
 
+/// Unwrap a value from the transforms surrounding it, gathering up the
+/// transforms
```
whchung: This method deserves an example to help illustrate its semantic.
Nit: Ending with a period `.`

whchung: And would it be better to make it returns a `std::tuple<Value, ArrayAttr>`? From the comment it is not clear `transforms` is also an output unless reading the implementation and the call sites.

krzysz00: Taking a reference like that feels like an indication of being an out parameter, but I'm willing to go tuple

mlir/lib/Dialect/MIOpen/MIOpenDialect.cpp
```diff
@@ -380,6 +380,486 @@ template <typename T> static LogicalResult verifyConvOp(T op) {
...
+  if (useIndexDiffs)
+    state.addAttribute(useIndexDiffsAttrName(state.name), b.getUnitAttr());
+
+  int32_t upperLen = bounds.size();
+  for (ValueRange upper : inits)
+    state.addOperands(upper);
+  state.addOperands(iterArgs);
+  state.addTypes(iterArgs.getTypes());
+
+  state.addAttribute(
```
whchung, line 482: Could you explain the idea behind this attribute?

mlir/lib/Dialect/MIOpen/MIOpenDialect.cpp
```diff
@@ -380,6 +380,486 @@ template <typename T> static LogicalResult verifyConvOp(T op) {
...
+      TransformingForOp::getOperandSegmentSizeAttr(),
+      b.getI32VectorAttr({upperLen * static_cast<int32_t>(inits.size()),
+                          static_cast<int32_t>(iterArgs.size())}));
+
+  // Set up region and block
+  Region *bodyRegion = state.addRegion();
+  Block &bodyBlock = bodyRegion->emplaceBlock();
+
+  SmallVector<int32_t> lowerStarts;
```
whchung, line 494: Some comments here would be helpful to indicate the purpose of the logic.

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -127,9 +127,6 @@ template <typename T> struct MILARewritePattern : public OpRewritePattern<T> {
     }
     nTWCopy->setAttr("bounds", b.getArrayAttr(twCopyBoundsAttr));
 
-    // 5. Adjust the copy to show the correct argument as global
-    nTWCopy->setAttr("globalArg", b.getIndexAttr(0));
```
whchung, line 131: @sjw36 Need your comment on this proposed change. I'm not that familiar with the expected syntax in this pass.

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -335,7 +176,7 @@ struct BlockwiseGemmRewritePattern : public OpRewritePattern<BlockwiseGemmOp> {
     auto loopOp = b.create<AffineForOp>(loc, 0, loopIteration);
 
     // inside the main loop.
-    auto lb = OpBuilder::atBlockTerminator(loopOp.getBody());
+    auto lb = OpBuilder::atBlockTerminator(loopOp.getBody(), b.getListener());
```
whchung, line 180: Do we really depend on methods in `OpBuilder::Listener`, such as `notifyOperationInserted` / `notifyOperationRemoved`?

krzysz00: The pattern rewriting infrastructure does, I'm pretty sure

whchung: Compared with the original implementation, why didn't we need it before, but need it now?

krzysz00: Because not doing that was wrong

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
krzysz00: I added that line back when I reworked `threadwise_copy`, and so I figure I can remove it since the `threadwise_copy` rewrite will handle this without the attribute.

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
whchung: Could you shed some lights on the reason? Quite a few call sites in `lib/Dialect/Affine/Utils/LoopUtils.cpp` do so without the `listener` argument. From the interface the argument is optional, and in the PR I don't see any overrides to `notifyOperationInserted/notifyOperationRemoved`.

krzysz00: The code in `LoopUtils` isn't meant to run within a pattern rewrite

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -260,74 +256,380 @@ def MIOpen_WorkitemIdOp:
...
+def MIOpen_TransformingForOp :
+    MIOpen_Op<"transforming_for", [RecursiveSideEffects,
+      DeclareOpInterfaceMethods<LoopLikeOpInterface>,
+      AttrSizedOperandSegments,
+      SingleBlockImplicitTerminator<"::mlir::miopen::YieldOp">]>,
+    Arguments<(ins Variadic<Index>:$upperInits,
+      ArgTransformsAttr:$transforms,
+      Variadic<AnyType>:$iterInits,
+      IndexArrayAttr:$bounds,
+      I32ElementsAttr:$lowerStarts,
```
whchung, line 349: It seems quite a few operations depends on the layout of this attribute. Could it be documented?

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -260,74 +256,380 @@ def MIOpen_WorkitemIdOp:
...
+    "ArrayAttr":$transforms,
+    "ArrayAttr":$bounds, "bool":$forceUnroll, "bool":$useIndexDiffs,
+    CArg<"ValueRange", "llvm::None">:$iterArgs)>];
+
+  let extraClassDeclaration = [{
+    uint32_t lowerStart(uint32_t n) {
+      return *(lowerStarts().getValues<uint32_t>().begin() + n);
+    }
+
+    Block::BlockArgListType getLowerCoords(uint32_t domain) {
```
whchung, line 448: From the call sites it appears this method is crucial so documentation is needed. It's not entirely obvious what each domain is for. And the concept of `domain` is not defined in the documentation of `transforming_for`.

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -260,74 +256,380 @@ def MIOpen_WorkitemIdOp:
...
+      uint32_t start = lowerStart(domain);
+      uint32_t end = lowerStart(domain + 1);
+      return getBody()->getArguments().slice(start, end - start);
+    }
+    Block::BlockArgListType getLowerCoords() {
```
whchung, line 453: It seems no one really depends on this?

krzysz00: I figured we'd add the shorthand. Do you want the dead code removed instead?

mlir/test/Dialect/MIOpen/vector_slices.mlir
```diff
@@ -0,0 +1,73 @@
+//RUN: miopen-opt -miopen-expand-shorthand %s | FileCheck %s
```
whchung: nit: `// RUN:`

mlir/test/Dialect/MIOpen/lowering_threadwise_copy_v2.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: miopen-opt -miopen-lowering-step4 %s | FileCheck %s
+// RUN: miopen-opt -miopen-lowering-step3 -miopen-lowering-step4 %s | FileCheck %s
```
whchung: Do we still need step4 here now the lowering for these ops have been moved to step3?

krzysz00: Yeah, since the tests check for raw_buffer_store

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
whchung: The enum itself might deserve a better name. Perhaps `InMemoryDataOperation` -> `DataStoreMethod`. Then both `storeOperand` and `dataOperand` can simply be called `method`.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -41,53 +48,66 @@ struct LowerMIOpenOpsStep2Pass
...
-                     .template cast<IntegerAttr>()
-                     .getInt();
-    storeLength = gop->getAttr("matrix_b_dest_data_per_write_dim_n")
-                      .template cast<IntegerAttr>()
-                      .getInt();
-    vectorDim = gop->getAttr("matrix_b_source_vector_read_dim")
-                    .template cast<IntegerAttr>()
-                    .getInt();
-  }
+void computeLoadStoreTypeInfo(OpBuilder &b, ArrayRef<int64_t> sliceLengths,
```
whchung, line 62: The inputs and outputs are mixed up in the parameter list. Could we rearrange them, or fallback to using tuples?

krzysz00: The outputs all come after the inputs (with the minor exception of `vectorDim` which is read-write to allow for KPack stuff.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
```diff
@@ -48,2298 +62,939 @@ struct LowerMIOpenOpsStep4Pass
...
-  }
-  // Return computed coordinate length.
-  return coordLength;
-}
-
-//===----------------------------------------------------------------------===//
-// Utility function to emit the logic to copy between naive tensors.
-// This function is used within the lowering logic of threadwise_copy.
-//===----------------------------------------------------------------------===//
-void emitNaiveTensorCopyLogic(OpBuilder &b, Location loc, int64_t nSliceRow,
```
whchung, line 577: Nice this function is getting eliminated. One of the relic porting codes from C++.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -128,22 +148,166 @@ void affixThreadwiseCopyV2Attributes(ThreadwiseCopyV2Op top,
...
+    storeLength = storingVectorType.getNumElements();
+
+  size_t nUpper = bufferStart.size();
+  bool complexVectorStore = (storeLength > 1) && (vectorDim != nUpper - 1);
+
+  Value zero = b.createOrFold<ConstantIndexOp>(loc, 0);
+  SmallVector<Value, 5> linearInit(nUpper, zero);
+
+  ArrayAttr bufferTransforms;
+  buffer = untransform(b, buffer, bufferTransforms);
```
whchung: Without a proper documentation or some examples for `untransform`, this line is hard to understand.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -128,22 +148,166 @@ void affixThreadwiseCopyV2Attributes(ThreadwiseCopyV2Op top,
...
+      loopBounds,
+      /*forceUnroll=*/true, /*useIndexDiffs=*/true);
+  OpBuilder::InsertionGuard guard(b);
+  b.setInsertionPointToStart(loop.getBody());
+
+  // We can use vector.transfer_write if the vectorized dimension of the
+  // load-store loops is the fastest-moving dimension of the store loop.
+  // Otherwise, we need to gather elements from the result vector in order to
+  // store them
+  if (fullyScalar)
```
whchung: This is a 3-pronged if statement, let's use curly braces here to conform with other branches.

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -378,22 +220,29 @@ struct BlockwiseGemmRewritePattern : public OpRewritePattern<BlockwiseGemmOp> {
...
+        loc,
+        ArrayRef<ValueRange>{matrixAThreadwiseCopySourceCoords,
+                             matrixAThreadwiseCopyDestCoords},
+        ArrayRef<Attribute>{transformsA, emptyArr},
+        ArrayRef<int64_t>{1, KPerThread, MPerThreadSubC},
+        /*forceUnroll=*/true, /*indexDiffs=*/false);
+    OpBuilder copyABuilder =
+        OpBuilder::atBlockTerminator(copyALoop.getBody(), lab.getListener());
+    Value aCopy = copyABuilder.create<memref::LoadOp>(
+        loc, matrixA, copyALoop.getLowerCoords(0));
```
whchung: I feel `0` / `1` passed to `getLowerCoords` ought to be in an enumeration.
It seems in this PR we generally follow an idiom that `0` means the lower coordinates for the source to be read, and `1` means the lower coordinates for the destination to be stored.

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -378,22 +220,29 @@ struct BlockwiseGemmRewritePattern : public OpRewritePattern<BlockwiseGemmOp> {
...
-        matrixAThreadwiseCopyDestCoords);
-    affixThreadwiseCopyAttributes(threadwiseCopyAMatrixOp, op, b,
-                                  /*isMatrixA=*/true);
+    auto copyALoop = lab.create<TransformingForOp>(
+        loc,
+        ArrayRef<ValueRange>{matrixAThreadwiseCopySourceCoords,
+                             matrixAThreadwiseCopyDestCoords},
+        ArrayRef<Attribute>{transformsA, emptyArr},
+        ArrayRef<int64_t>{1, KPerThread, MPerThreadSubC},
+        /*forceUnroll=*/true, /*indexDiffs=*/false);
```
whchung, line 230: Any particular reason why index diffs is not in use here?

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: I'm not sure this is the right context for an enum - these are, in effect, indices into the arrays passed to the `TransformingForOp` builder.

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: Because they weren't in the original

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: Yeah, I did deliberately choose that particular idiom, though fundamentally, nothing would go wrong if we went for something like "buffers first, then indexing into the vectors" as the ordering instead.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: Hopefully the new version is better?

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
krzysz00: Yep!

mlir/test/Dialect/MIOpen/lowering_threadwise_copy_v2.mlir
krzysz00: Would you rather I rework the tests?

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
whchung: Yes, and the use of `0` and `1` are almost universally visible in every call site after a `transforming_for` is emitted. Without a proper documentation of `getLowerCoords` it's hard to grasp what those values are for.

Would it be possible to do a key/value pair mapping to make the idea of `iteration domains` more visible?

mlir/include/mlir/Dialect/MIOpen/utility/loweringUtils.h
krzysz00: Update: went with a tuple, it does make sense.

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
krzysz00: I went with just `StoreMethod`

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -260,74 +256,401 @@ def MIOpen_WorkitemIdOp:
...
+    The bounds on a write are the dimensions of the memref, though only the
+    dimensions specified in `oobDims` are checked before the write is issued.
+
+    Writes that our out of bounds are ignored.
+
+    This op supports vector writes and uses buffer store intrinsics if needed.
+
+    `paddingInfo` is used to apply backwards data store hacks if they're needed.
+
+    `storeOperation` controls whether the data is written to memory, overwriting
```
whchung: nit: `storeOperation` -> `storeMethod`.

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: I'll argue it's very visible in the printed IR - each iteration domain is its own `(%argX, %argY, %argZ) = [#transform_mapA, #transform_mapB, ...](%start1, %start2)` section in the printed loop. And this has some symmetry with `getBody([unsigned int block])` on `Operation`.

Though I do agree it isn't immediately obvious from the code which one's zero and which one's one - you have to look back at the constructor for the loop to get that.

... then again, you have to look back at the loop constructor to know which iter_arg is which.

krzysz00: (and if you have thoughts on how to improve the documentation of `getLowerCoords()`, I'm all ears)

mlir/test/Dialect/MIOpen/lowering_threadwise_copy_v2.mlir
whchung: That could be a fairly good idea as existing tests are too loose.

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
whchung: Let's keep it for now. Perhaps it could be used in other use cases.

mlir/include/mlir/Dialect/MIOpen/utility/loweringUtils.h
```diff
@@ -47,6 +47,18 @@ inline int64_t calculateKBlockNum(int64_t n, int64_t ho, int64_t wo) {
   return gemmKBlocks;
 }
 
+/// Unwrap a value from the transforms surrounding it, gathering up the
+/// transforms.
+/// Given a Value `v` that is generated by
+///   %v1 = miopen.transform [#transform1] %v0
+///   %v = miopen.transform [#transform2] %v1
+/// this method will return %v0 and an ArrayAttr equal to [#transform2,
+/// #transform1]. If `existinng` is passed in, it must be an array of
```
whchung: nit: `existing`.

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
whchung: Perhaps add `/*domain=*/` before the `0` / `1`?

krzysz00: Good idea  - does it seem to work?

whchung (commented):
The PR is pretty big and I'm down to 2 files remain to be studied: `GridwiseGemmToBlockwise.cpp` and `ThreadwiseToStdlib.cpp`. Will continue finish them tomorrow.

In general, yet again a very good refactoring on the dialect. It makes the syntax and semantic of the dialect much more comprehensible. With the abstraction of `transforming_for`, `index_diff_update`, combined with previously instilled `TransformMap`. Part of the dialect are now becoming much more presentable and potentially upstreamable.

+ @asroy for awareness.

whchung (approved):
And, to help unblock fusion so we can do fusion with XDLOPS I'm approving hereby.

krzysz00:
I'll take a look at everything tomorrow morning

PR #535, opened 2022-03-02
Add int8 E2E test cases

yiqian1:


jerryyin (changes requested):
Thank you @yiqian1, I'm blocking this PR until all changes are done.

krzysz00:
@yiqian1 Is there any particular reason this PR is running nightly tests?

jerryyin:
@krzysz00 That's me kicking off nightly just so that we can do an initial regression on how well i8 is supported with existing implementation.

krzysz00:
Ah

Could we perhaps run that test by hand off on one of the Lockharts instead?

jerryyin:
@krzysz00 Won't be necessary since we already achieved full pass on int8 xdlops testing on resnet50. We'd flip back the switch and shoot for merging this PR so at least we have a good regression baseline.

jerryyin (approved):
LGTM

PR #536, opened 2022-03-02
Emit utility kernels to set 0 and truncation for backward weight kernels with uses atomic add.

whchung:
Teach MLIR MIOpen dialect to emit utility kernels to set 0 and truncation for backward weight kernels with uses atomic add.

For `f32`, a set 0 kernel will be invoked prior to the backward weight convolution kernel to set the output (filter tensor) be all 0.
For `f16`, a set 0 kernel will be invoked prior to the backward weight convolution kernel to set the workspace be all 0. After the backward weight convolution kernel is executed, a truncation kernel will be invoked to do elementwise conversion from `f32` in the workspace to `f16` in the output (filter tensor).

MLIR-side coding is complete, and passing all nightly CI unit tests locally.

MIOpen solver side needs minor additional work to follow what's done in backward data when it involves multiple kernels. 

Upon working on this PR, it is identified the same treatment is needed for backward data convolution as well. Will submit relevant logic in a separate PR.

Eventually, in `miopen-gen` we should be able to eliminate the `isOut` option when we to random initialization of tensors on CPU. And in MIOpen MLIR solver, all dependency to `SetTensor` / `CastTensor` which depend on OpenCL GPU kernels could all be eliminated.


mlir/include/mlir/Dialect/MIOpen/utility/builderUtils.h
```diff
@@ -11,6 +11,12 @@ namespace miopen {
 // Utility function to emit constant zero op. Can return scalars or vectors.
 Value createZeroConstantOp(OpBuilder &b, Location loc, Type type);
 
+// Utility function to emit type conversion ops.
+Value createTypeConversionOp(OpBuilder &b, Location loc, Value source,
```
krzysz00, line 23: You don't need `sourceType` - it's just `Type sourceType = source.getType()` in the function

I think I made that change when pulling that function out in transforming_for

Great minds think alike, it seems

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
```diff
@@ -45,20 +52,19 @@ void AffixTuningParameters::runOnOperation() {
...
 }
 
 template <typename T>
-void AffixTuningParameters::affixTuningParametersImpl(T &op) {
-  OpBuilder b(op.getContext());
-
-  ConvolutionContext convContext = populateConvContext(op);
-
+std::tuple<int64_t, int64_t, int64_t, int64_t, int64_t, int64_t, int64_t,
+           int64_t, int64_t>
```
krzysz00, line 63: Just as a sanity check, do we perhaps want to make this some sort of `struct ConvDims {...}` instead? Or at least to document the order of the arguments in this tuple. int64_t^{big} just feels like a bad return type to me.

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
```diff
@@ -115,17 +116,88 @@ void AffixTuningParameters::affixTuningParametersImpl(T &op) {
...
+    if (needExtraPad == true) {
+      assert(gemmId == 0);
+    } else {
+      assert((gemmId >= 0) && (gemmId < 3));
+      switch (gemmId) {
+      case 0:
+      case 2:
+        // Override grid_size and block_size be 1 for utility kernels.
+        // FIXME. Use better sizes for speedups.
+        op->setAttr("grid_size", b.getI32IntegerAttr(1));
```
krzysz00, line 167: What do grid_size and block_size do?

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -177,6 +178,81 @@ void affixGridwiseGemmAttributes(Operation *convOp, Operation *gop,
...
+  } else {
+    llvm_unreachable("Incorrect memref type supplied");
+  }
+  auto outputDataType = output.getType().cast<MemRefType>().getElementType();
+  auto zeroOp = createZeroConstantOp(b, loc, outputDataType);
+  auto collapsedOutput = createCollapseShapeOp(b, loc, output);
+  ArrayRef<int64_t> collapsedOutputShape =
+      collapsedOutput.getType().cast<MemRefType>().getShape();
+  llvm::SmallVector<AffineForOp, 1> affineLoops;
+  for (auto bound : collapsedOutputShape) {
```
krzysz00: `buildAffineLoopNest()` - this can be massively simplified

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -177,6 +178,81 @@ void affixGridwiseGemmAttributes(Operation *convOp, Operation *gop,
...
+      collapsedFilter.getType().cast<MemRefType>().getShape();
+  ArrayRef<int64_t> collapsedWorkspaceShape =
+      collapsedWorkspace.getType().cast<MemRefType>().getShape();
+  llvm::SmallVector<AffineForOp, 1> affineLoops;
+  for (unsigned iter = 0; iter < collapsedFilterShape.size(); ++iter) {
+    auto boundFilter = collapsedFilterShape[iter];
+    auto boundWorkspace = collapsedWorkspaceShape[iter];
+    assert((boundFilter == boundWorkspace) &&
+           "Filter tensor and workspace size mismatch");
+    auto loop = b.create<AffineForOp>(loc, 0, boundFilter);
```
krzysz00: Same as above. Also, this can be a `transforming_for` later

mlir/lib/Dialect/MIOpen/utility/builderUtils.cpp
```diff
@@ -80,5 +81,71 @@ Value createZeroConstantOp(OpBuilder &b, Location loc, Type type) {
...
 }
 
+//===----------------------------------------------------------------------===//
+// Utility function to emit type conversion ops.
+//===----------------------------------------------------------------------===//
+Value createTypeConversionOp(OpBuilder &b, Location loc, Value source,
+                             Type sourceType, Type destType) {
+  // Convert from sourceType to destType if necessary.
+  Value result = source;
+  Type sourceElemType = sourceType;
```
krzysz00: `Type sourceType = source.getType();` above

Also, cleanup note since we're touching this code: all of this could be simplified with `mlir::getElementTypeOrSelf()`

mlir/lib/Dialect/MIOpen/utility/builderUtils.cpp
```diff
@@ -80,5 +81,71 @@ Value createZeroConstantOp(OpBuilder &b, Location loc, Type type) {
...
+      llvm_unreachable("Only fp32, fp16, or bf16 targets for data conversion");
+    }
+  }
+  return result;
+}
+
+Value createCollapseShapeOp(OpBuilder &b, Location loc, Value source) {
+  auto ctx = b.getContext();
+  auto sourceType = source.getType().cast<ShapedType>();
+  if (!sourceType.hasStaticShape()) {
```
krzysz00: This could be a `assert(sourceType.hasStaticShape() && "Only [...]")`

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -177,6 +178,81 @@ void affixGridwiseGemmAttributes(Operation *convOp, Operation *gop,
...
+LogicalResult elementwiseConversion(Conv2DBwdWeightOp op, PatternRewriter &b) {
+  auto loc = op.getLoc();
+  assert(op.workspace() && "Op has no workspace");
+  auto filter = op.filter();
+  auto workspace = op.workspace();
+  auto filterDataType = filter.getType().cast<MemRefType>().getElementType();
+  auto workspaceDataType =
+      workspace.getType().cast<MemRefType>().getElementType();
+  auto collapsedFilter = createCollapseShapeOp(b, loc, filter);
+  auto collapsedWorkspace = createCollapseShapeOp(b, loc, workspace);
```
krzysz00, line 220: Why do we need this?

krzysz00 (commented):
I've got some questions and simplifications. Also, note that memrefs have a default indexing scheme, namely row-major

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
whchung: I'd like to keep this as-is for now. In https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/439 we'd consolidate data structures commonly used between the generator and the lowering pipeline.

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
whchung: These are the parameters sent to the runtime when we launch a kernel. It specifies the size of a workgroup (block_size), and how many workgroups (grid_size) we need. Here you can see we use an extremely conservative way to ease the programming of the code emission logic for those utility kernels.

Ideally, we want something like this:
```
    size_t offset = (hipBlockIdx_x * hipBlockDim_x + hipThreadIdx_x);
    size_t stride = hipBlockDim_x * hipGridDim_x;

    for (size_t i = offset; i < N; i += stride) {
       // do the elementwise operations we want
    }
```

I was hoping upstream transforms for linalg could help me achieve that but I couldn't find an end-to-end pipeline yet.

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
whchung: We are certain we do an element-wise conversion, so we can use `memref.collapse_shape` to make it a 1D memref to simplify further code emissions.

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
krzysz00: I'd've gone for a different breakup.

```
per_thread_work = N / (n_total_threads)
offset = global_thread_id * per_thread_work
for (int i = 0; i < per_thread_work; ++i)
  buffer[offset + i] = ...;
```

krzysz00: And our coordinate transform scheme should be able to support this?

whchung:
`Test selected MIOpen configs` is failing because of lack of perfdb. @jerryyin taught me how to get around that and I think we'll need to apply that again. Also I'll apply some simplifications advised by @krzysz00 .

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
krzysz00: Ah, yeah, fair enough.

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
whchung: on another round of examining samples in upstream, this pipeline may offer some help:
```
-convert-linalg-to-parallel-loops -convert-parallel-loops-to-gpu
```
I'll do some local experiments. Should we get something working end-to-end I'll submit a new PR.

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
whchung: Since we already collapsed the memref to 1D, we don't even need the logic to emit a nested loop here.

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
whchung: Same as above. I'm switching to a simple 1D affine loop.

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
```diff
@@ -334,35 +408,8 @@ bool Conv2dGenerator::hasWorkspace(OpBuilder &builder) const {
...
-      auto inDim = canonicalizeDims(config.inputDimension, config.inputLayout);
-      auto filDim =
-          canonicalizeDims(config.filterDimension, config.filterLayout);
-      auto outDim =
-          canonicalizeDims(config.outputDimension, config.outputLayout);
-
-      int64_t gemmMSize, gemmNSize, gemmKSize;
-      gemmMSize = filDim["k"];
-      gemmKSize = outDim["n"] * outDim["h"] * outDim["w"];
-      gemmNSize = filDim["c"] * filDim["y"] * filDim["x"];
```
krzysz00, line 349: Can you go find all the instances of this out in `ConvToGemmLowering` and clean them up as well?

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
```diff
@@ -37,6 +37,13 @@ struct AffixTuningParameters : public MIOpenOpsAffixTuningParametersPassBase<Aff
 
   // Actual implementation.
   template <typename T> void affixTuningParametersImpl(T &op);
+
+  void affixBackwardWeightUtilityKernels(miopen::Conv2DBwdWeightOp &op);
+
+  template <typename T>
+  std::tuple<int64_t, int64_t, int64_t, int64_t, int64_t, int64_t, int64_t,
+             int64_t, int64_t>
+  fetchDimensions(T &op);
```
krzysz00, line 46: Any particular reason this is a template? You could either take an `Operation *` or, in things we probably should've done a while ago, define a trait that all the convolution ops share.

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
krzysz00: Fair enough, and if it works, it works.

krzysz00 (commented):
Broadly speaking, LGTM. I have one or two comments that I wanted to leave, but if you're happy with how things are, I'll hit approve.

I'd like a followup ticket for making the zeroing/truncation kernels more efficient (ex. with having each thread do a N / total_threads-sized chunk of the work) but that doesn't need to be here.

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
whchung: tracking it at https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/466

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
whchung: Tracking it at https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/467 and I'll cover it in the next sprint.

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
whchung: Same reason `affixTuningParametersImpl` is a template. Different ops will invoke `fetchDimensions`. And we are explicitly walking on those different ops in `AffixTuningParameters::runOnOperation()`.

Using a trait is an interesting option which we can explore in https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/439

krzysz00: Fair enough, I haven't poked this stuff in a detail, and so on.

krzysz00 (approved):
LGTM, let's send this in!

PR #537, opened 2022-03-03
Hack for BF16 in the cpu verifier

yiqian1:


mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -1145,6 +1145,34 @@ const char *getTypeStr(const mlir::Type &type) {
...
+  mlir::Type i16 = bt0.getIntegerType(16);
+  mlir::Type i32 = bt0.getI32Type();
+  loadOp = bt0.create<arith::BitcastOp>(loc, i32, loadOp);
+  // <loadOp> = <loadOp> & FFFF0000
+  auto cst16Op = bt0.create<arith::ConstantIntOp>(loc, -65536, i32);
+  loadOp = bt0.create<arith::AndIOp>(loc, loadOp, cst16Op);
+  loadOp = bt0.create<arith::BitcastOp>(loc, bt0.getF32Type(), loadOp);
+
+  return loadOp;
+}
```
krzysz00, line 1170: In general, I'll note that there's logic for this sort of thing in `external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp`, see `SoftwareBF16Trunc` there. Could we take this as the time to factor that out into a general utility, something like `SoftwareBF16`, that can be applied to both the host and GPU code.

krzysz00 (commented):
I've got some thoughts on how to fix this up better

mlir/tools/miopen-gen/miopen-gen.cpp
yiqian1: Good thoughts. Let me try it. Thanks.

krzysz00: For full generality, having `SoftwareBF16` be a pass over `arith` might be the right call. There's really a bunch of design decisions to make here.

Also, if you want to convert a constant float to a bfloat, `APFloat` and its bitcasting methods might be what you're looking for instead of the function you added here.

Let me know how important it is to get this in so I can triage "let's put it in" vs "let's get this right"

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -1368,6 +1397,45 @@ createVerifierFunc(ModuleOp &module, const KernelIF &kernel,
...
+
+  auto getFVal = [&](float val) -> mlir::Value {
+    llvm::APFloat apVal(val);
+    bool ignored;
+    if (elemFType.isF16()) {
+      apVal.convert(llvm::APFloat::IEEEhalf(), APFloat::rmNearestTiesToEven,
+                    &ignored);
+    } else if (elemFType.isBF16()) {
+      apVal.convert(llvm::APFloat::BFloat(), APFloat::rmNearestTiesToEven,
+                    &ignored);
```
krzysz00, line 1413: bitcast to APInt here and return a `ConstantIntOp`

krzysz00 (approved):
LGTM, I'll not hold this up for minor style nits

PR #538, opened 2022-03-03
Add MI-200 support to CI

krzysz00:
This change allows us to use MI-200 in our CI. The PR tests will run on either MI-100 or MI-200, depending on node availability.

Notes on current status:
1. I can't fully verify `nightly` due to a lack of a gfx90a perfdb. I'm currently testing out the new `weekly` at http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2F2022-02-mlir-gfx90a-test/detail/2022-02-mlir-gfx90a-test/18/pipeline - once I've confirmed that works, I'll want to temporarily move `weekly` to use this branch in order to enable generating both gfx908 and gfx90a perf-db
2. I've removed stashing the libMLIRMIOpen build since getting a build artifact portable between different Jenkins workspaces might be rather tricky. This'll annoyingly make nightly take more time
3. The perf benchmarking is now one long sequential process. This has the advantage that we won't get results from multiple nodes in the same perf report, but has the disadvantage that the builds will take longer. This also reduces the degree of parallelism in the nightly CI to some extent
4. This PR is currently skip-ci since we haven't confirmed it's mergeable and since running PR CI is a waste of resources
5. Getting a combined MI-100 + MI-200 perf report is future work.


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -494,6 +560,9 @@ pipeline {
                 }
                 stage("Build MLIR") {
                     steps {
+                        // Clean up build to disable static library and allow
+                        // mlir-rocm-runner
+                        sh 'rm -rf build'
```
yiqian1: This will also delete the performance results created in the stage "Performance Test: Tuned vs Untuned". They are needed to create performance html.

krzysz00: Good point
Not sure how I missed that
Deleting CMakeCache.txt might be enough

yiqian1:
Thanks. LGTM

yiqian1 (approved):

krzysz00:
Note: this currently isn't mergeable due to issues around durable tasks and MI-200. See errors like http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir-weekly/detail/mlir-weekly/29/pipeline/150/

PR #539, opened 2022-03-03
[DO NOT SQUASH] Extend tuning parameters with kpack turned off

jerryyin:
This fix https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/465, as well as a number of issues listed below:

- Amend vectorization size decision to be dependent on data width only
- Amended valid k/wave as both xdlops for int8 are reduction ops and needs k/wave to be at least 2/4
- Extended support of i8 type to MIOpen::buffer_load
- Extended support of i32 type to MubufStore
- Corrected lowering pattern of MubufLoad and MubufStore to incorporate i8
- Make Buffer load lowering to emit scalar load when the load is too narrow or too wide

Refer to full nightly pass [here](http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-539/10/pipeline/).

### DIFF
--- a/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
+++ b/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
@@ -178,7 +178,7 @@ struct MubufLoadOpLowering : ConvertToLLVMPattern {
     Type LLVMResultType = typeConverter->convertType(resultType);
 
     // use standard load for:
-    // 1) loading scalar f16 and i16 (bf16) from global (addrspace 0).
+    // 1) loading scalar from global (addrspace 0).
     if ((srcElementType.getIntOrFloatBitWidth() != 32) &&
         (srcMemRefType.getMemorySpaceAsInt() == 0 &&
          !resultType.isa<VectorType>())) {
@@ -190,8 +190,8 @@ struct MubufLoadOpLowering : ConvertToLLVMPattern {
     }
 
     // use standard load for:
-    // 2) loading scalar and vector f16 and i16 (bf16) from LDS (addrspace 3).
-    // 3) loading scalar and vector f16 and i16 (bf16) from VGPR (addrspace 5).
+    // 2) loading scalar and vector from LDS (addrspace 3).
+    // 3) loading scalar and vector from VGPR (addrspace 5).
     if (srcMemRefType.getMemorySpaceAsInt() == 3 ||
         srcMemRefType.getMemorySpaceAsInt() == 5) {
       Value dataPtr =
@@ -295,18 +295,23 @@ struct MubufLoadOpLowering : ConvertToLLVMPattern {
         loc, LLVMI1Type, rewriter.getIntegerAttr(I1Type, 0));
 
     if (srcElementType.getIntOrFloatBitWidth() != 32) {
-      // for f16 and i16 (bf16) types, use f32 buffer_load and bitcast the
-      // result.
+      // For non-f32 types, use f32 buffer_load and bitcast the result.
       assert(resultType.isa<VectorType>());
-      // deduce the interim type for f16 / i16 (bf16).
       VectorType vectorResultType = resultType.template cast<VectorType>();
       auto vectorShape = vectorResultType.getShape();
+
+      // For int8, each 32 bits will fit 4;
+      // For f16 and i16, each 32 bits will fit 2
+      int64_t numFit32 = 32 / vectorResultType.getElementTypeBitWidth();
+      int64_t lastDimShape = vectorShape.back() / numFit32;
 
       SmallVector<int64_t, 1> interimShape;
       for (unsigned iter = 0; iter < vectorShape.size() - 1; ++iter)
         interimShape.push_back(vectorShape[iter]);
-      interimShape.push_back(vectorShape[vectorShape.size() - 1] >> 1);
-      bool useScalarF32 = (vectorShape.size() == 1) && (vectorShape[0] == 2);
+      interimShape.push_back(lastDimShape);
+      bool useScalarF32 =
+          (vectorShape.size() == 1) && (vectorShape[0] == numFit32);
+
       Type interimResultType;
       if (useScalarF32)
         interimResultType = rewriter.getF32Type();
@@ -315,6 +320,7 @@ struct MubufLoadOpLowering : ConvertToLLVMPattern {
             VectorType::get(interimShape, rewriter.getF32Type());
       Type interimLLVMResultType =
           typeConverter->convertType(interimResultType);
+      interimResultType.dump();
 
       Value interimLoad = rewriter.create<ROCDL::MubufLoadOp>(
           loc, interimLLVMResultType, rsrc, vindex, voffset, slc, glc);
@@ -352,7 +358,7 @@ struct MubufStoreOpLowering : ConvertToLLVMPattern {
     auto adaptorShift = adaptor.shift();
     Type valueType = mubufStoreOp.value().getType();
 
-    // use standard store for storing scalar f16 and i16 (bf16).
+    // use standard store for storing scalar
     if ((dstElementType.getIntOrFloatBitWidth() != 32) &&
         !valueType.isa<VectorType>()) {
       Value dataPtr =
@@ -450,18 +456,23 @@ struct MubufStoreOpLowering : ConvertToLLVMPattern {
         loc, LLVMI1Type, rewriter.getIntegerAttr(I1Type, 0));
 
     if (dstElementType.getIntOrFloatBitWidth() != 32) {
-      // for f16 and i16 (bf16) types, use f32 buffer_store and bitcast the
-      // result.
+      // for non-f32 types, use f32 buffer_store and bitcast the result.
       assert(valueType.isa<VectorType>());
-      // deduce the interim type for f16 / i16 (bf16).
       VectorType vectorResultType = valueType.template cast<VectorType>();
       auto vectorShape = vectorResultType.getShape();
+
+      // For int8, each 32 bits will fit 4;
+      // For f16 and i16, each 32 bits will fit 2
+      int64_t numFit32 = 32 / vectorResultType.getElementTypeBitWidth();
+      int64_t lastDimShape = vectorShape.back() / numFit32;
 
       SmallVector<int64_t, 1> interimShape;
       for (unsigned iter = 0; iter < vectorShape.size() - 1; ++iter)
         interimShape.push_back(vectorShape[iter]);
-      interimShape.push_back(vectorShape[vectorShape.size() - 1] >> 1);
-      bool useScalarF32 = (vectorShape.size() == 1) && (vectorShape[0] == 2);
+      interimShape.push_back(lastDimShape);
+      bool useScalarF32 =
+          (vectorShape.size() == 1) && (vectorShape[0] == numFit32);
+
       Type interimValueType;
       if (useScalarF32)
         interimValueType = rewriter.getF32Type();
@@ -582,16 +593,21 @@ struct RawbufStoreOpLowering : ConvertToLLVMPattern {
 
     if (dstElementType.getIntOrFloatBitWidth() != 32) {
       if (auto vectorResultType = valueType.dyn_cast<VectorType>()) {
-        // for f16 and i16 (bf16) types, use f32 buffer_store and bitcast the
-        // result.
-        // deduce the interim type for f16 / i16 (bf16).
+        // For non-f32 types, use f32 buffer_store and bitcast the result.
         auto vectorShape = vectorResultType.getShape();
+
+        // For int8, each 32 bits will fit 4;
+        // For f16 and i16, each 32 bits will fit 2
+        int64_t numFit32 = 32 / vectorResultType.getElementTypeBitWidth();
+        int64_t lastDimShape = vectorShape.back() / numFit32;
 
         SmallVector<int64_t, 1> interimShape;
         for (unsigned iter = 0; iter < vectorShape.size() - 1; ++iter)
           interimShape.push_back(vectorShape[iter]);
-        interimShape.push_back(vectorShape[vectorShape.size() - 1] >> 1);
-        bool useScalarF32 = (vectorShape.size() == 1) && (vectorShape[0] == 2);
+        interimShape.push_back(lastDimShape);
+        bool useScalarF32 =
+            (vectorShape.size() == 1) && (vectorShape[0] == numFit32);
+
         Type interimValueType;
         if (useScalarF32)
           interimValueType = rewriter.getF32Type();
@@ -605,8 +621,8 @@ struct RawbufStoreOpLowering : ConvertToLLVMPattern {
             loc, interimLLVMValueType, adaptorValue);
         rewriter.replaceOpWithNewOp<ROCDL::RawbufStoreOp>(
             op, bitcastedValue, rsrc, voffset_shift, vindex, zeroglcslc);
-      } else { // f16 and i16 (bf16) types scalar value
-               // FIXME:there is a problem with bf16 when in_h and in_w is odd
+      } else {
+        // FIXME:there is a problem with bf16 when in_h and in_w is odd
         rewriter.replaceOpWithNewOp<ROCDL::RawbufStoreOp>(
             op, adaptorValue, rsrc, voffset_shift, vindex, zeroglcslc);
       }
--- a/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
+++ b/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
@@ -320,7 +320,6 @@ struct MubufLoadOpLowering : ConvertToLLVMPattern {
             VectorType::get(interimShape, rewriter.getF32Type());
       Type interimLLVMResultType =
           typeConverter->convertType(interimResultType);
-      interimResultType.dump();
 
       Value interimLoad = rewriter.create<ROCDL::MubufLoadOp>(
           loc, interimLLVMResultType, rsrc, vindex, voffset, slc, glc);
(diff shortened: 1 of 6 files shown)


mlir/lib/Dialect/MIOpen/Tuning/GridwiseGemmParams.cpp
```diff
@@ -268,6 +268,12 @@ LogicalResult PopulateParamsXDL::populateDerived(
     return failure();
   }
 
+  res = isKpackValid(&params, gemmADerivedParam, gemmBDerivedParam);
+  if (failed(res)) {
+    LLVM_DEBUG(llvm::dbgs() << "Incoherent kpack tuning parameter.\n");
+    return failure();
+  }
```
jerryyin: Question: Should I check this in padding kernel as well?

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
```diff
@@ -366,12 +366,12 @@ class PopulateParamsBase {
...
-    if (dataType.isF32()) {
-      vectorizationSize = 4;
-    } else if (dataType.isF16() || dataType.isBF16()) {
-      // FIXME: figure out the best vectorization length for f16 and bf16.
-      vectorizationSize = 4;
-    }
+    unsigned dataWidth = dataType.getIntOrFloatBitWidth();
+    // 128 is the upper limit we support in vectorized load/store, which could
+    // be 4 fp32, 8 fp16, or 16 int8
+    const auto highestPotentialVectorizationLen = 128;
```
krzysz00: Can this be an actual concrete datatype, maybe `const size_t`?

Heck, let's make it `constexpr` while we're here

krzysz00 (commented):
I'll leave substantial review to someone who's worked with this stuff recently, but if it doesn't break anything it's _probably_ fine

Also, if we've been having cases where we page fault but then the error message gets swallowed ... that could be where our weird node hangs come from?

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
jerryyin: I made it `const size_t`

I don't see much value in addition to const in this context so prefer to just use const instead constexpr. I'd be declaring a constexpr if we use an expression to derive the value of `highestPotentialVectorizationLen`, and that we know for sure this can be done at compile time.

mlir/lib/Dialect/MIOpen/Tuning/GridwiseGemmParams.cpp
jerryyin: @whchung clarified that padding kernel will not use kpack feature.

Therefore, it is safe to skip this check in padding kernel.

jerryyin:
Note: Per jenkins log [here](http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-539/4/pipeline/124), we have a full pass in shared library round. This guarantees the PR will not incur regression to existing data types.

external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
```diff
@@ -315,6 +320,7 @@ struct MubufLoadOpLowering : ConvertToLLVMPattern {
             VectorType::get(interimShape, rewriter.getF32Type());
       Type interimLLVMResultType =
           typeConverter->convertType(interimResultType);
+      interimResultType.dump();
```
jerryyin: TODO: cleanup

jerryyin:
@whchung Gentle ping, this PR has passed both PR and nightly testing. I believe you are the only one that can review this now as @krzysz00 is out this week.

Do not squash if you decide to merge as there's a mixture of internal and external repo changes.

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
```diff
@@ -816,15 +815,18 @@ class PopulateParamsXDL : public PopulateParamsBase {
...
-      // TODO remove
-      {64, 64, 1, 32, 32, 1, false, false},
-
-      {64, 64, 1, 32, 32, 16, false, false},
-      {64, 64, 2, 32, 32, 16, false, false},
-      {64, 64, 4, 32, 32, 16, false, false},
-
-      // TODO Amend all kPack to be 16 for i8
-      {16, 16, 16, 16, 16, 1, false, false},
+      // TODO enable kpack
```
whchung, line 818: Do we need to change the heuristic values for non-XDLOPS path? Or are we concerned about XDLOPS path for now?

jerryyin: Good point, actually non-xdlops doesn't need to change because those tuning parameters are tailored for the two xdlops instructions for i8 only (and non-xdlops doesn't have this issue). M/N per block must be either 32 x 32 or 16 x 16 to be able to work.

On the other hand non-xdlops nightly is already all-passing with the existing tuning parameters.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
```diff
@@ -246,18 +246,51 @@ struct BufferLoadRewritePattern : public OpRewritePattern<BufferLoadOp> {
...
+        elementType = loadedVectorType.getElementType();
+        vectorLength = loadedVectorType.getShape()[0];
+
+        auto loadWidth = vectorLength * elementType.getIntOrFloatBitWidth();
+        bool isTooNarrow = loadWidth < 32;
+        bool isTooWide = loadWidth > 4 * 32;
+
+        // Use scalar load when load width is too narrow or too wide
+        // for a mubuf instruction
+        if (isTooNarrow || isTooWide) {
```
whchung, line 272: Is there any test case in the existing test suite to cover this code path?

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
```diff
@@ -246,18 +246,51 @@ struct BufferLoadRewritePattern : public OpRewritePattern<BufferLoadOp> {
...
-        // use buffer load since the source memref is on address space 0
-        SmallVector<Value, 4> srcLowerIndicesI32;
-        for (auto v : loadCoords)
-          srcLowerIndicesI32.push_back(
-              b.create<IndexCastOp>(loc, b.getIntegerType(32), v));
-        loadedValue = b.create<gpu::MubufLoadOp>(loc, loadedType, source,
-                                                 srcLowerIndicesI32);
+
+      Type elementType = loadedType;
+      int64_t vectorLength = 1;
```
whchung: These two lines seems to be used only in `if (loadedType.isa<VectorType>)` path.

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
```diff
@@ -966,6 +968,8 @@ class PopulateParamsXDL : public PopulateParamsBase {
     if ((param.gemmNPerBlock % param.gemmNPerWave) != 0)
       return failure();
 
+    // TODO Remove. Note KPerBlock and KPack are independent tuning parameters.
+    // There's no need to check if they are divide exactly
```
whchung, line 972: I think we have some discussion before that this check can be removed? Is it not the case after empirical testing?

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
jerryyin: @whchung Yes, with the tuning parameters in this PR, if this chunk of code doesn't exist, it would directly yield into a lowering failure that Mubuf instruction on 2 i8 is wrong for a number of resnet50 configs. (And now all resnet50 configs are passing)

Below config is how I used to test and bring up:

> ./bin/miopen-gen -batchsize=64 -in_channels=128 -in_h=16 -in_w=16 -out_channels=128 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=1 --padding_w=1 -t i8 -x2 --operation conv2d -pv | ./bin/mlir-miopen-driver -c | ./bin/mlir-rocm-runner --shared-libs=./lib/librocm-runtime-wrappers.so,./external/llvm-project/llvm/lib/libmlir_runner_utils.so --entry-point-result=void

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
```diff
@@ -246,18 +246,51 @@ struct BufferLoadRewritePattern : public OpRewritePattern<BufferLoadOp> {
...
+        VectorType loadedVectorType = loadedType.cast<VectorType>();
+        elementType = loadedVectorType.getElementType();
+        vectorLength = loadedVectorType.getShape()[0];
+
+        auto loadWidth = vectorLength * elementType.getIntOrFloatBitWidth();
+        bool isTooNarrow = loadWidth < 32;
+        bool isTooWide = loadWidth > 4 * 32;
+
+        // Use scalar load when load width is too narrow or too wide
+        // for a mubuf instruction
```
whchung, line 271: Would it be better to hide this logic in the lowering logic for `gpu.mubuf_load`?

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
jerryyin: Make sense will address

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
jerryyin: Correct this needs to remove. Removing this would allow more tuning parameters to run. For i8 this is fine, however, once I turn this check off it will result in a crash in fp16 condition. I'll file a new ticket for the context.

whchung (commented):
Please consider whether the added logic in `emitLoadInstruction` lambda better suited inside the lowering logic for `gpu.mubuf_load`. Or perhaps use `miopen.in_bounds_load` to reuse operators we have already defined. I'm open for discussion.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
jerryyin: I'm assuming `MubufLoadOp` is just a wrapper around the corresponding llvm intrinsic, do we intend to extend those gpu ops for them to do more?

At this case, `MubufLoadOp` can allow i8 type but with minimum `<4xi8>`. I think having `MubufLoadOp` to lower back to `memref::load` would make gpu ops' dependency looks strange (theoretically speaking gpu ops should not dependent on a "high level op" like memref op because it behave almost at llvm level.)

whchung: Actually, how about use `miopen.in_bounds_load` here? It was added recently along with `miopen.transforming_for`. The logic here actually looks very identical with what `miopen.in_bounds_load` provides.

jerryyin: Looks promising. Let me take a look at it. Thanks for the suggestion. I have no idea it exists.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
jerryyin: Done in [d66a280](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/539/commits/d66a280e92596aa61d7d49e43c1a507880e8a74e)

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
jerryyin: Swapping to in bound load does make the entire chunk of code much more succinct. We need more of those ops that can make our lives easier.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
```diff
@@ -246,18 +254,34 @@ struct BufferLoadRewritePattern : public OpRewritePattern<BufferLoadOp> {
         oobDims.push_back(pair.index());
     bool toEmitOobChecks = !oobDims.empty();
 
-    auto emitLoadInstruction = [&b, loc, loadedType,
+    auto emitLoadInstruction = [&b, loc, loadedType, sourceType,
```
whchung: We can get rid of `sourceType` now?

jerryyin: Oops you are right, thanks for the reminder

whchung (approved):

external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
```diff
@@ -295,18 +295,23 @@ struct MubufLoadOpLowering : ConvertToLLVMPattern {
...
+      int64_t numFit32 = 32 / vectorResultType.getElementTypeBitWidth();
+      int64_t lastDimShape = vectorShape.back() / numFit32;
+
       SmallVector<int64_t, 1> interimShape;
       for (unsigned iter = 0; iter < vectorShape.size() - 1; ++iter)
         interimShape.push_back(vectorShape[iter]);
-      interimShape.push_back(vectorShape[vectorShape.size() - 1] >> 1);
-      bool useScalarF32 = (vectorShape.size() == 1) && (vectorShape[0] == 2);
+      interimShape.push_back(lastDimShape);
+      bool useScalarF32 =
```
krzysz00, line 312: Minor nit: since there's bitcasts and such involved, could this be an `i32` load instead of an `f32` load? It'll make reading the resulting IR clearer.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
```diff
@@ -249,15 +257,31 @@ struct BufferLoadRewritePattern : public OpRewritePattern<BufferLoadOp> {
...
+      if (loadedType.isa<VectorType>()) {
+        VectorType loadedVectorType = loadedType.cast<VectorType>();
+        Type elementType = loadedVectorType.getElementType();
+        int64_t vectorLength = loadedVectorType.getShape()[0];
+
+        auto loadWidth = vectorLength * elementType.getIntOrFloatBitWidth();
+        bool isTooNarrow = loadWidth < 32;
+        bool isTooWide = loadWidth > 4 * 32;
+
+        // Use scalar load when load width is too narrow or too wide
```
krzysz00, line 270: Can we get a test for this in the fixup?

krzysz00 (commented):
I'd like to duly throw some minor nitpicks at this PR

external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
jerryyin: Filed https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/476 to track. I remember that only f32 intrinsic get exposed but will double check later.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
jerryyin: https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/476 tracked.

`isTooNarrow` are already tracked by new tuning parameter generated code from resnet50 e2e tests. `isTooWide` may never hit because of the input/output type of this op. My take away here is to come up with some infra beyond pass that can encode 32 and 32 * 4 correctly so it doesn't become one of those majic numbers.

krzysz00: I more meant adding to the unit tests for `buffer_load` and `buffer_store` to make sure calling for something too narrow works correctly.

external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
jerryyin: I'm starting to executing ROCmSoftwarePlatform/llvm-project-private#476.

Looking at where it is today, I believe:
 - This has moved to AMDGPU -> ROCDL lowering: `RawBufferOpLowering`
 - You have properly handle that in `llvmBufferValType` in treating fp16/i8 as a i32 type

I think there's nothing to do here. Please correct me if I'm wrong.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
jerryyin: I think again you have finished the work in `amdgpu-to-rocdl.mlir` and there's nothing left to do. Correct me if I'm wrong.

external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
krzysz00: Yeah, I think I did this implicitly.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
krzysz00: Yeah, this got fixed in the rewrite, near as I can tell

PR #540, opened 2022-03-03
[DO NOT SQUASH] Warnings cleanup

krzysz00:
This one's exactly what it says it is.

`skip-ci` has been applied so that we can get to merging it after some of the more substantial stuff goes in - we only have so many CI nodes at the moment, after all.

### DIFF
--- a/mlir/lib/CAPI/Dialect/MIGraphX.cpp
+++ b/mlir/lib/CAPI/Dialect/MIGraphX.cpp
@@ -57,7 +57,7 @@ void mlirGetKernelInfo(MlirModule module, int *size, void *data) {
     for (int i = 0; i < argSize; i++)
       argData[i + 1] = info[i];
     char *nameData = (char *)(argData + argSize + 1);
-    for (int i = 0; i < kernelName.size(); i++) {
+    for (size_t i = 0, e = kernelName.size(); i < e; ++i) {
       nameData[i] = kernelName[i];
     }
   }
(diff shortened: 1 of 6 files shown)


mlir/lib/CAPI/Dialect/MIGraphX.cpp
```diff
@@ -57,7 +57,7 @@ void mlirGetKernelInfo(MlirModule module, int *size, void *data) {
     for (int i = 0; i < argSize; i++)
       argData[i + 1] = info[i];
     char *nameData = (char *)(argData + argSize + 1);
-    for (int i = 0; i < kernelName.size(); i++) {
```
jungpark-mlir, line 60: I suppose size() only returns constant value and compiler can optimize this. good to have the correct type, thanks.

mlir/test/CAPI/mixr_cobj.cpp
```diff
@@ -202,11 +200,11 @@ static bool constructAndTraverseIr(MlirContext ctx) {
                        mlir::PassManager::Nesting::Implicit);
 
   mlir::migraphx::addHighLevelPipeline(pm);
-  pm.run(module);
+  (void)pm.run(module);
```
jungpark-mlir, line 203: why do we need this?

mlir/lib/CAPI/Dialect/MIGraphX.cpp
krzysz00: Yeah. I've seen the `i = 0, e = ...; i < e` style in a bunch of LLVM code and I'm assuming there's a reason for it

mlir/test/CAPI/mixr_cobj.cpp
krzysz00: `-Wunused-result` is the setting, IIRC. `LogicalResult` has a NODISCARD on it, so the compiler will yell if you don't check it or explicitly point out you're not checking it with something like that `(void)`

jungpark-mlir: Right, i'd have better consumed it. (void) should be fine for now.

jungpark-mlir (approved):

PR #514, opened 2022-02-16
[AMDGPU] Revert hack on AGPR allocation ordering.

whchung:
Revert a hack previously instilled on AGPR allocation ordering. Use CI to monitor whether we gain or lose performance.

PR #541, opened 2022-03-04
Fix clang-format in various codes checked in recently.

whchung:
These files are causing other outstanding PRs from completing CI runs.

### DIFF
--- a/mlir/lib/CAPI/Dialect/MIGraphX.cpp
+++ b/mlir/lib/CAPI/Dialect/MIGraphX.cpp
@@ -33,8 +33,9 @@ void mlirGetKernelInfo(MlirModule module, int *size, void *data) {
   int argIdx = 0;
   llvm::StringRef kernelName;
 
-  assert(size != nullptr ||
-         data != nullptr); // either of ptrs should be provided.
+  // Either of pointers should be provided.
+  assert((size != nullptr || data != nullptr) &&
+         "Either size or data pointer should be provided");
   std::vector<int> info;
   mod.walk([&](mlir::FuncOp f) {
     auto args = f.getArguments();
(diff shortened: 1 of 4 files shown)


krzysz00:
Isn't this fix for this to merge/rebase miopen-dialect so that these lines don't show up in the diff? They're presumably formatted correctly overall, otherwise they couldn't've been checked in themselves.

Unless this is a separate issue.

whchung:
I was trying to remove `skip-ci` label on #514, and think these changes deserves to be in a separate PR, to keep #514 succinct. As you can see in the files changed here, they are indeed not formatted correctly. I have no idea why they went in undetected.

mlir/lib/CAPI/Dialect/MIGraphX.cpp
```diff
@@ -33,7 +33,8 @@ void mlirGetKernelInfo(MlirModule module, int *size, void *data) {
   int argIdx = 0;
   llvm::StringRef kernelName;
 
-  assert (size != nullptr || data != nullptr); // either of ptrs should be provided.
+  assert(size != nullptr ||
```
krzysz00: Can we go ahead and make this a noisy assert? `assert((size != nullptr || data != nullptr) && "Either size or data pointer should be provided)`

krzysz00 (commented):
Yeah, very weird that these didn't get caught

One nitpick here, then I'm willing to toss these in CI-less

mlir/lib/CAPI/Dialect/MIGraphX.cpp
whchung: I'm merely fixing `clang-format` here. To improve the quality of the assert I think it's up to the owner of the module, @jungpark-mlir ?

jungpark-mlir: Yeah, that change (embedding comment) looks better

jungpark-mlir (approved):
Thanks for fixing this.

krzysz00 (approved):

PR #542, opened 2022-03-04
Add non-xdlops i8 GEMM and enable it for verification of xdlops gemm

krzysz00:


krzysz00:
Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/447

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -1767,9 +1767,10 @@ populateHostHarnessLogic(ModuleOp &module,
       // use non-xdlops kernels to verify xdlops kernels
       if (genConfig.xdlops)
         conv2dGenerator.flipXdlops();
-
-      // use f32 data type to verify non-f32 or xdlops f32 kernels
-      conv2dGenerator.setDataType("f32");
+      if (!genConfig.xdlops || genConfig.dataTypeStr != "i8")
+        // use f32 data type to verify non-f32 or xdlops f32 kernels
+        // except that i8 xdlops is verified with i8 non-xdlops
+        conv2dGenerator.setDataType("f32");
```
jerryyin, line 1773: Is this up-to-date? I feel like I just merged @yiqian1 PR who use CPU for validation only.

jerryyin (commented):
Overall looks good. However I'm paranoid with directly using non-xdlops for baseline before we fully verified it. Is there a way to test non-xdlops algorithm with CPU first before setting this as the baseline?

mlir/tools/miopen-gen/miopen-gen.cpp
krzysz00: Yeah, this is what's in `miopen-dialect`

krzysz00:
@jerryyin I'll note that the non-xdlops I8 implementation gets verified against the CPU code in non-xdlops CI (and manually)

jerryyin (approved):
LGTM

PR #543, opened 2022-03-04
Show more comparison results in the performance report

yiqian1:
Add MIOpen with tuned MLIR kernels vs. MIOpen without MLIR kernels in the performance report.
Combine two perfTest stages into one to avoid potential impact of different nodes on performance.

### DIFF
--- a/mlir/utils/jenkins/reportUtils.py
+++ b/mlir/utils/jenkins/reportUtils.py
@@ -41,7 +41,7 @@ def setCommonStyles(styler: 'pd.io.formats.style.Styler', speedupCols: list):
         {'selector': 'tbody tr:nth-child(odd)', 'props': [('background-color', '#e0e0e0')]},
         {'selector': 'tbody tr:nth-child(even)', 'props': [('background-color', '#eeeeee')]},
         {'selector': 'table', 'props': [('background-color', '#dddddd'), ('border-collapse', 'collapse')]},
-        {'selector': 'th, td', 'props': [('padding', '0.5em'), ('text-align', 'center')]}])
+        {'selector': 'th, td', 'props': [('padding', '0.5em'), ('text-align', 'center'), ('max-width', '150px')]}])
     styler.set_precision(ROUND_DIGITS)
     styler.set_na_rep("FAILED")
     for col in speedupCols:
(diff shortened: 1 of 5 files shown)


mlir/utils/jenkins/reportUtils.py
```diff
@@ -41,7 +41,7 @@ def setCommonStyles(styler: 'pd.io.formats.style.Styler', speedupCols: list):
         {'selector': 'tbody tr:nth-child(odd)', 'props': [('background-color', '#e0e0e0')]},
         {'selector': 'tbody tr:nth-child(even)', 'props': [('background-color', '#eeeeee')]},
         {'selector': 'table', 'props': [('background-color', '#dddddd'), ('border-collapse', 'collapse')]},
-        {'selector': 'th, td', 'props': [('padding', '0.5em'), ('text-align', 'center')]}])
+        {'selector': 'th, td', 'props': [('padding', '0.5em'), ('text-align', 'center'), ('max-width', '150px')]}])
```
krzysz00, line 44: Why?

krzysz00 (commented):
Huh, looks like parts of #538 ended up in here since there was a separate reason to do them.

I do have one question: how sure are we that when we say "Tuned MLIR kernels", that these are actually the MLIR kernels? Does the `MIOpenDriver.py` script set the `MIOPEN_DEBUG_ONLY_SOLVER` environment variable when doing those tests?

mlir/utils/jenkins/reportUtils.py
yiqian1: To wrap up long column names.

This is how it would look like:
http://ml-ci.amd.com:21096/job/MLIR/job/test-perf/Performance_20report/MLIR_vs_MIOpen.html

yiqian1:
> Huh, looks like parts of #538 ended up in here since there was a separate reason to do them.
> 
> I do have one question: how sure are we that when we say "Tuned MLIR kernels", that these are actually the MLIR kernels? Does the `MIOpenDriver.py` script set the `MIOPEN_DEBUG_ONLY_SOLVER` environment variable when doing those tests?

Yes. MIOPEN_DEBUG_FIND_ONLY_SOLVER are set in MIOpenDriver.py. I also manually checked to ensure MLIR solvers were invoked when I added the tests.

mlir/utils/jenkins/reportUtils.py
krzysz00: Seems reasonable to me, approving

krzysz00 (approved):

PR #545, opened 2022-03-08
Fix pattern rewrite paradigm to allow error to propagate

### LINKED ISSUE #1449: MLIR kernel fail to run while tuning
This is found while debugging [SWDEV-316705](https://ontrack-internal.amd.com/browse/SWDEV-316705)

on gfx908, latest develop branch, rocm-5.1

`MIOPEN_FIND_MODE=1  MIOPEN_FIND_ENFORCE=4   ./MIOpenDriver  conv -n 16 -c 186 -H 16 -W 16 -k 191 -y 3 -x 3 -p 1 -q 1 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -F 2 -t 1      2>&1 | tee log.txt`

This will fail in following solver:
after doing `export MIOPEN_DEBUG_CONV_MLIR_IGEMM_BWD_XDLOPS=0` before above command, this failure is gone.
```
MIOpenDriver conv -n 16 -c 186 -H 16 -W 16 -k 191 -y 3 -x 3 -p 1 -q 1 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -F 2 -t 1
MIOpen(HIP): Warning [FindSolutionImpl] Perf Db: load skipped: ConvOclDirectFwd, enforce: SEARCH_DB_UPDATE(4)
MIOpen(HIP): Warning [SearchImpl] Searching the best solution in the 9 dim space. Please, be patient...
MIOpen(HIP): Warning [SearchImpl] Runs left: 863, min time so far: 0.926557, curr time: 0.926557 16,16,16,16,1,1,1,1,1
MIOpen(HIP): Warning [SearchImpl] Runs left: 838, min time so far: 0.162239, curr time: 0.925437 16,8,16,16,1,2,1,1,2
MIOpen(HIP): Warning [SearchImpl] Runs left: 812, min time so far: 0.162239, curr time: 0.18672 8,16,16,16,2,1,1,2,1
MIOpen(HIP): Warning [SearchImpl] Runs left: 787, min time so far: 0.162239, curr time: 0.185919 8,8,16,16,2,2,1,2,2
MIOpen(HIP): Warning [SearchImpl] Default run, min time so far: 0.162239, default time: 0.550558 8,8,16,16,2,2,8,2,1
MIOpen(HIP): Warning [SearchImpl] ...Score: 3.3935
can't support backward data padding kernel when xdlops stride 2 pad_h,pad_w>0 and pad gemmM or gemmN due to store issue
MIOpen(HIP): Warning [FindSolutionImpl] Perf Db: load skipped: ConvMlirIgemmBwdXdlops, enforce: SEARCH_DB_UPDATE(4)
can't support backward data padding kernel when xdlops stride 2 pad_h,pad_w>0 and pad gemmM or gemmN due to store issue
can't support backward data padding kernel when both pad gemmN and gemmK due to load issue
can't support backward data padding kernel when both pad gemmN and gemmK due to load issue
can't support backward data padding kernel when both pad gemmN and gemmK due to load issue
MIOpen(HIP): Warning [GenericSearch] ConvMlirIgemmBwdXdlops: Searching the best solution among 0 (spare)...
MIOpen(HIP): Warning [GenericSearch] Done: 0/0/0, best #0 3.40282e+38 -1,-1,-1,-1,-1,-1,0,0
MIOpen(HIP): Error [FindSolutionImpl] Search failed for: ConvMlirIgemmBwdXdlops: /dockerx/repo/MIOpen/src/include/miopen/generic_search.hpp:475: Search failed
can't support backward data padding kernel when xdlops stride 2 pad_h,pad_w>0 and pad gemmM or gemmN due to store issue
can't support backward data padding kernel when both pad gemmN and gemmK due to load issue
can't support backward data padding kernel when both pad gemmN and gemmK due to load issue
can't support backward data padding kernel when both pad gemmN and gemmK due to load issue
can't support backward data padding kernel when xdlops stride 2 pad_h,pad_w>0 and pad gemmM or gemmN due to store issue
error: 'miopen.conv2d_bwd_data' op operand #0 must be 5D memref of 32-bit float or 16-bit float or 16-bit signless integer values, but got '!llvm.struct<(ptr<f32>, ptr<f32>, i32, array<5 x i32>, array<5 x i32>)>'
can't support backward data padding kernel when both pad gemmN and gemmK due to load issue
terminate called after throwing an instance of 'miopen::Exception'
  what():  /dockerx/repo/MIOpen/src/hipoc/hipoc_program.cpp:300: Code object build failed. Source: mlir_gen_igemm_conv2d_v4r1_bwd_xdlops0.mlir
```

jerryyin:
I have tested this locally and confirm it will fix https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1449. This PR is an urgency blocker, please review ASAP.

I corrected the common paradigm for pattern rewriting in this PR.

`applyPatternsAndFoldGreedily()`
 - This is inappropriate rewriting pattern for our case
 - It applies to lowering patterns that needs to be repeatedly applied and check whether it converged
 - It ignored any lowering failure from any lowering pattern
 - From API end, it won't be able to generate an error code even if the lowering pattern failed

Instead

`applyPartialConversion()`
 - Allow us to setup a success metric beforehand
 - Make sure target IR was what we expected

### DIFF
--- a/mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
+++ b/mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
@@ -2645,16 +2645,12 @@ struct GridwiseGemmV2RewritePattern
 void LowerMIOpenOpsStep2Pass::runOnOperation() {
   MLIRContext *ctx = &getContext();
   ConversionTarget target(*ctx);
-  target.addIllegalOp<miopen::GridwiseGemmOp>();
-  target.addIllegalOp<miopen::GridwiseGemmV2Op>();
-
-  target.addLegalDialect<arith::ArithmeticDialect>();
-  target.addLegalDialect<miopen::MIOpenDialect>();
-  target.addLegalDialect<AffineDialect>();
+  target.addIllegalOp<miopen::GridwiseGemmOp, miopen::GridwiseGemmV2Op>();
+  target.addLegalDialect<arith::ArithmeticDialect, miopen::MIOpenDialect,
+                         AffineDialect>();
 
   RewritePatternSet patterns(ctx);
-  patterns.add<GridwiseGemmRewritePattern>(ctx);
-  patterns.add<GridwiseGemmV2RewritePattern>(ctx);
+  patterns.add<GridwiseGemmRewritePattern, GridwiseGemmV2RewritePattern>(ctx);
   if (failed(applyPartialConversion(getOperation(), target,
                                     std::move(patterns)))) {
     signalPassFailure();
--- a/mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
+++ b/mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
@@ -27,7 +27,7 @@
 #include "mlir/Dialect/MIOpen/utility/builderUtils.h"
 
 #include "mlir/Support/LogicalResult.h"
-#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "mlir/Transforms/DialectConversion.h"
 
 #include "llvm/ADT/SmallVector.h"
 
@@ -1212,7 +1212,7 @@ struct GridwiseGemmRewritePattern : public OpRewritePattern<GridwiseGemmOp> {
         matrixCThreadwiseCopyDestCoords);
     affixThreadwiseCopyAttributes(threadwiseCopyCMatrixOp, op, b);
 
-    op.erase();
+    b.eraseOp(op);
 
     return success();
   }
@@ -2636,7 +2636,7 @@ struct GridwiseGemmV2RewritePattern
                                       enableOutSwizzles);
     }
 
-    op.erase();
+    b.eraseOp(op);
 
     return success();
   }
@@ -2644,10 +2644,21 @@ struct GridwiseGemmV2RewritePattern
 
 void LowerMIOpenOpsStep2Pass::runOnOperation() {
   MLIRContext *ctx = &getContext();
+  ConversionTarget target(*ctx);
+  target.addIllegalOp<miopen::GridwiseGemmOp>();
+  target.addIllegalOp<miopen::GridwiseGemmV2Op>();
+
+  target.addLegalDialect<arith::ArithmeticDialect>();
+  target.addLegalDialect<miopen::MIOpenDialect>();
+  target.addLegalDialect<AffineDialect>();
+
   RewritePatternSet patterns(ctx);
-  patterns.add<GridwiseGemmRewritePattern, GridwiseGemmV2RewritePattern>(ctx);
-  if (failed(applyPatternsAndFoldGreedily(getOperation(), std::move(patterns))))
+  patterns.add<GridwiseGemmRewritePattern>(ctx);
+  patterns.add<GridwiseGemmV2RewritePattern>(ctx);
+  if (failed(applyPartialConversion(getOperation(), target,
+                                    std::move(patterns)))) {
     signalPassFailure();
+  }
 }
 } // end anonymous namespace
 
--- a/mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
+++ b/mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
@@ -47,10 +47,13 @@ private:
     static MLIRContext context(getRegistry());
     static std::once_flag once;
     std::call_once(once, []() {
-      // FIXME: A conv op would ignore this context and instead use the
-      // context from the UnknownLoc
+      // Turn off all diagnotic printing on op and stacktrace
+      // Note: This is not necessary with below handler
       context.printOpOnDiagnostic(false);
       context.printStackTraceOnDiagnostic(false);
+      // Register a handler that swallows all diagnostic print
+      DiagnosticEngine &engine = context.getDiagEngine();
+      engine.registerHandler([](Diagnostic &diag) {});
       context.loadDialect<miopen::MIOpenDialect, StandardOpsDialect>();
     });
     return context;
--- a/mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
+++ b/mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
@@ -47,6 +47,10 @@ private:
     static MLIRContext context(getRegistry());
     static std::once_flag once;
     std::call_once(once, []() {
+      // FIXME: A conv op would ignore this context and instead use the
+      // context from the UnknownLoc
+      context.printOpOnDiagnostic(false);
+      context.printStackTraceOnDiagnostic(false);
       context.loadDialect<miopen::MIOpenDialect, StandardOpsDialect>();
     });
     return context;
--- a/mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
+++ b/mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
@@ -1784,22 +1784,18 @@ void LowerMIOpenOpsStep1Pass::runOnOperation() {
   MLIRContext *ctx = &getContext();
 
   ConversionTarget target(*ctx);
-  target.addIllegalOp<miopen::Conv2DOp>();
-  target.addIllegalOp<miopen::Conv2DBwdDataOp>();
-  target.addIllegalOp<miopen::Conv2DBwdWeightOp>();
-
-  target.addLegalOp<miopen::TransformOp>();
-  target.addLegalOp<miopen::GridwiseGemmOp>();
-  target.addLegalOp<miopen::GridwiseGemmV2Op>();
+  target.addIllegalOp<miopen::Conv2DOp, miopen::Conv2DBwdDataOp,
+                      miopen::Conv2DBwdWeightOp>();
+  target.addLegalOp<miopen::TransformOp, miopen::GridwiseGemmOp,
+                    miopen::GridwiseGemmV2Op>();
   // Below are required legalize for the lowering of Conv2DBwdWeightOp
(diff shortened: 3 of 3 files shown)


mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
```diff
@@ -47,6 +47,10 @@ struct MiirHandle_s {
     static MLIRContext context(getRegistry());
     static std::once_flag once;
     std::call_once(once, []() {
+      // FIXME: A conv op would ignore this context and instead use the
+      // context from the UnknownLoc
+      context.printOpOnDiagnostic(false);
+      context.printStackTraceOnDiagnostic(false);
```
whchung, line 53: @jerryyin could you explain what the current symptom is with potentially different `MLIRContext` instance in use?

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2636,18 +2636,29 @@ struct GridwiseGemmV2RewritePattern
...
+
+  target.addLegalDialect<arith::ArithmeticDialect>();
+  target.addLegalDialect<miopen::MIOpenDialect>();
+  target.addLegalDialect<AffineDialect>();
+
   RewritePatternSet patterns(ctx);
-  patterns.add<GridwiseGemmRewritePattern, GridwiseGemmV2RewritePattern>(ctx);
-  if (failed(applyPatternsAndFoldGreedily(getOperation(), std::move(patterns))))
+  patterns.add<GridwiseGemmRewritePattern>(ctx);
+  patterns.add<GridwiseGemmV2RewritePattern>(ctx);
```
krzysz00: Why not keep the old `add` call? Also, does `addIllegalOp` take multiple arguments?

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2636,18 +2636,29 @@ struct GridwiseGemmV2RewritePattern
...
+  if (failed(applyPartialConversion(getOperation(), target,
```
whchung, line 2654: It seems we are only changing the first 2 steps. Do we need to change other lowering steps too?

krzysz00 (approved):
Approved on an emergency basis, the approach seems reasonable.

Followup PR using the `add<>()` API properly, like it was before, requested. Probably same for `addIllegalOperation()` and friends.

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
krzysz00: Other notes: there's one context per thread, and MLIR can spin up multiple threads to handle rewrites in parallel.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: Some of the lower stages actually do rely on the repeated-application/convergency property, like `step4` with `TransformingFor` and the current step3 with `TransformOpRewritePattern`.

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
jerryyin: `emitError` or `emitOpError` (that we use, or populate by a pattern failure) both shall end up in `Operation::emitError()` (refer to `mlir/lib/IR/Operation.cpp`) and should not populate an error message when `printOpOnDiagnostic()` is turned off however it still does even after I have the setting here.

Either I have misunderstood how `emitError()` work or this is a different bug somewhere. But either way the side effect is just print out a couple of lines of error message which is much less critical than not returned the error code.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Correct, we need to change all lowering steps. But if the later stages are dead for any reason, we are doomed anyways (Because we are not allowed to, after applicability checks allows a certain config to proceed.). So I'd prioritize fixing the first two passes that are actually used in applicability checks.

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
krzysz00: printOpOnDiagnostic() is, IIRC, something that controls whether the operation itself is printed next to the error.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: I don't think we need to change the change the rest of the passes

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Yes for both. I don't see much of the difference from callee perspective. Looking at the `RewritePatternSet` source code I feel like they are more or less equivalent with each other. Or is there any compelling reason to adopt the variadic version?

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: > Some of the lower stages actually do rely on the repeated-application/convergency property

Good to know if that's the case. I had no idea that they do so repeatedly.

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
jerryyin: @krzysz00 So maybe it does end up to work properly. This is in terms that it only yields an error message but didn't print the op. Thanks for the the comment

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: Are you open to fixing up the calls to add() and the like here or do you want to do those later?

jerryyin: @krzysz00 We can follow up the `add()` call in the other thread brought by you. But for this thread on identifying such scenarios for other passes, I'd not do it in this PR and do it later.

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
krzysz00: Yeah

I think what you want to do instead is to set up a diagnostic handler that swallows the error around the lowering calls that are in the library.

krzysz00: And then to get rid of this bit

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: Because that's the style preferred upstream

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
jerryyin: @krzysz00 Sounds like a follow up ticket I'll document: This handler should print out everything when used from our internal tool like `mlir-miopen-driver` but swallow the error when used by `mlir-miopen-lib`.

krzysz00: The solution is at https://mlir.llvm.org/docs/Diagnostics/

krzysz00: I'd like to see it applied now, especially if we're backporting to a release

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Can't argue with that :-p, addressed in 2d45f48702c046ebe4ffd5aed75ca288afda9157

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
jerryyin: @krzysz00 I'm not sure the commit would be backported. But yah, it sounds fair to fix the diagnostic in the same PR. Please un-approve to wait for my upcoming fix.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Done in [72323fd](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/545/commits/72323fdad1b523f61d1396f8fc875badab6c9aa1)

jerryyin: Done in 72323fd

mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp
jerryyin: Done in https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/commit/72323fdad1b523f61d1396f8fc875badab6c9aa1

krzysz00 (approved):
This all looks good

PR #546, opened 2022-03-09
Emit utility kernels to set 0 for backward data kernels when needed.

whchung:
Teach MLIR MIOpen dialect to emit utility kernels to set 0 and truncation for backward data kernels when needed.

A set 0 kernel will be invoked prior to the backward data convolution kernel to set the output (input tensor) be all 0 when needed. The heuristic to determine is borrowed from MIOpen solver.

In `miopen-gen`, we also eliminate the `isOut` option when we to random initialization of tensors on CPU. And in MIOpen MLIR solver, all dependency to `SetTensor` which depend on OpenCL GPU kernels are all be eliminated.


whchung:
The meaning and ordering of `kernel_id` / `gemm_id` needs to be consolidated between `miopen-gen`, backward data lowering path, backward weight lowering path, as it's apparent `-1` has different meaning in each path. Fixing e2e unit tests.

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
```diff
@@ -708,8 +696,21 @@ LogicalResult Conv2dGenerator::genConvModule(ModuleOp &module, int kernel_id,
...
+  // For backward data convolution, additional processing is needed below.
+  int64_t gemmId = kernel_id;
+
+  // Obtain gemm ID from kernel_id for backward data convolution.
+  if (config.operation.getValue() == miopen::ConvOpType::BwdData) {
+    llvm::SmallVector<int64_t> gemmIds = populateBackwardDataGemmIds(
+        config.strideHeight, config.strideWidth, config.dilationHeight,
+        config.dilationWidth, config.filterHeight, config.filterWidth);
+    assert(gemmIds.size() > static_cast<size_t>(kernel_id));
+    gemmId = gemmIds[kernel_id];
```
krzysz00, line 709: SmallVector already has bounds checks, you don't need your own assert.

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
```diff
@@ -120,6 +124,26 @@ AffixTuningParameters::fetchDimensions(T &op) {
   return std::make_tuple(y, x, ho, wo, hi, wi, k, c, n);
 }
 
+void AffixTuningParameters::affixBackwardDataUtilityKernels(
+    miopen::Conv2DBwdDataOp &op) {
+  auto gemmIdAttr = op->template getAttrOfType<IntegerAttr>("gemm_id");
+
+  // In case the gemm ID is -1, override grid_size and block_size for the
+  // utility kernel.
+  if (gemmIdAttr.getInt() < 0) {
```
krzysz00, line 133: Note to future us: we really should consider a different approach to this sort of thing - maybe a `KernelType` enum/attribute, or actually teaching the runtime and library there are operations we generate code for that aren't convolution, and so on.

Actually, yeah, instead of having "convolutions" that aren't, we really should have `miopen.zero_kernel` and `miopen.truncate_kernel` as top-level ops. But that's a bigger refactor and this needs to go through now so ... can you pull out a ticket?

krzysz00 (approved):
LGTM. I have long term concerns about how we've implemented this, but correctness bugs are correctness bugs.

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
whchung: https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/477

PR #547, opened 2022-03-10
Compute out of bounds info from coordinate transforms

krzysz00:
Will address https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/470

This is currently just the core method - it hasn't been integrated into anything, and hasn't even been tested meaningfully. That's going to be the lengthy and annoying part.

I just wanted to make sure the work that did exist was somewhere visible so you didn't have to go scrounging for it.

@sjw36 @jungpark-mlir for the progress update.


krzysz00:
Note: I still need to get rid of the old dead code, but the core of the PR is ready for review

krzysz00:
This PR is now ready for review and took less time than I thought it would

krzysz00:
Future note to me: the arithmetic range analysis might clear up some copies of the bounds check code.

krzysz00:
Also, do we know if the hardware has an opinion about having OOB be true vs. having in bounds be true? I'm thinking of things like branch predictors (even though there are no branches)

whchung:
@krzysz00 sorry for late. I'll address this PR today.

whchung (approved):
Thanks for coming up this generic framework to compute OOB dimensions!

PR #548, opened 2022-03-14
Improve utility kernels perf

whchung:
The utility kernels for backward weight convolutions were naive and performed badly. Use a better grid size / block size to increase parallelism in this PR.

Benchmark commmand:
```
./bin/miopen-gen --x2 --operation conv2d_bwd_weight -t f16 --batchsize 64 --in_channels 1024 --out_channels 1024 --in_h 14 --in_w 14 --out_h 14 --out_w 14 --fil_h 1 --fil_w 1 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h 0 --padding_w 0 --groupsize 1 -ph | ./bin/mlir-miopen-driver -c | rocprof --stats ./bin/mlir-rocm-runner --shared-libs=./lib/librocm-runtime-wrappers.so,./external/llvm-project/llvm/lib/libmlir_runner_utils.so --entry-point-result=void
```

This is a backward weight convolution which uses XDLOPS, in fp16, no padding along Gemm M/N/K. Therefore a 0-init kernel and a conversion kernel will be used.

at tip of `miopen-dialect`:

```
# cat results.stats.csv
"miopen_conv2d_bwd_weight_gkcyx_ngchw_ngkhw_0.kd",1,290640047,290640047,51.83518016940135
"miopen_conv2d_bwd_weight_gkcyx_ngchw_ngkhw_2.kd",1,264794271,264794271,47.2256280123375
"miopen_conv2d_bwd_weight_gkcyx_ngchw_ngkhw_1.kd",1,5266052,5266052,0.9391918182611508
```

With this PR:
```
# cat results.stats.csv
"miopen_conv2d_bwd_weight_gkcyx_ngchw_ngkhw_1.kd",1,5289100,5289100,99.4823779017496
"miopen_conv2d_bwd_weight_gkcyx_ngchw_ngkhw_2.kd",1,20960,20960,0.3942354352953568
"miopen_conv2d_bwd_weight_gkcyx_ngchw_ngkhw_0.kd",1,6560,6560,0.12338666295503534
```

`miopen_conv2d_bwd_weight_gkcyx_ngchw_ngkhw_1.kd` which is the actual backward convolution, originally takes only about 0.94% of execution time, now takes up 99.48%. It means the execution time of other 2 utility kernels have been greatly reduced via this PR.

This PR only covers backward weight convolution. I'll update backward data in #546 directly.

### DIFF
--- a/mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
+++ b/mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
@@ -24,6 +24,7 @@
 #include "mlir/Dialect/MIOpen/MIOpen.h"
 #include "mlir/Dialect/MIOpen/Passes.h"
 #include "mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h"
+#include "mlir/Dialect/MIOpen/Tuning/UtilityParams.h"
 #include "mlir/Dialect/MIOpen/utility/builderUtils.h"
 #include "mlir/Dialect/MIOpen/utility/loweringUtils.h"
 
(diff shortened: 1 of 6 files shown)


mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -224,13 +244,30 @@ LogicalResult elementwiseConversion(Conv2DBwdWeightOp op, PatternRewriter &b) {
...
+
+  auto loop = b.create<scf::ForOp>(
+      loc, offset, b.create<ConstantIndexOp>(loc, collapsedWorkspaceShape[0]),
+      stride);
   b.setInsertionPointToStart(loop.getBody());
-  auto loadedValue = b.create<AffineLoadOp>(loc, collapsedWorkspace, iv);
+  auto iv = loop.getInductionVar();
+  auto loadedValue = b.create<memref::LoadOp>(loc, collapsedWorkspace, iv);
   auto convertedValue =
       createTypeConversionOp(b, loc, loadedValue, filterDataType);
```
jerryyin: I think `elementwiseConversion` is much more generic than the `zeroInit`. I'd prefer both to use a more generic implementation such that the way we generate `convertedValue` is an argument where:
 - In `zeroInit`, the functor just set 0
 - In `elementwiseConversion`, the functor does type conversion

This also makes future elementwise kernels to be more straightforward to implement.

whchung: Pushed a new commit to abstract the elementwise emission loop.

jerryyin (commented):
LGTM, thank you!

jerryyin (approved):

PR #549, opened 2022-03-14
Enable non-xdlops int8 E2E tests

yiqian1:
I've run fixed E2E tests on lockhart6 and they passed.


jerryyin:
LGTM for this PR.

~~Though I'm a bit worried that after incorporating #539 things may break. Could give it a shot to make sure this would pass with #539 incorporated?~~

Scratch that, let's me amend xdlops tuning parameters in a different PR if there's a need. Right now looks like there doesn't need additional tuning parameters around non-xdlops.

jerryyin (approved):

PR #553, opened 2022-03-15
MLIR#468: use miopen.transform instead of memref.expand_shape in TosaToMIOpen

sjw36:
Since the conversion of tosa.conv2d will generate an miopen.conv2d which requires 5D tensor parameters, we need to expand the shape of the tosa.conv2d parameters by adding a 'G; dimension of size 1. Currently this is accomplished with a memref.expand_shape operation but this causes problems if we need to insert an miopen.transform<broadcast> for bias-add. The chain of miopen.transforms cannot be broken by memref operations (as currently implemented).

This change will apply only to tosa.conv2d and tosa.matmul operation parameters.

See: https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/468

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/MIOpen.h
+++ b/mlir/include/mlir/Dialect/MIOpen/MIOpen.h
@@ -260,10 +260,9 @@ public:
   void addDim(StringRef name, uint32_t dim, int64_t size);
 
   void expand(StringRef name, uint32_t dim, int64_t size);
-  void expand(ArrayRef<StringRef> names,
-              ArrayRef<uint32_t> dims,
+  void expand(ArrayRef<StringRef> names, ArrayRef<uint32_t> dims,
               ArrayRef<int64_t> sizes);
-  
+
   void slice(ArrayRef<StringRef> upperNames, ArrayRef<StringRef> lowerNames,
              ArrayRef<int64_t> begins, ArrayRef<int64_t> ends);
 
--- a/mlir/include/mlir/Dialect/MIOpen/MIOpen.h
+++ b/mlir/include/mlir/Dialect/MIOpen/MIOpen.h
@@ -111,8 +111,8 @@ public:
   // The names live as long as the CoordTransformBuilder
   void getEndNames(SmallVectorImpl<StringRef> &names);
 
-  SmallString<8> startName(uint32_t dim);
-  SmallString<8> endName(uint32_t dim);
+  StringRef startName(uint32_t dim);
+  StringRef endName(uint32_t dim);
   uint32_t startIndex(StringRef name);
   uint32_t endIndex(StringRef name);
 
@@ -259,6 +259,11 @@ public:
   // Defines a dimension that is not mapped to any coordinates in the output
   void addDim(StringRef name, uint32_t dim, int64_t size);
 
+  void expand(StringRef name, uint32_t dim, int64_t size);
+  void expand(ArrayRef<StringRef> names,
+              ArrayRef<uint32_t> dims,
+              ArrayRef<int64_t> sizes);
+  
   void slice(ArrayRef<StringRef> upperNames, ArrayRef<StringRef> lowerNames,
              ArrayRef<int64_t> begins, ArrayRef<int64_t> ends);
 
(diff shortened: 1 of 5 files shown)


sjw36:
This change requires @krzysz00 fix for OOB checks in miopen.transform.

mlir/include/mlir/Dialect/MIOpen/MIOpen.h
```diff
@@ -111,8 +111,8 @@ class CoordTransformsBuilder {
   // The names live as long as the CoordTransformBuilder
   void getEndNames(SmallVectorImpl<StringRef> &names);
 
-  SmallString<8> startName(uint32_t dim);
-  SmallString<8> endName(uint32_t dim);
+  StringRef startName(uint32_t dim);
```
krzysz00, line 114: No, that's not acceptable, since we can't guarantee the lifetime of the returned string within the builder.

mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
```diff
@@ -76,30 +76,12 @@ static Value expandMemRef(ConversionPatternRewriter &rw, Operation *op,
...
-    expShape.push_back(shape[dim]);
-    if (dim == idx) {
-      expShape.push_back(1);
-      reassocs.push_back(
-          {getAffineDimExpr(d0++, context), getAffineDimExpr(d0++, context)});
-    } else {
-      reassocs.push_back({getAffineDimExpr(d0++, context)});
-    }
-  }
+  SmallVector<StringRef, 8> names{"a", "b", "c", "d", "e", "f", "h", "i"};
```
krzysz00, line 79: ... just generate `dim0`, `dim1`, ...

mlir/lib/Dialect/MIOpen/CoordTransformBuilder.cpp
```diff
@@ -523,6 +523,35 @@ void BottomUpCTBuilder::addDim(StringRef name, uint32_t dim, int64_t size) {
   addTransform(TransformType::AddDim, {size}, {}, {}, {name}, {dim});
 }
 
+void BottomUpCTBuilder::expand(StringRef name, uint32_t dim, int64_t size) {
```
krzysz00, line 526: This isn't a method that goes here

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -78,31 +78,39 @@ template <typename T> struct MILARewritePattern : public OpRewritePattern<T> {
...
-  Value backtraceToAlloc(Value inp) const {
-    if (auto bt = inp.getDefiningOp<miopen::TransformOp>())
-      return backtraceToAlloc(bt->getOperand(0));
-    if (auto bt = inp.getDefiningOp<memref::ExpandShapeOp>())
-      return backtraceToAlloc(bt->getOperand(0));
-    if (auto bt = inp.getDefiningOp<memref::AllocOp>())
-      return inp;
+  typedef SmallVector<miopen::TransformOp, 4> TransformList;
+
+  Value bwTraceTo(Value inp, TransformList *transforms = nullptr) const {
```
krzysz00, line 83: This is `untransform()`

krzysz00: with some extra bits

krzysz00 (changes requested):
I've got significant complaints about how this PR was written.

mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
krzysz00: This violates zero-one-infinity in a very silly way for no good reason

mlir/lib/Dialect/MIOpen/CoordTransformBuilder.cpp
krzysz00: I mean, maybe it could be, but you're looking for moving `expandNamesInPlace` out of `ConvToGemm` and the top names wrapper

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
krzysz00: Update: it has somewhat different semantics and so can probably stay its own thing

krzysz00:
Closing because #564 went in

PR #556, opened 2022-03-16
Add a 'render' group in Dockerfile, and use it with --group-add in Je…

pcf000:
…nkinsfile, because

some CI hosts have /dev/kfd under that group instead of 'video'.

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -61,7 +61,7 @@ void showEnv() {
 }
 
 String dockerArgs() {
-    return "--device=/dev/kfd --device=/dev/dri --group-add video -v /etc/passwd:/etc/passwd:ro -v /etc/group:/etc/group:ro"
+    return "--device=/dev/kfd --device=/dev/dri --group-add video --group-add render -v /etc/passwd:/etc/passwd:ro -v /etc/group:/etc/group:ro"
 }
 
 String dockerImage() {


pcf000:
I think I'll have to split this into two, because passing "--group-add render" will barf unless it has a docker image with the render group.  First push the Dockerfile and let the new image propagate, then push the Jenkinsfile.  But I'll do that tomorrow.

jerryyin:
@pcf000 Correct, the first dockerfile change needs to go first. Then when docker finished building, we can submit the second PR.

mlir/utils/jenkins/Dockerfile.rocm
```diff
@@ -65,6 +65,9 @@ RUN ln -s /usr/bin/clang-10 /usr/bin/clang;\
 
 # --------------------- Section 2: MIOpen dialect setups -----------------
 
+# Need "render" group because some CI hosts have /dev/kfd under it.
+RUN groupadd -g 109 render
```
jerryyin: [Question] What does 109 mean in the groupadd command?

pcf000: It's the group ID.  The render group happens to be 109 on the machines I checked, and I wanted it to match in the image.

jerryyin: What if a different machine happen to have a different group id than 109? We may not want to target the fix for one specific machine.

pcf000: If it's really an ubuntu 20 thing, that shouldn't arise.  However, if we have to accommodate it, we'd need to add code to our image to run at startup time, that ascertains the group of /dev/kfd and adds the current user to it.  Something like
```
sudo usermod -aG `ls -n /dev/kfd | cut -d' ' -f4` $USER
```
assuming we can do sudo there, and possibly a newgrp or su, though I don't know how to add startup code to a docker invocation.

krzysz00 (approved):
LGTM - has the Dockerfile been rebuilt?

pcf000:
Looks like both rocm/mlir:latest and rocm/mlir:rocm5.0-latest images have the render group, thanks to the earlier pull request.  This one now changes only Jenkinsfile.  And validations *finally* finished just a few minutes ago, so I'm going to merge it.

PR #558, opened 2022-03-16
Refactor how dimension names / sizes are fetched in tuning and lowering pipeline.

whchung:
Refactor how dimension names / sizes are fetched in tuning and lowering pipeline.

Introduce a `ConvolutionOp` trait.
Steal the idea of `canonicalizeDims` currently in use in the generator, and teach `fetchDimensions` to return 3 maps instead of a list of scalar values. Use the maps in the tuning and lowering pipeline.

No attempt is made to consolidate data structures between tuning and generator logic yet. It would be addressed in separate PRs.

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/MIOpenOpTraits.h
+++ b/mlir/include/mlir/Dialect/MIOpen/MIOpenOpTraits.h
@@ -1,0 +1,31 @@
+//===- MIOpenOpTraits.h - MLIR MIOpen dialect operation traits --*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file declares C++ classes for some of operation traits in the MIOpen
+// dialect.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef MLIR_DIALECT_MIOPEN_OPTRAITS_H_
+#define MLIR_DIALECT_MIOPEN_OPTRAITS_H_
+
+#include "mlir/IR/OpDefinition.h"
+
+namespace mlir {
+namespace OpTrait {
+namespace miopen {
+
+/// An operation trait for convolution ops.
+template <typename ConcreteType>
+class ConvolutionOp : public TraitBase<ConcreteType, ConvolutionOp> {};
+
+} // namespace miopen
+} // namespace OpTrait
+} // namespace mlir
+
+#endif // MLIR_DIALECT_MIOPEN_OPTRAITS_H_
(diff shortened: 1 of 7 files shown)


whchung:
Fixing unit tests.

mlir/include/mlir/Dialect/MIOpen/MIOpenOpTraits.h
```diff
@@ -0,0 +1,31 @@
...
+
+#include "mlir/IR/OpDefinition.h"
+
+namespace mlir {
+namespace OpTrait {
+namespace miopen {
+
+/// An operation trait for convolution ops.
+template <typename ConcreteType>
+class ConvolutionOp : public TraitBase<ConcreteType, ConvolutionOp> {};
```
krzysz00, line 25: First off, I'm not sure we need the `OpTrait` namespace here.

Second, ... can't we tablegen this whole thing?

mlir/lib/Dialect/MIOpen/utility/loweringUtils.cpp
```diff
@@ -22,5 +23,46 @@ std::tuple<Value, ArrayAttr> untransform(OpBuilder &b, Value transformed,
   }
   return {ret, b.getArrayAttr(transformList)};
 }
+
+static llvm::StringMap<int64_t> canonicalizeDims(ArrayRef<int64_t> dims,
```
krzysz00, line 27: Should this be in a `namespace {}` instead of `static`ing it? I can't tell if there's a style preference on that.

krzysz00 (commented):
This seems like a quite nice refactoring, thank you for taking the time to do it!

mlir/include/mlir/Dialect/MIOpen/MIOpenOpTraits.h
whchung: To my understanding, we need a subclass of `TraitBase` be defined in C++.

mlir/lib/Dialect/MIOpen/utility/loweringUtils.cpp
whchung: I'll move them to `namespace {}` after fixing unit tests.

mlir/include/mlir/Dialect/MIOpen/MIOpenOpTraits.h
krzysz00: Nope!

Here's how I'm adding an `InferIntRangeInterface` to `arith` in some background work, minus header guards

`InferIntRangeInterface.h`
```c++
#include "mlir/IR/Attributes.h"
#include "mlir/IR/OpDefinition.h"
#include "mlir/IR/Operation.h"

#include "mlir/Dialect/Arithmetic/IR/InferIntRangeInterface.h.inc"
```

`InferIntRangeInterface.td`

```
include "mlir/IR/OpBase.td"

def InferIntRangeInterface : OpInterface<"InferIntRangeInterface"> {
    let description = [{ ... }];
    let cppNamespace = "::mlir::arith";
    let methods = [...];
}
```
CMakeLists in the include directory
```
add_mlir_interface(InferIntRangeInterface)
```

`InferIntRangeInterface.cpp`
```c++
#include "mlir/Dialect/Arithmetic/IR/InferIntRangeInterface.h"
#include "mlir/Dialect/Arithmetic/IR/InferIntRangeInterface.cpp.inc"
```

And that's it!

whchung: Thanks for the pointers. Let me revise this PR.

PR #559, opened 2022-03-16
NFC. Remove unused codes.

whchung:
This attribute is never used anywhere. Remove it.


jerryyin (approved):
Indeed, those are preivously used in c++ generator, and now become obsolete.

PR #561, opened 2022-03-17
Reorganize E2E tests and reduce batch size

yiqian1:
Reduce batch size to 64 for most E2E tests. 2 Resnet50 tests and 3 Resnext101 tests remain batch size of 256.

Put E2E tests of same data type in the same folder. Each file includes all combinations for one config.


krzysz00 (changes requested):
Can you also commit the config files used to generate these tests so we can regenerate them in the future?

krzysz00 (approved):

krzysz00:
I'm assuming all the existing FIXME's in the tests are either preserved in the configs or aren't actually a bug anymore.

PR #564, opened 2022-03-18
Fork of MLIR#468: use miopen.transform instead of memref.expand_shape in TosaToMIOpen

krzysz00:
Superscedes https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/553 temporarily since @sjw36 is on vacation.

Thanks to @jungpark-mlir , the unrelated bug in the fusion pass has been fixed and now this PR is good pending review from Simon and merge of #547 

---

**MLIR#468: use miopen.transform instead of memref.expand_shape in TosaToMIOpen**

Since the conversion of tosa.conv2d will generate an miopen.conv2d which requires 5D tensor parameters, we need to expand the shape of the tosa.conv2d parameters by adding a 'G; dimension of size 1. Currently this is accomplished with a memref.expand_shape operation but this causes problems if we need to insert an miopen.transform for bias-add. The chain of miopen.transforms cannot be broken by memref operations (as currently implemented).

This change will apply only to tosa.conv2d and tosa.matmul operation parameters.

See: https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/468

### DIFF
--- a/mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
+++ b/mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
@@ -66,41 +66,23 @@ static bool isConstantZero(Value v) {
 }
 
 static Value expandMemRef(ConversionPatternRewriter &rw, Operation *op,
-                          Value operand, uint32_t idx = 4) {
-  auto context = rw.getContext();
-  auto oprType = operand.getType().template cast<ShapedType>();
-  if (!oprType.hasStaticShape()) {
-    (void)rw.notifyMatchFailure(
-        op, "tosa to miopen conversion expects statically shaped tensors");
-    return Value();
-  }
-  auto shape = oprType.getShape();
-
-  idx--;
-
-  // expShape = shape + 1 dim
-  //  ex:  <NxHxWxCxf32> -> <Nx1xHxWxCxf32>
-  SmallVector<int64_t, 5> expShape;
-  // reassocs = shape w/ 1 split
-  //  ex:  [0,1,[2,3],4]
-  uint32_t d0 = 0;
-  SmallVector<ReassociationExprs, 5> reassocs;
-  for (uint32_t dim = 0; dim < shape.size(); ++dim) {
-    expShape.push_back(shape[dim]);
-    if (dim == idx) {
-      expShape.push_back(1);
-      reassocs.push_back(
-          {getAffineDimExpr(d0++, context), getAffineDimExpr(d0++, context)});
-    } else {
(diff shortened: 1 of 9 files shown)


sjw36:
Thanks @krzysz00 , my PR wasn't really ready for review except for Jungwook to see changes in AlignTiling. I will review this today.

mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
```diff
@@ -66,40 +68,43 @@ static bool isConstantZero(Value v) {
...
+    ("dim" + Twine(i)).toVector(dimName);
+    dimNames.emplace_back(std::move(dimName));
+    dimNameRefs.push_back(StringRef(dimNames[i]));
   }
 
-  auto newType = MemRefType::get(expShape, oprType.getElementType());
-  auto oprExpand = rw.create<memref::ExpandShapeOp>(op->getLoc(), newType,
-                                                    operand, reassocs);
-  return oprExpand;
+  miopen::BottomUpCTBuilder transform(rw, dimNameRefs, shape, loc);
```
sjw36, line 92: Would be nice to have a mode in CTBuilder to generate names like this. And since it owns the names, this would remove the copying.

sjw36 (approved):
Looks good. Thanks.

mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
sjw36: And would greatly simplify CTBuilder user code.

krzysz00: Ok, that is a good point and the pattern has now come up enough times that I see a solid argument for adding it. Followup PR?

krzysz00: On the other hand, the caller needs those names to be able to refer to the dimensions ... probably the right call here is getting or making a better source of long-lived strings - maybe something with attributes

sjw36: Not in this instance and not for my broadcast. Can they not be referenced positionally?

krzysz00: We could probably add the API for that, it just never came up before.

You'd still need to give the names of their equivalents in upper space, unless ... yeah, we could do a slightly wonky thing and just have `.passThrough(endPosition, startPosition)`

PR #563, opened 2022-03-18
Reduce `ConvolutionContext` usage.

whchung:
Reduce `ConvolutionContext` usage to be a utility data structure inside the tuning logic.
Hide it from passes such as `-miopen-affix-params` and `-miopen-lowering`.
Move the implementation of populating `ConvolutionContext` from a header file to a source file.

This is a first step of several steps to consolidate `ConvolutionContext` and `Conv2dGenerator::Config` to remove duplicated logic.

PR #565, opened 2022-03-21
Introduce `DimIndexAndSize` and use it in the context of `ConvolutionContext`.

whchung:
Following #563, this PR use a custom structure instead of `std::pair` to somewhat make the tuning logic more comprehensible. Originally we rely on `.first` for dimension index, and `.second` for dimension size.

The next step is to start consolidate how dimension index / dimension values are used between `ConvolutionContext` and `Conv2dGenerator::Config`.

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/Tuning/ConvContext.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Tuning/ConvContext.h
@@ -22,11 +22,16 @@
 
 namespace mlir {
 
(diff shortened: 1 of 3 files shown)


jerryyin (approved):
LGTM, thanks!

whchung:
Follow-ups are documented at: https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/478

krzysz00:
Very nice cleanup, thanks!

PR #566, opened 2022-03-22
Remove AbstractElementwise trait, use a big isa<> instead.

pcf000:
Folks on discourse.llvm.org are not inclined to support new traits.  Here's a version of the abstract-elementwise concept that removes all the stuff for the trait itself, and uses an isa<> to call out the ops that aren't otherwise identifiable.  All the action (relative to the main line) is now in TosaPartition.cpp.


krzysz00:
I'm not entirely sure it was the new trait they were objecting to uniformly, I think part of the complaint may have been stuff like the name.

Re the thread upstream, I think some concrete examples of what the partitioning does might help, including why some ops don't have the property that was captured by `AbstractElementwise`

sjw36 (approved):
Looks good.

PR #567, opened 2022-03-22
Fix int8 verifier

yiqian1:
Fix a bug in cpu verifier that misused data type when --conv-config is used.

Also add a config in Resnext101Config.toml which was missing from my previous PR. E2E tests were generated correctly. Somehow I missed the config when I committed.


jerryyin (approved):
LGTM, thank you!

PR #568, opened 2022-03-22
Adding kpack xdlops gemm implementation for i8

jerryyin:
I have implemented and fixed majority of issues discovered from kpack + reduction combination. With this PR, all lowering compilation issues will be fixed. 

### Example: correct result with default kpack == 1 tuning parameters

```bash
./bin/miopen-gen -batchsize=64 -in_channels=128 -in_h=16 -in_w=16 -out_channels=128 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=1 --padding_w=1 -t i8 -x2 --operation conv2d -pv | ./bin/mlir-miopen-driver -c | ./bin/mlir-rocm-runner --shared-libs=./lib/librocm-runtime-wrappers.so,./external/llvm-project/llvm/lib/libmlir_runner_utils.so --entry-point-result=void

Unranked Memref base@ = 0x559a10d423f0 rank = 1 offset = 0 sizes = [1] strides = [1] data =
[1]
```

### Example: correct result with kpack == 8 tuning parameters:

```bash
./bin/miopen-gen -batchsize=64 -in_channels=128 -in_h=16 -in_w=16 -out_channels=128 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=1 --padding_w=1 -t i8 -x2 --operation conv2d --perf_config "64,64,8,32,32,8,0,0" -pv | ./bin/mlir-miopen-driver -c | ./bin/mlir-rocm-runner --shared-libs=./lib/librocm-runtime-wrappers.so,./external/llvm-project/llvm/lib/libmlir_runner_utils.so --entry-point-result=void

Unranked Memref base@ = 0x559a10d423f0 rank = 1 offset = 0 sizes = [1] strides = [1] data =
[1]
```

Done:
 - kicking off nightly on this PR to make sure this does not cause a regression on int8 resnet50 configs

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
@@ -1033,9 +1033,10 @@ private:
       return failure();
     }
 
-    // XXX FIXME: Ignore KReduction XDLOPS path for forward and backward weight
-    // convolution now. These M/NPerBlock combinations will result in lowering
-    // errors at tuning.
+    // TODO remove : Ignore KReduction XDLOPS path for forward and backward
+    // weight convolution now (unless int8). These M/NPerBlock combinations used
+    // to result in lowering errors at tuning. Once non-int8 types are tested,
+    // we should get rid of this limitation.
     if (param.gemmKPack > 1 && !dataType.isInteger(8) &&
         ((ctx.getOpType() == miopen::ConvOpType::Fwd) ||
          (ctx.getOpType() == miopen::ConvOpType::BwdWeight))) {
--- a/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
@@ -1033,29 +1033,6 @@ private:
       return failure();
     }
 
-    if (dataType.isInteger(8)) {
-      // kpack for int8 must be larger than kbase of 4, which means
-      // kpack must be at least 4, once enabled.
-      if (param.gemmKPack == 2) {
-        return failure();
-      }
-
-      // When kpack is tunred off, K tile size must be larger than or
-      // equal to num_input_blks * k_base to qualify for a xdlops gemm
-      if (param.gemmKPack == 1) {
-        if ((param.gemmMPerBlock == 32) && (param.gemmNPerBlock == 32)) {
-          if (param.gemmKPerBlock < 8) {
-            return failure();
-          }
-        }
-        if ((param.gemmMPerBlock == 16) && (param.gemmNPerBlock == 16)) {
-          if (param.gemmKPerBlock < 16) {
-            return failure();
-          }
-        }
-      }
-    }
-
     // XXX FIXME: Ignore KReduction XDLOPS path for forward and backward weight
     // convolution now. These M/NPerBlock combinations will result in lowering
     // errors at tuning.
--- a/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
@@ -1033,7 +1033,7 @@ private:
     // XXX FIXME: Ignore KReduction XDLOPS path for forward and backward weight
     // convolution now. These M/NPerBlock combinations will result in lowering
     // errors at tuning.
-    if (param.gemmKPack > 1 &&
+    if (param.gemmKPack > 1 && !dataType.isInteger(8) &&
         ((ctx.getOpType() == miopen::ConvOpType::Fwd) ||
          (ctx.getOpType() == miopen::ConvOpType::BwdWeight))) {
       if ((param.gemmMPerBlock == 16 || param.gemmMPerBlock == 32 ||
--- a/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
@@ -828,19 +828,22 @@ private:
   // clang-format off
   llvm::SmallVector<InitParamsXDL, 4> initParametersForwardI8 = {
       // M/block N/block K/block M/wave N/wave kPack aCopyMore bCopyMore
-      // TODO enable kpack
-      // The 32 x 32 xdlops k/block must be divisible by 2
+      // kpack for int8 must be larger than kbase, which means
+      // kpack must be at least 4, once enabled.
+      {64, 64, 8, 32, 32, 8, false, false},
+      {64, 64, 8, 32, 32, 4, false, false},
+      {32, 32, 8, 16, 16, 8, false, false},
+      {32, 32, 8, 16, 16, 4, false, false},
+      // The 32 x 32 xdlops k/block must be at least 8
+      {64, 64, 16, 32, 32, 1, false, false},
       {64, 64, 8, 32, 32, 1, false, false},
-      {64, 64, 4, 32, 32, 1, false, false},
-      {64, 64, 2, 32, 32, 1, false, false},
+      {32, 32, 16, 32, 32, 1, false, false},
       {32, 32, 8, 32, 32, 1, false, false},
-      {32, 32, 4, 32, 32, 1, false, false},
-      {32, 32, 2, 32, 32, 1, false, false},
-      // The 16 x 16 xdlops k/block must be divisible by 4
-      {32, 32, 8, 16, 16, 1, false, false},
-      {32, 32, 4, 16, 16, 1, false, false},
-      {16, 16, 8, 16, 16, 1, false, false},
-      {16, 16, 4, 16, 16, 1, false, false},
+      // The 16 x 16 xdlops k/block must be at least 16
+      {32, 32, 32, 16, 16, 1, false, false},
+      {32, 32, 16, 16, 16, 1, false, false},
+      {16, 16, 32, 16, 16, 1, false, false},
+      {16, 16, 16, 16, 16, 1, false, false},
   };
   // clang-format on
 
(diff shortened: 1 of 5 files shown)


krzysz00:
Please give me a ping in a day or two so I can take a look at this properly. Nothing obviously wrong jumped at me, but I haven't yet made an effort to understand the impacts of the change.

jerryyin:
@krzysz00 @whchung Please help review. With this PR all int8 heuristic configs can continue to pass: So kpack + kreduction can work after this PR.

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
```diff
@@ -1033,7 +1036,7 @@ class PopulateParamsXDL : public PopulateParamsBase {
     // XXX FIXME: Ignore KReduction XDLOPS path for forward and backward weight
     // convolution now. These M/NPerBlock combinations will result in lowering
     // errors at tuning.
-    if (param.gemmKPack > 1 &&
+    if (param.gemmKPack > 1 && !dataType.isInteger(8) &&
```
whchung: @jerryyin I can see you made some fixes to `xdlops_gemm_v2` on KReduction path. Could you revise the comment that we only reject them for types other than i8 now?

Once this PR lands, I can check KReduction path for other data types and perhaps we can remove this XXX block altogether.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
```diff
@@ -1456,45 +1467,122 @@ struct XdlopsGemmV2RewritePattern : public OpRewritePattern<XdlopsGemmV2Op> {
...
+            loc, innerLoopb.create<MulIOp>(loc, outerLoopiv, KBaseConstantOp),
+            innerLoopb.create<MulIOp>(loc, innerLoopiv, KBaseConstantOp));
+
+        if (k_base == 1) {
+          argA = innerLoopb.create<memref::LoadOp>(loc, argType, op.bufferA(),
+                                                   ValueRange{offset});
+          argB = innerLoopb.create<memref::LoadOp>(loc, argType, op.bufferB(),
+                                                   ValueRange{offset});
+        } else {
+          // k_base > 1, use transferRead to build the
```
whchung: The comment is incomplete.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
```diff
@@ -1456,45 +1467,122 @@ struct XdlopsGemmV2RewritePattern : public OpRewritePattern<XdlopsGemmV2Op> {
...
+              loc, bufferAElementType.cast<VectorType>(), bufferA,
+              ValueRange{outerLoopiv});
+          argB = innerLoopb.create<memref::LoadOp>(
+              loc, bufferBElementType.cast<VectorType>(), bufferB,
+              ValueRange{outerLoopiv});
+        } else {
+          // If xdlops argument vector size is smaller than the vgpr vector
+          // length, we have to temporarily construct smaller vgpr vector
+          // for xdlops to consume
+          constructArg(bufferA, argA, outerLoopiv, innerLoopiv, argType,
```
whchung: how about change to:

`argA = constructArg(bufferA, outerLoopiv, innerLoopiv, argType, k_base)` to make it more comprehensible about the input / output?

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
```diff
@@ -1456,45 +1467,122 @@ struct XdlopsGemmV2RewritePattern : public OpRewritePattern<XdlopsGemmV2Op> {
...
+          argB = innerLoopb.create<vector::TransferReadOp>(
+              loc, argType.cast<VectorType>(), op.bufferB(),
+              ValueRange{offset});
+        }
       } else {
-        argA = innerLoopb.create<memref::LoadOp>(loc, argType, op.bufferA(),
-                                                 ValueRange{offset});
-        argB = innerLoopb.create<memref::LoadOp>(loc, argType, op.bufferB(),
-                                                 ValueRange{offset});
+        auto constructArg = [&outerLoopb, &innerLoopb, &loc,
```
whchung: A comment to briefly describe the semantic of the lambda could be useful. Also, `constructArg` is too generic. Perhaps rename it to depict the idea that we are constructing a narrower vector from a wider vector could make the code more comprehensible. Also, it'd be better have the lambda return a `Value` instance so the call sites looks more or less the same compared with other routes to prepare `argA` and `argB`.

whchung (changes requested):
Thanks for fixing the KReduction path. I have made some comments regarding to `constructArg` and how we use it.

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
jerryyin: Sure, I'll revise the comment.

BTW, I was originally planning to do it such that:
 1. get this PR merged
 2. get reduction for other data types bring up
 3. Rewrite this lowering to consolidate the logic between kreduction on/off, kpack on/off
 4. Remove all non-reduction case, and keep only reduction (This is such that MI300 can work)

Just want to make sure we don't happen to work on the same thing later.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
jerryyin: Sure

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
jerryyin: Sounds good.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
```diff
@@ -1524,11 +1526,16 @@ struct XdlopsGemmV2RewritePattern : public OpRewritePattern<XdlopsGemmV2Op> {
...
-          arg = createZeroConstantOp(innerLoopb, loc, argType);
+        // This lambda create an xdlops argument from the vgpr buffer.
+        // Each time in the nested loop, it first loads a vector of size
+        // kpack guarateed to be a multiple of the size of the xdlops
+        // argument. Then it construct the xdlops argument by extracting
+        // elements from the kpack vector.
+        auto constructXdlopsArg = [&outerLoopb, &innerLoopb, &loc,
+                                   &b](Value &buffer, Value &outerLoopiv,
+                                       Value &innerLoopiv, Type &argType,
+                                       int64_t k_base) {
```
whchung: add `-> Value` for clarity?

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
```diff
@@ -1524,11 +1526,16 @@ struct XdlopsGemmV2RewritePattern : public OpRewritePattern<XdlopsGemmV2Op> {
...
-                                 Value &innerLoopiv, Type &argType,
-                                 int64_t k_base) {
-          arg = createZeroConstantOp(innerLoopb, loc, argType);
+        // This lambda create an xdlops argument from the vgpr buffer.
+        // Each time in the nested loop, it first loads a vector of size
+        // kpack guarateed to be a multiple of the size of the xdlops
+        // argument. Then it construct the xdlops argument by extracting
+        // elements from the kpack vector.
+        auto constructXdlopsArg = [&outerLoopb, &innerLoopb, &loc,
+                                   &b](Value &buffer, Value &outerLoopiv,
```
whchung: nitpick: `b` is not really used in this lambda it seems?

jerryyin: Nice catch!

whchung (approved):
LGTM

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
whchung: We would still need to keep the non-reduction case for MI-100/200, so perhaps it's time to tap into the unused `arch` attribute.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
```diff
@@ -1456,45 +1467,130 @@ struct XdlopsGemmV2RewritePattern : public OpRewritePattern<XdlopsGemmV2Op> {
...
+              loc,
+              buffer.getType()
+                  .cast<MemRefType>()
+                  .getElementType()
+                  .cast<VectorType>(),
+              buffer, ValueRange{outerLoopiv});
+
+          for (int64_t i = 0; i < k_base; ++i) {
+            Value iterOp = innerLoopb.create<ConstantIndexOp>(loc, i);
+            Value newOffset =
```
krzysz00: This is `miopen::ExtractSliceOp`

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
jerryyin: @whchung Good point. Hopefully I can find a way to suppress non-reduction xdlops for MI300, such that reduction xdlops would be the only ones available. To do this, we indeed need the `arch` field.

Coming to think of it, I think `-xdlops` field is just a application of `arch` field as well (It is almost the equivalent of if (`arch` >= `gfx908`).

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
jerryyin: Done in c5ae273. Nice! This just make the lambda trivial!

PR #569, opened 2022-03-23
Revert vectorization size changes to fix non-xdlops

jerryyin:
This will fix https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/479.

This change was originally added from #539. As it turns out, the particular change that allow vectorization size to become wider will cause following non-xdlops nightly to fail:

>  Failed Tests (6):
  MLIR-MIOpen :: mlir-miopen-driver/auto_e2e/Resnet50/F16_tests/Resnet50_f16_2.mlir
  MLIR-MIOpen :: mlir-miopen-driver/auto_e2e/Resnet50/F16_tests/Resnet50_f16_3.mlir
  MLIR-MIOpen :: mlir-miopen-driver/auto_e2e/Resnet50/F16_tests/Resnet50_f16_8.mlir
  MLIR-MIOpen :: mlir-miopen-driver/auto_e2e/Resnet50/Int8_tests/Resnet50_int8_2.mlir
  MLIR-MIOpen :: mlir-miopen-driver/auto_e2e/Resnet50/Int8_tests/Resnet50_int8_3.mlir
  MLIR-MIOpen :: mlir-miopen-driver/auto_e2e/Resnet50/Int8_tests/Resnet50_int8_8.mlir

Since this is irrelevant to a particular data type, I will revert this change and get it documented in a separate PR to investigate.

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
@@ -367,11 +367,22 @@ protected:
     // in the algorithm.
     int64_t vectorizationSize = 1;
     auto dataType = ctx.getDataType();
-    unsigned dataWidth = dataType.getIntOrFloatBitWidth();
-    // 128 is the upper limit we support in vectorized load/store, which could
-    // be 4 fp32, 8 fp16, or 16 int8
-    const size_t highestPotentialVectorizationLen = 128;
-    vectorizationSize = highestPotentialVectorizationLen / dataWidth;
+
+    // TODO: Revert the vectorizationSize decision with below commented code:
+    // unsigned dataWidth = dataType.getIntOrFloatBitWidth();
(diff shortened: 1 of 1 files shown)


whchung (approved):

krzysz00:
Is it possible to fix instead of reverting?

jerryyin:
@krzysz00 Definitely possible. But considering that the deadline is in two days, the realistic thing to do is to make sure nightly CI is working again. Otherwise we won't be able to branch when we should.

krzysz00:
In which case, should we revert my regression as well, or do we want to try and fix it so Simon's PR can land

jerryyin:
@krzysz00 Use your best judgement. Desirably we want CI to be passing tonight so we can branch tomorrow. If any new feature will cause instability to CI, I prefer to not push it in because who knows how much additional time it may take to fix. I'd rather consider backporting than dragging.

krzysz00:
@sjw36 How likely is it that the fusion stuff you're interested in will actually make it in?

PR #571, opened 2022-03-24
[DO NOT SQUASH] Better use of buffer intrinsics

krzysz00:
1) Use the new raw atomic add everywhere
2) Instead of fixing buffer sizes to 2 GB, use the actual buffer size everywhere
2b) Because of this, don't actually emit checks for OOB on the right in the slowest-moving dimension of a tensor
3) Add `targetIsRDNA` to buffer ops to be Navi-proof
4) Use the buffer load itself to handle oob instead of selecting values afterwards
5) Drop bwd padding hack, it was compensating for the lack of bounds check on the left in stores

Future PRs include removing the BwdPaddingStoreHack enum for being unused and allowing padding in atomic add kernels.

# Performance info (selected arbitrarily)
On the config
```
conv -F 1 -f Nxxx -n 256 -c 128 -H 28 -W 28 -k 128 -x 3 -y 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -g 1
```

New, NHWC: 17..36
New, NCHW: 28.86
Old, NHWC: 17.04
Old, NCHW: 27.40
Percent change, NCHW: 5..3%
Percent change, NHWC: 1.88%


krzysz00:
Update: I worked out why some really big tests were failing (the method I used to trigger OOB led to integer overflows) and we now have a nightly pass

whchung:
Looks like another great cleanup!

krzysz00:
Also, in misc notes that I might want to follow up on further, *nod* unrolling the loops that go to or from global memory bought us another TFlop on the NCHW case above with this change and kept the NHWC case flat - inspection of the ISA showed that things still got unrolled but that the address computation was handled differently. There might be some stuff to play around with here pending a more comprehensive perf test.

krzysz00:
In interesting fun with performance, the NCHW perf improvements are consistent on lockhart5, but for NHWC there's a perf regression of about 1 TFlops with the new code. Very weird.

krzysz00:
Other good news is that on MI-100, with the case I've been testing, the perf improvements are consistent when looking at bwd and wrw as well

whchung (approved):
LGTM!

PR #573, opened 2022-03-25
[DO NOT SQUASH] MLIR#217: Add broadcast support to miopen.transform

sjw36:
* use index maps to ensure minority identity maps for linalg.generic
* insert miopen.transform<Broadcast> before applying output transforms
* * generate affine map to access broadcast dimensions
* moved linalg-elementwise-fusion pass to after bufferization

* fixed tosa-to-linalg to prefer minor-identity matrix maps

Changes to CoordTransformBuilder:
* added constructor to assign names generically
* added broadcast method
* added expand macro (passThrough + addDim)

https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/217

### DIFF
--- a/external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalg.cpp
+++ b/external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalg.cpp
@@ -670,7 +670,7 @@ elementwiseMatchAndRewriteHelper(Operation *operation,
     SmallVector<AffineExpr, 4> affineExprs;
     newShape.reserve(type.getRank());
     for (const auto &it : llvm::enumerate(type.getShape())) {
-      if (it.value() == resultTy.getDimSize(it.index())) {
+      if (it.value() == resultTy.getDimSize(it.index()) && it.value() != 1) {
         newShape.push_back(it.value());
         affineExprs.push_back(
             mlir::getAffineDimExpr(it.index(), rewriter.getContext()));
--- a/mlir/include/mlir/Dialect/MIOpen/MIOpen.h
+++ b/mlir/include/mlir/Dialect/MIOpen/MIOpen.h
@@ -126,6 +126,8 @@ public:
   void passThrough(ArrayRef<StringRef> names);
   void passThrough(ArrayRef<StringRef> outNames, ArrayRef<uint32_t> outDims,
                    ArrayRef<StringRef> inNames);
+  void passThrough(ArrayRef<uint32_t> endIndices,
+                   ArrayRef<uint32_t> startIndices);
 
   // Parameters is the pre and post padding for each dimension in the order they
   // appear as arguments. For example, padding x with 2 on the left and 1 on the
@@ -266,8 +268,6 @@ public:
 
   // Defines a dimension that is not mapped to any coordinates in the output
   void addDim(StringRef name, uint32_t dim, int64_t size);
-
-  void expand(ArrayRef<uint32_t> dims, ArrayRef<int64_t> sizes);
 
   void broadcast(ArrayRef<uint32_t> bcastDims, ArrayRef<int64_t> endDims);
 
--- a/mlir/include/mlir/Dialect/MIOpen/MIOpen.h
+++ b/mlir/include/mlir/Dialect/MIOpen/MIOpen.h
@@ -142,6 +142,8 @@ public:
 protected:
   CoordTransformsBuilder(mlir::Builder &builder, ArrayRef<StringRef> startNames,
                          ArrayRef<int64_t> startShape, mlir::Location loc);
+  CoordTransformsBuilder(mlir::Builder &builder, ArrayRef<int64_t> startShape,
+                         mlir::Location loc);
 
   template <class T, typename = typename std::enable_if<std::is_base_of<
                          CoordTransformsBuilder, T>::value>::type>
@@ -160,6 +162,8 @@ protected:
                              SmallVectorImpl<int64_t> &lowerBounds) = 0;
 
   virtual int64_t paddingSign() const = 0;
+
+  llvm::SmallVector<SmallString<8>, 8> &nStartNames() { return startNames; }
 
   uint32_t nStartDims();
   uint32_t nEndDims();
@@ -250,6 +254,10 @@ public:
                     ArrayRef<int64_t> startShape, mlir::Location loc)
       : CoordTransformsBuilder(builder, startNames, startShape, loc) {}
 
+  BottomUpCTBuilder(mlir::Builder &builder, ArrayRef<int64_t> startShape,
+                    mlir::Location loc)
+      : CoordTransformsBuilder(builder, startShape, loc) {}
+
   static BottomUpCTBuilder above(BottomUpCTBuilder &previous,
                                  TransformMapAttr &result) {
     return CoordTransformsBuilder::nextTransforms(previous,
@@ -258,6 +266,10 @@ public:
 
   // Defines a dimension that is not mapped to any coordinates in the output
   void addDim(StringRef name, uint32_t dim, int64_t size);
+
+  void expand(ArrayRef<uint32_t> dims, ArrayRef<int64_t> sizes);
+
+  void broadcast(ArrayRef<uint32_t> bcastDims, ArrayRef<int64_t> endDims);
 
   void slice(ArrayRef<StringRef> upperNames, ArrayRef<StringRef> lowerNames,
              ArrayRef<int64_t> begins, ArrayRef<int64_t> ends);
--- a/mlir/include/mlir/Dialect/MIOpen/MIOpen.h
+++ b/mlir/include/mlir/Dialect/MIOpen/MIOpen.h
@@ -165,7 +165,7 @@ protected:
 
   virtual int64_t paddingSign() const = 0;
 
-  llvm::SmallVector<SmallString<8>, 8> &nStartNames() { return startNames; }
+  llvm::SmallVector<SmallString<8>, 8> &getStartNames() { return startNames; }
 
   uint32_t nStartDims();
   uint32_t nEndDims();
@@ -269,7 +269,7 @@ public:
   // Defines a dimension that is not mapped to any coordinates in the output
   void addDim(StringRef name, uint32_t dim, int64_t size);
 
-  void broadcast(ArrayRef<uint32_t> bcastDims, ArrayRef<int64_t> endDims);
+  void broadcast(ArrayRef<uint32_t> endDims, ArrayRef<int64_t> endSizes);
 
   void slice(ArrayRef<StringRef> upperNames, ArrayRef<StringRef> lowerNames,
              ArrayRef<int64_t> begins, ArrayRef<int64_t> ends);
--- a/mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
+++ b/mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
@@ -68,7 +68,7 @@ static bool isConstantZero(Value v) {
 }
 
 static Value expandMemRef(ConversionPatternRewriter &rw, Operation *op,
-                          Value operand, int idx = 4) {
+                          Value operand, uint32_t idx = 4) {
   auto loc = op->getLoc();
   auto oprType = operand.getType().template cast<ShapedType>();
   if (!oprType.hasStaticShape()) {
@@ -78,32 +78,9 @@ static Value expandMemRef(ConversionPatternRewriter &rw, Operation *op,
   }
   ArrayRef<int64_t> shape = oprType.getShape();
 
-  SmallVector<SmallString<8>> dimNames;
-  SmallVector<StringRef> dimNameRefs;
-  dimNames.reserve(shape.size());
-  dimNameRefs.reserve(shape.size());
-  for (size_t i = 0, e = shape.size(); i < e; ++i) {
-    SmallString<8> dimName;
-    ("dim" + Twine(i)).toVector(dimName);
-    dimNames.emplace_back(std::move(dimName));
-    dimNameRefs.push_back(StringRef(dimNames[i]));
-  }
+  miopen::BottomUpCTBuilder transform(rw, shape, loc);
+  transform.expand({idx}, {1});
 
-  miopen::BottomUpCTBuilder transform(rw, dimNameRefs, shape, loc);
-  llvm::StringMap<uint32_t> upperNames;
-  if (idx == 0) {
-    StringRef expandedDim = dimNameRefs[0];
-    upperNames = miopen::expandNamesInPlace(
-        dimNameRefs, {{expandedDim, {"g", expandedDim}}});
-  } else {
-    StringRef expandedDim = dimNameRefs[idx - 1];
-    upperNames = miopen::expandNamesInPlace(
-        dimNameRefs, {{expandedDim, {expandedDim, "g"}}});
-  }
-
-  miopen::BottomUpCTTopDimsWrapper wrapper(transform, upperNames);
-  wrapper.passThrough(dimNameRefs);
-  wrapper.addDim("g", 1);
   return rw.create<miopen::TransformOp>(loc, operand, transform.get());
 }
 
--- a/mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
+++ b/mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
@@ -78,8 +78,15 @@ static Value expandMemRef(ConversionPatternRewriter &rw, Operation *op,
   }
   ArrayRef<int64_t> shape = oprType.getShape();
 
+  SmallVector<uint32_t, 8> endDims;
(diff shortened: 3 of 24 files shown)


sjw36:
@krzysz00 this PR has some changes dropped from 468. I prefer to avoid generating and copying names around when it is irrelevant for my transforms. Let me know what you think.

external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalg.cpp
```diff
@@ -670,7 +670,7 @@ elementwiseMatchAndRewriteHelper(Operation *operation,
     SmallVector<AffineExpr, 4> affineExprs;
     newShape.reserve(type.getRank());
     for (const auto &it : llvm::enumerate(type.getShape())) {
-      if (it.value() == resultTy.getDimSize(it.index())) {
+      if (it.value() == resultTy.getDimSize(it.index()) && it.value() != 1) {
```
krzysz00, line 673: What's this do? Can it be upstreamed?

mlir/include/mlir/Dialect/MIOpen/MIOpen.h
```diff
@@ -259,6 +267,10 @@ class BottomUpCTBuilder : public CoordTransformsBuilder {
   // Defines a dimension that is not mapped to any coordinates in the output
   void addDim(StringRef name, uint32_t dim, int64_t size);
 
+  void expand(ArrayRef<uint32_t> dims, ArrayRef<int64_t> sizes);
```
krzysz00: expand() still feels too specific

mlir/include/mlir/Dialect/MIOpen/MIOpen.h
```diff
@@ -161,6 +163,8 @@ class CoordTransformsBuilder {
 
   virtual int64_t paddingSign() const = 0;
 
+  llvm::SmallVector<SmallString<8>, 8> &nStartNames() { return startNames; }
```
krzysz00: I don't think this needs an `n` on it

But it is a good method that's missing.

mlir/include/mlir/Dialect/MIOpen/MIOpen.h
krzysz00: What I'd suggest is the following: `passThrough(ArrayRef<uint32_t> endIndices, ArrayRef<uint32_t> startIndices)` - this lets you avoid having to talk about the names at all

mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
```diff
@@ -78,32 +78,9 @@ static Value expandMemRef(ConversionPatternRewriter &rw, Operation *op,
...
-  } else {
-    StringRef expandedDim = dimNameRefs[idx - 1];
-    upperNames = miopen::expandNamesInPlace(
-        dimNameRefs, {{expandedDim, {expandedDim, "g"}}});
-  }
+  miopen::BottomUpCTBuilder transform(rw, shape, loc);
+  transform.expand({idx}, {1});
 
-  miopen::BottomUpCTTopDimsWrapper wrapper(transform, upperNames);
-  wrapper.passThrough(dimNameRefs);
```
krzysz00, line 105: And so this'll be (not needing the wrapper) `.passThrough({0, 1, ... idx - 1, idx + 1, ... n + 1}, {0, 1, ... n})` and then `addDim("g", idx, size)`

mlir/lib/Dialect/MIOpen/CoordTransformBuilder.cpp
```diff
@@ -112,6 +112,17 @@ AffineMapAttr assembleMapFor(Builder &b, ArrayRef<TransformAttr> transforms,
       assert(upperDims.size() == 1 && lowerDims.size() == 0 &&
              "Invalid AddDim");
       // dimension is ignored, do nothing
+    } else if (type == TransformType::Broadcast) {
+      // Compute lower dimension strides.
+      for (auto pair : llvm::enumerate(lowerBounds)) {
```
krzysz00: Not like this, please

Proposed alternate design: broadcast takes a set of upper dimensions and a set of lower dimensions, and its parameters are the sizes of the lower dimensions (yes, it's duplicative, it's future-proofing for dynamic shapes)

That is, `broadcast{2, 1} ["g", "h"] at [0, 1] -> ["g", "h"] at [0, 1]` produces the map `(d0, d1) -> (d0 % 2, d1 % 1)`, aka `(d0, d1) -> (d0 % 2, 0)`. The upper size is specified separately.

The individual transforms doing this sort of global reasoning over the entire map makes the setup much less clean.

mlir/lib/Dialect/MIOpen/CoordTransformBuilder.cpp
```diff
@@ -523,6 +553,55 @@ void BottomUpCTBuilder::addDim(StringRef name, uint32_t dim, int64_t size) {
...
+    auto dim = std::get<0>(tuple);
+    auto size = std::get<1>(tuple);
+    SmallString<8> name;
+    ("exp" + Twine(dim)).toVector(name);
+
+    addDim(name, dim, size);
+  }
+}
+
+void BottomUpCTBuilder::broadcast(ArrayRef<uint32_t> bcastDims,
```
krzysz00: No, I don't agree with this signature. It should be (with some specialized versions, versions that don't mention names, etc.) `BottomUpCTBuilder::broadcast(ArrayRef<StringRef> endNames, ArrayRef<uint32_t> endDims, ArrayRef<StringRef> startNames, ArrayRef<int64_t> upperSizes)`

and also `TopDownCTBuilder::collapse(ArrayRef<StringRef> endNames, ArrayRef<uint32_t> endDims, ArrayRef<StringRef> startNames, ArrayRef<int64_t> lowerSizes)` which also produces a broadcast

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -138,27 +142,83 @@ template <typename T> struct MILARewritePattern : public OpRewritePattern<T> {
...
+        inpType = inp.getType().template cast<MemRefType>();
+        inpShape = inpType.getShape();
+        broadcastedDims = {expDims.begin(), expDims.end()};
+      } else {
+        inpIdxMap.isMinorIdentityWithBroadcasting(&broadcastedDims);
+      }
+
+      // 0. insert a broadcast miopen.transform
+      miopen::BottomUpCTBuilder transform(b, inpShape, loc);
+      transform.broadcast(broadcastedDims, outType.getShape());
```
krzysz00: Yeah, I don't like this. The broadcast should specify exactly which dimensions it applies to and the rest should be passed through. Not doing this makes the code hard to reason about.

mlir/lib/Dialect/MIOpen/utility/loweringUtils.cpp
```diff
@@ -27,6 +27,7 @@ void propagateTransformOob(TransformMapAttr transformMap,
 
     switch (transform.getType()) {
     case TransformType::PassThrough:
+    case TransformType::Broadcast: // is this right??
```
krzysz00: No, no it isn't.

The index diff update rule fro broadcast is as follows: `lowerDiff[lowerDims[i]] = upperDiff[upperDims[i]] % params[i]`

That is, suppose you've broadcast `[a, b]` into `[a, b, a, b, ...]` Then, adding 1 moves you to the other element, but adding 2 keeps where you are.

mlir/lib/Dialect/MIOpen/CoordTransformBuilder.cpp
krzysz00: Alternatively, allowing for fun times like, `[broadcast{1} ["d"] at [0] -> ["d"] at [0]] bounds [4] -> [2]` being the transform [a, a, a, a] <- [a, b], the `broadcast()` method could be most general by taking both the upper and lower sizes, but that's a stretch and I'm not entirely sold on it either.

krzysz00 (changes requested):
I strongly disagree with your design of broadcast, it does not fit into the coordinate transformation paradigm. In addition, your update to index diff maps is incorrect.

Also, please add unit tests for broadcast() to the coordinate transform unit tests.

Also also, now that we have that startNames() helper, please go fix up the code over in `GridwiseGemmToBlockwise` that also has the "dim[blah]" pattern

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
sjw36: The broadcast call here passes the dim indices that would be broadcast but it also needs the upper 'result' dim sizes since the builder takes the lower sizes only.

krzysz00: But you're not broadcasting every dimension, while passing every size

krzysz00: broadcast() should only take the upper sizes of the dimensions that are being subject to broadcast

mlir/lib/Dialect/MIOpen/CoordTransformBuilder.cpp
sjw36: It already has the lower dims from construction, but do you require each transform to pass both lower and upper sizes?

mlir/lib/Dialect/MIOpen/CoordTransformBuilder.cpp
sjw36: yes, I think I understand now. lowerBounds/upperBounds are global? and the lowerDims/upperDims are specific to the transform?

mlir/lib/Dialect/MIOpen/CoordTransformBuilder.cpp
krzysz00: Agreed, hence why I was hesitant about the general case. Just passing in the upper dim sizes might be sufficient, if we are willing to make the assumption that we always broadcast the lower dimension in its entirity - we could then just copy those lower sizes to the parameters. (My instinct towards generality says we should provide a way to override that default, but still).

I do, at the very least, want broadcast() to allow for the lower and upper dimensions to be in different places.

external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalg.cpp
sjw36: It should be upstreamed.

mlir/lib/Dialect/MIOpen/CoordTransformBuilder.cpp
krzysz00: Yep, lowerBounds is a global property that shouldn't come up in the affine map construction, and should be instead handled by setting parameters on the transform.

This is partly because, from where I'm standing, it might make sense to remove upperBounds and lowerBounds entirely, especially now that we have `transforming_for`

external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalg.cpp
sjw36: As stated in the comment, it encourages minor identity maps in tosa-to-linalg for elementwise.

krzysz00: What are those? Why do we want them?

sjw36: See AffineMap::isMinorIdentity(). Basically (d0, d1, d2, d3) -> (0, 0, d2, d3).

It is easier to reason about at this point but this code would create (d0, d1, d2, d3) -> (d0, 0, 0, d3) if d0 happens to be size '1'.

krzysz00: Ah, makes sense.

mlir/include/mlir/Dialect/MIOpen/MIOpen.h
sjw36: rename to getStartName()

sjw36:
Updated for new understanding of CoordTransformsBuilder.

mlir/include/mlir/Dialect/MIOpen/MIOpen.h
```diff
@@ -259,6 +269,8 @@ class BottomUpCTBuilder : public CoordTransformsBuilder {
   // Defines a dimension that is not mapped to any coordinates in the output
   void addDim(StringRef name, uint32_t dim, int64_t size);
 
+  void broadcast(ArrayRef<uint32_t> bcastDims, ArrayRef<int64_t> endDims);
```
krzysz00: The names I'd go with here are `endDims` and `endSizes`

mlir/include/mlir/Dialect/MIOpen/MIOpen.h
krzysz00: Reminder on rename

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
```diff
@@ -1020,6 +1020,9 @@ struct IndexDiffUpdateRewritePattern
         }
       } else if (transformation == TransformType::AddDim) {
         // Do nothing - the dimension will be dropped by the code below
+      } else if (transformation == TransformType::Broadcast) {
+        // lower broadcast dims, uses map
+        assert(0);
```
krzysz00:
```
for (i : 0..(e.size()) {
  int64_t lowerLen = e[i];
  Value lowerLenOp = b.create<ConstantIndexOp>(loc, lowerLen);
  auto mbUpperDiff = `isConstantValue(upperDiffs[p[i]]);
  Value wrappedDiff;
  if (mbUpperDiff.hasValue()) {
    wrappedDiff = b.create<ConstantIndexOp>(loc, *mbUpperDiff % lowerLen);
  } else {
    wrappedDiff = b.create<RemUIOp>(loc, upperDiffs[p[i]], lowerLenOp);
  }
  Value newLower = addToOriginal(lowerIndicesOriginal[q[i]], wrappedDiff);
  newLower = b.create<RemUIOp>(loc, newLower, lowerLenOp);
  Value lowerDiff = b.create<SubIOp>(loc, newLower, lowerIndicesOriginal[q[i]]);
  lowerDiffsUpdated[q[i]] = lowerDiff;
  lowerIndicesUpdated[q[i]] = newLower;
}
```
is about how I'd write this. Long-term, some of this could probably be massively simplified with some generous application of `createOrFold`, but that's a future us problem.

krzysz00 (changes requested):
Minor nitpicks re names, a suggestion for updateIndexDiff, and missing unit tests.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
sjw36: How can I exercise this code? Does it only show up with xdlops?

krzysz00: In the non-xdlops case, you need a gemm that doesn't have any extra padding.

(Getting that limitation more precise is a background todo item)

sjw36: so I changed all dimensions to 64 and the filter width/height to 1.. and still did not generate miopen.index_diff_update. Do you have a config that will do this?

krzysz00:
```
./bin/miopen-gen -fil_layout=kcyx -in_layout=nchw -out_layout=nkhw -batchsize=64 -in_channels=64 -in_h=64 -in_w=64 -out_channels=64 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=0 --padding_w=0 -t f32 --operation conv2d | ./bin/mlir-miopen-driver -miopen-affix-params -miopen-lowering -miopen-lowering-step{2,3}
```
generates a `transforming_for` around a `buffer_store` at the end of step 3 that has `useIndexDiffs` which means it'll give you an `index_diff_update`

If the `transforming_for` your `threadwise_copy` turns into doesn't have `useIndexDiffs`, you won't hit the index diff code

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
sjw36: test added

krzysz00 (commented):
Overall, looks quite good

I'd also like to draw your attention to `mlir/unittests/Dialect/MIOpen/CoordTransformBuilderTests.cpp` in case you spot some missing test coverage there

krzysz00 (approved):
I think you've gotten everything I wasn't happy about, and so approved

Thanks for all the work on this!

PR #551, opened 2022-03-15
Enable KPACK for backward weight convolution.

whchung:
This PR enables KPACK for backward weight convolution when XDLOPS it used.

Combined with #548, the speed of backward weight convolution would be greatly improved.

Benchmark commmand:
```
./bin/miopen-gen --x2 --operation conv2d_bwd_weight -t f16 --batchsize 64 --in_channels 1024 --out_channels 1024 --in_h 14 --in_w 14 --out_h 14 --out_w 14 --fil_h 1 --fil_w 1 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h 0 --padding_w 0 --groupsize 1 -ph | ./bin/mlir-miopen-driver -c | rocprof --stats ./bin/mlir-rocm-runner --shared-libs=./lib/librocm-runtime-wrappers.so,./external/llvm-project/llvm/lib/libmlir_runner_utils.so --entry-point-result=void
```

This is a backward weight convolution which uses XDLOPS, in fp16, no padding along Gemm M/N/K. Therefore a 0-init kernel and a conversion kernel will be used.

At tip of `miopen-dialect`:
```
# cat results.stats.csv
"Name","Calls","TotalDurationNs","AverageNs","Percentage"
"miopen_conv2d_bwd_weight_gkcyx_ngchw_ngkhw_0.kd",1,290640047,290640047,51.83518016940135
"miopen_conv2d_bwd_weight_gkcyx_ngchw_ngkhw_2.kd",1,264794271,264794271,47.2256280123375
"miopen_conv2d_bwd_weight_gkcyx_ngchw_ngkhw_1.kd",1,5266052,5266052,0.9391918182611508
```
Total time: `560700370`

With PR #548:
```
# cat results.stats.csv
"Name","Calls","TotalDurationNs","AverageNs","Percentage"
"miopen_conv2d_bwd_weight_gkcyx_ngchw_ngkhw_1.kd",1,5289100,5289100,99.4823779017496
"miopen_conv2d_bwd_weight_gkcyx_ngchw_ngkhw_2.kd",1,20960,20960,0.3942354352953568
"miopen_conv2d_bwd_weight_gkcyx_ngchw_ngkhw_0.kd",1,6560,6560,0.12338666295503534
```
Total time: `5316620`

With PR #548 + this PR:
```
# cat results.stats.csv
"Name","Calls","TotalDurationNs","AverageNs","Percentage"
[...]

PR #575, opened 2022-03-28
Enable KPACK for backward data convolution.

whchung:
This PR enables KPACK for backward data convolution when XDLOPS it used.

Benchmark commmand:

```
./bin/miopen-gen -p -x2 --operation conv2d_bwd_weight -t f16 -ph | ./bin/mlir-miopen-driver -c | rocprof --stats ./bin/mlir-rocm-runner --shared-libs=./lib/librocm-runtime-wrappers.so,./external/llvm-project/llvm/lib/libmlir_runner_utils.so --entry-point-result=void
```

At tip of miopen-dialect:
```
# cat results.stats.csv
"miopen_conv2d_bwd_data_gkcyx_ngchw_ngkhw_0.kd",1,8806609,8806609,100.0
```

With this PR:
```
# cat results.stats.csv
"miopen_conv2d_bwd_data_gkcyx_ngchw_ngkhw_0.kd",1,683203,683203,100.0
```

@jerryyin , this is one of the low-hanging fruit after #551 has been proved, and the meeting with @asroy earlier today.


mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -837,6 +842,32 @@ LogicalResult backwardData(Conv2DBwdDataOp op, PatternRewriter &b) {
...
+      // Use Unmerge to split gemmK (dim 1) into gemmK and gemmKPack, place
+      // gemmKPack at dim 3.
+      int64_t gemmKLength = sourceTransform.endSize(1);
+      auto gemmKName = sourceTransform.endName(1);
+      kpackGemmFilterTransform.unmerge({gemmKName, "gemmKPack"}, {1, 3},
+                                       gemmKName, {gemmKLength / KPack, KPack});
+      TransformMapAttr kpackGemmFilterTransformAttr =
+          kpackGemmFilterTransform.get();
+      gemmFilterKPack =
+          b.create<TransformOp>(loc, source, kpackGemmFilterTransformAttr);
```
jerryyin: I'd prefer this + line 1036 + below consolidated into a function that builds a K / Kpack transform.

https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/395e80539e3128362b3a1ec6659aff81df597db2/mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp#L1335-L1349

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -22,7 +22,7 @@ void buildMIOpen(String cmakeOpts) {
 }
 
 void getAndBuildMIOpen(String prefixOpt, String cmakeOpts) {
-    git branch: 'mlir-enlarge-tuning-search-space', poll: false,\
+    git branch: 'mlir-enlarge-tuning-search-space-bwd', poll: false,\
```
jerryyin, line 25: Did you regress through all bwd tuning configs? If not, we'd need to run that regression on the PR.

jerryyin (commented):
Looking good, with small nitpicks. Thanks for generalizing kpack to bwd data.

mlir/utils/jenkins/Jenkinsfile
whchung: I did locally. But it's a good idea to verify with CI too to avoid the confusion we had for #551.

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
whchung: Will do.

whchung:
@jerryyin I've factored out common logic applying KPACK transforms in backward data and backward weight atomic add kernels.

For forward + backward weight (non atomic add), the logic would be a bit different. I'll filed another PR for it.

jerryyin (approved):
LGTM, thanks for the kpack implementation +  cleanup

PR #576, opened 2022-03-29
xdlops fusion

jungpark-mlir:


mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -138,6 +140,82 @@ template <typename T> struct MILARewritePattern : public OpRewritePattern<T> {
...
+    SmallVector<ReassociationExprs, 1> reassocs;
+    collapsedShape.push_back(collapsedDim);
+    reassocs.push_back(exprs);
+
+    auto collapsedType =
+        MemRefType::get(collapsedShape, sourceType.getElementType(), {}, 5);
+    Value result =
+        b.create<memref::CollapseShapeOp>(loc, collapsedType, source, reassocs);
+    return result;
+  }
```
krzysz00, line 230: Maybe we should do this with coordinate transformations instead?

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -138,6 +140,82 @@ template <typename T> struct MILARewritePattern : public OpRewritePattern<T> {
...
+    for (uint i = 0; i < shape.size(); ++i) {
+      regShape.push_back(1);
+      indexVector.push_back(c0);
+    }
+    regShape[shape.size() - 1] = regVecType.getNumElements();
+    auto elemType = regVecType.getElementType();
+    auto regType = MemRefType::get(regShape, elemType, {}, 5);
+    auto clonedVec = b.create<miopen::GpuAllocOp>(loc, regType);
+    auto indices = ValueRange(indexVector);
+    b.create<vector::StoreOp>(loc, nVecSlice->getResult(0), clonedVec, indices);
```
krzysz00, line 324: Does it make a difference if you use vector store or miopen::InBoundsStore when you look at the generated ISA?

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -138,6 +140,82 @@ template <typename T> struct MILARewritePattern : public OpRewritePattern<T> {
...
+      uint inIdx = 2 + i;
+      uint outIdx = 2 + shape.size() + i;
+      auto inCoord = miTWCopy->getOperand(inIdx);
+      auto outCoord = miTWCopy->getOperand(outIdx);
+      nTWCopy->setOperand(outIdx, inCoord);
+      nTWCopy->setOperand(inIdx, outCoord);
+    }
+    // 4. keep bound and transforms, source/dest will be swapped in later stage.
+    // 5. Adjust the copy to show the correct argument as global
+    nTWCopy->setAttr("globalArg", b.getIndexAttr(0));
```
krzysz00: Not used anymore

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -376,35 +456,121 @@ template <typename T> struct MILARewritePattern : public OpRewritePattern<T> {
...
+      if (dataPerCopy > 1) {
+        Value loaded = b.create<miopen::BufferLoadOp>(
+            loc, vecType, source, srcLeftOob, srcRightOob,
+            copyLoop.getLowerCoords(/*domain=*/0));
+        SmallVector<Value, 6> indicies;
+        for (uint i = 0; i < shape.size() - 1; ++i) {
+          indicies.push_back(c0);
+        }
+        indicies.push_back(copyLoop.getLowerCoords(/*domain=*/1)[0]);
+        b.create<vector::StoreOp>(loc, loaded, dest, indicies);
```
krzysz00: Don't case split vector store and memref store, we have `miopen.in_bounds_store` for a reason

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -376,35 +456,121 @@ template <typename T> struct MILARewritePattern : public OpRewritePattern<T> {
...
+
+      return fail;
     }
+  };
+
+  //===----------------------------------------------------------------------===//
+  // ThreadwiseCopyV2 Conversion in Fusion.
+  //===----------------------------------------------------------------------===//
+
+  struct ThreadwiseCopyV2RewritePattern
```
krzysz00: Hey, hold on, why's this here? The threadwise copy lowering is over in `BlockwiseGemmToThreadwise` - copying it will lead to bugs

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2496,18 +2496,22 @@ struct GridwiseGemmV2RewritePattern
...
-    // deals with a distinct set of values.
-    llvm::SmallVector<Value, 4> vectors;
-    vectors.reserve(transformedTail.size() * iterationsPerVectorC);
+
+    // Merge the vectors using miopen.insertslice and store loop to use
+    // miopen.extractslice so offset parsing can be deferred after loop
+    // unrolling.
+    int64_t vectorLen = vectorType.getShape()[0];
+    VectorType mergedType =
+        VectorType::get({static_cast<long>(vectorLen * transformedTail.size())},
```
krzysz00, line 2504: You might want the clone method on vectorType here?

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2603,11 +2640,16 @@ struct GridwiseGemmV2RewritePattern
         //     index_t col = col_blk * mfma_type.n + blk_td + n_i * NPerXdlops;
         threadMtxColInBlock = blk_td_xdlops_gemm;
       }
-      int64_t thread_mtx_on_blk_col_const =
-          col_blk_xdlops_gemm * n + n_i_xdlops_gemm * NPerXdlops;
+
+      auto n_ConstantOp = b.create<arith::ConstantIndexOp>(loc, n);
+      auto NPerXdlops_ConstantOp =
+          b.create<arith::ConstantIndexOp>(loc, NPerXdlops);
+      auto TMOBC1 = b.create<MulIOp>(loc, col_blk_xdlops_gemm, n_ConstantOp);
```
krzysz00: Variable naming ...

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2496,18 +2496,22 @@ struct GridwiseGemmV2RewritePattern
...
-    vectors.reserve(transformedTail.size() * iterationsPerVectorC);
+
+    // Merge the vectors using miopen.insertslice and store loop to use
+    // miopen.extractslice so offset parsing can be deferred after loop
+    // unrolling.
+    int64_t vectorLen = vectorType.getShape()[0];
+    VectorType mergedType =
+        VectorType::get({static_cast<long>(vectorLen * transformedTail.size())},
+                        vectorType.getElementType());
+    Value resultMerged =
```
krzysz00, line 2506: I'd like a perf test here - I'm hoping this early post-swizzle manipulation doesn't screw up the instruction selector

krzysz00 (commented):
I've got some minor concerns about how this's been implemented, mainly the duplicate threadwise copy rewrite pattern

mlir/test/fusion/xdlops/tosa-to-miopen-add.mlir
```diff
@@ -0,0 +1,13 @@
+// RUN: mlir-miopen-driver -host-pipeline highlevel -kernel-pipeline rocdl %s -o -| FileCheck %s
+
+// CHECK-LABEL: test_fusion_x2
```
whchung: can we be more specific about the expected results of the test?

mlir/test/fusion/xdlops/tosa-to-miopen-relu.x2.e2e.mlir
```diff
@@ -0,0 +1,27 @@
+// RUN: mlir-miopen-driver -host-pipeline highlevel %s | miopen-gen -ph -print-results -rand 1 -rand_type float - | mlir-miopen-driver -c  | mlir-rocm-runner --shared-libs=%rocm_wrapper_library_dir/librocm-runtime-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s
+
+module {
+// CHECK: Unranked Memref base
```
whchung, line 4: can we be more specific about the expected test result?

also can a lowering test be added to check the expected IR without running the kernel?

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
jungpark-mlir: Actually I tried but wasn't able to find which one to use. Can you point me any example? this is simply flattening to 1d and all the other dimensions are dummies. for example [1,1,1,1,1,16] ->[16]

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
jungpark-mlir: This is casting vector to the same sized register as memref (offset is always {0,0,,}), does it still matter?

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
jungpark-mlir: So,  xdlops fusion consists of two steps. First the fusion logic fuses and it includes illegal threadwise copy v2 (mem ->vector) and this pattern only convert that copy.
At least, this needs more comment to explain. And it might be reworked as a function if there is a strong reason not to do this way.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jungpark-mlir: Can you give me more information about clone method? duplicating vector type and changing the shape still need same amount of code, I suppose.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jungpark-mlir: Yeah, I'm also interested if this causes any performance regression. Fortunately, few cases I've seen showed almost identical instructions before going into llvm already. Let's see the performance report and I'll investigate if it catches anything.

mlir/test/fusion/xdlops/tosa-to-miopen-relu.x2.e2e.mlir
jungpark-mlir: I suppose this is e2e test verifying the result against that from CPU, so expecting correct result of fused convolution and relu.

Lowering test is good point, I'll add it with specific ir and corresponding result.

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
krzysz00: You'll want to `addDim()` in the dummy dimensions and `PassThrough` the 16

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
krzysz00: Probably not,  and in this particular case it'll make the code clearer, so let's keep this `vector.store` but make a bunch of cases into `miopen.in_bounds_store`

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
krzysz00: Could this have been a function instead of a rewrite?

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: The `ShapedType`s have methods that, IIRC, let you do "I want this type but with a different shape (or datatype, or so on)"

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: I'd give some convolutions that use swizzles a `--debug-only=serialize-to-blob` from this PR and from `miopen-dialect`

Also, for perf, you can run the perf test script yourself: you need to `ninja ci-performance-scripts` and then do something like
```
./bin/MIOpenDriver.py -o nhwc-test-description.csv -F 1 -f NHWC -n 256 -c 128 -H 28 -W 28 -k 128 -x 3 -y 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -g 1
```

(that exact set of parameters is why I ended up not doing the swizzling on one big vector/all at once)

jungpark-mlir: Performance is not changing.
I've noticed that there are few convolutions sometimes run very slowly. including (x2~x4 slower than usual)
with or without this PR. haven't got chance to investigate it yet but includes these

bwd,f16,True,kcyx,nchw,nkhw,256,256,14,14,256,3,3,1,1,1,1,1,1,4.35545105968997
wrw,f32,True,kcyx,nchw,nkhw,256,512,16,16,512,3,3,1,1,2,2,0,0,5.58864260967153,

Excepting that average performance is 0.2% better but I interpret it's just the same.

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
sjw36: Wouldn't **merge()** be better

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -522,15 +667,96 @@ template <typename T> struct MILARewritePattern : public OpRewritePattern<T> {
...
 };
 
-//===- Passes -------------------------------------------------------------===//
-//===- MIOpenLinalgAlignPass - Align Tiling of Linalg Ops -----------------===//
+//===----------------------------------------------------------------------===//
+// ThreadwiseCopyV2 Conversion in Fusion.
+//===----------------------------------------------------------------------===//
+
+struct ThreadwiseCopyV2RewritePattern
+    : public OpRewritePattern<miopen::ThreadwiseCopyV2Op> {
```
sjw36: You should have the ThreadwiseCopyV2 already in the previous pass, is there some missing info that prevents this from applying then. I would prefer a function instead of retracing the whole graph.

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
jungpark-mlir: merge works but it's actually ignoring the dimensions. I still can't find existing set of functions. I though top down ct builder should be used for this case but passThrough() is missing. Maybe I need to call addTransform?

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jungpark-mlir: cloneWith just needs the same arguments, only shape is optional arg but this is changing shape.

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
jungpark-mlir: After have a talk with @sjw36, this is only feeding the result into linalg so better to keep as it is so that we can avoid unnecessary time for lowering/optimizing dummy dimensions.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2561,19 +2577,40 @@ struct GridwiseGemmV2RewritePattern
       //     + m_i * MPerXdlops; return MatrixIndex{row, col};
       // }
       //
-      int64_t xdlops_i_xdlops_gemm = iter / NumBlksPerXdlops;
-      int64_t j_xdlops_gemm = iter % NumBlksPerXdlops;
-      int64_t m_i_xdlops_gemm = xdlops_i_xdlops_gemm / NRepeats;
-      int64_t n_i_xdlops_gemm = xdlops_i_xdlops_gemm % NRepeats;
+      // int64_t xdlops_i_xdlops_gemm = iter / NumBlksPerXdlops;
```
sjw36: can remove these comments

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2603,11 +2640,18 @@ struct GridwiseGemmV2RewritePattern
         //     index_t col = col_blk * mfma_type.n + blk_td + n_i * NPerXdlops;
         threadMtxColInBlock = blk_td_xdlops_gemm;
       }
-      int64_t thread_mtx_on_blk_col_const =
-          col_blk_xdlops_gemm * n + n_i_xdlops_gemm * NPerXdlops;
+
+      auto n_ConstantOp = b.create<arith::ConstantIndexOp>(loc, n);
+      auto NPerXdlops_ConstantOp =
+          b.create<arith::ConstantIndexOp>(loc, NPerXdlops);
+      // thread_mtx_on_blk_col = threadMtxColInBlock +
```
sjw36: and these

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2624,11 +2668,17 @@ struct GridwiseGemmV2RewritePattern
                              b.create<arith::RemUIOp>(loc, blk_td_xdlops_gemm,
                                                       group_size_ConstantOp));
       }
-      int64_t thread_mtx_on_blk_row_const =
-          row_blk_xdlops_gemm * m + m_i_xdlops_gemm * MPerXdlops;
+      auto m_ConstantOp = b.create<arith::ConstantIndexOp>(loc, m);
+      auto MPerXdlops_ConstantOp =
+          b.create<arith::ConstantIndexOp>(loc, MPerXdlops);
+      // thread_mtx_on_blk_row = threadMtxRowInBlock +
```
sjw36: and these

sjw36 (approved):
Thanks for moving back into one rewriter. Looks good, just remove some commented code.

jungpark-mlir:
Hopefully everything commented have been addressed. Please have a look and let me know any change needed.

sjw36:
Looks good. Let's have it.

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -200,6 +202,136 @@ template <typename T> struct MILARewritePattern : public OpRewritePattern<T> {
...
+
+    SmallVector<int64_t, 1> collapsedShape;
+    SmallVector<ReassociationExprs, 1> reassocs;
+    collapsedShape.push_back(collapsedDim);
+    reassocs.push_back(exprs);
+
+    auto collapsedType =
+        MemRefType::get(collapsedShape, sourceType.getElementType(), {}, 5);
+    Value result =
+        b.create<memref::CollapseShapeOp>(loc, collapsedType, source, reassocs);
```
krzysz00, line 228: Are we not doing this with transforms? Or is that a future PR?

jungpark-mlir: Not doing it with transform because it's only consumed by linalg.

PR #577, opened 2022-03-29
[DO NOT SQUASH] Separate call-graph roots from kernels;  main() calls roots, kernels …

pcf000:
…get wrapped

* Add bufferization dialect to loaded dialects.
* Use a special 'wrapped_call' attribute to avoid replacing calls that are already wrapped.
* Control the print options for roots after making the CallOp.  Remove wrapped_call attribute.
* Don't count a created CPUConvFunc as a kernel, because those get GPU-wrapped.  It'll be a root.
* Recognise when a root is a kernel, and call with localVars instead of validation's valVars.
* Refactoring the emitPrintTensor flag flag.
* Let's not count -fut as a kernel.
* Sort the roots by name, because sometimes order matters and DenseSet isn't deterministic.
* Clean up root/kernel and printp predicates.

### DIFF
--- a/mlir/tools/miopen-gen/miopen-gen.cpp
+++ b/mlir/tools/miopen-gen/miopen-gen.cpp
@@ -10,6 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "mlir/Analysis/CallGraph.h"
 #include "mlir/Conversion/MIOpenPasses.h"
 #include "mlir/Conversion/MIOpenToGPU/MIOpenToGPU.h"
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
@@ -41,6 +42,7 @@
 #include "mlir/Support/FileUtilities.h"
 #include "mlir/Support/LogicalResult.h"
 
+#include "llvm/ADT/SetVector.h"
 #include "llvm/IR/Constants.h"
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/InitLLVM.h"
@@ -799,7 +801,8 @@ static FuncOp createGPUWrapper(ModuleOp &module, const KernelIF &kernel) {
   }
 
   // Emit kernel function call.
-  b.create<CallOp>(loc, kernel.func, paramCasts);
+  auto wrappedCall = b.create<CallOp>(loc, kernel.func, paramCasts);
+  wrappedCall->setAttr("wrapped_call", b.getUnitAttr());
 
   i = 0;
   for (auto &param : paramCasts) {
@@ -1285,34 +1288,32 @@ static void emitMemcpy(OpBuilder &b, mlir::Value src, mlir::Value dst) {
   b.create<CallOp>(loc, memcpyFunc, ValueRange{srcFlat, dstFlat, cstSize});
 }
 
-static void emitPrintTensor(OpBuilder &b, mlir::Value var, bool flag = true) {
-  if (flag) {
-    auto loc = b.getUnknownLoc();
-    auto varType = var.getType().template dyn_cast<MemRefType>();
-    auto elemType = varType.getElementType();
-    auto floatType = b.getF32Type();
+static void emitPrintTensor(OpBuilder &b, mlir::Value var) {
+  auto loc = b.getUnknownLoc();
+  auto varType = var.getType().template dyn_cast<MemRefType>();
+  auto elemType = varType.getElementType();
+  auto floatType = b.getF32Type();
 
-    // get print func
-    mlir::Value pvar = var;
-    if (elemType != floatType) {
-      // make copy
-      auto pvarType = MemRefType::get(varType.getShape(), floatType);
-      pvar = b.create<memref::AllocOp>(loc, pvarType);
-      emitMemcpy(b, var, pvar);
-    }
+  // get print func
+  mlir::Value pvar = var;
+  if (elemType != floatType) {
+    // make copy
+    auto pvarType = MemRefType::get(varType.getShape(), floatType);
+    pvar = b.create<memref::AllocOp>(loc, pvarType);
+    emitMemcpy(b, var, pvar);
+  }
 
-    auto module = b.getBlock()->getParentOp()->getParentOfType<ModuleOp>();
-    auto unrankedMRType = UnrankedMemRefType::get(b.getF32Type(), 0);
-    auto printFunc = makeFuncDecl(module, "print_memref_f32", {unrankedMRType});
+  auto module = b.getBlock()->getParentOp()->getParentOfType<ModuleOp>();
+  auto unrankedMRType = UnrankedMemRefType::get(b.getF32Type(), 0);
+  auto printFunc = makeFuncDecl(module, "print_memref_f32", {unrankedMRType});
 
-    // Emit cast + call print
-    auto printCast = b.create<memref::CastOp>(loc, unrankedMRType, pvar);
-    b.create<CallOp>(loc, printFunc, ValueRange{printCast});
+  // Emit cast + call print
+  auto printCast = b.create<memref::CastOp>(loc, unrankedMRType, pvar);
+  b.create<CallOp>(loc, printFunc, ValueRange{printCast});
 
-    if (pvar != var) {
-      // dealloc pvar
-      b.create<memref::DeallocOp>(loc, pvar);
-    }
+  if (pvar != var) {
+    // dealloc pvar
+    b.create<memref::DeallocOp>(loc, pvar);
   }
 }
 
@@ -1654,6 +1655,7 @@ static LogicalResult populateTensorFillLogic(mlir::OpBuilder &b,
 static LogicalResult
 populateHostHarnessLogic(ModuleOp &module,
                          const SmallVector<KernelIF, 8> &kernels,
+                         const SmallVector<KernelIF, 8> &roots,
                          const mlir::Conv2dGenerator::Config &genConfig) {
 
   auto context = module.getContext();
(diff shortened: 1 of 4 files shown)


mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -799,7 +800,8 @@ static FuncOp createGPUWrapper(ModuleOp &module, const KernelIF &kernel) {
   }
 
   // Emit kernel function call.
-  b.create<CallOp>(loc, kernel.func, paramCasts);
+  auto wrappedCall = b.create<CallOp>(loc, kernel.func, paramCasts);
+  wrappedCall->setAttr("wrapped_call", b.getUnitAttr());
```
pcf000, line 805: The previous version emits wrapped calls while building main(), which is fine when there are only immediate kernel calls.  This version scans the whole module and changes calls to wrapped functions to instead call the wrapper.  Since the wrapper functions themselves are in the module, they're also scanned, and we don't want to replace their calls.

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -1285,34 +1287,32 @@ static void emitMemcpy(OpBuilder &b, mlir::Value src, mlir::Value dst) {
   b.create<CallOp>(loc, memcpyFunc, ValueRange{srcFlat, dstFlat, cstSize});
 }
 
-static void emitPrintTensor(OpBuilder &b, mlir::Value var, bool flag = true) {
-  if (flag) {
```
pcf000, line 1289: A boolean parameter that controls the execution of a function offends my aesthetic sensibilities.  I removed it entirely and instead controlled whether emitPrintTensor() is called.

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -1654,6 +1654,7 @@ static LogicalResult populateTensorFillLogic(mlir::OpBuilder &b,
 static LogicalResult
 populateHostHarnessLogic(ModuleOp &module,
                          const SmallVector<KernelIF, 8> &kernels,
+                         const SmallVector<KernelIF, 8> &roots,
```
pcf000, line 1658: Call-graph roots are separate from kernels, though a function may be in both sets.

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -1706,13 +1707,16 @@ populateHostHarnessLogic(ModuleOp &module,
 
   // Create all local variables for each kernel param
   // - assumes all kernels read the same memrefs
-  auto kernel0 = kernels.front();
-  bool isCPUKernel = !kernel0.func->hasAttr("kernel");
+  if (roots.size() > 1) {
+    // TODO: verify that all parameter lists match
```
pcf000, line 1712: The assumption is made, but not checked.

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -1774,25 +1778,56 @@ populateHostHarnessLogic(ModuleOp &module,
...
-    if (kernel.func->hasAttr("kernel")) {
-      auto kernelWrapperFunc = createGPUWrapper(module, kernel);
-      b.create<CallOp>(loc, kernelWrapperFunc, localVars);
-    } else {
-      if (!valVars.empty()) {
-        b.create<CallOp>(loc, kernel.func, valVars);
+  // Call the roots.
+  for (auto &root : roots) {
+    // Is the root also a kernel?
+    bool rootKernel = std::find_if(kernels.begin(), kernels.end(),
```
pcf000: Hedging bets here, if a function without a "kernel" attribute can be flagged as a kernel.  I think it's an incomplete consideration of how to handle -fut.  The chief purpose here is to ensure that a non-validation function is called with non-validation arguments.

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -1993,8 +2028,7 @@ int main(int argc, char **argv) {
 
   if (!hasUserKernel) {
     if (genCPUKernel.getValue()) {
-      auto func = createCPUConvFunc(module, genConfig);
```
pcf000, line 1996: We don't want to GPU-wrap a CPU function.  Technically it'll be a kernel, but "kernel" conflates both "computational kernel" and "function run on GPU."

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -2024,25 +2058,50 @@ int main(int argc, char **argv) {
...
+
+  // Make KernelIFs for the roots, to pass to populateHostHarnessLogic().
+  SmallVector<KernelIF, 8> rootIFs;
+  for (auto node : roots) {
+    FuncOp func = dyn_cast<FuncOp>(node->getCallableRegion()->getParentOp());
+    rootIFs.emplace_back(func);
+  }
+
+  // Sort by name to be deterministic (because DenseSet above).
+  llvm::sort(rootIFs, [&](KernelIF &a, KernelIF &b) {
```
pcf000: There's a test that generates sub-kernels with names including "_0_", "_1_", and "_2_", and they pass results as arguments to the next one so they have to be called in the right order.  DenseSet, used to generate the roots, is not necessarily in the right order.  Could also sort by location, but it doesn't have a native ordering and converting to opaque pointer felt crude.

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -2024,25 +2058,50 @@ int main(int argc, char **argv) {
...
+                      });
+
   if (testFuncNameVal.empty()) {
     mod.walk([&](FuncOp func) -> WalkResult {
       if (func->hasAttr("kernel")) {
         kernels.emplace_back(func);
       }
       return WalkResult::advance();
     });
-  } else {
```
pcf000, line 2034: Don't treat -fut as a kernel.  If it's naturally one, it'll be one.  If it's naturally a root, it'll be one.  I'm not sure which cases -fut gives us now.

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -1774,25 +1778,56 @@ populateHostHarnessLogic(ModuleOp &module,
...
 
+  // Wrap the kernels and gather them to substitute in calls.
+  llvm::SmallDenseMap<FuncOp, FuncOp> wrappedFuncs;
+  for (auto &kernel : kernels)
+    wrappedFuncs[kernel.func] = createGPUWrapper(module, kernel);
+
+  // Redirect calls to kernel functions to point at wrapped functions.
+  module.walk([&](CallOp call) -> WalkResult {
+                // Don't substitute the call inside the wrapper.
+                if (call->hasAttr("wrapped_call")) {
```
sjw36: Seems like you would only encounter a CallOp once. Is the attribute necessary?

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -1774,25 +1778,56 @@ populateHostHarnessLogic(ModuleOp &module,
...
+      b.create<CallOp>(loc, root.func, localVars);
+      if (!root.func->hasAttr("kernel")) {
         printValidationResults.setValue(false);
         printResults.setValue(true);
       }
     }
   }
 
+  // Wrap the kernels and gather them to substitute in calls.
+  llvm::SmallDenseMap<FuncOp, FuncOp> wrappedFuncs;
```
sjw36, line 1807: This could be done as a post-process after populateHostHarness. To separate the processes and keep this more simple.

mlir/tools/miopen-gen/miopen-gen.cpp
krzysz00: Would putting together the actual def-use graph be too much effort?

krzysz00: Or does this not matter for the actual IR but only for printing it?

mlir/tools/miopen-gen/miopen-gen.cpp
sjw36: Then this func would not need to know about kernels.

mlir/tools/miopen-gen/miopen-gen.cpp
pcf000: If we have f() which calls g(), and g() is a kernel, then createGPUWrapper() will make a function g_gpu() which calls g().  g_gpu() is present in the module by the time we do the walk.  I want to replace the call to g() inside f() with a call to g_gpu().  I do not want to replace the call to g() inside g_gpu().

More pictorially:
1.  F -> G         (originally)
2. F -> G ; G_GPU -> G      (after createGPUWrapper())
3. F -> G_GPU ; G_GPU -> G     (after module.walk;  we don't want G_GPU->G_GPU)

mlir/tools/miopen-gen/miopen-gen.cpp
pcf000: The def-use graph should be easy, given the underlying SSA-ness of llvm, and that would certainly be more generally correct.  Let me look into that.

mlir/tools/miopen-gen/miopen-gen.cpp
pcf000: We could probably stand to split kernel generation out from host-code generation, which may help accomplish the same thing.  Probably in another pull request, though.

mlir/tools/miopen-gen/miopen-gen.cpp
pcf000: On further consideration, I think the right idea is simply to preserve the order of the roots.  The original problem case was a generated kernel, which made three kernels named _0, _1, and _2, so sorting by name kept the generated order.

Since the code before me called them in original order, and miopen-gen is not a general tool, preserving that order will at least not break anything new.

mlir/tools/miopen-gen/miopen-gen.cpp
sjw36: That's fine.

mlir/tools/miopen-gen/miopen-gen.cpp
sjw36: There could be many kernels and many roots in the file, but suppose we only want to test 1 of them.

pcf000: Yeah, I don't have a good answer for that yet.  I think we want to make -fut be a root, and the only root, but not a kernel unless it has the "kernel" attribute or there are no other kernels.

mlir/tools/miopen-gen/miopen-gen.cpp
pcf000: Changed DenseSet to SetVector to keep the original order, and then I can remove the sort() call entirely.

mlir/tools/miopen-gen/miopen-gen.cpp
pcf000: I think it also makes sense for -fut to be mutually exclusive with kernel generation.

krzysz00:
FYI, you're failing clang-format

sjw36 (approved):
Looks good. Please file tickets for follow-up enhancements (specifically -fut usage).

sjw36 (changes requested):
I just noticed that this has changes in the LLVM subtree. Those should be in a separate commit and then merged without squash.

pcf000:
Thanks for the reminder;  now split.

sjw36 (approved):
Looks good.

PR #578, opened 2022-03-29
Revert "Enable KPACK for backward weight convolution. (#551)"

krzysz00:
This reverts commit 395e80539e3128362b3a1ec6659aff81df597db2.

This commit causes the following kernel code to become incorrect:

```
./bin/miopen-gen "--conv-config=--x2 1 --operation conv2d_bwd_weight --num_cu 120 --arch amdgcn-amd-amdhsa:gfx908:sramecc+:xnack- --groupsize 1 --fil_layout GNCHW --fil_type fp32 --in_layout NGCHW --out_layout NGCHW --in_type fp32 --out_type fp32 --batchsize 256 --in_channels 1024 --out_channels 256 --in_h 14 --in_w 14 --out_h 14 --out_w 14 --fil_h 1 --fil_w 1 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h 0 --padding_w 0 --kernel_name mlir_gen_igemm_conv2d_v4r4_wrw_xdlops1 --perf_config 128,128,8,64,64,4,1,1" -pv | ./bin/mlir-miopen-driver -c | rocm-run | tail
```

with outputs
```
Values: 618.312500, 30314.312500
Values: 29696.000000, 30314.312500
Values: -298.875000, -14634.875000
Values: -14336.000000, -14634.875000
Values: -298.000000, -14634.000000
Values: -14336.000000, -14634.000000
Values: 618.312500, 30314.312500
Values: 29696.000000, 30314.312500
Unranked Memref base@ = 0x13e43a0 rank = 1 offset = 0 sizes = [1] strides = [1] data =
[0]
```

As far as I'm concerned, we should merge this unless we find the bug or we can conclusively prove no one will ever try to run this config (the CI failures on #571 show that the PR in MIOpen isn't doing that)

Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/490

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
@@ -807,10 +807,9 @@ public:
 
 class PopulateParamsXDL : public PopulateParamsBase {
 private:
-  // Initial tuning parameters for forward convolution and backward weight
-  // convolution.
+  // Initial tuning parameters for forward convolution.
   // clang-format off
-  llvm::SmallVector<InitParamsXDL, 4> initParametersFwdAndBwdWeight = {
+  llvm::SmallVector<InitParamsXDL, 4> initParametersForward = {
       // M/block N/block K/block M/wave N/wave kPack aCopyMore bCopyMore
       {128, 128, 4, 64, 64, 4, false, false},
       {32, 64, 4, 32, 64, 4, false, false},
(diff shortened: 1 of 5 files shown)


whchung:
Instead of revert, how about we change the heuristic tuning parameter set for fp32 backward weight?

krzysz00:
Please explain why kPack = 4 is not a valid tuning parameter and why it should be excluded from the search space, aside from "there's a bug in 551 that makes it not work"

I'd prefer to not reject values of tuning parameters unless we can demonstrate that they don't work for mathematical or architectural reasons

Barring some evidence that this is either not a valid input to the convolution generator or that we're hitting a bug outside of our control, I believe we should revert and investigate.

whchung:
Per discussion offline, I'll submit a hot fix to:
- reject KPACK=4 for fp32 backward weight convolution in the applicability check
- change the heuristic tuning parameter vector for fp32 backward weight convolution so KPACK=4 won't be supplied.

Meanwhile I'll continue investigate the reason why setting KPACK=4 introduces such issue. We track it at https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/490

PR #579, opened 2022-03-30
[HOTFIX] Reject and avoid KPACK=4 for fp32 XDLOPS backward weight convolution

whchung:
For fp32 XDLOPS backward weight convolutions:
- Reject KPACK=4.
- The heuristic tuning parameter vector avoids KPACK=4.

Combined with MIOpen PR #1474 ( https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1474 ) which alters the tuning search space and the default perf DB, this hot fix could temporarily get around intermittent failures in fp32 XDLOPS backward weight convolutions when `KPACK=4`, specifically the tuning parameter set `128,128,8,64,64,4,1,1`.

Will continue root cause the issue over https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/490

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
@@ -857,6 +857,23 @@ private:
       {16, 16, 4, 16, 16, 1, false, false},
   };
   // clang-format on
+
+  // XXX FIXME: Initial tuning parameters for fp32 backward weight convolution.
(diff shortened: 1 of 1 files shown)


krzysz00 (approved):
On the assumption that 1471 will prevent us from spanning the tuning logs with that error message (please do the `LLVM_DEBUG` thing otherwise) and since this is important, approved

whchung:
I'm thinking about using `emitError()` to allow more fine-grained negative tests, but that requires some overhaul in the existing tuning logic where we need to pass in some MLIR-specific objects.

mvermeulen (commented):
LGTM

PR #580, opened 2022-03-31
Add NHWC to performance results

krzysz00:
Also fix some typos in the script

Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/441

Note: local testing shows that MIOpen will give absolutely terrible TFlops numbers for NHWC, and so we'll get skewed results. We may not want to actually merge this, but I'm tossing it up so it's there.

### DIFF
--- a/mlir/utils/jenkins/MIOpenDriver.py
+++ b/mlir/utils/jenkins/MIOpenDriver.py
@@ -287,21 +287,18 @@ def benchmarkMLIR(commandLine, xdlops):
     nanoSeconds = getNanoSeconds(BENCHMARKING_RESULT_FILE_NAME)
     return config.tableEntry(nanoSeconds)
 
-def benchmarkMIOpen(commandLine, xdlops, envs=dict(), skip=False):
+def benchmarkMIOpen(commandLine, xdlops, envs=dict()):
     config = ConvConfiguration.fromCommandLine(commandLine, xdlops)
-    if not skip:
-        runConfigWithMIOpenDriver(commandLine, envs)
-        # get nanoseconds from rocprof output.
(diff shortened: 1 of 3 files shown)


mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -388,6 +383,8 @@ def tuneMLIRKernels(configs, xdlops):
     if len(sys.argv) == 1:
         # batch benchmark with MLIR and MIOpen.
         generatePerformanceResults(configs, xdlops)
+    elif sys.argv[1] == '-bmlir_only':
+        generatePerformanceResults(configs, xdlops, skipMIOpen=True)
```
yiqian1: This is the same as the option "-b" ?

yiqian1:
I also noticed that MIOpen shows much worse performance for NHWC than MLIR, and that's most for non-xdlops. For xdlops, MLIR is still slower for most configs. 

Also, if we do want to merge this, it might be helpful to group the results by layout in the performance report.

mlir/utils/jenkins/MIOpenDriver.py
krzysz00: No, because it'll generate performance reports acceptable to the report generator? 

Although I'll go and test if the output of -b goes through the perf report maker without problems

krzysz00:
Agreed on the grouping thing, this was just a first pass

yiqian1:
LGTM

krzysz00:
@yiqian1 Can I get an approval, or do you want to see some extra testing first?

yiqian1 (approved):

PR #582, opened 2022-03-31
Use index diffs everywhere

krzysz00:
The lack of index diff use on padding kernels was partly due to an
llvm issue that has been fixed. Another part of the problem was that
removing the loadStoreHack caused certain wrw kernels to fail in the
non-xdlops case. What I think happened there is that unfold() didn't
properly propagate index updates that would have required writes to be
dropped. For example, if we had c = 255, y = x = 3 and we started at
(gemmN = 254 * 9 + 2 * 3 + 2) -> (254, 2, 2), using index diff maps to
compute gemmN + 1 would give (254, 2, 3), which would pass the oob
checker and be written, though the correct coordinate, (255, 0, 0)
would be oob.

Therefore, we disable unfold() in certain padding kernels, allowing us
to enable index diffs everywhere.

Fixes
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/440

I can confirm that I've run non-xdlops, CPU verified tests on `lockhart5` and that they all passed.

[index_diff_enablement_impacts.html.txt](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/files/8393207/index_diff_enablement_impacts.html.txt) shows the perf impact from this change is generally flat (the sort of +- 1% that's benchmarking noise) except on some fp32 wrw kernels.

There's an interesting regression in one backwards data kernel that I'm curious about, but this might be good to go overall


krzysz00:
Update, that 30% drop on the one backward data config is very very weird - both `miopen-dialect` and this branch produce identical inputs to LLVM and yet there's a performance difference

whchung:
One potential reason about the performance difference is the clock frequency may be throttle down automatically by the system under low loading situation. I was also trying to figure out why nightly CI shows varying perf results between runs.
`rocm-smi --setsclk` and `rocm-smi --setmclk` could lock the frequency. 
https://rocmdocs.amd.com/en/latest/ROCm_System_Managment/ROCm-SMI-CLI.html

whchung (approved):
LGTM! Thanks for another great clean up!

PR #583, opened 2022-04-01
[DO NOT SQUASH] Supress hip semicolon warnings

krzysz00:
This removes tens of warnings from build logs that we can't do
anything about.

Differential Revision: https://reviews.llvm.org/D122927

(I suspect this is the sort of thing we can stick upstream on our own initiative without anyone complaining)


pcf000 (approved):
Looks good from here.  I had been searching for a way to do it from the cmake command, but this is better.

Can I just say "LGTM" over at reviews.llvm.org?

PR #584, opened 2022-04-05
Make bwd data kernel alwasy pad GEMM on the right, even for bwd data xdlops

krzysz00:
Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/288 , https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/289, https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/290 since it looks like the underlying compiler issue is gone.

Nightlies passed on MI-200.

This PR appeared to improve the performance of two fp16 NHWC kernels significantly, but, if that was a fluke, the overall conclusion is that performance is not degraded by this change. 
[bwd-data-changes-perf.html.txt](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/files/8420555/bwd-data-changes-perf.html.txt)

(skip-ci is on account of dependencies)


whchung (approved):
LGTM

PR #585, opened 2022-04-06
[Don't merge] Dump tuning

jungpark-mlir:
experimental PR to steal tuning parameters.

will delete this soon.


jerryyin:
@jungpark-mlir This isn't a good spend of our CI time. I'll do a presentation tmr on our weekly meeting about tuning parameters. Then we can meet if you have any questions from that point.

jungpark-mlir:
@jerryyin You're right, I'm just desperately looking for a piece of optimistic news before going away for a week. Also trying to intercept and run only necessary part but it's not accessible at the moment. We'll see. And of course looking forward to your presentation.

PR #587, opened 2022-04-07
[DO NOT SQUASH] Use LLVM dialect inline_asm op for LDS barrier.

whchung:
Currently `rocdl.lds_barrier` creates an LLVM fence with release ordering and workgroup syncscope. It would still introduce a `s_waitcnt vmcnt(0)` within the hot loop of a convolution kernel. From thread trace it can be seen we have stalled holes due to `s_waitcnt vmcnt(0)`.

In this PR, we reinstate using inline assembly for LDS barrier.

- Remove `rocdl.lds_barrier` and its lowering logic.
- Modify `gpu.lds_barrier` and use LLVM dialect `inline_asm` op to convey the
  instruction sequence we want in the hot loop:

```
  s_waitcnt lgkmcnt(0)
  s_barrier
```

Empirical testing shows 8~10% performance improvement in resnet50 configs.

As denoted at https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/503, a follow up ticket could be created to modify the behavior of the backend so we fallback to using intrinsics but still retain the same assembly output where no `s_waitcnt vmcnt(0)` is emitted. This requires more thorough understanding about the behavior in the backend.

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
@@ -116,17 +116,6 @@ def ROCDL_BarrierOp : ROCDL_Op<"barrier"> {
   let assemblyFormat = "attr-dict";
 }
 
-def ROCDL_LDSBarrierOp : ROCDL_Op<"lds_barrier"> {
-  // Guarantees LDS update until this point are all completed.
-  string llvmBuilder = [{
-    llvm::LLVMContext &llvmContext = builder.getContext();
-    builder.CreateFence(llvm::AtomicOrdering::Release,
-                       llvmContext.getOrInsertSyncScopeID("workgroup"));
-    createIntrinsicCall(builder, llvm::Intrinsic::amdgcn_s_barrier);
-  }];
-  let assemblyFormat = "attr-dict";
-}
-
 //===---------------------------------------------------------------------===//
 // Xdlops intrinsics
 
--- a/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
+++ b/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
@@ -722,14 +722,11 @@ struct SoftwareBF16Trunc : OpRewritePattern<LLVM::FPTruncOp> {
   }
 };
 
-struct LDSBarrierOpLowering : ConvertToLLVMPattern {
-  explicit LDSBarrierOpLowering(MLIRContext *context,
-                                LLVMTypeConverter &typeConverter)
-      : ConvertToLLVMPattern(gpu::LDSBarrierOp::getOperationName(), context,
-                             typeConverter) {}
+struct LDSBarrierOpLowering : public ConvertOpToLLVMPattern<gpu::LDSBarrierOp> {
+  using ConvertOpToLLVMPattern<gpu::LDSBarrierOp>::ConvertOpToLLVMPattern;
 
   LogicalResult
-  matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+  matchAndRewrite(gpu::LDSBarrierOp op, gpu::LDSBarrierOp::Adaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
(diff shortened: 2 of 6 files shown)


external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
```diff
@@ -720,6 +720,31 @@ struct SoftwareBF16Trunc : OpRewritePattern<LLVM::FPTruncOp> {
     return success();
   }
 };
+
+struct LDSBarrierOpLowering : ConvertToLLVMPattern {
+  explicit LDSBarrierOpLowering(MLIRContext *context,
+                                LLVMTypeConverter &typeConverter)
+      : ConvertToLLVMPattern(gpu::LDSBarrierOp::getOperationName(), context,
```
krzysz00: These days, there's a `ConvertOpToLLVMPattern` that provides better ergonomics and is preferred

external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
```diff
@@ -797,6 +822,7 @@ void mlir::populateGpuToROCDLConversionPatterns(
       GCNRawBufferOpLowering<gpu::GCNRawBufferStoreOp, ROCDL::RawBufferStoreOp>,
       GCNRawBufferOpLowering<gpu::GCNRawBufferAtomicFaddOp,
                              ROCDL::RawBufferAtomicFAddOp>>(converter);
+  patterns.insert<LDSBarrierOpLowering>(ctx, converter);
```
krzysz00: Add to the list of things being `.add()`ed above after using the OpPattern

krzysz00 (commented):
Minor style nits

external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
whchung: Adopted the pattern in the latest commit.

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
```diff
@@ -116,17 +116,6 @@ def ROCDL_BarrierOp : ROCDL_Op<"barrier"> {
   let assemblyFormat = "attr-dict";
 }
 
-def ROCDL_LDSBarrierOp : ROCDL_Op<"lds_barrier"> {
```
krzysz00, line 119: Quick sanity check, is this op upstream or not?

krzysz00 (approved):
LGTM

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
whchung: It's not in upstream.

whchung:
SI insert waitcnt pass in LLVM AMDGPU backend is doing things relatively conservatively:

Basically it tracks pending memory instructions, and:

- Upon seeing a `s_barrier`, it would modify the most recent `s_waitcnt` to have all outstanding memory ops to complete.
- Upon seeing an instruction which depends on an outstanding memory op, it would emit a new `s_waitcnt` on all outstanding memory ops to complete.

There are 2 issues in this approach:

- Amending `s_waitcnt` to force all outstanding memory ops to complete is too strong a guarantee for an `s_barrier`. The outstanding memory ops may not be immediately required in the instructions after the `s_barrier`.
- SI insert waitcnt pass doesn't consider the latency between the producers and consumers of memory ops, so it has to force emit `s_waitcnt` to force wait on outstanding memory ops. This may create stalled gaps in high-performance kernels.

We can't say the approach is incorrect, as they would render correct codes in all cases. But for high-performance kernels they indeed impede performance.

The problem is now somewhat clear. How to derive a proper fix instead of getting around it via inline asm is unknown yet.

whchung:
https://reviews.llvm.org/D120544 introduces a subtarget feature `FeatureBackOffBarrier`, but it's not available on gfx908. Plus it only solves 1 of the problem within a GEMM / conv hot loop.

PR #588, opened 2022-04-07
Make buffer load read potential garbage and then replace it with 0

krzysz00:
In order to reduce the number of instructions used in address
computation and to give the processor something to do while it waits
for values to come back, have buffer_load perform its loads even if
they're going to a bad address (since actual out of bounds access will
be rejected in hardware) and then replace oob values with zero.

This is inspired by @whchung's note about trying not to schedule consecutive buffer loads and a conversation I had with @jerryyin. Inspection of ISA shows that the kernel has definitely gotten shorter, but benchmark runs done on 112 are being rather unclear about whether it's any *faster* (some configs seem to be faster, some slower, maybe it's all a matter of tuning).

I'll try and get some results from MI-200 tomorrow (someone went and upgraded the ROCm version on lockhart5 in a way that apparently made /dev/dri go away)

### DIFF
--- a/mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
+++ b/mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
@@ -239,30 +239,39 @@ struct TransformingForRewritePattern
 //===----------------------------------------------------------------------===//
 // BufferLoad lowering.
 //===----------------------------------------------------------------------===//
-// TODO(kdrewnia): use "OOB reads = 0" from hardware to remove
-// hardcoded zero value
 struct BufferLoadRewritePattern : public OpRewritePattern<BufferLoadOp> {
   using OpRewritePattern<BufferLoadOp>::OpRewritePattern;
   LogicalResult matchAndRewrite(BufferLoadOp op,
                                 PatternRewriter &b) const override {
     Location loc = op.getLoc();
     Value source = op.source();
+    ValueRange coords = op.coords();
     auto sourceType = source.getType().cast<MemRefType>();
     ArrayRef<int64_t> sourceShape = sourceType.getShape();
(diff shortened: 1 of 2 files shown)


whchung:
Let's get some perf numbers on this one.

whchung:
Let's get some perf numbers on this one.

krzysz00:
I think this one can hang out in a holding pattern for a while, especially since perf was weak

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
```diff
@@ -239,30 +239,39 @@ struct TransformingForRewritePattern
...
+    // If the index here is outside the range of the underlying buffer,
+    // the load will return 0. If indexing produces a final index inside the
+    // buffer but some of the coordinates are out of bounds (ex, loading from an
+    // h or w coordinate that corresponded to implicit padding), correct afther
+    // the fact with a zero
+
+    // This approach allows us to move bounds checks on loads to after each load
+    // which improves performance by reducing memory stalls. It also simplifies
+    // the adress computation logic on loads, enabling potential compiler
+    // optimizations.
```
jerryyin, line 264: I doubt the impact of address computation logic on loads is as impactful on a gpu. I actually don't know if gpu has branch predictor or how many pipelines it generally has at all.

On occasion of oob, I suspect it will:
 - consume more power because of the additional load
 - pollute l1/l2 cache because of the useless value

But those effect may not be apparent enough in a convolution kernel and again I don't know how much impact to the performance does branch check has.

krzysz00: That would explain the very not-great perf numbers I got here

krzysz00:
Not a very useful experiment and is clogging up PR list, closing

PR #581, opened 2022-03-31
Remove restrictions on backward data padding kernels

krzysz00:
The problem cases appear to work correctly now.

@whchung Do you have other examples of previously-broken cases?

PR #589, opened 2022-04-08
Partially revert "Remove restrictions on backward data padding kernels (#581)"

krzysz00:
This reverts commit ac776e68bbd1cfce0938897b499a50c066da8c40.

However, it doesn't revert the xdlops padding changes, since those
weren't the problem.


jerryyin (approved):
Approving based on urgency in fixing nightly. I'm not capable in reviewing the change in this/previous PR though.

whchung (approved):

PR #590, opened 2022-04-12
Make Jenkinsfile play nicer with CI cluster

krzysz00:
1. Limit execution to nodes with a `mlir` label

2. When a job fails, reboot the underlying node, since the failure is
_probably_ due to a GPU hang

3. Use timeouts to kill jobs that produce no output after 30 minutes,
indicating a hang


jerryyin (approved):
Nice cleanup assuming CI continue to pass. Thank you, LGTM!

okakarpa (approved):
awesome i have added MLIR label for lockhart nodes as well

PR #591, opened 2022-04-15
Protect outlineConvPartOps() from empty secondOps, and call it for al…

pcf000:
…l Conv2DOp regardless of adjacent elementwise ops.


sjw36 (approved):
Looks good.

PR #593, opened 2022-04-19
Make load/store positioning consistent in gridwise gemm

krzysz00:
Now, both in the body of the loops and in the loop header, the ops go
load A ; load B ; store A ; store B

Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/457


whchung (approved):
LGTM.

PR #594, opened 2022-04-19
Add missing support for optional attributes in MIGraphX to Tosa to MIOpen pipeline

jungpark-mlir:
xdlops and perf_config are propagated through.


sjw36 (approved):
Looks good.

PR #595, opened 2022-04-19
Rename coordinate transform builder to transform map builder

krzysz00:
This makes the code less confusing

Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/483


mlir/include/mlir/Dialect/MIOpen/MIOpen.h
```diff
@@ -83,14 +83,14 @@ namespace miopen {
 /// shape down or the lower shape up, such as passThrough() or pad(), "start" is
 /// the coordinate space you began building from and end is the coordinate space
 /// you're building towards (whether that's the upper or lower coordinates)
-class CoordTransformsBuilder {
+class TransformMapBuilder {
```
sjw36: Move to it's own file. Many things that include MIOpen do not need this.

sjw36 (commented):
Nice. Fewer characters to type.. except mostly using the derived classes anyway.. :)

Please move this out of MIOpen.h.

mlir/include/mlir/Dialect/MIOpen/MIOpen.h
krzysz00: Yeah, that's a reasonable ask since we're already here

krzysz00: Done

sjw36 (approved):
Looks good.

PR #596, opened 2022-04-19
Adjustments to buildbot support after some experience

pcf000:
* Change buildbot login name, run docker as daemon.
  The login name used at the master isn't the original one I'd submitted, so change here to match it.
* Use python 3.8 to include some newer libraries.
* Use '--group-add root' to work around /dev/kfd group glitch, update PyYAML, add python3.8-dev.
  /dev/kfd and /dev/dri are coming into the docker with group root instead of video or render, and I don't know why, but I can compensate.
* Add README with buildbot startup instructions.

### DIFF
--- a/mlir/utils/buildbot/README
+++ b/mlir/utils/buildbot/README
@@ -1,0 +1,7 @@
+To start up the buildbot:
+
+  git clone https://github.com/ROCmSoftwarePlatform/llvm-project-mlir.git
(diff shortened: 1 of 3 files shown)


pcf000:
Buildbot has been running on lockhart12, via an ssh tunnel from ixt-rack-112, for about ten days with no trouble, using these changes.

krzysz00 (approved):

PR #598, opened 2022-04-28
Fixing pad op lowering pipeline from subtree merge

jerryyin:
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/521

### DIFF
--- a/mlir/lib/Dialect/MIOpen/Pipeline/Pipeline.cpp
+++ b/mlir/lib/Dialect/MIOpen/Pipeline/Pipeline.cpp
@@ -46,6 +46,10 @@ void miopen::addHighLevelPipeline(PassManager &pm, bool toMIOpen) {
     pm.addPass(tosa::createTosaToMIOpenPass());
   }
   mlir::tosa::addTosaToLinalgPasses(pm);
+  bufferization::OneShotBufferizationOptions options;
+  options.allowReturnAllocs = true;
+  options.testAnalysisOnly = true;
+  pm.addPass(createLinalgComprehensiveModuleBufferizePass(options));
   pm.addPass(createLinalgElementwiseOpFusionPass());
   pm.addPass(createLinalgBufferizePass());
   pm.addNestedPass<mlir::FuncOp>(createTensorBufferizePass());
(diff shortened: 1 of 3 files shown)


mlir/lib/Dialect/MIOpen/Pipeline/Pipeline.cpp
```diff
@@ -46,6 +46,10 @@ void miopen::addHighLevelPipeline(PassManager &pm, bool toMIOpen) {
     pm.addPass(tosa::createTosaToMIOpenPass());
   }
   mlir::tosa::addTosaToLinalgPasses(pm);
+  bufferization::OneShotBufferizationOptions options;
+  options.allowReturnAllocs = true;
+  options.testAnalysisOnly = true;
```
sjw36, line 51: Why is testAnalysisOnly necessary. From the description it seems to suggest that it won't do any of the transforms.

mlir/lib/Dialect/MIOpen/Pipeline/Pipeline.cpp
```diff
@@ -46,6 +46,10 @@ void miopen::addHighLevelPipeline(PassManager &pm, bool toMIOpen) {
...
+  pm.addPass(createLinalgComprehensiveModuleBufferizePass(options));
   pm.addPass(createLinalgElementwiseOpFusionPass());
   pm.addPass(createLinalgBufferizePass());
```
sjw36, line 54: Many of these will not become unnecessary. Only bufferization::createBufferResultsToOutParamsPass is still needed for Bufferization.

sjw36 (approved):
Overall looks good but there is some redundancy now.

mlir/lib/Dialect/MIOpen/Pipeline/Pipeline.cpp
jerryyin: They are actually still useful. I did a binary search in attempting to removing the rest of those passes and they will result in unit test failures one way or the other. For the most part, I think comprehensive module bufferize pass is still orthogonal to those other passes.

mlir/lib/Dialect/MIOpen/Pipeline/Pipeline.cpp
jerryyin: By looking at their pass pipeline:

https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/579526e08448ce42d89528f75c870175c337c34b/external/llvm-project/mlir/lib/Dialect/Linalg/Transforms/ComprehensiveBufferizePass.cpp#L117-L124

Everything below `opt.testAnalysisOnly` will break MIOpen dialect and the reason is not yet clear to me. But I can confirm that everything before the escape is equivalent to previous pipeline.

PR #599, opened 2022-04-29
Test suite minimization, round 1

krzysz00:
- Adds a script to scrape our end to end tests and gather information
about each config
- Adds the results of analyzing that information to reveal duplicate
tests
- Removes duplicate tests
- Transitions our lowering pipeline to use LLVM_DEBUG instead of
commented out prints

One remaining question is if the tests that have rand-type=float but
are otherwise redundant in the host validation section are still needed

Followup work will be running this analysis again, but with less focus
on duplicate tests and more on catching gaps in our testsuite


yiqian1 (approved):
LGTM

PR #600, opened 2022-05-02
Fix handling of right padding

krzysz00:
When attempting to ensure that our convolution input always has enough
values to successfully apply the filter by implicitly adding padding,
we were failing to include the already-specified right padding in the
new right padding value.

This commit fixes that issue and adjusts the tests that depended on it.


krzysz00:
(currently running nightly tests on this change)

krzysz00:
Nightlies pass with this change

whchung (commented):
The new logic indeed addresses the issue in the existing logic.

@sjw36 it'd be good if you can take a look at this PR and have the final approval.

sjw36 (approved):
Looks good. ( as long as we have adequate coverage ..)

PR #601, opened 2022-05-02
Adding support of int8 inference in miopen_validate.sh

jerryyin:
This PR adds the following capability:
 - inference run for convolution configs
 - int8 data type

To run the script in int8 inference:
```bash
 bash ~/llvm-project-mlir/mlir/utils/jenkins/miopen-tests/miopen_validate.sh --test-inf --dtype int8 --tuning  < ~/llvm-project-mlir/mlir/utils/jenkins/miopen-tests/resnet50-miopen-configs
```

TODO:
 - Adding int8 MIOpen (MLIR solver) support to its branch
 - Add int8 inference tuning to weekly test/ selected configs test

### DIFF
--- a/mlir/utils/jenkins/miopen-tests/miopen_validate.sh
+++ b/mlir/utils/jenkins/miopen-tests/miopen_validate.sh
@@ -5,7 +5,7 @@ declare -g TMPFILE
 declare -ig XDLOPS=-1
 declare -ig TUNING=0
 declare -ig TESTALL=0
-declare -ig TESTINF=0
+declare -ig TESTFWD=0
 declare -g DTYPE=""
 declare -g DIRECTION=""
 declare -g LAYOUT=""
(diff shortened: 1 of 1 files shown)


krzysz00 (commented):
Wouldn't it make more sense for this to be called `--test-fwd`? Or, more generally, you could just call the script twice - once with int8, nchw, fwd and once with int8, nchw, fwd

I'm not sure we need this extra option at all

jerryyin:
@krzysz00 `--test-fwd` sounds good, I'll amend the naming in the next commit.

> you could just call the script twice

On one hand I think the inference option would be useful with more collaboration with MIgrpahX team (which dosen't need backward kernels in general). On the other hand, this reduce the duplication in testing stage.

krzysz00:
Well, let's put the option in, then - makes sense

krzysz00 (approved):

PR #602, opened 2022-05-04
Add support for broadcast ops in MIXR

jungpark-mlir:
contains
- migraphx.broadcast
- migraphx.multibroadcast
- fix broadcast folding with nchw layout

### DIFF
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -135,6 +135,15 @@ public:
   }
 };
 
+bool isBroadcastable(operation op, operation operand) {
+  // tosa only broadcast implicitly on the second input of the binary operators.
+  if (op.getNumOperands() != 2)
+    return false;
+  if (op.getOperand(1) != operand)
+    return false;
+  return true;
+}
+
 class BroadcastConverter final
     : public OpConversionPattern<migraphx::BroadcastOp> {
 public:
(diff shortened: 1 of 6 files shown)


mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -135,9 +135,131 @@ class ConvConverter final
...
+                  ConversionPatternRewriter &rewriter) const final {
+    auto operands = adaptor.getOperands();
+    Location loc = op->getLoc();
+    auto input_t = operands[0];
+    auto axis = op->getAttr("axis").cast<IntegerAttr>().getInt();
+
+    for (auto &use : op->getResult(0).getUses()) {
+      auto expandedOp = use.getOwner();
+      // isa binary operation,
+      if (isa<migraphx::AddOp>(expandedOp) ||
```
sjw36: Shouldn't need to restrict to these ops. Can we just replace the migraphx.broadcast with the tosa.reshape?

jungpark-mlir: broadcast in tosa is restrictive. Just checked and it only broadcasts on the second input of the binary operations. Not clearly defined as a rule but only defined per each operator.
I'll change it to reflect that rule.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -135,9 +147,134 @@ class ConvConverter final
...
+
+    // get shape of the use
+    auto outShape = op->getResultTypes()[0].cast<ShapedType>().getShape();
+    auto outElemType =
+        op->getResultTypes()[0].cast<ShapedType>().getElementType();
+    uint32_t outRank = outShape.size();
+    auto inShape = input_t.getType().cast<ShapedType>().getShape();
+
+    Value newOperand = input_t;
+    if (outRank != inShape.size()) {
```
sjw36, line 171: Looks good.

sjw36 (approved):
LGTM

PR #606, opened 2022-05-05
Update parameter sweep script, add to weekly CI

krzysz00:
This'll try out a whole bunch of parameter combinations on a weekly basis to catch weirder regressions. Merging this won't cause CI failures, near as I can tell.


krzysz00:
@yiqian1 @jerryyin  This is now ready for review.

mlir/utils/jenkins/Dockerfile.rocm
```diff
@@ -22,16 +22,19 @@ RUN echo 'install build dependencies'; \
         gdb build-essential  \
         ninja-build \
         ccache \
-        python3 python3-psutil \
-        python3-pip python3-setuptools \
-        swig python3-dev libedit-dev libncurses5-dev libxml2-dev liblzma-dev golang rsync jq;
+        swig libedit-dev libncurses5-dev libxml2-dev liblzma-dev golang rsync jq;
```
jerryyin: There's a generic request here that in all changes to `Section 1`, could you move them to `Section 2`? (Unless the changes are from upstream).

This is due to the dockerfile has been manually composed and synced from upstream. It is easier to keep upstream/downstream separate and tell the difference.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -345,6 +345,59 @@ pipeline {
                 }
             }
         }
+        stage ("Parameter sweeps") {
```
jerryyin, line 348: [Question] How long does it typically take for this stage? I noticed that this new stage is in the PR test.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
```diff
@@ -200,7 +199,8 @@ static LogicalResult runMLIRPasses(ModuleOp &module,
           mlir::miopen::createAffixTuningParametersPass(blockSize, gridSize));
     } else {
       // Set up the default lowering pipeline which goes down to GPU dialect.
-      miopen::addPipeline(pm);
+      // If applicability is in the set, only go through gridwise GEMM.
+      miopen::addPipeline(pm, kernelPipelineSet.contains("applicability"));
```
jerryyin: I don't see `applicability` option can go together with others, therefore I prefer this to be scoped in another if condition that check for the `kernelPipelineSet` size to make sure the user does not try to put multiple pipelines after applicability.

mlir/utils/jenkins/parameterSweeps.py
```diff
@@ -0,0 +1,394 @@
...
+        print(f"{result.name}: {config!r}")
+    if result == TestResult.FAIL:
+        return config
+    return result
+
+IterType = TypeVar('IterType')
+async def sweepParameters(paramIter: Iterable[IterType],
+    toConfig: Callable[[IterType], MLIROnlyConfig]) -> Tuple[int, int, List[MLIROnlyConfig]]:
+
+    CONCURRENT_TESTS = 100
```
jerryyin: Why override this locally? And how is `100` decided?

mlir/utils/jenkins/parameterSweeps.py
```diff
@@ -0,0 +1,394 @@
...
+        for c in failures:
+            print(' '.join(c.generateMlirDriverCommandLine()))
+    print(f"Passed: {n_passes}, Invalid: {n_invalids}, Failed: {len(failures)}")
+    return len(failures) == 0
+
+def main() -> bool:
+    global CONCURRENT_TESTS
+    global DEBUG
+    global QUIET
+    global XDLOPS
```
jerryyin: I'm not a huge fan of using global. The script can grow larger and depend on state of those global variables.

mlir/utils/jenkins/parameterSweeps.py
```diff
@@ -0,0 +1,394 @@
...
+
+CONV_STRUCTURE = itertools.product(
+    # Small/large
+    [False, True],
+    # op
+    ['fwd', 'wrw', 'bwd'],
+    # layout
+    ['NCHW', 'NHWC'],
+    # dtype
+    ['f32', 'f16'],
```
jerryyin, line 260: Probably an independent topic(and I'm fine if the complete support is added later): what about bf16 and i8?

mlir/utils/jenkins/parameterSweeps.py
```diff
@@ -0,0 +1,394 @@
...
+    range(0, 4),
+    range(0, 4),
+    range(0, 4),
+    range(0, 4),
+    # Stride - 1 or 2
+    range(1, 3),
+    range(1, 3),
+    # Dilation in [1, 2]
+    range(1, 3),
+    range(1, 3))
```
jerryyin: Some fairly important constants here: What's the reasoning in choosing the ranges? This question applies to other Structure configurations.

To make sure I understand: we brute force and try all and every config in the range? I'm asking because I was assuming that we "randomly pick" parameters.

mlir/utils/jenkins/parameterSweeps.py
```diff
@@ -0,0 +1,394 @@
...
+    # NPerBlock (exponent)
+    range(4, 9),
+    # KPerBlock (exponent)
+    range(0, 4),
+    # MPerWave (exponent)
+    range(1, 8),
+    # NPerWave (exponent)
+    range(4, 8),
+    # KPack (exponent)
+    range(1, 4)
```
jerryyin, line 305: [Note] I'm reminding myself that the ranges need to be adjusted when it is amended from MIOpen. How can we remember to synchronize those fields?

mlir/utils/jenkins/parameterSweeps.py
```diff
@@ -0,0 +1,394 @@
...
+    range(1, 3),
+    range(1, 3))
+
+def to_conv_structure_type_test(params) -> MLIROnlyConfig:
+    size, op, layout, dtype, phl, phr, pwl, pwr, sh, sw, dh, dw = params
+    g, hi, wi, y, x = 1, 4, 4, 2, 2
+    if size:
+        n, c, k = 64, 64, 64
+    else:
+        n, c, k = 1, 7, 7
```
jerryyin: Hmmm not sure what does this chunk do, mind explaining?

jerryyin (commented):
Thank you for your work! Looks good overall, with small nits/clarifications.

mlir/utils/jenkins/Jenkinsfile
krzysz00: Wait, what? It should only be in weekly. And about 3 hours.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
krzysz00: I'm pretty sure the way the pipeline script is written, you can have `applicabality` and `gpu` together?

krzysz00: @sjw36

sjw36: Should be 'tuning' and no it should not be coupled with any other pipelines.

krzysz00: But `tuning` already exists and doesn't run the whole applicability check

krzysz00: Tuning just gets tuning parameters - `applicability` stops after gridwise gemm

mlir/utils/jenkins/Jenkinsfile
jerryyin: Sorry red-herring, I saw it come after static test and assumed this is in PR test.

> It should only be in weekly. And about 3 hours.

Sounds good

mlir/utils/jenkins/parameterSweeps.py
krzysz00: Fixed

mlir/utils/jenkins/parameterSweeps.py
krzysz00: Todo comment added

mlir/utils/jenkins/parameterSweeps.py
krzysz00: Yeah, no, this is very much a brute force check, hence running weekly

krzysz00: (and I added some comments to clarify where I got the numbers from)

mlir/utils/jenkins/parameterSweeps.py
krzysz00: Commented

mlir/utils/jenkins/parameterSweeps.py
krzysz00: The local override is a bug, which has been fixed, good catch

mlir/utils/jenkins/Dockerfile.rocm
krzysz00: I've moved things to section 2, which includes the previous contamination of section 1

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
krzysz00: I'll note that the other possibilities (things like `applicability,rocdl`) will be weird no-ops, which should be the expected behavior if you think about it.

krzysz00: Ok, I'm reworking this entire pile of if statements. The main change is that `rocdl` now just does gpu-to-rocdl, and so the former `rocdl` is now `gpu,rocdl`. I claim this leads to cleaner code overall.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
```diff
@@ -191,32 +193,42 @@ static LogicalResult runMLIRPasses(ModuleOp &module,
         return failure();
       }
     }
+    if ((kernelPipelineSet.contains("tuning") ||
+         kernelPipelineSet.contains("applicability")) &&
+        kernelPipelineSet.size() != 1) {
+      llvm::errs() << "The `tuning` and `applicability` pipelines cannot be "
+                      "combined with any other pipeline options.\n";
+      return failure();
+    }
```
jerryyin: I believe `tuning` should be equivalent to `applicability`. I think the creation of `pipeline.cpp` aims at creating a central location where we ensure all pipeline are in sync. But tuning is the outlier, that this is one of the few cases we explicitly add a single stage to the pass manager.

I'd prefer to options consolidated into one now that you are touching it.

mlir/utils/jenkins/parameterSweeps.py
```diff
@@ -0,0 +1,403 @@
...
+def to_conv_structure_type_test(params, options: Options) -> MLIROnlyConfig:
+    size, op, layout, dtype, phl, phr, pwl, pwr, sh, sw, dh, dw = params
+    # Fixed parameters, y = x = 2, hi = wi = 4, g = 1
+    g, hi, wi, y, x = 1, 4, 4, 2, 2
+    if size:
+        # Values of n, c, k that prevent hitting the padding kernel
+        n, c, k = 64, 64, 64
+    else:
+        # Values of n, c, k, meant to be small and to hit the padding kernel
+        n, c, k = 1, 7, 7
```
jerryyin, line 283: I'm want to seek for some clarifications:
 - `g: 1`: this makes group convolution to be always hard-coded to 1 therefore making other non-1 variants untested?
 - `n, c, k = 64, 64, 64`: So the reason you pick those is due to they are larger than `1,7,7` and will always processed by non-padding kernel? But no particular reason on why you don't pick another set of parameters like `128,128,128`?

I'm okay with we go with the hard-coded numbers, but I need to understand that the parameter sweeps excluded the sweep of  `g, hi, wi, y, x, n, c, k` parameters?

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
krzysz00: Ok, who actually uses "tuning" and why was it just that one pass? @sjw36

krzysz00: Having looked, `tuning` is used only in some of the mlir-miopen-lib tests, where it could probably be replaced by calling `-miopen-affix-params` directly

mlir/utils/jenkins/parameterSweeps.py
krzysz00: Yeah, I'm only sweeping two particular sets of values for g, n, w, h, y, x, c, and k . I could throw together another set of tests that sweeps that set, probably carefully picking some other sets of those values but keeping things like stride and dilation constant. 

And yeah, all-64 was the smallest input I could think of that would ensure there was no need for GEMM padding - since bigger inputs lead to more verification time.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
krzysz00: (and so that's the question for you, Jerry)

mlir/utils/jenkins/parameterSweeps.py
jerryyin: Two sets are a reasonable start. I think we can go with it unless there's a counter example from a different work load. Could you actually file a follow-up ticket with ROCm x.x just so that we can grab the ticket when there's a need to extend?

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
jerryyin: Well I can also argue that the creation of `applicability` option is unnecessary. But I don't feel this matter as much to hold back the merge because of this minor refactoring item.

jerryyin (approved):
LGTM, thank you!

mlir/utils/jenkins/parameterSweeps.py
krzysz00: No problem, will do

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
krzysz00: I think what I'll do here is get rid of the `tuning` option.

jerryyin: Sounds good to me

PR #607, opened 2022-05-05
Fix calculateKBlockNum for group convolution.

whchung:
In MLIR MIOpen dialect implementation, k (output channel) and c (input channel)
are independent of g (group count).

In MIOpen implementation, k and c are implicitly timed by g, so they have to be
divided with g. This computation is not necessary inside MLIR MIOpen dialect
implementation.

Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/531

Relevant MIOpen logic:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/include/miopen/conv/problem_description.hpp#L167
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/include/miopen/conv/problem_description.hpp#L201

It can be observed g is not a standalone dimension in MIOpen problem description.


whchung:
Commit history is a bit messed up. Will fix it and reopen.

PR #608, opened 2022-05-05
Fix `calculateKBlockNum` for group convolution.

whchung:
In MLIR MIOpen dialect implementation, k (output channel) and c (input channel)
are independent of g (group count).

In MIOpen implementation, k and c are implicitly timed by g, so they have to be
divided with g. This computation is not necessary inside MLIR MIOpen dialect
implementation.

Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/531

Relevant MIOpen logic:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/include/miopen/conv/problem_description.hpp#L167
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/include/miopen/conv/problem_description.hpp#L201

It can be observed g is not a standalone dimension in MIOpen problem description.

Per @jerryyin 's comment I've also made the function returning `LogicalResult` so it could also be used to reject invalid tuning parameters, instead of simply crashing due to assert.

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/utility/loweringUtils.h
+++ b/mlir/include/mlir/Dialect/MIOpen/utility/loweringUtils.h
@@ -42,10 +42,8 @@ inline int64_t calculateKBlockNum(int64_t n, int64_t ho, int64_t wo, int64_t g,
                                   int64_t MPerBlock, int64_t NPerBlock,
                                   int64_t KPerBlock, int64_t KPack,
                                   int64_t num_cu) {
-  const int64_t KPerGroup = k / g;
-  const int64_t CPerGroup = c / g;
-  const int64_t gemmM = KPerGroup;
-  const int64_t gemmN = CPerGroup * y * x;
+  const int64_t gemmM = k;
+  const int64_t gemmN = c * y * x;
 
   const int64_t gemmK = n * ho * wo;
   int64_t gemmKBlock = 1;
--- a/mlir/include/mlir/Dialect/MIOpen/utility/loweringUtils.h
+++ b/mlir/include/mlir/Dialect/MIOpen/utility/loweringUtils.h
@@ -69,9 +69,9 @@ inline LogicalResult calculateKBlockNum(int64_t n, int64_t ho, int64_t wo,
     break;
   }
   // not more than n
-  gemmKBlocks = std::min(n, gemmKBlocks);
+  gemmKBlock = std::min(n, gemmKBlock);
   // not less than 1
-  gemmKBlocks = std::max((__int64_t)1, gemmKBlocks);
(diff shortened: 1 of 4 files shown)


mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -384,8 +394,12 @@ LogicalResult backwardWeightAtomicAdd(Conv2DBwdWeightOp op,
...
   // calculate gemmKBlocks
 
-  // static const int64_t MaxSubBlockNum = 2048 / standardBockNum;
-  int64_t gemmKBlocks = calculateKBlockNum(n, ho, wo);
+  int64_t gemmKBlocks = 1;
+  if (failed(calculateKBlockNum(n, ho, wo, g, k, c, y, x, MPerBlock, NPerBlock,
+                                KPerBlock, KPack, numCu, &gemmKBlocks))) {
+    op->emitOpError("Invalid tuning parameters to compute KBlock.");
+    return failure();
+  }
```
jerryyin, line 402: Nit: I feel like this is a repeated calculation as it is already derived in `GridwiseGemmParams.cpp` one lowering earlier. Could you make this kBlock to be an attribute only in case of backward atomic add kernel. Such that here you don't have to calculate this attribute a second time.

mlir/include/mlir/Dialect/MIOpen/utility/loweringUtils.h
```diff
@@ -22,27 +22,59 @@
...
+// contraints:
+// - GemmK (before splitting) = KBlock * KPerBlock * KPack * GemmK (after
+// splitting).
+// - n (batch size) is divisible by KBlock.
+//
+// 20 is a magic number obtained in MIOpen after empirical testing. It offers a
+// reasonable reduction of GemmK after splitting, without incurring too much
+// overheads on atomic adds. One potential future work is to make this value be
+// tunable.
+inline LogicalResult calculateKBlockNum(int64_t n, int64_t ho, int64_t wo,
```
jerryyin, line 40: I'll abstain from reviewing this function. Still unclear about how line 59-69 works.

jerryyin (approved):
I'm assuming you have tested the configs that was failing the previous nightly. Overall LGTM with a nitpick.

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
whchung: Got it. Let me do it in a separate PR.

PR #609, opened 2022-05-06
Introduce `kblocks` attribute.

whchung:
Address request made by @jerryyin in #608 . Avoid duplicated computation of `kblocks` by keeping it in an attribute.

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
@@ -586,11 +586,10 @@ struct InitParamsNonXDL : InitParams, Serializable<InitParamsNonXDL> {
   InitParamsNonXDL(int64_t bSize, int64_t mPerBlock, int64_t nPerBlock,
                    int64_t kPerBlock, int64_t mPerThread, int64_t nPerThread)
       : InitParams{mPerBlock, nPerBlock, kPerBlock}, gemmMPerThread(mPerThread),
-        gemmNPerThread(nPerThread), blockSize(bSize), gemmKBlocks(1LL) {}
+        gemmNPerThread(nPerThread), blockSize(bSize) {}
   int64_t gemmMPerThread;
   int64_t gemmNPerThread;
   int64_t blockSize;
-  int64_t gemmKBlocks;
 
   InitParamsNonXDL() : InitParamsNonXDL(0LL, 0LL, 0LL, 0LL, 0LL, 0LL) {}
 
--- a/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
@@ -595,7 +595,6 @@ struct InitParamsNonXDL : InitParams, Serializable<InitParamsNonXDL> {
   InitParamsNonXDL() : InitParamsNonXDL(0LL, 0LL, 0LL, 0LL, 0LL, 0LL) {}
 
   template <class Self, class F> static void visit(Self &&self, F f) {
-    f(self.gemmKBlocks);
     f(self.blockSize);
     f(self.gemmMPerBlock);
     f(self.gemmNPerBlock);
@@ -632,7 +631,6 @@ struct InitParamsXDL : InitParams, Serializable<InitParamsXDL> {
     f(self.gemmKPack);
     f(self.gemmAThreadCopyMoreGemmK);
     f(self.gemmBThreadCopyMoreGemmKPack);
-    f(self.gemmKBlocks);
   }
 };
 
--- a/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
@@ -610,7 +610,7 @@ struct InitParamsXDL : InitParams, Serializable<InitParamsXDL> {
       : InitParams{mPerBlock, nPerBlock, kPerBlock}, gemmMPerWave(mPerWave),
         gemmNPerWave(nPerWave), gemmKPack(kPack),
         gemmAThreadCopyMoreGemmK(aThreadCopyMoreGemmK),
-        gemmBThreadCopyMoreGemmKPack(bThreadCopyMoreGemmKPack), gemmKBlocks(1LL) {}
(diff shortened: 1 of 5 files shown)


mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
```diff
@@ -586,10 +586,11 @@ struct InitParamsNonXDL : InitParams, Serializable<InitParamsNonXDL> {
   InitParamsNonXDL(int64_t bSize, int64_t mPerBlock, int64_t nPerBlock,
                    int64_t kPerBlock, int64_t mPerThread, int64_t nPerThread)
       : InitParams{mPerBlock, nPerBlock, kPerBlock}, gemmMPerThread(mPerThread),
-        gemmNPerThread(nPerThread), blockSize(bSize) {}
+        gemmNPerThread(nPerThread), blockSize(bSize), gemmKBlocks(1LL) {}
```
jerryyin: I believe `wrw + atomic` is only available in xdlops, because that's the only place where `getKBlocks()` get called. If that's the case, we should remove this unused parameter for non-xdlops.

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
```diff
@@ -325,6 +326,7 @@ void AffixTuningParameters::affixTuningParametersImpl(T &op) {
     op->setAttr("block_size", b.getI32IntegerAttr(validParams.blockSize));
     // For non-XDLOPS path, do not use KPack for now.
     op->setAttr("kpack", b.getI32IntegerAttr(1));
+    op->setAttr("kblocks", b.getI32IntegerAttr(validParams.gemmKBlocks));
```
jerryyin: Should remove it from here as well.

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
```diff
@@ -268,6 +268,7 @@ void AffixTuningParameters::affixTuningParametersImpl(T &op) {
     }
 
     op->setAttr("kpack", b.getI32IntegerAttr(validParams.gemmKPack));
+    op->setAttr("kblocks", b.getI32IntegerAttr(validParams.gemmKBlocks));
```
jerryyin: Add a condition such that we set this attribute only when op is of `Conv2DBwdWeightOp`.

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
krzysz00: Agreed re not having this in the non-xdlops case, especially since, IIRC, the InitParams* structs directly correspond to perf config strings.

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
whchung: will do.

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
whchung: will do.

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
whchung: will do.

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
```diff
@@ -610,7 +610,7 @@ struct InitParamsXDL : InitParams, Serializable<InitParamsXDL> {
       : InitParams{mPerBlock, nPerBlock, kPerBlock}, gemmMPerWave(mPerWave),
         gemmNPerWave(nPerWave), gemmKPack(kPack),
         gemmAThreadCopyMoreGemmK(aThreadCopyMoreGemmK),
-        gemmBThreadCopyMoreGemmKPack(bThreadCopyMoreGemmKPack) {}
+        gemmBThreadCopyMoreGemmKPack(bThreadCopyMoreGemmKPack), gemmKBlocks(1LL) {}
```
krzysz00: @jerryyin Shouldn't something like this go on the derived parameters? I could've sworn InitParams is supposed to be in 1:1 correspondence with the perf config

jerryyin: @krzysz00 Nice catch! Correct, both `initParams` and `InitParamsXDL` has 1:1 mappting with perf config and a pass-through attribute should go to the derived parameters list.

whchung: @jerryyin Just to understand this amended requirement correctly: it seems we want to make `InitParams*` only contain values from `--perf_config`, therefore derived values such as `kblocks` shall be kept in a different structure.

Then, some subsequent questions for discussion. I'm open to ideas and can act accordingly.

- in which structure shall this value be kept? right now it seems there are `DerivedParams` and `DerivedOutParams`. Should we put this field in one of them, or create a new structure?

- what should be the interface to compute this derived parameter? right now we have `calculateInputDerivedParams` and `calculateOutputDerivedParams`, and `populateParamsXDL::obtainTuningParameters` -> `populateParamsXDL::populateDerived` is the main interface to compute these derived parameters.

jerryyin: https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/060a944e06eb4395cb2d648906fb9ff4e9c11b20/mlir/lib/Dialect/MIOpen/Tuning/GridwiseGemmParams.cpp#L229-L233

In this function signature, we use the `InitParamsXDL` to derive the rest of parameters such that:
 - `gemmADerivedParam` store gemmA related derived parameters only
 - `gemmBDerivedParam` store gemmB related derived parameters only
 - `gemmCDerivedParam` store gemmC related derived parameters only

So it really depends on whether this parameter belong to a gemm argument or output. If not, then you'll likely need to follow on how `blockSize` or `gridSize` get passed along in `obtainTuningParameters()` routine.

whchung: @jerryyin please take a look at the latest commit. `kblocks` is not really tied to a specific operand so I'm making it a standalone value at the moment.

jerryyin (approved):

PR #611, opened 2022-05-09
Match i8 sanity tests with other data types

jerryyin:
I did the following in this PR: 
 - Matched i8 sanity tests with other data types
 - Amended applicability check to reject cases for i8 in corner cases

### DIFF
--- a/mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
+++ b/mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
@@ -440,9 +440,11 @@ LogicalResult Conv2dGenerator::parseConvConfig(const char *arguments) {
       return false;
     }
 
-    bool noMixedTypes = argMap["in_type"] == argMap["fil_type"] &&
-                        (argMap["out_type"] == "i32" ||
-                         argMap["out_type"] == argMap["in_type"]);
+    bool noMixedTypes =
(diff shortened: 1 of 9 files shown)


krzysz00 (approved):
LGTM, good catch!

Also, should we maybe split the sanity tests into more files?

jerryyin:
> Also, should we maybe split the sanity tests into more files?

My two cents is that we don't necessarily need to group sanity tests into different files based on that majority of them does not rely on IR but return status.

A second aspect is that since we have little type dependent logic, aggregating them together can actually help consolidate the testing logic && make developers more aware of the differences. Examples please look at the diff here: [mlir/test/mlir-miopen-driver/populate_host_print.mlir](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/611/files#diff-0f956ae259e1c761ab3561b9cfb7c58cae52206f58bccef509301baad864ca21). This unit test, if written independently, will be required to have a large number of duplications (as was originally written). Now, with type specific checks, we make the type dependent logic really stand out.

krzysz00:
I more meant the specific files `sanity.mlir` and the like - they've got these really long chains of compiler invocations, and we could parallelize those by type by splitting them into other files

jerryyin:
@krzysz00 That's probably a good idea - let me split it into multiple smaller ones for the sake of speed up testing.

PR #612, opened 2022-05-10
[DO NOT SQUASH] Backport changes to buffer ops from upstream

krzysz00:
The raw buffer load/store ops are now upstream in an `amdgpu` dialect.

I'd like to get those changes (which are just moving things around, compared to our version) into the repo now so that I can work on things like moving the mfma stuff to `amdgpu` as well.

The upstream changes (which is where the "yeah, this should go into AMDGPU" comment came from) are at https://reviews.llvm.org/D122765

and a small followup fix I pushed in real quick up there is at https://reviews.llvm.org/D125320

I moved the buffer ops over because those were an easy, self-contained change that other people upstream might need.

Overall, we should move other wrappers around clearly vendor-specific intrinsics into `amdgpu` and send them upstream. This includes things like `mfma`.


jerryyin (approved):
LGTM. For reference purpose, could you attach the following to PR's description:
 - The corresponding upstream PR link
 - Relevant discussions on how you decide to move just raw buffer ops into a new lowering pipeline

With those information, it can help guide our future refactoring as well.

krzysz00:
PR description updated

PR #613, opened 2022-05-10
MLIR#535: fixed miopen-copy-opt to fail when types don't match

sjw36:
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/535

### DIFF
--- a/mlir/test/xmir/mobilenetv1.mid.mlir
+++ b/mlir/test/xmir/mobilenetv1.mid.mlir
@@ -1,23 +1,21 @@
-// XFAIL: *
-// RUN: mlir-miopen-driver -host-pipeline partition %s | miopen-opt --miopen-async-launch - | FileCheck %s
-
+// RUN: mlir-miopen-driver -host-pipeline partition,highlevel %s | FileCheck %s
 
 module {
 
-// CHECK: func private @mobilenetv1_outlined_part_0(%arg0: tensor<1x224x224x3xf32>, %arg1: tensor<32x3x3x3xf32>) -> tensor<1x112x112x32xf32> attributes {kernel}
-// CHECK: func private @mobilenetv1_outlined_part_1(%arg0: tensor<1x112x112x32xf32>, %arg1: tensor<3x3x32x1xf32>) -> tensor<1x112x112x32xf32> attributes {kernel}
-// CHECK: func private @mobilenetv1_outlined_part_2(%arg0: tensor<1x112x112x32xf32>, %arg1: tensor<64x1x1x32xf32>) -> tensor<1x112x112x64xf32> attributes {kernel}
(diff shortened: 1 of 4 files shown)


mlir/test/xmir/mobilenetv1.mid.mlir
```diff
@@ -1,23 +1,21 @@
...
-// CHECK-NEXT:  %token_2, %results_3 = async.launch @mobilenetv1_outlined_part_2 [%token_0] (%results_1, %arg3) : (tensor<1x112x112x32xf32>, tensor<64x1x1x32xf32>) -> tensor<1x112x112x64xf32>
-// CHECK-NEXT:  %token_4, %results_5 = async.launch @mobilenetv1_outlined_part_3 [%token_2] (%results_3, %arg4) : (tensor<1x112x112x64xf32>, tensor<3x3x64x1xf32>) -> tensor<1x56x56x64xf32>
-// CHECK-NEXT:  %token_6, %results_7 = async.launch @mobilenetv1_outlined_part_4 [%token_4] (%results_5, %arg5) : (tensor<1x56x56x64xf32>, tensor<128x1x1x64xf32>) -> tensor<1x56x56x128xf32>
-// CHECK-NEXT:  async.await %token_6 : !async.token
-// CHECK:       return %results_7 : tensor<1x56x56x128xf32>
+// CHECK: func private @mobilenetv1_outlined_part_0(%arg0: memref<1x224x224x3xf32>, %arg1: memref<32x3x3x3xf32>, %arg2: memref<1x112x112x32xf32>) attributes {kernel} {
+
+// CHECK: func @mobilenetv1(%arg0: memref<1x224x224x3xf32>, %arg1: memref<32x3x3x3xf32>, %arg2: memref<3x3x32x1xf32>, %arg3: memref<64x1x1x32xf32>, %arg4: memref<3x3x64x1xf32>, %arg5: memref<128x1x1x64xf32>, %arg6: memref<1x56x56x128xf32>) {
+// CHECK:   %cst = arith.constant 0.000000e+00 : f32
```
krzysz00: Hey, let's not embed symbol names we don't have control over into tests.

krzysz00 (changes requested):
I'm going to assume the core logic is correct, so that's a LGTM.

One thing I realized I missed last time and that I don't like is that the CHECK lines make assumptions about what the MLIR pretty-printer will do. This isn't good, and I'd like to see those replaced with `%[[VAR:.*]]` and `%[[VAR]]` as needed

mlir/test/xmir/mobilenetv1.mid.mlir
sjw36: Ok, updated for local variables.

krzysz00 (approved):
Approved, though for generality it might be worth cleaning up the argument signatures as well (`CHECK-SAME` exists for this reason). It probably won't be an issue, but it's something to keep in mind when writing future tests.

PR #614, opened 2022-05-10
[HOTFIX] Fix incorrectly placed test configs.

whchung:
In https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/268 ,
2 configs were passing and 1 config was failing. In PR #610, the failing one was
incorrectly added instead of the passing 2. Fix in this commit.


krzysz00:
I'm creating a forced CI pass, so as soon as there's an approval it'll be mergeable

jerryyin (approved):

PR #615, opened 2022-05-11
[NFC] Move gridwise gemm stuff out of headers

krzysz00:


mlir/include/mlir/Dialect/MIOpen/Generator/Conv2dGenerator.h
```diff
@@ -22,6 +22,7 @@
 #include "llvm/ADT/StringRef.h"
 
 namespace mlir {
+namespace miopen {
```
jerryyin, line 25: Are you trying to scope every header into the `miopen` namespace as long as they belong to `miopen-dialect`?

mlir/include/mlir/Dialect/MIOpen/MIOpen.h
```diff
@@ -14,34 +14,21 @@
 #define MLIR_MIOPENOPS_OPS_H_
 
 #include "mlir/Dialect/Func/IR/FuncOps.h"
-#include "mlir/IR/Attributes.h"
```
jerryyin, line 17: Nice clean up in this universally included header.

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
```diff
@@ -65,10 +67,8 @@ Conv2dGenerator::Conv2dGenerator(
 Conv2dGenerator::Conv2dGenerator(const Conv2dGenerator::Config &_config)
     : config(_config) {}
 
-namespace {
-
-void strToTokens(const std::string &arguments,
-                 std::map<std::string, std::string> &argMap) {
+static void strToTokens(const std::string &arguments,
```
jerryyin, line 70: [Nit] I don't feel strong about this point. Although very much similar, I don't like static compared to anonymous namepace: The latter has wider applicable range and can allow to define a locally used type. I'd prefer to revert here and later changes that moved anonymous namespace to static.

mlir/lib/Dialect/MIOpen/Tuning/GridwiseGemmParams.cpp
```diff
@@ -6,14 +6,587 @@
 
 #define DEBUG_TYPE "miopen-tuning-parameter"
 
+using namespace mlir;
+using namespace mlir::miopen;
+
+static void
```
jerryyin, line 12: Changes looks good. If you want to continue the refactoring, my preference would be to make 3 independent header/cpp combinations on the following classes because this file really has blown up:
 - InitParams
 - InitParamsXDL
 - InitParamsNonXDL

mlir/lib/Dialect/MIOpen/Tuning/ConvContext.cpp
```diff
@@ -36,10 +36,9 @@ void populateSeqVal(const ArrayAttr &seqAttr,
     }
   }
 }
-} // namespace
 
-ConvolutionContext populateConvContext(Operation *op) {
-  miopen::ConvOpType opType = miopen::obtainConvDirection(op);
+ConvolutionContext mlir::miopen::populateConvContext(Operation *op) {
+  ConvOpType opType = obtainConvDirection(op);
```
jerryyin, line 41: Why do you need to add the namespace around this (with added using statements)

jerryyin (approved):
Some nits/clarifications. I'm good with merging with the state of PR as-is. Nice clean up and thank you!

mlir/include/mlir/Dialect/MIOpen/Generator/Conv2dGenerator.h
krzysz00: I was trying to put all the miopen related headers into the miopen namespace, yeah. I didn't go through them and get them systematically, but I went and touched the ones I was already editing.

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
krzysz00: When I was doing review with upstream, they wanted local functions as `static` and `namespace{}` only for anonymous structs.

krzysz00: Which I'm not the biggest fan of either, but if that's how they've been wanting it over there...

mlir/lib/Dialect/MIOpen/Tuning/ConvContext.cpp
krzysz00: I was getting linker errors if I didn't

mlir/lib/Dialect/MIOpen/Generator/Conv2dGenerator.cpp
jerryyin: Gotcha, I can understand why they want this. This gives an explicit static in the function signature so is more visible to developer that this is available in local scope only.

PR #616, opened 2022-05-12
Eliminate temporary tuning parameters for fp32 wrw.

whchung:
This should not be getting into our way now.


mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
```diff
@@ -203,12 +203,6 @@ class PopulateParamsXDL {
   static const InitParamsXDL initParametersForwardI8[nInitParametersForwardI8];
 
   static constexpr size_t nInitParametersBwdWeightF32 = 7;
```
krzysz00: Get rid of the size parameter as well

krzysz00 (changes requested):
You left in an integer constant (which is partly my bad for the ordering of the comments). LGTM otherwise.

krzysz00 (approved):

PR #617, opened 2022-05-12
Generalise --tosa-partition to allow other specified ops as primary op [DO NOT SQUASH]

pcf000:
Allow tosa.transpose and tosa.reshape to be gathered along with the element-wise ops.
Allow specifying the name of the "kernel" attribute.
Test the sub-options primary-ops, trailing-only, and attribute-name.
Test both command-line and programmatic interfaces.

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.h
+++ b/external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.h
@@ -31,9 +31,11 @@ std::unique_ptr<Pass> createTosaInferShapesPass();
 std::unique_ptr<Pass> createTosaMakeBroadcastablePass();
 std::unique_ptr<Pass> createTosaTestQuantUtilAPIPass();
 std::unique_ptr<Pass> createTosaPartitionPass();
-std::unique_ptr<Pass> createPostPartitionCollapsePass();
+std::unique_ptr<Pass>
+createTosaPartitionPass(llvm::function_ref<bool(Operation *)> &pred,
+                        std::string attributeName = "kernel",
+                        bool nofront = false);
 std::unique_ptr<Pass> createTosaOptionalDecompositions();
-void registerTosaPartitionPipelinePass();
 
 #define GEN_PASS_REGISTRATION
 #include "mlir/Dialect/Tosa/Transforms/Passes.h.inc"
--- a/external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.h
+++ b/external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.h
@@ -31,10 +31,8 @@ std::unique_ptr<Pass> createTosaInferShapesPass();
 std::unique_ptr<Pass> createTosaMakeBroadcastablePass();
 std::unique_ptr<Pass> createTosaTestQuantUtilAPIPass();
 std::unique_ptr<Pass> createTosaPartitionPass();
-std::unique_ptr<Pass>
-createTosaPartitionPass(llvm::function_ref<bool(Operation *)> &pred,
-                        std::string attributeName = "kernel",
-                        bool nofront = false);
+class PartitionConfig;
+std::unique_ptr<Pass> createTosaPartitionPass(PartitionConfig *config);
 std::unique_ptr<Pass> createTosaOptionalDecompositions();
 
 #define GEN_PASS_REGISTRATION
--- a/external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
@@ -52,7 +52,13 @@ def TosaPartition : Pass<"tosa-partition", "ModuleOp"> {
   let constructor = "createTosaPartitionPass()";
 
   let options = [
-    Option<"nofront", "nofront", "bool", /*default=*/"false",
+    ListOption<"primaryOps", "primary-ops", "std::string",
+               "One or more operations to be used as focus of partitioned "
+               "kernels",
+               "llvm::cl::ZeroOrMore">,
+    Option<"attributeName", "attribute-name", "std::string",
+           /*default=*/"\"kernel\"", "Attribute for outlined functions">,
+    Option<"noFront", "trailing-only", "bool", /*default=*/"false",
            "Don't gather ops ahead of Conv2D">
   ];
   let dependentDialects = ["tosa::TosaDialect"];
--- a/external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
@@ -50,17 +50,6 @@ def TosaPartition : Pass<"tosa-partition", "ModuleOp"> {
   }];
 
   let constructor = "createTosaPartitionPass()";
-
-  let options = [
-    ListOption<"primaryOps", "primary-ops", "std::string",
-               "One or more operations to be used as focus of partitioned "
-               "kernels",
-               "llvm::cl::ZeroOrMore">,
-    Option<"attributeName", "attribute-name", "std::string",
-           /*default=*/"\"kernel\"", "Attribute for outlined functions">,
-    Option<"noFront", "trailing-only", "bool", /*default=*/"false",
-           "Don't gather ops ahead of Conv2D">
-  ];
   let dependentDialects = ["tosa::TosaDialect"];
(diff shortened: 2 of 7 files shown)


mlir/tools/miopen-gen/CMakeLists.txt
```diff
@@ -5,9 +5,16 @@ set(LLVM_LINK_COMPONENTS
 get_property(dialect_libs GLOBAL PROPERTY MLIR_DIALECT_LIBS)
 get_property(conversion_libs GLOBAL PROPERTY MLIR_CONVERSION_LIBS)
 
+if(MLIR_INCLUDE_TESTS)
```
pcf000, line 8: This set of changes, with the one for mlir-miopen-driver, fixes compiling with -DMLIR_INCLUDE_TESTS=OFF.  It's the same pattern used in external/llvm-project/mlir/tools/*/CMakeLists.txt.

The default is ON, but I was trying to make sure my tests didn't cause trouble in the other case.  As it happens, the whole test directory is omitted when OFF, so my concern was a non-problem, but I found this build failure.

external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.h
```diff
@@ -31,9 +31,11 @@ std::unique_ptr<Pass> createTosaInferShapesPass();
 std::unique_ptr<Pass> createTosaMakeBroadcastablePass();
 std::unique_ptr<Pass> createTosaTestQuantUtilAPIPass();
 std::unique_ptr<Pass> createTosaPartitionPass();
-std::unique_ptr<Pass> createPostPartitionCollapsePass();
+std::unique_ptr<Pass>
```
pcf000: Remove the obsolete declarations for passes we no longer have, and add a second variant of createTosaPartitionPass() to allow specifying the predicate, attribute name, and trailing-only option at the program level.  Typical command-line usage will call the no-arguments version and interpret the command-line options;  calling within a program can supply them as arguments.

external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
```diff
@@ -52,7 +52,13 @@ def TosaPartition : Pass<"tosa-partition", "ModuleOp"> {
   let constructor = "createTosaPartitionPass()";
 
   let options = [
-    Option<"nofront", "nofront", "bool", /*default=*/"false",
+    ListOption<"primaryOps", "primary-ops", "std::string",
```
pcf000: See the test file for usage examples.  I could be persuaded to use "anchor" instead of "primary" but I'm sort-of waiting for something better.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -80,10 +80,12 @@ bool isElementwiseOp(Operation *op) {
         tosa::NegateOp,
```
pcf000: One of my lost earlier versions had this function as "isFuseableOp" instead of "isElementwiseOp," since they aren't all element-wise any more.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -374,10 +377,22 @@ void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
   }
 }
 
+bool isConv2D(Operation *op) { return isa<tosa::Conv2DOp>(op); }
+
 // Inspired by / adapted from TestSCFIfUtilsPass in
 // test/lib/Transforms/TestSCFUtils.cpp.
 class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
+  std::function<bool(Operation *)> isPrimaryOp;
```
pcf000: The command-line option specifies op names but the create-function takes a predicate;  both lead to setting isPrimaryOp.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -374,10 +377,22 @@ void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
...
+
 public:
+  TosaPartitionPass() = default;
+  TosaPartitionPass(std::function<bool(Operation *)> pred,
+                    const std::string &attrName, bool nofront)
+      : isPrimaryOp(std::move(pred)) {
+    attributeName = attrName;
```
pcf000: Something about type conversions required that I assign these here rather than initialise them like isPrimaryOp.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -539,3 +569,82 @@ class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
...
+std::unique_ptr<Pass>
+mlir::tosa::createTosaPartitionPass(std::function<bool(Operation *)> &pred,
+                                    std::string attrName, bool noFront) {
+  return std::make_unique<TosaPartitionPass>(pred, attrName, noFront);
+}
+
+
+namespace {
+
+class TestTosaPartitionOptionsPass
```
pcf000, line 671: This pass exists only for testing.  Each of its sub-options sets up a particular createTosaPartitionPass() call to test out the argument-passing.  The combinations are arbitrarily arranged to fit the command-line-options test so I can use the same .mlir file for both.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
krzysz00: Can we go for `llvm::function_ref<>` instead of `std::function<>`?

pcf000: I did not know that was a thing.  I made the change.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -267,6 +269,7 @@ findOutliningCandidate(OutliningCandidate &newCandidate,
 void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
                         ArrayRef<Operation *> frontOps, ArrayRef<Value> params,
                         ArrayRef<Value> returnVals, StringRef partFnName,
+                        StringRef attrName,
```
sjw36, line 305: how about 'kernelAttrName'.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -418,18 +433,33 @@ class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
   }
 
   void runOnOperation() override {
+    // No supplied predicate.  If no --primary-ops, use default, isConv2D().
+    // Otherwise, use the generic list-search predicate.
+    if (!isPrimaryOp) {
```
sjw36: I would prefer 'anchor' over 'primary' since it is more descriptive and used similarly in other dialects such as Linalg. It's ok if you want to do that in a follow-up PR but please file a ticket.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -539,3 +569,77 @@ class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
 std::unique_ptr<Pass> mlir::tosa::createTosaPartitionPass() {
   return std::make_unique<TosaPartitionPass>();
 }
+
+std::unique_ptr<Pass>
+mlir::tosa::createTosaPartitionPass(llvm::function_ref<bool(Operation *)> &pred,
+                                    std::string attrName, bool noFront) {
```
sjw36: use StringRef

sjw36: Also should have a test for predecessors and successors. I suggest a class in the header that has overrideable methods to test each.

sjw36 (changes requested):
I would like a class in the header that could be override-able with methods: isAnchor(), isFuseable(). Or similar. Maybe separate test for predecessors and successors.

pcf000:
Turned skip-ci on while I'm working on things.

PR #618, opened 2022-05-12
Adding int8 tuning and perf results

jerryyin:
This PR added int8 performance tests into the weekly CI and enabled performance report for int8. I have to amend the MIOpen default tracking branch before all changes are checked into develop.

### DIFF
--- a/mlir/utils/jenkins/MIOpenDriver.py
+++ b/mlir/utils/jenkins/MIOpenDriver.py
@@ -36,11 +36,19 @@ def getConfigurations(fileName):
     configFile = open(fileName, 'r')
     lines = configFile.readlines()
     configs = [];
+    # All combinations of conv direction, type and layouts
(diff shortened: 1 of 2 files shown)


mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -36,11 +36,19 @@ def getConfigurations(fileName):
     configFile = open(fileName, 'r')
     lines = configFile.readlines()
     configs = [];
+    # All combinations of conv direction, type and layouts
```
krzysz00, line 39: How about adding int8 to `DATA_TYPES` and then going with an
```python
if datatype == `convint8` and direction != `-F 1`:
    continue
```

It'll reduce code duplication

krzysz00: (also, can you fix the extra space in the format string while you're here?)

jerryyin: Nice catch! Thanks for catching those.

krzysz00 (approved):

PR #619, opened 2022-05-12
Fix gemm size calculation, probably fixing some bwd data bugs

krzysz00:
The value for leftPadW was actually the value for rightPadH

This fixes the two configs specifically called out in
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/534
but I'll need to re-test to confirm if that is the entire issue.

For the second suspicious config (with strides=2 in both h and w)
we get a change in grid_size from 72 to 96 on the kernels, which
matches the failure state (which was that no one was writing to
input[48:64]).

(I am going to duly suspect that this might be the reason we've got some of those bwd data padding kernel configs excluded as well, but that's something for future me to check)


jerryyin (approved):
LGTM but why isn't right pad H/W used here at all? I'm unfamiliar with this part of code to tell.

krzysz00:
None of the ytilda/htilda/... computations, etther here or in `ConvToGemm`, use the right pad. My guess is this is because, since the right pad doesn't show up as a change to coordinates, having it be skipped because you didn't grab enough threads as opposed to because of a bounds check doesn't make a difference.

krzysz00:
Update: this actually fixes all the configs the parameter sweep was complaining about.

PR #620, opened 2022-05-12
[ROCm 5.2] Get rid of unnecessary warning messages from tuning

jerryyin:
MIOpen caught those messages in tuning. Requirement: those error messages should not be visible in tuning process or at least suppressible. The messages violate this requirement.

I'm doing this change directly upon release branch because there's little risk on this PR.

Please review/approve.

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
@@ -989,14 +989,10 @@ private:
     // For fp32: reject anything wider than 4.
     // For fp16/bf16: reject anything narrower than 4, or greater than 8.
     if (dataType.isF32() && param.gemmKPack > 4) {
-      llvm::errs() << "Invalid KPACK tuning parameter: " << param.gemmKPack
-                   << "\n";
       return failure();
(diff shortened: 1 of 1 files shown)


junliume (approved):
LGTM

whchung (commented):
instead of getting rid of the legitimate outputs, perhaps suppress them by switching to `LLVM_DEBUG(llvm::dbgs())`?

krzysz00 (commented):
Can we make these `LLVM_DEBUG(llvm::dbgs() << ...)`s instead? I apparently forgot to do that during yesterday's refactor - and, for the record, they were like this before that.

jerryyin:
> Can we make these LLVM_DEBUG(llvm::dbgs() << ...)s instead?

Nope. That has to be done in an implementation cpp file, not in a header. Otherwise the `#define DEBUG_TYPE` will spill over to all other implementation that included this header. 

Right now I figure that just getting rid of the error message is the least risky way to do it. Otherwise we'd have to pull in @krzysz00 's refactoring PR to be able to do `LLVM_DEBUG`. I'd prefer not to do that in a release branch.

krzysz00 (approved):
Oh, this is onto the release branch?! In which case, yeah, that's the best fix, approved

PR #621, opened 2022-05-13
Refactor implementation of SoftwareBF16

yiqian1:
Add a SoftwareBF16 pass to handle bf16->i16 conversion and replace bf16 arithmetic operations with f32 operations. This pass is called from within ConvertGPUToRocdl and from the pmHost in mlir-rocm-runner. Remove BF16 hacks from miopen-gen.


external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/Transforms/SoftwareBF16.h
```diff
@@ -0,0 +1,33 @@
...
+
+namespace LLVM {
+
+/// Since AMD GPU targets don't natively support bfloat as a type, even though
+/// LLVM does, populate patterns and type conversions to convert uses of bf16 to
+/// i16. To be removed when/if the backend has been taught about bfloat.
+void populateBF16ToLLVMConversionPatterns(LLVMTypeConverter &,
+                                          RewritePatternSet &);
+/// Creates a pass that converts BF16 type
+std::unique_ptr<Pass> createSoftwareBF16Pass();
```
krzysz00: This declaration goes in `Passes.h` by convention.

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/Transforms/SoftwareBF16.h
```diff
@@ -0,0 +1,33 @@
+//===- LegalizeForExport.h - Prepare for translation to LLVM IR -*- C++ -*-===//
```
krzysz00: Update filename in comment

external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
```diff
@@ -107,7 +109,8 @@ struct LowerGpuOpsToROCDLOpsPass
     // Keep last to prioritize newly-added type conversions, just in case
```
krzysz00: I'd reword the comment here

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/Transforms/SoftwareBF16.h
```diff
@@ -0,0 +1,33 @@
...
+class LLVMTypeConverter;
+class RewritePatternSet;
+class Pass;
+
+namespace LLVM {
+
+/// Since AMD GPU targets don't natively support bfloat as a type, even though
+/// LLVM does, populate patterns and type conversions to convert uses of bf16 to
+/// i16. To be removed when/if the backend has been taught about bfloat.
+void populateBF16ToLLVMConversionPatterns(LLVMTypeConverter &,
```
krzysz00: Rename to something like `populateSoftwareBF16Patterns`.

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/Transforms/SoftwareBF16.h
```diff
@@ -0,0 +1,33 @@
...
+
+namespace mlir {
+class Operation;
+class LLVMTypeConverter;
+class RewritePatternSet;
+class Pass;
+
+namespace LLVM {
+
+/// Since AMD GPU targets don't natively support bfloat as a type, even though
```
krzysz00: Reword comment to mention that this applies to other targets, including X86

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,364 @@
+//===- LegalizeForExport.cpp - Prepare for translation to LLVM IR ---------===//
```
krzysz00: Filename

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,364 @@
...
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+using namespace mlir;
+
+namespace mlir {
+namespace LLVM {
+/// Rewrites bf16 constants to their i16 equivalents
+/// This is relying on the fact that the vector, i16, and bf16 types used in the
+/// LLVM dialect are the standard ones and not weird custom wrappers
+struct BF16ConstCasting : OpRewritePattern<LLVM::ConstantOp> {
```
krzysz00, line 37: These structs go into an anonymous namespace

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,364 @@
...
+#include "PassDetail.h"
+#include "mlir/Conversion/LLVMCommon/ConversionTarget.h"
+#include "mlir/Conversion/LLVMCommon/Pattern.h"
+#include "mlir/Conversion/LLVMCommon/TypeConverter.h"
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+using namespace mlir;
+
```
krzysz00, line 17: Might as well chuck a `using namespace mlir::LLVM;` here

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,364 @@
...
+namespace mlir {
+namespace LLVM {
+/// Rewrites bf16 constants to their i16 equivalents
+/// This is relying on the fact that the vector, i16, and bf16 types used in the
+/// LLVM dialect are the standard ones and not weird custom wrappers
+struct BF16ConstCasting : OpRewritePattern<LLVM::ConstantOp> {
+  explicit BF16ConstCasting(MLIRContext *context) : OpRewritePattern(context) {}
+
+  llvm::APInt toInt(llvm::APFloat value) const {
```
krzysz00: Name: `castBf16ToInt` or such.

Also, make this a `static` helper outside the struct so [the people I've mainly been working with] upstream don't complain about why you didn't do that.

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,364 @@
...
+    if (auto valFloat = val.dyn_cast<mlir::FloatAttr>()) {
+      if (valFloat.getType() != bf16)
+        return failure();
+      APInt newVal = toInt(valFloat.getValue());
+      rewriter.replaceOpWithNewOp<LLVM::ConstantOp>(
+          rawOp, retType, rewriter.getIntegerAttr(retType, newVal));
+      return success();
+    }
+
+    if (auto valDense = val.dyn_cast<mlir::DenseElementsAttr>()) {
```
krzysz00: I had/you have some spurious `mlir::`s here?

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,364 @@
...
+
+template <typename Op>
+struct BF16AsF32 : OpRewritePattern<Op> {
+  explicit BF16AsF32(MLIRContext *context) : OpRewritePattern<Op>(context) {}
+
+  LogicalResult matchAndRewrite(Op op,
+                                PatternRewriter &rewriter) const override {
+
+    Location loc = op.getLoc();
+    Type opdType = op->getOperand(0).getType();
```
krzysz00, line 91: Might as well expand to `operandType` and `resultType`

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,364 @@
...
+    return success();
+  }
+};
+
+Value getLlvmI32Const(Location loc, PatternRewriter &rewriter, Type type,
+                      int32_t value) {
+  Attribute ret = rewriter.getI32IntegerAttr(value);
+  if (LLVM::isCompatibleVectorType(type))
+    ret = SplatElementsAttr::get(type.cast<ShapedType>(), ret);
+  return rewriter.create<LLVM::ConstantOp>(loc, type, ret);
```
krzysz00: Can this be `createOrFold`?

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,364 @@
...
+      Op operation =
+          rewriter.create<Op>(loc, resType, extended, op->getAttrs());
+      rewriter.replaceOp(op.getOperation(), {operation});
+    }
+
+    return success();
+  }
+};
+
+Value getLlvmI32Const(Location loc, PatternRewriter &rewriter, Type type,
```
krzysz00: `static`

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,364 @@
...
+  for (Operation &op : block.getOperations())
+    replaceBF16InOp(&op, converter);
+}
+
+void replaceBF16InRegion(Region &region, LLVMTypeConverter &converter) {
+  for (Block &block : region.getBlocks())
+    replaceBF16InBlock(block, converter);
+}
+
+void replaceBF16InOp(Operation *op, LLVMTypeConverter &converter) {
```
krzysz00: MLIR prefers a worklist over recursion - you can probably merge these into one function by using block and operand worklists

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,364 @@
...
+  return;
+}
+
+} // namespace LLVM
+} // namespace mlir
+
+void mlir::LLVM::populateBF16ToLLVMConversionPatterns(
+    LLVMTypeConverter &converter, RewritePatternSet &patterns) {
+  MLIRContext *ctx = converter.getDialect()->getContext();
+  // AMD GPUs don't have a backend that understands bfloat, even though LLVM's
```
krzysz00, line 263: Make comment slightly more generic - maybe "Some LLVM targets, such as"

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,364 @@
...
+  // Override for vector/struct types since they get caught by
+  // isCompatibleType(), which doesn't convert the element type
+  converter.addConversion(
+      [llvmI16, bf16](mlir::VectorType type) -> Optional<Type> {
+        if (type.getElementType() == bf16 && type.getRank() == 1)
+          return type.clone(llvmI16);
+        return llvm::None; // continue search
+      });
+
+  converter.addConversion([&](LLVM::LLVMStructType type) -> Optional<Type> {
```
krzysz00: Use the three-argument form of conversion that takes a worklist of types in order to handle the recursion

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,364 @@
...
+      return type;
+
+    if (type.isIdentified()) {
+      auto convertedType = LLVM::LLVMStructType::getIdentified(
+          type.getContext(), ("_Converted_" + type.getName()).str());
+      unsigned counter = 1;
+      while (convertedType.isInitialized()) {
+        convertedType = LLVM::LLVMStructType::getIdentified(
+            type.getContext(),
+            ("_Converted_" + std::to_string(counter++) + type.getName()).str());
```
krzysz00: I think there might be some usage of `Twine` possible here to avoid the `to_string`

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,364 @@
...
+
+namespace {
+struct SoftwareBF16Pass : public SoftwareBF16Base<SoftwareBF16Pass> {
+  void runOnOperation() override {
+    auto m = getOperation();
+    MLIRContext *ctx = m->getContext();
+    LowerToLLVMOptions options(ctx);
+    LLVMTypeConverter converter(ctx, options);
+    RewritePatternSet bf16fixupPatterns(ctx);
+    RewritePatternSet llvmPatterns(ctx);
```
krzysz00: Dead variable

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,364 @@
...
+    LLVMConversionTarget target(getContext());
```
krzysz00: You shouldn't need one of these

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/Transforms/Passes.td
```diff
@@ -16,4 +16,9 @@ def LLVMLegalizeForExport : Pass<"llvm-legalize-for-export"> {
   let constructor = "mlir::LLVM::createLegalizeForExportPass()";
 }
 
+def SoftwareBF16 : Pass<"llvm-software-bf16"> {
+  let summary = "Convert BF16 to I16 in LLVM IR";
```
krzysz00, line 20: Add `description`?

external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
```diff
@@ -107,7 +109,8 @@ struct LowerGpuOpsToROCDLOpsPass
...
     // Note that since these patterns touch LLVM ops, they'll need to run after
     // conversion
-    populateBF16ToROCDLConversionPatterns(converter, bf16fixupPatterns);
+    mlir::LLVM::populateBF16ToLLVMConversionPatterns(converter,
```
krzysz00: Let's not do it this way. Instead, use `NestedPassManager` (I think I remember the name correctly) to run the pass after the conversion call

krzysz00 (commented):
I've got a bunch of comments, some minor nitpicking, some less so - this is from recent experience polishing up a contribution to upstream.

On top of this, here's a bigger one: add tests for the new pass down under `external`.

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/Transforms/Passes.h
```diff
@@ -10,12 +10,14 @@
...
 
 #include "mlir/Dialect/LLVMIR/Transforms/LegalizeForExport.h"
+#include "mlir/Dialect/LLVMIR/Transforms/SoftwareBF16.h"
 #include "mlir/Pass/Pass.h"
 
 namespace mlir {
 
 namespace LLVM {
 
+std::unique_ptr<Pass> createSoftwareBF16Pass();
```
krzysz00, line 20: Comment - and a newline after the declaration.

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/Transforms/Passes.td
```diff
@@ -16,4 +16,15 @@ def LLVMLegalizeForExport : Pass<"llvm-legalize-for-export"> {
...
+  let description = [{
+    This pass handles BF16 types.
```
krzysz00:
```
This pass erases the BF16 type from LLVM IR.

Some LLVM targets do not support LLVM's `bfleat` type, or only support it incompletely.
To allow using the `bf`16` type on such targets, this pass replaces all of its uses by `i16`
and then replaces operations on `bf16` by extending the 16-bit values into `f32`, then
computes the floating-point operation on the extended value, and then truncates the results.
```

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/Transforms/SoftwareBF16.h
```diff
@@ -0,0 +1,31 @@
...
+
+namespace mlir {
+class Operation;
+class LLVMTypeConverter;
+class RewritePatternSet;
+class Pass;
+
+namespace LLVM {
+
+/// Since some LLVM targets like AMD GPU and X86 don't natively support bfloat
```
krzysz00: It's spelled `AMDGPU`.

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/Transforms/SoftwareBF16.h
```diff
@@ -0,0 +1,31 @@
...
+/// as a type, even though LLVM frontend does, populate patterns and type
+/// conversions to convert uses of bf16 to i16. To be removed when/if the
+/// backend has been taught about bfloat.
+void populateSoftwareBF16Patterns(LLVMTypeConverter &, RewritePatternSet &);
```
krzysz00: I still don't think this is the right architecture.

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/CMakeLists.txt
```diff
@@ -8,4 +9,5 @@ add_mlir_dialect_library(MLIRLLVMIRTransforms
   MLIRIR
   MLIRLLVMIR
   MLIRPass
+  MLIRLLVMCommonConversion
```
krzysz00, line 12: Do we need this?

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,367 @@
...
+#include "mlir/Conversion/LLVMCommon/TypeConverter.h"
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "mlir/Dialect/LLVMIR/Transforms/Passes.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+using namespace mlir;
+
+namespace {
+static llvm::APInt castBF16toInt(llvm::APFloat value) {
```
krzysz00: `static`s go outside of the anon namespace.

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,367 @@
...
+  }
+};
+
+void replaceBF16WithInt8(Operation *op, const LLVMTypeConverter &cvt) {
+  auto converter = const_cast<LLVMTypeConverter &>(cvt);
+  if (auto func = dyn_cast<LLVM::LLVMFuncOp>(op)) {
+    auto funcType = func.getFunctionType();
+    func.setType(converter.convertType(funcType));
+    for (auto arg : func.getArguments())
+      arg.setType(converter.convertType(arg.getType()));
```
krzysz00, line 242: I don't think this needs to be a `LLVMTypeConverter` - just a `TypeConverter`.

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,367 @@
...
+        loc, bitcastType, addValue, constantSixteen);
+    Value truncValue =
+        rewriter.create<LLVM::TruncOp>(loc, destType, shiftBeforeTruncValue);
+    rewriter.replaceOp(op.getOperation(), {truncValue});
+
+    return success();
+  }
+};
+
+void replaceBF16WithInt8(Operation *op, const LLVMTypeConverter &cvt) {
```
krzysz00: WithInt16

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,367 @@
...
+namespace LLVM {
+
+void populateSoftwareBF16Patterns(LLVMTypeConverter &converter,
+                                  RewritePatternSet &patterns) {
+  MLIRContext *ctx = converter.getDialect()->getContext();
+  // AMD GPUs don't have a backend that understands bfloat, even though LLVM's
+  // frontend does. Remove this if/when that changes. Note that adding
+  // conversions after the default constructor runs gives them priority
+  // over the defaults.
+  Type llvmI16 = converter.convertType(IntegerType::get(ctx, 16));
```
krzysz00: No need to grab a `llvmI16` here - it's the native type

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+  converter.addConversion(
+      [&](LLVMStructType type, SmallVectorImpl<Type> &results,
+          ArrayRef<Type> callStack) -> Optional<LogicalResult> {
+        bool converted = false;
+        SmallVector<Type> convertedElemTypes;
+        convertedElemTypes.reserve(type.getBody().size());
+        for (auto t : type.getBody()) {
+          SmallVector<Type, 1> element;
+          if (failed(converter.convertType(t, element)))
```
krzysz00, line 295: Shouldn't this take the call stack?

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,367 @@
...
+  }
+  return;
+}
+
+} // namespace
+
+namespace mlir {
+namespace LLVM {
+
+void populateSoftwareBF16Patterns(LLVMTypeConverter &converter,
```
krzysz00: I think this can be a static method and not exposed in the API

external/llvm-project/mlir/test/Conversion/SoftwareBF16/softwareBF16.mlir
```diff
@@ -0,0 +1,57 @@
...
+//CHECK:    %{{.*}} = llvm.bitcast %arg1 : f32 to i32
+//CHECK-NEXT:    %{{.*}} = llvm.lshr %{{.*}}, %{{.*}}  : i32
+//CHECK-NEXT:    %{{.*}} = llvm.and %{{.*}}, %{{.*}}  : i32
+//CHECK-NEXT:    %{{.*}} = llvm.add %{{.*}}, %{{.*}}  : i32
+//CHECK-NEXT:    %{{.*}} = llvm.add %{{.*}}, %{{.*}}  : i32
+//CHECK-NEXT:    %{{.*}} = llvm.lshr %{{.*}}, %{{.*}}  : i32
+//CHECK-NEXT:    %{{.*}} = llvm.trunc %{{.*}} : i32 to i16
+
+    %4 = llvm.fsub %arg0, %3  : bf16
+//CHECK:    %{{.*}} = llvm.zext %arg0 : i16 to i32
```
krzysz00: Variables exist, `%[[thing:.*]]` defines one, `%[[thing]]` references it

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,363 @@
...
+#include "mlir/Conversion/LLVMCommon/Pattern.h"
+#include "mlir/Conversion/LLVMCommon/TypeConverter.h"
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "mlir/Dialect/LLVMIR/Transforms/Passes.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+using namespace mlir;
+
+static llvm::APInt castBF16toInt(llvm::APFloat value) {
```
krzysz00: Very minor note - this can be spelled just `APInt` (not sure about `APFloat`)

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
yiqian1: The call stack contains the types that are being converted (not the converted types). The current type is the last one in the call stack.

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,363 @@
...
+    ret = SplatElementsAttr::get(type.cast<ShapedType>(), ret);
+  return rewriter.createOrFold<LLVM::ConstantOp>(loc, type, ret);
+}
+
+namespace {
+/// Rewrites bf16 constants to their i16 equivalents
+/// This is relying on the fact that the vector, i16, and bf16 types used in the
+/// LLVM dialect are the standard ones and not weird custom wrappers
+struct BF16ConstCasting : OpRewritePattern<LLVM::ConstantOp> {
+  explicit BF16ConstCasting(MLIRContext *context) : OpRewritePattern(context) {}
```
krzysz00, line 38: I think this can be handled by `using` the constructors from OpRewritePattern, but that might be personal preference...

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,363 @@
...
+      return success();
+    }
+    // No match otherwise
+    return failure();
+  }
+};
+
+template <typename Op>
+struct BF16AsF32 : OpRewritePattern<Op> {
+  explicit BF16AsF32(MLIRContext *context) : OpRewritePattern<Op>(context) {}
```
krzysz00, line 85: Same on `using`

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/CMakeLists.txt
yiqian1: Yes. Otherwise the compiler cannot find the type-conversion, and op-rewriter related methods.

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+    if (resElementType == i16) {
+      Op operation =
+          rewriter.create<Op>(loc, extType, extended, op->getAttrs());
+      rewriter.replaceOpWithNewOp<LLVM::FPTruncOp>(op, resType,
+                                                   operation.getResult());
+    } else { // FCmp
+      Op operation =
+          rewriter.create<Op>(loc, resType, extended, op->getAttrs());
+      rewriter.replaceOp(op.getOperation(), {operation});
```
krzysz00: Can this be just one `replaceOpWithNewOp`?

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+    Type i16 = rewriter.getIntegerType(16);
+    if (auto destShaped = destType.dyn_cast<ShapedType>()) {
+      if (destShaped.getElementType() != i16)
+        return failure();
+    } else if (destType != i16)
+      return failure();
+
+    // a = bitcast f32 value to i32
+    // b = (a + 32767) << 16
```
krzysz00: These comments have the `<<` in the wrong direction, and I somehow didn't notice until now. They should show `>>`

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,363 @@
...
+        rewriter.create<LLVM::TruncOp>(loc, destType, shiftBeforeTruncValue);
+    rewriter.replaceOp(op.getOperation(), {truncValue});
+
+    return success();
+  }
+};
+
+} // namespace
+
+static void replaceBF16WithI16(Operation *op, const TypeConverter &cvt) {
```
krzysz00: Why's this take a const reference in the first place?

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,363 @@
...
+  Type llvmI16 = IntegerType::get(ctx, 16);
+  Type bf16 = BFloat16Type::get(ctx);
+  converter.addConversion(
+      [llvmI16](mlir::BFloat16Type type) -> Type { return llvmI16; });
+
+  // Override for vector/struct types since they get caught by
+  // isCompatibleType(), which doesn't convert the element type
+  converter.addConversion(
+      [llvmI16, bf16](mlir::VectorType type) -> Optional<Type> {
+        if (type.getElementType() == bf16 && type.getRank() == 1)
```
krzysz00: Why only rank 1? I'm not sure that's a restriction we have

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
krzysz00: That is, you want to add the `t` you're converting to the call stack ... which feels like it should be a `SmallVectorImpl &` or something

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,363 @@
...
+      [&](LLVM::LLVMStructType type, SmallVectorImpl<Type> &results,
+          ArrayRef<Type> callStack) -> Optional<LogicalResult> {
+        bool converted = false;
+        SmallVector<Type> convertedElemTypes;
+        convertedElemTypes.reserve(type.getBody().size());
+        for (auto t : type.getBody()) {
+          SmallVector<Type, 1> element;
+          if (failed(converter.convertType(t, element)))
+            return llvm::None;
+          convertedElemTypes.push_back(element[0]);
```
krzysz00, line 298: Can we get an assert here that the conversion was 1 to 1?

external/llvm-project/mlir/test/Conversion/SoftwareBF16/softwareBF16.mlir
```diff
@@ -0,0 +1,57 @@
+// RUN: mlir-opt --llvm-software-bf16 %s| FileCheck %s
+
+module attributes {llvm.data_layout = ""} {
+  llvm.func @verify_bf16_f32(%arg0: bf16, %arg1: f32) -> i32 {
+//CHECK:  llvm.func @verify_bf16_f32(%arg0: i16, %arg1: f32) -> i32 {
+
+    %0 = llvm.mlir.constant(0 : i32) : i32
+    %1 = llvm.mlir.constant(1 : i32) : i32
+//CHECK:    %[[V0:.*]] = llvm.mlir.constant(32767 : i32) : i32
```
krzysz00: CHECK-DAG on the constants. Also, I'm not convinced by the extra whitespace here

krzysz00 (commented):
This is looking pretty solid to me.

I might be of the opinion that having the tests be split into different functions will help readability, but still.

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
yiqian1: The lambda function used to capture converter by value. I should have used capture by reference I think.

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
yiqian1: And the shift in (a+32767)>>16 was missing.

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
yiqian1: This is how it is defined in TypeConverter class. What I'm doing here is almost identical with the default StructType converter.

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/CMakeLists.txt
krzysz00: I think it's possible those live in a smaller dependency - something like `MLIRSupport` and `MLIRTransforms` ... but it's probably fine as is.

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,364 @@
...
+    return success();
+  }
+};
+
+struct SoftwareBF16Ext : OpRewritePattern<LLVM::FPExtOp> {
+  explicit SoftwareBF16Ext(MLIRContext *context) : OpRewritePattern(context) {}
+
+  LogicalResult matchAndRewrite(LLVM::FPExtOp op,
+                                PatternRewriter &rewriter) const override {
+    auto loc = op.getLoc();
```
krzysz00: Can we de-`auto` here? We'll get complained at for good reason (that being, upstream likes `auto` only when not using it would repeat obvious info)

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,364 @@
...
+  }
+};
+
+} // namespace
+
+static void replaceBF16WithI16(Operation *op, TypeConverter &converter) {
+  if (auto func = dyn_cast<LLVM::LLVMFuncOp>(op)) {
+    auto funcType = func.getFunctionType();
+    func.setType(converter.convertType(funcType));
+    for (auto arg : func.getArguments())
```
krzysz00: Let's go `Value arg` here

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,364 @@
...
+      arg.setType(converter.convertType(arg.getType()));
+  } else if (auto globalOp = dyn_cast<LLVM::GlobalOp>(op)) {
+    auto globalType = globalOp.getType();
```
krzysz00: Unclear `auto`, as above

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,364 @@
...
+    }
+    for (unsigned idx = 0; idx < op->getNumResults(); idx++) {
+      auto type = converter.convertType(op->getResult(idx).getType());
+      op->getResult(idx).setType(type);
+    }
+  }
+  return;
+}
+
+static void populateSoftwareBF16Patterns(LLVMTypeConverter &converter,
```
krzysz00: I insist this can just be a `TypeConverter` - do you have counterexamples?

krzysz00: (you _might_ need to add the fallback rule that types convert to themselves if not otherwise specified ... I can't quite remember how the type conversion machinery handles the default case)

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
krzysz00: Yeah, I looked, you're right

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
```diff
@@ -0,0 +1,364 @@
...
+               BF16AsF32<LLVM::FCeilOp>, BF16AsF32<LLVM::FFloorOp>,
+               BF16AsF32<LLVM::FMAOp>, BF16AsF32<LLVM::FMulAddOp>>(ctx);
+}
+
+namespace {
+struct SoftwareBF16Pass : public SoftwareBF16Base<SoftwareBF16Pass> {
+  void runOnOperation() override {
+    auto m = getOperation();
+    MLIRContext *ctx = m->getContext();
+    LowerToLLVMOptions options(ctx);
```
krzysz00: (To state my point differently - I claim that, since we're not converting to LLVM, we don't need of the machinery for that conversion ... at this point, it's already happened, and so the regular `TypeConverter` API will be enough)

krzysz00 (commented):
I think we're pretty much there, plus-minus fiddling with the type of the converter

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
yiqian1: LLVMTypeConverter can handle the default case, while TypeConverter cannot

krzysz00: LLVMTypeConverter doesn't actually handle the default case - it just has a bunch of conversions. You can get much the same behavior by having the first conversion you add be `(Type t) { return t; }`

krzysz00: ... sorry, didn't see your commit

external/llvm-project/mlir/lib/Dialect/LLVMIR/Transforms/SoftwareBf16.cpp
krzysz00: This here comment can probably disappear, since the pass documentation explains what it does.

krzysz00 (approved):
I have the one minor nitpick about a comment, but I think this is in a good state to land.

Would you be wiling to start the upstreaming process for this as well?

yiqian1:
> I have the one minor nitpick about a comment, but I think this is in a good state to land.
> 
> Would you be wiling to start the upstreaming process for this as well?

Yes, I will. Thank you for valuable feedbacks.

PR #622, opened 2022-05-13
Remove bwd data restriction, take 2

krzysz00:
Depends on #619 

Testing with #619 applied revealed that that bug was also a cause of the store issue - which explains why in my previous attempts to debug the "store issue" showed that entire rows of batches (particular `n` coordinates) were missing.

In addition, the kPackLogic for backward data wasn't being given the correct transformation when the output tensor was padded, so I fixed that up too.

I can report a successful parameter sweep of backward data kernels and a successful nightly run on MI-200 .

Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/259 , https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/529 , and https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/297


whchung (approved):
Nice!

PR #623, opened 2022-05-16
The big GemmContext+ refactoring

krzysz00:
Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/525

- Define a GemmContext struct with m, k, and n fields and make various
utility functions use it.
- Similarly, define ConvolutionDims struct, migrate utility methods to
it.
- Add methods to broader structs like Conv2DGenerator and ConvContext
that produce a ConvolutionDims from that struct
- Migrate the lowering passes to use ConvContext because we apparently
hadn't done that yet.
- Factor out the fetching of filter/input/output names in ConvToGemm


mlir/include/mlir/Dialect/MIOpen/Tuning/ConvContext.h
```diff
@@ -27,33 +28,57 @@ struct DimIndexAndSize {
...
-        strideVal(stride), dilationVal(dilation), paddingVal(padding),
-        gemmId(gemmid), dataType(type) {}
+        strideVal(), dilationVal(), paddingVal(), gemmId(gemmid),
+        dataType(type) {
+    strideVal.reserve(stride.size());
+    llvm::copy(stride, std::back_inserter(strideVal));
+    dilationVal.reserve(dilation.size());
+    llvm::copy(dilation, std::back_inserter(dilationVal));
+    paddingVal.reserve(padding.size());
+    llvm::copy(padding, std::back_inserter(paddingVal));
```
jerryyin: Why do the copies manually?

mlir/lib/Dialect/MIOpen/Tuning/ConvContext.cpp
```diff
@@ -17,7 +17,7 @@ populateDimIndexAndSize(const ArrayAttr &layoutAttr,
 }
 
 static void populateSeqVal(const ArrayAttr &seqAttr,
-                           llvm::SmallVector<int64_t, 0> &seqVal) {
+                           SmallVectorImpl<int64_t> &seqVal) {
```
jerryyin, line 20: Good job

mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
```diff
@@ -156,23 +148,16 @@ void AffixTuningParameters::affixBackwardWeightUtilityKernels(
...
+        GemmContext::fromConvolution(ConvOpType::BwdWeight, convDims);
     int64_t gemmMExtra, gemmNExtra, gemmKExtra;
     gemmMExtra = gemmNExtra = gemmKExtra = 0;
 
     bool needExtraPad = false;
     PopulateParamsXDL populateParamsXDL;
     std::tie(needExtraPad, gemmMExtra, gemmNExtra, gemmKExtra) =
-        calculatePaddingKernelSize(gemmMSize, gemmNSize, gemmKSize,
-                                   obtainConvDirection(op),
+        calculatePaddingKernelSize(gemmSize, obtainConvDirection(op),
```
jerryyin: Looking at how this goes, I think we can make the padding values in as another gemm context because it provides better organization of data compared to a tuple.

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -82,6 +84,49 @@ LogicalResult checkNames(ArrayRef<StringRef> actual,
   return success();
 }
 
+/// Get the dimension names for the given `op` into `filterNames`, `inputNames`
+/// and `outputNames`, returning failure if `op`'s layout doesn't contain all of
+/// the expected dimension names.
+template <typename T>
+LogicalResult getConvDimNames(T op, SmallVectorImpl<StringRef> &filterNames,
+                              SmallVectorImpl<StringRef> &inputNames,
+                              SmallVectorImpl<StringRef> &outputNames) {
```
jerryyin, line 93: I feel like all use case of `vector<StringRef> *Name` can be covered easier with the `dimIndexAndSize` data structure that actually associate name with layout and size. 

(In its use case) "expandNamesInPlace", for example, would not need another loop just to find the corresponding matching layout from the vector.

https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/d60ec55b81c06c3ade787bca799970a16312ec9a/mlir/lib/Dialect/MIOpen/Tuning/ConvContext.cpp#L6-L17

And its use case

https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/d60ec55b81c06c3ade787bca799970a16312ec9a/mlir/lib/Dialect/MIOpen/Tuning/ConvContext.cpp#L71-L82

mlir/include/mlir/Dialect/MIOpen/Tuning/ConvContext.h
krzysz00: I don't think there's a SmallVector from ArrayRef constructor, unless I'm missing one

jerryyin: I'll give an example based on iterator initialization. Could below work in the initializer list: `strideVal(stride.begin(), stride.end()), dilationVal(...) ...`

krzysz00: ... good point, it probably can.

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
krzysz00: To fix in the future.

jerryyin (approved):
LGTM thanks

PR #624, opened 2022-05-17
[hotfix] Move around installs in the Dockerfile so it can build

krzysz00:


krzysz00:
(this is what i get for not testing the Docker changes)

krzysz00:
(Update, having worked out how to test this stuff, we're still missing something)

jerryyin (approved):
Approved but please leave some inline explanations on downgrade.

PR #628, opened 2022-05-20
[DO NOT SQUASH][NFC] Create amdgpu.mfma and corresponding enum, use it

krzysz00:
This includes
- Creating `amdgpu.mfma`, a `MFMAInstr` enum that it uses to know which MFMA we want, and lowering to ROCDL + tests
- Removing `gpu.mfma`
- Switching xdlops code selection to `amdgpu::MFMAInstr` instead of strings

The first commit, which creates `amdgpu.mfma`, is meant to be upstreamed.

This also moves the vector of i8 -> i32 transformation all the way to the LLVM conversion where it makes sense.

Finally, since I was working off the AMDGPU intrinsics table for this, this adds support for all xdlops currently defined, whether or not we use them.


external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
```diff
@@ -129,25 +129,35 @@ class ROCDL_Mfma_IntrOp<string mnemonic, list<Trait> traits = []> :
...
+def ROCDL_mfma_f32_32x32x4bf16_1k : ROCDL_Mfma_IntrOp<"mfma.f32.32x32x4bf16.1k">;
+def ROCDL_mfma_f32_16x16x4bf16_1k : ROCDL_Mfma_IntrOp<"mfma.f32.16x16x4bf16.1k">;
+def ROCDL_mfma_f32_4x4x4bf16_1k : ROCDL_Mfma_IntrOp<"mfma.f32.4x4x4bf16.1k">;
+def ROCDL_mfma_f32_32x32x8bf16_1k : ROCDL_Mfma_IntrOp<"mfma.f32.32x32x8bf16.1k">;
+def ROCDL_mfma_f32_16x16x16bf16_1k : ROCDL_Mfma_IntrOp<"mfma.f32.16x16x16bf16.1k">;
+def ROCDL_mfma_f64_16x16x4f64 : ROCDL_Mfma_IntrOp<"mfma.f64.16x16x4f64">;
+def ROCDL_mfma_f64_4x4x4f64 : ROCDL_Mfma_IntrOp<"mfma.f64.4x4x4f64">;
+def ROCDL_mfma_i32_16x16x32_i8 : ROCDL_Mfma_IntrOp<"mfma.i32.16x16x32.i8">;
+def ROCDL_mfma_i32_32x32x16_i8 : ROCDL_Mfma_IntrOp<"mfma.i32.32x32x16.i8">;
+def ROCDL_mfma_f32_16x16x8_xf32 : ROCDL_Mfma_IntrOp<"mfma.f32.16x16x8.xf32">;def ROCDL_mfma_f32_32x32x4_xf32 : ROCDL_Mfma_IntrOp<"mfma.f32.32x32x4.xf32">;
```
jerryyin: Missing an enter between last two lines?

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
```diff
@@ -129,25 +129,35 @@ class ROCDL_Mfma_IntrOp<string mnemonic, list<Trait> traits = []> :
...
+def ROCDL_mfma_f32_32x32x2bf16 : ROCDL_Mfma_IntrOp<"mfma.f32.32x32x2bf16">;
+def ROCDL_mfma_f32_16x16x2bf16 : ROCDL_Mfma_IntrOp<"mfma.f32.16x16x2bf16">;
+def ROCDL_mfma_f32_4x4x2bf16 : ROCDL_Mfma_IntrOp<"mfma.f32.4x4x2bf16">;
+def ROCDL_mfma_f32_32x32x4bf16 : ROCDL_Mfma_IntrOp<"mfma.f32.32x32x4bf16">;
+def ROCDL_mfma_f32_16x16x8bf16 : ROCDL_Mfma_IntrOp<"mfma.f32.16x16x8bf16">;
+def ROCDL_mfma_f32_32x32x4bf16_1k : ROCDL_Mfma_IntrOp<"mfma.f32.32x32x4bf16.1k">;
+def ROCDL_mfma_f32_16x16x4bf16_1k : ROCDL_Mfma_IntrOp<"mfma.f32.16x16x4bf16.1k">;
+def ROCDL_mfma_f32_4x4x4bf16_1k : ROCDL_Mfma_IntrOp<"mfma.f32.4x4x4bf16.1k">;
+def ROCDL_mfma_f32_32x32x8bf16_1k : ROCDL_Mfma_IntrOp<"mfma.f32.32x32x8bf16.1k">;
+def ROCDL_mfma_f32_16x16x16bf16_1k : ROCDL_Mfma_IntrOp<"mfma.f32.16x16x16bf16.1k">;
```
jerryyin, line 157: Could you remind me what `_1k` (and `_i8`) means?

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -234,6 +237,130 @@ struct RawBufferOpLowering : public ConvertOpToLLVMPattern<GpuOp> {
...
+
+/// Return the `rocdl` intrinsic corresponding to a `MFMAInstr` value.
+/// This conversion happens here to allow code up the stack to handle the choice
+/// of mfma by picking between enum variants, which is much more ergonomic than
+/// picking between ops, at the cost of some long switch statements in this
+/// pass.
+static StringRef mfmaInstrToIntrinsicName(MFMAInstr instr) {
+  switch (instr) {
+  case MFMAInstr::f32_32x32x1f32:
+    return ROCDL::mfma_f32_32x32x1f32::getOperationName();
```
jerryyin: This switch case make spelling correctness essential for correctness. Could we do some macro expansion like below so as to not worry about possible mismatches?

```
#define mapMfmaToROCDL(INSTR) case MFMAInstr::INSTR : return ROCDL::mfma_INSTR::getOperationName();
```

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -234,6 +237,130 @@ struct RawBufferOpLowering : public ConvertOpToLLVMPattern<GpuOp> {
...
+/// vectors of bytes instead integers. Since we do not want to expose this
+/// implementation detail to MLIR, we correct for it here.
+static Value mfmaConcatIfNeeded(ConversionPatternRewriter &rewriter,
+                                Location loc, Value input) {
+  Type inputType = input.getType();
+  if (auto vectorType = inputType.dyn_cast<VectorType>()) {
+    if (vectorType.getElementType() != rewriter.getI8Type())
+      return input;
+    int64_t numBytes = vectorType.getNumElements();
+    Type destType = rewriter.getIntegerType(numBytes * 8);
```
jerryyin, line 253: I wonder if this is too relaxed in the condition that allows some untested case to go through? I think we want to make sure:
 - If this is not i8 but is vector, throw a debug message that we didn't do anything, but the result may not be expected.
 - If this is i8 but numBytes != 4, throw an error that this is unexpected, just to be defensive

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
```diff
@@ -129,25 +129,35 @@ class ROCDL_Mfma_IntrOp<string mnemonic, list<Trait> traits = []> :
 }
 
 def ROCDL_mfma_f32_32x32x1f32 : ROCDL_Mfma_IntrOp<"mfma.f32.32x32x1f32">;
+def ROCDL_mfma_f32_16x16x1f32 : ROCDL_Mfma_IntrOp<"mfma.f32.16x16x1f32">;
+def ROCDL_mfma_f32_4x4x1f32 : ROCDL_Mfma_IntrOp<"mfma.f32.4x4x1f32">;
```
jerryyin, line 134: In what sequence did you reordered these intrinsics? Looks like you regrouped them towards reduction/non-reduction. Mind adding some comments in between each group?

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/AMDGPU.td
```diff
@@ -163,4 +164,98 @@ def AMDGPU_RawBufferAtomicFaddOp :
...
+                             VectorOfLengthAndType<[4, 8], [I8]>]>;
+def MFMAOutTypes : AnyTypeOf<[F64,
+                              VectorOfLengthAndType<[4, 16, 32], [F32]>,
+                              VectorOfLengthAndType<[4, 16, 32], [I32]>,
+                              VectorOfLengthAndType<[4], [F64]>]>;
+
+def AMDGPU_MFMAOp :
+    AMDGPU_Op<"mfma", [AllTypesMatch<["sourceA", "sourceB"]>,
+                        AllTypesMatch<["destC", "destD"]>]>,
+    Arguments<(ins AMDGPU_MFMAInstrAttr:$instr,
```
jerryyin, line 226: Should we refactor the rocdl intrinsics to follow the same convention? Such that there's only a single op, but the first field of the op is a enum the specifies what it does.

external/llvm-project/mlir/lib/Dialect/AMDGPU/IR/AMDGPUDialect.cpp
```diff
@@ -44,17 +52,145 @@ static LogicalResult verifyRawBufferOp(T &op) {
...
+  StringRef instrName = stringifyMFMAInstr(instr());
+
+  Type inType = sourceA().getType();
+  switch (instr()) {
+  case MFMAInstr::f32_32x32x1f32:
+  case MFMAInstr::f32_16x16x1f32:
+  case MFMAInstr::f32_4x4x1f32:
+  case MFMAInstr::f32_32x32x2f32:
+  case MFMAInstr::f32_16x16x4f32:
+    if (inType != b.getF32Type())
```
jerryyin, line 77: I guess add the same inspection on
 - sourceB
 - and DestC

This applies to below cases

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
krzysz00: No idea what `_1k` is - it seems to be a new xdlops form.

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
krzysz00: I might be able to do that, depending on how token concat works, good catch

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
krzysz00: If we get a non-i8 vector, that's either a valid input or LLVM will yell.

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
krzysz00: I used the exact order of the LLVM tablegen file for AMDGPU intrinsics.

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/AMDGPU.td
krzysz00: I'd like the things in `rocdl` to be 1-1 wrappers around their LLVM counterparts

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
krzysz00: Also, there is an xdlop that takes an i64 to stand for `8xi8` now

external/llvm-project/mlir/lib/Dialect/AMDGPU/IR/AMDGPUDialect.cpp
krzysz00: "AllTypesMatch" in the tablegen will verify that

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
krzysz00: And done

jerryyin: Thank you!

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
jerryyin: Sounds good.

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/AMDGPU.td
jerryyin: That's right, it is a convention we've been following.

external/llvm-project/mlir/lib/Dialect/AMDGPU/IR/AMDGPUDialect.cpp
jerryyin: Thanks for letting me know

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
jerryyin: Make sense, the idea seem to be that we only need to periodically regenerate it according to perl and keep its natural sequence. I hope there's a way to guard different intrinsic according to ROCm versions. (which makes us backward compatible to an older ROCm)

krzysz00: Fundamentally, if someone uses an intrinsic for hardware they're not compiling for, the backend will complain at them. Before that, everything should be fine in that we're tracking upstream pretty closely so all these intrinsics will at least be defined, if not necessarily available.

Knowing your `-mcpu` and picking appropriate `xdlops` is probably more for `XdlopsCodeSelection.h` than here

jerryyin (commented):
From a review perspective, I'm ready to merge it after last clarification.

Could you also help clarify the requirement of backward compatibility between us and older ROCm? It looks like after this PR gives us the possibility of generating some ISA that would be impossible to run with older rocm/hardware. To what degree is this a requirement?

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
jerryyin: How'd we make sure to generate the correct isa? The thing is that we don't seem to worry about that even with user giving us the "architecture, features" bit. If the machine supports xdlops it looks like we tend to use the fastest possible instructions without worrying about whether this is legal to the user.

krzysz00:
I'm not sure why you're asking me about the backcompat requirements :P - I didn't set that policy

But yes, this adds wrappers around intrinsics that don't exist on all hardware and may not exist on hardware anyone can buy. However, it doesn't make us use them - if we want to avoid a "your hardware doesn't support this xdlop" situation, we either don't use the newer ones (how we currently do this) or have some if statements so we don't generate any `amdgpu.mfma` ops that aren't supported on the current target.

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
krzysz00: The thing is, currently, we never invoke any of the 90a-only xdlops. If we want to start using them, that's a separate PR

jerryyin: Fair enough. sooner or later we might need to do that and start to use the "architecture" field to decide what intrinsic to choose from. We may actually do want to use the newer ones for gfx90a, this sounds at least like an experiment to do, but has nothing to do with this PR.

jerryyin (approved):

PR #631, opened 2022-05-24
MLIR#545: Use upstream libmlir_rocm_runtime instead of librocm-runtime-wrappers

sjw36:
...for MLIR supported calls

- updated all mlir-rocm-runner tests

https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/545


sjw36:
Using %linalg_test_lib_dir since it points to the external/llvm-project/llvm/lib dir but perhaps we should rename it..

krzysz00:
Quick question: did you get the test generation files in `mlir/utils/`?

krzysz00:
I take it there's a future PR where we migrate to the `gpu.*` ops that lower to the external calls?

sjw36:
> I take it there's a future PR where we migrate to the `gpu.*` ops that lower to the external calls?

Exactly!

sjw36:
> Quick question: did you get the test generation files in `mlir/utils/`?

Everything in there seems to be using our locally added routines..

krzysz00:
@sjw36 What I'm asking about are the `.toml` files that are used to the e2e tests - those have a "suffix" parameter that includes the rocm-runner string. That needs to be changed to the new string to ensure we don't break things when someone uses the script to regenerate the tests.

sjw36:
@krzysz00 please re-review/approve.

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
```diff
@@ -276,14 +222,6 @@ extern "C" void mcpuMemset(float *allocated, float *aligned, int64_t offset,
...
 
-extern "C" StridedMemRefType<float, 1>
-mgpuMemAlloc(float *allocated, float *aligned, int64_t offset, int64_t size,
-             int64_t stride) {
-  float *gpuPtr;
-  hipMalloc((void **)&gpuPtr, size * sizeof(float));
-  return {gpuPtr, gpuPtr, offset, {size}, {stride}};
-}
-
 extern "C" void mgpuMemDealloc(float *allocated, float *aligned, int64_t offset,
```
krzysz00, line 225: I take it they don't have a free() upstream?

krzysz00 (approved):

krzysz00:
Good cleanup, let's get this in!

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
sjw36: Nope..

PR #633, opened 2022-05-25
convert passthrough linalg.generic to miopen.transform

jungpark-mlir:


mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -1582,10 +1583,85 @@ template struct Conv2DRewritePattern<Conv2DOp>;
...
 
+//===- MITPRewritePattern -------------------------------------------------===//
+//===-  ------------------------------------------------===//
+template <typename T> struct MITPRewritePattern : public OpRewritePattern<T> {
+  using OpRewritePattern<T>::OpRewritePattern;
+
+  miopen::TransformOp makeTranspose(PatternRewriter &b, Value inp,
+                                    const AffineMapAttr &inMap,
+                                    const AffineMapAttr &outMap) const {
+    auto inpIdxMap = inMap.getAffineMap();
```
krzysz00: Can we spell out these `auto`s please?

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -1582,10 +1583,85 @@ template struct Conv2DRewritePattern<Conv2DOp>;
...
+    auto inpShape = inpType.getShape();
+
+    SmallVector<uint32_t, 8> endDims;
+    SmallVector<uint32_t, 8> startDims;
+    for (uint32_t i = 0, e = inpShape.size(); i < e; ++i) {
+      startDims.push_back(i);
+      auto inMapped = inpIdxMap.getDimPosition(i);
+      endDims.push_back(outpIdxMap.getDimPosition(inMapped));
+    }
+    miopen::BottomUpTMBuilder transform(b, inpShape, loc);
```
krzysz00, line 1608: This may be easier to read if it were a TopDownTMBuilder

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -1582,10 +1583,85 @@ template struct Conv2DRewritePattern<Conv2DOp>;
...
+    transform.passThrough(endDims, startDims);
+    auto tfOp = b.create<miopen::TransformOp>(loc, inp, transform.get(),
+                                              inpType.getMemorySpaceAsInt());
+    return tfOp;
+  }
+
+  LogicalResult matchAndRewrite(T laGeneric,
+                                PatternRewriter &b) const override {
+    LogicalResult fail = failure();
```
krzysz00: I think having `fail` like this is an anti-pattern. Just call `failure()` - it's cheap.

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -1582,10 +1583,85 @@ template struct Conv2DRewritePattern<Conv2DOp>;
...
+    auto loc = laGeneric.getLoc();
+
+    // 0. Test compatibility
+    // 0.0. Only fully parallel for now
+    for (auto itr : laGeneric.iterator_types()) {
```
krzysz00: If there's a range, there's a method on `ArrayAttr` that does the casting for you.

Heck, `.getAsValueRange<StringAttr>()` should let you write this as `for (StringRef iter : ...)`

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -1582,10 +1583,85 @@ template struct Conv2DRewritePattern<Conv2DOp>;
...
+      if (itr.template cast<StringAttr>().getValue() != "parallel") {
+        return fail;
+      }
+    }
+
+    bool bPassing = false;
+    laGeneric.getRegion().walk([&](linalg::YieldOp yieldOp) {
+      auto laReturn = yieldOp->getOperand(0);
```
krzysz00: Let's say `Value` instead of `auto`. Also, is there a more specific accessor on `yieldOp()`? It's good to use names instead of indices where possible.

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -1582,10 +1583,85 @@ template struct Conv2DRewritePattern<Conv2DOp>;
...
+    });
+
+    // 0.1. Test if 1 to 1 pass through
+    if (laGeneric.inputs().size() != 1 || laGeneric.outputs().size() != 1 ||
+        !bPassing) {
+      return fail;
+    }
+
+    // 0.2. Test it only passes through and no other calculation
+    auto idxMaps =
```
krzysz00, line 1646: I think you're save to say `cast` instead of `template cast` - you only need that extra word in a few cases involving templated code. (and you'll get yelled at by the compiler if you do need it, so leave it off unless told otherwise)

krzysz00: On that note, is there an accessor function for this attribute?

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -1582,10 +1583,85 @@ template struct Conv2DRewritePattern<Conv2DOp>;
 template struct Conv2DRewritePattern<Conv2DBwdDataOp>;
 template struct Conv2DRewritePattern<Conv2DBwdWeightOp>;
 
+//===- MITPRewritePattern -------------------------------------------------===//
+//===-  ------------------------------------------------===//
+template <typename T> struct MITPRewritePattern : public OpRewritePattern<T> {
```
krzysz00: Why's this templated? Just operate on `linalg::Generic` by name

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -1582,10 +1583,85 @@ template struct Conv2DRewritePattern<Conv2DOp>;
...
+      return fail;
+    }
+
+    // 0.2. Test it only passes through and no other calculation
+    auto idxMaps =
+        laGeneric->template getAttrOfType<ArrayAttr>("indexing_maps");
+    auto inIdxMap = idxMaps[0].template cast<AffineMapAttr>();
+    auto outIdxMap = idxMaps[1].template cast<AffineMapAttr>();
+
+    auto tpTransform =
```
krzysz00, line 1650: Why do we need this? Can I get a comment?

mlir/test/fusion/tosa-tp-highlevel-fused.mlir
```diff
@@ -0,0 +1,14 @@
...
+func @test_fusion(%arg0: tensor<256x128x28x28xf32>, %arg1: tensor<64x128x3x3xf32>, %arg2: tensor<256x64x28x28xf32>) -> tensor<256x28x28x64xf32> attributes {kernel} {
+    %cst = arith.constant dense<[0, 2, 3, 1]> : tensor<4xi64>
+    %cst_0 = arith.constant dense<0.000000e+00> : tensor<1xf32>
+    %a = "tosa.transpose"(%arg0, %cst) : (tensor<256x128x28x28xf32>, tensor<4xi64>) -> tensor<256x28x28x128xf32>
+    %b = "tosa.transpose"(%arg1, %cst) : (tensor<64x128x3x3xf32>, tensor<4xi64>) -> tensor<64x3x3x128xf32>
+    %0 = "tosa.conv2d"(%a, %b, %cst_0) {dilation = [1, 1], expected_filter_layout = "kyxc", expected_input_layout = "nhwc", expected_output_layout = "nhwk", pad = [1, 1, 1, 1], stride = [1, 1]} : (tensor<256x28x28x128xf32>, tensor<64x3x3x128xf32>, tensor<1xf32>) -> tensor<256x28x28x64xf32>
+    %1 = "tosa.transpose"(%arg2, %cst) : (tensor<256x64x28x28xf32>, tensor<4xi64>) -> tensor<256x28x28x64xf32>
+    %2 = "tosa.add"(%0, %1) : (tensor<256x28x28x64xf32>, tensor<256x28x28x64xf32>) -> tensor<256x28x28x64xf32>
+    return %2 : tensor<256x28x28x64xf32>
+}
```
krzysz00: Newline

krzysz00 (commented):
Seems reasonable overall, here's your onslaught of style nits

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
jungpark-mlir: Doesn't seem like there is an accessor for it, this is literally checking the first operand of the yieldOp

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
jungpark-mlir: checked files but it doesn't have immediate accessor

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
jungpark-mlir: I put the wrong comment there, it might confused you. comment fixed.

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
jungpark-mlir: It doesn't seem that we don't have that version of passThrough for topdown currently

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -28,10 +28,12 @@
 #include "mlir/Dialect/MIOpen/Tuning/UtilityParams.h"
 #include "mlir/Dialect/MIOpen/utility/builderUtils.h"
 #include "mlir/Dialect/MIOpen/utility/loweringUtils.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
```
krzysz00, line 31: Do we need this? (that is, check what happens if you remove it)

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -1582,10 +1583,89 @@ template struct Conv2DRewritePattern<Conv2DOp>;
 template struct Conv2DRewritePattern<Conv2DBwdDataOp>;
 template struct Conv2DRewritePattern<Conv2DBwdWeightOp>;
 
+//===- MITPRewritePattern -------------------------------------------------===//
+//===-  ------------------------------------------------===//
```
krzysz00: Please fix the formatting of this header

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
```diff
@@ -1582,10 +1583,89 @@ template struct Conv2DRewritePattern<Conv2DOp>;
...
+struct MITPRewritePattern : public OpRewritePattern<linalg::GenericOp> {
+  using OpRewritePattern<linalg::GenericOp>::OpRewritePattern;
+
+  miopen::TransformOp makeTranspose(PatternRewriter &b, Value inp,
+                                    const AffineMapAttr &inMap,
+                                    const AffineMapAttr &outMap) const {
+    AffineMap inpIdxMap = inMap.getAffineMap();
+    AffineMap outpIdxMap = outMap.getAffineMap();
+    auto loc = inp.getLoc();
```
krzysz00: Same deal - `Location` here, `ArrayRef<int64_t>` on the shape and so on ... standards upstream are to only use `auto` where the type is mentioned in the thing that generates the variable or it's a rather long and unwieldy type to write down.

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemmLowering.cpp
krzysz00: Fair, let's leave it as it is, then. Unless you want to fix up TopDownTM builder

mlir/test/fusion/tosa-tp-highlevel-fused.mlir
krzysz00: Still missing a trailing newline

krzysz00 (commented):
We're definitely in "annoying minor style changes" territory, but this should be good to merge shortly

jungpark-mlir:
Thanks @krzysz00 !

krzysz00 (approved):
LGTM, thanks!

PR #632, opened 2022-05-25
Make the xdlops writeback loop stop having implicit transforms

krzysz00:
1. This PR replaces all the hard-coded math in the gridwise_gemm_v2 output loop with coordinate transformations.
2. It replaces threadwise_copy_v2 with a much simpler op that just does [vector load], [typecast], [store], without introducing a loop or coordinate transforms. The main reason this is there is to handle the typecasts and to ensure fusion has something to look for
3. Fixes up the fusion pass to the new scheme. This has the advantage of moving the fused operation **before** the writeback loop and of allowing us to load the elements corresponding to all tensor elements the gemm computed in one go.

This PR only touches xdlops in order to ensure we have something to validate against while refactoring the gemm code.

Once people are happy with this approach, I will produce a non-xdlops version, which, from what I can tell, will allow the fusion code to be substantially simplified.

The changes in this PR have passed our nightly e2e tests run manually on Lockhart with `-pv-with-gpu`

Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/543

Depends on https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/629 and  https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/630

PR #635, opened 2022-05-25
Fix up IR aesthetics

krzysz00:
1. Print a space between the "by" and the opening bracket in a
transform_map
2. Make the storeMethod attribute actually print out as either `set`
or `atomic_add` instead of as a random integer.
3. Don't generate enum attributes for enums that don't get used as
attributes.

(This is currently skip-ci not because I don't want it merged, but because we have a backlog of PRs and I also want this one to eat the merge conflict with #632 )


jerryyin (approved):
Reasonable cleanup, and LGTM

PR #645, opened 2022-06-02
[DO NOT SQUASH] XMIR (phase 2):  xmir-runner for GPU launch

sjw36:
MLIR#546: Added new func enum for AccessMode and corresponding attribute
- https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/546

MLIR#546: Apply access_modes for parameters of partitioned kernel funcs
- https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/546

MLIR#539: Add async.launch conversion to gpu.launch_func
    * converts async.await to gpu.wait
    * converts local memref.alloc's to gpu.alloc
    * insert gpu.memcpy to copy to/from gpu
-    https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/539

MLIR#488: added xmir-runner
    * lowers from XMIR execution model
    * only targets GPU (for now)
    * cleanup miopen-gen
    * fixed CPU tests
-    https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/488


external/llvm-project/mlir/include/mlir/Conversion/AsyncToGPU/AsyncToGPU.h
```diff
@@ -0,0 +1,25 @@
...
+#include <memory>
+
+namespace mlir {
+
+class ModuleOp;
+template <typename T>
+class OperationPass;
+
+/// Create a pass to convert Async operations to the GPU dialect.
+std::unique_ptr<OperationPass<ModuleOp>> createConvertAsyncToGPUPass();
```
krzysz00: I don't think that's the usual way you write these headers - basically everywhere I've looked you just return `unique_ptr<Pass>` - though I'm open to counterexamples.

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,363 @@
...
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/IR/ImplicitLocOpBuilder.h"
+#include "mlir/IR/TypeUtilities.h"
+#include "mlir/Interfaces/CallInterfaces.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/DialectConversion.h"
+#include "llvm/ADT/TypeSwitch.h"
+
+#define DEBUG_TYPE "convert-async-to-llvm"
```
krzysz00: Change here?

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,363 @@
...
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "mlir/Conversion/AsyncToGPU/AsyncToGPU.h"
+
+#include "../PassDetail.h"
+#include "mlir/Dialect/Arithmetic/IR/Arithmetic.h"
```
krzysz00: Can we try and trim the includes?

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,363 @@
...
+/// AsyncGPUTypeConverter only converts types from the Async dialect to
+/// the corresponding GPU type and does not convert any other types.
+class AsyncGPUTypeConverter : public TypeConverter {
+public:
+  AsyncGPUTypeConverter() {
+    addConversion([](Type type) { return type; });
+    addConversion(convertAsyncTypes);
+  }
+
+  static Optional<Type> convertAsyncTypes(Type type) {
```
krzysz00: Could've been a lambda that takes `TokenType`? I think the `isa<>` is already implicitly handled by the type converter infrastructure.

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,363 @@
...
+// Convert async.launch ops with 'gpu' target to gpu.launch_func ops with
+// required memory staging.
+//===----------------------------------------------------------------------===//
+
+namespace {
+class LaunchOpConversion : public OpConversionPattern<async::LaunchOp> {
+public:
+  using OpConversionPattern::OpConversionPattern;
+
+  static Optional<FuncOp> getCalledFunc(async::LaunchOp op) {
```
krzysz00: static helpers like this can probably be moved outside the class - I got bothered about that a bunch when I was getting range analysis in

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,363 @@
...
+    CallOpInterface callIf(op);
+    auto *callable = callIf.resolveCallable();
+    if (!callable)
+      return llvm::None;
+    return dyn_cast<FuncOp>(callable);
```
krzysz00: I don't think that's how you want this one.
Try 
```c++
if (auto funcOp = dyn_cast_or_null<FuncOp>(callable))
  return funcOp;
return llvm::None
```

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,363 @@
...
+  }
+
+  // Get target{gpu} attribute from called func
+  static Optional<DictionaryAttr> getGPUTarget(async::LaunchOp op) {
+    auto func = getCalledFunc(op);
```
krzysz00: Auto that's not clear from context

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,363 @@
...
+  }
+
+  static Value makeWait(OpBuilder b, Location loc, ArrayRef<Value> deps={}) {
+    auto tokenType = b.getType<gpu::AsyncTokenType>();
+    return b.create<gpu::WaitOp>(loc, tokenType, deps).asyncToken();
+  }
+
+  Value moveMemory(OpBuilder b, Value opr, uint32_t fidx,
+                   func::AccessMode accessMode,
+                   llvm::SmallVector<Value, 8> &copyBackOprs,
```
krzysz00: `SmallVectorImpl<Value> &`

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,363 @@
...
+                   llvm::SmallVector<Value, 8> &asyncDeps) const {
+    auto loc = opr.getLoc();
```
krzysz00: `Location loc`

krzysz00: (and so on for unclear auto throughout)

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,363 @@
...
+    auto gridSize = gpuAttr->get("grid_size").template cast<IntegerAttr>();
+
+    auto func = *getCalledFunc(op);
+    auto floc = func.getLoc();
+
+    // Also capture the accessMap for the params, default all to read-write
+    SmallVector<func::AccessMode, 8> accessVec(func.getNumArguments(),
+                                               func::AccessMode::ReadWrite);
+    if (auto accessMap =
+            func->template getAttrOfType<ArrayAttr>("access_map")) {
```
krzysz00: I don't think you need to say the word `template` here ... and pretty much everywhere throughout.

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,363 @@
...
+      std::transform(
```
krzysz00: `llvm::map_range`

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,363 @@
...
+          accessMap.begin(), accessMap.end(), accessVec.begin(),
+          [](Attribute a) {
+            return a.template cast<func::AccessModeAttr>().getValue();
+          });
+    }
+
+    // 2. create dummy gpu.module
```
krzysz00: I'm going to want some explanation here

The binary attribute holds the compiled code. There probably won't be compiled code at this stage?

Inside a gpu module you have a `gpu.func`, unless you're literally in the middle of the lowering/compiling process, in which case you have a `llvm.func`

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,363 @@
...
+    //    - and llvm.func (may not be necessary
+    //    gpu.module @<func_name>_module attributes {arch = "gfx908", gpu.binary
+    //    = "\7FELF\..."} {
+    //      llvm.func @<func_name> (...) attributes {block_size = 256 : i32,
+    //      gpu.kernel, grid_size = 900 : i32, rocdl.kernel}
+    //        // does any of that matter?
+
+    FunctionOpInterface funcIF(func);
+    auto funcName = funcIF.getName();
+    auto gpuModuleName = funcName + "_module";
```
krzysz00, line 168: Check for collisions, add a suffix if needed.

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
krzysz00: ... Ok, no, actual review feedback: update this comment

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,363 @@
...
+          b.create<gpu::GPUFuncOp>(floc, funcName, func.getFunctionType());
+      gpuFunc->setAttr("block_size", blockSize);
+      gpuFunc->setAttr("grid_size", gridSize);
+      gpuFunc->setAttr(gpu::GPUDialect::getKernelFuncAttrName(),
+                       b.getUnitAttr());
+
+      SymbolTable symbolTable(gpuModule);
+      symbolTable.insert(gpuFunc);
+
+      // Emit gpu convolution logic.
```
krzysz00: ... why mention convolution here

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,363 @@
...
+    auto operands = adaptor.getOperands();
+    llvm::SmallVector<Value, 8> asyncDeps;
+    llvm::SmallVector<Value, 8> gpuOperands;
+    size_t diff = operands.size() - func.getNumArguments();
+    size_t i = 0;
+    if (diff > 0) {
+      for (; i < diff; ++i)
+        asyncDeps.push_back(operands[i]);
+    } else
+      assert(diff == 0);
```
krzysz00, line 226: `assert(diff == 0 && "Message here");`

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+    SmallVector<Value, 8> copyBackOprs(func.getNumArguments(), Value());
```
krzysz00: Perhaps just `SmallVector<Value>`

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,363 @@
...
+void ConvertAsyncToGPUPass::runOnOperation() {
+  ModuleOp module = getOperation();
+  MLIRContext *ctx = module->getContext();
+
+  // Convert async dialect types and operations to LLVM dialect.
+  AsyncGPUTypeConverter converter;
+  RewritePatternSet patterns(ctx);
+
+  // Convert return operations inside async.execute regions.
+  patterns.add<LaunchOpConversion>(converter, ctx);
```
krzysz00: You can and should put all the patterns into one `add<>` call. Explanatory comments go on the passes.

mlir/tools/xmir-runner/xmir-runner.cpp
```diff
@@ -0,0 +1,102 @@
+//===- xmir-runner.cpp - MLIR ROCM Execution Driver-------------------===//
```
krzysz00: Now wait just a darn minute, why is this a separate executable instead of being part of `mlir-rocm-runner`?

krzysz00 (changes requested):
1) A whole bunch of style nits that'll get in our way when we upstream this
2) Why `xmir-runner`?

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
sjw36: It does check for collision, but re-uses the same GPU module/func when found.

mlir/tools/xmir-runner/xmir-runner.cpp
sjw36: This is a general runner (not platform specific) and when it has CPU support as well we will upstream it.

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,354 @@
...
+//===----------------------------------------------------------------------===//
+
+namespace {
+/// AsyncGPUTypeConverter only converts types from the Async dialect to
+/// the corresponding GPU type and does not convert any other types.
+class AsyncGPUTypeConverter : public TypeConverter {
+public:
+  AsyncGPUTypeConverter() {
+    addConversion([](Type type) { return type; });
+    addConversion(convertAsyncTypes);
```
krzysz00: That is, why isn't this `addConversion([](TokenType type) { return gpu::AsyncTokenType::get(type.getContext()); });`

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,354 @@
...
+  }
+};
+} // namespace
+
+//===----------------------------------------------------------------------===//
+// Convert async.launch ops with 'gpu' target to gpu.launch_func ops with
+// required memory staging.
+//===----------------------------------------------------------------------===//
+
+namespace {
```
krzysz00: To rephrase, these don't go in a `namespace {}` but do get marked static.

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
krzysz00: Unresolved?

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
krzysz00: Ah

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,354 @@
...
+    //    %15 = gpu.wait async
+    //    %16 = gpu.launch_func async [%15] @test_fusion_module::@test_fusion
+    //    blocks in (%c900, %c1, %c1) threads in (%c256, %c1, %c1)
+    //    dynamic_shared_memory_size %c0_i32 args(%4 : memref<128x32x32x8xf32>,
+    //    %9 : memref<128x3x3x8xf32>, %14 : memref<128x30x30x128xf32>)
+
+    auto tokenType = rewriter.getType<gpu::AsyncTokenType>();
+
+    auto zeroIdx =
+        rewriter.create<arith::ConstantOp>(loc, rewriter.getIndexAttr(0));
```
krzysz00: `createOrFold`, perhaps?
And sometimes it makes sense to go `Value zeroIdx = ...` to make it clearer you just want the `Value`

krzysz00 (commented):
I'm pretty sure the opinion on `xmir-runner` upstream will by "why isn't this just an alias for `mlir-cpu-runner`?" Which is a fair point, since `mlir-rocm-runner` _almost_ a wrapper around `mlir-cpu-runner` with a hardcoded pipeline.

It may make sense to define the xmir-runner as a pass pipeline

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
sjw36: I guess the rule is nameless namespace for local classes/types and static for local funcs.

sjw36:
I am looking into the pipeline idea.

sjw36:
After reviewing the PassPipelineRegistration flow, looks like a good idea but only used in one dialect 'SparseTensor'.
Furthermore, the xmir-runner is still very much in flux (especially the pipeline) so not ready for upstream at this point.

We should try this approach but in a different PR. See ticket below.

https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/549

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -0,0 +1,345 @@
...
+    }
+    asyncDeps.push_back(dstToken);
+    return dstMem;
+  }
+
+  LogicalResult matchAndRewrite(async::LaunchOp op, OpAdaptor adaptor,
+                                ConversionPatternRewriter &rw) const override {
+    Location loc = op.getLoc();
+    auto module = op->getParentOfType<ModuleOp>();
+    auto ctx = module.getContext();
```
krzysz00: At the very least, `auto* ctx`, but really `MLIRContext *ctx`

external/llvm-project/mlir/lib/Dialect/Async/IR/Async.cpp
```diff
@@ -143,7 +144,7 @@ void LaunchOp::print(OpAsmPrinter &p) {
   p << ")";
 
   // -> (return.type, ...)
-  p.printOptionalArrowTypeList(llvm::drop_begin(getResultTypes()));
+  p.printArrowTypeList(llvm::drop_begin(getResultTypes()));
```
krzysz00, line 147: Why?

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -378,6 +378,10 @@ template <typename T> struct MILARewritePattern : public OpRewritePattern<T> {
     // generic and a copy operation or transform.
     int inpUses = 0;
     for (Operation *use : inp.getUsers()) {
+      if (dyn_cast<memref::DeallocOp>(use)) {
```
krzysz00: `isa<>` over a `dyn_cast<>` that's never used ... come to think it, I'm surprised there's not a warning about that

krzysz00 (commented):
Ok, so, given the ticket about pipeline pass registration, I'm willing to allow xmir-runner on the theory that we'll get to that later.

I still have ongoing style nitpicks about excessive `auto` in code that is liable to be upstreamed ... they'll complain, so I'm complaining first.

external/llvm-project/mlir/lib/Dialect/Async/IR/Async.cpp
sjw36: FuncInterface wants this.

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
sjw36: I see plenty of auto upstream, especially when it's obvious and only used to pass thru. But fine.

external/llvm-project/mlir/lib/Dialect/Async/IR/Async.cpp
krzysz00: Fair enough

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
krzysz00: The standard I've seen (which, well, there's definitely stuff that's gone in that doesn't meet this rule, as everywhere, but I've seen reviewers upholding it and I see their point) is to use `auto` in cases only where the type is already mentioned nearby or when it's a long monstrosity that isn't really important (the exact types of iterators, for example).

But things like `Location loc = ...` seem preferred over `auto loc = ` because it means less running out to the API docs to find the type of the thing.

sjw36: I get that. But generally 'loc' is only ever passed to core methods, the details are never needed in this code.

krzysz00: But that's not exactly the point - being able to generally tell what type a thing is (or what type it's being used as) is what's useful. Sure, it's not always important - a lot of type-inference-happy languages will lean on editor support for that question - but C++ doesn't have that culture, and so spelling out types that aren't clear from context (or way too long, or both) is a reasonable request to make of developers.

And so saying `Location`, `MLIRContext *` and so on explicitly helps people work their way around a large core codebase.

sjw36: 'auto' has been part of the language for 10 years now. I think it is useful when details are not needed such as pass thru variables and avoids massive updates when type names change. 

In any case, the requested changes have been made and the tests are finally green.

sjw36:
All requested changes have been made and tests are green. Please approve asap.

krzysz00 (approved):
The main remaining complaints are
- There's a high probability someone's going to ask you to deal with the rest of the `auto`s (I marked the Location as representative) when this gets upstreamed; and
- xmir-runner doesn't need to exist (but it's fine for now since we have plans to clean it up)

I'm submitting an approval for if you really want this merged, but it might make more sense to get things into a better state now instead of "oh, we'll go back and fix it later"

Also, re "tests are finally green" - you can re-trigger [partial] CI runs if there were transient failures

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
krzysz00: On the other hand, the people we'll be sending this code to think (from what I can tell) that while `auto` can help readability in some cases (ex. iterator types, the exact types of lambdas, some iterator variables, etc.), using it all over the place hurts readability and probably makes it easier to miss a spot during refactorings.

And so the general preference for naming your types (esp. for stuff like `Location`  or `MLIRContext *`)

... speaking of, you probably want `auto*` at the least on `getContext()` so it doesn't accidentally copy the context or something.

PR #646, opened 2022-06-02
[release branch] Code object abi released

krzysz00:
As of ROCm 5.2, device library users must set the code object ABI
version, either by defining the __oclc_ABI_version variable or linking
in a device library that does so.

MLIR did not do this, this commit fixes the issue by backporting the
upstream fix.

Corresponding upstream fix https://reviews.llvm.org/D126913

### DIFF
--- a/external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp
+++ b/external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp
@@ -282,6 +282,10 @@ SerializeToHsacoPass::translateToLLVMIR(llvm::LLVMContext &llvmContext) {
                          .getZExtValue();
     uint32_t isaNumber = minor + 1000 * major;
     addControlConstant("__oclc_ISA_version", isaNumber, 32);
+
+    // This constant must always match the default code object ABI version
+    // of the AMDGPU backend.
+    addControlConstant("__oclc_ABI_version", 400, 32);
   }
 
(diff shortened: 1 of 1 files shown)


whchung (approved):
LGTM. Thanks for patching it!

krzysz00:
@sjw36 Jerry has pointed out that backports to the release branch are a pain at this point, and after some discussion I've observed that the only way that not having this patch in could be a problem is if anyone, ever, hits the math function lowering calls defined at https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/miopen-dialect/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp#L360 

With the state of the MIGraphX integration in 5.2, could any external client ever send in requests that would trigger those `math::` to `__ocml` conversions?

If not, I can close this PR as unnecessary.

sjw36:
I believe we could get math calls from fusions. See test fusion/e2e/tosa-to-miopen-rsqrt.e2e.mlir.

krzysz00:
Are there APIs in MIGraphX 5.2 that would allow a user to hit that fusion?

krzysz00:
(and confirmed that trying to fuse a sqrt() in will cause the math libraries to be linked in, and so, if we don't get this commit in, anyone trying to fuse in sqrt() will get linker errors from their fused kernels)

krzysz00:
Closing - the only way someone could hit this linker error in prod would be to run the fusion code, which isn't exposed in MIGraphX 5.2 at all.

PR #647, opened 2022-06-02
Set the ABI version constant to v5 for ROCm 5.3

krzysz00:
In the ROCm 5.3 release, the default ABI version will be v5 instead of
v4. This means that the ABI version control constant must be set to
500 instead of 400.

The change of the default version has not yet landed upstream, but I
am assured it will land in time for the ROCm 5.3 release.

Note that, while setting this constant correctly will ensure that any
users linking against ROCm 5.3 will have their code function correctly
if they, for example, use exp(), it will also mean that usage of ROCm
5.2 device libraries in conjunction with code containing this
commit (as we might do in the interval between the 5.2 and 5.3
releases) will be incorrect.

This issue won't affect anyone external, but it does mean that the
process of bumping the ROCm version in the CI will need to skip 5.2.


jerryyin:
Do we have a unit test such that if the flag doesn't exist, it fails?

krzysz00:
No, because our Docker container only tests against ROCm 5.0, where this flag doesn't matter

I don't even know how to set up a testing environment that'll let me check for this problem without stomping on a dev machine

jerryyin (approved):
Approving based on this does not matter for existing tested ROCm, and is forward looking towards a future release.

PR #649, opened 2022-06-03
MLIR#382: Remove hip dependency from mlir/lib

zhanglx13:
1. All components in mlir/lib (except the rocm-runtime-wrappers) do not depend on BackendUtils, hence they do not depend on hip.
2. IsaNameParser is renamed as IsaNameSplitter and moved to Dialect/MIOpen/utility. 
4. IsaNameSplitter is compiled into MLIRMIOpenUtility.
5. mlir/include/mlir/ExecutionEngine is empty, so it is removed.
6. hipGetDeviceProperties() is invoked directly in mlir-rocm-runner

Any code that uses IsaNameSplitter needs to
1. Add #include "mlir/Dialect/MIOpen/utility/IsaNameSplitter.h", and
2. link to MLIRMIOpenUtility


mlir/lib/Conversion/GPUToMIGraphX/CMakeLists.txt
```diff
@@ -19,5 +19,4 @@ target_link_libraries(MLIRGPUToMIGraphX
   MLIRMIGraphX
   MLIRGPUToGPURuntimeTransforms
   MLIRGPUToROCDLTransforms
-  LLVMROCmBackendUtils
```
krzysz00, line 22: Let's call it MLIRRocmBackendUtils?

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
```diff
@@ -1,1376 +0,0 @@
-//===- rocm-runtime-wrappers.cpp - MLIR ROCM runner wrapper library -------===//
```
krzysz00: ... hold on, don't we still use these functions in tests? @sjw36 

I'm surprised this removal didn't break the build

krzysz00 (commented):
The movement looks reasonable, but I have one naming nitpick and I'm surprised that the removal of the runtime wrappers went through ...

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
zhanglx13: Now the tests are using mlir_rocm_runtime from upstream according to a recent commit PR631

mlir/lib/Conversion/GPUToMIGraphX/CMakeLists.txt
zhanglx13: @jerryyin What do you think about the naming?

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
krzysz00: Not all the functions the tests need are in that file though

zhanglx13: Thank you for pointing that out. I'll try to fix it.

krzysz00: The question will be ... do the tests actually run with this file deleted?

zhanglx13: No. It turns out I was testing without removing the rocm-runtime-wrapper library.
I brought it back and will submit again soon after the testing is done.

mlir/lib/Dialect/MIOpen/Generator/CMakeLists.txt
```diff
@@ -37,3 +41,7 @@ target_link_libraries(MLIRMIOpenConv2dGenerator
   MLIRMIOpenOps
   ${LIBS}
 )
+
+add_llvm_library(MLIRRocmBackendUtils
```
krzysz00: add_llvm_library is absolutely not the function you want - that has the implication that the component is part of LLVM, not MLIR. `add_mlir_dialect_library`, maybe?

mlir/tools/mlir-rocm-runner/mlir-rocm-runner.cpp
```diff
@@ -33,9 +32,25 @@
 #include "mlir/InitAllDialects.h"
 #include "mlir/Transforms/DialectConversion.h"
 
+#include "mlir/Conversion/GPUCommon/GPUCommonPass.h"
+#include "mlir/Dialect/GPU/GPUDialect.h"
+#include "mlir/Dialect/GPU/Passes.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Pass/PassManager.h"
```
krzysz00, line 38: Can we check to make sure that these includes are all actually needed?

mlir/tools/mlir-rocm-runner/mlir-rocm-runner.cpp
```diff
@@ -61,6 +76,25 @@ static cl::opt<bool>
                cl::desc("input is in the MLIR LLVM/ROCDL dialect"),
                cl::init(false));
 
+static constexpr const char kTargetTriple[] = "amdgcn-amd-amdhsa";
+
+#if __INCLUDE_HIP__
+namespace {
+    void getGpuGCNArchName(hipDevice_t device, std::string &gcnArchName) {
```
krzysz00: `static` over anonymous namespaces for things like this, per upstream style we haven't always known about or followed well

And you'll want to `git clang-format` after staging and before committing (or `git clang-format miopen-dialect` after committing) or else your changes will get bounced back by CI for not being formatted correctly.

krzysz00 (commented):
Overall looks reasonable and it's definitely an important cleanup, just a few things here and there

mlir/lib/Dialect/MIOpen/Generator/CMakeLists.txt
zhanglx13: According to the definition of mlir_add_dialect_library, this function declares a library associated with a dialect. Should we instead use the more general function mlir_add_library for IsaNamePaser since it is more like a utility function?

mlir/tools/mlir-rocm-runner/mlir-rocm-runner.cpp
zhanglx13: Actually the GPUDialect.h and Pass.h are not needed.

mlir/tools/mlir-rocm-runner/mlir-rocm-runner.cpp
zhanglx13: Thank you @krzysz00 for the advice. I'll make sure I run it before submitting commits.

As per the discussion with @jerryyin, the next step will be to remove calling hipGetDeviceProperties() inside mlir-rocm-runner to avoid the out-of-handle problem when running multiple rocm instances concurrently. 
I'll check with the coding style to learn more about things like this.

mlir/lib/Conversion/GPUToMIGraphX/CMakeLists.txt
jerryyin: I'm no pro to naming but `backend utils` seems too broad and vague compared to what it does. The only utility in this component has nothing to do with `backend`. I'd just get rid of `backend` in the naming.

mlir/lib/Dialect/MIOpen/Generator/CMakeLists.txt
```diff
@@ -37,3 +41,7 @@ target_link_libraries(MLIRMIOpenConv2dGenerator
   MLIRMIOpenOps
   ${LIBS}
 )
+
+add_mlir_library(MLIRRocmBackendUtils
+  IsaNameParser.cpp
+)
```
jerryyin: Can we make the `IsaNameParser` to be a compilation unit of `Conv2dGenerator`?

mlir/lib/Dialect/MIOpen/Generator/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+set(LLVM_OPTIONAL_SOURCES
+  Conv2dGenerator.cpp
+  IsaNameParser.cpp
+)
```
jerryyin: I'd prefer to get rid of this (and not use `LLVM_*` in mlir CmakeLists)

mlir/tools/mlir-rocm-runner/CMakeLists.txt
```diff
@@ -4,14 +4,48 @@ set(LLVM_OPTIONAL_SOURCES
...
+    PATHS "${HIP_ROOT_DIR}/lib")
+  if (NOT ROCM_RUNTIME_LIBRARY)
+    message(SEND_ERROR "Could not locate ROCm HIP runtime library")
+  else()
+    message(STATUS "ROCm HIP runtime lib: ${ROCM_RUNTIME_LIBRARY}")
+  endif()
+
+  # Set HIP compile-time flags.
+  add_definitions(-D__HIP_PLATFORM_HCC__)
+  add_definitions(-D__INCLUDE_HIP__=1)
```
jerryyin: We probably can get rid of these.

mlir/tools/mlir-rocm-runner/mlir-rocm-runner.cpp
```diff
@@ -61,6 +74,28 @@ static cl::opt<bool>
                cl::desc("input is in the MLIR LLVM/ROCDL dialect"),
                cl::init(false));
 
+static constexpr const char kTargetTriple[] = "amdgcn-amd-amdhsa";
+
+#if __INCLUDE_HIP__
```
jerryyin: ditto here, we can probably get rid of the macro together with (#if, #else...) here.

mlir/include/mlir/ExecutionEngine/ROCm/BackendUitls.h
```diff
@@ -1,46 +0,0 @@
-//===- BackendUtils.h - MLIR to C++ option parsing ---------------===//
```
jerryyin, line 1: Just to make sure, the entire `ROCm` sub-directory is gotten rid of, correct?

jerryyin (commented):
This cleanup makes all the entangled dependencies nicer and cleaner, thanks for your work.

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
jerryyin: Potentially I'd like to see us not having to maintain an additional "rocm-runtime-wrapper` (versus the upstream duplicate). I think what can be done (after this PR is merged) is:
 - file an upstream PR to add in whatever is necessary there
 - file another PR to get rid of `rocm-runtime-wrapper` duplicate in our side

krzysz00: I agree, but that's non-trivial in that `rocm-runtime-wrapper` here has the convolution validation code.

We'd need to do the thing where the convolution code gets moved into what I'm thinking of as the "mlir-miopen-test-lib.mlir" - a bunch of static MLIR functions that we can copy into tests. It's something I've been wanting to do for a while but haven't gotten around to.

(This also relates to @sjw36's plan to move more of the miopen-gen generated code to `gpu.*` ops ... which is vaguely in progress and I think requires a `gpu.free` op being added upstream)

mlir/lib/Dialect/MIOpen/Generator/CMakeLists.txt
krzysz00: This is a library related to the `MIOpen` dialect, given which folder it's in

mlir/tools/mlir-rocm-runner/mlir-rocm-runner.cpp
krzysz00: If you could drop them?

mlir/tools/mlir-rocm-runner/mlir-rocm-runner.cpp
krzysz00: There's no need to dive deeply into coding style (except for the bits that can't be auto-checked - there's an LLVM style guide we probably all collectively want to review, myself included) - `clang-format` is good at its job

mlir/lib/Dialect/MIOpen/Generator/CMakeLists.txt
krzysz00: `LLVM_OPTIONAL_SOURCES` is a pretty widely-used thing, IIRC

But I don't think we'll need id if we're using `add_mlir_dialect_library` calls

mlir/tools/mlir-rocm-runner/CMakeLists.txt
krzysz00: ... we do need to define HCC when including HIP

mlir/include/mlir/ExecutionEngine/ROCm/BackendUitls.h
krzysz00: Not anymore, I think

mlir/lib/Dialect/MIOpen/Generator/CMakeLists.txt
krzysz00: Overall, we could just add the library to the `Conv2DGenerator`. Or move it ta `utils/IsaNameParses.cpp` and add it to the MIOpen utils library (next to `loweringUtils`)

mlir/tools/mlir-rocm-runner/mlir-rocm-runner.cpp
```diff
@@ -33,9 +32,23 @@
...
+#include "mlir/Dialect/GPU/Passes.h"
+#include "mlir/Pass/PassManager.h"
+#include "mlir/Support/FileUtilities.h"
+#include "mlir/Transforms/Passes.h"
+#include "mlir/Dialect/MIOpen/Generator/IsaNameParser.h"
+
 #include <cstdlib>
 #include <mutex>
 
+#if __INCLUDE_HIP__
```
krzysz00: We can probably drop the `#if __INCLUDE_HIP_` in that this file won't be compiled if HIP isn't a dependency.

krzysz00 (commented):
Ok, so, we're down to bikeshedding about names. Which is important, but, still, fundamentally, good work!

mlir/tools/mlir-rocm-runner/mlir-rocm-runner.cpp
zhanglx13: Yes, I did.

mlir/lib/ExecutionEngine/ROCm/rocm-runtime-wrappers.cpp
jerryyin: @krzysz00 Good point. I just forgot about the fact that validation related code are in here as well. 

So let me rephrase, in the upcoming refactor to decouple these two:
 - all `mgpu.*` function moved (added) to external repo
 - all `mcpu.*` function and its utils remain (and potentially, we can put this to a new runtime library `rocm-validation-wrapper`)

Just so that we are clear on the boundary of the two components.

mlir/tools/mlir-rocm-runner/CMakeLists.txt
jerryyin: We still do now? `HCC` should be a deprecated old flag and we should be using `hip-clang` now. But who knows.. if we still need it then leave it as-is.

On the other hand, get rid of `__INCLUDE_HIP__` because that's defined by us and is no longer necessary.

mlir/lib/Conversion/GPUToMIGraphX/CMakeLists.txt
jerryyin: Just as FYI, we talked a second ago and sounds like a better suited name is: `IsaNameSplitter`. This is as close as it can get in what this component does.

mlir/tools/mlir-rocm-runner/CMakeLists.txt
krzysz00: We never use `hipe-clang` anywhere in the build.
So we have to specify we're targetting HCC with the flag

jerryyin: Just for clarification, by default ROCM ship with `hip-clang`, not hcc, so I'm surprised we still need the "hcc" flag.

zhanglx13: In hip_runtime.h (rocm 5.0.1), it requires the definition of __HIP_PLATFORM_AMD/HCC__ or __HIP_PLATFORM_NVIDIA/NVCC__.  Later it does say that __HCC__ and __HIP_PLATFORM_HCC__ are deprecated.

jerryyin: @zhanglx13 Sounds good, and thanks for looking into this. Please go ahead with `__HIP_PLATFORM_AMD__`.

jerryyin (approved):
LGTM

krzysz00 (approved):
LGTM!

zhanglx13:
@jerryyin @krzysz00 There is a connection reset error when building MIOpen. Should we rerun the check or there is something wrong on the server?

jerryyin:
@zhanglx13 That's a fairly typical MIOpen dependency (boost) failure, you can re-trigger it.

zhanglx13:
@jerryyin @krzysz00 Now all tests have passed. Could one of you merge the pull request? Or should I do it?

jerryyin:
@zhanglx13 Feel free to do it if:
 - All tests are passing
 - At least 1 reviewer approved
 - No foreseeable breakage to nightly tests

PR #650, opened 2022-06-03
Fix up shared library arguments in performance tests and parameter sw…

krzysz00:
…eeps

The recent change to use more of upstream's copy of the runtime failed to update the performance test script.


sjw36 (approved):
Looks good.

PR #651, opened 2022-06-03
Generalise --tosa-partition to allow other specified ops as primary op [DO NOT SQUASH]

pcf000:
[in external]
Allow tosa.transpose and tosa.reshape to be gathered along with the element-wise ops.
Allow specifying the name of the "kernel" attribute.
Test the sub-options primary-ops, trailing-only, and attribute-name.
Test both command-line and programmatic interfaces.
Also include constant operand of TransposeOp in outlined kernels.
A couple of details from use within torch-mlir.
Initial implementation of partition-config.
Options in Passes.td, create config at runtime from options if any are set.
Change conv/second/front terminology to anchor/trailing/leading.

[in our mlir]
Fix MLIR_INCLUDE_TESTS usage for our tools.
Fix a couple of compiler warnings.

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.h
+++ b/external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.h
@@ -31,9 +31,9 @@ std::unique_ptr<Pass> createTosaInferShapesPass();
 std::unique_ptr<Pass> createTosaMakeBroadcastablePass();
 std::unique_ptr<Pass> createTosaTestQuantUtilAPIPass();
 std::unique_ptr<Pass> createTosaPartitionPass();
-std::unique_ptr<Pass> createPostPartitionCollapsePass();
+class PartitionConfig;
+std::unique_ptr<Pass> createTosaPartitionPass(PartitionConfig *config);
 std::unique_ptr<Pass> createTosaOptionalDecompositions();
-void registerTosaPartitionPipelinePass();
 
 #define GEN_PASS_REGISTRATION
 #include "mlir/Dialect/Tosa/Transforms/Passes.h.inc"
--- a/external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
@@ -59,7 +59,7 @@ def TosaPartition : Pass<"tosa-partition", "ModuleOp"> {
     Option<"attributeName", "attribute-name", "std::string",
            /*default=*/"\"kernel\"", "Attribute for outlined functions">,
     Option<"trailingOnly", "trailing-only", "bool", /*default=*/"false",
-           "Don't gather ops ahead of Conv2D">
+           "Don't gather ops ahead of anchor op">
   ];
   let dependentDialects = ["tosa::TosaDialect"];
 }
--- a/external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
@@ -52,7 +52,13 @@ def TosaPartition : Pass<"tosa-partition", "ModuleOp"> {
   let constructor = "createTosaPartitionPass()";
 
   let options = [
-    Option<"nofront", "nofront", "bool", /*default=*/"false",
+    ListOption<"anchorOps", "anchor-ops", "std::string",
+               "One or more operations to be used as focus of partitioned "
+               "kernels",
+               "llvm::cl::ZeroOrMore">,
+    Option<"attributeName", "attribute-name", "std::string",
+           /*default=*/"\"kernel\"", "Attribute for outlined functions">,
+    Option<"trailingOnly", "trailing-only", "bool", /*default=*/"false",
            "Don't gather ops ahead of Conv2D">
   ];
   let dependentDialects = ["tosa::TosaDialect"];
--- a/external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
+++ b/external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
@@ -499,7 +499,7 @@ public:
       if (anchorOps.hasValue() || attributeName.hasValue() ||
           trailingOnly.hasValue()) {
         if (anchorOps.empty()) // ListOption doesn't have a default value.
-          anchorOps = {"tosa.conv2d"};
+          anchorOps = {"tosa.conv2d","tosa.matmul","tosa.depthwise_conv2d"};
         config = new mlir::tosa::PartitionConfigFromOptions(
             anchorOps, attributeName, trailingOnly);
       } else {
@@ -680,9 +680,6 @@ public:
       class AttributeOnePartitionConfig
           : public mlir::tosa::SimpleDefaultPartitionConfig {
       public:
-        bool isAnchorOp(Operation *op) override {
-          return isa<tosa::Conv2DOp>(op);
-        }
         std::string attributeName() override { return "one"; }
       };
       pm.addPass(
--- a/external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
+++ b/external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
@@ -11,6 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "mlir/Analysis/SliceAnalysis.h"
+#include "mlir/Dialect/Arithmetic/IR/Arithmetic.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/Tosa/IR/TosaOps.h"
 #include "mlir/Dialect/Tosa/Transforms/PassDetail.h"
@@ -88,26 +89,53 @@ bool isElementwiseOp(Operation *op) {
   // clang-format on
 }
 
+bool isFusibleOp(Operation *op) {
+  return isElementwiseOp(op) || isa<tosa::TransposeOp, tosa::ReshapeOp>(op);
+}
+
+bool isZeroAttribute(Attribute value) {
+  if (auto intValue = value.dyn_cast<IntegerAttr>())
+    return intValue.getValue().isNullValue();
+  if (auto fpValue = value.dyn_cast<FloatAttr>())
+    return fpValue.getValue().isZero();
+  if (auto splatValue = value.dyn_cast<SplatElementsAttr>())
+    return isZeroAttribute(splatValue.getSplatValue<Attribute>());
+  if (auto elementsValue = value.dyn_cast<ElementsAttr>())
+    return llvm::all_of(elementsValue.getValues<Attribute>(), isZeroAttribute);
+  if (auto arrayValue = value.dyn_cast<ArrayAttr>())
+    return llvm::all_of(arrayValue.getValue(), isZeroAttribute);
+  return false;
+}
+
+bool isConstantZero(Operation *op) {
+  // test for zero
+  if (auto cst = dyn_cast<arith::ConstantOp>(op)) {
+    return isZeroAttribute(cst.getValue());
+  }
+  return false;
+}
+
 ////////////////////////////////////////////////////////////////////////////////
 
 // Inspired by / adapted from outlineIfOp() in SCF/Transforms/Utils.cpp
 // and mergeIdenticalBlocks() in Utils/RegionUtils.cpp.
 
 struct OutliningCandidate {
-  OutliningCandidate(Operation *_convOp, ArrayRef<Operation *> &_secondOps,
-                     ArrayRef<Operation *> &_frontOps, ArrayRef<Value> &_params,
-                     ArrayRef<Value> &_returnVals, StringRef _partFnName);
+  OutliningCandidate(Operation *anchorOp_, ArrayRef<Operation *> &trailingOps_,
+                     ArrayRef<Operation *> &leadingOps_,
+                     ArrayRef<Value> &params_, ArrayRef<Value> &returnVals_,
+                     StringRef partFnName_);
 
   unsigned addOp(Operation *op, unsigned orderIt);
 
-  Operation *convOp;
-  SmallVector<Operation *> secondOps;
-  SmallVector<Operation *> frontOps;
+  Operation *anchorOp;
+  SmallVector<Operation *> trailingOps;
+  SmallVector<Operation *> leadingOps;
   SmallVector<Value> params;
   SmallVector<Value> returnVals;
   std::string partFnName;
   llvm::hash_code hash;
-  FuncOp function;
+  func::FuncOp function;
 
   /// Return the order index for the given value that is within the block of
   /// this data.
@@ -134,19 +162,19 @@ unsigned OutliningCandidate::addOp(Operation *op, unsigned orderIt) {
   return orderIt;
 }
 
-OutliningCandidate::OutliningCandidate(Operation *convOp_,
-                                       ArrayRef<Operation *> &secondOps_,
-                                       ArrayRef<Operation *> &frontOps_,
+OutliningCandidate::OutliningCandidate(Operation *anchorOp_,
+                                       ArrayRef<Operation *> &trailingOps_,
+                                       ArrayRef<Operation *> &leadingOps_,
                                        ArrayRef<Value> &params_,
                                        ArrayRef<Value> &returnVals_,
                                        StringRef partFnName_)
-    : convOp(convOp_), partFnName(partFnName_), hash(0), function(nullptr) {
+    : anchorOp(anchorOp_), partFnName(partFnName_), hash(0), function(nullptr) {
   // We'll need to grab the cloned ops to avoid use-after-free.
-  for (auto *op : secondOps_) {
-    secondOps.push_back(op);
+  for (auto *op : trailingOps_) {
+    trailingOps.push_back(op);
   }
-  for (auto *op : frontOps_) {
-    frontOps.push_back(op);
+  for (auto *op : leadingOps_) {
+    leadingOps.push_back(op);
   }
   for (auto val : params_) {
     params.push_back(val);
@@ -156,11 +184,11 @@ OutliningCandidate::OutliningCandidate(Operation *convOp_,
   }
 
   unsigned orderIt = params.size();
-  for (auto *op : frontOps) {
+  for (auto *op : leadingOps) {
     orderIt = addOp(op, orderIt);
   }
-  orderIt = addOp(convOp, orderIt);
-  for (auto *op : secondOps) {
+  orderIt = addOp(anchorOp, orderIt);
+  for (auto *op : trailingOps) {
     orderIt = addOp(op, orderIt);
   }
 }
@@ -235,15 +263,15 @@ bool outliningCandidatesEquivalent(OutliningCandidate &one,
     }
   }
 
-  for (auto ops : llvm::zip(one.frontOps, two.frontOps)) {
+  for (auto ops : llvm::zip(one.leadingOps, two.leadingOps)) {
     if (!opsMatch(std::get<0>(ops), std::get<1>(ops), one, two)) {
       return false;
     }
   }
-  if (!opsMatch(one.convOp, two.convOp, one, two)) {
+  if (!opsMatch(one.anchorOp, two.anchorOp, one, two)) {
     return false;
   }
-  for (auto ops : llvm::zip(one.secondOps, two.secondOps)) {
+  for (auto ops : llvm::zip(one.trailingOps, two.trailingOps)) {
     if (!opsMatch(std::get<0>(ops), std::get<1>(ops), one, two)) {
       return false;
     }
@@ -264,19 +292,20 @@ findOutliningCandidate(OutliningCandidate &newCandidate,
 
 // Given a convolution op and its fuse-able trailing (second) and leading
 // (front) ops, remove them into a separate function.
-void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
-                        ArrayRef<Operation *> frontOps, ArrayRef<Value> params,
-                        ArrayRef<Value> returnVals, StringRef partFnName,
+void outlineConvPartOps(Operation *anchorOp, ArrayRef<Operation *> trailingOps,
+                        ArrayRef<Operation *> leadingOps,
+                        ArrayRef<Value> params, ArrayRef<Value> returnVals,
+                        StringRef partFnName, StringRef attrName,
                         std::vector<OutliningCandidate> &candidates) {
   ValueRange values(params);
-  OpBuilder b(convOp);
-  Location loc = convOp->getLoc();
-  FuncOp outlinedFunc;
+  OpBuilder b(anchorOp);
+  Location loc = anchorOp->getLoc();
+  func::FuncOp outlinedFunc;
 
   // ------------------------------------------------------------
   // Merging part.
 
-  OutliningCandidate newCandidate(convOp, secondOps, frontOps, params,
+  OutliningCandidate newCandidate(anchorOp, trailingOps, leadingOps, params,
                                   returnVals, partFnName);
 
   if (OutliningCandidate *found =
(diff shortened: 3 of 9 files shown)


mlir/tools/miopen-gen/CMakeLists.txt
```diff
@@ -5,9 +5,16 @@ set(LLVM_LINK_COMPONENTS
 get_property(dialect_libs GLOBAL PROPERTY MLIR_DIALECT_LIBS)
 get_property(conversion_libs GLOBAL PROPERTY MLIR_CONVERSION_LIBS)
 
+if(MLIR_INCLUDE_TESTS)
```
pcf000, line 8: This set of changes, with the one for mlir-miopen-driver, fixes compiling with -DMLIR_INCLUDE_TESTS=OFF. It's the same pattern used in external/llvm-project/mlir/tools/*/CMakeLists.txt.

The default is ON, but I was trying to make sure my tests didn't cause trouble in the other case. As it happens, the whole test directory is omitted when OFF, so my concern was a non-problem, but I found this build failure.

external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.h
```diff
@@ -31,9 +31,9 @@ std::unique_ptr<Pass> createTosaInferShapesPass();
 std::unique_ptr<Pass> createTosaMakeBroadcastablePass();
 std::unique_ptr<Pass> createTosaTestQuantUtilAPIPass();
 std::unique_ptr<Pass> createTosaPartitionPass();
-std::unique_ptr<Pass> createPostPartitionCollapsePass();
+class PartitionConfig;
```
pcf000, line 34: Remove the obsolete declarations for passes we no longer have, and add a second variant of createTosaPartitionPass() to allow specifying the partitioning configuration at the program level. Typical command-line usage will call the no-arguments version and interpret the command-line options; calling within a program can supply the configuration as an argument.  PartitionConfig is forward-declared because that's the abstract interface for users that write their own.

pcf000, line 34: See the test file for usage examples.

external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
```diff
@@ -52,7 +52,13 @@ def TosaPartition : Pass<"tosa-partition", "ModuleOp"> {
   let constructor = "createTosaPartitionPass()";
 
   let options = [
-    Option<"nofront", "nofront", "bool", /*default=*/"false",
+    ListOption<"anchorOps", "anchor-ops", "std::string",
```
pcf000, line 55: See the test file for usage examples.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -88,26 +89,53 @@ bool isElementwiseOp(Operation *op) {
   // clang-format on
 }
 
+bool isFusibleOp(Operation *op) {
```
pcf000: Separating "element-wise" from "other ops we want to fuse."  And "fusible" is the dictionary spelling, though I keep thinking of "fusible links."

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -88,26 +89,53 @@ bool isElementwiseOp(Operation *op) {
...
+  return isElementwiseOp(op) || isa<tosa::TransposeOp, tosa::ReshapeOp>(op);
+}
+
+bool isZeroAttribute(Attribute value) {
```
pcf000, line 99: Moved out because they don't need to be member functions.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -88,26 +89,53 @@ bool isElementwiseOp(Operation *op) {
...
 ////////////////////////////////////////////////////////////////////////////////
 
 // Inspired by / adapted from outlineIfOp() in SCF/Transforms/Utils.cpp
 // and mergeIdenticalBlocks() in Utils/RegionUtils.cpp.
 
 struct OutliningCandidate {
-  OutliningCandidate(Operation *_convOp, ArrayRef<Operation *> &_secondOps,
-                     ArrayRef<Operation *> &_frontOps, ArrayRef<Value> &_params,
-                     ArrayRef<Value> &_returnVals, StringRef _partFnName);
+  OutliningCandidate(Operation *anchorOp_, ArrayRef<Operation *> &trailingOps_,
```
pcf000, line 127: A whole bunch of nomenclature changes start here:
  secondOps -> trailingOps
  frontOps -> leadingOps
  convOp -> anchorOp

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -264,19 +292,20 @@ findOutliningCandidate(OutliningCandidate &newCandidate,
 
 // Given a convolution op and its fuse-able trailing (second) and leading
 // (front) ops, remove them into a separate function.
-void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
-                        ArrayRef<Operation *> frontOps, ArrayRef<Value> params,
-                        ArrayRef<Value> returnVals, StringRef partFnName,
+void outlineConvPartOps(Operation *anchorOp, ArrayRef<Operation *> trailingOps,
+                        ArrayRef<Operation *> leadingOps,
+                        ArrayRef<Value> params, ArrayRef<Value> returnVals,
+                        StringRef partFnName, StringRef attrName,
```
pcf000, line 301: Kernel attribute name is an option now.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -360,53 +391,95 @@ void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
...
     }
   }
 }
 
+} // namespace
+
+namespace mlir {
+namespace tosa {
+
+class PartitionConfig {
```
pcf000, line 416: New stuff.  A PartitionConfig has three predicates to define anchor, leading, and trailing ops, and a string attribute name.  No explicit "trailing-only" at this level;  instead, isLeadingOp() will be made to return false always.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -360,53 +391,95 @@ void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
...
+public:
+  virtual bool isAnchorOp(Operation *) = 0;
+  virtual bool isLeadingOp(Operation *) = 0;
+  virtual bool isTrailingOp(Operation *) = 0;
+  virtual std::string attributeName() = 0;
+  virtual ~PartitionConfig() = default;
+};
+
+class SimpleDefaultPartitionConfig : public PartitionConfig {
```
pcf000, line 425: The simple case we've used most recently:  Conv2DOps, "kernel," and the isFusibleOp() predicate in both directions.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -360,53 +391,95 @@ void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
...
+public:
+  bool isAnchorOp(Operation *op) override { return isa<tosa::Conv2DOp>(op); }
+  bool isLeadingOp(Operation *op) override {
+    return isConstantZero(op) || isFusibleOp(op);
+  }
+  bool isTrailingOp(Operation *op) override { return isFusibleOp(op); }
+  std::string attributeName() override { return "kernel"; }
+};
+
+class PartitionConfigFromOptions : public PartitionConfig {
```
pcf000, line 437: A configuration based on the command-line sub-options (defined in Passes.td).  The creator of the configuration object will look at the Option objects and pass their values in to the constructor, which saves them in the object.  Any defaulting happens in the caller.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -360,53 +391,95 @@ void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
...
-      return fpValue.getValue().isZero();
-    if (auto splatValue = value.dyn_cast<SplatElementsAttr>())
-      return isZeroAttribute(splatValue.getSplatValue<Attribute>());
-    if (auto elementsValue = value.dyn_cast<ElementsAttr>())
-      return llvm::all_of(elementsValue.getValues<Attribute>(),
-                          isZeroAttribute);
-    if (auto arrayValue = value.dyn_cast<ArrayAttr>())
-      return llvm::all_of(arrayValue.getValue(), isZeroAttribute);
-    return false;
+  mlir::tosa::PartitionConfig *config = nullptr;
```
pcf000, line 465: Default the config in the pass to nullptr because we can't create it in the pass constructor, because the options aren't parsed yet.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -360,53 +391,95 @@ void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
...
+
+  // Special case:  TransposeOp's second operand must be a
+  // constant, which means we must include it too if we include
+  // the TransposeOp.  "ops" here may be either leadingOps or trailingOps.
+  void specialCaseForTranspose(Operation *op, SetVector<Operation *> &ops) {
```
pcf000, line 470: Compilation will barf if the operand isn't a constant op.  This is the operand that specifies how it transposes, and I guess it encodes into the instruction later.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -360,53 +391,95 @@ void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
...
-    }
-    return false;
-  }
+public:
+  TosaPartitionPass() = default;
+  TosaPartitionPass(mlir::tosa::PartitionConfig *config) : config(config) {}
+  ~TosaPartitionPass() override { delete config; }
 
-  void traceInputs(Operation *op, SmallVector<Operation *> &predecessors,
+  void traceInputs(Operation *op, SetVector<Operation *> &predecessors,
```
pcf000, line 480: Made the parameter a SetVector to unify the interface for specialCaseForTranspose().

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -418,18 +491,35 @@ class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
   }
 
   void runOnOperation() override {
+    // Must set config here, because at pass-construction time the options
+    // haven't been parsed yet.
+    if (!config) {
+      if (anchorOps.hasValue() || attributeName.hasValue() ||
```
pcf000, line 502: If *any* sub-option is set, go this way.  Option defaults make sure it behaves rationally.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -418,18 +491,35 @@ class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
...
-      if (func->hasAttr("kernel"))
+      if (func->hasAttr(config->attributeName()))
         continue;
 
       int count = 0;
       // (Problems with node mismatches and unexpected uses if we have the
       // candidates list at module level.)
       std::vector<OutliningCandidate> candidates;
-      auto callback = [&](tosa::Conv2DOp convOp) {
+      auto callback = [&](Operation *op) {
```
pcf000, line 524: Can't specify the anchor op at compile time, alas.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -443,35 +533,29 @@ class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
...
         // Grab a useful set of leading ops, like we do for trailing.
-        // Let's limit it to only first arguments, with single uses.
-        SmallVector<Operation *> frontOps;
-        if (!nofront) {
-          traceInputs(convOp, frontOps, inputNodes);
-        } else {
-          inputNodes.insert(convOp->getOperands().begin(),
-                            convOp->getOperands().end());
-        }
+        SetVector<Operation *> leadingOps;
```
pcf000, line 556: !nofront absorbed into isLeadingOp().

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -488,40 +572,40 @@ class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
...
               }
             }
           }
         }
 
         // Make the outlined function from the ops we've gathered.
-        outlineConvPartOps(convOp, secondOps.getArrayRef(), frontOps,
-                           inputNodes.getArrayRef(), resultNodes.getArrayRef(),
+        outlineConvPartOps(anchorOp, trailingOps.getArrayRef(),
+                           leadingOps.getArrayRef(), inputNodes.getArrayRef(),
```
pcf000, line 614: Added .getArrayRef() when leadingOps became a SetVector.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -539,3 +623,92 @@ class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
 std::unique_ptr<Pass> mlir::tosa::createTosaPartitionPass() {
   return std::make_unique<TosaPartitionPass>();
 }
+
+std::unique_ptr<Pass>
+mlir::tosa::createTosaPartitionPass(mlir::tosa::PartitionConfig *config) {
```
pcf000, line 637: New create-function with the configuration parameter.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -539,3 +623,92 @@ class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
...
+  return std::make_unique<TosaPartitionPass>(config);
+}
+
+namespace {
+
+class TestTosaPartitionOptionsPass
```
pcf000, line 643: New options-test pass, to be able to test partition configurations specified in the program instead of on the command line.  It defines some arbitrary sub-options that lead to setting up various configurations, using the same test file that the command-line-option testing does.  Adapted from somebody else's test pass.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -539,3 +623,92 @@ class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
...
+        }
+      };
+      pm.addPass(
+          tosa::createTosaPartitionPass(new DepthwiseOnlyPartitionConfig()));
+    } else if (both) {
+      class DepthwiseAlsoPartitionConfig
+          : public mlir::tosa::SimpleDefaultPartitionConfig {
+      public:
+        bool isAnchorOp(Operation *op) override {
+          return isa<tosa::Conv2DOp, tosa::DepthwiseConv2DOp>(op);
```
pcf000: Two anchors.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -539,3 +623,92 @@ class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
...
+          : public mlir::tosa::SimpleDefaultPartitionConfig {
+      public:
+        std::string attributeName() override { return "one"; }
+      };
+      pm.addPass(
+          tosa::createTosaPartitionPass(new AttributeOnePartitionConfig()));
+    } else if (nofrontArg) {
+      // Another way is to pass the values to PartitionConfigFromOptions.
+      mlir::tosa::PartitionConfig *config =
+          new mlir::tosa::PartitionConfigFromOptions({"tosa.depthwise_conv2d"},
```
pcf000, line 697: Or, use this constructor and specify all the arguments.

external/llvm-project/mlir/test/Dialect/Tosa/tosa-partition-options.mlir
```diff
@@ -0,0 +1,72 @@
+// RUN: mlir-opt --tosa-partition %s | FileCheck %s
```
pcf000, line 1: A test file I happened to have laying around.  One batch of tests using command line, one using the API, same set of outputs.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
sjw36: Please move this to a header file: mlir/Dialect/Tosa/Transforms/TosaPartition.h
so other users can derive from this class for customized partitioning.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
sjw36: Would be good to have this as a default also available in the header. Users could use directly use the Simple or derive from it. Maybe rename to just DefaultPartitionConfig also.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -360,53 +391,95 @@ void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
...
+  virtual bool isAnchorOp(Operation *) = 0;
+  virtual bool isLeadingOp(Operation *) = 0;
+  virtual bool isTrailingOp(Operation *) = 0;
+  virtual std::string attributeName() = 0;
+  virtual ~PartitionConfig() = default;
+};
+
+class SimpleDefaultPartitionConfig : public PartitionConfig {
+public:
+  bool isAnchorOp(Operation *op) override { return isa<tosa::Conv2DOp>(op); }
```
sjw36: Please add tosa::MatMulOp and tosa::DepthwiseConv2dOp.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
sjw36: If it is required, you can pass as a const reference.

sjw36 (changes requested):
Looking good so far. Just need to have a header for source users to derive Configs.

external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.h
```diff
@@ -31,9 +31,9 @@ std::unique_ptr<Pass> createTosaInferShapesPass();
...
+std::unique_ptr<Pass> createTosaPartitionPass(PartitionConfig *config);
```
krzysz00, line 35: Why's this by-pointer?

external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
```diff
@@ -52,7 +52,13 @@ def TosaPartition : Pass<"tosa-partition", "ModuleOp"> {
...
+               "One or more operations to be used as focus of partitioned "
+               "kernels",
+               "llvm::cl::ZeroOrMore">,
+    Option<"attributeName", "attribute-name", "std::string",
+           /*default=*/"\"kernel\"", "Attribute for outlined functions">,
+    Option<"trailingOnly", "trailing-only", "bool", /*default=*/"false",
            "Don't gather ops ahead of Conv2D">
```
krzysz00: Drop reference to Conv2D here?

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
krzysz00: We could do some good-old fashioned programmer spelling and say fusable anyway

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
krzysz00: In which case they get a `static`

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -88,26 +89,53 @@ bool isElementwiseOp(Operation *op) {
...
+  if (auto elementsValue = value.dyn_cast<ElementsAttr>())
+    return llvm::all_of(elementsValue.getValues<Attribute>(), isZeroAttribute);
+  if (auto arrayValue = value.dyn_cast<ArrayAttr>())
+    return llvm::all_of(arrayValue.getValue(), isZeroAttribute);
+  return false;
+}
+
+bool isConstantZero(Operation *op) {
+  // test for zero
+  if (auto cst = dyn_cast<arith::ConstantOp>(op)) {
```
krzysz00, line 115: We could pull out the ConstantLike interface here?

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
krzysz00: And then add an function that creates and passes in the default config

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
krzysz00: We could default it to a dummy object or something?

krzysz00: Alternatively, have all the internal functions pass around a `const PartitionConfig &` and then `runOnOperation()` creates one.

That is, we don't store this config into the pass, but pass it around as the partitioning is being performed.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -173,7 +173,7 @@ class BroadcastConverter final
       SmallVector<Attribute, 5> newShapeAttr;
 
       // align the dimensions - by the given axis
-      for (int i = 0; i < outRank; i++) {
+      for (uint i = 0; i < outRank; i++) {
```
krzysz00: I'm pretty sure we usually spell that `unsigned` upstream (or perhaps `unsigned int`)

krzysz00 (commented):
This all looks like a nice generalization, though I'm not sure the `PartitionConfig` is set up the right way ... Simon also noticed some of what I was complaining about.

krzysz00:
Also, as a note, `Optional<PassConfig&>` might be better than a pointer.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
pcf000: As a kid, I competed in spelling bees.  "Fusable" would make me twitch.  Though I did consider "fuseable."

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
pcf000: Yes.  I just saw the anonymous-namespace part of the coding-standards document.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
pcf000: Yes.  I realised I'd forgotten that part of your original requests.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
krzysz00: Entirely fair. I'd argue for "fuseable", as it doesn't send people running to a dictionary for a form that's not super-recognized.

(and it's neat to learn about the spelling bee thing!)

external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.h
pcf000: Maybe I have a blind spot, but I'm not aware that you can make virtual calls on things that aren't pointers.

external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
pcf000: Yes, should become "anchor op."

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
pcf000: It's been a while since I wrote this, but I think it didn't quite fit.  But I'll reexamine.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
pcf000: Old habits.  I'll check.

external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.h
krzysz00: Yeah, you can make virtual calls on references (which are a very funky way to spell non-null pointers, except when they aren't)

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
pcf000: Should be uint32_t, to match the definition of outRank.

pcf000:
Superseded by #664 .

PR #652, opened 2022-06-07
Reduce number of parameter sweep workers to [# CPUs] / 2

krzysz00:
Also, in case it helps fix the non-xdlops failures, set the heartbeat
property within the parameter sweep job.


pcf000:
Looks okay from here.

pcf000 (approved):

PR #653, opened 2022-06-08
[DO NOT SQUASH] Remove downstream rocm-runtime-wrappers

zhanglx13:
1.  Merged all `mgpu*` functions to upstream: `RocmRuntimeWrappers.cpp`
    - Removed `hipCtxPopCurrent` and `hipCtxPushCurrent` since they are deprecated
    - Wrapped hip API calls with error checking macro
2. Moved the remaining `mcpu*` functions and their helper functions to `mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp` and compiled it as `libconv-validation-wrappers`.
3. Changed CMakeLists and lit configuration variables that are related to the validation library to reflect the fact that it does not depend on Rocm.
4. Changed all tests to use the new name of the library


external/llvm-project/mlir/lib/ExecutionEngine/RocmRuntimeWrappers.cpp
```diff
@@ -155,6 +154,28 @@ extern "C" void mgpuMemHostRegister(void *ptr, uint64_t sizeBytes) {
   HIP_REPORT_IF_ERROR(hipHostRegister(ptr, sizeBytes, /*flags=*/0));
 }
 
+// Allows to register a MemRef with the ROCM runtime. Initializes array with
+// value. Helpful until we have transfer functions implemented.
+template <typename T>
+void mgpuMemHostRegisterMemRef(T *pointer, llvm::ArrayRef<int64_t> sizes,
```
krzysz00: This is probably a general nitpick, but doesn't this function do too much? It's both a allocation and a memset

Where are we using this?

mlir/lib/ExecutionEngine/CMakeLists.txt
```diff
@@ -1,2 +1,13 @@
-# Placeholder to keep the original directory structure.
-add_subdirectory(ROCm)
+add_llvm_library(conv-validation-wrappers SHARED
```
krzysz00: This still isn't `add_llvm_library`. `add_mlir_library` at best

krzysz00 (commented):
I think this PR may be premature and will impose merge-conflict burdens. I'm not exactly opposed, but I want to check a few things.

@sjw36 , what is progress on moving the code in `miopen-gen` to using gpu dialect ops instead of direct calls to these runtime library functions? Is that expected soon?

external/llvm-project/mlir/lib/ExecutionEngine/RocmRuntimeWrappers.cpp
zhanglx13: It is used by `mgpuMemHostRegisterFloat` and `mgpuMemHostRegisterInt32`, which are not used by any test. So can we just remove them?

krzysz00: Yeah, let's go ahead and delete anything we're not actually using - there's a bunch of functions like that we never cleaned up

external/llvm-project/mlir/lib/ExecutionEngine/RocmRuntimeWrappers.cpp
```diff
@@ -38,19 +38,18 @@ class ScopedContext {
...
+    hipCtx_t ctx;
+    HIP_REPORT_IF_ERROR(hipDevicePrimaryCtxRetain(&ctx, device));
   }
 
-  ~ScopedContext() { HIP_REPORT_IF_ERROR(hipCtxPopCurrent(nullptr)); }
+  ~ScopedContext() {
+    hipDevice_t device;
+    HIP_REPORT_IF_ERROR(hipDeviceGet(&device, /*ordinal=*/defaultDevice));
+    HIP_REPORT_IF_ERROR(hipDevicePrimaryCtxRelease(device));
+  }
```
krzysz00: Honestly, the fix here is to completely drop the ScopedContext stuff and write in the commit message that it's all deprecated.

external/llvm-project/mlir/lib/ExecutionEngine/RocmRuntimeWrappers.cpp
```diff
@@ -141,11 +140,6 @@ extern "C" void mgpuMemcpy(void *dst, void *src, size_t sizeBytes,
       hipMemcpyAsync(dst, src, sizeBytes, hipMemcpyDefault, stream));
 }
 
-extern "C" void mgpuMemset32(void *dst, int value, size_t count,
-                             hipStream_t stream) {
```
krzysz00: Please check to see if this is used or referenced upstream - there are other people who rely on these RocmRuntimeWrappers

external/llvm-project/mlir/lib/ExecutionEngine/RocmRuntimeWrappers.cpp
```diff
@@ -202,7 +196,116 @@ mgpuMemGetDeviceMemRef1dInt32(int32_t *allocated, int32_t *aligned,
   return {devicePtr, devicePtr, offset, {size}, {stride}};
 }
 
-extern "C" void mgpuSetDefaultDevice(int32_t device) {
```
krzysz00: For example, setDefaultDevice() is used by the lowering of `gpu.set_default_device`

external/llvm-project/mlir/lib/ExecutionEngine/RocmRuntimeWrappers.cpp
```diff
@@ -202,7 +196,116 @@ mgpuMemGetDeviceMemRef1dInt32(int32_t *allocated, int32_t *aligned,
...
+                                     unsigned short *aligned, int64_t offset,
+                                     int64_t size0, int64_t size1,
+                                     int64_t size2, int64_t size3,
+                                     int64_t size4, int64_t stride0,
+                                     int64_t stride1, int64_t stride2,
+                                     int64_t stride3, int64_t stride4) {
+  HIP_REPORT_IF_ERROR(hipFree(aligned));
+}
+
+extern "C" void mgpuMemCopy5DHalf(
```
krzysz00, line 281: Aside from the deallocator, can we investigate ways to not use these? If not here, then could you file a tiket so we get around to it later?

krzysz00 (commented):
Let's not break upstream with this, but it looks pretty good so long as we remember to follo up on either upstreaming or deleting the custom stuff.

external/llvm-project/mlir/lib/ExecutionEngine/RocmRuntimeWrappers.cpp
zhanglx13: @krzysz00 so why is it not good to use these?

external/llvm-project/mlir/lib/ExecutionEngine/RocmRuntimeWrappers.cpp
zhanglx13: I did find that mgpuMemset32 and mgpuSetDefaultDevice are used in rewrite patterns for gpu.memset and gpu.set_default_device. I'll add them back.

external/llvm-project/mlir/lib/ExecutionEngine/RocmRuntimeWrappers.cpp
zhanglx13: @krzysz00 I noticed that for some gpu runtime functions, their high level IR functions are used in mlir (like `gpu.*` functions). And for others ( which seem to be not basic), the runtime `mgpu*` functions are generated directly. Is there a way to determine when to use the basic functions (`gpu.*`) and when to use non-basic functions (`mgpu*`)?
Is this related to the new ticket [MLIR#555](https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/555)?

krzysz00: Yeah

The historical situation is that we called those runtime wrappers directly when generating tests (I wasn't around when we did that, and this was before we started having a split between `external/` and `mlir/`). Then, upstream, at some point, added `gpu.*` wrappers around these functions, and we didn't move to them.

The long-term plan is, I think, to move to these `gpu.*` wrappers to make things less fragile and to make our code clearer. However, there are some functions that don't have `gpu.*` equivalents yet, and for those we need to create them or find a way to not use them anymore. (ex. these 5D memcopy functions could probably be replaced by a wrapper that extracts the pointer from the memref and calls a 1D memcopy, keeping the C API small.

krzysz00 (approved):
Nice cleanup, approved

Let's make a plan for healing our divergences with upstream's wrappers.

external/llvm-project/mlir/lib/ExecutionEngine/RocmRuntimeWrappers.cpp
sjw36: Apologies, I already started this on https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/555.

I can pass off to you @zhanglx13 if you like. It is pretty simple change but there are alot of tests to change.

zhanglx13: @sjw36 Yes you can assign that ticket to me.

jerryyin: @zhanglx13 Just a reminder that if you decide to do that one, implement that from a independent PR. I feel like there's already enough of the code changes in this one, and we want to regress it as its own feature.

zhanglx13: @jerryyin Sure. I'll merge this one as soon as I get all tests passing.

PR #657, opened 2022-06-13
MLIR#556 Remove rocprof when using MIOpenDriver for benchmarking

zhanglx13:
[MLIR#556](https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/556)
When MIOpenDriver fails to run with a config, NULL is used as the TFlops in the output csv files


mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -288,6 +286,11 @@ def runConfigWithMIOpenDriver(commandLine, envs):
         p1.kill()
         print("MIOpen benchmark timed out")
         outs, errs = p1.communicate()
+    else:
+        outs=outs.decode()
+        # Extract Elapsed time in ms
+        elapsedTimeInMs = outs.split("Elapsed: ")[1].split("ms (average)")[0]
```
jerryyin: Several comments:
 - Please leave a more detailed comment on what format is expected here
 - Needs better error handling: Could you check if the elapsed time will be populated when MIOpenDriver fail to run? We do have some int8 nhwc conv unsupported by MIOpenDriver and we want to make sure that doesn't break this script.

zhanglx13: @jerryyin Can you elaborate about the unsupported configs?
MIOpenDriver.py auto generates different combinations of {directions, data_types, layouts} but it skips int8 non-forward convolutions, i.e. only convint8 -F 1 ... is used as configs.  Does this setup still contains configurations that are not supported by MIOpenDriver?
MIOpenDriver only generates one run for int8 nhwc:
`convint8 -F 1 -f NHWC -I NHWC -O NHWC -n 256 -c 256 -H 14 -W 14 -k 256 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1`

jerryyin: @zhanglx13 We used two different ways to build MIOpen binary. When `MIOPEN_USE_MLIR=0` setup, all NHWC configs will fail. Please take a look at nightly results int8 + NHWC section to observe this more clearly. Because of that scenario, we want to make sure this script does not give up if the pattern is not observed.

zhanglx13: @jerryyin when MIOpenDriver fails to run, do you want to remove the whole table entry for this config or just the TFlops field of the table entry?

zhanglx13: or put NaN as the TFlops?

jerryyin: @zhanglx13 How do we handle this for existing failures like this? I suspect that we just don't populate anything for that field and then later we get nothing on the table. I'd prefer to be consistent here.

To test, maybe pull that nightly trigger on and see if it can continue to generate the performance report correctly?

yiqian1: Currently the table shows "FAILURE" for failed configs. Such entries are automatically excluded when computing the average.

zhanglx13: In the very first report (miopen_tuned/untuned_perf.csv), empty string is used in the TFlops column if the config fails. The empty strings are further "processed" in later reports, such as the [performance report for mlir-nightly-all](http://rocmhead.amd.com:8080/job/MLIR/job/mlir-nightly-all/395/Performance_20report/).

jerryyin: @zhanglx13 Just FYI I kicked off a perf report run on your 1d85eed commit [here](http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-657/5/pipeline/360) and looks like it is bumping into some computation failures. You can do the same to regress your newer commit 60e2eb3

zhanglx13: @jerryyin The error you saw was caused by MIOpenDriver timed out (timeout is set to be 180 in MIOpenDriver.py). When it is timed out, none is return (instead of np.nan when there is an error) and later it caused an error in the computation. So
1. I tried `60e2eb3` on ixt-rack-112 and everything is good.
2. The current CI check is running `60e2eb3` and it is still waiting for available node.
3.  How do you kick off a perf report run? Is there a quick way to do the ci check on jenkins without doing the building process?

jerryyin: 1/2: Make sense, thanks for your explanation. I think with gfx908 it is less likely to hit such timeout. 

3: Please ping me when you are on site so I can demo it to you. Kicking off a perf run is generally not required. Though for this particular PR, it is going to be helpful to see at least one passing perf run.

zhanglx13: @jerryyin I just realized that the timeout happened on gfx908. I tried it again with `60e3eb3` and it still timed out on gfx908. Were you trying to say it is less likely to happen on gfx90a?

jerryyin: @zhanglx13 Correct, things with gfx90a is that its filesystem is slow, but this isn't an issue on gfx908 in general. I just help you kicked off another run.

BTW, my observation is both failures happen in the same host of `ixt-hq-99`, it can be due to a host specific failure. This usually wind into one of the two roads:
 - Reboot the node (or ask Omkar to do it)
 - (Ask Omkar the login credential of the node), then see if you can reproduce it in that machine.

krzysz00: Re the on-failure question: existing code puts in a NaN

mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -273,17 +273,21 @@ def runConfigWithMLIR(config: ConvConfiguration):
...
     print("Running MIOpen Benchmark: ", ' '.join(commandLine))
-    # invoke rocprof + MIOpenDriver.
-    p1 = subprocess.Popen(profilerCommand, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=envs)
+    # invoke MIOpenDriver.
+    p1 = subprocess.Popen(MIOpenDriverCommand, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=envs)
     # get output.
     try:
         outs, errs = p1.communicate(timeout=180)
         if len(errs) > 0:
-            print("MIOpen benchmark produced errors: ", errs.decode('utf-8'))
```
krzysz00, line 286: It's still useful to have the errors printed out?

mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -273,17 +273,21 @@ def runConfigWithMLIR(config: ConvConfiguration):
...
+            return np.nan
+        else:
+            # convert bytes to str
+            outs=outs.decode()
```
krzysz00: Explicitly say `utf-8` here, please

Alternatively, `Popen` takes an `encoding` argument

mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -273,17 +273,21 @@ def runConfigWithMLIR(config: ConvConfiguration):
...
+            # Extract Elapsed time in ms
+            elapsedTimeInMs = outs.split("Elapsed: ")[1].split("ms (average)")[0]
```
krzysz00: Can we make this a regex (and use `re.compile`) instead? That feels like it'd be cleaner

mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -273,17 +273,21 @@ def runConfigWithMLIR(config: ConvConfiguration):
...
+            return float(elapsedTimeInMs)*1000000
```
krzysz00: Nit: would it be unreasonable to write the mili-to-nano conversion factor as `1.0e6`?

krzysz00 (commented):
Some thoughts

mlir/utils/jenkins/MIOpenDriver.py
zhanglx13: Previously the error message was consumed by rocprof silently and it was reflected in the final report (as FAILED or --). Do we really want to print these error messages when running the tests? Maybe people will check the report anyway to see if there are unsupported configs.

mlir/utils/jenkins/MIOpenDriver.py
jerryyin: @zhanglx13 According to performance report [here](http://rocmhead.amd.com:8080/job/MLIR/job/mlir/job/PR-657/13/Performance_20report/), MLIR vs MIOpen
 - MLIR tflops for i8 should display values
 - Below, in MLIR perf test per kernel based values, the i8 columns for mlir should also have values

mlir/utils/jenkins/MIOpenDriver.py
krzysz00: It may be good for debugging to know if there's a failure failure vs an unsupported failure vs something else.

But we could put that behind some flavor of `--debug` switch, or drop it entirely. Not too important

jerryyin: Let's keep it. It's useful that this can help differentiate between different type of failures right from the log.

mlir/utils/jenkins/MIOpenDriver.py
zhanglx13: @jerryyin I tried the given config and I got the error message: Error: only single f32 function result supported
And this is how I did the experiment:
1. run `./bin/MIOpenDriver.py convint8 -F 1 -f NHWC -I NHWC -O NHWC -n 256 -c 256 -H 14 -W 14 -k 256 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1` and I printed out the command line options for miopen-gen, which is `--operation conv2d -t i8 -x2 --fil_layout kyxc --in_layout nhwc --out_layout nhwk --batchsize 256 --in_channels 256 --in_h 14 --in_w 14 --out_channels 256 --fil_h 3 --fil_w 3 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h 1 --padding_w 1`
2. Then I run miopen-gen | mlir-miopen-driver | mlir-rocm-runner as follows
`./bin/miopen-gen -ph ${miopen_gen_CMD} | ./bin/mlir-miopen-driver -c | ./bin/mlir-rocm-runner --shared-libs=./external/llvm-project/llvm/lib/libmlir_rocm_runtime.so,./lib/libconv-validation-wrappers.so,./external/llvm-project/llvm/lib/libmlir_runner_utils.so', '--entry-point-result=void` where ${miopen_gen_CMD} is the above miopen-gen command option
3. And I got the error message.

I build mlir-rocm-runner following the step on jenkins
`cmake -G Ninja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DMLIR_MIOPEN_DRIVER_ENABLED=1`
`ninja miopen-gen mlir-miopen-driver mlir-rocm-runner ci-performance-scripts`
FWIW, it passes all mlir and mlir-miopen tests.
Can you see what is going wrong here?

jerryyin: Try with `miopen-gen --conv-config "--operation conv2d -t i8 -x2 --fil_layout kyxc --in_layout nhwc --out_layout nhwk --batchsize 256 --in_channels 256 --in_h 14 --in_w 14 --out_channels 256 --fil_h 3 --fil_w 3 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h 1 --padding_w 1"` instead.

I'm unclear about why you bumped into the issue but if `--conv-config` can pass please give a comparison between passed/failed IR and let me know what makes it different using the command directly.

krzysz00: Hold on, the variable substitution might've led to there being one long argument instead of a bunch of arguments. Perhaps structure `miopen_gen_CMD` as `miopen_gen_CMD=(-a 1 -b 2)` as opposed to `miopen_gen_CMD="-a 1 -b 2"`

zhanglx13: @jerryyin @krzysz00 Thank you for the suggestions. Actually the error is caused with an extra `', '` in the argument for mlir-rocm-runner.

krzysz00 (approved):

mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -273,17 +274,25 @@ def runConfigWithMLIR(config: ConvConfiguration):
...
         if len(errs) > 0:
             print("MIOpen benchmark produced errors: ", errs.decode('utf-8'))
+            return np.nan
+        else:
+            # convert bytes to str
+            outs=outs.decode('utf-8')
+            # Extract Elapsed time in ms from the output of MIOpenDriver
+            # Use regular expression to match the contents between
+            # "Elasped: " (note the space at the end) and "ms"
+            pattern = re.compile(r"Elapsed: (.*)ms");
```
krzysz00: Ok, actually, I know I just said "approved", but one minor point before this goes in if possible - move this to a global constant so it only gets compiled once.

mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -30,6 +31,8 @@
 DIRECTIONS = ['-F 1', '-F 2', '-F 4']
 DATA_TYPES = ['conv', 'convfp16', 'convint8']
 LAYOUTS = ['NHWC', 'NCHW']
+# Compiled regexp object used for extracting elapsed time from MIOpenDriver's output
+pattern = re.compile(r"Elapsed: (.*)ms")
```
krzysz00: This sort of thing should really be an `ALL_CAPS_CONSTANT` - perhaps `ELAPSED_TIME_RE`

mlir/utils/jenkins/MIOpenDriver.py
zhanglx13: @jerryyin Can you please take a look at the performance report [here](http://rocmhead.amd.com:8080/job/MLIR/job/mlir/job/PR-657/46/Performance_20report/). If it looks ok then this PR can be merged.

jerryyin (approved):
Thank you for your work!

mlir/utils/jenkins/MIOpenDriver.py
jerryyin: @zhanglx13 LGTM!

PR #659, opened 2022-06-15
HOTFIX: disable e2e tests for unsupported machines

sjw36:


krzysz00 (approved):
For the record, the fix here isn't super hot, but I noticed it was a potential issue

PR #660, opened 2022-06-16
Non xdlops fusion writeback and GEMM rework

krzysz00:
1. Analogously to the xdlops version, use coordinate transforms that start at (block id, thread id, iteration #) to compute the tensor coordinates for writing matrix C
2. As a consequence of this, make the C register matrix a 1D memref, which also makes the LLVM optimizer happy
3. As a consequence of *that*, rewrite `threadwise_gemm` to use `transforming_for`
4. Having make this change, make the changes needed to the new way of doing fusion to enable cases the xdlops tests didn't cover
5. Remove the old way of doing fusion
6. Remove `threadwise_copy` since it has no users

One result of this change was identifying the bug in blockwise_gemm that might've caused kPack to be broken for non-xdlops. We should do a followup experiment to see if kPack works now.

Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/547


mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -83,221 +85,150 @@ struct BlockwiseGemmRewritePattern : public OpRewritePattern<BlockwiseGemmOp> {
...
-
-    auto iva = loopReadMatrixAOp.getInductionVar();
-
-    // Threadwise copy from LDS (naive tensor) to register (generic tensor).
-
-    // Set copy sorce and dest coordinate acoording to original C++ logic:
-    SmallVector<Value, 4> matrixAThreadwiseCopySourceCoords;
-    if (KPack > 1) {
-      matrixAThreadwiseCopySourceCoords = {
-          zeroConstantOp, zeroConstantOp, iv,
```
krzysz00, line 205: This and its anologue for B is the suspected bug in non-xdlops kPack

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -43,624 +45,433 @@
...
-        transforms = transforms_l;
-        break;
-      } else {
-        // 2.1. Test aligned input with output type
-        auto inpIdxMap = idxMaps[idx].template cast<AffineMapAttr>();
-        if (!checkCompatibleTypes(inp.getType(), inpIdxMap, out.getType(),
-                                  outIdxMap)) {
-          return fail;
-        }
+      if (!inpIdxMap.isIdentity()) {
```
sjw36, line 402: @jungpark-mlir can you confirm this case with MIXR (NCHW->NHWC).

sjw36 (approved):
Overall it's great simplification. Thx

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
krzysz00: I could've sworn there's special handling for that way further up the pipe - but yeah, good call

krzysz00:
@whchung @jerryyin Can I get a look at the gemm changes?

jerryyin:
@krzysz00 Will take a look today, please hold off merging.

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -750,18 +730,27 @@ def MIOpen_BlockwiseGemmV2Op:
...
-                   MemRefRankOf<[F32, F16, BF16, I32], [3, 4]>:$matrixB,
-                   MemRefRankOf<[F32, F16, BF16, I32], [3, 4]>:$matrixC)> {
+    Arguments<(ins MemRefRankOf<[F32, F16, BF16, I32], [1]>:$matrixA,
+                   MemRefRankOf<[F32, F16, BF16, I32], [1]>:$matrixB,
+                   MemRefRankOf<[F32, F16, BF16, I32], [1]>:$matrixC,
+                   IndexAttr:$g,
+                   IndexAttr:$k,
+                   IndexAttr:$m,
+                   IndexAttr:$n,
+                   IndexAttr:$kPack)> {
```
jerryyin: Those are statically decided, do we want to use them as arguments? Also, they seems to be able to be deducible from the memref rank of A/B/C. Thinking of this, I actually prefer they'd stay as attributes. (Marginal benefit of performance is that constant is better than consuming register bandwidth).

mlir/include/mlir/Dialect/MIOpen/TransformMapBuilder.h
```diff
@@ -166,6 +170,36 @@ class TopDownTMBuilder : public TransformMapBuilder {
   int64_t paddingSign() const override final;
 };
 
+/// A wrapper around a TopDownTMBuilder that looks up end dimensions in a
+/// provided map, used for cases such as when merge()ing will create extra
+/// dimensions. This takes the builder by reference and does modefiations there
+/// and thus doesn't expose its own get() method. Everything is defined here
+/// to increase inlineability
+struct TopDownTMBottomDimsWrapper {
```
jerryyin, line 178: I actually don't see a use case of this. Do we actually need this wrapper?

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -83,221 +85,150 @@ struct BlockwiseGemmRewritePattern : public OpRewritePattern<BlockwiseGemmOp> {
...
+                            << "N: " << nC << "\n"
+                            << "NRepeat: " << nRepeat << "\n"
+                            << "NPerThread: " << nPerThread << "\n");
+
+    TopDownTMBuilder strideLDSBufferA(
+        b, {"g", "k", "mRepeat", "mPerThread", "kpack"},
+        {g, k, mRepeat, m / mRepeat, kPack}, loc);
+    strideLDSBufferA.passThrough({"g", "k"});
+    strideLDSBufferA.embed("m", 2, m, {"mRepeat", "mPerThread"},
+                           {mRepeatStride, 1});
```
jerryyin: Looks like a nice approach! Instead of having implicit inner and outer loop I like the approach of converting them to 2 dimensions.

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -83,221 +85,150 @@ struct BlockwiseGemmRewritePattern : public OpRewritePattern<BlockwiseGemmOp> {
...
-              loc, lbb.create<MulIOp>(loc, ivb, NPerLevel1ClusterConstantOp),
-              op.threadOffsetB())};
+    // Main loop.
+    LLVM_DEBUG(llvm::dbgs() << "Outer loop:\n "
+                            << "k =  " << k << "\n"
+                            << " kPerThread = " << kPerThread << "\n");
+    auto loopOp = b.create<AffineForOp>(loc, 0, k, kPerThread);
+    OpBuilder::InsertionGuard guard(b);
+    b.setInsertionPointToStart(loopOp.getBody());
+    Value kOffset = loopOp.getInductionVar();
```
jerryyin, line 181: Shouldn't outer loop be `g * k`? Both stay the same for the size required to copy to lds for A and B matrix.

krzysz00:
@jerryyin These changes have passed nightly tests. Do you want a manual run of non-xdlops tuning to validate them as well?

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -83,221 +85,150 @@ struct BlockwiseGemmRewritePattern : public OpRewritePattern<BlockwiseGemmOp> {
...
+
+    SmallVector<Value, 5> registerStartCoords(5, zeroConstantOp);
+    SmallVector<Value, 5> ldsBufferAStartCoords = {
+        zeroConstantOp, kOffset, zeroConstantOp, op.threadOffsetA(),
+        zeroConstantOp};
+    auto copyALoop = b.create<TransformingForOp>(
+        loc, ArrayRef<ValueRange>{ldsBufferAStartCoords, registerStartCoords},
+        ArrayRef<Attribute>{transformsA, b.getArrayAttr(threadACopyViewAttr)},
+        ArrayRef<int64_t>{g, kPerThread, mRepeat, mPerThread, kPack},
```
jerryyin: I might be just confused. I think the outer loop is already looping through the k dimension, now in the transforming for op, are we loop through the k dimension a second time by `kPerThread`?

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1261,69 +1244,114 @@ struct GridwiseGemmRewritePattern : public OpRewritePattern<GridwiseGemmOp> {
...
+    Type destType = op.c().getType().cast<MemRefType>().getElementType();
+    if (destType != accumulatorType) {
+      auto convertedCType =
+          threadCRegisterMemRefType.clone(destType).cast<MemRefType>();
+      Value convertedC = b.create<miopen::GpuAllocOp>(loc, convertedCType);
+      auto convertLoop = b.create<TransformingForOp>(
+          loc, ArrayRef<ValueRange>{{zeroConstantOp}},
+          ArrayRef<Attribute>{b.getArrayAttr({})},
+          /*bounds=*/convertedCType.getShape(), /*strides=*/llvm::None,
+          /*useIndexDiffs=*/true, /*forceUnroll=*/true);
```
jerryyin, line 1289: Just want to double check, was this loop previously inside threadwise gemm? (And now you raised it outside?)

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
```diff
@@ -906,116 +907,81 @@ struct ThreadwiseGemmRewritePattern
...
+    int64_t m = op.mAttr().getInt();
+    int64_t n = op.nAttr().getInt();
+    int64_t kPack = op.kPackAttr().getInt();
+    LLVM_DEBUG(llvm::dbgs() << "Threadwise gemm:\n"
+                            << "g = " << g << "\n"
+                            << "k = " << k << "\n"
+                            << "m = " << m << "\n"
+                            << "n = " << n << "\n"
+                            << "kPack = " << kPack << "\n");
+    SmallVector<int64_t> dimensions = {g, k, m, n, kPack};
```
jerryyin: I like it that you are using coordinate transfer for the inner k/m/n computation. However, for the inner most kpack dimension, will it get consolidated as a vectorized load/store/computation?

jerryyin (commented):
Looks like a fairly aggressive but clean refactor, thanks for the work.

A decent amount of questions are due to my in-familiarity with the coordinate transfer and hoping to get clarified.

jerryyin:
> Do you want a manual run of non-xdlops tuning to validate them as well?

I'm fine either ways and trust you for already testing some of the combinations. If you are uncertain of the coverage, please run through nightly non-xdlops when CI is back.

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
krzysz00: The thing is, the memrefs A, B, and C are all 1D buffers, and so they don't carry a rank. Having them be 1D buffers makes the alloca()-to-registers promoter that has to exist somewhere happy, and only comes at the cost of having to manually keep track of m, n, and k.

mlir/include/mlir/Dialect/MIOpen/TransformMapBuilder.h
krzysz00: We need this wrapper for sanity when `expandNamesInPlace()` is called. It's the missing analogue to `BottomUpTMTopDimsWrapper`, which sees a lot of use in ConvToGemm

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: No, since the `threadwise_gemm` also takes an arbitrary `g` setting. We could force `g = 1` for `threadwise_gemm`, but I elected not to so there's no massive surprises later

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: Yep - the outer loop goes through the k dimension with a stride of `kPerThread`, the inner one goes with a bound of `kPerThread` with a stride of `1`

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: This loop used to be in `threadwise_copy` - it's been moved up for consistency with the xdlops gemm and so fusion behaves in a mixed-type context

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
krzysz00: Ok, so, with `kPack = 2`, the code will be 
```
c[0 * N + 0] = c [0 * N + 0] + a[(0 * M + 0) * kPack + 0] + b[(0 * N + 0) * kPack + 0]
c[0 * N + 0] = c[0 * N + 0] + a[(0 * M + 0) * kPack + 1] + b[(0 * N + 0) * kPack + 1]
c[0 * N + 1] = c[0 * N + 1] + a[(0 * M + 0) * kPack + 0] + b[(0 * N + 1) * kPack + 0]
...
```
And once this is all SSA-ified, that intermediate store/load pair will go away.

krzysz00:
What I'm uncertain of is coverage with non-default tuning

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
jerryyin: Talking of which... Please forgive me for asking more design decisions: Why making the original threadwise_gemm dimension from 3 or 4 to 1? I think to enforce the kpack dimension exists, we can enforce it to be 4 dimension tensor.

mlir/include/mlir/Dialect/MIOpen/TransformMapBuilder.h
jerryyin: I only see `TopDownTMBuilder` used in ConvToGemm lowering. So in other words this is not strictly needed in this PR but can be good to have in the future?

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseToStdlib.cpp
jerryyin: Sounds good

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
jerryyin: I get confused between bounds and strides and didn't realize an empty strides means all 1s for each dimension

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Sounds good

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
jerryyin: I don't quite follow here: What's the problem if g is not equal to 1 && we move it to the outer loop?

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
krzysz00: No, no, the kpack is a view of the buffer - the gemm operators take pointers to their underlying buffers, and the attributes specify how to interpret the buffers

mlir/include/mlir/Dialect/MIOpen/TransformMapBuilder.h
krzysz00: No, there's uses of `TopDownTMBottomDimsWrapper` in `GridiwiseToBlockwise`

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: It's an arbitrary choice. I've chosen to allow threadwise_gemm to have `g != 1` and to handle that case, even though gridwise_gemm sets `g = 1` for blockwise gemm. I could - maybe should? - just not have a `g` dimension in blockwise or threadwise gemm.

mlir/include/mlir/Dialect/MIOpen/TransformMapBuilder.h
jerryyin: Ahh that's right, silly question.

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
jerryyin: I'm not sure I made the point clear:
 - In the left side of this diff, `ThreadwiseGemmOp` used to take `MemRefRankOf<[F32, F16, BF16, I32], [3, 4]>`
 - Now you have make the decision instead to take `MemRefRankOf<[F32, F16, BF16, I32], [1]>`

The question is why you do prefer to pass in a flat 1 dimensional buffer instead of buffer that carries that dimension information? That approach that carries the dimension information sounds more natural to a threadwise gemm.

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
jerryyin: Would allowing g!=1 add the capability for certain use cases? I'm not familiar with it enough to tell.

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
krzysz00: A flat 1D buffer doesn't freak out LLVM

krzysz00: And I'm planning to move the arguments to blockwise_gemm to be 1D buffers as well

krzysz00: We could have blockwise_gemm construct the 2D view of the buffer and then pass _that_ to threadwise_gemm, and it'd probably work fine, but that doesn't align with how we handle the C buffer (which needs to be a view of something 1D for the writeback to work)

jerryyin (approved):
LGTM, thank you!

sjw36 (approved):
Looks good.

sjw36:
clang format errors..

jerryyin:
@sjw36 @jungpark-mlir For http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-660/10/pipeline/128 I'm pretty sure the clang format error was introduced from: https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/665. (Because that `PassDetail.h` is introduced from the PR).


 - I have seen it from a number of PRs. Now the strange thing is that why clang-format didn't block 665?
 - Could you take a further look on how/why clang-format is broken? My theory seems to point at when there needs to be a merge commit happens and the changes on 665 show up on the diff list compared with an existing PR.

krzysz00:
Update: thanks to @jerryyin , I've been able to run the non-xdlops code on lockhart, and I'm waiting to get results on that. Good news is: nightlies have passed.

krzysz00:
I've also done the non-xdlops parameter sweep, and so am broadly confident that we didn't flub any tuning-dependent logic here

PR #661, opened 2022-06-20
[DO NOT SQUASH] MLIR#557: assign access_modes to func.argAttrs and func.resultAttrs

sjw36:
convert from an array attribute on the func to use argAttrs.


krzysz00 (approved):
Seems reasonable to me

Come to think it - are the pass manager CL options supposed to be applied before or after you define the pipeline?

sjw36:
> Seems reasonable to me
> 
> Come to think it - are the pass manager CL options supposed to be applied before or after you define the pipeline?

It's mostly for --print-ir-after-all, etc.

krzysz00:
Does it also mean we get `--stop-after` now?

sjw36:
> Does it also mean we get `--stop-after` now?

Possibly, haven't tried that. But since there are multiple PassManagers, maybe not.

krzysz00:
FYI, merge conflict

sjw36:
Significant changes (I just didn't like how verbose it was).. please re-review.

krzysz00 (commented):
This looks fine to me, and seems to fit in with the same scheme as for attributes like `kernel`.

Do we know if there's a way to make it clearer that these attributes can be put on arguments in tablegen? If not, can this PR also update the documentation to mention these attributes?

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -230,9 +218,13 @@ class LaunchOpConversion : public OpConversionPattern<async::LaunchOp> {
       auto fidx = i - diff;
       Value opr = operands[i];
       // move input memories to GPU
-      if (!opr.getDefiningOp<gpu::AllocOp>()) {
-        opr =
-            moveMemory(rw, opr, fidx, accessVec[fidx], copyBackOprs, asyncDeps);
+      if (opr.getType().isa<MemRefType>() &&
+          !opr.getDefiningOp<gpu::AllocOp>()) {
+        bool readAccess{func.getArgAttr(fidx, FuncOp::getReadAccessAttrName())};
```
pcf000, line 223: Why are these two otherwise-identical operations done in two different ways?

sjw36: Forgot to update the second. I wanted to get rid of the C-style cast.

PR #662, opened 2022-06-20
Replacing llvm::errs with llvm::dbg static lib path

jerryyin:
Follow-up to https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/620. This is a breeze thanks to @krzysz00 's refactors that moved the implementation to appropriate cpp files.

This should get rid of redundant error messages in the tuning process.


krzysz00 (approved):
Approved, good cleanup!

(I can bikeshed all day whether `conv2d-generator` is a better or worse debug key than `conv2d-gen`, but you've made a reasonable choice so let's go with it and change it later if we get irritated)

PR #663, opened 2022-06-20
[NFC] Rearrange and rework MIOpen lowering passes

krzysz00:
- Rename the lowering passes to have clearer names and to match the
files they're defined in
Closes
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/563

- Make all the lowering passes that can accommodate it FuncOp passes
Fixes
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/514

- Separate out the gemm lowering from the transforming_for -> affine
pass, making it easier to peek under the hood

- Make all the gemm lowering passes partial dialect conversions so we
get better errors when lowering fails ( Thanks @jerryyin for this
observation )

- Update tests to match the new names

- While I was there, fix an issue in the xdlops lowering where a value
is defined after it is used


sjw36 (approved):
Much better. Thanks.

PR #664, opened 2022-06-21

pcf000:
One more variation on the theme.  This one uses Krzysztof's idea of moving the partition-config interface into the Pass itself, and subclassing the Pass to change behavior.  I think I like this one.

There is no abstract class this time -- the base TosaPartitionPass implements the simple interface but allows it to be overridden.  I override for the with-options variant and in the test pass for its little variations.  I could see an argument that the base version should use the options;  what I have feels natural, but on the other hand the options are already defined in Passes.td.

I'm sorry I can't show a direct comparison with the last version, but I'll try to highlight differences.

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
@@ -52,8 +52,14 @@ def TosaPartition : Pass<"tosa-partition", "ModuleOp"> {
   let constructor = "createTosaPartitionPass()";
 
   let options = [
-    Option<"nofront", "nofront", "bool", /*default=*/"false",
-           "Don't gather ops ahead of Conv2D">
+    ListOption<"anchorOps", "anchor-ops", "std::string",
+               "One or more operations to be used as focus of partitioned "
+               "kernels",
+               "llvm::cl::ZeroOrMore">,
+    Option<"partitionTagOpt", "partition-tag", "std::string",
+           /*default=*/"\"kernel\"", "Attribute for outlined functions">,
+    Option<"trailingOnly", "trailing-only", "bool", /*default=*/"false",
+           "Don't gather ops ahead of anchor op">
   ];
   let dependentDialects = ["tosa::TosaDialect"];
 }
--- a/external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
+++ b/external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
@@ -6,11 +6,15 @@
 //
 //===----------------------------------------------------------------------===//
 //
-// Replace conv2d followed by elementwise with call to function containing them.
+// Replace conv2d followed by elementwise op with call to function containing
+// them.  Generalised, outline any anchor op, all its trailing elementwise ops,
+// and all its leading elementwise ops.  (Where "elementwise" itself is
+// generalised to include transpose and reshape ops.)
 //
 //===----------------------------------------------------------------------===//
 
 #include "mlir/Analysis/SliceAnalysis.h"
+#include "mlir/Dialect/Arithmetic/IR/Arithmetic.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/Tosa/IR/TosaOps.h"
 #include "mlir/Dialect/Tosa/Transforms/PassDetail.h"
@@ -36,6 +40,7 @@
 using llvm::SmallVector;
 
 using namespace mlir;
+using namespace mlir::tosa;
 
 namespace {
 
@@ -88,26 +93,53 @@ bool isElementwiseOp(Operation *op) {
   // clang-format on
 }
 
+bool isFuseableOp(Operation *op) {
+  return isElementwiseOp(op) || isa<tosa::TransposeOp, tosa::ReshapeOp>(op);
+}
+
+bool isZeroAttribute(Attribute value) {
+  if (auto intValue = value.dyn_cast<IntegerAttr>())
+    return intValue.getValue().isNullValue();
+  if (auto fpValue = value.dyn_cast<FloatAttr>())
+    return fpValue.getValue().isZero();
+  if (auto splatValue = value.dyn_cast<SplatElementsAttr>())
+    return isZeroAttribute(splatValue.getSplatValue<Attribute>());
+  if (auto elementsValue = value.dyn_cast<ElementsAttr>())
+    return llvm::all_of(elementsValue.getValues<Attribute>(), isZeroAttribute);
+  if (auto arrayValue = value.dyn_cast<ArrayAttr>())
+    return llvm::all_of(arrayValue.getValue(), isZeroAttribute);
+  return false;
+}
+
+bool isConstantZero(Operation *op) {
+  // test for zero
+  if (auto cst = dyn_cast<arith::ConstantOp>(op)) {
+    return isZeroAttribute(cst.getValue());
+  }
+  return false;
+}
+
 ////////////////////////////////////////////////////////////////////////////////
 
 // Inspired by / adapted from outlineIfOp() in SCF/Transforms/Utils.cpp
 // and mergeIdenticalBlocks() in Utils/RegionUtils.cpp.
 
 struct OutliningCandidate {
-  OutliningCandidate(Operation *_convOp, ArrayRef<Operation *> &_secondOps,
-                     ArrayRef<Operation *> &_frontOps, ArrayRef<Value> &_params,
-                     ArrayRef<Value> &_returnVals, StringRef _partFnName);
+  OutliningCandidate(Operation *anchorOp_, ArrayRef<Operation *> &trailingOps_,
+                     ArrayRef<Operation *> &leadingOps_,
+                     ArrayRef<Value> &params_, ArrayRef<Value> &returnVals_,
+                     StringRef partFnName_);
 
   unsigned addOp(Operation *op, unsigned orderIt);
 
-  Operation *convOp;
-  SmallVector<Operation *> secondOps;
-  SmallVector<Operation *> frontOps;
+  Operation *anchorOp;
+  SmallVector<Operation *> trailingOps;
+  SmallVector<Operation *> leadingOps;
   SmallVector<Value> params;
   SmallVector<Value> returnVals;
   std::string partFnName;
   llvm::hash_code hash;
-  FuncOp function;
+  func::FuncOp function;
 
   /// Return the order index for the given value that is within the block of
   /// this data.
@@ -134,19 +166,19 @@ unsigned OutliningCandidate::addOp(Operation *op, unsigned orderIt) {
   return orderIt;
 }
 
-OutliningCandidate::OutliningCandidate(Operation *convOp_,
-                                       ArrayRef<Operation *> &secondOps_,
-                                       ArrayRef<Operation *> &frontOps_,
+OutliningCandidate::OutliningCandidate(Operation *anchorOp_,
+                                       ArrayRef<Operation *> &trailingOps_,
+                                       ArrayRef<Operation *> &leadingOps_,
                                        ArrayRef<Value> &params_,
                                        ArrayRef<Value> &returnVals_,
                                        StringRef partFnName_)
-    : convOp(convOp_), partFnName(partFnName_), hash(0), function(nullptr) {
+    : anchorOp(anchorOp_), partFnName(partFnName_), hash(0), function(nullptr) {
   // We'll need to grab the cloned ops to avoid use-after-free.
-  for (auto *op : secondOps_) {
-    secondOps.push_back(op);
+  for (auto *op : trailingOps_) {
+    trailingOps.push_back(op);
   }
-  for (auto *op : frontOps_) {
-    frontOps.push_back(op);
+  for (auto *op : leadingOps_) {
+    leadingOps.push_back(op);
   }
   for (auto val : params_) {
     params.push_back(val);
@@ -156,11 +188,11 @@ OutliningCandidate::OutliningCandidate(Operation *convOp_,
   }
 
   unsigned orderIt = params.size();
-  for (auto *op : frontOps) {
+  for (auto *op : leadingOps) {
     orderIt = addOp(op, orderIt);
   }
-  orderIt = addOp(convOp, orderIt);
-  for (auto *op : secondOps) {
+  orderIt = addOp(anchorOp, orderIt);
+  for (auto *op : trailingOps) {
     orderIt = addOp(op, orderIt);
   }
 }
@@ -235,15 +267,15 @@ bool outliningCandidatesEquivalent(OutliningCandidate &one,
     }
   }
 
-  for (auto ops : llvm::zip(one.frontOps, two.frontOps)) {
+  for (auto ops : llvm::zip(one.leadingOps, two.leadingOps)) {
     if (!opsMatch(std::get<0>(ops), std::get<1>(ops), one, two)) {
       return false;
     }
   }
-  if (!opsMatch(one.convOp, two.convOp, one, two)) {
+  if (!opsMatch(one.anchorOp, two.anchorOp, one, two)) {
     return false;
   }
-  for (auto ops : llvm::zip(one.secondOps, two.secondOps)) {
+  for (auto ops : llvm::zip(one.trailingOps, two.trailingOps)) {
     if (!opsMatch(std::get<0>(ops), std::get<1>(ops), one, two)) {
       return false;
     }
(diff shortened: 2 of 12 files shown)


mlir/test/xmir/mobilenetv1.mid.mlir
```diff
@@ -6,20 +6,22 @@ module {
...
-// CHECK:   async.await %token : !async.token
-
-// CHECK:   %token_0 = async.launch @mobilenetv1_outlined_part_1 (%{{.*}}, %arg3, %{{.*}}) : (memref<1x112x112x32xf32>, memref<64x1x1x32xf32>, memref<1x112x112x64xf32>)
-// CHECK:   async.await %token_0 : !async.token
-        
-// CHECK:   %token_1 = async.launch @mobilenetv1_outlined_part_2 (%{{.*}}, %arg5, %arg6) : (memref<1x56x56x64xf32>, memref<128x1x1x64xf32>, memref<1x56x56x128xf32>)
-// CHECK:   async.await %token_1 : !async.token
-    
+
+// CHECK:   %token_0 = async.launch @mobilenetv1_outlined_part_1 [%token] (%{{.*}}, %arg2, %{{.*}}) : (memref<1x112x112x32xf32>, memref<3x3x32x1xf32>, memref<1x112x112x32xf32>)
```
pcf000, line 10: With the changed defaults, there are more outlined kernels and thus more launches.  I don't know why the original version had awaits after each call when the newer version doesn't.  These are changes to make the tests pass, and may not match the spirit of the original test.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -6,11 +6,15 @@
 //
 //===----------------------------------------------------------------------===//
 //
-// Replace conv2d followed by elementwise with call to function containing them.
+// Replace conv2d followed by elementwise op with call to function containing
```
pcf000, line 9: Comments are cleaned up and the nomenclature changed to match the generalised code.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -385,53 +419,59 @@ void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
...
+  assert(anchorOp->use_empty() && "expected 'op' to have no uses");
+  anchorOp->erase();
+  for (auto &op : llvm::make_early_inc_range(leadingOps)) {
     if (op->use_empty()) {
       op->erase();
     }
   }
 }
 
+} // namespace
```
pcf000, line 437: I guess I don't need to close and reopen the namespace.  I used to have PartitionConfig in here, which needed to be accessible (before it was moved into an include file).  Anyway, there is no PartitionConfig now.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -385,53 +419,59 @@ void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
...
+
+namespace {
+
 // Inspired by / adapted from TestSCFIfUtilsPass in
 // test/lib/Transforms/TestSCFUtils.cpp.
 class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
-public:
-  static bool isZeroAttribute(Attribute value) {
```
pcf000, line 406: Moved out of the class, see above.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -385,53 +419,59 @@ void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
...
-    if (auto elementsValue = value.dyn_cast<ElementsAttr>())
-      return llvm::all_of(elementsValue.getValues<Attribute>(),
-                          isZeroAttribute);
-    if (auto arrayValue = value.dyn_cast<ArrayAttr>())
-      return llvm::all_of(arrayValue.getValue(), isZeroAttribute);
-    return false;
+  // Special case:  TransposeOp's second operand must be a
+  // constant, which means we must include it too if we include
+  // the TransposeOp.  "ops" here may be either leadingOps or trailingOps.
+  void specialCaseForTranspose(Operation *op, SetVector<Operation *> &ops) {
```
pcf000: Abstracted because it's used twice.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -385,53 +419,59 @@ void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
...
-  static bool isConstantZero(Operation *op) {
-    // test for zero
-    if (auto cst = dyn_cast<arith::ConstantOp>(op)) {
-      return isZeroAttribute(cst.getValue());
-    }
-    return false;
+public:
+  TosaPartitionPass() = default;
+
+  virtual bool isAnchorOp(Operation *op) {
```
pcf000: Partitioning interface.  Defaults here, but overridable.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -385,53 +419,59 @@ void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
...
+  }
+  virtual bool isTrailingOp(Operation *op) {
+    return isFuseableOp(op);
+  }
+  virtual std::string attributeName() {
+    return "kernel";
   }
 
-  void traceInputs(Operation *op, SmallVector<Operation *> &predecessors,
+  void traceInputs(Operation *op, SetVector<Operation *> &predecessors,
```
pcf000: Leading ops (predecessors) became SetVector<> so I could abstract the TransposeOp special case.  Doesn't matter much besides that.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -444,59 +484,60 @@ class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
...
         SetVector<Value> inputNodes;
-        SetVector<Value> resultNodes(convOp->getResults().begin(),
-                                     convOp->getResults().end());
+        SetVector<Value> resultNodes(anchorOp->getResults().begin(),
+                                     anchorOp->getResults().end());
 
         // Grab a useful set of leading ops, like we do for trailing.
-        // Let's limit it to only first arguments, with single uses.
-        SmallVector<Operation *> frontOps;
-        if (!nofront) {
```
pcf000, line 484: The "nofront" test is absorbed into isLeadingOp() -- see the with-options version below.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -513,54 +554,159 @@ class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
...
 
-      // Walk until we've outlined all the Conv2D ops we can.
+      // Walk until we've outlined all the anchor ops we can.
       while (func.walk(callback).wasInterrupted()) {
       }
     }
   }
 };
 
+class TosaPartitionPassWithOptions
```
pcf000: The variant of TosaPartitionPass that obeys the command-line sub-options.  It's a subclass of TosaPartitionPass and overrides the key predicates.  The no-args constructor uses the option member variables as they are, and there's another constructor to pass them in.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -513,54 +554,159 @@ class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
...
+  TosaPartitionPassWithOptions() = default;
+  TosaPartitionPassWithOptions(ArrayRef<std::string> anchorOps_,
+                               std::string attrName, bool trailingOnly_) {
+    anchorOps = anchorOps_;
+    attributeNameOpt = attrName;
+    trailingOnly = trailingOnly_;
+  }
+
+  bool isAnchorOp(Operation *op) override {
+    if (anchorOps.empty())   // ListOption doesn't have a default value.
```
pcf000: Have to test it here, because it'll always be empty at pass-creation time.  I think that's an artifact of registering the passes before the command line is parsed.  In the last version, I did this in runOnOperation();  I'd have to override that method to do the same with this version.

Either way, the anchor-op test is always a list-membership test in this version.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -513,54 +554,159 @@ class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
...
+      anchorOps = {"tosa.conv2d","tosa.matmul","tosa.depthwise_conv2d"};
+
+    return llvm::is_contained(anchorOps, op->getName().getIdentifier().str());
+  }
+  bool isLeadingOp(Operation *op) override {
+    return !trailingOnly && (isConstantZero(op) || isFuseableOp(op));
```
pcf000: There's the "nofront" test -- !trailingOnly.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -513,54 +554,159 @@ class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
...
+  }
+  std::string attributeName() override {
+    return attributeNameOpt;
```
pcf000: I had name collisions, so "...Opt".

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -513,54 +554,159 @@ class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
...
+  }
+};
+
 } // namespace
 
 std::unique_ptr<Pass> mlir::tosa::createTosaPartitionPass() {
-  return std::make_unique<TosaPartitionPass>();
+  return std::make_unique<TosaPartitionPassWithOptions>();
```
pcf000: The default at the user level is the with-options version.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -513,54 +554,159 @@ class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
...
+  TestTosaPartitionOptionsPass() = default;
+  TestTosaPartitionOptionsPass(const TestTosaPartitionOptionsPass &) {}
+
+  void runOnOperation() override {
+    ModuleOp module = getOperation();
+    PassManager pm(module.getContext(), mlir::PassManager::Nesting::Implicit);
+    if (defaultCase) {
+      pm.addPass(tosa::createTosaPartitionPass());
+    } else if (depthwiseOnly) {
+      class DepthwiseOnlyPartitionPass
```
pcf000: The last version made specialised PartitionConfig subclasses, where this version makes specialised TosaPartitionPass subclasses.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -513,54 +554,159 @@ class TosaPartitionPass : public TosaPartitionBase<TosaPartitionPass> {
...
+    } else if (attrOne) {
+      class AttributeOnePartitionPass
+          : public TosaPartitionPass {
+      public:
+        std::string attributeName() override { return "one"; }
+      };
+      pm.addPass(std::make_unique<AttributeOnePartitionPass>());
+    } else if (nofrontArg) {
+      // Another way is to pass the values to the pass constructor.
+      pm.addPass(std::make_unique<TosaPartitionPassWithOptions,ArrayRef<std::string>>({"tosa.depthwise_conv2d"},
```
pcf000: The syntax is a bit awkward.  I needed to add the ArrayRef to the template parameters or it wouldn't find a constructor.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -88,26 +92,53 @@ bool isElementwiseOp(Operation *op) {
   // clang-format on
 }
 
+bool isFuseableOp(Operation *op) {
+  return isElementwiseOp(op) || isa<tosa::TransposeOp, tosa::ReshapeOp>(op);
+}
+
+bool isZeroAttribute(Attribute value) {
+  if (auto intValue = value.dyn_cast<IntegerAttr>())
+    return intValue.getValue().isNullValue();
```
krzysz00, line 102: Nit: I think `isZero()` exists?

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -264,19 +295,20 @@ findOutliningCandidate(OutliningCandidate &newCandidate,
 
 // Given a convolution op and its fuse-able trailing (second) and leading
 // (front) ops, remove them into a separate function.
-void outlineConvPartOps(Operation *convOp, ArrayRef<Operation *> secondOps,
-                        ArrayRef<Operation *> frontOps, ArrayRef<Value> params,
-                        ArrayRef<Value> returnVals, StringRef partFnName,
+void outlineConvPartOps(Operation *anchorOp, ArrayRef<Operation *> trailingOps,
```
krzysz00: Do we want this to be something like `outlineAnchorPartOpss`?

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
krzysz00: Does `SetVector` have ordering guarantees?

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
krzysz00: Silly question - could this be a `StringAttr` (so the string gets cached and we avoid copies) or a `StringRef` return value?

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
krzysz00: Test passes go in `[...]/mlir/test/lib/[...]` (in this case, `mlir/test/lib/Dialect/Tosa/`)

mlir/test/xmir/mobilenetv1.mid.mlir
krzysz00: Re awaits - one thing to check is for dependencies between the partitioned parts. If you took `f(x) { y = A(x); B(y) }` to `f_1(x) { A(x) }` and `f_2(y) { B(y) }`, you'll need to await for `f_1` to be finished before you can launch `f_2`.

krzysz00 (commented):
I'm going to leave the actual approval to @sjw36, but I like this approach and just have some minor style nits (and a concern about the test).

So, fundamentally, if Simon's happy with it, I've got no objections

mlir/test/xmir/mobilenetv1.mid.mlir
sjw36: @krzysz00 that happens in miopen-async-launch. The tosa-partition pass just does the outlining with tensor sequential dataflow.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
sjw36: I like this better so that derived partitioners can re-use some of these test funcs.

external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
```diff
@@ -52,8 +52,14 @@ def TosaPartition : Pass<"tosa-partition", "ModuleOp"> {
   let constructor = "createTosaPartitionPass()";
 
   let options = [
-    Option<"nofront", "nofront", "bool", /*default=*/"false",
-           "Don't gather ops ahead of Conv2D">
+    ListOption<"anchorOps", "anchor-ops", "std::string",
+               "One or more operations to be used as focus of partitioned "
+               "kernels",
+               "llvm::cl::ZeroOrMore">,
+    Option<"attributeNameOpt", "attribute-name", "std::string",
```
sjw36: How about 'partition-tag' instead of attribute-name.

sjw36 (changes requested):
Looks good. One comment to rename the name-attribute.

sjw36:
If the TosaPartitionPass is not visible outside the cpp file then it cannot be derived externally. Can you add a header to make it visible.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
pcf000: Yes, it preserves the order of addition.  That's one reason I used it originally.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
pcf000: I made it return StringRef.  So far so good.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
pcf000: I didn't find a suitable one;  most just test ConstantIndexOp.  This isZeroAttribute() I copied from LLVMDialect.cpp.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
pcf000: I went with outlinePartitionOps.

external/llvm-project/mlir/include/mlir/Dialect/Tosa/Transforms/Passes.td
pcf000: Yes.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
pcf000: Added to TosaTestPasses.cpp.  That also depended on moving class declarations to Passes.h, and required adding MLIRTosaTransforms to CMakeLists.txt in the test directory.

pcf000:
Added the two pass class declarations to Passes.h.

pcf000:
Made the suggested changes, rebased and updated and force-pushed, so the diffs may be a mess.

sjw36 (approved):
Looks good. Thanks.

PR #667, opened 2022-06-23
Force use of pull-request cget that works better on lockhart.

pcf000:
What this commit does is have the Jenkinsfile MIOpen steps do

  pip install git+https://github.com/pcf000/cget.git@use-urlretrieve

instead of

  pip install cget

The funky syntax fetches the branch I made for a cget pull request, which updates cget to use requests.urlretrieve which works with the lockhart proxy better than the existing version which uses URLopener.retrieve.  (TIL that pip can install straight from github.)

I don't know how long it'll take for a pull request to make its way to pypi.org, but this is intended to be a short-term solution.  Preloading the MIOpen dependences in the docker image is impractical because we have two places in Jenkinsfile that need them and they need different versions (because one is MIOpen's main branch and one is the int8-perf-config branch).

I've tested by running the "cmake -P install_deps.cmake" command, but haven't done a full Jenkins run yet, because that would require reenabling the lockhart hosts.

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -10,6 +10,10 @@ void buildProject(String target, String cmakeOpts) {
     }
 }
 
+void MIOpen_cget_hack() {
+      sh 'sed -i "s|virtualenv_install(cget)|virtualenv_install(git+https://github.com/pcf000/cget.git@use-urlretrieve)|" install_deps.cmake'
+}
+
 void buildMIOpen(String cmakeOpts) {
     sh '[ ! -d build ] || rm -rf build'
     cmakeBuild generator: 'Unix Makefiles',\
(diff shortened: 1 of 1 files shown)


jerryyin:
@pcf000 Thank you for poking it through! I'm all for it if this can address our issue.

> but haven't done a full Jenkins run yet, because that would require reenabling the lockhart hosts

I'm confused, do we need to do anything special to re-enable lockhart nodes? I'm assuming if this approach works then then nightly/weekly CI will be just back to normal, is that right?

krzysz00:
@pcf000 I disagree on the impracticality. With the exception of us, MIOpen has a policy of allowing their dependencies to be +- a bit, and so we can, I expect, pull in whatever `develop` calls for at the time we happen to build the Docker image.

krzysz00 (approved):

jerryyin (approved):

pcf000:
@jerryyin , Omkar said he would disable them.  On the other hand, 6 and 15 aren't responding and may have hardware issues, 16 is up but doesn't have an agent, and 19 is up but doesn't have the mlir label.  I'll try to relaunch 16 and see if we're supposed to stay off 19.

@krzysz00 , they list specific commits for rocm-cmake and boost is either 1.72 or 1.79, so I figured they cared. I didn't want to mess with that.

PR #670, opened 2022-06-24
Add MIOpen dependency installer to our Dockerfile

krzysz00:


krzysz00:
(Note, I have successfully built this image, and am in the process of manually ensuring we can run a static library + MIOpen build inside it)

jerryyin (approved):
Yep, that's the approach I'd take, let's push it in. Thank you!

krzysz00:
Update: yeah, I can build MIOpen on here (after I remembered to set `CXX_COMPILER` like we do in the Jenkinsfile)

PR #671, opened 2022-06-24
Don't install MIOpen dependencies in CI

krzysz00:


jerryyin (approved):
Wired, why does the diff still include `Dockerfile.rocm` after you have merged?

pcf000:
I thought it would be more complicated.  Assuming the int8-perf-whatever branch isn't bothered, this looks good to me.

pcf000 (approved):

krzysz00:
I went and rebased onto `miopen-dialect` to fix the issue

pcf000:
The ml-ci.amd.com failure is

CMake Error at CMakeLists.txt:53 (find_package):
  Could not find a configuration file for package "ROCM" that is compatible
  with requested version "0.7.3".

  The following configuration files were considered but not accepted:

    /opt/rocm/share/rocm/cmake/ROCMConfig.cmake, version: 0.6.1
    /opt/rocm-5.0.1/share/rocm/cmake/ROCMConfig.cmake, version: 0.6.1

I didn't look very hard, but I remember that one of the differences between the two MIOpen branches was the rocm.cmake version.

PR #672, opened 2022-06-27
Update tip of miopen to develop

jerryyin:
Reverting to develop to help with correct miopen dependency version.

This will temporarily make int8 tuning run slower until https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1589 merged.

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -28,7 +28,7 @@ void buildMIOpen(String cmakeOpts) {
 }
 
 void getAndBuildMIOpen(String prefixOpt, String cmakeOpts) {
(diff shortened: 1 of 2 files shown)


krzysz00:
On the one hand, we should be up to develop for release purposes (so that the 5.3 branch can go to MIOpen's 5.3 commit). On the other hand, the MIOpen dependencies thing doesn't need to happen right now. 

So, ... do you expect that `int8-perf-config` PR to get in by Thursday?

jerryyin:
> So, ... do you expect that int8-perf-config PR to get in by Thursday?

I can guarantee it will go in by ROCm 5.3 in MIOpen.

krzysz00 (approved):

PR #673, opened 2022-06-28
Replace mgpu with gpu ops

zhanglx13:
In the gpu wrapper function, use gpu.alloc, gpu.memcpy, and gpu.dealloc instead of `mgpuMemAlloc5D*`, `mgpuMemCopy5D*`, and `mgpuDealloc5D*`.

Now we should be safely delete these `mgpu*5D*` functions restored in the following PR:
https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/658

### DIFF
--- a/mlir/tools/miopen-gen/miopen-gen.cpp
+++ b/mlir/tools/miopen-gen/miopen-gen.cpp
@@ -690,141 +690,31 @@ static FuncOp createGPUWrapper(ModuleOp &module, const KernelIF &kernel) {
         loc, b.create<arith::ConstantIntOp>(loc, deviceNum.getValue(),
                                             b.getIntegerType(32)));
 
-  // Emit some constant values for HIP runtime API calls.
-  auto oneConstantI32Op =
-      b.create<arith::ConstantIntOp>(loc, 1, b.getIntegerType(32));
-  auto twoConstantI32Op =
-      b.create<arith::ConstantIntOp>(loc, 2, b.getIntegerType(32));
-
-  auto getGpuAllocFunc = [&](mlir::Type elemType) -> FuncOp {
-    StringRef allocFuncName = "mgpuMemAlloc5DFloat";
-    if (elemType.isF32()) {
-    } else if (elemType.isF16()) {
-      allocFuncName = "mgpuMemAlloc5DHalf";
-    } else if (elemType.isBF16()) {
-      allocFuncName = "mgpuMemAlloc5DBF16";
-    } else if (elemType.isInteger(8)) {
-      allocFuncName = "mgpuMemAlloc5DInt8";
-    } else if (elemType.isInteger(32)) {
-      allocFuncName = "mgpuMemAlloc5DInt32";
-    }
-    auto unk5DType = MemRefType::get({-1, -1, -1, -1, -1}, elemType);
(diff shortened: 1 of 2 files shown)


mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -690,141 +690,31 @@ static FuncOp createGPUWrapper(ModuleOp &module, const KernelIF &kernel) {
...
-    cpuMem.push_back(castOp);
+  for (auto pair : llvm::enumerate(kernel.params)) {
+    mlir::Value arg = block->getArgument(pair.index());
+    cpuMem.push_back(arg);
 
     // Emit GPU memory allocation function calls.
-    auto gpuAllocOp = b.create<func::CallOp>(loc, getGpuAllocFunc(elemType),
-                                             ValueRange{castOp});
+    auto gpuAllocOp =
+        b.create<gpu::AllocOp>(loc, arg.getType(), mlir::Type(), ValueRange{},
```
krzysz00: Would it be possible to give those default-valued arguments (the `ValueRange{}` and the like) a `/*parameterName=*/` comment so we know what they're meant to be? Also, are they required arguments?

(Note: this isn't at all blocking, feel free to ignore this suggestion and merge)

krzysz00 (approved):
Great cleanup, approved!

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -690,141 +690,31 @@ static FuncOp createGPUWrapper(ModuleOp &module, const KernelIF &kernel) {
...
-    StringRef allocFuncName = "mgpuMemCopy5DFloat";
-    if (elemType.isF32()) {
-    } else if (elemType.isF16()) {
-      allocFuncName = "mgpuMemCopy5DHalf";
-    } else if (elemType.isBF16()) {
-      allocFuncName = "mgpuMemCopy5DBF16";
-    } else if (elemType.isInteger(8)) {
-      allocFuncName = "mgpuMemCopy5DInt8";
-    } else if (elemType.isInteger(32)) {
-      allocFuncName = "mgpuMemCopy5DInt32";
```
jerryyin, line 741: If no other use cases, get rid of those `mgpu*` functions?

krzysz00: I'd go make sure of what exactly the `gpu.*` ops lower to

zhanglx13: @jerryyin I'll make sure they can be safely removed and do it in a separate pr if that sounds good to you?

mlir/tools/miopen-gen/miopen-gen.cpp
zhanglx13: looks like they are not required if alloc is not called in the async manner.

mlir/tools/miopen-gen/miopen-gen.cpp
sjw36: The gpu.* ops will lower to mgpu* funcs, so they are still needed.

sjw36: Sorry, the 5D funcs are not needed..

sjw36 (approved):
Looks good!

PR #674, opened 2022-06-28
Define a shell script for building our Dockerfile

krzysz00:
Currently, the Dockerfile build is inline in a Jenkins job config,
which makes it impossible to track changes and. Move the logic from
that build pipeline (with minor refactoring) to a script that the
Jenkins job can call.

This PR will have no effect until the rocm-docker job on ml-ci.amd.com
is changed to use the new script.

### DIFF
--- a/mlir/utils/jenkins/build-dockerfile.sh
+++ b/mlir/utils/jenkins/build-dockerfile.sh
@@ -2,13 +2,15 @@
 
 set -x
 
-function err() {
+declare -a ARGS=("$@")
+
+err() {
   echo "$*" >&2
 }
 
 # Clean up the docker images
 # Note: includes a workaround for the occasional hung docker daemon
-function start_clean_docker() {
+start_clean_docker() {
   systemctl status docker | grep 'Active:'
   sudo /usr/bin/pkill -f docker
   sudo /bin/systemctl restart docker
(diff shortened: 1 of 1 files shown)


mlir/utils/jenkins/build-dockerfile.sh
```diff
@@ -0,0 +1,68 @@
...
+  fi
+
+  docker push "${docker_new_img_name}"
+  docker push "${docker_new_img_name_rocm_latest}"
+  docker push "${docker_new_img_name_latest}"
+  docker logout
+}
+
+function main() {
+  if git diff --name-only HEAD^ HEAD | grep -q "Dockerfile.rocm"; then
```
jerryyin: [Nit] I'd prefer to have a jenkins end controllable option to skip the dependency check. This means: When a build failed (and it happened from case to case), we have a way to force another build without having to change the source.

jerryyin (approved):
Looks good. Before we transition to this method, I'd prefer to have the [Nit] addressed so there's more control over the jenkins end.

mlir/utils/jenkins/build-dockerfile.sh
```diff
@@ -0,0 +1,71 @@
...
+
+err() {
+  echo "$*" >&2
+}
+
+# Clean up the docker images
+# Note: includes a workaround for the occasional hung docker daemon
+start_clean_docker() {
+  systemctl status docker | grep 'Active:'
+  sudo /usr/bin/pkill -f docker
```
manupak, line 15: @krzysz00 what was the intention behind this line ? was it to stop and remove the running containers ?

PR #677, opened 2022-07-06
[DO NOT SQUASH] Add --chipset option to AMDGPUToROCDL

krzysz00:
Because the buffer descriptor structure (the V#) has no backwards-compatibility
guarentees, and since said guarantees have been violated in practice
(see https://github.com/llvm/llvm-project/issues/56323 ), and since
the `targetIsRDNA` attribute isn't something that higher-level clients can set
in general, make the lowering of the amdgpu dialect to rocdl take a --chipset
option.

Note that this option is a string because adding a parser for the Chipset
struct to llvm::cl wasn't working out.


sjw36 (approved):
Looks good.

krzysz00:
https://reviews.llvm.org/D129228 is the upstreaming

PR #678, opened 2022-07-06
[mlir][AMDGPU] Use the correct values for OOB_SELECT on gfx10

krzysz00:
Fixes (implicitly - I'd misidentified the issue) https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/580

### DIFF
--- a/external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
+++ b/external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
@@ -168,13 +168,13 @@ struct RawBufferOpLowering : public ConvertOpToLLVMPattern<GpuOp> {
     // bit 24: Reserved to 1 (RDNA) or 0 (CDNA)
     // bits 25-26: Reserved (0)
     // bit 27: Buffer is non-volatile (CDNA only)
-    // bits 28-29: Out of bounds select (0 = structured, 1 = raw, 2 = none, 3 =
-    // swizzles) RDNA only
+    // bits 28-29: Out of bounds select (0 = structured, 1 = check index, 2 =
+    //  none, 3 = either swizzles or testing against offset field) RDNA only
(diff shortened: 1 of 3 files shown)


sjw36 (approved):
Looks good.

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -174,7 +174,7 @@ struct RawBufferOpLowering : public ConvertOpToLLVMPattern<GpuOp> {
     uint32_t word3 = (7 << 12) | (4 << 15);
     if (chipset.majorVersion == 10) {
       word3 |= (1 << 24);
-      uint32_t oob = adaptor.boundsCheck() ? 1 : 2;
+      uint32_t oob = adaptor.boundsCheck() ? 3 : 2;
       word3 |= (oob << 28);
```
jerryyin, line 178: Just trying to understand, I'm confused on: In above comments you mentioned this out of bounds select is RDNA only, now instead of base this bit on architecture, this is based on whether or not we perform bounds check. Why is that? And how does it impact non-RDNA arch?

krzysz00: RDNA has an `OOB_SELECT` field - 2 = no checks, 3 = `offset + lengthOfOp > numRecords - sgproffset` . CDNA has its oob checks inferred from the parameters to the instruction and no way to outright disable them

jerryyin: If that's the case, could you amend your comments in 171 to incorporate those? Right now the comment still points to `3 = swizzles` which is much less clear compared to your previous PR comment.

jerryyin (approved):

krzysz00:
Will be landing upstream (without further review, but with passing builds). See https://reviews.llvm.org/D129320

PR #679, opened 2022-07-06
[mlir][AMDGPU] Explicitly truncate memory addresses in buffer ops

krzysz00:
As a percaution, truncate memory addresses passed to kernels to 48 bits,
since bits 48-63 of the buffer descriptor are used for the stride field
and, on gfx10, to control swizzling.


krzysz00:
I'd like to poke this along

jerryyin (approved):
LGTM

krzysz00:
Submitted upstream at https://reviews.llvm.org/D131016

PR #682, opened 2022-07-08
Update Jenkinsfile to use rocm5.1-latest image.

pcf000:
Image has been built, and tested by hand.  Let's use it.


krzysz00 (changes requested):
You'll need to get `Jenkinsfile.downstream` as well

And probably also the Buildbot docker?

pcf000:
> You'll need to get `Jenkinsfile.downstream` as well
> And probably also the Buildbot docker?

True, and updated.

krzysz00 (approved):
Thanks for taking the time to do this and test it!

PR #683, opened 2022-07-08
MLIR#569 Export libMLIRMIOpen as searchable by find_package() in static library mode

zhanglx13:
In static library mode, the MLIRMIOpen library is exported and can be found by `find_package()` in MIOpen (or any other projects) when `-DCMAKE_PREFIX_PATH` is set to the installation dir of the static library.
Version checking is performed automatically by the package version file.


mlir/tools/mlir-miopen-lib/CMakeLists.txt
```diff
@@ -97,4 +90,56 @@ if( BUILD_FAT_LIBMLIRMIOPEN )
   install(TARGETS MLIRMIOpen
     ARCHIVE DESTINATION lib)
 
+  # Install Miir.h to ${CMAKE_INSTALL_PREFIX}/include/<project-name>-<version>/
+  # as part of component libMLIRMIOpen
+  set(PACKAGE_DIR MLIRMIOpen-${CMAKE_PROJECT_VERSION})
```
krzysz00: Do we want `CMAKE_PROJECT_VERSION` here? I don't think we're using that field, and I don't want to give us a trailing dash.

mlir/tools/mlir-miopen-lib/CMakeLists.txt
```diff
@@ -97,4 +90,56 @@ if( BUILD_FAT_LIBMLIRMIOPEN )
...
+  install(FILES Miir.h
+    DESTINATION include/${PACKAGE_DIR}
```
krzysz00: Let's keep the `TYPE` in here?

mlir/tools/mlir-miopen-lib/cmake/MLIRMIOpenConfig.cmake
```diff
@@ -0,0 +1,4 @@
+# Auto compute the include dir for the header file
+get_filename_component(SELF_DIR "${CMAKE_CURRENT_LIST_FILE}" PATH)
+include("${SELF_DIR}/MLIRMIOpenTargets.cmake")
+get_filename_component(MLIRMIOpen_INCLUDE_DIR "${SELF_DIR}/../../../include/@PACKAGE_DIR@" ABSOLUTE)
```
krzysz00: What's the last line here for?

krzysz00 (changes requested):

krzysz00 (commented):
I'm curious about a few things here

mlir/tools/mlir-miopen-lib/CMakeLists.txt
zhanglx13: If we keep TYPE then we can only install it to include.
Or do you think we should get rid of the ${PACKAGE_DIR}?

mlir/tools/mlir-miopen-lib/cmake/MLIRMIOpenConfig.cmake
zhanglx13: It sets the include_dir of Miir.h so that MIOpen should be able to include it like
`target_include_directories(MIOpen SYSTEM PRIVATE ${MLIRMIOpen_INCLUDE_DIR}) `

mlir/tools/mlir-miopen-lib/CMakeLists.txt
krzysz00: Ok, agreed on the DESTINATION bit. I do think we shouldn't put the version in `PACKAGE_DIR` though. (and see https://cmake.org/cmake/help/latest/command/install.html and the mention of `CMAKE_INSTALL_INCLUDEDIR`

mlir/tools/mlir-miopen-lib/cmake/MLIRMIOpenConfig.cmake
krzysz00: Is that the usual way other CMake packages handle this?
And shouldn't we be exporting the path to the library as well?

zhanglx13: The exported library is loaded by add_library as
`add_library(MLIR::MLIRMIOpen STATIC IMPORTED)` when find_package() locates and loads the packageConfig.cmake file. The path to the library is also set as a property of the target as `IMPORTED_LOCATION_CONFIG`.

In the config mode of find_package(), no variable is set to hold the include dir. Therefore, we need to set the INCLUDE_DIR manually. 
(I am new to cmake and please correct me if I mis-understood anything)

krzysz00: It looks like https://cmake.org/cmake/help/v3.15/module/CMakePackageConfigHelpers.html might be of interest, so as to make the whole installation setup more general?

mlir/tools/mlir-miopen-lib/CMakeLists.txt
```diff
@@ -97,4 +90,56 @@ if( BUILD_FAT_LIBMLIRMIOPEN )
...
+    install(TARGETS ${target_name}
+      EXPORT ${export_set}
+      ARCHIVE DESTINATION lib/${PACKAGE_DIR}
+      COMPONENT ${component_name} EXCLUDE_FROM_ALL)
+
+    # Generate package version file
+    include(CMakePackageConfigHelpers)
+    write_basic_package_version_file(
+      "${CMAKE_CURRENT_BINARY_DIR}/${pkg_version_file}"
+      VERSION ${CMAKE_PROJECT_VERSION}
```
jerryyin, line 142: What version is it currently? I actually don't know if we have explicitly set this.

mlir/tools/mlir-miopen-lib/CMakeLists.txt
```diff
@@ -97,4 +90,56 @@ if( BUILD_FAT_LIBMLIRMIOPEN )
...
 
+  # Install Miir.h to ${CMAKE_INSTALL_PREFIX}/include/<project-name>-<version>/
+  # as part of component libMLIRMIOpen
+  set(PACKAGE_DIR MLIRMIOpen)
+  install(FILES Miir.h
+    DESTINATION include/${PACKAGE_DIR}
+    COMPONENT libMLIRMIOpen EXCLUDE_FROM_ALL)
+
+  # Install lib${target_name} as part of ${component_name} and export it to be searchable by find_package()
+  function(export_target target_name component_name)
```
jerryyin, line 124: Could you give a summary of what artifacts it install to file system in the comments above? Give each entry a location and a brief description on why the file is needed and what it does?

jerryyin: Actually for the review purpose. Would you mind to actually attach all of those files as a artifact to github? I want to take a look at all exported (text) files.

mlir/tools/mlir-miopen-lib/CMakeLists.txt
zhanglx13: It is set in the top-level CMakeLists.txt
`project(miopen-mlir VERSION 0.202108 LANGUAGES CXX C) `

zhanglx13:
Here are the generated files
[MLIRMIOpen.zip](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/files/9074477/MLIRMIOpen.zip)

krzysz00:
Having looked at the generated files, it looks like there's a reference to `PACKAGE_PREFIX_DIR` that's not defined anywhere.

And I wonder if the checking of `RelWithDebInfo` will cause issues

zhanglx13:
The `PACKAGE_PREFIX_DIR` is set at the beginning of MLIRMIOpenConfig.cmake
The MLIRMIOpenTargets-CONFIG.cmake will be generated when MLIR is built with `-DCMAKE_BUILD_TYPE=CONFIG`.
And MIOpen will look for it if MIOpen is also built with `-DCMAKE_BUILD_TYPE=CONFIG`.

krzysz00:
... right, I'm very good at that noticing that we did set that variable.

In any case, this all seems like it should work, but the only way to really find out would be to try building MIOpen with it.

One other note - MIOpen has our directory as part of the `#include` directive, and so we don't need to include it in setting up the path (last I checked)

zhanglx13:
> One other note - MIOpen has our directory as part of the `#include` directive, and so we don't need to include it in setting up the path (last I checked)

You mean our installation dir? And we do not need to setup the header path for include?

krzysz00:
Ok, actually, I take that back, I'd misremembered. We need to be in the path because the include is https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/conv/invokers/mlir_impl_gemm.cpp#L35 , and so if we're installing to a subdirectory, we need to set that up.

And given that we previously, last I checked, weren't installing to a subdirectory, this PR is going to have CI fail until we amend MIOpen to use the CMake - so this PR will need one of those "temporarily change with branch of the MIOpen repo we point at" changes.

zhanglx13:
Yes. @jerryyin told me to tweak the CI to use a temp branch of MIOpen that can find MLIR using find_package(). We'll see how it goes.

mlir/tools/mlir-miopen-lib/CMakeLists.txt
jerryyin: Could you figure out if `0.202108` is the right version? Where does this number come from to start with?
 - What's the upstream to bump it? and what's the upstream corresponding version now?
 - How often do you think we should bump it? and How should we bump it?
 - Anyway to make it related with the ROCm release version?

mlir/tools/mlir-miopen-lib/CMakeLists.txt
```diff
@@ -97,4 +90,78 @@ if( BUILD_FAT_LIBMLIRMIOPEN )
...
+  # Install Miir.h to ${CMAKE_INSTALL_PREFIX}/include/<project-name>-<version>/
+  # as part of component libMLIRMIOpen
+  set(PACKAGE_DIR MLIRMIOpen)
+  install(FILES Miir.h
+    DESTINATION include/${PACKAGE_DIR}
+    COMPONENT libMLIRMIOpen EXCLUDE_FROM_ALL)
+
+  # Install lib${target_name} as part of ${component_name} and
+  # export it to be searchable by find_package()
+  #
```
jerryyin, line 102: Naming conventions below looks strange:
 - The generated files in this PR use \<package\>Targets.cmake
 - On the other hand, regular naming convention come from /opt/rocm/lib follows \<package\>-targets.cmake

Could you find out the recommended we and let us know?

mlir/tools/mlir-miopen-lib/CMakeLists.txt
```diff
@@ -97,4 +90,78 @@ if( BUILD_FAT_LIBMLIRMIOPEN )
   install(TARGETS MLIRMIOpen
     ARCHIVE DESTINATION lib)
 
+  # Install Miir.h to ${CMAKE_INSTALL_PREFIX}/include/<project-name>-<version>/
+  # as part of component libMLIRMIOpen
+  set(PACKAGE_DIR MLIRMIOpen)
```
jerryyin: Not fully convinced on the naming of the package here. Could you double check with @sjw36 what the correct short hand package name is? I remember it is somewhat `Miir`?

mlir/tools/mlir-miopen-lib/CMakeLists.txt
```diff
@@ -97,4 +90,78 @@ if( BUILD_FAT_LIBMLIRMIOPEN )
...
+  #
+  # As an example, when CMAKE_INSTALL_PREFIX is set to /a/dummy/ and
+  # CMAKE_BUILD_TYPE is set to RelWithDebInfo
+  # The following files is generated and installed for MLIRMIOpen
+  # /a/dummy/lib/MLIRMIOpen/libMLIRMIOpen.a
+  # /a/dummy/lib/cmake/MLIRMIOpen/MLIRMIOpenTargets.cmake
+  #   This file is autogenerated by install(EXPORT).
+  #   It creates the imported library as a target that can be used by
+  #   target_link_libraries()
+  # /a/dummy/lib/cmake/MLIRMIOpen/MLIRMIOpenTargets-relwithdebinfo.cmake
```
jerryyin: If a client build with release target then this would generate a `<package>-release.cmake`, is that right?

mlir/tools/mlir-miopen-lib/CMakeLists.txt
```diff
@@ -97,4 +90,78 @@ if( BUILD_FAT_LIBMLIRMIOPEN )
...
+  # export it to be searchable by find_package()
+  #
+  # As an example, when CMAKE_INSTALL_PREFIX is set to /a/dummy/ and
+  # CMAKE_BUILD_TYPE is set to RelWithDebInfo
+  # The following files is generated and installed for MLIRMIOpen
+  # /a/dummy/lib/MLIRMIOpen/libMLIRMIOpen.a
+  # /a/dummy/lib/cmake/MLIRMIOpen/MLIRMIOpenTargets.cmake
+  #   This file is autogenerated by install(EXPORT).
+  #   It creates the imported library as a target that can be used by
+  #   target_link_libraries()
```
jerryyin: This file claims the minimum cmake version is 2.6.0. I think we require a higher version than this. Could you double check on how it get populated?

jerryyin (commented):
Good work, and sorry I mostly have questions/nitpicks comparing the populate cmake with existing ones from `/opt/rocm`. I want to make sure this exported static lib target is done right.

Also CC @atamazov in case he has any feedback to this review.

mlir/tools/mlir-miopen-lib/CMakeLists.txt
zhanglx13: Yes

mlir/tools/mlir-miopen-lib/CMakeLists.txt
krzysz00: That's probably specifically the minimum version of CMake needed for the module to work, which is separate from what we need to build our project (which gets checked elsewhere)

mlir/tools/mlir-miopen-lib/CMakeLists.txt
krzysz00: We could probably go with `Miir` to match the header, but I haven't a clue where we got that name

krzysz00:
@jerryyin the version number is set in `project(miopen-mlir VERSION 0.202108 LANGUAGES CXX C)` in CMakeLists.txt in our repo. It can be whatever the heck we want it to be, and I'd advocate for it to be the ROCm release number - that is, we'd change it to 5.4 in the dev branch, then change it to 5.5 after branching, and so on.

mlir/tools/mlir-miopen-lib/CMakeLists.txt
jerryyin: That sounds fair, I do find this `cmake_minimum_required` in our project root that specify it to be 3.15.1.

jerryyin:
> and I'd advocate for it to be the ROCm release number

I think that's a good idea. However, having looked at how other components of ROCm setup their version, this seems to be a number irrelevant from ROCm's release version. I'm not sure whether we want to bump our version number all the way to 5.3, for example. (What if someone ask us to provide to 5.2 version of our project?)

zhanglx13:
@jerryyin The existing version number (0.202108) was created by @krzysz00 on Aug 2021 and that was how he chose the number. I feel like we do not have our own versioning mechanism since we branch in align with the ROCm release numbers. So we either need to figure out our own versioning system showing that we are (or intend to be) decoupled with ROCm or we can go with the ROCm version number showing that we are coupled with ROCm.

jerryyin:
@mvermeulen @whchung Could you give us some guidance on the project versioning? Is it okay for us to re-use rocm versioning, or shall we start from 1.0?

mlir/tools/mlir-miopen-lib/CMakeLists.txt
zhanglx13: I do not see any preference between `<package>Targets.cmake` and `<package>-targets.cmake`. The same goes for `<package>Config.cmake` and `<package>-config.cmake`. I saw both are used among components under `/opt/rocm/lib/cmake`. 
From find_package()'s perspective, it does not have a preference between finding `<package>Config.cmake` and `<package>-config.cmake`. So I think both should be fine. Let me know if you have other concerns.

zhanglx13: The only thing I found is [this doc](https://docs.salome-platform.org/latest/dev/cmake/html/config.html#) about configuration file conventions. It uses the naming that I used without explaining why it is used. Apparently the focus of this doc is not about the naming. 
Again, I think it should not matter.

jerryyin: Sounds good to me, thanks for looking it up!

mlir/tools/mlir-miopen-lib/CMakeLists.txt
zhanglx13: So I changed the package name to Miir. 
And MIOpen now can find libMLIRMIOpen.a by `find_package(Miir)`.

mvermeulen:
A few of my thoughts on versioning.

1. We should definitely use a different version number every time things change.
2. Default for ROCm releases is to tag a version number including the ROCm release, e.g. 5.3, 5.4.  Some things get tagged with minor releases e.g. 5.2.1 even when they haven't changed but this applies more to packaging or similar things where one needs to rebuild/redeliver.  That is preferable to using a date based version we currently use.
3. I don't have a problem with jumping from a 0.x number all the way to 5.3.
4. I don't know what versions the compiler team uses; but assume this is separate from topic of commit hashes - and a different but useful topic.

krzysz00 (approved):

PR #686, opened 2022-07-08
Fix buildbot problems with an upstream CMakeLists.txt patch and our own miopen-deps patch [DO NOT SQUASH]

pcf000:
One simple change is to install cmake without limiting it to 3.15.  This is a workaround for a problem introduced upstream when a commit used features that first appear in 3.18 or later.  They're working on retrofitting it, but for now let's cope by using a newer cmake.

The rest of the changes in Dockerfile remove a redundant installation of python3.8 and fix some library issues by removing /usr/local/bin/python3.  The rocm/mlir Dockerfile installs python3.8 in /usr/bin, but then the section that installs MIOpen dependences installs it again, in /usr/local because of --prefix.  When building MLIR, cmake's find_package finds python3 in /usr/bin even though it's not the first one in $PATH, and thus we have a python3 that doesn't have the libraries that were installed earlier.  Let's flush /usr/local/bin/python3 because we already have the same thing in /usr/bin.  Yeah, it's ugly, but the alternative is "--prefix /usr" and I like that less.

Finally, add --time-tests to the MLIR regression tests so we might spot outliers in CI.  Optional.

### DIFF
--- a/mlir/utils/buildbot/mlir-rocm-mi200/Dockerfile
+++ b/mlir/utils/buildbot/mlir-rocm-mi200/Dockerfile
@@ -15,25 +15,18 @@
 FROM rocm/mlir:rocm5.1-latest
 
 RUN apt-get update; \
-    apt-get install -y dumb-init; \
+    apt-get install -y dumb-init cmake; \
     update-alternatives --install /usr/bin/clang clang /usr/bin/clang-10 100 ;\
     update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-10 100 ;\
     update-alternatives --install /usr/bin/lld lld /usr/bin/lld-10 100
 
 # LTS releases often bundle obsolete pip versions that cannot access newest
 # Linux binary wheels.
-RUN python3 -m pip install --upgrade pip
-
-RUN apt-add-repository ppa:deadsnakes/ppa && apt-get update \
-     && apt-get install -y --no-install-recommends \
-            python3.8 python3.8-dev python3.8-venv \
-     && update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.8 10 \
-     && python3 -m pip install --upgrade pip \
-     && python3 -m pip install --ignore-installed PyYAML
+RUN rm /usr/local/bin/python3 && python3 -m pip install --upgrade pip && python3 -m pip install --ignore-installed PyYAML
 
 # Refer to mlir/lib/Bindings/Python/requirements.txt. Listed explicitly here
 # and version pinned for consistency as this is a bot.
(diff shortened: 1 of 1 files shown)


krzysz00:
I think the actual fix to the python3 thing is to patch MIOpen's `install_deps.cmake` to give it the option to not create a vintualenv and use system Python instead

external/llvm-project/mlir/test/CMakeLists.txt
```diff
@@ -170,6 +170,7 @@ add_custom_target(check-mlir-build-only
 add_lit_testsuite(check-mlir "Running the MLIR regression tests"
   ${CMAKE_CURRENT_BINARY_DIR}
   DEPENDS ${MLIR_TEST_DEPENDS}
+  ARGS "--time-tests"
```
krzysz00: This doesn't match how we handled `--time-tests` in CI - namely, by adding to the lit argumentss.

We could probably handle both these issues by removing our special-purpose `--time-tests` support from `mlir/test/CMakeLists.txt` and then setting `LIT_OPTS=--time-tests` in the environment

mlir/utils/buildbot/mlir-rocm-mi200/Dockerfile
```diff
@@ -15,25 +15,18 @@
...
     update-alternatives --install /usr/bin/lld lld /usr/bin/lld-10 100
 
 # LTS releases often bundle obsolete pip versions that cannot access newest
 # Linux binary wheels.
-RUN python3 -m pip install --upgrade pip
-
-RUN apt-add-repository ppa:deadsnakes/ppa && apt-get update \
-     && apt-get install -y --no-install-recommends \
-            python3.8 python3.8-dev python3.8-venv \
-     && update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.8 10 \
```
krzysz00, line 30: I think we might still need the `update-alternatives` call to set system `/usr/bin/python3`

external/llvm-project/mlir/test/CMakeLists.txt
pcf000: I tried to parallel the other set of tests, just to see the results.  I'll try your suggestion.

mlir/utils/buildbot/mlir-rocm-mi200/Dockerfile
pcf000: Already done in mlir/utils/jenkins/Dockerfile, which has a similar section.

krzysz00:
@pcf000 I'm willing to approve after the time-tests bits are moved elsewhere

pcf000:
@krzysz00 , we're on the same mental schedule -- I was just signing in to split off the time-test part.  Using LIT_OPTS did work, by the way.

krzysz00 (approved):

PR #688, opened 2022-07-09
Use set() instead of list() to append rpath flags

linedot:
As described in https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/issues/687 - in `cmake/llvm-project.cmake rpath` flags are appended to `CMAKE_EXE_LINKER_FLAGS` with `list(APPEND ...`. This inserts a ";" between the previous flags (if present) and the newly appended flags, breaking compilation. This can be avoided by using `set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ..."` to append the flags instead.

### DIFF
--- a/cmake/llvm-project.cmake
+++ b/cmake/llvm-project.cmake
@@ -48,8 +48,6 @@ list(APPEND LLVM_INCLUDE_DIRS
 )
 
 # Linker flags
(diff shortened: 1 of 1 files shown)


krzysz00 (approved):
Approved, thank you!

PR #689, opened 2022-07-11
Consolidate mfma loop in reduction/non-reduction

jerryyin:
I made the following changes in this PR:
 - Consolidated the gemm calculation between reduction and non-reduction cases
   - Both share the same implementation as the register structure from lds -> register are the same
   - Added `KPerThread` definition to share K dimension calculation between the two
 - Added test coverage in `lowering_xdlops_gemm_v2.mlir`
 - Misc cleanups:
   - Removed redundant `emitLoadLogic`, and use `InBoundsLoadOp` instead

### DIFF
--- a/mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
+++ b/mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
@@ -57,14 +57,6 @@ struct MIOpenThreadwiseGemmLoweringPass
 };
 
 //===----------------------------------------------------------------------===//
-// Utility function to emit load instructions for local buffers
-//===----------------------------------------------------------------------===//
-Value emitLoadLogic(OpBuilder &b, Location loc, Type loadedType,
-                    const Value source, ValueRange coords) {
-  return b.create<InBoundsLoadOp>(loc, loadedType, source, coords);
-}
-
-//===----------------------------------------------------------------------===//
 // ThreadwiseGemm lowering.
 //===----------------------------------------------------------------------===//
 struct ThreadwiseGemmRewritePattern
@@ -238,14 +230,20 @@ struct XdlopsGemmV2RewritePattern : public OpConversionPattern<XdlopsGemmV2Op> {
 
     bool IsKReduction = (num_output_blks == 1) && (num_input_blks > 1);
 
-    // Original C++ logic.
+    if (KPack > 1 && (KPack < k_base || KPack % k_base != 0)) {
+      llvm_unreachable(
+          "Tuning parameter selection guarantees kPack is multiple of k_base,"
+          "this should never happen");
+    }
+
     // const index_t laneId = get_thread_local_1d_id() % mfma_type.wave_size;
-    // FloatA a[K * MRepeats];
-    // FloatB b[K * NRepeats];
-    // constexpr index_t KRepeats = sizeof(FloatA) / (sizeof(data_type) *
-    // mfma_type.k_base); auto pa = reinterpret_cast<const data_type*>(&a); auto
-    // pb = reinterpret_cast<const data_type*>(&b); constexpr index_t AStride =
-    // K * KRepeats; constexpr index_t BStride = K * KRepeats;
+    // FloatA a[KPerThread * MRepeats];
+    // FloatB b[KPerThread * NRepeats];
+    // constexpr index_t KRepeats = KPack / mfma_type.k_base;
+    // auto pa = reinterpret_cast<const data_type*>(&a);
+    // auto pb = reinterpret_cast<const data_type*>(&b);
+    // constexpr index_t AStride = KPerThread * KRepeats;
+    // constexpr index_t BStride = KPerThread * KRepeats;
 
     auto tid = b.create<WorkitemIdOp>(loc, b.getIndexType());
     constexpr int64_t waveSize = 64;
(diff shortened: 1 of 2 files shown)


mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -274,31 +272,27 @@ struct XdlopsGemmV2RewritePattern : public OpConversionPattern<XdlopsGemmV2Op> {
...
+      // store bufferA logic.
+      // for(index_t m_i = 0; m_i < MRepeats; ++m_i)
+      //   for(index_t k_i      = 0; k_i < K; ++k_i)
+      //     a[k_i + m_i * K] = p_a_wave[k_i * M + laneId + MPerXdlops * m_i];
+      // Note: p_a_wave need to be offseted by waveOffsetA.
 
       auto outerLoopM = b.create<AffineForOp>(loc, 0, MRepeats);
-      auto olmb = ConversionPatternRewriter::atBlockTerminator(
-          outerLoopM.getBody(), b.getListener());
+      auto olmb = OpBuilder::atBlockBegin(outerLoopM.getBody());
```
krzysz00: We need `ConversionPatternRewriter` (since the object above is a `ConversionPatternRewriter`) and we need to pass the listener to make the `--debug` outputting of rewrites work

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -274,31 +272,27 @@ struct XdlopsGemmV2RewritePattern : public OpConversionPattern<XdlopsGemmV2Op> {
...
       auto olmiv = outerLoopM.getInductionVar();
       auto mOffset = olmb.create<AddIOp>(
           loc, aBase, olmb.create<MulIOp>(loc, MPerXdlopsConstantOp, olmiv));
       auto kOffsetA = olmb.create<MulIOp>(loc, olmiv, KConstantOp);
 
-      auto innerLoopMK = olmb.create<AffineForOp>(loc, 0, K);
-      auto ilmkb = ConversionPatternRewriter::atBlockTerminator(
-          innerLoopMK.getBody(), olmb.getListener());
+      auto innerLoopMK = olmb.create<AffineForOp>(loc, 0, KPerThread);
+      auto ilmkb = OpBuilder::atBlockBegin(innerLoopMK.getBody());
```
krzysz00: Same re builder

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -311,40 +305,29 @@ struct XdlopsGemmV2RewritePattern : public OpConversionPattern<XdlopsGemmV2Op> {
...
+      // for(index_t n_i = 0; n_i < NRepeats; ++n_i)
+      //   for(index_t k_i      = 0; k_i < KPerThread; ++k_i)
+      //     b[k_i + n_i * KPerThread] = p_b_wave[k_i * N + laneId + NPerXdlops
+      //     * n_i];
+      // Note: p_b_wave need to be offseted by waveOffsetB.
 
       auto outerLoopN = b.create<AffineForOp>(loc, 0, NRepeats);
-      auto olnb = ConversionPatternRewriter::atBlockTerminator(
-          outerLoopN.getBody(), b.getListener());
+      auto olnb = OpBuilder::atBlockBegin(outerLoopN.getBody());
```
krzysz00: And so on

krzysz00 (commented):
The "simplification" of the loop builders will make things annoying in the future, otherwise ... if it works, it works, and this is much simpler!

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
jerryyin: First Just trying to clarify here: Why does debug not work using `OpBuilder`? Does it not populate debug messages in the `LLVM_DEBUG`? Wants to understand the symptom to it so I can identify this better.

If this indeed break debug in certain way, I'd still prefer `OpBuilder::atBlockBegin(outerLoopM.getBody(), b.getListener());`, because `atBlockBegin` is a child class function that eventually fall back on `OpBuilder`. Since it is a static function anyway, why not use the parent `OpBuilder` version of this?

krzysz00: Ok, so the situation is that pattern rewriting itself has debug messages - these are handled through the listener methods on the rewriter. Creating new builders/rewriters without preserving the listener field breaks debug.

Second, I want to use the child class's version both in case it does something fancy, so we don't erase the actual type of the child, and so we have access to the conversion-rewriter methods in their context.

I agree with `atBlockBegin`, but I want that to be `ConversionPatternRewriter::atBlockBegin(..., b.getListener())`

And lastly, eventually, we should move to insertion guards and just using one builder object, but that might not be a refactor for this PR. (Last I checked, builder construction is not guaranteed to be cheap and so setting the insertion point on an existing builder is prefered)

jerryyin: > these are handled through the listener methods on the rewriter

Good to know. I didn't notice this before

> I agree with atBlockBegin

Sounds good.

> Second, I want to use the child class's version both in case it does something fancy

I was not accurate in previous comment. Actually, this `ConversionPatternRewriter` does not reroute a implementation of `atBlockBegin()` at all, and it inherits from the `OpBuilder` interface. Searching through the entire code base I don't find another instance where some consumer would ask for `atBlockBegin()` from a rewriter, and all instances of its use are from the `OpBuilder`.

At this point I don't think this matters to invoke either - it is likely by design insertion point setting and getting are from `OpBuilder` (and rewriter's responsibility is orthogonal from it) . But I'd prefer `OpBuilder`'s version to be consistent with the rest of the code base.

krzysz00: The rest of the code base got rather systematically adjusted to use `ConversionPatternRewriter::` in places where there's indeed a `ConversionPatternRewriter`, and I'd like to not have the static type of the object we're using for building get changed out from under us. I strongly prefer that, if we start with a `ConversionPatternRewriter` as the argument to our `matchAndRewrite`, we stick with a `ConversionPatternRewriter` throughout (except in utility methods that can downgrade, but then C++ is doing the right thing and adding vtables and so on).

krzysz00: That is, when I was turning this into conversion patterns, I had some very weird bugs arise from using a `ConversionPatternRewriter` to create an `OpBuilder`, and so I want to keep that consistency.

jerryyin: Sounds good! Thanks for patiently walking me through this.

krzysz00: No problem, thanks for doing this massive cleanup!

krzysz00 (approved):
LGTM, approved

PR #690, opened 2022-07-11
Enable using bare pointers for GPU kernels with static shape, and then use that support

krzysz00:
The changes needed to make this work upstream are at https://reviews.llvm.org/D130716 and have landed - a copy of that commit is on this branch

### DIFF
--- a/mlir/include/mlir-c/Dialect/MIGraphX.h
+++ b/mlir/include/mlir-c/Dialect/MIGraphX.h
@@ -18,6 +18,10 @@
 extern "C" {
 #endif
 
+// Version 2: Use bare pointer ABI (kernels take just a pointer to the data
+// buffer, not an entire memref struct). Also introduces this constant.
+#define MLIR_MIGRAPHX_DIALECT_API_VERSION 2
+
 MLIR_DECLARE_CAPI_DIALECT_REGISTRATION(MIGraphX, migraphx);
 
 // Phase 0 functions : Assuming the given module contains only one function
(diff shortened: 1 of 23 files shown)


sjw36 (approved):
Looks good. Thanks for doing that.

krzysz00:
Update: looks like we need to poke at how MIOpen calls us...

krzysz00:
We're now `skip-ci` because there's no point in wasting CI resources while we're blocked on working out exactly how to change MIOpen.

krzysz00:
@pcf000 , can you help me navigate making a MIGraphX PR that would accomodate the ABI change here?

@jerryyin , is there anything in particular we need to do over in MIOpen?

krzysz00:
(Update, @jungpark-mlir  might be the person to tag in for MIGraphX? )

jungpark-mlir:
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/7dcae03776938bf983ef6184a13bb9bcada678b5/src/targets/gpu/mlir.cpp#L623
This is where MIGraphX puts the memref descriptor fields into the kernel arguments.

krzysz00:
Blocked on @zhanglx13 getting his changes to finding MLIR into MIOpen, since we don't want both those changes to be one MIOpen PR https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1673

mlir/include/mlir-c/Dialect/MIGraphX.h
```diff
@@ -18,6 +18,8 @@
 extern "C" {
 #endif
 
+#define MLIR_MIGRAPHX_DIALECT_API_VERSION 2
+
```
jungpark-mlir, line 24: Can we have a comment here to keep the features added/deleted per each version?

jungpark-mlir (approved):

PR #691, opened 2022-07-11
Infer vectorization from maps for xdlops writeback

krzysz00:
This adds a vectorization inference utility to our system and uses it in the writeback loop in xdlops gemm.

(Doing this to nonxdlops gemm is a separate PR)


krzysz00:
Manual nightly pass (and now some ops have a vectorization length of 8 - that is, we allow writeback to be two `buffer_store` intrinsics to the same base address with a constant offset)

jungpark-mlir:
Out of curiosity, does it create single buffer store for the vec8 of 16bit types? That'd be really nice.

krzysz00:
... no, I should fix that, actually

krzysz00:
(mind, it only doesn't do that for very trivial reasons, namely that I forgot that the maximum store length isn't 4, it's (4 * (32 / bitwidth)))

krzysz00:
FYI this needs a merge/rebase to clear off clang-format

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -400,19 +402,43 @@ struct ThreadwiseCopyV2RewritePattern
...
-                                        op.rightOobDims(), op.destCoord(),
-                                        op.storeMethodAttr());
+    Type sourceElemType = sourceType.getElementType();
+    Type destElemType = op.dest().getType().cast<MemRefType>().getElementType();
+    int64_t elemsPerWord = (32 / destElemType.getIntOrFloatBitWidth());
+    int64_t maxWriteLen = 4 * elemsPerWord;
+    int64_t remainingLength = op.length().getSExtValue();
+    int64_t offset = 0;
+    while (remainingLength > 0) {
+      int64_t copyLength = std::min(remainingLength, maxWriteLen);
```
jerryyin, line 645: IR readability perspective... Is there a way to populate affine and then give it a mark to force it to unroll?

Feel free to ignore as this may not make sense. I'm hoping our populate IR at threadwise level doesn't get flooded by load/stores.

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -400,19 +402,43 @@ struct ThreadwiseCopyV2RewritePattern
...
-
-    Value loaded =
-        b.create<InBoundsLoadOp>(loc, typeToLoad, source, sourceCoord);
-    b.replaceOpWithNewOp<BufferStoreOp>(op, loaded, op.dest(), op.leftOobDims(),
-                                        op.rightOobDims(), op.destCoord(),
-                                        op.storeMethodAttr());
+    Type sourceElemType = sourceType.getElementType();
+    Type destElemType = op.dest().getType().cast<MemRefType>().getElementType();
+    int64_t elemsPerWord = (32 / destElemType.getIntOrFloatBitWidth());
+    int64_t maxWriteLen = 4 * elemsPerWord;
```
jerryyin, line 641: I've seen very similar logic somewhere and also the hard-coded 4 x word size. Sometime in the future we need to clean it up by defining the max load/store write in some header file.

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
```diff
@@ -0,0 +1,215 @@
+//===- transformMapUtils.cpp - transform map utilities --------------===//
```
jerryyin, line 1: It's a bit hard to see if you have updated any implementations while you are moving things around around this commit. Have you?

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
```diff
@@ -213,3 +225,171 @@ TransformOp mlir::miopen::reshapeBuffer(OpBuilder &b, Location loc,
...
+                      .getMap()
+                      .getValue()
+                      .getNumInputs());
+  maxLengths[dim] = len;
+  for (auto transformMap : transforms.getAsRange<TransformMapAttr>()) {
+    maxLengths = propagateVectorizationInfo(transformMap, maxLengths);
+  }
+
+  int64_t result = 1;
+  int64_t stride = 1;
```
jerryyin: Could we get a better name for this. 
 - Maybe `maxAccumulatedStride` for `result`?
 - `accumulatedStride` for `stride`?

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
```diff
@@ -213,3 +225,171 @@ TransformOp mlir::miopen::reshapeBuffer(OpBuilder &b, Location loc,
...
+  for (size_t i = 0, e = outputShape.size(); i < e; ++i) {
+    uint32_t dimension = (e - 1) - i;
+    int64_t length = outputShape[dimension];
+    if (maxLengths[dimension].hasValue()) {
+      int64_t maxLength = maxLengths[dimension].getValue();
+      if (result == stride) {
+        result *= maxLength;
+      } else {
+        break;
+      }
```
jerryyin: I feel like if we could instead check maxLength of a dimension is smaller than actual dimension length, would be a more straightforward logic. Then stride would just naturally be the longest possible vectorization size.

So I'd prefer the "embed" version of algorithm than this one.

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
```diff
@@ -0,0 +1,413 @@
...
+
+    switch (transform.getType()) {
+    case TransformType::PassThrough:
+    case TransformType::AddDim:
+    case TransformType::Slice:
+      for (auto pair : llvm::zip(upperDims, lowerDims)) {
+        result[std::get<1>(pair)] = input[std::get<0>(pair)];
+      }
+      break;
+      // Padding limits the max length to gcd(length, length - padding)
```
krzysz00: Needs to account for both l and r separately

krzysz00: The goal is to ensure that a vector chunk is always fully in bounds or fully out of bounds.

krzysz00: Shrinking until issue is avoided

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
```diff
@@ -0,0 +1,413 @@
...
+        std::tie(upper, lower) = data.value();
+        Optional<int64_t> upperLen = input[upper];
+        if (upperLen.hasValue()) {
+          int64_t lowerLen = (*upperLen) - leftPad - rightPad;
+          int64_t newMaxVectorization = math_util::gcd(*upperLen, lowerLen);
+          result[lower] = newMaxVectorization;
+        }
+      }
+      break;
+    case TransformType::Broadcast:
```
krzysz00, line 388: Same sort of boundary-crossing issue as padding

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
```diff
@@ -0,0 +1,413 @@
...
+      result[lowerDims[0]] = maxLength;
+      break;
+    }
+    // For merges, the input vectorization length is split among the output
+    // dimensions, with a dimension getting a vectorization length equal
+    // to the gcd of the remaining vectorization length and that dimension's
+    // length. The remaining length gets lowered by the full lenght of the
+    // output dimension so that Merge{3, 4} with a starting vectorization lenght
+    // of 6 gets the results [1, 2] and not [3, 2].
+    case TransformType::Merge:
```
krzysz00, line 451: Take NCHW and merge M <- W x H x N ?

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
```diff
@@ -213,3 +225,171 @@ TransformOp mlir::miopen::reshapeBuffer(OpBuilder &b, Location loc,
...
+        }
+      }
+      break;
+    // The embed rule: as we walk from smaller to larger coefficients, we
+    // accumulate the vectorization coefficient, `maxLength`, initially 1. If a
+    // dimension's coefficient in the embedding is equal to `maxLength`, we
+    // multiply `maxLength` by that dimension's vectorization length and
+    // continue. During this processes, dimensions that have vectorization
+    // information equal to None (those held constant) are ignored entirely.
+    case TransformType::Embed: {
```
jerryyin, line 407: Can we consolidate all 3 scenarios into a function:
 - embed
 - unmerge
 - derive steps inside of getMaxVectorization()

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
jerryyin: I'm talking about commit 83b5e6d204f9476da092f7660b03409e30460403

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
```diff
@@ -0,0 +1,413 @@
...
+  result.grow(map.getMap().getValue().getNumResults() - 1);
+  for (TransformAttr transform : map.getOps()) {
+    ArrayRef<uint32_t> upperDims = transform.getUpperDims();
+    ArrayRef<uint32_t> lowerDims = transform.getLowerDims();
+    ArrayRef<int64_t> params = transform.getParams();
+
+    switch (transform.getType()) {
+    case TransformType::PassThrough:
+    case TransformType::AddDim:
+    case TransformType::Slice:
```
jerryyin: For slice, if the upper dimension is 16 x 16 with vectorization size of 4 and lowering dimension is 15 x 15, can lower dimension continue to vectorize at length of 4? I'm assuming not because it crossed the boundary between.

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: There isn't a general way to mark unroll (though `transforming_for` can do it). And this'll be 2, maybe 4, stores at most if it comes to it

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
krzysz00: No, that's an nfc move - I just copied stuff

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
krzysz00: The problem there is vectorization probes where the final dimension was held constant, for example. 
That is, suppose we have NHWK output, and we want to vectorize the GemmN -> (N x H x W) dimension.
We can't do this unless K = 1, but your simplied logic would allow vectors in W.

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
krzysz00: I managed to pull in two of the three cases

krzysz00:
@jerryyin I've checked that nightly passes with my recent changes

Do you think I've added enough tests?

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
```diff
@@ -252,34 +297,39 @@ using VectorizationData = llvm::IndexedMap<Optional<int64_t>>;
...
   for (auto pair : dimAndLength) {
     uint32_t upperDim = std::get<0>(pair);
     int64_t dimLength = std::get<1>(pair);
     if (input[upperDim].hasValue()) {
-      if (!maxVectorizationLength.hasValue())
-        maxVectorizationLength = 1;
-      if (maxVectorizationLength == previousDimsStride) {
-        maxVectorizationLength =
-            (*maxVectorizationLength) * input[upperDim].getValue();
+      VectorizationInfo upperInfo = *input[upperDim];
```
jerryyin: I'm afraid it is not safe to directly reference a element from index without checking using `inBounds()` method of indexedMap. Otherwise it'd hit onto an assert [here](https://llvm.org/doxygen/IndexedMap_8h_source.html).

My question is: why would you use `indexedMap` to hold two integers? And how does the two integers get translated to the actual indexes in the map? It sounds to me that if not handling appropriately, we can implicitly be allocating a very large `SmallVector` without knowing it.

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
```diff
@@ -302,7 +352,10 @@ propagateVectorizationInfo(TransformMapAttr map,
         uint32_t upper, lower;
         std::tie(upper, lower) = data.value();
         if (input[upper].hasValue()) {
-          result[lower] = math_util::gcd(*input[upper], sliceEnd - sliceBegin);
+          int64_t maxLength =
```
jerryyin, line 355: Just want to make sure: when we slice, do we always slice equivalent length from left and right?

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
```diff
@@ -334,47 +391,52 @@ propagateVectorizationInfo(TransformMapAttr map,
...
         if (input[upper].hasValue()) {
-          result[lower] = math_util::gcd(*input[upper], modulus);
+          int64_t lowerMaxLen =
+              math_util::gcd(input[upper]->maxLength, modulus);
+          result[lower] =
+              VectorizationInfo(lowerMaxLen, input[upper]->needsCoefficient);
         }
       }
       break;
     // The embed rule: as we walk from smaller to larger coefficients, we
```
jerryyin, line 401: For clarifications, does coefficients means dimension size or dimension index? I think the rule is correct when it means index.

Also want to double confirm that `params` is referenced by dimension index such that `std::sort()` does the right thing?

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
```diff
@@ -334,47 +391,52 @@ propagateVectorizationInfo(TransformMapAttr map,
         int64_t modulus;
         std::tie(upper, lower, modulus) = data;
```
jerryyin, line 392: Could you remind me what `modulus` or `params` is?

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
```diff
@@ -334,47 +391,52 @@ propagateVectorizationInfo(TransformMapAttr map,
...
-      Optional<int64_t> maxLength = propagateUnmergeVectorization(
+    case TransformType::Unmerge:
+      result[lowerDims[0]] = propagateUnmergeVectorization(
           llvm::zip(llvm::reverse(upperDims), llvm::reverse(params)), input);
-      result[lowerDims[0]] = maxLength;
       break;
-    }
     // For merges, the input vectorization length is split among the output
     // dimensions, with a dimension getting a vectorization length equal
     // to the gcd of the remaining vectorization length and that dimension's
```
jerryyin, line 445: Sorry I'm not sure I get it for the merge rule: If the starting vectorization len is 6 and the output dimensions get split to `{3, 4}`, shouldn't the vectorization length be `{1, 4}`?
 - For the 4: since 6 > 4, the last split dimension is split to be fully vectorized
 - For the 1: 6 / 4 = 1, there's not enough vectorization length carried over to the next dimension

jerryyin (commented):
Left mostly questions, I still feel myself not familiar with coordinate transfer infrastructure enough to be able to review to the level of detail I'd like.

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
krzysz00: We won't hit that assert because of the call to `grow()` at the top.

And I'm using `IndexedMap` to be able to have a nice interface around `SmellVector<Option<VectorizationInfo>>` so that I don't have to manually write out the grow() method - why not use the type that's already there?

I'd be willing to go in for some other small dense map primitive, but I figure the space costs of summoning up some `None`s are much less than the bother of having to go through hash functions.

`IndexedMap` is a map from the dimension to the two integers we're storing as `VectorizationInfo`

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
krzysz00: Slice is specified by the beginning and end indices

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
krzysz00: coefficients is neither - it means the coefficient that particular input dimension is multiplied by in order to embed it

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
krzysz00: params is the general parameters to the coordinate transform.

modulus is the `B` in the `d_i -> d_i mod B` that is broadcast - it's always 1 in current applications but we've left in room to generalize it

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
krzysz00: No, we can't make that split

After every 6 moves in the upper dimension, there may be an arbitrary jump. This means that some of those length-4 "vectors" you propose may not be contiguous.

Consider that our data will end up in
```
x x+1 x+2 x+3 x+4 x+5 y y+1 y+2 y+3 ... y+5
```
when merge makes this 3 x 4 we get
```
x x+1 x+2 x+3
x+4 x+5 y y+1
y+2 y+3 y+4 y+5
```

Note that the second row of that is not a valid vector write.

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
jerryyin: I didn't comprehend the index conversion from pair of integer to the smallVector index. How does that work? If there is a pair of (5, 10), does that get converted to 15 or 50? 
 - I think it does matter if we allocate 50 elements for 1 value. 
 - On the other hand, if it get converted to 15, then (10, 5) also get converted to 15 so how does it handle collision?

krzysz00: ... that's not what IndexedMap is

krzysz00: `IndexedMap<T>` == `SmallVector<T>` with the constraint that `T` has a "default" value.

That is, in this case `input[upperDim]` gives you an `Option<VectorizationInfo>` from the underlying array.

The pairs of integers are the _values_, not the keys

jerryyin: Ahhh that's right, I did see `IndexedMap` using `SmallVector` under the hood, but was confused by what that `identity<T>` does. I was assuming it is doing vector version of hashmap (maybe because of there's a "map" in its naming?) and that identity function is doing something smart. Until today when I poked I confirm that index is really just a integer. Sorry for the confusion.

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
jerryyin: Thank you for the explanation.

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
jerryyin: Make sense.

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
jerryyin: Should you also gcd against begin length and end length? Otherwise the vectorization may cross the boundary of sliced chunk.

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
jerryyin: I don't think you handled a corner case, coefficient of 0:
 - In sorting, you'd handle that 0 coefficient dimension first: If that 0 coefficient isn't the lowest dimension, that would reorder the handling sequence to the wrong order
 - When coefficient of 0 does existing on the lowest dimension, it'd act like slicing. So that should use gcd instead of multiplication.

krzysz00: A coefficient of 0 is equivalent to not including the dimension at all ... and I probably should normalize our maps to not allow that

krzysz00: (and I'll note that the resin we sort is that Embed doesn't impose an order on its arguments - unlike Unmerge, which does

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
krzysz00: Well, no, because the map is `(dX) -> (dX + begin)`

Shifting values doesn't affect chunk validity

jerryyin: Does it matter if the begin happen to in middle of the vectorization length? For example, if we are to shift the element by 3, and the vectorization length is 10. It can still vectorize from the 4th elements going forward?

krzysz00: Hm ... 

Considering [a, b, c, d] having been [x, a, b, c, d, y] in the underlying space, you can still vectorize by 4 - you'll be starting to read at `a` and can still go 4 (or 2) or so on at a time. You'll definitely run into `needsCoefficient` problems if you try to recombine this longer dimension with something else - same deal as padding, but the vectorization itself is safe.

jerryyin (approved):
LGTM, appreciate helping me to walk through some implementation details. I'm starting to feel like coordinate transfer infrastructure is worthy of a document to talk about its designs now. I could potentially write a summary in the short future.

PR #695, opened 2022-07-13
Fb option

jungpark-mlir:
Add Affix Tunig Param pass an option to recover from invalid perf_config.
This improves user experience when calling the pass as a part of the pipeline.

This also includes wrapping a call to registerGpuSerializeToHsaSoPass with mutex lock to avoid crash when called by multithreaded MIGraphX

### DIFF
--- a/mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
+++ b/mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
@@ -18,11 +18,13 @@ using namespace mlir;
 using namespace mlir::miopen;
 
 namespace {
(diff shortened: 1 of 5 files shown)


mlir/lib/Dialect/MIOpen/Transforms/AffixTuningParameters.cpp
```diff
@@ -210,7 +213,16 @@ void AffixTuningParameters::affixTuningParametersImpl(T &op) {
         gemmBDerivedParam, gemmCDerivedParam, blockSize, gridSize, gemmKBlocks);
 
     if (failed(status)) {
-      signalPassFailure();
+      // Try again if allowed.
+      if (fallBackNoConfig) {
```
krzysz00, line 217: Nit: test for fallback and for a perf config being specified

krzysz00 (approved):
I see no reason not to do this, approved

PR #696, opened 2022-07-13
Bump f16 tolerance to 25%

zhanglx13:
This PR is a temp fix for some f16 tests that fail with `-rand 1` as
input.

### DIFF
--- a/mlir/tools/miopen-gen/miopen-gen.cpp
+++ b/mlir/tools/miopen-gen/miopen-gen.cpp
@@ -350,6 +350,11 @@ static cl::opt<std::string> randomSide(
              "By default, populate random numbers to both tensors."),
     cl::value_desc("tensor"), cl::init("both"));
 
+static cl::opt<float> f16Threshold(
+    "threshold",
+    cl::desc("Threshold used for CPU verification function for f16 datatype."),
+    cl::value_desc("error"), cl::init(0.25f));
+
 static cl::opt<int> deviceNum(
     "device",
     cl::desc("Device index on which to run the kernel (only with host code)"),
@@ -1276,7 +1281,7 @@ createVerifierFunc(ModuleOp &module, const KernelIF &kernel,
   };
 
   mlir::Value fval00, fval000001, fval001, fval025;
-  float f16_threshold = 0.25f;
+  float f16_threshold = f16Threshold.getValue();
   // Create constants needed for verification
   if ((randomSeed.getValue() != "none" &&
        randomDataType.getValue() == "float") ||
@@ -1373,8 +1378,10 @@ createVerifierFunc(ModuleOp &module, const KernelIF &kernel,
     auto absCpuVal = testBody.create<math::AbsOp>(loc, cpuLoadVal);
 
     mlir::Value maxPercentVal;
-    if (elemType.getIntOrFloatBitWidth() < 32) {
-      maxPercentVal = fval025; // 25%
+    if (elemType.isF16()) {
+      // The threshold for f16 datatype is controlled by -threshold
+      // The default value is set to 0.25
+      maxPercentVal = fval025;
     } else {
       maxPercentVal = fval000001; // 0.0001 %
     }
--- a/mlir/tools/miopen-gen/miopen-gen.cpp
+++ b/mlir/tools/miopen-gen/miopen-gen.cpp
@@ -1275,7 +1275,8 @@ createVerifierFunc(ModuleOp &module, const KernelIF &kernel,
     return b.create<arith::ConstantFloatOp>(loc, apVal, elemFType);
   };
 
-  mlir::Value fval00, fval000001, fval001;
+  mlir::Value fval00, fval000001, fval001, fval025;
+  float f16_threshold = 0.25f;
   // Create constants needed for verification
   if ((randomSeed.getValue() != "none" &&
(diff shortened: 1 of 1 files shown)


mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -1275,14 +1275,16 @@ createVerifierFunc(ModuleOp &module, const KernelIF &kernel,
     return b.create<arith::ConstantFloatOp>(loc, apVal, elemFType);
   };
 
-  mlir::Value fval00, fval000001, fval001;
+  mlir::Value fval00, fval000001, fval001, fval025;
+  float f16_threshold = 0.25f;
```
krzysz00: Can we get away with `0.0001f`?

zhanglx13: which number are you referring to? Both 0.001f and 0.000001f are used but not 0.0001f.

krzysz00: I'm saying the threshold shouldn't be 25% but some smaller number

... come to think it, while we're here, these long strings of 0s could use scientific notation.

But good to know

krzysz00 (approved):
Hotfix approved.

Long-term, could we fiddle with the CPU wrapper to have it do the computations in half-float? We have access to `APFloat` for that sort of thing...

zhanglx13:
Definitely. It could be part of the validation and verification function redesign.

jungpark-mlir:
> Hotfix approved.
> 
> 
> 
> Long-term, could we fiddle with the CPU wrapper to have it do the computations in half-float? We have access to `APFloat` for that sort of thing...

The problem is, gpu doesn't do half float calculation in mfma. It's repetition of dot4, each 16b x 16b multiplication results ~32bit and passed into ~80 bit accumulator. At the end of the dot4, the result is downcast to fp32.
In order to emulate this on CPU correctly, we need to implement the same igemm on the CPU as well. This is not something we want to do.
Also I think @yiqian1 tried emulating half calculation on cpu before.

jungpark-mlir (approved):
Also fine to have temporary fix

jerryyin:
> The problem is, gpu doesn't do half float calculation in mfma. It's repetition of dot4, each 16b x 16b multiplication results ~32bit and passed into ~80 bit accumulator. At the end of the dot4, the result is downcast to fp32.

@jungpark-mlir Any insights we can have something more permanent than a hard-coded 25%?

krzysz00:
@jungpark-mlir To make sure I'm understanding you right, each [4x4xfp16] * [4x4xfp16] -> [4x4xfp16] dot product is computed in half-floats, and then the result of that is brought up to high precision (and eventually cast down to fp32)? Or are the floats expanded before the product?

jungpark-mlir:
@jerryyin 
My personal preference is using a controlled input. Unfortunately there is no standard for the ML operator accuracy criteria as far as I know, it's more driven by the application (models and data).
Easy way is to use the convolution and data of real network model, it was once refused before but we may want to reconsider it.

Another way is to create a numerical model to determine the range of random input or a synthetic pattern that can lead the similar results between cpu and gpu.

More popular way is to introduce more statistics skill to check the validity, for example, adding mean squared error, using custom weight to scale relative error, etc.
This approach is more widely accepted because this is the easiest to apply and there is no absolute solution in anyway.

jungpark-mlir:
@krzysz00 
Basic unit constructing mfma matrix multiplication is dot4 ALU which calculate dot4(A, B) = A.0xB.0 + A.1xB.1 + A.2xB.2 + A.3xB.3

When A and B are half type, each multiplication results 32bit float which doesn't lose any precision. And sum of those four FP32 values is calculated using FP80 accumulator. The final result of this dot4 ALU is downcast as FP32 value.

I can't remember exactly but dot4 also takes additional FP32 to accumulate value in the matmul. i.e., ALU operation might be 
(float)dot4(half4 A, half4 B) + (float)C

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -1370,7 +1372,12 @@ createVerifierFunc(ModuleOp &module, const KernelIF &kernel,
 
     auto absCpuVal = testBody.create<math::AbsOp>(loc, cpuLoadVal);
 
-    mlir::Value maxPercentVal = fval000001; // 0.0001%
+    mlir::Value maxPercentVal;
+    if (elemType.getIntOrFloatBitWidth() < 32) {
+      maxPercentVal = fval025; // 25%
```
jerryyin: I know this was the pre-existing logic for error checking but I'd prefer us to be more precise in this conditional: If this issue only exists in fp16, use fp16 as the exact conditional. Otherwise, int8 validation check will also be bumped to a very high threshold.

jerryyin:
> for example, adding mean squared error

I think MIOpen was adopting exactly RMS for error checking. I should be able to poke into their code and see if they treat fp16 backward weight specially.

mlir/tools/miopen-gen/miopen-gen.cpp
zhanglx13: Done.
I also added -threshold as a command line option to `miopen-gen`. The default value is 0.25.

jerryyin (approved):

PR #697, opened 2022-07-15
Run parameter sweep only on multi_gpu

jungpark-mlir:
Try to avoid time out in weekly test

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -362,7 +362,7 @@ pipeline {
                     docker {
                         image dockerImage()
                         args dockerArgs()
-                        label "mlir && ${params.canXdlops ? '(gfx908 || gfx90a)' : 'rocm' }"
+                        label "mlir && ${params.canXdlops ? '(gfx908 || gfx90a) && multi_gpu' : 'rocm' }"
                         alwaysPull true
                     }
                 }


jerryyin:
Per most recent run here: http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir-weekly/detail/mlir-weekly/76/pipeline/122. 

It looks to be consistent that: 
 - `conv_structure` parameter sweep take around 4 hours
 - `perf_config` parameter sweep take around 50 mins

Therefore, I'm not convinced making multi-gpu to run will reduce the time it spend, because this is taking as long as single gpu. Instead there'are one of the two ways:
 - Break down `conv_structure` into two subtests, each sweep one config, CC: @krzysz00 
 - Or, talk with Omkar, and make sure jenkins don't timeout for a run over 4 hours.

jungpark-mlir:
Total runtime doesn't trigger the time out, it times out when it doesn't come back within an hour.
Currently not sure if a set of run might be taking more than an hour or there is a real issue.
Probably we can make each run to print a dot as a heartbeat, let me see

krzysz00:
If a convolution that doesn't have any parameter larger than 64 is taking more than an hour to come back, that is one heck of a bug - though it's probably The Doorbell Issue (tm)

krzysz00 (approved):

PR #698, opened 2022-07-15
[DONOT SQUASH]Upstream merge

yiqian1:
This PR merges upstream updates till the commit b12930e1338b.

Remaining issues:

- A fusion test is now disabled. [#594](https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/594)

- A static assertion is disabled. [#595](https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/595)

A couple of patches need to be revisited [#596](https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/596) and [597](https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/597).

These changes have passed public CI: http://ml-ci.amd.com:21096/job/MLIR/job/test-nightlyCI/28/

These changes have passed private CI: http://rocmhead.amd.com:8080/job/MLIR/job/test-nightlyCI/6/

and weekly CI: http://rocmhead.amd.com:8080/job/MLIR/job/test-weeklyCI/13/


mlir/lib/Dialect/MIOpen/MIOpenDialect.cpp
```diff
@@ -442,6 +442,21 @@ LogicalResult InsertSliceOp::verify() {
 //===-----------------------------------------------------===//
 // TransformingForOp
 //===-----------------------------------------------------===//
+
+  /// Parse a list of assignments of the form
+  ///   (%x1 = %y1 : type1, %x2 = %y2 : type2, ...)
+static ParseResult
+  parseAssignmentListWithTypes(SmallVectorImpl<OpAsmParser::Argument> &lhs,
```
jerryyin: This commit ad2af476ffcf7b09a2f2ba4b5648d851e7c21c75 looks a bit suspicious. Why do we have to bring over a deleted method from upstream? Any other alternatives from `parser`?

jerryyin:
Thanks for your work!

I'll abstain from the rest of the review as it is extremely hard to follow through/open some of the diffs! Consider my review as approved if you manage to get full pass in PR + nightly + weekly test.

A piece of advice for the future:
 - Try to avoid merge commits (Use rebase instead), just so that recently merged PRs doesn't show up in the code diffs. It can get really confusing for merged PR diffs to show up again.
 - Try with as small number of commits as possible, I'd organize the commit as below. I find that the less commit you keep, the easier it is for you to rebase the changes and not ended up to be a mess.
   - First commit for merge/squash + conflicts
   - Second commit for fixing external repo unit tests
   - Third commit for fixing internal repo unit tests

mlir/lib/Dialect/MIOpen/MIOpenDialect.cpp
yiqian1: TransformingForOp::parse() needs this function.

yiqian1: And the issue #596 will revisit this patch and see if we can do it without affecting the upstream.

jerryyin: But there must be a reason upstream get rid of this. Are there no substitute left? And how does other places handle the mismatch when it also needs to parse something?

jerryyin: > And the issue https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/596 will revisit this patch and see if we can do it without affecting the upstream.

Gotcha, I didn't realize you have already had a ticket open. Sounds good.

krzysz00: I've found the commit and I can take the fix unless someone desperately wants a learning experience

krzysz00:
@jerryyin - if it helps
1) You can filter out commits from the diff viewer 
2) My usual approach is `diff -rup ~/llvm-project/mlir ./external/llvm-project-mlir/ | less` with the copy of upstrem checked out to the commit Yi used for the merge, and analogously for our repo

krzysz00:
I'll run the nightly and weekly tests by hand and approve?

jerryyin:
> I'll run the nightly and weekly tests by hand and approve?

@krzysz00 For a change as large as this, desirably it shall be a full CI run that includes both xdlops and nonxdlops (on different supported architectures)

@yiqian1 Would you mind kicking out those from this pull request CI?

krzysz00:
I'd like to not be using PR CI as the place to kick off nightly/weekly runs because those flags are persistent and I don't want to be firing off a weekly every time someone merges anything.

krzysz00:
I'd say it makes more sense to use one of the clones of `mlir-nightly-all` to do that bit

yiqian1:
I've run nightlies and weekly already. Public nightly CI passed. Private CI had a timeout. I'm digging more and will rerun again.

krzysz00:
@yiqian1 You might want to pull in the `multi_gpu` PR that Jungwook just merged in - there seems to be a single-GPU node that makes our parameter sweeps misbehave.
Another option is to hotfix and turn the parallelism in the parameter sweep script (look for the default value of `-j`  and make it something like (number of CPUs) / 4)

yiqian1:
> @yiqian1 You might want to pull in the `multi_gpu` PR that Jungwook just merged in - there seems to be a single-GPU node that makes our parameter sweeps misbehave. Another option is to hotfix and turn the parallelism in the parameter sweep script (look for the default value of `-j` and make it something like (number of CPUs) / 4)

Thanks. I'll try.

jerryyin:
> I'd like to not be using PR CI as the place to kick off nightly/weekly runs because those flags are persistent and I don't want to be firing off a weekly every time someone merges anything.

I think upstream merge needs special treatment that we'd not run into that situation:
 - We have declared to hold off merging other PRs until this one get in since last weekly meeting
 - Owner of upstream merge should actively check run result and merge ASAP once it is a full pass

I'm also all for it if you are able to have it validated in a separate CI job. Once you get the full pass please attach the job id and mark the job to keep log forever. This way we don't risk losing the log should be debug something across upstream merge.

jerryyin (approved):

PR #699, opened 2022-07-18
Moving lds to register load into blockwisegemm

jerryyin:
This PR moved lds->register transfer into blockwisegemm and significantly simplified the xdlops gemm responsibility.

Follow-up items after this PR:
- fix 598
- Follow-up cleanups in xdlopsgemm that incorporate M/N/K as threadwisegemm does
- Move xdlopsgemm into coordinate transfer
- Move blockwisegemmV2 into coordinate transfer


mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -240,10 +241,17 @@ struct BlockwiseGemmV2RewritePattern
                                 ConversionPatternRewriter &b) const override {
     Location loc = op.getLoc();
 
+    int64_t M = op->getAttr("m").template cast<IntegerAttr>().getInt();
```
krzysz00, line 244: Is this a temporary measure until we've done some stuff with transforms, or should we go ahead and put `m`, `n`, and `k` as attributes in the operation definition?

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -265,8 +500,8 @@ struct BlockwiseGemmV2RewritePattern
 
       auto xdlopsGemmV2Op = b.replaceOpWithNewOp<XdlopsGemmV2Op>(
           op, resultTypes, adaptor.matrixA(), adaptor.matrixB(),
-          op.ldsBufferOffsetA(), op.ldsBufferOffsetB(), adaptor.waveOffsetA(),
-          adaptor.waveOffsetB(), adaptor.bufferA(), adaptor.bufferB(),
+          op.ldsBufferOffsetA(), op.ldsBufferOffsetB(), zeroConstantOp,
+          zeroConstantOp, adaptor.bufferA(), adaptor.bufferB(),
```
krzysz00, line 504: Why pass 0 instead of just eliminating these?

krzysz00 (commented):
Wouldn't it make sense to do a broader rewrite while we're here?

krzysz00:
(If nothing else, can the wave offset be computed inside BlockwiseGemmToThreadwise)?

jerryyin:
> Wouldn't it make sense to do a broader rewrite while we're here?

Personal in preference: go with PR that does as little and as easy to review as possible. I can consolidate all changes in a PR but that would accumulate significant amount of risk and explode the reviewer.

> (If nothing else, can the wave offset be computed inside BlockwiseGemmToThreadwise)?

It's likely that both register offset and wave offset can be computed inside. If you don't mind, I'll do this in next PR now that I've updated the responsibility between blockwise/xdlops gemm.

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
jerryyin: Because there are a couple of branches below that we'd actually need the offset that's non-zero.

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
jerryyin: I'm not sure I'm following, I think m, n and k are already in the op's attributes, and also, we are populating those attributes?

Do you mean something differently on `attributes in the operation definition`?

krzysz00: I mean that, instead of implicitly assuming they're there, they should be listed as required attributes in `MIOpenOps.td` ... unless we're getting rid of them through coordinate transforms stuff and so can get `M`, `N`, `K`, and `KPack` from the argument types ... which could be either this PR or a future PR

krzysz00:
Yeah, I see the argument for small PRs - I just tend not to go in for this sort of very partial refactoring because it's tricky to find intermediate states that feel like good PR candidates.

krzysz00 (approved):

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
jerryyin: >  they should be listed as required attributes in MIOpenOps.td ... which could be either this PR or a future PR

Thanks for the reminder and I can confirm with you that this is on my list. The goal is to catch up with nonxdlops side of things you have done.

PR #700, opened 2022-07-19
Backport recent MIGraphX fixes to rel5.3

jungpark-mlir:


krzysz00 (approved):
Seems like a small and reasonable backport to fix known issues to me, approved

jerryyin:
@jungpark-mlir We have already tagged the first dot release of `rocm-5.3.0`. This will become a new one only if there's a revision in general for ROCm. However, if you only want to make this into ROCm-5.3 branch and use it in the branch, it is okay.

jungpark-mlir:
@jerryyin Yep, I think that all that we can do for now.
This is only needed to MIGraphX integration and I'll inform this for the versioning in MIGraphX.

jerryyin (approved):

PR #704, opened 2022-07-21
Add cmake dependences needed for single-threaded Unix-Makefile builds.

pcf000:
rocm-arch #785 and #787 report build failures when using -DBUILD_FAT_LIBMLIRMIOPEN=1, Unix Makefiles, and single-threaded compilation.  This patch adds some dependences to CMakeLists.txt to make sure those included files are built before their includers.  Tested with the 5.1.0 and 5.2.0 release source and the latest main line, patched on the main line.

### DIFF
--- a/mlir/lib/Dialect/MIOpen/Tuning/CMakeLists.txt
+++ b/mlir/lib/Dialect/MIOpen/Tuning/CMakeLists.txt
@@ -9,6 +9,8 @@ add_mlir_dialect_library(MLIRMIOpenTuning
 
   DEPENDS
   MLIRSupport
+  MLIRGPUOpInterfacesIncGen
+  MLIRMIOpenPassIncGen
 )
 
 if (MLIR_MIOPEN_SQLITE_ENABLED)
--- a/mlir/lib/Dialect/MIOpen/Tuning/CMakeLists.txt
+++ b/mlir/lib/Dialect/MIOpen/Tuning/CMakeLists.txt
@@ -9,7 +9,6 @@ add_mlir_dialect_library(MLIRMIOpenTuning
 
   DEPENDS
   MLIRSupport
-  MLIRGPUOpInterfacesIncGen
   MLIRMIOpenPassIncGen
 )
 
--- a/mlir/lib/Conversion/GPUToMIGraphX/CMakeLists.txt
+++ b/mlir/lib/Conversion/GPUToMIGraphX/CMakeLists.txt
@@ -8,6 +8,7 @@ add_mlir_conversion_library(MLIRGPUToMIGraphX
 
   DEPENDS
   MLIRMIOpenConversionPassIncGen
+  MLIRMIGraphXTypeIncGen
 )
 
 target_link_libraries(MLIRGPUToMIGraphX
--- a/mlir/lib/Conversion/MIOpenToGPU/CMakeLists.txt
+++ b/mlir/lib/Conversion/MIOpenToGPU/CMakeLists.txt
@@ -7,6 +7,7 @@ add_mlir_conversion_library(MLIRMIOpenToGPU
   DEPENDS
   MLIRConversionPassIncGen
   MLIRMIOpenConversionPassIncGen
+  MLIRMIOpenPassIncGen
 )
 target_link_libraries(MLIRMIOpenToGPU
   PUBLIC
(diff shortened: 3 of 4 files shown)


mlir/lib/Dialect/MIOpen/Tuning/CMakeLists.txt
```diff
@@ -9,6 +9,8 @@ add_mlir_dialect_library(MLIRMIOpenTuning
 
   DEPENDS
   MLIRSupport
+  MLIRGPUOpInterfacesIncGen
```
jerryyin: I'd be surprised to see `MLIRGPUOpInterfacesIncGen` is necessary inside `MLIRMIOpenTuning` library. This targets use no gpu ops. Could you double check whether/why this is needed?

krzysz00: I think it's to do with the GPU async interface?

jerryyin: `MLIRMIOpenTuning` Shouldn't use any gpu async stuff? It only derived tuning parameters based on conv/grid level ops.

krzysz00: It might have dependencies that do and that don't declare their dependencies correctly?

If nothing else, `#include`ing the GPU dialect somewhere, even if we don't actually use it in that file anymore, could do this.

We might want to go hunting for redundant includes

jerryyin: > It might have dependencies that do and that don't declare their dependencies correctly?

That's sounds to be the case if this is truly needed to make compilation pass. And if so I'd prefer this PR to address the redundant `#include` from `MLIRMIOpenTuning` component only, such that we don't have to pull in gpu ops dependencies. Otherwise we'd be creating new problem out from the fixing of the old one.

krzysz00: Yeah, good idea

pcf000:
[AMD Official Use Only]

I'll dig up the include chain that led to this dependence.
Unfortunately, it'll take a couple of hours, because I don't have it
recorded and will need to reproduce.

jerryyin:
> I'll dig up the include chain that led to this dependence. Unfortunately, it'll take a couple of hours

I'm not sure you'll need to spend hours and test the single thread compilation again, instead, you can just get rid of gpu ops dependency and see if compilation can proceed or break. And if it breaks, follow the chain to get rid of redundant header.

krzysz00:
The thing is, that dependency was implictly being generated fast enough that we didn't notice the issue in the multi-threaded context, so running the single-threaded job is important

jerryyin:
> The thing is, that dependency was implictly being generated fast enough that we didn't notice the issue in the multi-threaded context

But we can do a make clean and compile just the tuning target. At this point getting rid of a dependency shouldn't matter in compilation threads context.

I guess once the fix is done, you'll probably need another compilation run to make sure things are still good.

krzysz00:
Good point, but I'd still go for a single-threaded compile of just the target

pcf000:
I removed the added dependence and a -j128 build still worked, mostly as expected.

I went back to single-threaded on a 5.1.0 base, removed the dependence, and so far I've removed apparently-unnecessary includes of GPUDialect.h from BackwardDataPaddingKernel.h, BackwardWeightV4R4Helper.h, and XdlopsCodeSelection.h (and XdlopsCodeSelection.h from BackwardDataPaddingKernel.h).

That may be all of them, but it's still building and I'll need to run through a couple other scenarios before I'm certain.

krzysz00:
(If it helps, I'm hurling `include-what-you-use` at our repo to try and discover unneeded includes)

pcf000:
Good, I was postulating such a tool, but didn't know if one existed.
Meanwhile, the two Backward*.h files went away between 5.1.0 and 5.2.0.  Do we patch old releases?

krzysz00:
(Said tool's output hasn't been _too_ helpful, I suspect the real approach here is someone going through and removing includes until things break and then adding either the includes of forward declarations back)

I figure we ideally ought to get into a point release, but I'm not sure how important that actually is.

pcf000:
The net change on the main line is merely to remove the one MLIRGPUOpInterfacesIncGen dependence.  The other files that were causing trouble either don't exist since 5.2.0 or don't include the GPU file.  I pushed that one change.

krzysz00:
One broader comment: I'm pretty sure it's not typical to depend on the generation of the include files - things like `FuncOpsIncGen` maybe should be a dependency on `FuncDialect` or whatever the top-level CMake object for the `Func` dialect is

pcf000:
All but one of the dependences I added are adjacent to other dependences on *IncGen, so I'm at least following the local pattern.  I see about 200 dependences on *IncGen in CMakeLists.txt under external/llvm-project/mlir.

jerryyin (approved):
LGTM, thank you!

krzysz00:
Ok, I was wrong about how to do the dependencies.

PR #706, opened 2022-07-22
Backport to 5.3: use set() instead of list() to append rpath flags (#688)

krzysz00:
In cmake/llvm-project.cmake rpath flags are appended to CMAKE_EXE_LINKER_FLAGS with list(APPEND .... This inserts a ";" between the previous flags (if present) and the newly appended flags, breaking compilation. This can be avoided by using set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ..." to append the flags instead.

This breaks people externally and needs to be in 5.3. It's already in master as #688

### DIFF
--- a/cmake/llvm-project.cmake
+++ b/cmake/llvm-project.cmake
@@ -48,8 +48,6 @@ list(APPEND LLVM_INCLUDE_DIRS
 )
 
 # Linker flags
-list(APPEND CMAKE_EXE_LINKER_FLAGS
-  " -Wl,-rpath -Wl,${CMAKE_CURRENT_BINARY_DIR}/external/llvm-project/llvm/lib"
-)
+set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-rpath -Wl,${CMAKE_CURRENT_BINARY_DIR}/external/llvm-project/llvm/lib")
 
 add_subdirectory("${LLVM_PROJ_SRC}/llvm" "external/llvm-project/llvm" EXCLUDE_FROM_ALL)


pcf000 (approved):
Looks fine to me, especially with all the history.

jerryyin:
@krzysz00 As much as I understand you want to help the rocm-arch repository, this backport would actually not help them. 

Because of how internal repositories are pulling us in, I have to be consistent with their approach and tag the commit that actually get released. What that means is that if there's a 5.3.1 revision, it will not pull in the new commit but the similar commit as 5.3.0.

Therefore, for consistency purpose, I suggest to drop this PR and wait till it officially go in from next release.

krzysz00:
@jerryyin So, what you're saying is that 5.3.0 is already frozen?

If so, then we should release this in 5.3.1 - we have breakage in our build system

jerryyin:
> If so, then we should release this in 5.3.1 - we have breakage in our build system

The problem is, we are bind with our client MIOpen. There needs a strong reason to backport commit into their release branch, because of QA devops and process ... In other words, we can't just backport because there are minor breakages. We can backport only when everyone agrees we should. Allowing https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/700 in is a mistake, but from this point on-wards I will not approve any backport commits unless @mvermeulen and @junliume together approved and agreed it is necessary. If you insist, please prepare a strong enough case to the component owners that it is important enough to be backported.

From another aspect, what that means is that we are stuck with 5.3.0, unless it is a crucial fix that can actually impact the past release. Now, if we continue to backport new commits into past release branch:
 1. It will not help anyone because newer commits will not make its way into the revisions
 2. It will confuse ourselves at best because then there's this inconsistency between the release branch and the tag

krzysz00:
I'd argue that #700 - which, as I understand it, fixes broadcast functionality that is legitimately needed by someone - should be pushed into 5.3.x - if not 5.3.0 because the commit there is frozen, that 5.3.1 .

Isn't "fixing minor breakages" what the point releases are for?

pcf000:
[AMD Official Use Only]

    Date: Tuesday, July 26, 2022  10:56AM CDT
    From: Krzysztof Drewniak ***@***.***>

    Isn't "fixing minor breakages" what the point releases are for?

Does ROCm do point releases?  It seems to make minor releases at a
fairly quick pace, and have point releases only for customer needs.  We
also don't seem to maintain prior releases.

(Looking at it through the experience of a different project with a
different release cadence, where customers would sometimes use the same
release for years.)

krzysz00:
Whereas there's some pretty strict guidelines about what goes into a patch release and this fix doesn't obviously qualify (that is, you could darn well make the argument that this is a customer breakage etc, it's nonobvious enough that it may not be worth the effort) - closing the attempted backport. If more information comes about concrete end-users being broken by this or so on, I'd be willing to reopen so we can try and fight the QA folks about this.

Barring that, the patch to the CMake will be part of ROCm 5.4 but not earlier. @linedot

mvermeulen:
> Does ROCm do point releases?
Barely...

ROCm doesn't do point releases in a traditional sense of maintaining a previous release branch and providing general purpose fixes for all customers, particularly for a more extended time.

Instead ROCm has more of the following model
- Cadence of more frequent minor releases, 5.3, 5.4, 5.5, etc with an emphasis of putting fixes/updates in that next release.
- Putting out a minor release and then having a very targeted list of fixes for a specific customer that somehow missed the initial release, e.g. 5.2.3 has a target of a very specific customer.

Because of this the list of things to be fixed in a 5.x.x is pretty tightly controlled and typically justified most as being required by specific target customer.  In addition because the regular minor release also have a somewhat frequent cadence - there tends to be some tension in MIOpen of whether to fix things for specific customer in 5.2.x or put the fix in the next minor release - particularly when there are tradeoffs in running the release crank again or having to go back and retune.

ROCm doesn't currently have a release cadence where we put out a minor point release and support customers using that release for years.  Instead for most all situations/customers there is incentive that if you need a fix you have a choice to live with the defect or move to further point release that fixes the defect.

PR #707, opened 2022-07-26
Enable vectorized stores for non-xdlops

krzysz00:


jerryyin (approved):
LGTM

However, I'd be very careful with this. In my past experience. When there's a comment of `TODO` or temporary fix it usually ended up to be there because of a past bug/failure of some sort. So could you watch closely to the nightly result after this get merged?

krzysz00:
That particular TODO was mine, marking out future work

PR #711, opened 2022-07-28
Making memref to capture vectorC/D in gemm from xdlops lowering

jerryyin:
First commit:
 - Amended blockwisegemm and xdlopsgemm op signature to include return
 - Fixed xdlopsgemm out of bound access when kPack = 1 and k_base > 1
 - Moved xdlops register offset to attributes
 - Moved gridwisegemm allocation before blockwisegemm compuation

Second commit:
 - Replaced vectorC/D as memref in blockwise/xdlops gemm

Third commit: 
 - Added vectorC/D type conversion before write back

Forth commit
 - Amended memref of scalars to memref of vectors
 - Fixed the redundantly allocated vgpr in populated ISA

Fifth commit
 - Fixed all unit tests failures

### DIFF
--- a/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
+++ b/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
@@ -793,7 +793,7 @@ def MIOpen_XdlopsGemmV2Op:
     Matrices A and B reside in LDS, the buffers live in registers, C is a vector
   }];
   let assemblyFormat = [{
-    `(` operands `)` attr-dict `:` type(operands) `->` type(results)
+    `(` operands `)` attr-dict `:` type(operands)
   }];
 }
 
--- a/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
+++ b/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
@@ -199,9 +199,10 @@ def MIOpen_GpuAllocOp:
 
 // TBD: eventually replace this with linalg.fill?
 def MIOpen_FillOp:
-    MIOpen_Op<"fill", [AllElementTypesMatch<["input", "value"]>]>,
+    MIOpen_Op<"fill">,
     Arguments<(ins AnyMemRef:$input,
-                   AnyTypeOf<[AnyInteger, AnyFloat]>:$value)> {
+                   //AnyTypeOf<[AnyInteger, AnyFloat]>:$value)> {
+                   AnyTypeOf<[AnyInteger, AnyFloat, AnyVector]>:$value)> {
   let summary = "Fill memory with constant value on GPU";
   let description = [{
     The `miopen.fill` op fills a memref on GPU with a constant value.
@@ -730,7 +731,8 @@ def MIOpen_BlockwiseGemmV2Op:
                              VectorOfLengthAndType<[2, 4], [F32]>,
                              VectorOfLengthAndType<[2, 4, 8], [F16, BF16]>,
                              VectorOfLengthAndType<[4, 8, 16], [I8]>]>:$bufferB,
-                   MemRefOf<[F32, F16, BF16, I32]>:$matrixRes)>{
+                   MemRefOf<[F32, F16, BF16, I32,
+                             VectorOf<[F32, F16, BF16, I32]>]>:$matrixC)>{
   let summary = "Blockwise GEMM XDLOPS version";
   let description = [{
     The `miopen.block_gemm` op does GEMM at workgroup (block) level.
@@ -784,7 +786,8 @@ def MIOpen_XdlopsGemmV2Op:
                              VectorOfLengthAndType<[2, 4], [F32]>,
                              VectorOfLengthAndType<[2, 4, 8], [F16, BF16]>,
                              VectorOfLengthAndType<[4, 8, 16], [I8]>]>:$bufferB,
-                   MemRefOf<[F32, F16, BF16, I32]>:$matrixRes)>{
+                   MemRefOf<[F32, F16, BF16, I32,
+                             VectorOf<[F32, F16, BF16, I32]>]>:$matrixC)>{
   let summary = "XDLOPS GEMM V2";
   let description = [{
     The `miopen.xdlops_gemm_v2` op is an abstraction of doing GEMM based on XDLOPS.
--- a/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
+++ b/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
@@ -730,9 +730,7 @@ def MIOpen_BlockwiseGemmV2Op:
                              VectorOfLengthAndType<[2, 4], [F32]>,
                              VectorOfLengthAndType<[2, 4, 8], [F16, BF16]>,
                              VectorOfLengthAndType<[4, 8, 16], [I8]>]>:$bufferB,
-                   MemRefOf<[F32, F16, BF16, I32]>:$matrixRes,
-                   Variadic<VectorOfRankAndType<[1], [F32, F16, I32]>>:$vectorCs)>,
-    Results<(outs Variadic<VectorOfRankAndType<[1], [F32, F16, I32]>>: $vectorDs)> {
+                   MemRefOf<[F32, F16, BF16, I32]>:$matrixRes)>{
   let summary = "Blockwise GEMM XDLOPS version";
   let description = [{
     The `miopen.block_gemm` op does GEMM at workgroup (block) level.
@@ -743,7 +741,7 @@ def MIOpen_BlockwiseGemmV2Op:
     not slices of a larger object.
   }];
   let assemblyFormat = [{
-    `(` operands `)` attr-dict `:` type(operands) `->` type($vectorDs)
+    `(` operands `)` attr-dict `:` type(operands)
   }];
 }
 
@@ -786,9 +784,7 @@ def MIOpen_XdlopsGemmV2Op:
                              VectorOfLengthAndType<[2, 4], [F32]>,
                              VectorOfLengthAndType<[2, 4, 8], [F16, BF16]>,
                              VectorOfLengthAndType<[4, 8, 16], [I8]>]>:$bufferB,
-                   MemRefOf<[F32, F16, BF16, I32]>:$matrixRes,
-                   Variadic<VectorOfRankAndType<[1], [F32, F16, I32]>>:$vectorCs)>,
-    Results<(outs Variadic<VectorOfRankAndType<[1], [F32, F16, I32]>>:$vectorDs)> {
+                   MemRefOf<[F32, F16, BF16, I32]>:$matrixRes)>{
   let summary = "XDLOPS GEMM V2";
   let description = [{
     The `miopen.xdlops_gemm_v2` op is an abstraction of doing GEMM based on XDLOPS.
--- a/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
+++ b/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
@@ -743,10 +743,10 @@ def MIOpen_BlockwiseGemmV2Op:
     not slices of a larger object.
   }];
   let assemblyFormat = [{
-    $matrixC `+` `` `=` $bufferA `by` $matrixA `[` $waveOffsetA `]` `*`
-                        $bufferB `by` $matrixB `[` $waveOffsetB `]` attr-dict
-    `:` type($matrixC) `+` `` `=` type($bufferA) `by` type($matrixA) `*` 
-                                  type($bufferB) `by` type($matrixB)
+    $matrixC `+` `` `=` $bufferA `from` $matrixA `[` $waveOffsetA `]` `*`
+                        $bufferB `from` $matrixB `[` $waveOffsetB `]` attr-dict
+    `:` type($matrixC) `+` `` `=` type($bufferA) `from` type($matrixA) `*` 
+                                  type($bufferB) `from` type($matrixB)
   }];
 }
 
--- a/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
+++ b/mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
@@ -743,7 +743,10 @@ def MIOpen_BlockwiseGemmV2Op:
     not slices of a larger object.
   }];
   let assemblyFormat = [{
-    `(` operands `)` attr-dict `:` type(operands)
+    $matrixC `+` `` `=` $bufferA `by` $matrixA `[` $waveOffsetA `]` `*`
+                        $bufferB `by` $matrixB `[` $waveOffsetB `]` attr-dict
+    `:` type($matrixC) `+` `` `=` type($bufferA) `by` type($matrixA) `*` 
+                                  type($bufferB) `by` type($matrixB)
   }];
 }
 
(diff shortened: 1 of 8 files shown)


mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -736,7 +737,7 @@ def MIOpen_BlockwiseGemmV2Op:
     not slices of a larger object.
   }];
   let assemblyFormat = [{
-    `(` operands `)` attr-dict `:` type(operands) `->` type($vectorDs)
+    `(` operands `)` attr-dict `:` type(operands)
```
krzysz00: Could I talk you into creating a pretty assembly format here?
Something like
```
$matrixC `+` `=` $matrixA `*` $matrixB
```
?
It's not _required_, but since we're here...

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -769,8 +770,8 @@ def MIOpen_XdlopsGemmV2Op:
                    MemRefOf<[F32, F16, BF16, I8]>:$matrixB,
                    IndexAttr:$ldsBufferOffsetA,
                    IndexAttr:$ldsBufferOffsetB,
-                   Index:$regOffsetA,
-                   Index:$regOffsetB,
+                   IndexAttr:$regOffsetA,
+                   IndexAttr:$regOffsetB,
```
krzysz00, line 779: Thanks!

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -789,7 +790,7 @@ def MIOpen_XdlopsGemmV2Op:
     Matrices A and B reside in LDS, the buffers live in registers, C is a vector
   }];
   let assemblyFormat = [{
-    `(` operands `)` attr-dict `:` type(operands) `->` type(results)
+    `(` operands `)` attr-dict `:` type(operands)
```
krzysz00: Simalarly on assembly format

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2331,9 +2336,8 @@ struct GridwiseGemmV2RewritePattern
     int64_t numBlksPerXdlops = (MPerXdlops * NPerXdlops) / (m * n);
     const auto &tailResults = blockwiseGemmV2TailOp->getResults();
     int64_t wavesInKernelBlock = kernelBlockSize / waveSize;
-    int64_t resultCVectorLen = vectorType.getNumElements();
-    int64_t numElements = resultCVectorLen * tailResults.size();
 
+    int64_t numElements = regCVectorLen * vectorNumber;
```
krzysz00: Can we name `vectorNumber` something like `numVectors`?

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2512,32 +2516,50 @@ struct GridwiseGemmV2RewritePattern
...
-    for (const auto &pair : llvm::enumerate(transformedTail)) {
-      Value cast = createTypeConversionOp(b, loc, pair.value(), castVectorType);
-      Value offset = b.createOrFold<arith::ConstantIndexOp>(
-          loc, pair.index() * resultCVectorLen);
-      b.create<miopen::InBoundsStoreOp>(loc, cast, resultMerged, offset);
+    Value convertedC = b.create<miopen::GpuAllocOp>(loc, convertedCType);
+
+    // Convert from memref<?xvector<?xT>> to memref<?xT> where the source T
+    // is the accumulatorType and destination type is destType
+    auto convertLoop = b.create<TransformingForOp>(
```
krzysz00, line 2398: Instead of the `MulIOp`, can we have the `bounds` be the size of the destination memref and the `strides` be the size of the vector?

krzysz00 (commented):
I've got a few initial comments - the main one is the loop you've added to gridwise gemm

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
jerryyin: Good idea! Indeed I should, and I will actually amend (or make sure) all gemm signatures are in this consistent format.

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
jerryyin: No problem. If I don't do this, we'd have to generate redundant IRs to calculate the right offset in this PR. Now looking forward to get rid of those in upcoming PRs...

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Yes agreed that this is an awkward name, but I was following the naming convention from `XdlopsSelection.h`. To limit the scale of refactoring in this PR, I'd rename the local variable name and leave the more comprehensive renaming to when we touch `XdlopsSelection.h`

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Nice suggestion! I don't like this `MulIOp` to start with, I'm glad you point this out to me, thank you!

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
krzysz00: I've been amending them as I refactor ... which reminds me, I should make sure I got `gridwise_gemm`

krzysz00: One note, if you're writing `+=`, you want 
```
`+` `` `=`
```
to get rid of a random space

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Wait, how does it work to make the source and destination in two coordinate space?

This transforming for op would:
 - `load registerC[coord]`
 - `type convert`
 - `store loaded convertedC[coord * vectorLen]` 

What should I do such that the load on registerC is still on linear coordinate space whereas store is strided by `vectorLen`

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2222,15 +2221,31 @@ struct GridwiseGemmV2RewritePattern
...
+    // i8: i32, since we have an i32 output
+    Type accumulatorType = destType;
+    if (elementType == b.getF16Type() || elementType == b.getBF16Type()) {
+      accumulatorType = b.getF32Type();
+    } else if (elementType == b.getI8Type()) {
+      accumulatorType = b.getI32Type();
+    }
+
+    VectorType accumulatorVectorType =
+        vectorType.cloneWith({}, accumulatorType);
```
krzysz00, line 2093: Is this spelled `clone(accumulatorType)` also?

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: Good question. Earlier idea withdrawn. My gut is saying to make a quick map that `Embed`s the destination coordinate with the appropriate coefficient to scale it from `[0, 1, ...]` to `[0, L, 2L, ...]` so that the multiplication is part of the loop structure, and thus analyzable.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: Agreed

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -736,7 +739,10 @@ def MIOpen_BlockwiseGemmV2Op:
     not slices of a larger object.
   }];
   let assemblyFormat = [{
-    `(` operands `)` attr-dict `:` type(operands) `->` type($vectorDs)
+    $matrixC `+` `` `=` $bufferA `by` $matrixA `[` $waveOffsetA `]` `*`
```
krzysz00: Nit: is `by` the right word here? Maybe `from` or `<-`

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
jerryyin: Sounds good, I was following your original practice anyways.

So if I do `+=` it automatically add a space in between? Haha, wired.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -198,12 +201,12 @@ struct XdlopsGemmV2RewritePattern : public OpConversionPattern<XdlopsGemmV2Op> {
 
     bool IsKReduction = (num_output_blks == 1) && (num_input_blks > 1);
 
-    Value bufferA = adaptor.bufferA();
-    Value bufferB = adaptor.bufferB();
-    auto bufferAType = adaptor.bufferA().getType().cast<MemRefType>();
-    auto bufferBType = adaptor.bufferB().getType().cast<MemRefType>();
-    Type bufferAElementType = bufferAType.getElementType();
-    Type bufferBElementType = bufferBType.getElementType();
+    Value matrixA = adaptor.matrixA();
```
krzysz00, line 203: I see things seem to have been renamed, but the one place I didn't see a renaming is in the names of the operation arguments themselves?

krzysz00 (commented):
Ok, so, we're definitely in the trivial nitpicking zone here - the main bit I'm staring at is the changes between what's called "buffer" and what's called "matrix"

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Hmm you are right, it looks like clone is the high level wrapper that allows either shape or type argument.

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
jerryyin: Yes, `from` is better, I used `by` just so that it is as short as possible.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
jerryyin: If I understand your question right, there used to be a `matrixA` argument, and whose name I reused (and replaced `bufferA`). So it is for `matrixB` and `bufferB`. You can see the updated diff on `MIOpenOps.td` that has those renames.

krzysz00: `MIOpenOps.td` doesn't rename the arguments at all.

What I was trying to work out: was it the case that, before this, "buffer" meant the LDS memref and "matrix" meant the in-register stuff and now you've switched it? It feels like you've switched it

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -765,22 +771,18 @@ def MIOpen_ThreadwiseGemmOp:
...
+                             VectorOfLengthAndType<[4, 8, 16], [I8]>]>:$matrixA,
                    MemRefOf<[F32, F16, BF16, I8,
                              VectorOfLengthAndType<[2, 4], [F32]>,
                              VectorOfLengthAndType<[2, 4, 8], [F16, BF16]>,
-                             VectorOfLengthAndType<[4, 8, 16], [I8]>]>:$bufferB,
-                   Variadic<VectorOfRankAndType<[1], [F32, F16, I32]>>:$vectorCs)>,
-    Results<(outs Variadic<VectorOfRankAndType<[1], [F32, F16, I32]>>:$vectorDs)> {
+                             VectorOfLengthAndType<[4, 8, 16], [I8]>]>:$matrixB,
+                   MemRefOf<[F32, F16, BF16, I32,
+                             VectorOf<[F32, F16, BF16, I32]>]>:$matrixC)>{
```
jerryyin, line 789: anchor

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
jerryyin: See [here](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/711/files#r940651714) for the rename.

Previously, matrix was lds and buffer is register. Now, matrix is register (and lds is no longer needed in new xdlops because blockwisegemm did the copy already).

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
krzysz00: ... right, I was looking at blockwise gemm's names. I'm very good at this.

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
krzysz00: I'd still go for `from` or something like `<-` on the theory that it'll make reading the IR clearer, which is rather important.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: To rephrase, instead of having one iteration domain, have two
The second one is
```
int64_t numOutputs = vectorType.getNumElements() * numVectors;
BottomUpTMBuilder vectorToScalar(b, {"scalar"}, {numOutputs}, loc);
vectorDest.embed({"vector"}, {0}, {numVectors}, "vector", {vectorType.getNumElements()});
TransformMapAttr vectorToScalarAttr = vectorToScalar.get();

...
// Set up the TransformingFor to take both the no-transforms array (vector coordinates) and the array containing the above transform (which gives scalar coordinates)
```

krzysz00 (commented):
Just thought I'd go back and elaborate on some things here

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
jerryyin: Addressed in c59cfa8

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Addressed in c59cfa8

krzysz00 (approved):
LGTM, thank you!

PR #712, opened 2022-07-30
In the TransposeOp special case, make sure const operand is placed ahead of all uses.

pcf000:
Also keep that operand out of inputNodes.

### DIFF
--- a/external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
+++ b/external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
@@ -447,6 +447,8 @@ void outlinePartitionOps(Operation *anchorOp, ArrayRef<Operation *> trailingOps,
 void TosaPartitionPass::specialCaseForTranspose(Operation *op,
                                                 SetVector<Operation *> &ops) {
   auto *operand = op->getOpOperand(1).get().getDefiningOp();
+  if (ops.contains(operand)) // If already present, move it for new use.
+    ops.remove(operand);
   ops.insert(operand);
 }
 
@@ -475,7 +477,8 @@ void TosaPartitionPass::traceInputs(Operation *op,
         // depth first
         traceInputs(usedOp, predecessors, inputNodes);
       }
-    } else {
+    } else if (!predecessors.contains(
+                   usedOp)) { // special-case consts aren't inputs
       inputNodes.insert(opnd);
     }
   }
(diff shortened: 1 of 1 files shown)


external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -447,6 +447,8 @@ void outlinePartitionOps(Operation *anchorOp, ArrayRef<Operation *> trailingOps,
 void TosaPartitionPass::specialCaseForTranspose(Operation *op,
                                                 SetVector<Operation *> &ops) {
   auto *operand = op->getOpOperand(1).get().getDefiningOp();
+  if (ops.contains(operand)) // If already present, move it for new use.
+    ops.remove(operand);
```
pcf000, line 451: The problem situation was two TransposeOps that used the same constant.  Tracing backward from the anchor, we'd save one transpose, then its constant (because it was tracing depth-first), and then the other transpose and its constant.  But the constant was already saved, so we wouldn't save it again, and it would end up between the two transposes instead of ahead of both.

By removing it, the usual insertion always places the constant ahead of all its uses.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -475,7 +477,8 @@ void TosaPartitionPass::traceInputs(Operation *op,
         // depth first
         traceInputs(usedOp, predecessors, inputNodes);
       }
-    } else {
+    } else if (!predecessors.contains(
+                   usedOp)) { // special-case consts aren't inputs
```
pcf000, line 481: This change and the one below avoid adding a node to inputNodes if it's already present in the saved nodes.  That generalises the problem of the special-case transpose constants -- since they don't qualify to be saved under normal circumstances, they'd otherwise fall into the inputNodes case, but we don't want to do that because it would just be an extra unused argument to the outlined function.  Cosmetic, but it bugged me.

pcf000:
I'm still not completely happy with how the special-case TransposeOp constants are handled, but this code works and all my attempts to change it so far have failed.  I haven't given up yet.

sjw36:
Seems like we should generally capture small constant vectors (like constant zeros). Then maybe we wouldn't need a special case.

sjw36 (approved):
Looks good.

sjw36:
May be simpler in the future to have a single list of ops and push_front for leading ops and push_back for trailing ops.

PR #713, opened 2022-08-01
[DO NOT SQUASH] MLIR-610: fixed bug in async-to-gpu

sjw36:
* added assert on hipRuntime error in debug
* updated test

https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/610
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/608

### DIFF
--- a/mlir/lib/Dialect/MIOpen/Pipelines/XMIRPipelines.cpp
+++ b/mlir/lib/Dialect/MIOpen/Pipelines/XMIRPipelines.cpp
@@ -1,4 +1,4 @@
-//===- Pipeline.cpp - Create MIOpen compilation pipeline ---------------===//
+//===- XMIRPipeline.cpp - Create XMIR runtime pipeline --------------------===//
 //
 // Copyright 2021 The MLIR Authors.
 //
@@ -21,26 +21,13 @@
 //===----------------------------------------------------------------------===//
 
 #include "mlir/Dialect/MIOpen/XMIRPipelines.h"
-#include "mlir/Dialect/MIOpen/Pipelines.h"
 
-#include "mlir/Conversion/MIOpenPasses.h"
-#include "mlir/Conversion/SCFToControlFlow/SCFToControlFlow.h"
-#include "mlir/Dialect/Arithmetic/Transforms/Passes.h"
-#include "mlir/Dialect/Async/Passes.h"
 #include "mlir/Dialect/MIOpen/Passes.h"
-#include "mlir/Dialect/Tensor/Transforms/Passes.h"
-#include "mlir/IR/Builders.h"
-#include "mlir/IR/BuiltinOps.h"
+#include "mlir/InitAllDialects.h"
 #include "mlir/InitAllPasses.h"
 #include "mlir/Pass/PassManager.h"
 #include "mlir/Pass/PassRegistry.h"
-
-#include "mlir/Conversion/FuncToLLVM/ConvertFuncToLLVM.h"
-#include "mlir/Conversion/FuncToLLVM/ConvertFuncToLLVMPass.h"
-#include "mlir/Conversion/GPUToROCDL/GPUToROCDLPass.h"
-#include "mlir/InitAllDialects.h"
 #include "mlir/Transforms/Passes.h"
-#include "llvm/Support/TargetSelect.h"
 
 using namespace mlir;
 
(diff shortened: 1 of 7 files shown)


mlir/test/xmir/e2e/resnet50_blk.e2e.mlir
```diff
@@ -14,10 +15,8 @@ module {
     }
      : (tensor<1x32x32x64xf32>, tensor<64x3x3x64xf32>, tensor<64xf32>) -> tensor<1x32x32x64xf32>
 
-    %1 = "tosa.clamp"(%0) {
-      min_fp = 0.0 : f32,
+    %1 = "tosa.reluN"(%0) {
```
jungpark-mlir, line 19: I heard reluN is deprecated in the latest TOSA specification because it's fully representable using clamp.

jungpark-mlir (commented):
Looks good to me

mlir/test/xmir/e2e/resnet50_blk.e2e.mlir
sjw36: I still see it upstream. Maybe it is a discussion.

jungpark-mlir: I mean the original TOSA specification. MLIR upstream seems slightly behind it, so we can still have it but need to change later when they update it.
I can confirm it's not present in the latest version 0.30.0 (older versions, e.g. 0.22.0 has ReluN)
https://developer.mlplatform.org/file/download/vdc2wr43p7p3b6nqgwxk/PHID-FILE-andd6yb2cozdie3ihy7h/tosa_spec.pdf

mlir/lib/Dialect/MIOpen/Pipelines/XMIRPipelines.cpp
```diff
@@ -21,26 +21,13 @@
...
 
-#include "mlir/Conversion/MIOpenPasses.h"
-#include "mlir/Conversion/SCFToControlFlow/SCFToControlFlow.h"
-#include "mlir/Dialect/Arithmetic/Transforms/Passes.h"
-#include "mlir/Dialect/Async/Passes.h"
 #include "mlir/Dialect/MIOpen/Passes.h"
-#include "mlir/Dialect/Tensor/Transforms/Passes.h"
-#include "mlir/IR/Builders.h"
-#include "mlir/IR/BuiltinOps.h"
+#include "mlir/InitAllDialects.h"
```
krzysz00, line 26: Is there any particular reason we're going for `InitAllDialects` here instead of including references to just the stuff we need? I could've sworn `InitAllDialects` was a convenience header for stuff like `mlir-opt` and that you weren't supposed to run around including it

mlir/lib/Dialect/MIOpen/Pipelines/XMIRPipelines.cpp
```diff
@@ -53,15 +40,23 @@ void xmir::buildModelPipeline(OpPassManager &pm,
...
+
+  } else {
     pm.addPass(createConvertAsyncToGPUPass());
     pm.addPass(createSymbolDCEPass());
+    pm.addPass(createGpuToLLVMConversionPass());
+    pm.addNestedPass<func::FuncOp>(createGpuAsyncRegionPass());
   }
   pm.addNestedPass<func::FuncOp>(createConvertMathToLLVMPass());
-  pm.addPass(createGpuToLLVMConversionPass());
+  pm.addPass(createConvertLinalgToLLVMPass());
```
krzysz00, line 55: Shouldn't these guys be taking an LLVM type converter so that we can set index-bitwidth=32?

krzysz00 (commented):
Minor nitpicks here, and we probably want to stick a [DO NOT SQUASH] on the PR

mlir/lib/Dialect/MIOpen/Pipelines/XMIRPipelines.cpp
sjw36: Could probably speed up compile a few seconds. If that gets to be a problem we should run a header minimizer on all the code.

mlir/lib/Dialect/MIOpen/Pipelines/XMIRPipelines.cpp
sjw36: This isn't for the kernel generation, just host code.

mlir/lib/Dialect/MIOpen/Pipelines/XMIRPipelines.cpp
krzysz00: Yeah, I've been vaguely wanting to run a header minimizer for a while...

krzysz00:
Broadly LGTM, but I'll let @jungpark-mlir do the actual approving since he knows this code better

jungpark-mlir (approved):
Looks good to me

PR #714, opened 2022-08-02
Move the computation of the thread offset in LDS to blockwise_gemm

krzysz00:
In order to make progress on removing hardcoded math in gridwise_gemm,
moves the computation of thread offsets into a map that lives in
blockwise_gemm. This means that gridwise_gemm no longer needs to
compute the offset in the m and n dimensions that each particular
thread should use.

In addition
- Rename the level0 and level1 clusters to be more descriptive. The
level 0 cluster represented the 16 threads of a CU wave being arranged
into a mPerCuwave x nPerCuwave grid, and the level 1 cluster was a 2D
grid made of the cuwaves in a block. More descriptive names will
hopefully save future refactorers the headache of understanding what
all this cluster stuff is about
- Change the IR format of blockwise_gemm in pursuit of the general
aesthetics project

### DIFF
--- a/mlir/lib/Dialect/MIOpen/TransformMapBuilder.cpp
+++ b/mlir/lib/Dialect/MIOpen/TransformMapBuilder.cpp
@@ -376,15 +376,41 @@ TransformMapBuilder::operator=(const TransformMapBuilder &other) {
     result = other.result;
     loc = other.loc;
 
-    startIndices = other.startIndices;
     startNames = other.startNames;
     startShape = other.startShape;
-    endIndices = other.endIndices;
     endNames = other.endNames;
     endShape = other.endShape;
     frozen = other.frozen;
+
+    startIndices.clear();
+    for (uint32_t i = 0, e = startNames.size(); i < e; ++i)
+      startIndices.insert({StringRef(startNames[i]), i});
+
+    endIndices.clear();
+    for (const auto &pair : endNames)
+      endIndices.insert({StringRef(pair.second), pair.first});
   }
   return *this;
+}
+
+TransformMapBuilder::TransformMapBuilder(const TransformMapBuilder &other)
+    : b(other.b), result(other.result), loc(other.loc), startIndices(),
+      startNames(other.startNames), startShape(other.startShape), endIndices(),
+      endNames(other.endNames), endShape(other.endShape), frozen(other.frozen) {
+  for (uint32_t i = 0, e = startNames.size(); i < e; ++i)
+    startIndices.insert({StringRef(startNames[i]), i});
+  for (const auto &pair : endNames)
+    endIndices.insert({StringRef(pair.second), pair.first});
+}
+
+TransformMapBuilder::TransformMapBuilder(TransformMapBuilder &&other)
+    : b(other.b), result(other.result), loc(other.loc), startIndices(),
+      startNames(other.startNames), startShape(other.startShape), endIndices(),
+      endNames(other.endNames), endShape(other.endShape), frozen(other.frozen) {
+  for (uint32_t i = 0, e = startNames.size(); i < e; ++i)
+    startIndices.insert({StringRef(startNames[i]), i});
+  for (const auto &pair : endNames)
+    endIndices.insert({StringRef(pair.second), pair.first});
 }
 
 /// Building from a defined set of upper dimensions
--- a/mlir/lib/Dialect/MIOpen/TransformMapBuilder.cpp
+++ b/mlir/lib/Dialect/MIOpen/TransformMapBuilder.cpp
@@ -403,16 +403,6 @@ TransformMapBuilder::TransformMapBuilder(const TransformMapBuilder &other)
     endIndices.insert({StringRef(pair.second), pair.first});
 }
 
-TransformMapBuilder::TransformMapBuilder(TransformMapBuilder &&other)
-    : b(other.b), result(other.result), loc(other.loc), startIndices(),
-      startNames(other.startNames), startShape(other.startShape), endIndices(),
-      endNames(other.endNames), endShape(other.endShape), frozen(other.frozen) {
-  for (uint32_t i = 0, e = startNames.size(); i < e; ++i)
-    startIndices.insert({StringRef(startNames[i]), i});
-  for (const auto &pair : endNames)
-    endIndices.insert({StringRef(pair.second), pair.first});
-}
-
 /// Building from a defined set of upper dimensions
 void TopDownTMBuilder::addTransform(TransformType type,
                                     ArrayRef<int64_t> params,
--- a/mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
+++ b/mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
@@ -112,42 +112,80 @@ struct BlockwiseGemmRewritePattern
     int64_t kPerThread = op.kPerThreadAttr().getInt();
     int64_t mPerThread = op.mPerThreadAttr().getInt();
     int64_t nPerThread = op.nPerThreadAttr().getInt();
-    int64_t mRepeatStride = op.mRepeatStrideAttr().getInt();
-    int64_t nRepeatStride = op.nRepeatStrideAttr().getInt();
+
+    int64_t mThreadsPerCuwave = op.mThreadsPerCuwaveAttr().getInt();
+    int64_t nThreadsPerCuwave = op.nThreadsPerCuwaveAttr().getInt();
+    int64_t cuwaveLen = mThreadsPerCuwave * nThreadsPerCuwave;
+
+    int64_t mCuwavesPerBlock = op.mCuwavesPerBlockAttr().getInt();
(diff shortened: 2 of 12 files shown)


krzysz00:
Note: this currently generates worse ISA now but won't have much impact if the math cleanup branch is applied.

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -112,42 +112,80 @@ struct BlockwiseGemmRewritePattern
...
+                              int64_t perThreadLen) -> TopDownTMBuilder {
+      TopDownTMBuilder splitTidForLDS(
+          b, {"k", repeatName, "tid", perThreadName, "kpack"},
+          {k, repeatLen, blockSize, perThreadLen, kPack}, loc);
+      splitTidForLDS.passThrough({"k", repeatName});
+      splitTidForLDS.merge({"m_cuwaves", "n_cuwaves", "m_cuwave", "n_cuwave"},
+                           {2, 3, 4, 5}, "tid",
+                           {mCuwavesPerBlock, nCuwavesPerBlock,
+                            mThreadsPerCuwave, nThreadsPerCuwave},
+                           /*isUnfold=*/true);
```
jerryyin, line 153: [Question] Are you trying to construct fewer builder by incorporating A and B's dimensions together and then later ignore `m` or `n`?

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -112,42 +112,80 @@ struct BlockwiseGemmRewritePattern
...
+    };
+
+    TopDownTMBuilder splitTidA =
+        ldsTidSplitter("m_repeat", mRepeat, "m_thread", mPerThread);
+    TransformMapAttr splitTidAAttr = splitTidA.get();
+    auto toLdsIndexA = TopDownTMBuilder::below(splitTidA, splitTidAAttr);
+    toLdsIndexA.passThrough("k");
+    toLdsIndexA.unmerge(
+        "m", 1, {"m_repeat", "m_cuwaves", "m_cuwave", "m_thread"},
+        {mRepeat, mCuwavesPerBlock, mThreadsPerCuwave, mPerThread});
```
jerryyin, line 166: I'm a bit lost on why you merge and then unmerge.

If the past and current `TopDownTMBuilder` are equivalent, I'd honestly think the past is more readable: I can immediately figure out the layout of `K, M, KPack` or `K, N, KPack`. Or my assumption is wrong? Did you do anything in addition to past builder?

mlir/lib/Dialect/MIOpen/TransformMapBuilder.cpp
```diff
@@ -376,17 +376,43 @@ TransformMapBuilder::operator=(const TransformMapBuilder &other) {
...
+    : b(other.b), result(other.result), loc(other.loc), startIndices(),
+      startNames(other.startNames), startShape(other.startShape), endIndices(),
+      endNames(other.endNames), endShape(other.endShape), frozen(other.frozen) {
+  for (uint32_t i = 0, e = startNames.size(); i < e; ++i)
+    startIndices.insert({StringRef(startNames[i]), i});
+  for (const auto &pair : endNames)
+    endIndices.insert({StringRef(pair.second), pair.first});
+}
+
+TransformMapBuilder::TransformMapBuilder(TransformMapBuilder &&other)
```
jerryyin: Why do you have to come up with the additional rvalue referenced constructor? I don't see an instance where you'd have to use move upon them. Plus this looks equivalent with the lvalue version above?

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: The way this math works is you split the thread ID into which CU wave (group of 16) you're on and which thread you're in within that wave. Then, both the larger waves and the threads within them form a 2D grid.

That is, which `m` you have is `tid / (cuwaveLen * nCuwavesPerBlock) + (tid % cuwaveLen) / nCuwavesPerBlock` , while the `n` is `(tid / cuwaveLen) % nCuwavesPerBlock + (tid % nThreadsPerCuwave)`

So we need to split the TID up into the 4-way grid (m_cuwaves x n_cuwaves x m_cuwave x n_cuwave) and then grab the relevant pair of numbers.

mlir/lib/Dialect/MIOpen/TransformMapBuilder.cpp
krzysz00: I'm adding it in case something calls the move constructor in the future. We don't want to move over dangling pointers (since `TransformMapBuilder` is self-referential)

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: This isn't equivalent to the old version. In the old version, there was an implicit thread offset being passed in from `gridwise_gemm`, which is now gone. Now, we have to compute our offset ourselves, which leads to the fiddlyness. I could go further here and make this a `tid x iteration number` function that then splits into `k`, `m_repeat`, `m_thread`, and `kpack`, but since the buffer layout is already `k x [m total] x kpack`, sticking to those coordinates made sense

mlir/lib/Dialect/MIOpen/TransformMapBuilder.cpp
jerryyin: If I remember it right, there's something like:
`TransformMapBuilder::TransformMapBuilder(TransformMapBuilder &&other)=delete;`

krzysz00: Which is incorrect. `startIndices` and `endIndices` hold pointers to data within `startNames` and `endNames` - if you naively move/copy the values in `startIndices` or `endIndices`, they'll point to junk data because `startNames` and `endNames` have moved.

jerryyin: When it is deleted, you'd not be able to compile if you call move on this class.

krzysz00: Right, I saw `=default` and not `=delete`. I'd argue for still *having* a move constructor, since there's nothing here we can't move, but I can delete it instead. I don't have a strong preference on this one.

jerryyin: It can go either ways: 
 - use =delete to make sure people cannot move it, or otherwise have to make the implementation right
 - Implement the move constructor correctly, which means you'll have to use move operator on all of the fields, unless the reference case which you customize the implementation to make sure it is moved correctly.

I think the existing implementation is deceiving in the sense that: If someone only read the header file and see a move constructor and try to use it to gain performance benefit, the move constructor actually is no different than the lvalue constructor. And actually he/she would pay the price of invoking a `std::move` to do this redundant type conversion.

krzysz00: I'd've expected the initialization syntax to do the moves, but apparently not? Will be deleting.

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
jerryyin: Just to make sure I understand. The builder previously is:
 - `k - <m_cuwaves, m_cuwave> - kpack`

Now, it is:
 - `k - < m_repeat, m_cuwaves, m_cuwave, m_thread > - kpack`

Since the dimension size of `m_cuwaves` and `m_cuwave` are the same, does that mean you added two dimensions of 1 into the transform?

Further down into `viewA`, it looks like you have: `k - < m_repeat, tid, m_thread > - kpack`. In which `tid` dimension is equivalent to `m_cuwaves, m_cuwave`?

Sorry if I'm asking noob questions on coordinate transfer, I don't quite see all transformations add up through the way it is composed.

krzysz00: Both transformations start from `k, m_repeat, tid, m_thread, kpack`
When walking our way towards the vector, we ignore `tid`.

When working towards matrix A, we split `tid` into `m_cuwaves, n_cuwaves, m_cuwave n_cuwave` and then drop the `n_*` ones in order to get a `m` coordinate. 

`m_cuwaves` and `m_cuwave` don't have the same dimension: although currently `m_cuwave` is always `4`, `m_cuwaves` is a function of `blockSize`

krzysz00: And previously we had `k, m_repeat, m_thread, kpack`, where `m_thread`'s initial value was based on the function currently being computed from `m_cuwaves` and `m_cuwave`.

jerryyin (approved):
LGTM, thanks!

PR #716, opened 2022-08-03
Remove dead code for dest_data_per_write

krzysz00:
Review of tuning parameters shows that the write to LDS is always
tuned to a scalar write. Therefore, all the code in
gridwise-gemm-to-blockwise designed to handle the non-scalar case is
dead and has been for a quite a while. So, remove the tuning
parameter and all the code that depends on it.

Also, get rid of a test that's been doing nothing for years.


zhanglx13 (approved):

jerryyin (commented):
I'm doubtful that it is the right thing to do in both fronts:

> shows that the write to LDS is always tuned to a scalar write. 

If you discovered that this is always turned to scalar probably we need to fix it rather than eliminate it. I believe that this is indeed a tuning parameter that can be adjusted on MIOpen end that has concrete benefit.

> Also, get rid of a test that's been doing nothing for years.

That gridwise gemm tests was come up with good intent and actually checks the arguments of `miopen.transform` to start with. Now, somebody ended up removing all the checks in an unnoticeable way (maybe for the convenience of pushing in some feature). I think the best action is to add back that check instead of getting rid of this test.

jerryyin (changes requested):
Considering this is already approved, I'll add a blocker in case it got in before discussion is done.

krzysz00:
@jerryyin The LDS write thing is a derived tuning parameter of ours - MIOpen can't touch it. The LDS writes vectorize anyway by way of the backend knowing how `load` in LLVM IR works, and so this code is very dead. It adds a bunch of complexity to the store-to-LDS logic that never gets used, and there's no clarity for how it might ever get used. It's getting in the way much more than anything.

On the test: I can't for the life of me think what that test is meant to do. The comments and the `RUN` line indicate it's about conv-to-gemm, but the body of the test passes in a gridwise_gemm invocation that hasn't been valid for a very long time.

jerryyin:
> The LDS writes vectorize anyway by way of the backend knowing how load in LLVM IR works

Ha, so this is a follow up of your vectorization derive PR (that instead of we derive it manually from tuning parameter, this is derived from the coordinate transfer infrastructure)? So effective what you mean is that we still **can** do lds write in vector but it isn't governed by this tuning parameter?

What about `srcDataPerRead`? Do you plan on to get rid of it as lds read should also be derived from your vectorization derive logic.

> On the test: I can't for the life of me think what that test is meant to do.

I checked on history a bit and think I mis-remembered and confused this one with https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/f6d2c6c34470545201ec77cbe3017f93b1de1bda/mlir/test/Dialect/MIOpen/lowering_top_level.mlir. You can see the header of the test is exactly similar to that one. Looks like it get copied over and no one ever did anything further.

It does look like a meaningless tests... Agreed on getting rid of it.

krzysz00:
So, this is an intermediate step in using vectorization derive logic for `srcDataPerRead`. I found this while trying to understand how `srcDataPerRead` and `destDataPerWrite` worked so I could replace all that with the vectorization derivation code and some coordinate transforms ... which led to the discovery that `destDataPerWrite` was 1 and has been since af70757947147c5b13b053ea908861f764630842 way back in the day.

Since the compiler vectorizes these stores awyway ... it seemed safe to remove.

krzysz00:
Basically, the `destDataPerWrite` hasn't been a function of tuning parameters for quite a while now, and since it's always 1 and that hasn't really impacted the quality of our generated code, we might as well get rid of the infrastructure for `destDataPerWrite` not being 1.

Unless @whchung wants to comment on the history here or @zjing14 (hopefully I got the right Jing) has context on the MIOpen implementation... I'd say we can ditch it

jerryyin (approved):
Could you file a ticket in which, it double check if there are any more vectorization opportunity that we haven't taken advantage of? I think we are on the same page that this existing way is not the best mechanism taking advantage of llvm infrastructure. So I'm less worried about your change here but more of leaving a reminder in case no one remembers/knows how well we are already vectorizing on lds

krzysz00:
Ok, so, wrt vectorization:

1. When we're using buffer load/buffer store, we have to do our own vectorizing - LLVM doesn't know how to combine those operations (and perhaps can't - there _might_ be soundness issues)
2. When it comes to regular loads/stores, LLVM has been doing a decent job - what do you want documented?

PR #717, opened 2022-08-04
Prefer TransposeOps as leading ops rather than trailing.

pcf000:
Generalise the TransposeOp special case to absorb all smallish constants.
Protect against operands with no defining op (eg, parameters).

The problem scenario in issue #611 is conv2d -> transpose -> elementwise ops -> transpose -> conv2d.  The second transpose is grabbed into the trailing ops of the first conv2d, which confuses the shape-inference pass later.  Avoid that by only accepting transposes into trailing ops if their defining op is an anchor;  that'll tend to place them in leading-ops for another anchor.

While we're in there, remove the special-case-for-transpose handling of the constant second operand, and instead generalise to grab any smallish constant-like op into leading or trailing ops.


sjw36:
Please add some testing to make sure we are getting what we want.

sjw36 (approved):
Looks good.

PR #718, opened 2022-08-08
Rework gridwise load to LDS (non-xdlops)

krzysz00:
- Introduce a `global_load` operation, analogous to current
`threadwise_copy_v2` (which will probably be renamed to `global_store`
in the future) that allows for vectorization >= 128 bits. This also
enables fusing with global loads in the future.
- Update vectorization inference to:
  - Fix handling of negative coefficients in Embed
  - Take advantage of unfold() marking
- Replace the hardcoded math for global load coordinates (and the
almost-identical LDS store coordinates). This also removes duplicate
code that does basically the same thing for matrices A and B. The
functions that are used to do this should be useable for the xdlops
gemm.

Notes:
- In the future, we may want to review the semantics of k_per_block in
the presence of kpack.
- This rewrite uses unKpack() as a temporary measure. kPack is a LDS
tuning parameter, and should not have been applied to the matrix
during gridwise gemm. A broader change will be a future PR.


krzysz00:
Manual nightly passed

krzysz00:
Also, in immediate impacts, the loads of the input tensor to the default convolution (`-p`) move from being unvectorized to having a 2x vectorization

krzysz00:
Between this and its dependency #714 , we have straight-up perf wins on non-xdlops (on gfx90a). For example, nhwc WRW perf is up on average of 7% for NCHW and 11% for NHWC. fp16 NCHW is up 8%, and we see improvements throughout - often an average of 3-4%. There are some benchmarking outliers that tanked the averages, but that always seems to happen with perf runs so I'm not too worried about them.

[global-refactor-perf-no-xdlops.zip](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/files/9283476/global-refactor-perf-no-xdlops.zip) contains the perf report.

krzysz00:
I except that the implementation of analogous changes in the xdlops path will bring similar performance improvements.

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -581,6 +580,66 @@ struct BlockwiseGemmV2RewritePattern
...
+    int64_t offset = 0;
+
+    Value result = createZeroConstantOp(b, loc, resType);
+
+    while (remainingLength > 0) {
+      int64_t copyLength = std::min(remainingLength, maxLoadLen);
+
+      // Clean up bad copy lengths
+      if (copyLength != maxLoadLen && copyLength > (2 * elemsPerWord))
+        copyLength = 2 * elemsPerWord;
```
jerryyin, line 613: I don't get the "2" part. Any reasoning to why we start trying from twice (no more, no less) length of elements per word?

I think you meant to start from the closest multiple of elemsPerWord every time? In that sense, if copyLength is 13 and maxLoadLen is 4, the maximum we can go is 3 * elemsPerWord = 12.

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -581,6 +580,66 @@ struct BlockwiseGemmV2RewritePattern
...
+      if (copyLength > elemsPerWord && copyLength % elemsPerWord != 0)
+        copyLength = elemsPerWord;
```
jerryyin, line 615: Now if `copyLength == 2 * elemsPerWord` then this condition check will make the previous one a no-op.

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
```diff
@@ -410,19 +409,30 @@ propagateVectorizationInfo(TransformMapAttr map,
...
             continue;
 
           int64_t upperLen = input[upperDim]->maxLength;
           int64_t needsCoeff = input[upperDim]->needsCoefficient;
 
-          if (!ourResult.hasValue())
+          if (!ourResult.hasValue()) {
             ourResult = VectorizationInfo(1, coefficient);
+            if (hasNegativeCoefficients)
+              break;
```
jerryyin, line 437: Replace this with `if (coefficient < 0) break;` so that we don't need to maintain that redundant bool flag?

mlir/test/Dialect/MIOpen/lowering_global_moves.mlir
```diff
@@ -53,3 +53,55 @@ func.func @miopen_threadwise_copy_v2_7xf16(%source : memref<32xf16, 5>,
     : memref<32xf16, 5> -> memref<64xf16>, index
   func.return
 }
+
```
jerryyin, line 56: Could we have one or two int8 test cases in global load?

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -261,6 +262,295 @@ static TransformingForOp createLdsStoreLoop(OpBuilder &b, Location loc,
...
+/// introduced in conv-to-gemm. kPack is a LDS-specific transformation and
+/// should not have been applied to the gridwise gemm argument. Future work will
+/// make the conversion stop doing that, and we'll be able to get rid of this.
+static Value unKpack(OpBuilder &b, Value matrix) {
+  ArrayRef<int64_t> shape = matrix.getType().cast<MemRefType>().getShape();
+  if (shape.size() != 4)
+    return matrix;
+  BottomUpTMBuilder unkpack(b, {"g", "k", "d", "kpack"}, shape,
+                            matrix.getLoc());
+  unkpack.passThrough({"g", "m"});
```
jerryyin: Do you meant to passThrough `"g", "d"`?

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -261,6 +262,295 @@ static TransformingForOp createLdsStoreLoop(OpBuilder &b, Location loc,
...
+  TransformMapAttr unkpackAttr = unkpack.get();
+  return b.create<TransformOp>(matrix.getLoc(), matrix, unkpackAttr);
+}
+
+/// Given a G x K x D matrix and the block tuning parameters and how much data
+/// each thread will load,
+/// return the dimension in which the load of this matrix from global memory
+/// should be vectorized and the length of that vector load. Also takes
+/// `tiebreaker`, the vectorization dimension to be used when both choises are
+/// equal, which should be the vectorization dimension of the store to LDS.
```
jerryyin, line 270: [question] I get the logic of trying to use the best vectorization between either K or MorN dimension. But I don't quite get it why we select `kpack > 1` to always fallback to K dimension when they are equal? Does it give us some concrete benefit or is it your personal preference for A and B to adopt vectorization from the same K dimension?

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -261,6 +262,295 @@ static TransformingForOp createLdsStoreLoop(OpBuilder &b, Location loc,
...
+  // That threads are grouped [other dim, k] is important: it menas we can
+  // ignore kPack here but then account for it when writing to LDS.
+  splitId.merge({dThreadName, "k_thread"}, {4, 5}, "tid", {dThreads, kThreads});
+  if (vectorDim == GemmDimension::K) {
+    splitId.merge({dIterName, "k_iter"}, {6, 7}, "iter",
+                  {dPerThread, kPerThread});
+  } else {
+    splitId.merge({"k_iter", dIterName}, {6, 7}, "iter",
+                  {kPerThread, dPerThread});
+  }
```
jerryyin, line 343: [Question] The way of the coordinate is used make me think this is a bottom up builder. The impression I have for topDownBuilder is that the coordinate transform start from `{"k_loop", "bid", "tid", "iter"}` and build to the actual target layout (like the ones we have in threadwisegemm lowering). Whereas this seems to be using it reversely?

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: Can we go for the closest multiple? I could've sworn we don't have load 3 words instructions ... or if we do, the `amdgpu` dialect doesn't allow for them. As far as I'm concerned, you can load 1, 2, or 4 words, or you can load a subword of the right size, but you can't, for example, pull in 3xf32 buffer load or a 3xi8

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: I'm not sure I follow.

(But yeah, if you call for a load of 3 or 7, you'll get a correction to a load of 2 above, and so this branch will be a noop. I expect the compiler to do the right thing here.)

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
krzysz00: We do need the flag, because this test is only being run when the current dimmension has vectorizatwion info attached. The negative coefficient might be on a dimesnion that's being assumed to be held constant.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: The tiebreak is based on the LDS layout. If we can't decide based on vectorization length, we want to vectorize using the dimension that matches the fastest-moving LDS dimension, which is m/n if kpack is 1 or k otherwise.

jerryyin (approved):
LGTM, I have kicked off the nonxdlops weekly CI: http://ml-ci.amd.com:21096/job/MLIR/job/test_weeklyCI/11/. Should be good to go once it has all pass.

---

Note: Looks like nonxdlops CI tuning has a full pass. This is ready to land.

PR #720, opened 2022-08-10
Fold duplicated dims before miopen fusion

jungpark-mlir:
Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/594
by eliminating shape change to the fusing in/out.

### DIFF
--- a/mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
@@ -497,11 +497,16 @@ LogicalResult LGDropDimsPattern::matchAndRewrite(linalg::GenericOp laGeneric,
   if (laGeneric.outputs().size() > 1)
     return failure();
   MemRefType regType;
-
+  Value laOut;
   // Check memref alloc as output and eliminate it with all users
   for (Operation *user : out.getUsers()) {
     if (isa<memref::CollapseShapeOp>(user)) {
-      regType = user->getResult(0).getType().template cast<MemRefType>();
+      for (Operation *copyOp : user.getUsers()) {
+        regType = user->getResult(0).getType().template cast<MemRefType>();
+        if (isa<memref::CopyOp>(copyOp)) {
+          laOut = copyOp->getTarget();
+        }
+      }
       eliminateAll(user);
       break;
     }
@@ -529,8 +534,6 @@ LogicalResult LGDropDimsPattern::matchAndRewrite(linalg::GenericOp laGeneric,
     }
   }
 
-  Value laOut;
-  bool bFirst = true;
   // Reconfigure the linalg.generic, this tries to restore the original rank
   // which was already compatible elementwise fusion duplicated some dims.
   for (auto pair : llvm::zip(idxMaps, laGeneric.inputs())) {
@@ -547,18 +550,12 @@ LogicalResult LGDropDimsPattern::matchAndRewrite(linalg::GenericOp laGeneric,
         newValue = inp;
       }
       newInputs.push_back(newValue);
-
-      // The first input comes from the anchor op and to be used as the final
-      // output
-      if (bFirst)
-        laOut = newValue;
       laGenericAMaps.push_back(
           AffineMap::getMultiDimIdentityMap(commonRank, ctx));
     } else {
       newInputs.push_back(inp);
       laGenericAMaps.push_back(inpIdxMap);
     }
-    bFirst = false;
   }
 
   // set output map and in/out
--- a/mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
@@ -59,6 +59,13 @@ struct MILARewritePattern : public OpRewritePattern<linalg::GenericOp> {
   LogicalResult matchAndRewrite(linalg::GenericOp laGeneric,
                                 PatternRewriter &b) const override;
 };
+
+struct LGDropDimsPattern : public OpRewritePattern<linalg::GenericOp> {
+  using OpRewritePattern<linalg::GenericOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(linalg::GenericOp laGeneric,
+                                PatternRewriter &b) const override;
+};
 } // end anonymous namespace
 
 /// If `inpMap` is a map of the form
@@ -172,10 +179,10 @@ static void insertCopyFromOtherArg(PatternRewriter &b, Location loc,
                                    Value dest) {
   LLVM_DEBUG(llvm::dbgs() << "Src type: " << srcOp.getType()
                           << " dest type: " << op.dest().getType() << "\n");
-  assert(srcOp.getType().cast<ShapedType>().getShape() ==
-             op.dest().getType().cast<ShapedType>().getShape() &&
-         "shape of extra fusion arguments matches shape of C tensor");
-
+  /*  assert(srcOp.getType().cast<ShapedType>().getShape() ==
+               op.dest().getType().cast<ShapedType>().getShape() &&
+           "shape of extra fusion arguments matches shape of C tensor");
+  */
   auto writeCLoop = op->getParentOfType<TransformingForOp>();
   assert(writeCLoop && "threadwise_copy_v2 must be in a transforming_for");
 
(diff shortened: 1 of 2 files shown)


krzysz00 (changes requested):
Preemptively, I'd like more of an explanation of what this is doing and why you needed to comment out a rather useful assert. I'll have more detailed thoughts tomorrow.

jungpark-mlir:
Currently it's not working as I set a wrong output. @krzysz00  I'll give you more details once this got fixed, also I'm not sure about the assertion and need to discuss it with you.

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -172,10 +179,10 @@ static void insertCopyFromOtherArg(PatternRewriter &b, Location loc,
                                    Value dest) {
   LLVM_DEBUG(llvm::dbgs() << "Src type: " << srcOp.getType()
                           << " dest type: " << op.dest().getType() << "\n");
-  assert(srcOp.getType().cast<ShapedType>().getShape() ==
-             op.dest().getType().cast<ShapedType>().getShape() &&
-         "shape of extra fusion arguments matches shape of C tensor");
-
+  /*  assert(srcOp.getType().cast<ShapedType>().getShape() ==
```
krzysz00, line 182: This assertion is, IIRC, a key invariant in the code below - it's there for good reason. If you're removing this assertionn (which should be removed, not comment out) please explain how either
1. The asserted invariant did not need to be true; or
2. You have modified the code below to relax the invariant

One note I have is that, in the code as I remember writing it, the fix for the extra argument's shape not matching the output tensor's shape was to `TransformOp` said extra argument tensor until it's shape matched. Why can this approach not be used for this new bug?

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -469,9 +476,109 @@ LogicalResult MILARewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
   return failure();
 }
 
+void eliminateAll(Operation *op) {
```
krzysz00, line 479: The existence of this method bothers me

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -469,9 +476,109 @@ LogicalResult MILARewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
...
+  for (Operation *user : op->getUsers()) {
+    eliminateAll(user);
+  }
+  op->dropAllUses();
+  op->erase();
+}
+
+// Dropping expanded dims from linalg elementwise fusion
+LogicalResult LGDropDimsPattern::matchAndRewrite(linalg::GenericOp laGeneric,
```
krzysz00, line 488: Please explain why this can't be done during the main fusion process. Furthermore, please explain why this sort of expansive rewriting of the generic, instead of a more targeted transformation, is required.

krzysz00 (changes requested):
Now that I'm properly at work, my "ugly hack" sensor is still going off, and I'd like to see justifications for the ugly hack.

In comments that don't fit onto a line of code, can I get an example of what the input to the fusion pass is (not necessarily the whole GEMM - the writeback loop and the `linalg.generic` being fused in are sufficient) and what goes wrong on `miopen-dialect`?

jungpark-mlir:
@krzysz00 Sure, I agree this looks suspicious in many ways and open to discussion. Apparently some parts could be improved.
Let me start from the problem description.

This happens when we try to fuse an element-wise operator with broadcast, fusion fails in AlignTiling.

Example)
when the shape of the convolution is : memref<1x30x30x16xf32>
and we add a bias with this shape : memref<16xf32>, expanded to memref<1x1x1x16xf32>

** The problem is in the linalg pass ElementwiseOpFusion, it expands the dims to memref<1x30x30x1x1x1x16xf32> even though it's compatible with the broadcasted bias.

```
    %5 = memref.expand_shape %0 [[0], [1], [2], [3, 4, 5, 6]] : memref<1x30x30x16xf32> into memref<1x30x30x1x1x1x16xf32>
    %6 = memref.alloc() {alignment = 128 : i64} : memref<1x30x30x1x1x1x16xf32>
    linalg.generic {indexing_maps = [#map1, #map2, #map1], iterator_types = ["parallel", "parallel", "parallel", "parallel", "parallel", "parallel", "parallel"]} ins(%5, %4 : memref<1x30x30x1x1x1x16xf32>, memref<1x1x1x16xf32>) outs(%6 : memref<1x30x30x1x1x1x16xf32>) {
    ^bb0(%arg4: f32, %arg5: f32, %arg6: f32):
      %8 = arith.addf %arg4, %arg5 : f32
      linalg.yield %8 : f32
    }
```

As you know, the new simplified fusion doesn't expect any transform to the input and can't handle this.

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
jungpark-mlir: That's fair, I'll remove it

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
jungpark-mlir: I want 2, because this change is trying to support the case when the extra arg's shape is different to the output shape.
Instead, can we check if they're compatible?

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
jungpark-mlir: It might be possible to include this into the main fusion align but I'm trying not to.
The reason is, I like the way you changed the fusion recently and want to keep the main fusion as simple as possible.

Linalg already has very similar pass but it doesn't perfectly fit to my purpose.
https://mlir.llvm.org/docs/Passes/#-linalg-fold-unit-extent-dims-remove-unit-extent-dimension-in-linalg-ops-on-tensors

I don't think this is a hacky solution but a normal use case of the multi-level compilation to tame the IR from different level of compilation. Main point is to completely separate this pattern from the main fusion and let it handle the broadcast (and other transforms if we encounter in the future) and main fusion pattern to focus on optimizing the memory read/write.
Maybe there is still a better way to apply this transform as you mentioned above, please let me know.

krzysz00:
We _do_ support transforming the extra fusion arg (though we don't support transforming the output tensor). See https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/miopen-dialect/mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp#L71

krzysz00:
If it's the output tensor being expanded by the `linalg` passes, I'd go up and add a `miopen.transform` around the output tensor to get it back into the form the convolution expects. Then, we'd be using the same broadcasting logic we already have to handle the bias tensor.

Though if it's a bias tensor being expanded, we use the same "add a transform" trick that we used to handle there being too few bias dimensions to handle the case where there's too many. We already have the infrastructure for that and shouldn't add a new rewrite for it.

jungpark-mlir:
OK, let me try adding transform to the output.

krzysz00:
To rephrase:

1. The input to conv-to-gemm should have the layout specified in the convolution operation
2. That aside, you should prefer transforming the bias tensor to match the shape of the final output tensor, which is what the existing code does.

jungpark-mlir:
Agree to the idea.
But actually the linalg pass transforms the convolution output itself and we're trying to fold the broadcast into the linalg generic.
Another difficulty is, linalg passes mess up the dimensions in tensor and there's not many option in that stage.

Anyway, 
While I was looking for a solution in the earlier stage, I found more interesting clue about the root cause.
It turns out the root cause is TosaToLinalg, not LinalgElementwiseFusion. There was a hidden implicit insertion of a tosa.reshape in TosaToLinalg pass which causes all this unexpected dimensions.

Fix for that is only few lines, will come to another PR soon.

krzysz00:
Well then if the convolution output is having its shape expanded, maybe add a thing to conv-to-gemm that replaces shape expansion with an equivalent `miopen.transform`

Which I thought we already had but maybe it broke

jungpark-mlir:
#729 is opened instead.

PR #722, opened 2022-08-10
[DO NOT SQUASH] Rework MFMA API

krzysz00:


krzysz00:
Manual nightly run passes (earlier issue was a typo) and this is now mergeable

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -305,49 +306,100 @@ static Value mfmaConcatIfNeeded(ConversionPatternRewriter &rewriter,
...
+/// on the architecture you are compiling for.
+static Optional<StringRef> mfmaOpToIntrinsic(MFMAOp mfma, Chipset chipset) {
+  uint32_t m = mfma.m(), n = mfma.n(), k = mfma.k(), b = mfma.nBlks();
+  Type sourceElem = mfma.sourceA().getType();
+  if (auto sourceType = sourceElem.dyn_cast<VectorType>())
+    sourceElem = sourceType.getElementType();
+  Type destElem = mfma.destC().getType();
+  if (auto destType = destElem.dyn_cast<VectorType>())
+    destElem = destType.getElementType();
+
```
jerryyin, line 320: Could we rename `nBlks` to `outBlks`, and instead of use `nBlks` as input, use it as a derived parameter from `bool reduction`. I think this is going to impact a number of things from your PR.

Derive m from n or n from m, make sure they are always equivalent.

We can compute the nBlks using: 
 - If nonreduction, this must satisfy: `nBlks x m(or n) == 64`
   - If chipset >= 0x40, `nBlks x m(or n) == 128` 
 - if reduction: `nBlks = 1`

For example, in fp32:
 - 32x32: nBlocks = 2
 - 16x16x1 (nonreduction): nBlks = 4
 - 16x16x4 (reduction): nBlks = 1

krzysz00: I'm not sure that if statement is accurate - many of the gfx940 ops still have reduction-less ops that have (m/n) * nBlks = 64

krzysz00: And there are reduction cases where `nBlks` is not 1, IIRC. Also, the number of input and output blocks is always equal, so `nBlks` (or `blocks`, or `numBlocks`) is better than `outputBlocks`

krzysz00: Ah, concrete example of k > 1, nBlks > 1, `mfma_f32_32x32x4f16`

jerryyin: > I'm not sure that if statement is accurate - many of the gfx940 ops still have reduction-less ops that have (m/n) * nBlks = 64

That's right, I guess for backward compatibility purpose it has to make a shorter argument xdlops available.

> Ah, concrete example of k > 1, nBlks > 1, mfma_f32_32x32x4f16

For this case m = 32, nBlks = 2, so it satisfies 32 x 2 = 64.

> And there are reduction cases where nBlks is not 1, IIRC. Also, the number of input and output blocks is always equal.

For existing code base, input blocks and output blocks means different things and can be unequal (xdlops selection). For example, 
 - mfma_f32_16x16x4f16 is considered to have 4 input blocks and 4 output blocks
 - mfma_f32_16x16x1f16 is considered to have 4 input blocks and 1 output block, that's what made it an reduction versus the first one.

So when you say input and output blocks are always equal which part of code base are you referring to?

Do you see a straightforward way to incorporate `reduction` as a parameter at this abstraction layer?

krzysz00: I'm referring to the individual `mfma` instructions themselves. They have parameters `m`, `n`, `k`, `blocks` (which is probably a better name than `nBlks`) and types. 

So, the point of this whole setup is that an individual MFMA is not selected by name but instead by m/n/k/blocks, which makes upstream happier.

I could absolutely see `bool reduction;` being a basic object at the level of xdlops selection, but I don't think it'll work correctly as part of `amdgpu.mfma`.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -467,221 +441,62 @@ struct XdlopsCodeSelection {
...
-      cycles = 32;
-      k_base = 4;
-    } else {
-      llvm_unreachable("Unsupported case as mfmaInstr not selected!\n");
-    }
+    // Rename this field in future refactoring
+    int64_t vectorNumber = imms.size();
+    int64_t numXdlops = vectorNumber / (MRepeats * NRepeats);
+    int64_t mCandidate = MPerXdlops / numXdlops;
+    int64_t nCandidate = NPerXdlops / numXdlops;
```
jerryyin: `MPerXdlops * numXdlops` and `NPerXdlops * numXdlops`?

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -467,221 +441,62 @@ struct XdlopsCodeSelection {
...
+    int64_t mfmaOutDim = std::min(mCandidate, nCandidate);
```
jerryyin: I understand `mCandidate` as "number of elements in m direction"? Is that correct? If that's the case, then could you rename `mCandidate` to `mLen`?

I don't quite get what `mfmaOutDim` is. For A * B and there's a broadcast in A, then `mCandidate == 64` and `nCandidate == 32`, why is `mfmaOutDim` defined as the minimal of the two?

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -467,221 +441,62 @@ struct XdlopsCodeSelection {
...
+    int64_t m = mfmaOutDim, n = mfmaOutDim;
+
+    int64_t kPerInputArg =
+        math_util::integer_divide_ceil(waveSize, mfmaOutDim * blocksMfma);
```
jerryyin: `KPerInputArg` is effectively `kLen` according to above naming convention? But then `kLen` should be different per A or B matrix (if broadcast happens, then one matrix effectively need only 1/block `k`s compared the other)?

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -467,221 +441,62 @@ struct XdlopsCodeSelection {
...
+    // k_base is the length of the step you must take in the k dimension of
+    // your input when preparing the next argument on a thread.
+    int64_t k_base = k / kPerInputArg;
+
+    int64_t nInputs = (m * blocksMfma * k) / waveSize;
```
jerryyin: Question, are all `m` and `n` interchangeable in the computations below? I was assuming you are referring to m, n, k definition in the mfma definition, not the actual `m`, `n`, and `k` that account for broadcast. In that regard m and n are always the same, therefore the question.

And if that's the case, I'd prefer to consistently use only m or n, in case it get confusing on whether it matters to only A or only B matrix.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -467,221 +441,62 @@ struct XdlopsCodeSelection {
...
+
+    int64_t nOutputs = (m * n * blocksMfma) / waveSize;
+    Type vectorElem;
+    if (dataType.isa<IntegerType>())
+      vectorElem = b.getI32Type();
+    else
+      vectorElem = b.getF32Type();
+    VectorType vectorType = VectorType::get({nOutputs}, vectorElem);
+
+    constexpr int64_t group_size = 4;
```
jerryyin: I might not have comprehended in the old definition what `group_size` is and why it is always 4, mind some explanations?

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
jerryyin: Gotcha, so blocks is effectively `num_input_blks` in the old context.

> but I don't think it'll work correctly as part of amdgpu.mfma

No I think you are good. I was debating yesterday whether blocks is the more "fundamental" information but now I see each can be derived from the other parameter. Plus reduction does sound like a derived information on the abstraction level higher than mfma instructions.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
krzysz00: Nope. That would just be `MPerWave`/`NPerWave`. These are the possible values for dimension of each `mfma` instruction (bearing in mind all MFMAs have the same m and n dimension) based on the values of `MPerXdlops`/`NPerXdlops`

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
krzysz00: mCandidate is the number of m elements _for each MFMA_ if we take MPerXdlops as the source of that number. Similarly for `nCandidate`.

`mfmaOutDim` is the m and n dimension of each MFMA instruction.

The reason we take the minimum is the 4x4 instructions. If we set `MPerXdlops = 4, NPerXdlops = 64` and have one MFMA, we need the MFMA to be a 4x4xK one, hence `mfmaOutDim = min(4, 64) = 4`

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
krzysz00: Nope, kPerInputArg is a property of the MFMA and is the same between A and B. It's the number of different values for k that can appear in a single MFMA input across all threads.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
krzysz00: Yeah, the m and n are interchangable, and could've been called mfmaOutDim instead, but that's longer.

I could go ahead and make them all `mfmaOutDim` (or maybe that's `mfmaNonKLen` or something ... names are hard)

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
krzysz00: group_size is the number of registers in a register tile of the MFMA output - it's always been fixed to 4, now that's not a function of if statements.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
jerryyin: I think the naming is becoming more confusing here.
 - The previous defined `MPerXdlops` should be just `MPerWave`
 - Now defined `mCandidate` should be `MPerXdlops`

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
jerryyin: Thanks once you clarified the meaning of it it all starts to make sense.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
jerryyin: Is `KPerInputArg` effectively `KPerXdlops`? But then `kPerXdlops` is effectively k for mfma instruction. This make me confused on why you don't directly take the k?

Let's take an example of `f32_16x16x4f16`: It does mfma of 16x4 * 4x16 for register A and register B. kPerInputArg should be 4 for both A and B.

Now according to your computation: `kPerInputArg = 64 / (16 * 4) =1`.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
jerryyin: I actually think it is better if you just use `mfmaOutDim` consistently without using `m` or `n`.

So here, `nInputs` is the actual `k` for mfma instruction's k?

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
jerryyin: Gotcha now.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -467,221 +441,62 @@ struct XdlopsCodeSelection {
...
+    int64_t rowsPerOutput = std::min(waveSize / n, m / group_size);
```
jerryyin: Is this effectively `vectorNumber`?

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -467,221 +441,62 @@ struct XdlopsCodeSelection {
...
+    int64_t blocksPerOutput =
+        math_util::integer_divide_ceil(waveSize, rowsPerOutput * n);
```
jerryyin: I start to get lost form here on again...

Could I ask you to draw a 1:1 reference mapping according to just one sample of xdlops?

Something like `16x16x4f16` with  below would work:

image-20220121163541641

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
krzysz00: Nope, `nInputs` is the number of logical values needed to hold the arguments to the MFMA - it's the number of inputs.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
krzysz00: Nope. `vectorNumber` operates at a higher level of loopng.

krzysz00:
Re `kPerInputArg` the 16x16x4xf16 MFMA takes four blocks, and so there are 16 * 4 * 4 = 256 total input values, which get split into 4 input arguments (each input argument has 64 values). Those values are arranged in [k, block, m] order, and so each input argument contains one k, hence kPerInputArg = 1

krzysz00:
I can work on a diagram on Thursday, but to clear up terminology as I've been using it:
- The MperWave/NPerWave are implemented with one ore more *xdlops*, each of which consists of one or more MFMA instructions. (Yes I know an xdlop is a hardware thing, but that's not how we use the word)
- You have multiple xdlops for one (M/N)PerWave by setting the repeats to be > 1

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -467,221 +442,68 @@ struct XdlopsCodeSelection {
...
+    // and N are needed for each MFMA.
+    int64_t mfmaNonKDim = std::min(MPerXdlops, NPerXdlops) / mfmasPerXdlops;
+
+    int64_t kPerInputArg =
+        math_util::integer_divide_ceil(waveSize, mfmaNonKDim * blocksMfma);
+    // k_base is the length of the step you must take in the k dimension of
+    // your input when preparing the next argument on a thread.
+    int64_t k_base = k / kPerInputArg;
+
+    // Number of logical values neened to hold the niputs to each MFMA.
```
jerryyin: ... to each MFMA. `->` ... to each thread in its MFMA invokation.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -467,221 +442,68 @@ struct XdlopsCodeSelection {
...
-      llvm_unreachable("Unsupported case as mfmaInstr not selected!\n");
-    }
+    // Rename this field in future refactoring
+    int64_t vectorNumber = imms.size();
+    int64_t mfmasPerXdlops = vectorNumber / (MRepeats * NRepeats);
+    // Since one of MPerXdlops or NPerXdlops can implicitly encode the usage of
+    // broadcasting by the MFMA sequence underlying the xdlops, we take the
+    // smaller of those two values in ordet to determine how many valuse for M
+    // and N are needed for each MFMA.
+    int64_t mfmaNonKDim = std::min(MPerXdlops, NPerXdlops) / mfmasPerXdlops;
```
jerryyin: This sounds like a static (non-calculated) value to me. `mfmaNonKDim` sounds like a fundamental value used to derive others, not to be derived from how we define the aggregation of several mfmas.

jerryyin: I think this follows a number of other deriving process: To actually check how to set the parameter of a 64 x 64 convolution, we start from a single xdlops instruction and work its way up. However, a lot of the calculation use the opposite direction to take the final product of the derived results and work its way down. If we want to support a new tile size it will make it very hard to come up with those fields of the first big if statement without the inherent values defined in each independent xdlops instruction.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -467,221 +440,86 @@ struct XdlopsCodeSelection {
...
+
+    // Number of logical values each thread needs to pass in to the MFMA in
+    // order for the correct number of input values to be passed to the MFMA.
+    int64_t nInputsToMfma = (mfmaNonKDim * blocksMfma * k) / waveSize;
+    Type argType = nInputsToMfma == 1
+                       ? dataType
+                       : VectorType::get({nInputsToMfma}, dataType);
+
+    int64_t nOutputsOfMfma =
+        (mfmaNonKDim * mfmaNonKDim * blocksMfma) / waveSize;
```
jerryyin, line 458: I know we have discussed about this. But could you remind me that for reduction xdlops, does the result occupy the same amount of vgprs as non-reduction xdlops? Intuitively, since reduction calculate less end result, it shouldn't need to occupy that many?

I think it either occupy same amount of registers in same amount of lanes as the non-reduction xdlops, but results are duplicated. Or it occupy less amount of registers in smaller number of lanes and results are all unique.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -467,221 +440,86 @@ struct XdlopsCodeSelection {
...
+    Type vectorElem;
+    if (dataType.isa<IntegerType>())
+      vectorElem = b.getI32Type();
+    else
+      vectorElem = b.getF32Type();
+    VectorType vectorType = VectorType::get({nOutputsOfMfma}, vectorElem);
+
+    constexpr int64_t group_size = 4;
+    // The number of rows in each MFMA result:
```
jerryyin: I think it's better if you add the definition of rows in the context before the implementation details section.

Something like, within a single output block, how many divisions of *non-contiguous* chunk of data across the two output vgprs.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -467,221 +440,86 @@ struct XdlopsCodeSelection {
...
+    // that you can fit into a wave (ex. a 32x32x2 mfma can fit two rows per
+    // result). However, in the case of 4x4xk (16 blocks) MFMAs, counting that
+    // number would produce the inaccurate result 64 / 4 = 16 rows per output,
+    // even though there is only one row available to place in each output
+    // (remembering that the rows are first tiled into groups of group_size
+    // outputs). Therefore, we need to impose the bound m_mfma / group_size on
+    // the number of rows per output.
+    int64_t rowsPerOutput = std::min(waveSize / /*n_mfma=*/mfmaNonKDim,
+                                     /*m_mfma=*/mfmaNonKDim / group_size);
+    // The number of blocks in each MFMA output. If rowsPerOutput followed the
```
jerryyin: ... in each MFMA output *vgpr*.

I think adding the scope of `vgpr` will point more clear at what the output unit is.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -467,221 +440,86 @@ struct XdlopsCodeSelection {
...
+    // typical case and was computed using waveSize / n_mfma, this will be 1.
+    // However, in the 4x4 case, where we do have 16 blocks packed into each
+    // output blocksPerOutput will be > 1 (namely 16).
+    int64_t blocksPerOutput = math_util::integer_divide_ceil(
+        waveSize, rowsPerOutput * /*n_mfma=*/mfmaNonKDim);
+    // The number of register groups (of four rows) per block of output
```
jerryyin, line 487: Note that I think the concept of `row` here is different than `rowsPerOutput` because here the 4 rows are contiguous and in the register dimension direction only (Previously it was in lane direction and non-contiguous). I suggest to clarify or do not use the `rows` terminology.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -467,221 +440,86 @@ struct XdlopsCodeSelection {
...
+    // Note that the inclusion of blocksPerOutput forces this value to be 1 in
+    // the 4x4 case, as it should be.
+    int64_t num_groups_blk = math_util::integer_divide_ceil(
+        /*m_mfma=*/mfmaNonKDim, group_size * rowsPerOutput * blocksPerOutput);
+    int64_t num_regs_blk = group_size * num_groups_blk;
+    int64_t num_output_blks = blocksMfma / blocksPerOutput;
+    int64_t num_regs_xdlops = num_regs_blk * num_output_blks;
```
jerryyin: I think there's a degree of which we are using xdlops and mfma interchangeably in naming, shall we use only one? Or add a comment at the very top to clarify xdlops = mfma.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -467,221 +440,86 @@ struct XdlopsCodeSelection {
...
+    assert(nOutputsOfMfma == num_regs_xdlops &&
+           "Registers per xdlops equal to vector size");
+
+    // Properties of th selected bundle of MFMA instructions, or of how they
+    // are used to implement GEMM.
+
+    // Rename this field in future refactoring
+    int64_t vectorNumber = imms.size();
+    int64_t MPerXdlops = MPerWave / MRepeats;
+    int64_t NPerXdlops = NPerWave / NRepeats;
```
jerryyin: Per our discussion before, I'd prefer those fields to be calculated based on xdlops own properties, not based on a group of mfmas. I think the right derive order is that knowing `MPerXdlops` and `MRepeats`, we derive `MPerWave` by multiplying two together.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -467,221 +440,86 @@ struct XdlopsCodeSelection {
...
+    // lane in a wave.
+    int64_t k_base = k / kPerInputArg;
+
+    // Number of logical values each thread needs to pass in to the MFMA in
+    // order for the correct number of input values to be passed to the MFMA.
+    int64_t nInputsToMfma = (mfmaNonKDim * blocksMfma * k) / waveSize;
+    Type argType = nInputsToMfma == 1
+                       ? dataType
+                       : VectorType::get({nInputsToMfma}, dataType);
+
```
jerryyin, line 456: I'm giving a summary of my understanding of output section variable definition. Please feel free to adjust and attach them in the beginning of the output section. Also, correct me if any of those are ambiguous.

- nOutputsOfMfma: Total number of vgprs across lane 0 to lane 63
  - Note, it looks like `num_regs_xdlops` is equivalent to this, please get rid of one or another
- rowsPerOutput: Number of divisions of non-contiguous chunk of data across the lane direction in a single output block. 
- blocksPerOutput: Number of output blocks across the lane direction
- num_groups_blk: Number of groups of 4 contiguous chunk of data across the register direction in a single output block.
- num_regs_blk: Number of output registers in a single output block
- num_output_blks: Number of output blocks (same concept versus the blocksMfma for input).
   - I'd suggest to unify the naming convention between input and output here

jerryyin (commented):
No functional obstructions, but sorry, more nitpicks after finally understanding the full picture. I'd advise more renaming/comments just so that a 3rd person can go in and hopefully take less effort in figuring out what's going on behind those computations.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
krzysz00: Reduction occupies fewer registers than the corresponding non-reduction, yeah

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
krzysz00: Except we don't know `MPerXdlops`, we just know `MPerWave` and we selected a `mfmaNonKDim` and an `MRepeat` from it

jerryyin: Ahh, if that's the case, can we make `MPerXdlops` and `NPerXdlops` the fundamental parameters and `MRepeats` and `NRepeats` to be derived? 

It makes me feel very un-intuitive to have known the repeat as the known condition to derive back at how much does one xdlops consumes. It makes more sense that we have already decided which xdlops instruction we pick to execute this (by specifying the PerXdlops lengths), then, repeat is just about how many times on each direction to fill the gap.

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
jerryyin: That makes sense, thanks! So `nOutputsOfMfma` has the special condition to not be completely filled up towards lane 63, but we occupy the same amount of registers regardless. Since you are describing the number of registers occupied, it doesn't matter for this computation.

krzysz00: There is filling up to lane 63. The computation is this: We know each MFMA will produce mfmaNonKDim (aka m_mfma) * mfmaNonKDim (aka n_mfma) * blocksMfma outputs, and that those outputs will be put into some number of outputs, with each output containing waveSize (aka 64) values.

jerryyin: Gotcha now, thank you!

jerryyin (approved):
I can understand the general idea behind the XdlopsCodeSelection now (with all the QAs across this PR), but I still think additional pictures are due to completely describe the idea behind all computations. I'm not too worried based on: 1. XdlopsSelection is not in its final form yet. 2. The component is very scoped to itself.

Approved, overall this PR improved the Mfma intrinsic by providing more useful metadata around it.

PR #723, opened 2022-08-11
[DO NOT SQUASH] Support kernel generation for multiple target chips and select at runtime

sjw36:
MLIR#616: enhance ExecutionEngines with a system detect
* including ROCm system detect
* add 'async' dialect name to "targets" attribute

MLIR#614: support multiple targets in compilation/xmodel/runner
* updated miopen-clone-kernes and miopen-apply-impls
* shift toward 'chip' instead of 'arch'

* use RocmSystemDetect in xmir-runner

https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/614
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/616


sjw36:
See test xmir/e2e/resnet50_blk.e2e.mlir which now builds kernels for multiple targets and xmir-runner selects the appropriate chip at runtime.

krzysz00:
Review will happen tommorow, probably (just thought I'd let you know I've seen this)

sjw36:
I am on PTO for a week so this can wait.

external/llvm-project/mlir/include/mlir/ExecutionEngine/SystemDetect.h
```diff
@@ -0,0 +1,54 @@
...
+// This file provides a system detection interface with default CPU detection
+// for the execution engine.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef MLIR_EXECUTIONENGINE_SYSTEMDETECT_H_
+#define MLIR_EXECUTIONENGINE_SYSTEMDETECT_H_
+
+#include <string>
+#include <unordered_map>
```
krzysz00: Why not use LLVM data structures here?

external/llvm-project/mlir/include/mlir/ExecutionEngine/SystemDetect.h
```diff
@@ -0,0 +1,54 @@
...
+/// SystemDetect finds devices on the current system.
+///
+
+struct SystemDevice {
+  enum class Type : int32_t { ECPU, EGPU, ENPU, EALT };
+
+  Type type;
+  uint32_t count;
+  std::string chip;
+  std::unordered_map<std::string, std::string> properties;
```
krzysz00: For example, this is almost certainly better as a `StringMap`

external/llvm-project/mlir/include/mlir/ExecutionEngine/SystemDetect.h
```diff
@@ -0,0 +1,54 @@
...
+
+  SystemDevice(Type _type = Type::ECPU, const std::string &_chip = {});
```
krzysz00: StringRef, maybe?

external/llvm-project/mlir/include/mlir/ExecutionEngine/SystemDetect.h
```diff
@@ -0,0 +1,54 @@
...
+};
+
+class SystemDetect : public std::unordered_map<std::string, SystemDevice> {
```
krzysz00: Why inherit from maps? I'd argue it'd be better to provide an API around the map (esp. if we move to the llvm container, which is, last I checked, broadly preferred over std equivalents in a lot of cases)

external/llvm-project/mlir/include/mlir/ExecutionEngine/SystemDetect.h
```diff
@@ -0,0 +1,54 @@
...
+public:
+  static const SystemDetect &get();
+
+  virtual ~SystemDetect() {}
+
+protected:
+  SystemDetect();
+
+  SystemDevice *addDevice(SystemDevice::Type _type, const std::string &_chip,
```
krzysz00: Returning pointers is raising red flags. As does this method in general. What if this is called in a multi-threaded context?

We might want to look into mlir contexts and `StorageUniquer` here.

external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt
```diff
@@ -29,15 +32,20 @@ add_mlir_library(MLIRExecutionEngineUtils
...
+  MC
   ScalarOpts
+  Target
   Vectorize
   TransformUtils
   IPO
   Passes
   )
+  
+if(CMAKE_SYSTEM_NAME MATCHES "Linux")
```
krzysz00, line 50: Why?

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
```diff
@@ -0,0 +1,73 @@
...
+      std::string features;
+
+      auto idx = chipName.find_first_of(':');
+      if (idx != std::string::npos) {
+        features = chipName.substr(idx + 1);
+        chipName.erase(idx);
+      }
+
+      auto *dev = addDevice(SystemDevice::Type::EGPU, chipName);
+      dev->properties.insert(
```
krzysz00: Couldn't this property list be passed to the constructor?

mlir/lib/Dialect/MIOpen/Transforms/ApplyImpl.cpp
```diff
@@ -37,53 +38,63 @@ struct MIOpenApplyImplPass
...
+              if (auto kernelFunc = mod.lookupSymbol<func::FuncOp>(attr)) {
+                std::vector<NamedAttribute> attributes{
+                    b.getNamedAttr("type", b.getStringAttr("gpu")),
+                    b.getNamedAttr("arch", gpuMod->getAttr("arch")),
+                    b.getNamedAttr("grid_size", func->getAttr("grid_size")),
+                    b.getNamedAttr("block_size", func->getAttr("block_size")),
+                    b.getNamedAttr("binary", binaryAttr)};
+
+                kernelImpls[kernelFunc].push_back(
+                    b.getDictionaryAttr(attributes));
```
krzysz00, line 74: I'm not sure this here bit of nested attribute spaghetti is the right approach.  I'd at least like to see a `GpuKernel` attribute that combines grid_size, block_size, and binary. Then we could map from `(SystemType, StringAttr) -> Attribute` , where for a GPU kernel, the attribute would be a `GPUKernel` attribute.

This would also probably mean making `SystemType` a tablegen'd enum, so you can have a `SystemTypeAttr` but I'd say that's a solid thing anyway.

mlir/lib/Dialect/MIOpen/Transforms/CloneKernels.cpp
```diff
@@ -28,50 +28,65 @@
 #include "llvm/ADT/SmallVector.h"
 
 using namespace mlir;
+using namespace llvm;
```
krzysz00: I think instead of this, there's a header you can include that reexports most of the common llvm bits, like StringRef, into MLIR. `using namespace llvm;` has caused too many cases of otherwise-unneeded prefixing, IMO.

mlir/lib/Dialect/MIOpen/Transforms/CloneKernels.cpp
```diff
@@ -28,50 +28,65 @@
...
       return;
 
-    auto *ctx = &getContext();
-    auto gpuMod = mod.lookupSymbol<ModuleOp>(modName);
-    if (!gpuMod) {
-      OpBuilder b(ctx);
-      // create a GPUModuleOp in case the GPU module specified does not exist.
-      gpuMod = b.create<ModuleOp>(mod.getLoc(), modName);
+    for (auto chip : chips) {
+      std::string modName = ("__kernel_" + chip).str();
```
krzysz00: Maybe this should be a `SmallString` constructed via a `Twine`

mlir/lib/Dialect/MIOpen/Transforms/CloneKernels.cpp
```diff
@@ -28,50 +28,65 @@
...
+          symbolTable.insert(kernelFunc);
+
+          // TODO: also find all calls and import callee
+        }
       }
     }
   }
+
+private:
+  SmallVector<StringRef, 4> chips;
```
krzysz00: The storage of `StringRef` requires knowledge of the lifetimes of the underlying strings, which I don't think you have. This'll probably cause dangling pointers. `SmallVector<SmallString<8>>` or such instead?

krzysz00 (commented):
As an initial pass, I'm somewhat skeptical of some of the parts of the system detector

Furthermore, when we've been testing, we've had a persistent issue with doing too many `hipGetDeviceProperties()` calls in parallel (or at the very least, this wasn't helping our GPU hang issues). Can xmir-runner and friends have a mode that explicitly bypasses system detection and lets us pass in the specific chipset we want to run on?

external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt
```diff
@@ -280,4 +288,33 @@ if(MLIR_ENABLE_ROCM_RUNNER)
...
+    ${MLIR_MAIN_INCLUDE_DIR}/mlir/ExecutionEngine
+
+    DEPENDS
+    intrinsics_gen
+  )
+
+  llvm_update_compile_flags(MLIRRocmExecutionEngineUtils)
+  target_include_directories(MLIRRocmExecutionEngineUtils
+    PRIVATE
+    ${HIP_PATH}/include
```
jerryyin, line 315: Repercussions because of this:
 1. You can't assume HIP must exist in the current environment, therefore this has to be guarded by the same macro from mlir ROCm runner.
 2. Because of previous bullet point, you have to have a fall back for any component that rely on that this utils shared lib API in case the building environment doesn't have HIP.
 3. Thinking of this, I don't believe it is easy enough to come up with the fall back that populate those fields in a meaningful way. Therefore, I'm against of the idea to populate this field in a shared library. This should be a standalone executable that (1) If it functions, it gives the values for existing system (2) If now, user must enter in those field.

jerryyin (dismissed):
Sorry I have to put a blocker because of https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/723#discussion_r948040714. We should make sure to agree on that review comment before proceed with implementation details.

external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt
jerryyin: Reminded by @krzysz00's comments, I think a better idea instead is to make a minimal executable for obtaining information from HIP, and make sure that executable run only once. Then the rest of test can skip calling HIP API again and again, which tends to cause trouble.

external/llvm-project/mlir/include/mlir/ExecutionEngine/SystemDetect.h
sjw36: StringMap does not own the keys (StringRef) and there are cases where property keys are not constant.

krzysz00: StringMap does own the keys - it just takes `StringRef`s in the API to allow passing keys by reference

external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt
sjw36: This is guarded by the same macro (MLIR_ENABLE_ROCM_RUNNER), the same place mlir_rocm_runtime is built.

This lib (MLIRRocmExecutionEngineUtils) is only needed by xmir-runner which is also guarded.

sjw36:
Updated to make the SystemDetect conditional (use -target-chip on CLI to bypass).

external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt
jerryyin: Gotcha! That's right somehow I missed the point that the newly added block is already wrapped in the macro. Sorry!

external/llvm-project/mlir/include/mlir/ExecutionEngine/CpuSystemDetect.h
```diff
@@ -0,0 +1,39 @@
...
+#include "mlir/ExecutionEngine/SystemDevices.h"
+#include <vector>
+
+namespace mlir {
+
+/// CpuSystemDetect finds cpus on the current system.
+///
+
+class CpuSystemDetect : public std::vector<SystemDevice> {
+  CpuSystemDetect();
```
krzysz00, line 26: As a general note, is this being a singleton safe? It feels like something like this should depend on a context

external/llvm-project/mlir/include/mlir/ExecutionEngine/SystemDevices.h
```diff
@@ -0,0 +1,75 @@
...
+  template <class...>
+  struct types {
+    using type = types;
+  };
+
+public:
+  // Singleton per SystemDetect Types
+  template <typename... Targs>
+  static const SystemDevices &get() {
+    static SystemDevices s_systemDevices(types<Targs...>{});
```
krzysz00, line 44: Can you define a singleton in a header like this?

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
```diff
@@ -0,0 +1,77 @@
...
+            chip,
+            1,
+            {{"features", features},
+             {"major", to_str(deviceProps.major)},
+             {"minor", to_str(deviceProps.minor)},
+             {"multiProcessorCount", to_str(deviceProps.multiProcessorCount)},
+             {"sharedMemPerBlock", to_str(deviceProps.sharedMemPerBlock)},
+             {"regsPerBlock", to_str(deviceProps.regsPerBlock)},
+             {"warpSize", to_str(deviceProps.warpSize)}}};
+        push_back(std::move(dev));
```
krzysz00: Ok, yeah, I can feel the concurrency issues here

external/llvm-project/mlir/include/mlir/ExecutionEngine/SystemDevices.h
```diff
@@ -0,0 +1,75 @@
...
+
+  Type type;
+  llvm::SmallString<32> chip;
+  uint32_t count = 1;
+  llvm::StringMap<llvm::SmallString<8>> properties = {};
+};
+
+/// SystemDevices captures devices for the current system.
+///
+class SystemDevices : public llvm::StringMap<const SystemDevice &> {
```
krzysz00, line 39: If we're going to have a map of pointers, how do we ensure lifetimes? That is, can we get a better sense of who owns this data and how long it'll live for?

Secondly, is `SystemDevices` really a map? It feels better for it to have a map and to provide wrappers around the map API that fit our usecase.

external/llvm-project/mlir/lib/ExecutionEngine/CpuSystemDetect.cpp
```diff
@@ -0,0 +1,51 @@
...
+                   {{"llvm_triple", llvm::StringRef(targetTriple)}}};
+
+  llvm::StringMap<bool> hostFeatures;
+  if (llvm::sys::getHostCPUFeatures(hostFeatures)) {
+    for (auto &f : hostFeatures) {
+      llvm::StringRef val(f.second ? "1" : "0");
+      dev.properties.insert({f.first().str(), val});
+    }
+  }
+  push_back(std::move(dev));
```
krzysz00, line 50: Wait a minute, the map is storing references ... either we're staring at a weird quirk of the C++ or StringMap semantics or this is a dangling pointer waiting to happen

mlir/lib/Dialect/MIOpen/Transforms/CloneKernels.cpp
krzysz00: Yeah.  `mlir/include/mlir/Support/LLVM.h` So strong disapprove of `using namespace llvm;` - it causes a _lot_ of headaches.

krzysz00 (commented):
I've got some serious concerns about the design of SystemDetect - I don't think it'll be safe and therefore it probably won't be upstreamable.

Wrt the internal bits, are the changes you made to `mlir-miopen-driver` meaning that, in the future, we could have `mlir-rocm-runner` go away and be replaced by `mlir-cpu-runner` and having `mlir-miopen-driver` go all the way to compiled GPU modules + LLVM dialect host code?

external/llvm-project/mlir/include/mlir/ExecutionEngine/SystemDevices.h
krzysz00: Ok, yeah, this storing of references is reasonable, since they're pointing into a const singleton, but this storing of pointers feels like a way for things to break awkwardly - can we get some comments annotating lifetimes, etc.?

external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt
```diff
@@ -29,15 +32,20 @@ add_mlir_library(MLIRExecutionEngineUtils
   LINK_COMPONENTS
   Analysis
   Core
-  Coroutines
```
krzysz00, line 32: ?

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
```diff
@@ -0,0 +1,77 @@
...
+      } else {
+        auto to_str = [](auto v) {
+          return llvm::SmallString<8>(std::to_string(v));
+        };
+
+        SystemDevice dev{
+            SystemDevice::Type::EGPU,
+            chip,
+            1,
+            {{"features", features},
```
krzysz00: We might want a "AMD" in here somewhere

mlir/lib/Dialect/MIOpen/Transforms/CloneKernels.cpp
krzysz00: Reiterating the style nit

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
```diff
@@ -78,6 +78,9 @@ static cl::opt<std::string>
     tripleName("triple", cl::desc("target triple: amdgcn-amd-amdhsa"),
                cl::value_desc("triple string"), cl::init(""));
 
+static cl::opt<std::string> targets("targets", cl::desc("list of target chips"),
+                                    cl::init(""));
+
 static cl::opt<std::string> targetChip("target", cl::desc("target chip"),
```
krzysz00: Can we make "target" an alias for "targets" or is that premature?

krzysz00 (commented):
Ok, I'm no longer particularly concerned about the soundness of this design

Would you be opposed to [RFC]ing this upstream now to get discussion started?

I still have a few strong style concerns

And with how substantial this PR is I'd like a second reviewer

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
krzysz00: I'm going to call for a backend name here - we need to know this is a ROCm device and not a CUDA device

krzysz00 (commented):
One point reiterated (marking the detected GPU with its ROCm-ness), but, overall, I don't see anything fundamentally unsound with this design. I might not be the biggest fan of it, but it works, and I'll leave sending in the patch series upstream to you.

@jerryyin for a second pair of eyes on all this.

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
sjw36: Added a vendor=AMD field

external/llvm-project/mlir/lib/ExecutionEngine/CpuSystemDetect.cpp
jerryyin, line 50: It looks like you will end up with exactly 1 device into the vector. If that's the case, why bother using a vector as the type to inherit from?

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
```diff
@@ -0,0 +1,79 @@
...
+#define DEBUG_TYPE "execution-engine-rocm-system-detect"
+
+using namespace mlir;
+
+RocmSystemDetect::RocmSystemDetect() {
+  // collect all GPUs
+  int count = 0;
+  hipError_t herr = hipGetDeviceCount(&count);
+  if (herr != hipSuccess) {
+    llvm::errs() << "hipGetDeviceCount() should never fail\n";
```
jerryyin, line 35: Does the client want to handle the corner case that hip call failed yet we still want to compile for cpu? My assumption is that whenever someone use a `ExecutionEngine` with a ROCm header then he expects certain gpu to compile with, therefore, I'd suggest to use  `LLVM_UNREACHABLE` here.

This goes the same for `hipGetDeviceProperties()` call.

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
```diff
@@ -0,0 +1,79 @@
...
+  for (int i = 0; i < count; ++i) {
+    hipDeviceProp_t deviceProps;
+    herr = hipGetDeviceProperties(&deviceProps, i);
+    if (herr == hipSuccess) {
+      llvm::SmallString<32> chipName(deviceProps.gcnArchName);
+      llvm::StringRef chip = chipName;
+      llvm::StringRef vendor("AMD");
+      llvm::StringRef features;
+
+      auto idx = chipName.find_first_of(':');
```
jerryyin: `find_first_of` is not supposed to be used this way, I think it is a bit misleading that you are really only finding one character (versus its originally designed to find the first out of several). Please use `find()` instead.

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
```diff
@@ -0,0 +1,79 @@
...
+      if (idx != llvm::StringRef::npos) {
+        features = chipName.substr(idx + 1);
```
jerryyin: I doubt you can later use the features string as is from hip returned chip name, you are likely need to use a format conversion while reading it in. I think we have addressed this problem from `IsaNameSplitter.cpp`.

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
```diff
@@ -0,0 +1,79 @@
...
+        chip = chipName.substr(0, idx);
+      }
+      auto itr = std::find_if(begin(), end(), [&](const SystemDevice &dev) {
+        return dev.chip == chip;
+      });
+      if (itr != end()) {
+        itr->count++;
+      } else {
+        auto to_str = [](auto v) {
+          return llvm::SmallString<8>(std::to_string(v));
```
jerryyin: Indentation is broken from this line on. Looks like you need to rerun clang-format.

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
```diff
@@ -0,0 +1,79 @@
...
+        };
```
jerryyin: I don't quite get what the local to_str() lambda is for since that it can only be used in the local scope.

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
```diff
@@ -0,0 +1,79 @@
...
+      llvm::StringRef features;
+
+      auto idx = chipName.find_first_of(':');
+      if (idx != llvm::StringRef::npos) {
+        features = chipName.substr(idx + 1);
+        chip = chipName.substr(0, idx);
+      }
+      auto itr = std::find_if(begin(), end(), [&](const SystemDevice &dev) {
+        return dev.chip == chip;
+      });
```
jerryyin: I doubt we can consolidate different gpus together with only the chip field. To achieve the optimal performance, we have to compile with both chip and features.

mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
```diff
@@ -113,7 +116,7 @@ makeMIOpenConv2D(ConversionPatternRewriter &rw, Operation *op, Value input,
   if (xdlopsV2)
```
jerryyin, line 116: I happen to spot this here. 

I think you need to file a ticket to match the implementation of https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/733 for migraphx. In MIOpen scope, we use this gemm features enum to capture a field like xdlops.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
```diff
@@ -78,9 +78,10 @@ static cl::opt<std::string>
     tripleName("triple", cl::desc("target triple: amdgcn-amd-amdhsa"),
                cl::value_desc("triple string"), cl::init(""));
 
-static cl::opt<std::string> targetChip("target", cl::desc("target chip"),
-                                       cl::value_desc("AMDGPU ISA version"),
-                                       cl::init(""));
+static cl::opt<std::string>
+    targetChips("targets", cl::desc("list of target chips"), cl::init(""));
+
+static cl::alias aliasTargetChips("target", cl::aliasopt(targetChips));
```
jerryyin, line 84: We don't have backward compatibility concerns and in case it creates confusions later, I'd prefer:
 - Either open a ticket and make a single option between the two
 - Make the update in this PR now

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
```diff
@@ -200,53 +216,75 @@ static LogicalResult runMLIRPasses(ModuleOp &module,
...
+    LogicalResult kernelResult = success();
+    module->walk([&](ModuleOp kernelModule) {
+      auto chip = kernelModule->getAttrOfType<StringAttr>("kernel.chip");
+      if (chip && chips.contains(chip)) {
+        hasKernels = true;
+        kernelResult = runKernelPipeline(chip.getValue(), kernelModule,
+                                         isHighLevel, kernelPipelineSet);
+      }
+    });
+    if (!hasKernels) {
```
jerryyin, line 260: This is wired. I think both the set and get of `kernel.chip` attribute take advantage of the `targetChips` field so how can we end up in a scenario where they don't match? I think we'd only need to consider this in the rocm runner. Now, taking one step back, I haven't seen any update on rocm runner so how do we handle the situation in which we compiled from a different machine than we run from?

mlir/test/Conversion/MIOpenToGPU/multiple_kernels.mlir
```diff
@@ -12,22 +12,22 @@
 // CHECK-LABEL: gpu.func @step1
 
 module  {
-  func.func @step1(%arg0: memref<1x1216x16x3x3xf32>, %arg1: memref<1216x1x16x32x32xf32>, %arg2: memref<1216x1x1216x30x30xf32>) attributes {kernel} {
+  func.func @step1(%arg0: memref<1x1216x16x3x3xf32>, %arg1: memref<1216x1x16x32x32xf32>, %arg2: memref<1216x1x1216x30x30xf32>) attributes {kernel, arch = "gfx906"} {
```
jerryyin, line 15: [Question] This arch field in unit test doesn't do anything (since it is using miopen-opt, it does just blind compilation), is that correct?

external/llvm-project/mlir/lib/ExecutionEngine/CpuSystemDetect.cpp
sjw36: The SystemDevices will 'registerDevices' on this object and will iterate over all elements to add to it's map.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
sjw36: This is supported in xmir-runner. It is getting complicated for mlir-miopen-driver to support all these paths.. Let me take another look.

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
sjw36: It's is supposed to find the first match in the string and this is the version for a single char:

size_t find_first_of (char c, size_t pos = 0) const;

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
sjw36: Makes sense. I guess we should upstream this utility as well for use here and elsewhere.

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
sjw36: Yes, I was thinking we need to use a <chip>:<normalized_features> and have a special lookup for the case when a feature is not listed but matches when <feature>-. 

Question: can a feature with '-' be removed in normalization?

mlir/test/Conversion/MIOpenToGPU/multiple_kernels.mlir
sjw36: The 'arch' field (I would like to change to 'chip') is generally passed on the conv2d, not sure why it needs to be there. In the XMIR flow I create a sub-module that is 'chip' specific, so you could generate kernels for multiple targets (ie. gfx90a, gfx90a:sramecc+), along with features, num_cus, any override values from the chip.db (discussed in https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/607).

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
sjw36: Oh, it is set and get here because you can call mlir-miopen-driver multiple times for different parts of the pipeline.

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
sjw36: clang-format doesn't complain for me..

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
sjw36: I cannot find a good SmallString creation for converting an int to a string.. Ironically, the only llvm versions, also create std::string..

external/llvm-project/mlir/lib/ExecutionEngine/CpuSystemDetect.cpp
jerryyin: But looks like its `registerSystem()` call can handle either a single object or a vector of objects. 

Either ways, feel free to leave this unaddressed as there are systems that have more than 1 cpu although it is unclear to me if llvm system detection has support for that.

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
jerryyin: Yes, but some developer can come in later and instead make that a string argument should the splitter changed to be more than 1 character. Once they do that, they'd hit into this very unobvious bug caused by `find_first_of`.

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
jerryyin: Sounds reasonable.

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
jerryyin: We discussed offline and decided that `-` features are still necessary because its behavior is different between different code object versions.

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
jerryyin: It only work when you stage your change for the latest commit. I believe this update is from an earlier commit which is already in your branch.

So the way to handle this situation is to add an additional indention to this entire file then invoke clang-format. Then it is forced to do its job and fix the indentation.

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
jerryyin: Yes I got that intention... It's just that I don't quite get a lambda defined in a local else scope can be used outside below between line 66 to 71. Could you move the definition to outside of this local scope?

mlir/test/Conversion/MIOpenToGPU/multiple_kernels.mlir
jerryyin: Just want to make sure I understand, `arch` showing up on the func op is an unexpected result and that it shouldn't be here. It should rather be in sub-module or conv op.

If that's the case, could you file a ticket so we can keep track and remove it later?

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
jerryyin: Okay this is indeed very confusing... Is that a legit use case we want to support? Even for unit testing it shouldn't try to mislead the compiler by passing different chip attribute at different stage of the compilation. My assumption is that is can end up in very messy result if not carefully handled.

If we do want to support it, could you add a comment above just to describe this situation?

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
jerryyin: Nvm sorry, I think I misread and forget this is inside a lambda therefore the confusion.

mlir/test/Conversion/MIOpenToGPU/multiple_kernels.mlir
sjw36: It is valid to override on the 'func' level. When looking up 'chip' (or other features), it should look up hierarchically; first on the conv2d op, then on the enclosing func, then the enclosing module (but no further).

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
sjw36: I think it is still valid to continue when this happens. The client can fall-back to CPU jit and still run the code. They could also decide if no ROCm device available to bail.

jerryyin: Sure, it's up to you what the behavior of failing you want it to be. I'm from the CI angle where when HIP is dead in a CI machine, I'd typically let it to error out right way so machine problems doesn't propagate through.

mlir/test/Conversion/MIOpenToGPU/multiple_kernels.mlir
jerryyin: Gotcha

sjw36:
@jerryyin I think most things are addressed. Please review.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
sjw36: Added comments.

mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp
sjw36: Filed ticket https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/648

sjw36:
Actually, going to remove IsaNameSplitter and replace with RocmDeviceName.

sjw36:
Ok, cleaned up commit history again. @jerryyin please review.

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
krzysz00: Yeah, `-feature` means "assume feature is not a thing", `+feature` is "assume feature is a thing", and not mentioning feature means "it could go either way, make the conservative assumptions", last I checked

krzysz00: For example, `-sramecc` will allow generating some code that `+sramecc` or no mention of sramecc wil not.

mlir/test/Conversion/MIOpenToGPU/multiple_kernels.mlir
krzysz00: We don't look up. `arch` on the conv2d (feel free to rename) is the single source of truth for which chip that conv2d/gemm to lower to. We currently don't use that information explicitly inside any gemm implementations ... but we should. Namely, we should be allowing certain additional MFMA instructions when the arch isn't gfx908.

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
sjw36: removed.

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
```diff
@@ -0,0 +1,158 @@
...
+}
+
+/*
+Full list of properties:
+  name = '\000' <repeats 255 times>,
+  totalGlobalMem = 0x3ff000000,
+  sharedMemPerBlock = 0x10000,
+  regsPerBlock = 0x10000,
+  warpSize = 0x20,
+  maxThreadsPerBlock = 0x400,
```
jerryyin: If you plan to document the entire possible section of properties here, I don't think this is the best place for that documentation purpose. Could you open a ticket and add those details in?

external/llvm-project/mlir/lib/ExecutionEngine/RocmDeviceName.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include "mlir/ExecutionEngine/RocmDeviceName.h"
+#include "llvm/Support/Error.h"
+
+using namespace mlir;
+
+static constexpr const char kGcnDefaultTriple[] = "amdgcn-amd-amdhsa";
+
+// This function converts the arch name to target features:
+// sramecc+:xnack- to +sramecc,-xnack
+static LogicalResult parseTargetFeatures(ArrayRef<StringRef> tokens,
```
jerryyin, line 22: Thanks for simplifying the implementation here.

external/llvm-project/mlir/lib/ExecutionEngine/RocmDeviceName.cpp
```diff
@@ -0,0 +1,76 @@
...
+      return failure();
+    }
+    gcnArchFeatures += modifier;
+    gcnArchFeatures += token.substr(0, token.size() - 1);
+  }
+  return success();
+}
+
+RocmDeviceName::RocmDeviceName(StringRef devName)
+    : status(failure()), triple(kGcnDefaultTriple) {
```
jerryyin: Okay, I have to say I dislike the idiom here of trying to make constructor do all the work. From my background in oop, constructor is for basic setup, not full computation. Instead, build the fields on demand. 

In particular I dislike about the class carrying a side effect (the `status`) information in order to capture the failing case in construction. I doubt anyone that contributes to a client of this splitter are expected to handle the possibility of construction failures.

I tend to ask you to move this into a function like `init()` that populate those fields.

mlir/test/Conversion/MIOpenToGPU/multiple_kernels.mlir
sjw36: There are more ops in the scope of a kernel that may need chip specific details to help lowering. The whole kernel-func will be compiled for a specific chip. So no, this should be inherited from scope.

jerryyin (approved):
LGTM, thanks for all the work.

external/llvm-project/mlir/lib/ExecutionEngine/RocmSystemDetect.cpp
sjw36: Will remove from here. Captured in https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/607

external/llvm-project/mlir/lib/ExecutionEngine/RocmDeviceName.cpp
sjw36: This is just a Helper utility object but yes it can fail without the user knowing. I guess I will revert it back to a function that updates string parameters and returns a LogicalResult.

external/llvm-project/mlir/lib/ExecutionEngine/SystemDevices.cpp
```diff
@@ -0,0 +1,41 @@
+//===- RocmSystemDetect.cpp - Detect ROCm devices -------------------------===//
```
zhanglx13, line 1: @sjw36  I guess you forgot to change this comment after copying it from `RocmSystemDetect.cpp`.

PR #724, opened 2022-08-11
Refactor accumulator data type between xdlops/nonxdlops

jerryyin:
Leftover from https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/711. Accumulator selection should be consolidated between xdlops/nonxdlops.

### DIFF
--- a/mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
+++ b/mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
@@ -56,7 +56,7 @@ struct MIOpenGridwiseGemmToBlockwisePass
 
 // TODO(kdrewnia): Could rank-0 vectors clear some of this up?
 // Utility function for crafting optional vector types
-Type vectorTypeOrSelf(Type elementType, int64_t len) {
+static Type vectorTypeOrSelf(Type elementType, int64_t len) {
   if (len == 1)
     return elementType;
   return VectorType::get({len}, elementType);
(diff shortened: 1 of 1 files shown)


mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -84,6 +84,21 @@ void computeLoadStoreTypeInfo(OpBuilder &b, ArrayRef<int64_t> sliceLengths,
   loadType = vectorTypeOrSelf(elementType, loadLength);
 }
 
+Type obtainAccumulatorType(OpBuilder &b, Type &elementType, Type &destType) {
```
krzysz00: In minor nitpicks, we could put a `static` on this.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -84,6 +84,21 @@ void computeLoadStoreTypeInfo(OpBuilder &b, ArrayRef<int64_t> sliceLengths,
...
+  // Determine the type used on VGPR to act as accumulator.
+  // f32: f32.
+  // f16, bf16: f32 to prevent overflow from happening.
+  // i16 : i16.
+  // i8: i32, since we have an i32 output
+  Type accumulatorType = destType;
+  if (elementType == b.getF16Type() || elementType == b.getBF16Type()) {
```
krzysz00: Minor comments I realized, we have `elementType.isF15()`  and such. We could also, in principle, not pass in the builder and use stuff like the `get()` methods on the types we want to construct, passing in the context from the elementType, but idk about if that's the right call here.

krzysz00 (approved):
Minor style complaints, but approved

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Good idea. I debated over this as I see none of other local function carry static. While I'm here, better fix those as well.

Fixed in ac59dec

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: I like the idea of check type from themselves instead of creating a useless temporary and have them immediately destroyed, good point!

On the second one, all type construction require a context, which is probably why those have been abstracted away in `Builders.h` as a convenient wrapper. I'll skip this one and continue to pass in the builder.

PR #726, opened 2022-08-12
Restore tosa partition tests

pcf000:
I noticed the other day, while porting the tosa-partition code into torch-mlir, that the tosa-partition.mlir and tosa-partition-run.mlir tests had disappeared.  Turns out that they used --linalg-comprehensive-module-bufferize, which was recently removed.  There's a --one-shot-bufferize, but it isn't a direct replacement, and following the most obvious path foundered when a tosa.conv2d wouldn't bufferise.

This patch updates the tests to use the appropriate options, and I regenerated tosa-partition-run.mlir using torch and a simple torch.nn.Conv2d model to avoid the failure-to-bufferize error.

### DIFF
--- a/external/llvm-project/mlir/test/Dialect/Tosa/tosa-partition-run.mlir
+++ b/external/llvm-project/mlir/test/Dialect/Tosa/tosa-partition-run.mlir
@@ -1,0 +1,52 @@
+// RUN: mlir-opt %s -pass-pipeline="func.func(tosa-to-linalg-named)" \
+// RUN: -pass-pipeline="func.func(tosa-to-linalg)" --tosa-to-tensor \
+// RUN: -tosa-to-arith -arith-bufferize \
+// RUN: -linalg-comprehensive-module-bufferize="allow-return-allocs" \
+// RUN: -linalg-bufferize -vector-bufferize -tensor-bufferize \
+// RUN: -func-bufferize -finalizing-bufferize --convert-linalg-to-loops \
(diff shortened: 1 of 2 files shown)


pcf000:
Oops, forgot to rebase-and-squash before making the request.  Oh, well, it'll happen with the merge.

krzysz00 (commented):
Shouldn't these pipelines match https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/miopen-dialect/mlir/lib/Dialect/MIOpen/Pipelines/Pipelines.cpp#L89 (and nearby)?

pcf000:
> Shouldn't these pipelines match https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/miopen-dialect/mlir/lib/Dialect/MIOpen/Pipelines/Pipelines.cpp#L89 (and nearby)?

Separate origins, independent evolution, etc, etc.  But thanks for leading me to that file and to mlir::tosa::addTosaToLinalgPasses(), which may be handy in the future.  For my simple case, I don't need all the passes in that pipeline, only enough to make it run.

krzysz00:
I'd argue we should strive for consistency here, but fair enough.

If you want these in as is, I'll approve, but it might make things easier to debug if the tosa-to-linalg pipelines match.

krzysz00 (approved):
Approved - there's no fundamental reason not to test with the minimal pass set, and given the nitpick status of adding more passes, I defer to @pcf000's preferences here.

PR #727, opened 2022-08-15
Decouple between kPack/non-kPack in xdlopsgemm

jerryyin:
This is a partial work that makes xdlops and its IR cleaner. This is also as far as I can go for refactoring xdlops gemm for now. I will leave the next section of refactoring to when I will finish:
 - https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/623
 - https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/624
Then, existing xdlopsgemm will become a coordinate transfer on `[K, KRepeats, k_base]`

I will shift focus to adjusting the responsibility between blockwisegemmv2 and xdlopsgemm further in my next PR to unblock larger M/N repeats.

### DIFF
--- a/mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
+++ b/mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
@@ -249,7 +249,7 @@ struct XdlopsGemmV2RewritePattern : public OpConversionPattern<XdlopsGemmV2Op> {
       //   argB = a[k_i];
       //   p_c_thread = mfma(argA, argB, p_c_thread);
       // }
-      Value zeroConstantOp = b.create<ConstantIndexOp>(loc, 0);
+      Value zeroConstantOp = b.createOrFold<ConstantIndexOp>(loc, 0);
       auto mfmaLoop = b.create<TransformingForOp>(
           loc, ArrayRef<ValueRange>{{zeroConstantOp}},
           ArrayRef<Attribute>{b.getArrayAttr({})},
@@ -261,26 +261,25 @@ struct XdlopsGemmV2RewritePattern : public OpConversionPattern<XdlopsGemmV2Op> {
         b.setInsertionPointToStart(mfmaLoop.getBody());
         Value coord = mfmaLoop.getLowerCoords(/*domain=*/0)[0];
 
-        Value offset = coord;
-        Value regAIdx = b.create<AddIOp>(loc, regAConstantOp, offset);
-        Value regBIdx = b.create<AddIOp>(loc, regBConstantOp, offset);
+        auto loadArg = [&](Value regOffset, Value matrix) {
+          Value regIdx = b.create<AddIOp>(loc, regOffset, coord);
+          Value arg;
+          if (k_base == 1) {
+            // xdlops needs only 1 element, load directly from buffer.
+            arg = b.create<InBoundsLoadOp>(loc, argType, matrix,
+                                           ValueRange{regIdx});
+          } else {
+            // k_base > 1, use transferRead to load a vector length equivalent
+            // with a xdlops argument.
+            arg = b.create<vector::TransferReadOp>(
+                loc, argType.cast<VectorType>(), matrix, ValueRange{regIdx},
+                /*InBounds*/ ArrayRef<bool>(true));
+          }
+          return arg;
+        };
 
-        Value argA;
-        Value argB;
-        if (k_base == 1) {
-          // xdlops needs only 1 element, load directly from buffer.
-          argA = b.create<memref::LoadOp>(loc, argType, matrixA,
-                                          ValueRange{regAIdx});
-          argB = b.create<memref::LoadOp>(loc, argType, matrixB,
-                                          ValueRange{regBIdx});
-        } else {
-          // k_base > 1, use transferRead to load a vector length equivalent
-          // with a xdlops argument.
-          argA = b.create<vector::TransferReadOp>(
-              loc, argType.cast<VectorType>(), matrixA, ValueRange{regAIdx});
-          argB = b.create<vector::TransferReadOp>(
-              loc, argType.cast<VectorType>(), matrixB, ValueRange{regBIdx});
-        }
+        Value argA = loadArg(regAConstantOp, matrixA);
+        Value argB = loadArg(regBConstantOp, matrixB);
         populateMfma(b, argA, argB);
       }
     } else {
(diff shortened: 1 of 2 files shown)


mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -212,119 +212,123 @@ struct XdlopsGemmV2RewritePattern : public OpConversionPattern<XdlopsGemmV2Op> {
...
+        b.setInsertionPointToStart(mfmaLoop.getBody());
+        Value coord = mfmaLoop.getLowerCoords(/*domain=*/0)[0];
+
+        Value offset = coord;
+        Value regAIdx = b.create<AddIOp>(loc, regAConstantOp, offset);
+        Value regBIdx = b.create<AddIOp>(loc, regBConstantOp, offset);
+
+        Value argA;
+        Value argB;
+        if (k_base == 1) {
```
krzysz00: This might be spelled `miopen.in_bounds_load`?

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -212,119 +212,123 @@ struct XdlopsGemmV2RewritePattern : public OpConversionPattern<XdlopsGemmV2Op> {
...
+          // xdlops needs only 1 element, load directly from buffer.
+          argA = b.create<memref::LoadOp>(loc, argType, matrixA,
+                                          ValueRange{regAIdx});
+          argB = b.create<memref::LoadOp>(loc, argType, matrixB,
+                                          ValueRange{regBIdx});
+        } else {
+          // k_base > 1, use transferRead to load a vector length equivalent
+          // with a xdlops argument.
+          argA = b.create<vector::TransferReadOp>(
```
krzysz00: And if the above comment is false, can we mark the transfer_read with an explicit "inbounds" attribute for clarity?

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -212,119 +212,123 @@ struct XdlopsGemmV2RewritePattern : public OpConversionPattern<XdlopsGemmV2Op> {
...
-    Value matrixBElement;
+    // TODO: zyin adopt generic layout: K, KRepeats, k_base
+    // After that, we'd be able to uniform between kPack/nonKPack
+    if (KPack == 1) {
+      // for(index_t k_i = 0; k_i < KPerThread; k_i += k_base) {
+      //   argA = a[k_i];
+      //   argB = a[k_i];
+      //   p_c_thread = mfma(argA, argB, p_c_thread);
+      // }
+      Value zeroConstantOp = b.create<ConstantIndexOp>(loc, 0);
```
krzysz00: Maybe a `createOrFold` to save the pattern rewriter some work?

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -212,119 +212,123 @@ struct XdlopsGemmV2RewritePattern : public OpConversionPattern<XdlopsGemmV2Op> {
...
+          /*bounds=*/ArrayRef<int64_t>{KPerThread},
+          /*strides=*/ArrayRef<int64_t>{k_base},
+          /*useIndexDiffs=*/true, /*forceUnroll=*/true);
+      {
+        OpBuilder::InsertionGuard guard(b);
+        b.setInsertionPointToStart(mfmaLoop.getBody());
+        Value coord = mfmaLoop.getLowerCoords(/*domain=*/0)[0];
+
+        Value offset = coord;
+        Value regAIdx = b.create<AddIOp>(loc, regAConstantOp, offset);
```
krzysz00: Here's a thought. What if the loop had transforms where we `Slice` to apply the regAOffset and regBOffset, if any? Not sure that'd make it any clearer, but it would lead to interesting things.

We could also have one loop for a and one loop for b, each of which  starts at the relevant register offset.

... in fact, yeah, what if there was a function that gave you this load loop for A or B, as relevant - the only difference between those loops is whether you use `regAOffset` or `regBOffset`. Sure, that'll leave the loads in a different order, but it does mean not needing to copy-paste the A and B logic.

krzysz00 (commented):
Overall, nice PR. I think there might be some ways to reduce the amount of duplication for handling `a` and `b`, but that might not be worth it.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
jerryyin: Yes, that would work. I guess both'd work interchangeably but we'd prefer a higher abstraction op?

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
jerryyin: Tuning parameter selection will guarantee this satisfy, but agreed that an explicit inbounds will be a more defensive way to make sure the behavior is as expected when going out of bounds.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
jerryyin: Sounds good.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
jerryyin: Actually I'd prefer instead to make a single loop and call a lambda for a and b each. 

Reasoning being that we know that the right form of code should consolidate both loops into one for max performance. I find it defeat the purpose if we instead make two loops for the hope that compiler would do the right thing for us.

krzysz00: Entirely fair. We could have a loop that gives you an a coordinate and a b coordinate.

Although the pattern further up in GEMM is to have a loop for a and a loop for b and it seems to work fine

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
krzysz00: It's actually for the opposite of the reason you said ... we know the load will be in bounds, and so we want to make extra-sure the lowering pipeline won't generate if statements.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
krzysz00: I have `in_bounds_load` to handle that abstracted "maybe we're loading a vector, maybe we're not" case.

jerryyin: Done in 1405899

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
jerryyin: Thanks for the explanation! Done in 1405899

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
jerryyin: Done in 1405899

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
jerryyin: Done in 1405899

krzysz00 (approved):
Nice cleanup! LGTM

PR #728, opened 2022-08-15
Remove dead PaddingInfoAttr _and_ use structs for tuning parameters

krzysz00:
Since we now use bounds inference to determine when bounds checks are
needed, and since the bugs that prevented us from using index diffs in
padding kernels are long gone, there are no uses of the
PaddingInfoAttr in any of the gemm implementations ... but we still
pass it in.

Remove the attribute, remove the code setting it, and rework any
relevant tests to account for the loss of the information about
padding being applied.

This is also setup for a miopen.gemm operation.


jerryyin (commented):
@whchung Do you remember why `PaddingInfo` is there? What part of padding kernel require this information, or what do we want to address by adding this attribute in the first place?

I find it strange that this never get consumed below gridwisegemm level but figure that this must exist for a reason from start.

krzysz00:
@jerryyin This info used to be used in some backwards data kernels to determine if we could use index diffs. We stopped needing to make that decision back in https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/582 , and the padding info's been dead since.

jerryyin (approved):
Thanks for the explanation @krzysz00

PR #729, opened 2022-08-16
[DO NOT SQUASH] Fix fusion test failure by modifying TosaToLinalg.

jungpark-mlir:
TosaToLinalg folds its dim first when a dim size of the operands is 1 and add reshape ops later to match with the original shape.
This might complicate the later passes.

For example,
when tosa.add has two compatible tensor operands A[1,30,30,16] and B[1,1,1,16]
TosaToLinalg reshapes B to [16]
and LinalgElementwiseOpFusion pass reshapes A and B to A[1,30,30,1,1,1,16] and B[1,1,1,16] and result also needs to be reshaped from [1,30,30,1,1,1,16] to [1,30,30,16]

This PR prevents TosaToLinalg pass from dropping the dims with size=1 so that the operands of elementwise ops keep their shapes through the converting passes.
Not sure everyone upstream would be happy with this change but this looks more sensible to me (If something need this existing dimension folding, that should be optional)

This also fixes the failed fusion tests.

### DIFF
--- a/external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalg.cpp
+++ b/external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalg.cpp
@@ -631,7 +631,7 @@ elementwiseMatchAndRewriteHelper(Operation *operation,
     SmallVector<AffineExpr, 4> affineExprs;
     newShape.reserve(type.getRank());
     for (const auto &it : llvm::enumerate(type.getShape())) {
-      if (it.value() == resultTy.getDimSize(it.index()) && it.value() != 1) {
+      if (it.value() == resultTy.getDimSize(it.index())) {
         newShape.push_back(it.value());
         affineExprs.push_back(
             mlir::getAffineDimExpr(it.index(), rewriter.getContext()));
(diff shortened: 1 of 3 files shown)


jungpark-mlir:
Need to modify the external test as well.

jungpark-mlir:
Well, 
I need to cover the option to generate the code in the original way so it doesn't break the compatibility in upstream.

krzysz00 (changes requested):
I don't think this is the right approach either, and I don't want us maintaining this patch on upstream.

Why not have a pass (or a rewrite pattern in an existing pass) somewhere after the linalg elementwise stuff that either replaces the `expand_shape`/`collapse_shape` calls with coordinate transformations? Or something that works the way the rest of our existing broadcast handling does?

Could you also identify upstream commit that causes this problem? (and maybe, in the process, find some sort of flag for restoring the old behavior if we get really lucky)

I think we should meet Thursday or Friday to discuss this at this point.

jungpark-mlir:
Yeah, I should have checked the history first.

It's not the upstream that introduced folding the dims. The change is only on our repo and was made by @sjw36 in https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/commit/45ecde6aedeb2085abee1785abbd4189be7c7a71

Need to discuss this once he's back.

sjw36:
The reason this still fails is that MIOpenLinalgAlign is failing.

jungpark-mlir:
Yes, I'm reworking this differently now. This needs to be discussed with upstream later.
Closing.

PR #730, opened 2022-08-17
Update cmake to 3.22.1.  Update buildbot to match, and for production.

pcf000:
Update cmake for general use, since the old 3.15 would break on the next update-llvm-from-main-line exercise.  Buildbot was already loading a newer one, but it's based on the rocm/mlir image so now it won't need to.  And update buildbot port to production and add a pointer to confluence documentation.


jerryyin (approved):
Thanks! The dockerfile changes will not be tested until the new build is on dockerhub. Could you take a look for that and successive build to make sure it is working?

PR #731, opened 2022-08-18
Store tuning parameters in structs

krzysz00:
1. Define attributes to contain the expected tuning parameters for xdlops/non-xdlops ("general", because we might end up with a Navi one) GEMM
2. Actually specify that gridwise gemm etc. have their tuning parameters as an attribute, so that we can't construct a gridwise gemm without tuning by accident
3. Make convolutions have an optional set of tuning parameters attached
4. Similarly, the gridwise GEMM and below take the blockSize and gridSize as parameters, and those are now optional parameters on the convolutions ops
5. The target architecture and number of CUs are now mandatory operands to a `miopen.conv*` and the gridwise GEMM below
6. Since we now explicitly say block and grid sizes are i32, make our code talk about them as `uint32_t`s to avoid accidental widening

The next step here will be to define a `GemmFeatures` enum to no longer need the ad-hoc `xdlopsV2` attribute.

Note that the `matrix_*_data*` attributes haven't been ported into this scheme because they're all slated for deletion anyway.


jungpark-mlir:
Is it an optional attribute on the miopen.conv because tuning parameters firstly come as perf_config string and affixTuningParam creates the new tuning params structure attribute?
So, formally registering the tuning parameter attributes now.

Looks good to me,
I'm currently planning to add perf_config directly on the miopen.conv2d (i.e., just before visiting the AffixTuningParams pass)
in the MIGraphX integrated pipeline, this should be fine with it.

krzysz00:
Yeah, it's an optional parameter because we have `--affix-tuning-pararms`.

In the future, I'm thinking `perf_config` will get turned directly into the struct once we've gotten rid of all the tuning parameter derivation stuff.

jungpark-mlir (approved):

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
```diff
@@ -40,7 +40,12 @@ def MIOpen_Conv2DOp :
     MIOpen_Op<"conv2d">,
     Arguments<(ins MemRefRankOf<[F32, F16, BF16, I8], [5]>:$filter,
                    MemRefRankOf<[F32, F16, BF16, I8], [5]>:$input,
-                   MemRefRankOf<[F32, F16, BF16, I32], [5]>:$output)> {
+                   MemRefRankOf<[F32, F16, BF16, I32], [5]>:$output,
+                   StrAttr:$arch,
+                   I32Attr:$numCu,
+                   OptionalAttr<I32Attr>:$blockSize,
+                   OptionalAttr<I32Attr>:$gridSize,
+                   MIOpen_OptionalGemmParams:$params)> {
```
jerryyin, line 48: I know you've had a conversation about @jungpark-mlir about this but I still feel this is very awkward to be as a convolution attribute. What's your plan on using them? Are you plan on them to be there for debugging purpose? It also looks like they are ignored in conv->gemm lowering.

 If they are for debugging purpose, then it is a duplication of `blocksizeoverrid` or `gridsizeoverride`. If it isn't for debugging I can't figure out how `blockSize` or `gridSize` should be derived without yet lowering to gemm yet?

jerryyin (commented):
Overall nice cleanup, thanks!

mlir/include/mlir/Dialect/MIOpen/MIOpenOps.td
krzysz00: This isn't for debugging. The block size, grid size, and tuning parameters *are* used in the gridwise GEMM lowering, and so must be present for gridwise gemm. They're derived by `--affix-tuning-params`, hence them being optional attributes.

We've always been implicitly setting these attributes: this change just makes their existence explicit.

jerryyin: I think I've misunderstood current vs future. Yes, right now affix tuning parameters derive all that before a conv op therefore this is all reasonable stuff.

jerryyin (approved):

PR #732, opened 2022-08-19
Embedded INCLUDE_DIR info as property of the exported library

zhanglx13:
So that:
- No need to set variable `${MLIRMIOpen_INCLUDE_DIR}` explicitly to hold
the INCLUDE_DIR
- No need to call `target_include_directories()` in MIOpen

Some cleanup
- Stop installing Miir.h when BUILD_FAT_LIBMLIRMIOPEN is not set
- Install libMLIRMIOpen.a only once when BUILD_FAT_LIBMLIRMIOPEN is set

### DIFF
--- a/mlir/tools/mlir-miopen-lib/CMakeLists.txt
+++ b/mlir/tools/mlir-miopen-lib/CMakeLists.txt
@@ -47,9 +47,6 @@ llvm_update_compile_flags(mlir-miopen-lib-test)
 target_link_libraries(mlir-miopen-lib-test PRIVATE MLIRMIOpenThin ${LIBS})
 mlir_check_link_libraries(mlir-miopen-lib-test)
 
-install(FILES "Miir.h"
-  DESTINATION include)
-
 llvm_canonicalize_cmake_booleans(BUILD_FAT_LIBMLIRMIOPEN)
 # Static library target, enabled only when building static libs
 if( BUILD_FAT_LIBMLIRMIOPEN )
@@ -86,9 +83,6 @@ if( BUILD_FAT_LIBMLIRMIOPEN )
     DEPENDS
     MLIRMIOpen
     )
-
-  install(TARGETS MLIRMIOpen
-    ARCHIVE DESTINATION lib)
 
   # Install Miir.h to ${CMAKE_INSTALL_PREFIX}/include/<package>/
   # as part of component libMLIRMIOpen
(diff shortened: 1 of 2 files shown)


krzysz00 (approved):
Nice bit of cleanup, approved

mlir/tools/mlir-miopen-lib/CMakeLists.txt
```diff
@@ -130,10 +124,15 @@ if( BUILD_FAT_LIBMLIRMIOPEN )
...
+    # The INTERFACE_INCLUDE_DIRECTORIES will be set as the property of the exported library
+    # so that when comsumer libraries (MIOpen) link in ibMLIRMIOpen, ${INCLUDE_DIR} is
+    # automatically propagated to the consumer library's include_dirs, which has the same
+    # effects as calling target_include_directories(MIOpen ${MLIRMIOpen_INCLUDE_DIR})
     install(TARGETS ${target_name}
       EXPORT ${export_set}
       ARCHIVE DESTINATION lib/${PACKAGE_NAME}
-      COMPONENT ${component_name})
+      COMPONENT ${component_name}
+      INCLUDES DESTINATION ${INCLUDE_DIR})
```
jerryyin, line 135: Just for clarification, is it the `INCLUDES DESTINATION` that will set `INTERFACE_INCLUDE_DIRECTORIES`? Also, which exported cmake file will help MIOpen to automatically find the `${INCLUDE_DIR}`?

zhanglx13: Yes.
The `find_package()` in MIOpen first locates `MiirConfig.cmake`, which includes `MiirTargets.cmake`, which further set the `INTERFACE_INCLUDE_DIRECTORIES` property of the `Miir::MLIRMIOpen` targets as follows
```cmake
add_library(Miir::MLIRMIOpen STATIC IMPORTED)
set_target_properties(Miir::MLIRMIOpen PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include/Miir")
```
Later in MIOpen, when `target_link_libraries(MIOpen Miir::MLIRMIOpen)` is called, MIOpen obtains MLIRMIOpen's `INTERFACE_INCLUDE_DIRECTORIES` automatically.

jerryyin: Make sense and thanks for the detailed explanation!

PR #733, opened 2022-08-19
Define a GemmFeatures enum, use it

krzysz00:
This sets up for the architecture DB stuff that we're planning and allows for having more features like a Navi kernel.


jerryyin (commented):
I think this is a reasonable step forward but if I'm to implement this I'd probably pick a different way. Do you see this PR as an execution of https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/607 (which I'm WIP on a proposal)? Was planning to have this arch and features mapping added to module op (instead of conv op, because other ops inside can potentially also use this information).

Also we may need somewhat more explicit planning sessions in case our work collide into each other.

krzysz00:
@jerryyin I'm seeing this as a step towards executing that issue, but mainly as getting rid of `xdlopsV2 = true` as an attribute. 

I'm operating on the theory that the attributes about what we're targeting should be on each kernel operation and not on the module: if nothing else, you might have kernel functions for the same architecture in the same module. We could also go for putting features (and stuff like `blockSize`) only on the containing function, but that imposes a requirement that our kernels live inside particular functions that I'm not sure is a good move.

jerryyin:
> I'm seeing this as a step towards executing that issue

Sounds good. Agreed this is the first step in execution of that ticket. Will likely needs a few adjustment in the future and more fusion side stuff to kick in and clarify the requirements.

> We could also go for putting features (and stuff like blockSize) only on the containing function

I think block size is already put on the funcop?

> but that imposes a requirement that our kernels live inside particular functions that I'm not sure is a good move

In that ticket only architecture specific attributes will be on moduleop. This does not include block/grid size, but xdlops, since xdlops attribute can be shared among any AST child under the moduleop.

krzysz00:
block size is both on the function and on the kernel - I'm willing to let either the kernel or the function be the source of truth ... but if it's the function then we'll be erroring out during lowering if the function containing the gridwise_gemm[_v2] doesn't have a block size set.

And re attributes ...
1. Can't we have functions targetting multiple architectures in the same module? We already have modules that have both CPU and GPU code in them.
2. If we want to go arch-specific, we should have a `miopen.module`
3. This again imposes requirements that kernels be nested inside specific outer blocks. While I'm not opposed to this, I'd like to be clear that that's what you're asking for?

krzysz00:
... here's why we don't want this to just be a module-level thing: we can compile for `gfx908` but have xdlops on for one kernel and off for another. IN fact, we commonly do that for verification.

jerryyin (approved):
I have discussed this with @krzysz00 offline and our decision is documented [here](https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/607#issuecomment-1224863139). @sjw36 Please take a look at this review as well as the updated proposal.

PR #735, opened 2022-08-22
[DO NOT SQUASH] Move fusion fixup rewrites to their own pass + #742

krzysz00:
These patterns shouldn't be in conv-to-gemm, both because they'll be
applicable to a GEMM by itself and because they're only used when
fusion is enabled.


krzysz00:
@sjw36 @jungpark-mlir Can I get this poked along so the GEMM operator can land?

jungpark-mlir:
> @sjw36 @jungpark-mlir Can I get this poked along so the GEMM operator can land?

I'm fine with that

krzysz00:
@jungpark-mlir To rephrase, the only actual change in this PR is the fusion-fixup pass. That still needs review (although the only objections I can see is naming or a broad opposition to moving those rewrites to their own pass). The GEMM operator got merged in here because it depended on this PR and got approved first.

jungpark-mlir (commented):
Actually let's try to find another name for the pass.
miopen-fold-to-transform
miopen-fold-transpose

krzysz00:
The question is, and this is an @sjw36 question since he's working on some fusion stuff, do we see this pass either gaining more patterns or being removed? If so, we might want to keep the current name because it'll either become relevant or because it'll be deleted

sjw36 (approved):
Approved once you change the name (as per @jungpark-mlir 's comment).

krzysz00:
Thanks for the suggested names, `-miopen-fold-transpose` will be landing soon

PR #736, opened 2022-08-22
[MLIR#620] Use RMS and element wise metrics together in the verification function

zhanglx13:
This PR serves as the first step ([MLIR#620](https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/620)) towards the better E2E testing framework.

- The new verification function computes RMS, maxAbsDiff, and maxRelDiff when comparing the results from gpu kernel and the validation function. It checks whether these metrics are below the given tolerance. The test is considered to be pass if all metrics are below their tolerance.
- Thresholds for the metrics are hard coded as follows
  - RMS: 0.00003f for all data types
  - absDiff: 100.0f for all data types, i.e. the absDiff metric is disabled
  - relDiff: 
    - 100.0f for f16, i.e. the relDiff metric is disabled for f16 tests
    - 0.000001f for other data types
- The verification function can print the following debug information if it is told to do so
  - Values of the three metrics: RMS, maxAbsDiff, and maxRelDiff
  - Histogram of the relDiff metric
  - Errors of elements if their difference is larger than the tolerance in the form
    `index: expectedVal  computedVal  absDiff  relDiff`
- `miopen-gen` emits code to call the new verification function based on the data types of the gpu kernel and the validation function.
- An option `-print-verify-results|-p_verify` is added to `miopen-gen` to control when to print verification debug information
  - `-p_verify=always`: print debug info
  - `-p_verify=failure`: print debug info only if the test fails
  - `-p_verify=off`: do not print debug info

The next PR will remove the hard coded thresholds and make them test-specific parameters.


krzysz00 (commented):
In high-level questions, is there a good way to avoid the memcpy() calls by instead having the type conversion in the verifier?

In low-level questions: why do you have tho comments showing very detailed output of the generated code? Do we need those?

Just to check, @yiqian1, is this broadly independent of your work on changing the CPU verifier?

krzysz00 (changes requested):
Ok, actual blocker: you've edited the E2E tests, please edit the E2E test generation script in `mlir/utils` as well. If some of the E2E tests needed some metrics disabled, please extend that script to allow for that sort of disabling.

zhanglx13:
> In high-level questions, is there a good way to avoid the memcpy() calls by instead having the type conversion in the verifier?

On the CPU side, there are helper c functions to convert f32 to f16/bf16 and convert bf16 to f32. No helpers to convert f16 to f32. So if we want the conversion to happen in the verification function, we have to do either of the following
1. Implement a helper function in C to convert f16 to f32 so that we can still use the C implementation of the verification function. A good reference is [this doc](http://www.fox-toolkit.org/ftp/fasthalffloatconversion.pdf).
2. Emit mlir code to implement the verification function (as the case of the old verification function).

So what's you concern with those `memcpy()`'s for conversion?
> In low-level questions: why do you have tho comments showing very detailed output of the generated code? Do we need those?

Because I found this kind of comments in the old `createVerifierFunc()`. I figured it is more convenient for reviewers to see what is generated. But let me know if they are not necessary and there is a better way to review the generated code.
> Just to check, @yiqian1, is this broadly independent of your work on changing the CPU verifier?

I believe @yiqian1's work happens before generating the verification function in question here. But please let me know if there is any overlap.

> Ok, actual blocker: you've edited the E2E tests, please edit the E2E test generation script in mlir/utils as well. If some of the E2E tests needed some metrics disabled, please extend that script to allow for that sort of disabling.

Can you clarify that if the E2E generation script is used in any of the CI checks? I do not find it being called in the Jenkins file so I think it should be fine to manually change the tests **just for this PR**.

The general modification of the E2E generation script will happen in the next PR, which is planned to address the issues mentioned in [#621](https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/621) and [#619](https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/619).

krzysz00:
Ok, so
1. I don't have any fundamental issue with the memcpy()s, just a gut feeling that we might not need to allocate an entire temporary buffer. However, linking in `half` would be A Pain, so I figure we might as well go with the memcopy approach.
2. Yeah, Yi's work is probably unrelated, I just wanted to check with her.
3. The script isn't run anywhere in the CI - currently it's just used to manually regenerate the tests. My concern was that I didn't want the test generator to become useless on account of this PR, but if you're planning to restore that functionality (or make it unnecessary) soon, I'm good with letting this merge.

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -1195,248 +1204,136 @@ createVerifierFunc(ModuleOp &module, const KernelIF &kernel,
...
 
-  // call mcpuPrintF32(float f1, float f2)
-  func::FuncOp printFunc;
-  if (elemType.isIntOrIndex()) {
-    printFunc = makeFuncDecl(module, "mcpuPrintInt32", {intType, intType});
+  // obtain function name of the verifier wrapper
+  std::string verifyFuncName = "mcpuVerify5D";
+  if (valElemType.isF32())
+    verifyFuncName += "Float";
+  else
```
krzysz00: In the interests of paranoia, can we get this to be `else if .isIntegerType(32) ... else (emit error)`

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -1195,248 +1204,136 @@ createVerifierFunc(ModuleOp &module, const KernelIF &kernel,
...
+    // call @_memcpy_f32_f16(%7, %9, %c9408_2) : (memref<?xf32>, memref<?xf16>, index) -> ()
+    // %10 = memref.collapse_shape %5 [[0, 1, 2, 3, 4]] : memref<1x64x3x7x7xf16> into memref<9408xf16>
+    // %11 = memref.cast %10 : memref<9408xf16> to memref<?xf16>
+    // %12 = memref.collapse_shape %arg1 [[0, 1, 2, 3, 4]] : memref<1x64x3x7x7xf32> into memref<9408xf32>
+    // %13 = memref.cast %12 : memref<9408xf32> to memref<?xf32>
+    // %c9408_3 = arith.constant 9408 : index
+    // call @_memcpy_f16_f32(%11, %13, %c9408_3) : (memref<?xf16>, memref<?xf32>, index) -> ()
+    // clang-format on
+
+    valResultTmp = b.create<memref::AllocOp>(loc, testType);
```
krzysz00: Ok, here, I don't think we need to allocate. Why not instead loop through the test result, cast each element to the narrower type, then cast it back. `extf` and `truncf` should do it. (along with `createAffineLoopNest`)

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
```diff
@@ -842,3 +843,188 @@ mcpuConv2dInt8(int64_t rank1, void *f_ptr, int64_t rank2, void *i_ptr,
...
+  double sumRelDiff = 0.0;
+  float maxVAL_rel = 0.0f;
+  float maxGPU_rel = 0.0f;
+  // Metric RMS
+  float maxMag = 0.0f;
+  double sumDiffSq = 0.0;
+  // histogram
+  // hist_relDiff[9]++ when cpuVal == 0
+  int hist_relDiff[10] = {0};
+  for (int64_t i = 0; i < dataSize; ++i) {
```
krzysz00, line 939: Could we extend this loop sot hat, if a value goes over the absolute or relative threshold, we print the index of the value, the expected result, and the actual result?

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -1753,7 +1650,9 @@ populateHostHarnessLogic(ModuleOp &module,
     mlir::Value testResult = localVars[outIdx];
     mlir::Value valResult = valVars[outIdx];
 
-    auto verifierFunc = createVerifierFunc(module, root0, genConfig);
+    auto testType = testResult.getType().template dyn_cast<MemRefType>();
```
krzysz00: I don't think you need `template` here.

krzysz00 (commented):
Ok, so now that I'm low on fundamental concerns, some commentss.

mlir/tools/miopen-gen/miopen-gen.cpp
zhanglx13: How about reusing `arg0` to hold the narrower type of the validation result?
```cpp
emitMemcpy(b, arg1, arg0);
emitMemcpy(b, arg0, arg1)
```
Since `arg0`'s values are copied (and extended) to `testResultNew`, it should be safe to overwrite it.

krzysz00: Yeah, it should be safe, but that sort of aliasing feels like a trap that'll come back to bite us when we forgot we did it. It feels like it'd be simpler to write an in-place elementwise truncate + extend. Heck, you could probably use a `linalg.generic` for it.

(and part of the reason I was wanting to cut down on extra buffers was that these buffers are already on the large side)

krzysz00 (approved):
Looks good to me, though can we get the printing of the failing value?

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
```diff
@@ -930,6 +931,9 @@ void mcpuVerify(T *gpuResults, T *validationResults, int64_t dataSize,
       }
       // Accumulate square root
       sumDiffSq += static_cast<double>(absDiff) * static_cast<double>(absDiff);
+      // Print out values if difference is larger than threshold
+      if (absDiff > thr_absDiff || relDiff > thr_relDiff)
+        printf("%ld: %f %f %f %lf\n", i, valNum, gpuNum, absDiff, relDiff);
```
zhanglx13, line 979: @krzysz00 It is printed here.

krzysz00: Oh, lol, I'm absolutely brilliant at this. And so that's basically all my comments, I think this is a good improvement!

jerryyin:
Please hold off merging for now, I'll review today to peek into some implementation details.

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
```diff
@@ -842,3 +843,192 @@ mcpuConv2dInt8(int64_t rank1, void *f_ptr, int64_t rank2, void *i_ptr,
       stride_h, stride_w, padding_h_l, padding_h_r, padding_w_l, padding_w_r,
       dilation_h, dilation_w, xdlops);
 }
+
+int findIdxHistRelDiff(double relDiff) {
```
jerryyin: Some nits:
 - Return should be of `size_t`
 - Implementation could take advantage of a statically populated 1:1 mapping 
   - This use relDiff to reference the mapping (in a while loop)
   - print functions can reference both index and relative diff (in a while loop)

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
```diff
@@ -842,3 +843,192 @@ mcpuConv2dInt8(int64_t rank1, void *f_ptr, int64_t rank2, void *i_ptr,
...
+    } else {
+      float absDiff = fabs(valNum - gpuNum);
+      // Update maxAbsDiff and its correspinding pair of values
+      if (absDiff > maxAbsDiff) {
+        maxVAL_abs = valNum;
+        maxGPU_abs = gpuNum;
+        maxAbsDiff = absDiff;
+      }
+      sumAbsDiff += static_cast<double>(absDiff);
+      // Update maxRelDiff only if cpuVal != 0
```
jerryyin, line 957: [Question] So we abandoned the `Relative difference (ignore small denominators)` but used the consistent `relative difference` method in validation?

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
```diff
@@ -842,3 +843,192 @@ mcpuConv2dInt8(int64_t rank1, void *f_ptr, int64_t rank2, void *i_ptr,
...
+            static_cast<double>(absDiff) / (static_cast<double>(fabs(valNum)));
+        hist_relDiff[findIdxHistRelDiff(relDiff)]++;
+        if (relDiff > maxRelDiff) {
+          maxVAL_rel = valNum;
+          maxGPU_rel = gpuNum;
+          maxRelDiff = relDiff;
+        }
+        sumRelDiff += relDiff;
+      } else {
+        hist_relDiff[9]++;
```
jerryyin: Could you add a line of comment in the conversion function that last index represent Nan?

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
```diff
@@ -842,3 +843,192 @@ mcpuConv2dInt8(int64_t rank1, void *f_ptr, int64_t rank2, void *i_ptr,
...
+      }
+      // Accumulate square root
+      sumDiffSq += static_cast<double>(absDiff) * static_cast<double>(absDiff);
+      // Print out values if difference is larger than threshold
+      if (absDiff > thr_absDiff || relDiff > thr_relDiff)
+        printf("%ld: %f %f %f %lf\n", i, valNum, gpuNum, absDiff, relDiff);
```
jerryyin: Since this is still in the loop, could this end up printing too much messages?

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
```diff
@@ -842,3 +843,192 @@ mcpuConv2dInt8(int64_t rank1, void *f_ptr, int64_t rank2, void *i_ptr,
...
+    }
+  }
+  double aveAbsDiff = sumAbsDiff / static_cast<double>(dataSize);
+  double aveRelDiff = sumRelDiff / static_cast<double>(dataSize);
+  double err_RMS = sqrt(sumDiffSq) / (static_cast<double>(maxMag) *
+                                      sqrt(static_cast<double>(dataSize)));
+  // Verbose information about the difference
```
jerryyin: I'd prefer:
 - An on/off switch to printing
 - An independent function to print overall
 - And in which `hist_relDiff` abstracted to another function
   - histogram printing implemented in a loop 

BTW, why not use the more modern `cout` for printing?

mlir/tools/miopen-gen/miopen-gen.cpp
```diff
@@ -1195,248 +1204,148 @@ createVerifierFunc(ModuleOp &module, const KernelIF &kernel,
...
-  if (elemType.isIntOrIndex()) {
-    printFunc = makeFuncDecl(module, "mcpuPrintInt32", {intType, intType});
+  auto mr5DUnkType = MemRefType::get({-1, -1, -1, -1, -1}, valElemType);
+  bool isTestAndValSameType =
+      (testOutType.isInteger(32) || testOutType.isF32());
+
+  mlir::Value testResult, valResult; // Values passed to the verify function
+  mlir::Value testResultNew;         // Values used for type conversion
+  if (!isTestAndValSameType) {
+    // When gpu kernel output data type = f16 | bf16, type conversions
```
jerryyin, line 1301: [Question] I'm trying to understand why we need those type conversions? If gpu kernel output data type is f16, can't we also have cpu computation output data type to be f16?

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
zhanglx13: Can you elaborate on the "statically populated 1:1 mapping" point?

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
zhanglx13: Yes. It is hard to define what should be treated as "small".
However, it is expected that the threshold for relDiff should be larger than that of relDiff (ignore small denominators) for both f16 and f32.

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
jerryyin: It looks like in both cases you'd need `relDiff` to index mapping so it doesn't have to be both fwd and backward, and a c++ map would work.

```cpp
const map<double, int> histRelDiffConversion = {(0, 0}, {1e-6, 1}, {1e-5, 2} .... }
```

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
zhanglx13: Yes, just like the old verification function. 
It is a little bit annoying when the threshold is set too small, the verification function keeps printing out nearly all elements. But on the other hand, it provides information about how badly is the error. 
Maybe we can just print out the 10 largest errors with the total number of elements whose difference is above the threshold.
Since @krzysz00 explicitly requested this to be printed, what do you think?

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
zhanglx13: For the switching point, would it be too much details for `miopen-gen` users?
When `check-mlir-miopen` is invoked, `lit` only matches the last line of the output, i.e. `[1 1 1]`. 
When `miopen-gen` is invoked explicitly, I think the developer expects to see those debug information. Therefore, the on/off functionality is fulfilled by the way `miopen-gen` is used.

I'll refactor the histogram printing process.
> BTW, why not use the more modern cout for printing?

I do not see any reasons not using it. I can switch to it.

mlir/tools/miopen-gen/miopen-gen.cpp
zhanglx13: From the validation function perspective, since @yiqian1 has implemented different validation functions ([#738](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/738)), we may have some options:

1. C/C++ implementation: it is much easier to use f32 as the data type
2. MLIR implementation: I am not sure if it is possible to perform the computation in f16. But as a validation function, f32 is generally used to validate f16 computations. Can you confirm, @yiqian1?

From the verification function perspective, since it is implemented in C/C++, f32 is the best option for the computation. Hence the conversion.

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
zhanglx13: Each element in `hist_relDiff[]` represents an interval. I think we have to compare `relDiff` with the boundaries to determine which interval it falls into. Did I miss something here?

jerryyin: Ah somehow I get myself confused on this. If that's the case, forget about the map, could you use a while loop to implement this?

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
jerryyin: Here are my thoughts on the use case when we want to turn print debug messages to off:
 - In automated testing, if we'd only match [1 1 1] then we should try to limit the error printing to only those. From robustness point of view.
 - In development setting, it could be tedious to always see a bunch of information printed on screen for a passing test.

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
zhanglx13: Sure. I can do that.

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
krzysz00: I do want that full error trace on error - not having those in can make debugging tricky, though it does require a `| tail`

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
krzysz00: I agree on the development case.

However, I'll note that `FileCheck` suppresses output of passing tests.

If we're worried about false-positive matches, we can add some sort of prefix to the result line - say `RESULT: `, to make it harder to accidentally write a debug print that makes the test pass

krzysz00: Also, @jerryyin , I'd actually prefer `printf()` here. It makes it much easier to control things like float precision. `cout`'s API for that is ugly

mlir/tools/miopen-gen/miopen-gen.cpp
krzysz00: I will say that, given that MIOpen already has it and that it's a header-only library, it's not inherently unreasonable to pull in `half.hpp` and make this all templatized on datatype. We could probably similarly pull in a bfloat type

krzysz00:
On the printing question, I'd be in favor of a switch that allows (although implementing it cleanly might require moving to the string stream infrastructure, negating my printf() preferences)
- Print debug output always
- Print debug output on failure (the default?)
- And _maybe_ print debug output never

But the on-failure case means `stringstream` or futzing with the `?s?printnf()`s

mlir/tools/miopen-gen/miopen-gen.cpp
jerryyin: @zhanglx13 I wouldn't worry too much on having to implement something for this point. It isn't unreasonable for fp16 to be validated in fp32 your and your comments reminded me the reason of it. Thanks.

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
```diff
@@ -844,41 +845,55 @@ mcpuConv2dInt8(int64_t rank1, void *f_ptr, int64_t rank2, void *i_ptr,
...
-    2: 1e-6 - 1e-5
-    3: 1e-5 - 1e-4
-    4: 1e-4 - 1e-3
-    5: 1e-3 - 1e-2
-    6: 1e-2 - 0.1
-    7: 0.1 - 1
-    8: >= 1
-   */
+size_t findIdxHistRelDiff(double relDiff,
+                          const llvm::SmallVector<double, 7> bucketBoundary) {
```
krzysz00: I don't think you want to pass vectors by value like this.
If anything, I'd go and make this a `static const double BUCKET_BOUNDARIES[NUM_BUCKETS] = {...};`
with a `constexpr size_t NUM_BUCKETS = 7;` before it

krzysz00: But, failing that, take an `ArrayRef`

krzysz00 (commented):
Minor quibbling re bucket boundaries

Also, would it make sense to define an `enum class` for the debug print types and `static_cast` into it?

jerryyin (approved):
Nothing past @krzysz00's comments. LGTM and thanks for your work.

krzysz00 (approved):
This all looks good to me, approved

PR #737, opened 2022-08-23
Adding Softmax Conversion

manupak:
This commit adds the dialect conversion of
MIGraphX operator Softmax to a set of operators
in TOSA dialect.

* Added conversion tests
* Added a cpu e2e test

### DIFF
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -42,12 +42,12 @@ static TosaOp createOpAndInfer(mlir::PatternRewriter &rewriter, mlir::Location l
                         Args &&...args){
     auto op = rewriter.create<TosaOp>(loc, UnrankedTensorType::get(elemType), args...);
     InferShapedTypeOpInterface shapeInterface = cast<InferShapedTypeOpInterface>(op.getOperation());
-    SmallVector<ShapedTypeComponents> ReturnShape;
+    SmallVector<ShapedTypeComponents> returnShape;
     LogicalResult shapeInferenceStatus = shapeInterface.inferReturnTypeComponents(op.getContext(), op.getLoc(),
                                       op->getOperands(), op->getAttrDictionary(),
-                                      op->getRegions(), ReturnShape);
+                                      op->getRegions(), returnShape);
     assert(shapeInferenceStatus.succeeded());
-    Type newOutTy = RankedTensorType::get({ReturnShape[0].getDims()}, elemType);
+    Type newOutTy = RankedTensorType::get({returnShape[0].getDims()}, elemType);
     auto result = op->getResult(0);
     result.setType(newOutTy);
     return op;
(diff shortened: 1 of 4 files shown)


krzysz00:
Hi. While I'm not necessarily the best person to approve, since this isn't my chunk of the codebase, I'd like to say

1. You have good style! (although I haven't seen too many `final` classes in the MLIR codebase, but I don't see any reason not to have them)
2. I'm curious, why infer the shape instead of setting the shape right off the bat?

manupak:
Hi @krzysz00,

Thanks for taking a look!.

> I'm curious, why infer the shape instead of setting the shape right off the bat?

I think that is a usecase of the op implementer defining InferShapedTypeOpInterface in the operator definition. TLDR;

My rationale (Something I have discussed with @jungpark-mlir as well) : 

In this case, I felt we maintaining our bespoke logic in the conversion (there can be good reasons in other cases) is a duplication of what goes into InferShapedTypeOpInterface definition in a sense and just becomes a maintenance burden.

Some related discussions that I followed : https://discourse.llvm.org/t/better-layering-for-infershapedtypeopinterface/3823
"InferReturnTypeComponents : This method is used to get information of the return values when they are shaped type. This is meant to get the return type during operation creation."

krzysz00:
Huh. I wasn't expecting that to be what that interface was for, but it makes sense and I don't see any good reason not to use it. Good find!

(also that shape result vector's name should probably start in lowercase to match MLIR style, but that's a trivial nit)

manupak:
Ah thanks! I was getting confused between LLVM vs MLIR coding guidelines :) -- switched to MLIR now.

sjw36 (approved):
Looks good @manupak !

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -333,10 +352,43 @@ class DotConverter final : public OpConversionPattern<migraphx::DotOp> {
...
+
+    auto tosaMax = createOpAndInfer<tosa::ReduceMaxOp>(
+        rewriter, loc, elementType, input, axisAttr);
+    auto tosaSub = createOpAndInfer<tosa::SubOp>(rewriter, loc, elementType,
+                                                 input, tosaMax);
+    auto tosaExp =
+        createOpAndInfer<tosa::ExpOp>(rewriter, loc, elementType, tosaSub);
+    auto tosaReduceSum = createOpAndInfer<tosa::ReduceSumOp>(
+        rewriter, loc, elementType, tosaExp, axisAttr);
+    auto tosaReciprocal = createOpAndInfer<tosa::ReciprocalOp>(
```
sjw36, line 378: Could just use tosa.div

manupak: If Im not mistaken, tosa.div is for integer divide : https://mlir.llvm.org/docs/Dialects/TOSA/#tosadiv-mlirtosadivop

Also here : https://www.mlplatform.org/tosa/tosa_spec.html#:~:text=int32_t-,2.5.6.%20INTDIV,-Elementwise%20integer%20divide

sjw36: Ah, ok. Wow, that is strange... since it could easily be decomposed to RECIP and MUL.

manupak:
Thanks for the reviews @sjw36 @krzysz00 !

I did go ahead and merged but later doubting whether is a policy around merging. Please do let me know if there is any :)

PR #739, opened 2022-08-24
Improving MIOpenDriver.py Path resolutions 

manupak:
Currently, the MIOpenDriver.py needs to be run
in the build directory to correctly resolve the paths.

This commit does the following changes :

* Adds argparse to parse top-level arguments.
  (In future might want to consider adding a subparser
   for CLI-based kernel parameters)
* Adds mlir-build-dir and miopen-build-dir arguments
  that defaults to path finding that happens in
  either WORKSPACE directory or home directory.


manupak:
I came across this script trying to do some perf. experiments of my own and found it annoying that is was cluttering the build dir and had to run it from the build dir.

Let me know whether this is a good idea.

krzysz00 (changes requested):
One problem: In nightly CI, this is installed to the build directory using the `ci-performance-scripts` target and run from there.

So while I'm definitely in support of a change like this in principle, we'll need to handle both the case where the script is installed and the case where it isn't.

I'd be open to doing some filesystem crawling to find the binaries

manupak:
to keep it simple , how about we provide an optional arg to define the build dir path, because it is already oriented around the build dir ?

manupak:
More precisely my suggestion would be to change the following : 

https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/e182fc6336cfe8732a727e000dfc0b31f822155c/mlir/utils/jenkins/Jenkinsfile#L653-L657

to 

```
 dir('build') {
   sh 'python3 ./bin/MIOpenDriver.py -bdir ${WORKSPACE}/build -bmiopen_use_tuned_mlir'
   sh 'rm -rf ${WORKSPACE}/MIOpen/build/MIOpenUserDB'
   sh 'python3 ./bin/MIOpenDriver.py -bdir ${WORKSPACE}/build -bmiopen_use_untuned_mlir'
}
```

While the bdir argument could default to '{repo_root}/build' -- where '{repo_root} is discovered relatively to script location within the repo.

Im open to reverting the polarity as well -- that is to let default be the current dir. However, IMO, just reading the script it is not obvious that it had to be run on the build dir if one did not look at jenkins.

Let me know what you think.

krzysz00:
I'd argue for a Python version of my `rocm-run` script (though looking for source directory and build directory)

```
#!/bin/bash

declare -g BASEDIR="none-such"
for dir in "$PWD/.." "$PWD" "$PWD/build" \
                  "$HOME/llvm-project-mlir/build"; do
    if [[ -x "$dir/bin/mlir-rocm-runner" ]]; then
        BASEDIR="$dir"
        break
    fi
done

exec "${BASEDIR}/bin/mlir-rocm-runner" "--shared-libs=${BASEDIR}/external/llvm-project/llvm/lib/libmlir_rocm_runtime.so,${BASEDIR}/lib/libconv-validation-wrappers.so,${BASEDIR}/external/llvm-project/llvm/lib/libmlir_runner_utils.so" --entry-point-result=void "$@"
```

manupak:
Having another look at the script, basically the script requires to know two things : 

A0. The build dir (the runner executables and libs are based on this)
A1. The location where MIOpen build dir and bins are (this is currently assumed to be relative to A1, which I think is also not right even in this PR)

I think above two are actually a choice of the user (that choice could even be not named as "build") of MIOpenDriver.py script, that user being a developer or Jenkins.

Therefore, I'd argue we should put two arguments specify the following : 

B0. mlir-build-dir
B1. miopen-build-dir

However, I suppose your concern is coming from the angle that the user (dev or Jenkins) has extra burden to provide these as well, though it could be argued it is as if asking the script to do work to find what the user consciously decided on (i.e. the build dirs of MLIR and MIOpen).

Therefore, accommodating both sides, I think we could compromise on having the B0 and B1 as arguments to scripts that defaults to path finding as you suggested, but I would push us to have them as arguments.

krzysz00:
Yeah, you're right there. Having those directories be arguments with some default path-finding (and the note that if the MIOpen directory doesn't exist, you can still be running performance tests)

Heck, we probably should give this script proper command-line argument parsing since we're here...

manupak:
Ok, I ve done what we discussed and added a top-level argparse infra as well.
(Though there is a underlying arg parsing happens that looks like it needs of subparser -- however, that would require a breaking change, therefore thought this would be good for now).

PTAL when you have some time.

mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -11,29 +12,91 @@
...
 DIRECTIONS = ['-F 1', '-F 2', '-F 4']
 DATA_TYPES = ['conv', 'convfp16', 'convint8']
 LAYOUTS = ['NHWC', 'NCHW']
+
 # Compiled regexp object used for extracting elapsed time from MIOpenDriver's output
 ELAPSED_TIME_RE = re.compile(r"Elapsed: (.*)ms")
 
+class Paths(NamedTuple):
+    """This structure is used to hold paths needed to perform the tests"""
+    miopen_gen_path : str
```
krzysz00, line 35: Can we make this a `Path` object or, failing that, `bytes`? Paths aren't necessarily UTF-anything.

mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -11,29 +12,91 @@
...
+    mlir_miopen_driver_path : str
+    miopen_driver_path : str
+    rocm_runner_path : str
+    libmlir_rocm_runtime_path : str
+    libconv_validation_wrappers_path : str
+    libmlir_runtime_utils_path : str
+    configuration_file_path : str
+
+
+def find_mlir_build_dir() -> str:
```
krzysz00, line 45: Types aside, I'd like to see this first check for `./bin/miopen-gen` and `../../../build/bin/miopen-gen`. A full autodetector will complain about my home directory, which has both `~/llvm-project-mlir/` and `~/llvm-project-mlir-upstream/`

mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -11,29 +12,91 @@
...
+    workspace_dir = os.environ.get('WORKSPACE', str(Path.home()))
+    assert workspace_dir, "Cant find WORKSPACE env arg or home directory"
+    miopen_opt_path = glob.glob(workspace_dir + '/**/bin/miopen-opt', recursive=True)
+    assert len(miopen_opt_path) == 1, "Multiple paths found to contain */bin/miopen-opt"
+    miopen_opt_path = miopen_opt_path[0]
+    build_dir = Path(miopen_opt_path).parent.parent
+    return str(build_dir)
+
+
+def find_miopen_build_dir() -> str:
```
krzysz00, line 76: I'd start this by looking relative to `mlir_build_dir` first

mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -302,24 +366,24 @@ def runConfigWithMIOpenDriver(commandLine, envs):
...
     config = ConvConfiguration.fromCommandLine(commandLine, xdlops)
     # get nanoseconds from MIOpenDriver output
-    nanoSeconds = runConfigWithMIOpenDriver(commandLine, envs)
+    nanoSeconds = runConfigWithMIOpenDriver(commandLine, paths, envs)
     return config.tableEntry(nanoSeconds)
 
 #Generate MLIR vs. MIOpen performance results
-def generatePerformanceResults(configs, xdlops):
-    mlir_df = pd.DataFrame(benchmarkMLIR(testVector.split(sep=' '), xdlops)
+def generatePerformanceResults(configs, xdlops, paths: Paths):
```
krzysz00, line 413: Ooh, type annotations. I/we should've added those earlier...

mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -378,54 +442,134 @@ def tuneMLIRKernels(configs, xdlops):
...
+    xdlop_supported_gpus = ['gfx908', 'gfx90a']
+    xdlop_supported_gpus_str = xdlop_supported_gpus[0]
+    for gpu in xdlop_supported_gpus[1:]:
+        xdlop_supported_gpus_str += '|' + xdlop_supported_gpus_str
+    r = subprocess.run(f"/opt/rocm/bin/rocm_agent_enumerator -t GPU | grep -E '{xdlop_supported_gpus_str}'", shell=True)
+    if r.returncode == 0:
+        return True
+    return False
+
+def add_double_dash_convention(args):
```
krzysz00, line 486: Let's just make them double-dash arguments. This is an internal script that gets called from ~1 place, we can break stuff

mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -378,54 +442,134 @@ def tuneMLIRKernels(configs, xdlops):
                 print("MIOpen tuning timed out")
                 outs, errs = p1.communicate()
 
+def is_xdlops_present() -> bool:
+    """This function checks whether a GPU with xdlops support is present"""
+    xdlop_supported_gpus = ['gfx908', 'gfx90a']
+    xdlop_supported_gpus_str = xdlop_supported_gpus[0]
+    for gpu in xdlop_supported_gpus[1:]:
+        xdlop_supported_gpus_str += '|' + xdlop_supported_gpus_str
+    r = subprocess.run(f"/opt/rocm/bin/rocm_agent_enumerator -t GPU | grep -E '{xdlop_supported_gpus_str}'", shell=True)
```
krzysz00: Might as well make this a quiet grep?

krzysz00 (commented):
Overall:
1. Thank you! It's good to see this code cleaned up, and I hope that "hey, let's get the perf measuring scripts cleaned up" wasn't a overly-pointless initial bit of work
2. Since this is all internal, we really don't have too strong of an API commitment - you can break the arguments formats and such just fine
3. I would still like to see some initial tests before we break out the globs
4. Speaking of breaking changes, can you make sure the other clients of MIOpenDriver (I think `parameterSweeps.py` is one) aren't broken here? (and that script in particular could probably use `is_xdlops_present()`

mlir/utils/jenkins/MIOpenDriver.py
manupak: Will the upstream one has miopen-opt as well ? 

Or

what is special about miopen-gen ?

mlir/utils/jenkins/MIOpenDriver.py
manupak: any reason ?
It could be anywhere right ?

mlir/utils/jenkins/MIOpenDriver.py
manupak: we can -- it is just the consumers of this has to convert them to strings to be used by sub process.
Do you see a benefit in doing so ?

mlir/utils/jenkins/MIOpenDriver.py
krzysz00: Sorry, `llvm-project-mlir-upstream` is a second clone of our repository I use to have a clean copy of the `miopen-dialect` branch to make testing easier.

And I'd just misspoken about which executable you were looking for, none of them matter

mlir/utils/jenkins/MIOpenDriver.py
manupak: I was worried about me not knowing all the consumers.
Who are consumers ?

I can see Jenkinsfile  and  parametersSweeps.py being the consumers.

mlir/utils/jenkins/MIOpenDriver.py
krzysz00: If `subprocess` will only take strings, then, yeah, ignore all that and use strings. However, I think you can pass a `Path` subclass to the subprocesses runner, but it might take `shell=False`

mlir/utils/jenkins/MIOpenDriver.py
krzysz00: While it could be anywhere, I'd look just under the root of the MLIR repository first. That's where the CI puts it and where a lot of people have it symlinked in, so that'll save on glob calls

mlir/utils/jenkins/MIOpenDriver.py
manupak: Any suggestion to differentiate between those ?
or 
do we want to ? because the alternative is to provide a path to build dir

krzysz00: Start from the path where the script is located. From there, test if there's a `./miopen-[something]`. If not, check to see if you're running from source by testing for a `../../../build/bin/miopen-[something]`. And then, as a final fallback, either tell the user to pass a path or do the search you're currently doing, either's probably fine.

Similarly, to get the MIOpen path (which you probably only want to test for if the user's test scheme asked for MIOpen runs), check for `[script directory]/../../MIOpen` and `[script directory]/../../../MIOpen/` to cover the built and source case, respectively.

mlir/utils/jenkins/MIOpenDriver.py
krzysz00: Yeah, they're it. Anything using this file is in this repository.

mlir/utils/jenkins/MIOpenDriver.py
manupak: Makes sense! I ll change this tommorow when Im bored trying to figuring out other things.

mlir/utils/jenkins/MIOpenDriver.py
manupak: Makes sense! I ll change this tommorow when Im bored trying to figuring out other things.

mlir/utils/jenkins/MIOpenDriver.py
manupak: Ack

mlir/utils/jenkins/MIOpenDriver.py
manupak: Having thought about this more, I feel keeping them in strings is much useful.
It is true that Path object can be passed onto subprocess but there are not passed as it is -- there are additional arguments concat'd to them.
Moreover, there library paths here that will form into composite arguments such as "--shared-libs".

Therefore, to keep consistency, I feel keeping them as strings is a better choice.

mlir/utils/jenkins/MIOpenDriver.py
manupak: Done (I think)

mlir/utils/jenkins/MIOpenDriver.py
manupak: Done (I think)

mlir/utils/jenkins/MIOpenDriver.py
manupak: I had a look but I feel we should do the breaking change in a isolated PR. This PR is already seeing symptoms of scope creep.

manupak (commented):
Thanks for the review.

I changed path finding logic as discussed and changed parameterSweeps that was affected due to path finding, as discussed.

Due to scope creep, I feel we should do the breaking change in a follow up (to be revertible incase).

krzysz00 (approved):
Fantastic cleanup, approved!

PR #740, opened 2022-08-24
Remove unused runners and toy from CMake

manupak:
This commit removes unused runners and toy example 
from the cmake build system and tests.


krzysz00 (approved):
Thanks for taking the hot minute to clean this up!

Since you're already in these files, could you clear out stuff like the mentions of `toy`, which don't show up in our test suite?

manupak:
I think I've removed them now. (Lets see what the CI says about it)

krzysz00:
Just as a minor process note re commits
- We have a rough preference for squash-and-merge, and so it's not particularly required to `--amend`/rebase your changes
- However, if a branch mixes changes to `external/llvm-project/` with changes to the rest of the repository, it needs a "[DO NOT SQUASH]" on it and it's placed in via merge commit or rebase
- That being said, it's not a bad thing to rebase etc, it's just not necessary (upstream seems to have a stronger desire for large single commits, but we don't have as much of that here)

manupak:
Thanks for the info about about the process :).

> We have a rough preference for squash-and-merge, and so it's not particularly required to --amend/rebase your changes

Ack, this make sense.

What is the policy for the commit message though then ? More particularly, on the commits following the first one.. Do we use bullet points ?

> However, if a branch mixes changes to external/llvm-project/ with changes to the rest of the repository, it needs a "[DO NOT SQUASH]" on it and it's placed in via merge commit or rebase

In my experience, DO NOT SQUASH approaches have led into trouble when we want to bisect the branch as they could leave the tree broken at places unless individual commits are verified. Do we verify each commit ?

> That being said, it's not a bad thing to rebase etc, it's just not necessary (upstream seems to have a stronger desire for large single commits, but we don't have as much of that here)

Agreed. In my experience, this usually comes from having better commit messages and verifiable tree at each commit.

krzysz00:
Yeah, you got us. We don't have the best commit message discipline, and our [DO NOT SQUASH] aren't too bisectable. In the [DO NOT SQUASH] case, I try and `rebase -i` my commits into a coherent set of commits, but that's not done universally.The reason for not squashing is that we don't want upstream merges to have to deal with commits that also include internal changes.

PR #745, opened 2022-08-30
Make miopen.transform into a more typical view op, fix memory space

krzysz00:
Thanks to Simon for noticing these issues with how transform was
defined, which causes issues with bufferization.


sjw36 (approved):
Look good.

PR #746, opened 2022-08-30
[DO NOT SQUASH] MLIR#633: Support multiple or independent transposes on miopen.conv2d params

sjw36:
* removed memory space param on miopen.transform, gets from input
        * removed affine_map on output of miopen.transpose, encoded in op
    
        https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/633

### DIFF
--- a/mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
+++ b/mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
@@ -356,6 +356,7 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
         convOp->setAttr("expected_filter_layout", b.getStringAttr(layout));
         top.replaceAllUsesWith({tInput});
       }
+      top.erase();
     }
 
     return success();
--- a/mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
+++ b/mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
@@ -192,13 +192,13 @@ public:
     Value output = rw.create<memref::AllocOp>(loc, outputType);
 
     SmallString<5> filterLayout("kyxcg");
-    if (auto attr = op->getAttrOfType<StringAttr>("expected_filter_layout"))
+    if (auto attr = op->getAttrOfType<StringAttr>("filter_layout"))
       filterLayout = Twine(attr.getValue() + "g").str();
     SmallString<5> inputLayout("nhwcg");
-    if (auto attr = op->getAttrOfType<StringAttr>("expected_input_layout"))
+    if (auto attr = op->getAttrOfType<StringAttr>("input_layout"))
       inputLayout = Twine(attr.getValue() + "g").str();
     SmallString<5> outputLayout("nhwkg");
-    if (auto attr = op->getAttrOfType<StringAttr>("expected_output_layout"))
+    if (auto attr = op->getAttrOfType<StringAttr>("output_layout"))
       outputLayout = Twine(attr.getValue() + "g").str();
 
     if (failed(makeMIOpenConv2D(rw, op, input, inputLayout, filter,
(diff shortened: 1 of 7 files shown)


krzysz00 (commented):
From a quick skim, high-level thughts:
1. This PR accidentally almost-duplicates #745 - looks like you'd done the fix and were waiting to push it as part of this, and like I'd woken up today and gone "I've got a moment, let me fix up that issue Simon found". Would you like me to drop #745, or (to give my prefference) should we merge #745 and then this PR gets cleaner?
2. Looking at the actual TosaToMIOpen changes, I'm not sure I'm a fan of all the precise matching and if statements. I'll note that, even though our outer infrastructure is only set up for NCHW and NHWC, the conv-to-gemm pass can (legacy tuning heuristics for vectorization aside) handle any layout - that is, if you called for a WHCN input, it should by all rights work.
3. So, wouldn't it make more sense to implement transposes by permuting the expected layout? That is, you'd start with `"nhwc"` or even `['n', 'h', 'w', 'c']` and rewrite transposes by permuting the letters in the string.

sjw36:
@krzysz00

1. Yes, that's fine, let's take yours. thanks
2. I thought there was a hard requirement that HW must be together.. is that gone?
3. If 2 is no longer a req, it could easily be done

krzysz00:
2. I'm not aware of such a requirement - or at least I didn't see anything that would impose it in ConvToGemm. It might be bouncing around `--affix-tuning-params`, but I don't remember seeing it there either. I know we do rather precisely guard the `isUnfold` checks in `--conv-to-gemm`, but that's a matter of performance, not correctness

sjw36:
'verifyConvOp' does the check for HW.

But I am adding the permutation code anyway, perhaps we can remove that check.

krzysz00:
Thanks for finding the check - we'll want to reexamine it in a release or two

jungpark-mlir (approved):

mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
```diff
@@ -297,148 +281,89 @@ class MatMulConverter final : public OpConversionPattern<tosa::MatMulOp> {
...
+    }
+    if (Operation *cval = v.getDefiningOp<tosa::ConstOp>()) {
+      auto cattr = cval->getAttr("value").cast<DenseElementsAttr>();
+      return SmallVector<int32_t>(cattr.getValues<int32_t>());
     }
-    // all matches.
-    return true;
+    // May be bufferization cast
+    //  but this is no longer a bufferization pass, so assert
+    assert(0);
```
krzysz00, line 298: Nit: can we get an error in the assert?

mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
```diff
@@ -297,148 +281,89 @@ class MatMulConverter final : public OpConversionPattern<tosa::MatMulOp> {
...
+      // tosa.conv2d output is transpose
+      permuteLayout(convOp, "output_layout", "nhwk", dims);
+      convOp->getResult(0).setType(tOutput.getType());
+      top->replaceAllUsesWith(convOp);
+    } else {
+      // trace output to tosa.conv2d
+      for (auto &use : tOutput.getUses()) {
+        if (auto op = dyn_cast<tosa::Conv2DOp>(use.getOwner())) {
+          if (convOp)
+            return failure();
```
krzysz00, line 346: Why return failure?

How about collecting up all the convolution ops that use this and doing the permutation logic to each of them?

(Especially since, IIRC, transpose is NoSideEffect, and so, if you rewrite away every user of the transpose, the transpose will disappear in the next `--canonicalize` or earlier)

mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
```diff
@@ -297,148 +281,89 @@ class MatMulConverter final : public OpConversionPattern<tosa::MatMulOp> {
...
+        } else {
+          return failure();
+        }
+      }
+
+      // conv Input Modifier
+      if (convOp.getOperand(0) == tOutput) {
+        // input feature map
+        permuteLayout(convOp, "input_layout", "nhwc", dims, true);
+        top.replaceAllUsesWith({tInput});
```
krzysz00, line 357: And per above, for the multi-convolution case etc, you could just replace that particular use.

(and you'd return failure() on the next go-around through the pattern when you find no convolutions that can be rewritten)

krzysz00 (commented):
I've got some thoughts on how to make this a bit more general, but I don't see any reason this couldn't land

mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
sjw36: For now, this case should not be encountered as we only have 1 conv2d per kernel.

mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
sjw36: see comment above.

mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
sjw36: should not encounter this case and the message is in the comment.

mlir/lib/Conversion/TosaToMIOpen/TosaToMIOpen.cpp
krzysz00: Why not prepare for the possibility now instead of having mysterious "Lowering failed" when someone tries it in the future?

krzysz00: That being said, I'm not going to insist on actually handling the general case. It'd be nice if we could print some sort of diagnostic when failing here though

PR #747, opened 2022-08-30
Use FMA in non-xdlops kernel

krzysz00:
Since we have access to a FMA intrinsic that lowers to instructions avaialable
on all GCN+ (gfx900+) chips via LLVM's built-in FMA type, we might as well use
it.

This also changes the non-xdlops threadwise gemm to always use vector
transfer reads and writes so that we have vector-typed values, thus
making the FMA intrinsic work (though we could've used 0-D vectors instead)
*and* enabling future work on using dot product intrinsics (those intrinsics
will want to set loadKpackLen to some higher value that reflects the dot
product being carried out).

Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/634

### DIFF
--- a/mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
+++ b/mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
@@ -93,7 +93,6 @@ struct ThreadwiseGemmRewritePattern
     SmallVector<int64_t, 4> dimensions = {k, m, n, kPack};
     SmallVector<int64_t, 4> strides = {1, 1, 1, loadKpackLen};
     auto abType = VectorType::get(loadKpackLen, dataType);
-    auto cType = VectorType::get(1, dataType);
 
     TopDownTMBuilder aView(b, {"k", "m", "n", "kpack"}, dimensions, loc);
     aView.ignore("n");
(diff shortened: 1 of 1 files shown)


mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -117,29 +128,30 @@ struct ThreadwiseGemmRewritePattern
...
+      ValueRange cCoords = gemmLoop.getLowerCoords(/*domain=*/2);
+      Value cVal =
+          b.create<vector::TransferReadOp>(loc, cType, bufferC, cCoords,
+                                           /*inBounds=*/ArrayRef<bool>(true));
+      Value result;
+      if (dataType.isa<IntegerType>()) {
+        Value mul = b.create<MulIOp>(loc, aVal, bVal);
+        result = b.create<AddIOp>(loc, mul, cVal);
+      } else if (dataType.isa<FloatType>()) {
+        result = b.create<vector::FMAOp>(loc, aVal, bVal, cVal);
```
jerryyin: This is a different approach than xdlops end of "force" it to use a specific intrinsic. In terms of hit or miss rate, how much would this actually ended up in the dot instructions? I think this is a reasonable patch so that nonxdlops kernel can "occasionally" use dot ops when size happen to matches, but ultimately we'd need a new threadwisegemm to do it exclusively using that.

sjw36 (approved):
Nice improvement.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
krzysz00: Ok, so, this doesn't do dot instructions at all. Code-structure wise, I'm pretty sure blockwise-to-threadwise extends everything to F32/I32 before we reach this point. More fundamentally, I'm not sure the compiler will optimize 
```
%v1 = fma %a0, %b0 : f16
%v2 = fma %a1, %b1 : f16
%v1_ext = ext %v1 : f32
%v2_ext = ext %v2 : f32
%acc= fadd %acc, %v1_ext, %v2_ext : f32
```
to a dot product

jerryyin (approved):

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
jerryyin: That was my wrong assumption since we had that dot instruction conversation so close before...

krzysz00:
Perf numbers for this show an average 1.5x improvement - a lot of kernels are now trice as fast, and those that aren't are still substantially quicker.

Also, in interesting notes, the FP16 kernel sees the compiler optimizing to these instructions on Navi21
```
        v_fma_mix_f32 v85, v86, v90, v85 op_sel_hi:[1,1,0]
        v_fma_mix_f32 v84, v86, v90, v84 op_sel:[0,1,0] op_sel_hi:[1,1,0]
```
and so on
[fma-perf.zip](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/files/9464112/fma-perf.zip)

PR #748, opened 2022-08-31
Fusion fix for broadcast

jungpark-mlir:
- Optimize the copy with collapseShape, replace with expandShape to the original memory
- handle broadcast input and reshape in the align pattern
- type checking assertion needs to be changed since this case covers elementwise input has different shape as the C-tensor (but it's broadcast and checking if it's broadcastable)

### DIFF
--- a/mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
@@ -169,6 +169,16 @@ static void insertCopyFromOtherArg(PatternRewriter &b, Location loc,
                                    Value dest) {
   LLVM_DEBUG(llvm::dbgs() << "Src type: " << srcOp.getType()
                           << " dest type: " << op.dest().getType() << "\n");
+  ArrayRef<int64_t> sType, dType;
+  sType = srcOp.getType().cast<ShapedType>().getShape();
+  dType = op.dest().getType().cast<ShapedType>().getShape();
+  assert(sType.size() == dType.size() &&
+         "Rank of extra fusion arguments matches shape of C tensor");
+  for (unsigned i = 0; i < sType.size(); i++) {
+    assert((sType[i] == dType[i] || sType[i] == 1) &&
+           "shape of extra fusion arguments matches shape of C tensor or "
+           "broastable");
+  }
 
   auto writeCLoop = op->getParentOfType<TransformingForOp>();
   assert(writeCLoop && "threadwise_copy_v2 must be in a transforming_for");
@@ -438,10 +448,13 @@ LogicalResult MILARewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
   }
   // 2. Apply if input found
 
-  // move reshape before la access it.
-  if (out.getDefiningOp<memref::ExpandShapeOp>()) {
-    auto parentTransff = copyOp->getParentOp();
-    out.getDefiningOp()->moveBefore(parentTransff);
+  // point back to original memory.
+  Value delOut = nullptr;
+  if (memref::ExpandShapeOp expanded =
+          out.getDefiningOp<memref::ExpandShapeOp>()) {
+    if (out.hasOneUse())
+      delOut = out;
+    out = expanded.getSrc();
   }
 
   Value gemmV2Outs = copyOp.source();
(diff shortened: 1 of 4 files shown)


mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -466,6 +503,10 @@ LogicalResult MILARewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
     return success();
   }
 
+  // Delete expandShapeOp whose single use has been dropped
+  if (delOut)
```
krzysz00: Do we need this? Won't memref.expand_shape die by canonicalization?

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -291,6 +298,21 @@ static ThreadwiseCopyV2Op traceToThreadwiseCopy(Value inp) {
       allValidUses = false;
     }
   }
+
+  // broadcast input
```
krzysz00: Can we expand this comment a bit to mention that this handles the case where the output of our conv2d is `expand_shape`d

krzysz00 (approved):
Minor nits, but they're nothing big, at least from my angle. Mostly calling for slightly more detailed comments.

I'd wait for @sjw36 to look at the copy-opt changes

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -172,9 +172,16 @@ static void insertCopyFromOtherArg(PatternRewriter &b, Location loc,
...
-         "shape of extra fusion arguments matches shape of C tensor");
+  ArrayRef<int64_t> sType, dType;
+  sType = srcOp.getType().cast<ShapedType>().getShape();
+  dType = op.dest().getType().cast<ShapedType>().getShape();
+  assert(sType.size() == dType.size() &&
+         "Rank of extra fusion arguments matches shape of C tensor");
+  for (unsigned i = 0; i < sType.size(); i++) {
+    assert((sType[i] == dType[i] || sType[i] == 1) &&
+           "shape of extra fusion arguments matches shape of C tensor or "
+           "broastable");
```
sjw36: sp. broadcastable

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -423,6 +451,15 @@ LogicalResult MILARewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
   }
   // 2. Apply if input found
 
+  // point back to original memory.
+  Value delOut = nullptr;
```
sjw36: Value defaults to null. Assignment unnecessary.

sjw36 (approved):
Overall good! Just some nits.

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
sjw36: And this code is preceded by an unconditional return.

krzysz00: Ok, so we definitely don't need this and can remove `delOut` entirely.

mlir/lib/Dialect/MIOpen/Transforms/AlignTiling.cpp
```diff
@@ -317,6 +339,12 @@ static Value reconfigureLAGeneric(PatternRewriter &b,
       Value newInput;
       if (inp == twout) {
         newInput = laIn;
+      } else if (auto expanded = inp.getDefiningOp<memref::ExpandShapeOp>()) {
```
sjw36: You can eliminate this special case by calling traceToThreadwiseCopy to check if this is the threadwisecopy output.

mlir/lib/Dialect/MIOpen/Transforms/CopyOpt.cpp
```diff
@@ -59,6 +59,7 @@ template <typename T> struct MICORewritePattern : public OpRewritePattern<T> {
     // 1. Capture allocation->copy pattern
     Operation *writer = nullptr;
     memref::CopyOp reader = nullptr;
+    memref::CollapseShapeOp csop = nullptr;
```
sjw36, line 62: This should be generalized for all subsequent collapse/expand ops. Not required for this commit but please file a ticket with testcase for multiple reshapes in various combinations.

jungpark-mlir: https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/644 filed.

PR #749, opened 2022-09-01
Bump dockerfile to ROCm 5.2.3

jerryyin:


krzysz00:
Have the MIOpen dependencies changed recently, and should we bump that commit as well while we're here?

zhanglx13:
I am just curious about is there a way to `wget` the latest MIOpen instead of maintaining a commit? Or is there a good reason to do so?

jerryyin:
@zhanglx13 That's actually a good idea. By looking at when the docker is populated, we know which version of MIOpen dependency we get.

zhanglx13 (approved):

krzysz00 (approved):
@zhanglx13 , great idea with the wget! I'm in favor of all of this

PR #751, opened 2022-09-01
Disable relative tests in random host e2e test for fp32

krzysz00:
due to fma and host code disagreement on small values


zhanglx13 (approved):
Let me leave a note here.
- For now, the f32 tests could fail with random inputs between [-1, 1] with the relDiff metric. This is because the new relDiff metric does not ignore small element values.
- f16 tests works fine with relDiff since it is disabled for f16.
- absDiff works since now it is disabled for all data types.

More details can be found at [PR#736](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/736)
In the future, these tests will be ported using the auto-gen mechanism. And we will figure out different metric+tolerance combination for different groups of tests.

jerryyin:
@krzysz00 Maybe file a "TODO" ticket to remind us to revert this to [1, 1, 1] once @zhanglx13 has made the default relative diff range to [-1, -3] or [1, 3]. I believe this is just another example of denorm behavior that shall be fixed once the range is migrated beyond [-1, 1].

PR #754, opened 2022-09-01
Moving M/NRepeats to xdlops gemm as looping constructs

jerryyin:
This PR did the following:
 - Get rid of the hard-coded repeats definition in blockwisegemmV2
   - It is xdlops gemm's responsibility to perform the execution in register level
 - M/N Repeats become a loop so we can potentially go with arbitrary repeats numbers (with larger tile size)
 - Abstracted the Repeats in coordinate transform infrastructure
   - The common coordinate space is [m, n, k, v] with:
     - m: MRepeats
     - n: NRepeats
     - k: KPerThread
     - v: Vector length along the result dimension, such that if the group of xdlops generate 2 result vectors, the number is 2
   - Inconsistencies between K dimension caused by divergence between reduction/non-reduction will be address later
   - Note: I struggled with how to use the proper coordinate space on especially v dimension and ended up deciding it is key to describe a xdlops gemm.
 - Fix and amend unit tests

### DIFF
--- a/mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
+++ b/mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
@@ -533,9 +533,21 @@ struct BlockwiseGemmV2RewritePattern
       lklb.create<memref::StoreOp>(loc, valueB, bufferB, ValueRange{lkliv});
     }
 
-    b.replaceOpWithNewOp<XdlopsGemmV2Op>(
-        op, b.getIndexAttr(0), b.getIndexAttr(0), adaptor.bufferA(),
-        adaptor.bufferB(), adaptor.matrixC(), tuningParams);
+    // Workload of either MPerWave and NPerWave that are larger
+    // than wave size of 64 will be executed by repeats
+    // TODO: amend this for tuning parameter selection as well
+    xcs = XdlopsCodeSelection::get(dataType, MPerXdlops, NPerXdlops, b);
+    Value reshapedARegisters = reshapeBuffer(
+        b, loc, adaptor.bufferA(), {"m", "k"}, {MRepeats, KPerThread});
+    Value reshapedBRegisters = reshapeBuffer(
+        b, loc, adaptor.bufferB(), {"n", "k"}, {NRepeats, KPerThread});
+    Value reshapedCRegisters =
+        reshapeBuffer(b, loc, adaptor.matrixC(), {"m", "n", "v"},
+                      {MRepeats, NRepeats, xcs.nResultVectors});
+
+    b.replaceOpWithNewOp<XdlopsGemmV2Op>(op, reshapedARegisters,
+                                         reshapedBRegisters, reshapedCRegisters,
+                                         tuningParams);
     return success();
   }
 };
--- a/mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
+++ b/mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
@@ -533,50 +533,9 @@ struct BlockwiseGemmV2RewritePattern
       lklb.create<memref::StoreOp>(loc, valueB, bufferB, ValueRange{lkliv});
     }
 
-    XdlopsGemmParamsAttr threadwiseParams = tuningParams;
-    if (MRepeats > 1 || NRepeats > 1) {
-      // Hard-coded m_per_wave/n_per_wave as 64 when MRepeat>1 or NRepeat>1.
-      // So each xdlops_gemm_v2 handles a 64x64 GEMM.
-      threadwiseParams = b.getAttr<XdlopsGemmParamsAttr>(
-          tuningParams.getKPerBlock(), tuningParams.getMPerBlock(),
-          tuningParams.getNPerBlock(), tuningParams.getKpack(), /*mPerWave=*/64,
-          /*nPerWave=*/64);
-    }
-    if (MRepeats == 1 && NRepeats == 1) {
-      b.create<XdlopsGemmV2Op>(loc, b.getIndexAttr(0), b.getIndexAttr(0),
-                               adaptor.bufferA(), adaptor.bufferB(),
-                               adaptor.matrixC(), threadwiseParams);
-      b.eraseOp(op);
-    } else if (MRepeats == 2 && NRepeats == 1) {
-      // Original C++ logic.
-      // p_c_thread.s.x.l = XdlopsGemm.template Run<M, N, K>(
-      // p_a_block, p_b_block, p_c_thread.s.x.l);
-      // p_c_thread.s.y.l = XdlopsGemm.templateRun<M, N, K>(
-      // p_a_block + MPerXdlops, p_b_block, p_c_thread.s.y.l);
-
-      b.create<XdlopsGemmV2Op>(loc, b.getIndexAttr(0), b.getIndexAttr(0),
-                               adaptor.bufferA(), adaptor.bufferB(),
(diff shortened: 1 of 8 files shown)


mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -533,50 +533,21 @@ struct BlockwiseGemmV2RewritePattern
...
-                               adaptor.matrixC(), threadwiseParams);
-      b.create<XdlopsGemmV2Op>(
-          loc, b.getIndexAttr(0), b.getIndexAttr(KPerThread), adaptor.bufferA(),
-          adaptor.bufferB(), adaptor.matrixC(), threadwiseParams);
-      b.eraseOp(op);
-    }
-
+    // Workload of either MPerWave and NPerWave that are larger
+    // than wave size of 64 will be executed by repeats
+    // TODO: amend this for tuning parameter selection as well
```
krzysz00, line 538: Ok, so, these are genuinely repeats (having looked at the XdlopsCodeSelection code). So we might want to do some stuff with how the results of XdlopsCodeSelection are structured so that, for example, we don't have to go back in and hardcode that 128x128 is permitted.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -163,6 +163,14 @@ struct XdlopsGemmV2RewritePattern : public OpConversionPattern<XdlopsGemmV2Op> {
     int64_t MPerWave = tuningParams.getMPerWave();
     int64_t NPerWave = tuningParams.getNPerWave();
 
+    // Workload of either MPerWave and NPerWave that are larger
+    // than wave size of 64 will be executed by repeats
+    // TODO: amend this for tuning parameter selection as well
+    int64_t MRepeats = (MPerWave > 64) ? (MPerWave / 64) : 1;
```
krzysz00: ... hold on a moment, couldn't you pull MRepeats out of the xdlops code selection and compute the PerXdlops as MPerWave / MRepeats? 

I think the high-level change might be having `imms` only mention the instructions for one repeat and then making `nResultVectors` include the repeat terms.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -194,136 +202,163 @@ struct XdlopsGemmV2RewritePattern : public OpConversionPattern<XdlopsGemmV2Op> {
...
-          return arg;
+    auto generateMfmaOnKDim = [&](Value &regAOffset, Value &regBOffset,
+                                  Value &regCOffset) {
+      if (KPack == 1) {
+        auto mfmaLoop = b.create<TransformingForOp>(
+            loc, ArrayRef<ValueRange>{{zeroConstantOp}},
+            ArrayRef<Attribute>{b.getArrayAttr({})},
+            /*bounds=*/ArrayRef<int64_t>{KPerThread},
+            /*strides=*/ArrayRef<int64_t>{k_base},
+            /*useIndexDiffs=*/true, /*forceUnroll=*/true);
```
krzysz00: Do me a favor and double check the ordering of useIndexDiffs and forceUnroll? They're used inconsistently in this PR and I can't remember what order they go in

krzysz00 (commented):
I have a higher-level thought:
Can we get rid of all this kPack switch-case by having blockwise gemm load up a scalar memref and then having the loop just go through k with a stride of k_base?
I have a sense that making these intermediate buffers vector-typed isn't needed and might be leading to extra complexity.
Especially since kPack is a LDS optimization that doesn't necessarily affect the xdlops gemm ... seeing as, unlike the general one, we don't have an outer loop over k.

jerryyin:
> Can we get rid of all this kPack switch-case by having blockwise gemm load up a scalar memref and then having the loop just go through k with a stride of k_base?

Yes that's possible and if I understand it right then I'd expect that to be addressed in: https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/623.

I think once the definition on the K are unified between kpack and non-kpack then naturally there is one of the two ways to refactor this:
 - Make nonKPack as a KPack of 1
 - Or, in the opposite way, make KPack not encoded as part of the type and use scalar instead (as you suggested)

I like the second approach better as well so I'll prioritize and experimenting the second one, assuming it doesn't cause any additional vgpr allocation we should end in a unified case here and across many other places in the stack.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
krzysz00: Poking this comment, especially since I'm not liking these harcoded 64s

Though an immediate fix would be calling that number `waveSize` so we can keep an eye on it later.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -194,136 +202,163 @@ struct XdlopsGemmV2RewritePattern : public OpConversionPattern<XdlopsGemmV2Op> {
...
-    }
+    };
+
+    auto gemmLoop = b.create<TransformingForOp>(
+        loc, ArrayRef<ValueRange>{startCoords, startCoords, startCoords},
+        ArrayRef<Attribute>{aTransforms, bTransforms, cTransforms}, dimensions,
+        /*strides=*/ArrayRef<int64_t>{1, 1, KPerThread, nResultVectors},
+        /*forceUnroll=*/true, /*useIndexDiffs=*/false);
+    {
+      Value regAOffset = gemmLoop.getLowerCoords(/*domain=*/0)[0];
```
krzysz00, line 358: Could we maybe go for `regAIdx` or `regAIndex` and so on throughout? Offset makes it sound like we're going to be doing another layer of indexing on top of that

krzysz00 (commented):
See minor nits above and in the previous round of comments, but I think all this makes sense and am generally good to approve it

mlir/lib/Dialect/MIOpen/Transforms/BlockwiseGemmToThreadwise.cpp
jerryyin: Yes, that's the goal. The upper limit where we need to hard code to should be 64x64. Any larger tile size should be derived by repeats. I'll address it in next release time frame.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
jerryyin: You are right about repeats, they should be gone since that was the point of reshaping from blockwisegemm. I'd have to keep M/N Per Xdlops computation just so that it is able to inquire on the xdlops code selection again.

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
jerryyin: That's right this one is swapped.

krzysz00 (approved):
Minor naming quibble about `offset` for coordinates, which isn't that big a deal esp. since we might end up rewriting this whole thing later, approved

mlir/lib/Dialect/MIOpen/Transforms/ThreadwiseGemmLowering.cpp
jerryyin: Oops sorry I think I didn't see this comment while working on it, otherwise this'd be already included in latest commit. But judging that this is a naming change I'll skip it for now.

PR #757, opened 2022-09-06
Fixed perf-config apply for padded gemms

manupak:
Currently the perf-config application relies on
params to be aligned to per wave parameters.

This commit changes them to be padded and tried
out because it would be required for kernel dims
that are not alignable (e.g. 7x7).

* Fold the semi-duplicated populate padding derived to a
   single populate derived.
* This incurs a behavorial change in the compiler to pad
   the where possible rather than trying the non-pad
   configs first.


manupak:
This improves the 7x7 conv2d by 21% (at least on 112) with xdlops by allowing the tuning to actually happen on 7x7 as every tunable instance requires padding.

mlir/utils/jenkins/miopen-tests/resnet50-miopen-configs
```diff
@@ -20,4 +20,4 @@
 -n 256 -c 64 -H 56 -W 56 -k 256 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
 -n 256 -c 64 -H 56 -W 56 -k 64 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
 -n 256 -c 64 -H 56 -W 56 -k 64 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
-# FIXME: -n 256 -c 3 -H 230 -W 230 -k 64 -y 7 -x 7 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -t 1
+-n 256 -c 3 -H 230 -W 230 -k 64 -y 7 -x 7 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -t 1
```
krzysz00, line 23: Could you also add this test to the e2e test lists in the `.toml`s in `mlir/test/e2e/`?

krzysz00 (commented):
Is this perhaps https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/627 ? That is, support for using any set of tuning parameters with padding?

It looks like it is, and, if so, thank you!

However, it looks like this induces an increase in the number of times we pick a padding kernel.

So, I'd like to check
1. Performance numbers
2. Do we still have the loop that checks if there exists a set of tuning parameters in our heuristic list that can be used without padding? I'd like to see that, as we want to pick a non-padding kernel over a padding kernel where possible, I claim. (or maybe further experimentation will reveal we don't).

manupak:
Thanks for taking a look and this is new territory for me -- so feel free to point out things :) 

> Is this perhaps https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/627 ? That is, support for using any set  of tuning parameters with padding?

Looks like a happy accident!

> Performance numbers

I think from a MIOpen solver find PoV, it does not really matter (Does it ?).
I think that because it will search the space anyway and will pick non-padded one if that is what is perfomant. right ?

I think what gets affected might be non-tuned performance (if its a valid user usecase, is it ?).
If so I can insert re-ordering function to order the padded ones to come last in the init parameter list.

> Do we still have the loop that checks if there exists a set of tuning parameters in our heuristic list that can be used without padding? I'd like to see that, as we want to pick a non-padding kernel over a padding kernel where possible, I claim. (or maybe further experimentation will reveal we don't).

Same point as above.

krzysz00:
Yeah, from the tuned case, it'll just pick the best one, agreed.

But, for the non-tuned case, I'd like to see a function like `Optional<InitParamsSubclass> findUnpaddedParams(GemmSize)` that'll go through our list and give you the largest non-padding kernel, if any.

I'd also then argue for the heuristic that, if we must pad and we haven't been given tuning parameters, we should go for a known-good set of parameters (which is what I think the universal parameters were).

mlir/utils/jenkins/miopen-tests/resnet50-miopen-configs
manupak: Done

manupak:
Update : I've restored the existing behavior that is to attempt the non-padding param sets first and then attempt padding based param sets -- which I think was behavior before this PR.

I think we should do further re-ordering based on the padding value in a next PR that is slightly orthogonal compared to being able to tune kernels (e.g. 7x7) that are not tunable at all -- which is what is being fixed here (among few cleanups :) ).

I ll probably look at that as the next task in the scope of https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/627 which I have assigned myself now.

update : created PR https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/760 for that.

jerryyin (changes requested):
I'm putting a blocker now in case this get approved and merged. Between me and @manupak we synced up offline. I'm putting a blocker here just because it enabled padding cases for tuning, an area we had too little test coverage to make me confident on shipping this in production environment.

mlir/test/Dialect/MIOpen/affix_tuning_params_conv2d.mlir
```diff
@@ -226,3 +228,24 @@ func.func @miopen_conv2d_bwd_weight_padALL_f16(%filter : memref<1x20x8x3x3xf16>,
...
+    gemm_id = 0 : i32,
+    input_layout = ["ni", "gi", "ci", "hi", "wi"],
+    numCu = 120 : i32,
+    output_layout = ["no", "go", "ko", "ho", "wo"],
+    padding = [0 : i32, 0 : i32, 0 : i32, 0 : i32],
+    perf_config = "64,256,8,64,64,4,1,1",
+    strides = [2 : i32, 2 : i32]
+  } : memref<1x64x3x7x7xf32>, memref<256x1x3x230x230xf32>, memref<256x1x64x112x112xf32>
+  return
+}
```
krzysz00: Nit: missing final newline?

krzysz00 (commented):
While I suspect we might not want to land this right before the release ... this all seems pretty good to me

manupak:
Yea, I agree with landing this closer to release is risky -- hence not a good idea.

However, it would be great for us to sketch out / quantify what it means to achieve testing parity with non-padded kernels.
So we can rely on testing to de-risk and improve confidence of this change.

Me and @jerryyin discussed certain aspects related solvers and invoker APIs using padded kernels -- specifically the existence of different solvers for padding. 

Having thought about it overnight, 
Im not 100% understanding / buying the design choice of bifurcations of solvers based on padding as I believe whether to use padding is a function(conv_params, tuning_params) -- therefore there is not really a choice between solvers when the conv_params+tuning_params decide whether to use padding or not (i.e. they are mutually exclusive) -- but I must admit I am yet to learn about why the invoker APIs for padded kernel would be different (possibly due to workspace requirements). I think I recognize the need for a different tuning parameter space -- both dimensionality and range -- might be a reason but that would not necessarily mean we need to do propagate that down to MLIR.

I think I need have to more discussions to understand how we end up different solvers (non-padding and padding) in MIOpen and whether we should follow the same for MLIR-compiler based solution.

krzysz00:
Re padding kernels
1. There used to be a bigger split, especially when we didn't have the map-analysis-based detection of when we needed to test for OOB coordinates and so it was easier to have padding bugs
2. These days, having padding needed locks you into a particular set of parameters (something I'm pretty confident we can remove, but *shrugs*) and enables the usage of the atomic add backwards weight kernel
3. Said backwards weight kernel is still probably going to need to be a no-padding situation for the immediate future: when I tried to enable using that algorithm in the presence of GEMM padding, things mysteriously didn't work, and I don't recall why. I think it was some interaction with the `KBlocks` parameter and how padding in the GEMM's k dimension interacted with that ...
4. The implementations from `gridwise_gemm` on down don't care about padding - they have the invariant that `K % KPerBlock == 0 && M % MPerBlock == 0 && N % NPerBlock == 0`. So I'd argue there's little risk for enabling padding for more kernels, but there is the possibility that there's some weird interaction between tuning parameters, the presence of bounds checks, and so on, that would reveal a bug.
5. I think that if we ran the parameter sweep script for perf configs with the new changes, we could get pretty solid test coverage

jerryyin:
@krzysz00 @manupak 

While I agree all the technical analysis of why and how they *should* work, I continue to think we need more proof on it is working on all the combinations we supported. Until I see that proof coming I'll have to keep this PR blocked. I'm acting conservative here in that enabling of padding applicability range will expose us to a much larger config space. I don't want us to be the reason that crash tensorflow and pytorch. That will do more damage than help because MIOpen already have CK based padding kernels that get tuned.

#### My concerns in test and validation are:

Our only config that will run into padding is the 7x7 filter yet it is still disabled for some unknown reason. I claim that in order to enable padding, at least half the size versus the non-padding cases configs needs to be added and tested. I don't think the two configs in parameter sweep script is enough judging by the how limited the situation we could run into (plus, int8).

The other things is, the applicability range of padding and non-padding kernels can change. To make sure testing has good and orthogonal coverage on both, we may need a special flag that enforce padding and other cases non-padding as a long term solution.

#### My concerns in design are:

Can we definitively guarantee that, between padding the non-padding kernel:
- Will kernel interface always the same (in particular the additional workspace requirement)
- Will tuning parameters always the same (indeed the existing tuning parameters manipulation is an ugly hack and padding kernel's tuning parameters may play better on its own)

I need a strong guarantee on that otherwise we'd open ourselves to potential bugs or degraded performance in this area. Once this PR merge, we'd have a strong coupling between the two kinds of kernels.

krzysz00:
Ok, so, I'm not sure what you're worried about
1. We already have tests for the padding case in e2e - we know it works. Sure, they're not canonical resnet50 or anything, but they're there.
2. I meant the sweep of tuning parameters, not the sweep of convolution configs. This is a change to how padding requirements interact with GEMM size and tuting parameters - the convolution is irrelevant.
3. What workspace? In what universe do we need a workspace for GEMM?
4. There isn't a "padding kernel". There's a particular set of tuning parameters we use when there's no set of tuning parameters that's applicable to a config (either from our heuristic list or from tuning when the user called for a fallback).

manupak:
Again Im actually agreeing that we should do the proper testing and need to assert that this wont do additional breakages.

Update on testing : I've verified the following using MIOpen and MLIR solvers (locally) for 7x7 : 

* {fp32, fp16} x {fwd, bwd, wrw}
* int8 x fwd -- seems like MIOpen does not support any backward implementions.

bf16 does not seem to work with MIOpen + MLIR solvers in the trunk even in a case where padding not needed as follows :

i.e. : 
```
MIOPEN_FIND_ENFORCE=4 MIOPEN_DRIVER_USE_GPU_REFERENCE=1 MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvMlirIgemmFwdXdlops ../../../MIOpen/build/bin/MIOpenDriver convbfp16 -F 1 -f NCHW -I NCHW -O NCHW -n 256 -c 64 -H 56 -W 56 -k 64 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1 

MIOpenDriver convbfp16 -F 1 -f NCHW -I NCHW -O NCHW -n 256 -c 64 -H 56 -W 56 -k 64 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
MIOpen Error: /home/mkarunar/work/MIOpen/src/ocl/convolutionocl.cpp:538: Forward Convolution cannot be executed due to incorrect params
RunForwardGPU() FAILED, rc = 0x7
Forward Convolution FAILED: 1.79769e+308 > 0.0656
```

Im not very much familiar with parameterSweeps and I ll take a look at that next.

jerryyin:
@krzysz00 

> We already have tests for the padding case in e2e - we know it works. Sure, they're not canonical resnet50 or anything, but they're there.

If we are talking about the same thing, then 90% of tests under `mlir/test/mlir-miopen-driver/auto_e2e/*padding*` are that 7x7 config. I think it is not enough.

> I meant the sweep of tuning parameters, not the sweep of convolution configs.

This is consistent with last one. I think sweep of only tuning parameters is not enough. Without enough regression on configs, there are certain conditions in the code we'd never hit. Like with existing parameter sweep, for the two configs it exercised it'd never hit the reduction xdlops.

> What workspace? In what universe do we need a workspace for GEMM?

Well, we don't need workspace for gemm but we do for convolution. For backward weight convolution, depends on the situation we may or may not need a workspace. I'm not sure padding kernel has been looked as closely on the situation as non-padding ones.

> There isn't a "padding kernel". There's a particular set of tuning parameters we use when there's no set of tuning parameters that's applicable to a config

This is true but not necessarily the best implementation. CK's version implementation of padding kernel has a different set of tuning parameters and range. There must be a reason that they need to tune differently. That's also why I call our version a hack based on it derived its range from non-padding ones.

jerryyin:
@manupak Sounds like a good start. Just so you know you can try to have CI do it for you for auto-e2e tests. Also, we'd need some more configs than 7x7 to test with tuning to be sure.

krzysz00:
The workspace we need for the atomic add case because fp16 atomic add is incorrect, it's nothing to do with padding.

manupak:
# Update on testing : 

From scanning the codebase, I have found three types of e2e testing : 

T1) test/mlir-miopen-driver/auto_e2e/ : This folder seems to contain a bunch .mlir files for chosen few configs and these are direct LIT tests and does not have codified parameterization.

T2) test/e2e/ : This folder contains two .toml files that seems to contain the layers of Resnet50 and Resnext101. These only define the convolution structure and auto generation of tests will cover data_type and direction coverage.

T3) parameterSweeps.py : this seems to do a codified parameterization of conv structures (among others).

## Existing coverage for padding kernels 

 coverage of padding kernels of conv dimension of (k, x, y, hi, wi, c)

T1) :

(64, 3, 3, 11, 11, 7) , 
(17, 3, 3, 11, 11, 13), 
(13, 3, 3, 11, 11, 64), 
(6, 7, 7, 32, 32, 3), 
(19, 3, 3, 11, 11, 64), 
(64, 7, 7, 230, 230, 3),  
(11, 3, 3, 7, 7, 13), 
(64, 7, 7, 224, 224, 3)

T2) :

(256, 3, 3, 28, 28, 256), 
(128, 3, 3, 56, 56, 128),
(64, 7, 7, 230, 230, 3),
(256, 3, 3, 56, 56, 256),
(64, 7, 7, 224, 224, 3),
(512, 3, 3, 28, 28, 512),
(512, 3, 3, 14, 14, 512),
         
T3) :  

(7, 2, 2, 4, 4, 7) 

Few observations : 

T1) It is not obvious to figure out the coverage just looking at *.mlir ones and the cross product for dtypes and directions with these configs seemed sparse.

T2) A much better auto-generation and a way to ensure coverage in terms of config x dtypes x directions (where its legal and known to be supported)

T3) Though its only using a single tuple for filter dimensions, it is much straightforward to way to guarantee testing parity with non-padded kernels.

## Non-padded test coverage : 

T1) None -- seems like only padding cases are here (correct me if I am wrong)
T2) Altogether there were 39 cases and padded cases were 7 as shown above thus have 32 more in .toml files.
     
However, if you look at https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/miopen-dialect/mlir/test/e2e/Resnext101Config.toml, the **non-padded cases** they mostly cover are :  
(1x1 , 3x3) x input/output_channels (128, 256, 512, 1024) x in_h/w (7, 14, 28, 56) 

T3) same coverage as padded kernels

## Summary and Questions 

Firstly, It is not entirely clear/evident where the 90% coverage of all e2e tests for non-padded kernels are coming from -- Surely I must be missing something. Therefore if you can point me where and which test cases that is there for non-padded kernels, I could work towards adding them.

Secondly, If we are to add test cases, I feel T3) seems to be a better option to guarantee the testing parity as compared to T1) and T2). Is there any historical precedence/preference in which way the e2e test should be added out of the three ways : T1, T2 and T3 ? 

##  References : 

T3) : https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/04079bf93c191d7e03e0e27ad965e4d3e7aa37f9/mlir/utils/jenkins/parameterSweeps.py#L280-L285

The logic to determine where the lowering will result in a padded gemm : 

https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/04079bf93c191d7e03e0e27ad965e4d3e7aa37f9/mlir/lib/Dialect/MIOpen/Tuning/GemmContext.cpp#L8-L27

The current tuning parameter space lower bounds : 

| dtype    | xdlops | MPerBlock | KPerBlock | NPerBlock |
|----------|--------|-----------|-----------|-----------|
| int8     | Yes    | 4         | 8         | 16        |
| int8     | No     | 32        | 4         | 32        |
| non-int8 | Yes    | 4         | 1         | 16        |
| non-int8 | No     | 32        | 4         | 32        |

So the tuning parameter space is power of 2 above those bounds. Therefore, combining the above table with GemmContext conversion logic -- one could determine whether lowered GEMM will be padded.

manupak:
## Update on design concerns 

> Will kernel interface always the same (in particular the additional workspace requirement)

My current understanding from above and parameterization of T2 and T3 based testing all the directions are covered and all invoker APIs are tested. I would expect those to fail some of those if the invoker API is different. Correct me if I am wrong.

> Will tuning parameters always the same (indeed the existing tuning parameters manipulation is an ugly hack and padding kernel's tuning parameters may play better on its own)

There are only two special solvers for padded for Igemm in MIOpen that is for fwd and wrw.

### MIOpen's fwd tuning parameter space : 

non-padded vs padded : it seems the the only difference in tuning space is the addition of https://github.com/ROCmSoftwarePlatform/MIOpen/blob/62b4df14325a60201a2e208abac563d5cbd4eadf/src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops_padded_gemm.cpp#L127-L128

References : 

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/62b4df14325a60201a2e208abac563d5cbd4eadf/src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/62b4df14325a60201a2e208abac563d5cbd4eadf/src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops_padded_gemm.cpp

### MIOpen's wrw tuning parameter space 

non-padded vs padded : there is no difference

References : 

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/62b4df14325a60201a2e208abac563d5cbd4eadf/src/solver/conv_hip_implicit_gemm_wrw_v4r4_xdlops.cpp#L99-L118

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/62b4df14325a60201a2e208abac563d5cbd4eadf/src/solver/conv_hip_implicit_gemm_wrw_v4r4_xdlops_padded_gemm.cpp#L104-L119

jerryyin:
### Re: Existing coverage for padding kernels

There are in total of T1: 8 configs, T2: 7 configs and T3: 1 config. Have you tested that there are indeed that many (8 + 7 + 1) configs that will exercise padding case? I can't comment on T1 and T3, but for T2 last time when I run benchmarking for resnet50 or resnext101 I think there's only 1 or 2 configs that will actually fall back to padding kernel. Because the padding decision always comes when the kernel is not supported by non-padding kernels, the only way to tell seem to run through them and see how they get lowered.

### Re: Questions

1. I think I was mostly referring to files from `padding_kernel_all_fwd`, `padding_kernel_all_wrw`, `padding_kernel_gemmK` and so forth files. Taking a second look after following your summary, I agree that `90%` statement is an overstatement. But I also want to point out that padding cases are highly spare/uneven, for example:
   - I do not see any int8 padding cases. I assume we do have support for that but it isn't tested.
   - It is not clear to me what `direction` + `dtype` support we actually have there. I'm assuming `bwd` + `xdlops` are never supported, and if that's the case are they rejected properly?

2. Agreed we should prioritize in adding T2 or T3 cases, but there may be some quirks mentioned on 1) that disallow you from doing that because of our uneven support in padding cases.

### Re: Design concerns

Agreed that for the tested cases should work, but we should focus on supported && untested cases. I'm not worried too much about unsupported cases as long as they get rejected properly.

Thanks for looking at the MIOpen repo for the padding kernels. I don't think the difference matters so far because we are not using them. The unanswered questions for me are:
 - If we ever *have to* include more tuning parameters for padding kernels, does the existing design prevent us from doing that?
 - I must point out those files are a stale version of CK (https://github.com/ROCmSoftwarePlatform/composable_kernel). To be really future proof I'd also take a look at the latest padding kernels and their tuning parameters there.

manupak:
Thanks @jerryyin for the quick response!

> There are in total of T1: 8 configs, T2: 7 configs and T3: 1 config. Have you tested that there are indeed that many (8 + 7 + 1) configs that will exercise padding case? 

All the configs 8 + 7 + 1 fails based on the following experiment that would otherwise pass. Therefore, I would take that as evidence that they are using padded kernels.

```
LogicalResult PopulateParams::populatePaddingKernelDerived(
    ConvolutionContext &ctx, const InitParamsNonXDL &param, GemmSize &gemmSize,
    DerivedParams &gemmADerivedParam, DerivedParams &gemmBDerivedParam,
    DerivedBlockGemmParams &blockGemmDerivedParam,
    DerivedOutParams &gemmCDerivedParam, uint32_t &gridSize) {

  LogicalResult res = failure();
  InitParams paddingParam = getUniversalParameters();

  // find complete config for paddingParam
  // if mblock,nblock,kblock is the same with this tuning parameter
  if (paddingParam.gemmMPerBlock != param.gemmMPerBlock ||
      paddingParam.gemmNPerBlock != param.gemmNPerBlock ||
      paddingParam.gemmKPerBlock != param.gemmKPerBlock)
    return failure();

  if (gemmSize.gemmM % param.gemmMPerBlock != 0){
    gemmSize.gemmM = gemmSize.gemmM + (param.gemmMPerBlock -
                                       gemmSize.gemmM % param.gemmMPerBlock);
    llvm::errs() << "M dim needs padding.\n";
    assert(false);
  }

  if (gemmSize.gemmN % param.gemmNPerBlock != 0){
    gemmSize.gemmN = gemmSize.gemmN + (param.gemmNPerBlock -
                                       gemmSize.gemmN % param.gemmNPerBlock);
    llvm::errs() << "N dim needs padding.\n";
    assert(false);
  }

  if (gemmSize.gemmK % param.gemmKPerBlock != 0){
    gemmSize.gemmK = gemmSize.gemmK + (param.gemmKPerBlock -
                                       gemmSize.gemmK % param.gemmKPerBlock);
    llvm::errs() << "K dim needs padding.\n";
    assert(false);
  }
```

>  But I also want to point out that padding cases are highly spare/uneven

I agree with this statement for T1-style testing, therefore I think it make sense to look for T2 or T3 style tests for ascertain coverage where it is easy to make sure it is added.

> It is not clear to me what direction + dtype support we actually have there. I'm assuming bwd + xdlops are never supported, and if that's the case are they rejected properly?

In T1-style testing it is explicit but sparse.

In T2 all directions x dtype are tested. An example generated .mlir file would look like as follows : 

```
//This file is automatically generated. Do not edit!

// RUN: miopen-gen -batchsize=64 -in_channels=2048 -in_h=7 -in_w=7 -out_channels=512 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=0 --padding_w=0 --operation conv2d -fil_layout=gkcyx -in_layout=ngchw -out_layout=ngkhw -t f32  %pv %random_data %xdlops | mlir-miopen-driver -c | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CHECK_config_7_1
// CHECK_config_7_1: [1 1 1]

// RUN: miopen-gen -batchsize=64 -in_channels=2048 -in_h=7 -in_w=7 -out_channels=512 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=0 --padding_w=0 --operation conv2d -fil_layout=gkcyx -in_layout=ngchw -out_layout=ngkhw -t f16  %pv %random_data %xdlops | mlir-miopen-driver -c | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CHECK_config_7_2
// CHECK_config_7_2: [1 1 1]

// RUN: miopen-gen -batchsize=64 -in_channels=2048 -in_h=7 -in_w=7 -out_channels=512 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=0 --padding_w=0 --operation conv2d -fil_layout=gkcyx -in_layout=ngchw -out_layout=ngkhw -t i8  %pv %random_data %xdlops | mlir-miopen-driver -c | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CHECK_config_7_3
// CHECK_config_7_3: [1 1 1]

// RUN: miopen-gen -batchsize=64 -in_channels=2048 -in_h=7 -in_w=7 -out_channels=512 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=0 --padding_w=0 --operation conv2d -fil_layout=gkyxc -in_layout=nhwgc -out_layout=nhwgk -t f32  %pv %random_data %xdlops | mlir-miopen-driver -c | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CHECK_config_7_4
// CHECK_config_7_4: [1 1 1]

// RUN: miopen-gen -batchsize=64 -in_channels=2048 -in_h=7 -in_w=7 -out_channels=512 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=0 --padding_w=0 --operation conv2d -fil_layout=gkyxc -in_layout=nhwgc -out_layout=nhwgk -t f16  %pv %random_data %xdlops | mlir-miopen-driver -c | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CHECK_config_7_5
// CHECK_config_7_5: [1 1 1]

// RUN: miopen-gen -batchsize=64 -in_channels=2048 -in_h=7 -in_w=7 -out_channels=512 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=0 --padding_w=0 --operation conv2d -fil_layout=gkyxc -in_layout=nhwgc -out_layout=nhwgk -t i8  %pv %random_data %xdlops | mlir-miopen-driver -c | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CHECK_config_7_6
// CHECK_config_7_6: [1 1 1]

// RUN: miopen-gen -batchsize=64 -in_channels=2048 -in_h=7 -in_w=7 -out_channels=512 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=0 --padding_w=0 --operation conv2d_bwd_data -fil_layout=gkcyx -in_layout=ngchw -out_layout=ngkhw -t f32  %pv %random_data %xdlops | mlir-miopen-driver -c | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CHECK_config_7_7
// CHECK_config_7_7: [1 1 1]

// RUN: miopen-gen -batchsize=64 -in_channels=2048 -in_h=7 -in_w=7 -out_channels=512 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=0 --padding_w=0 --operation conv2d_bwd_data -fil_layout=gkcyx -in_layout=ngchw -out_layout=ngkhw -t f16  %pv %random_data %xdlops | mlir-miopen-driver -c | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CHECK_config_7_8
// CHECK_config_7_8: [1 1 1]

// RUN: miopen-gen -batchsize=64 -in_channels=2048 -in_h=7 -in_w=7 -out_channels=512 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=0 --padding_w=0 --operation conv2d_bwd_data -fil_layout=gkyxc -in_layout=nhwgc -out_layout=nhwgk -t f32  %pv %random_data %xdlops | mlir-miopen-driver -c | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CHECK_config_7_9
// CHECK_config_7_9: [1 1 1]

// RUN: miopen-gen -batchsize=64 -in_channels=2048 -in_h=7 -in_w=7 -out_channels=512 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=0 --padding_w=0 --operation conv2d_bwd_data -fil_layout=gkyxc -in_layout=nhwgc -out_layout=nhwgk -t f16  %pv %random_data %xdlops | mlir-miopen-driver -c | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CHECK_config_7_10
// CHECK_config_7_10: [1 1 1]

// RUN: miopen-gen -batchsize=64 -in_channels=2048 -in_h=7 -in_w=7 -out_channels=512 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=0 --padding_w=0 --operation conv2d_bwd_weight -fil_layout=gkcyx -in_layout=ngchw -out_layout=ngkhw -t f32  %pv %random_data %xdlops | mlir-miopen-driver -c | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CHECK_config_7_11
// CHECK_config_7_11: [1 1 1]

// RUN: miopen-gen -batchsize=64 -in_channels=2048 -in_h=7 -in_w=7 -out_channels=512 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=0 --padding_w=0 --operation conv2d_bwd_weight -fil_layout=gkcyx -in_layout=ngchw -out_layout=ngkhw -t f16  %pv %random_data %xdlops | mlir-miopen-driver -c | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CHECK_config_7_12
// CHECK_config_7_12: [1 1 1]

// RUN: miopen-gen -batchsize=64 -in_channels=2048 -in_h=7 -in_w=7 -out_channels=512 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=0 --padding_w=0 --operation conv2d_bwd_weight -fil_layout=gkyxc -in_layout=nhwgc -out_layout=nhwgk -t f32  %pv %random_data %xdlops | mlir-miopen-driver -c | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CHECK_config_7_13
// CHECK_config_7_13: [1 1 1]

// RUN: miopen-gen -batchsize=64 -in_channels=2048 -in_h=7 -in_w=7 -out_channels=512 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=0 --padding_w=0 --operation conv2d_bwd_weight -fil_layout=gkyxc -in_layout=nhwgc -out_layout=nhwgk -t f16  %pv %random_data %xdlops | mlir-miopen-driver -c | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CHECK_config_7_14
// CHECK_config_7_14: [1 1 1]
```

> I do not see any int8 padding cases. I assume we do have support for that but it isn't tested.

It is tested in T2 style tests for the combinations mentioned above and also you could see them above.

> It is not clear to me what direction + dtype support we actually have there. I'm assuming bwd + xdlops are never supported, and if that's the case are they rejected properly?

All T2 and T3 style tests cover direction x dtype combos. I think xdlops is checked on different CI job -- as I understood from @krzysz00 -- i.e. rocmhead vs ml-ci.

> Agreed we should prioritize in adding T2 or T3 cases, but there may be some quirks mentioned on 1) that disallow you from doing that because of our uneven support in padding cases.

Im not sure that I follow what is meant by "uneven support in padding cases" because I would not expect configx.mlir generated from T2 for padding case to pass if that was the case.


## Design Concerns

> If we ever have to include more tuning parameters for padding kernels, does the existing design prevent us from doing that?

Hmmm my take is padding of the gemm is an dependent effect of the choice of gemm dimensions and tuning params rather than an independent choice. Having said that, unless we have new tuning knob, we can still have different ranges for padding kernels only in MIOpen but we dont need to further bifurcate the MLIR compiler based on those ranges, prematurely.

At least the current MIOpen's tuning parameter space does not highlight such a divergent space is being used as opposed to non-padding.

> I must point out those files are a stale version of CK (https://github.com/ROCmSoftwarePlatform/composable_kernel). To be really future proof I'd also take a look at the latest padding kernels and their tuning parameters there.

This is new to me, Thanks! I ll take a look.

## Outstanding questions

I am still trying to find out what tests should be added in the light of T2 and T3 based testing to reach parity with non-padded kernels. Im saying this mainly because I still see T1) tests to be extra because there is no counter-part for non-padded tests for those and I see T2/T3 tests are where non-padded kernels are tested.

So : 

* Is there more places where non-padded e2e tests are run that I have missed in my analysis ?
* The difference between T1 and T2 (by looking at the conv properties), seemed to be diff in dims of input. 
  * Do we want T1 differential configs to be added to new .toml file (e.g.  PaddedGemmConfig.toml) ?
  * Do we want T1 differential configs to be part of the test space of T3 ?

krzysz00:
I'm not sure talking about a "padding kernel" is meaningful here. Whether or not we have padding in the (implicit) GEMM is a function of the tuning parameters and of the specified problem size. Since the tuning parameters can come from anywhere in a range, but not all points in tuning parameter space are valid ones, I don't think you can determine if a particular parameter + problem combination will require padding until you've checked.

And we already have a padding-dependent API bit: the workspace size for 16-bit wrw, which you only need if the kernel will be an atomic add kernel, which you only can use if there'll be no padding. Currently, we handle this via `calculatePaddingKernelSize()` getting called inside Miir.h (through some indirection), which flips through our list of heuristic parameters in order to identify if one doesn't need padding.

... Having reviewed all that logic, I think there's something more fundamental here: the selection of tuning parameters should be happening in Conv2DGenerator - either by the user specifying a perf config or by running our heuristic logic, and so `--affix-tuning-params` should more or less go away (or, as it stands right now, should be limited to computing derived parameters). Then, we know, as soon as we have a `Conv2DGenerator`, we have tuning parameters, which means we know if the kernel is a padding kernel or not.

krzysz00:
So, @jerryyin: potential bugs in Miir.h aside, the solution to padding vs. non-padding is that we know at he time the solver's being invoked if we'll need padding or not and we can act accordingly. If there are separate sets of tuning parameters for padding and non-padding kernels in some meaningful way (which I don't think we've seen any evidence of above), that what we actually want is the "if you'd need to pad, fail" solver and the "pad if you have to" solver.

jerryyin:
@krzysz00 I think we are in general on the same page.

> I think there's something more fundamental here: the selection of tuning parameters should be happening in Conv2DGenerator

Yes, that's been discussed for 3 releases now but never get bumped to enough priority. Though I think you also had the idea that we should do it after it is converted to gemm. So we may want to decide on that too.

> If there are separate sets of tuning parameters for padding and non-padding kernels in some meaningful way

I know for a fact that:
 - Only non-padding can be tuned, and there are logic to make sure if there's an incoming tuning parameter, it can only go to non-padding kernel
 - Padding kernel can only operate on heuristic mode, as backups

That's the reason why I get nervous because this PR has the potential to break the assumption and to cause bugs.

manupak:
Me @krzysz00 and @jerryyin had an offline sync and aligned our understanding on the changes done here.

As next steps, we agreed to add more configs that will results in padded gemm from T1 to T2-style .toml file (maybe named as PaddedGemmConfig.toml) to ascertain the direction x dtype coverage.

manupak:
@krzysz00 @jerryyin,

When you have a minute, PTAL. 
See if there anything more to be done here.

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
```diff
@@ -139,7 +139,41 @@ struct DerivedBlockGemmParams {
...
+  // otherwise maintaining the provided order.
+  // TODO(@manupak) : improve this to order based on amount of padding added.
+  std::vector<InitParamType>
+  orderInitParams(const ArrayRef<InitParamType> &initParams,
+                  const GemmSize &gemmSize) {
+    std::vector<InitParamType> paddingRequiredParams;
+    std::vector<InitParamType> paddingNotRequiredParams;
+    for (const InitParamType &param : initParams) {
+      if (isValidGemm(param, gemmSize).succeeded()) {
+        paddingRequiredParams.emplace_back(param);
```
krzysz00: Isn't this backwards?

krzysz00: Ok, the names are backwards

mlir/test/e2e/PaddedGemmConfig.toml
```diff
@@ -0,0 +1,48 @@
...
+values = ["-fil_layout=gkcyx -in_layout=ngchw -out_layout=ngkhw", "-fil_layout=gkyxc -in_layout=nhwgc -out_layout=nhwgk"]
+
+[[axis]]
+name = "data type"
+values = ["f32", "f16", "i8"]
+prefix = "-t"
+
+## PaddedGemm
+[[suite]]
+name = "config"
```
krzysz00: Can this be named "padded_gemm"?

krzysz00 (commented):
I think this works as written, broadly, but can you check the hasWorkspace() flow to make sure it'll still work correctly with these changes? (I think it will but I'm reviewing on mobile and so can't quite trace the code right now)

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
manupak: good catch the names are backwards

mlir/test/e2e/PaddedGemmConfig.toml
manupak: ack

mlir/test/e2e/PaddedGemmConfig.toml
```diff
@@ -0,0 +1,48 @@
+directory = "PaddedGemm"
```
jerryyin, line 1: Do you mind to get rid of any redundant configs that were sitting under auto-e2e directory? I think you have provided a fairly complete set of test in here.

mlir/test/e2e/PaddedGemmConfig.toml
manupak: done.

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
manupak: done.

mlir/test/e2e/PaddedGemmConfig.toml
manupak: Done.

krzysz00 (changes requested):
Ok, so, having checked, you'll need to fix needExtraPadBwdWeight in Conv2DGenerator.cpp to account for the fact that applying padding doesn't return a failure() anymore.

Long-term, the fix'll be to not have derived parameters and to have the tuning heuristic return a parameter vector that can then be checked with `requiredPadding()` but that's its own PR. 

Short term, I think adding some mechanism to signal if you had to add padding or not is the right call

krzysz00 (approved):
Having been over this a bunch, the workspace stuff is future work because this PR has revealed an underlying problem with that code which is outside of the scope of this refactor

Namely
1. calculatePaddingKernelSize(), which is used during, among other things, determining whether we need a workspace for a WRW kernel, doesn't account for perf config and just looks at the heuristic list
2. So, if we're in heuristic mode, everything works fine, and this PR doesn't change any of the correctness there
3. But, if we've been given a perf config, there are the following cases
  - The perf config requires padding. Either cacluratePaddingKernelSize() correctly determines that padding is needed (because nothing on the list is padding-free) or it doesn't. In the former case, we don't call for a workspace. In the latter case, we call for a workspace and never use it.
  - The perf config doesn't require padding. Then, if the perf config is on our list, we correctly call for a workspace. If the config is **not** on the heuristic list, we may fail to request a workspace. If this occurs, then `--miopen-conv-to-gemm` will fail (or fall back to the non-atomic-add kernel, haven't checked which) because of the lack of a workspace.
  - That last bit is the bug
 3. This PR's change to what can be a padding kernel has no impact on whether the bug can happen

So, therefore, given that we've got successful testing (aka, a weekly run), this PR's approed

krzysz00:
@jerryyin I think you still need to take your hold off

jerryyin (approved):
Nothing past @krzysz00's comments. I'd like a ticket on padding + workspace reinspection.

manupak:
> Nothing past @krzysz00's comments. I'd like a ticket on padding + workspace reinspection.

https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/653

manupak:
http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-757/38/pipeline/

parameterSweeps seems successful now...

mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
```diff
@@ -167,15 +167,13 @@ void AffixTuningParameters::affixBackwardWeightUtilityKernels(
...
+        op->getAttrOfType<XdlopsGemmParamsAttr>(op.paramsAttrName());
     Optional<GemmContext> extraPadSizes =
-        calculatePaddingKernelSize(gemmSize, obtainConvDirection(op),
-                                   obtainConvDataType(op), populateParamsXDL);
+        calculatePadding(gemmParams.getKPerBlock(), gemmParams.getMPerBlock(),
+                         gemmParams.getNPerBlock(), gemmSize);
 
-    // For padding cases, gemmId must be 0.
-    if (extraPadSizes.hasValue()) {
-      assert(gemmId == 0);
```
manupak, line 178: I had to remove this to make a case : mlir/test/rocmlir-driver/auto_e2e/conv2d_f16_wrw_perf_config.mlir (in this PR to work). I encountered this case in a tuning run on the weekly which is only triggered when a padded perf-config case was given.

manupak (commented):
@krzysz00 I don't fully follow why this assertion was here.

mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
manupak: It is worth noting that this path was hardly taken before as it used to run calculatePaddingKernelSize that out right ignores perf-config as discussed before

manupak:
PR-CI passed here : http://rocmhead.amd.com:8080/job/MLIR/job/mlir/view/change-requests/job/PR-757/45/

Kicking off a weekly job : http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-757/46/pipeline

mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
krzysz00: This assertion is there to assert that we haven't added the zeroing utility kernel when there's no padding.

So this means you may need to fix the detection if whether there'll be padding or not to account for perf configs in order to get this PR to work.

I'd put the assert back and fix the underlying problem.

krzysz00 (changes requested):
I'm adding a hold so as to ensure the assert goes back in

mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
manupak: That does not make sense to me.
 
> This assertion is there to assert that we haven't added the zeroing utility kernel when there's no padding.

extraPadSizes.hasValue() reads as a scenario where there is padding as opposed to a scenario where there is no padding. if anything it is checking when there is padding, it seems to generate only a single kernel ? if not what does gemmId means here ?

krzysz00: gemmId is the ID of the kernel.
In backwards weight, the cases are this:
1. There  is padding: there is only one kernel, the gemmId is 0
2. There is no padding, the datatype is FP32: There are two kernels, one that zeroes the result buffer and a second one that does the commputation. Their gemmIDs are 0 and 1, respectively
3. There is no padding, the datatype is FP16: There are three kernels, one that zeroes the workspace, one that computes the convolution to the workspace, and one that copies the workspace to the output with truncation

The number of kernels that the system is generating must match those cases, and the assertion ensures that

krzysz00: (ok, yeah, I had it backwards: you need to only generate one kernel when there is padding)

The way I could see this assert tripping is that if you have a case where one of the heuristic parameter sets leads to a no-padding case (giving you a zeroing kernel) but the perf config leads to the padding case.

Therefore, the code to check if a backwards weight kernel needs padding (as used in Miir -> hasWorkspace()) needs to be modified to take the tuning parameters.

This might also involve finally writing the function that goes from a perf_config to a [Something]GemmParamsAttr

manupak: Thanks for enlightenment of what is going in this part of the code...

I managed to get it working with assert.
(See below -- I changed it to include a message)

manupak (commented):
Thanks! I got the assert back in.

Yes it seems like the discrepancy between hasWorkspace API and kernel generation seems to be causing the issue.

krzysz00 (commented):
Looks like there was some fun with the fact that kPerBlock is in units of kpack, causing much confusion everywhere. We really need to fix that.

I'll approve when we have passing weeklies.

manupak:
@krzysz00 after a significant amount of debugging I found out the weekly (tuning) is failing due to incorrect code in MIOpen.
It does not pass the perf_config when determining kernel count.

I had to do the following to get the current failures to pass in the MIOpen side. 

```
diff --git a/src/solver/conv_mlir_igemm_fwd_xdlops.cpp b/src/solver/conv_mlir_igemm_fwd_xdlops.cpp
index 6ea3335c2..5378aa8f0 100644
--- a/src/solver/conv_mlir_igemm_fwd_xdlops.cpp
+++ b/src/solver/conv_mlir_igemm_fwd_xdlops.cpp
@@ -127,7 +127,7 @@ bool PerformanceConvMlirIgemmXdlops::IsValid(const ConvolutionContext& ctx) cons
     if(*this == MlirHeuristicInitRequest())
         return true;
 
-    int kernel_count = MiirGetKernelCount(mlir::ConstructBuildOptions(ctx, true));
+    int kernel_count = MiirGetKernelCount(mlir::ConstructBuildOptions(ctx, *this, true));
     bool isValid     = false;
     for(int kernel_id = 0; kernel_id < kernel_count; ++kernel_id)
     {
diff --git a/src/solver/conv_mlir_igemm_wrw_xdlops.cpp b/src/solver/conv_mlir_igemm_wrw_xdlops.cpp
index cfaa58400..72c4bd519 100644
--- a/src/solver/conv_mlir_igemm_wrw_xdlops.cpp
+++ b/src/solver/conv_mlir_igemm_wrw_xdlops.cpp
@@ -86,7 +86,7 @@ ConvSolution ConvMlirIgemmWrWXdlops::GetSolution(const ConvolutionContext& ctx,
 {
 #if MIOPEN_USE_MLIR
     ConvSolution result;
-    int kernel_count = MiirGetKernelCount(mlir::ConstructBuildOptions(ctx, true));
+    int kernel_count = MiirGetKernelCount(mlir::ConstructBuildOptions(ctx, config, true));
 
     for(int kernel_id = 0; kernel_id < kernel_count; ++kernel_id)
     {
```

krzysz00:
Whoof

The fix for things that require MIOpen patches is usually to have this commit point at a branch of the MIOpen repo that contains the fix and then to file the patch as a PR against MIOpen.

The tricky thing is that we already have one of those PRs open (@zhanglx13's find_package branch), which means you'll need to file your PR against that branch instead of against `develop`.

At least we found the root cause here!

krzysz00:
(You'll also want to bump the API version in `Miir.h` on our end so you can `#ifdef` both versions of the API instead of just having a patch)

zhanglx13:
My [PR](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1673) is blocked by another MIOpen PR and we do not know when it can merge.
@manupak I think you can start from `rename-MLIR`, which is the MIOpen branch we use for testing, add your changes, push the new branch, and change the MIOpen branch on our jenkins to your new one. 
And later when our PRs are merged into their `develop` branch, we can change the MIOpen branch on jenkins to `develop`.

manupak:
Thanks for the pointers! @krzysz00 @zhanglx13!

> @manupak I think you can start from rename-MLIR, which is the MIOpen branch we use for testing, add your changes, push the new branch, and change the MIOpen branch on our jenkins to your new one.

This makes sense as a start. Thanks!

> (You'll also want to bump the API version in Miir.h on our end so you can #ifdef both versions of the API instead of just having a patch)

I could bump the API version but I dont follow the breaking change introduced here, at least from our end.
It occurs to me as MIOpen bug to me. 

I suppose the backward compatibility scenario is if MIOpen 5.5 wants to use with rocMLIR 5.4 ? 
In that case, I dont see a harm in passing the perf-config additionally to rocMLIR 5.4 where rocMLIR 5.4 would just out-right ignore perf-config in hasWorkspace APIs in conjunction with AffixParameters pass would just simply fail if there was padding -- Therefore I dont see a problem that requires us to use a ifdef.

jerryyin:
@manupak Nice catch! I think the reason why this happened can trace back to the decision to have MLIR undertake more of the kernel executions in wrw pass. So while we have made it possible for MLIR to execute more than 1 kernel, it doesn't occur to anyone that the requirement can change due to different tuning parameters.

In your branch, I'd suggest you make the same change to all MLIR solvers to avoid this from happening in the future:
 - ConvMlirIgemm<Fwd/Bwd/WrW>
 - ConvMlirIgemm<Fwd/Bwd/WrW>Xdlops

manupak:
Thanks @jerryyin! I made changes to everywhere MiirGetKernelCount is called to include that : https://github.com/ROCmSoftwarePlatform/MIOpen/commit/6ee35cdcf36df014391db2d7eb2efcc366dc9c60

manupak:
now closes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/653

manupak:
The last weekly run seems promising 🤞  : http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-757/54/pipeline/

It fails on something that I guess is unrelated: 

```
[2022-09-28T19:25:54.893Z] wrapper script does not seem to be touching the log file in /var/z1_miciadmin/workspace/MLIR_mlir_PR-757/MIOpen/build@tmp/durable-5eefefa8

[2022-09-28T19:25:54.893Z] (JENKINS-48300: if on an extremely laggy filesystem, consider -Dorg.jenkinsci.plugins.durabletask.BourneShellScript.HEARTBEAT_CHECK_INTERVAL=86400)

script returned exit code -1
```

Anyway, there were conflicts with the trunk now -- so kicked off a weekly (and at the mercy of the CI infra now ...)

krzysz00:
@manupak That error is the result of the script lagging out and spending some reasonably large amount of time without producing output.

This probably means that the node got its GPU into a bad state and needs rebooted - I don't remember if the weekly tuning job does that automatically or not.

Also, for future CI runs, since we have parameter sweeps working, I _think_ there might be a parameter you can uncheck to skip them so we don't need to keep rerunning them

manupak:
Ah it hit the same error again :(. : http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-757/55/pipeline

I think only way to run the  Tune MLIR kernels stage is by using weekly setting on (we might want to modularize that in future)

https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/ac36fec6d6dd8a64d05b7770fffc497bb4576baf/mlir/utils/jenkins/Jenkinsfile#L405-L410.

Is there anything else could be done here other than re-triggering ?

zhanglx13:
If some node is always picked and failing, like lockhart14, maybe you can manually pick another node by tweaking the `label` option in the agent.

manupak:
Thanks @zhanglx13! It seemed like both cases picked lockhart14. 
I am not very familiar on doing "tweaking the label option in the agent". Where should I do this ? 

Also, what other options are there ? would any lockhart node would do?

zhanglx13:
On Jenkins, you can replay a build by pressing the `replay` button on the left panel. It leads you to a page that you can modify the jenkinsfile for the replay. 
For each stage, there is an `agent` section
https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/ac36fec6d6dd8a64d05b7770fffc497bb4576baf/mlir/utils/jenkins/Jenkinsfile#L418-L425
Inside, there is a label command in the docker section. For now, it is set according to the `${ARCH}` value. You can either reset the label when `${ARCH} == gfx90a`, or you can change the `${ARCH}` values to something like
```
'gfx906', 'gfx908', 'gfx90a && MI250X'
```
This will rule out lockhart14 when tuning for gfx90a nodes.

You can check nodes and their labels or the other way around on Jenkins: http://rocmhead.amd.com:8080/label/gfx90a/

manupak:
Yey! Weekly CI is green finally!

(Noted with thanks @zhanglx13, now I know and I ll keep that in mind for next time)

krzysz00 (approved):
LGTM, feel free to land this now

manupak:
Thanks @krzysz00  @jerryyin @zhanglx13!

mlir/test/rocmlir-driver/auto_e2e/conv2d_f32_wrw_perf_config.mlir
```diff
@@ -0,0 +1,3 @@
+// RUN: rocmlir-gen --operation conv2d_bwd_weight --perf_config 64,64,2,64,64,2,1,1 -t f32 -x2 --fil_layout kcyx --in_layout nchw --out_layout nkhw --batchsize 512 --in_channels 512 --in_h 1 --in_w 1 --out_channels 512 --fil_h 1 --fil_w 1 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h 0 --padding_w 0 %pv %random_data -x2 | rocmlir-driver -c | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CHECK_F32_WRW_PERF_CFG
```
zhanglx13, line 1: @manupak Do you expect this test to be xdlops-only? 
- If so, it needs to be put in a directory, like the deleted PaddingXDLOPS, with a lit.local.cfg to make sure this test runs only when xdlops is enabled, i.e. Jenkins picked gfx908 or gfx90a.
- If this test needs to be checked for both xdlops and non-xdlops, `-x2` needs to be replaced by `%xdlops`.

Just want to check which is the case. I can fix them in my current PR.

krzysz00: Yeah, good catch @zhanglx13. Making this %xdlops or moving it to a relevant directory needs to happen today. I'm willing to approve a hotfix PR and to cheat it through CI so we don't have broken nightlies/weeklies

zhanglx13: Sure.
Please correct me if I am wrong, since I am not familiar with `perf_config`. From the number of values provided to `perf_config`, it seems that this is a xdlops-only test, is that true @krzysz00?

krzysz00: Yep, it is

manupak: My bad.. good catch! Thanks guys!

PR #758, opened 2022-09-06
[DO NOT SQUASH] Set up and use an architecture database, add extra feature flags

krzysz00:
1. Create an AmdArchInfo struct and a lookupArchInfo() function that
returns the AmdArchInfo for a given chipset
2. Define the "dot", "wave32", and "atomic_add" features, we may need
them in the future
3. Populate the database with cases for the architectures the backend
currently knows about.
4. Update the convolution generator and TosaToMIOpen to get their
default features by querying the database
5. Update tests

This implements, as far as I'm concerned,
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/607


krzysz00:
Feel free to review but please don't merge yet - I plan to pull in https://reviews.llvm.org/D133374 onto this branch to simplify things since it just landed upstream.

mlir/lib/Dialect/MIOpen/Generator/AmdArchDb.cpp
```diff
@@ -0,0 +1,39 @@
...
+AmdArchInfo mlir::miopen::lookupArchInfo(StringRef chip) {
+  return llvm::StringSwitch<AmdArchInfo>(chip)
+      .Cases("gfx908", "gfx90a", "gfx940", cdnaFeatures)
+      // gfx906 has the dot product instructions, uniquely
+      .Case("gfx906", AmdArchInfo(GemmFeatures::dot))
+      .StartsWith("gfx9", gcnFeatures)
+      .Cases("gfx1011", "gfx1013", rdnaNoDotFeatures)
+      .Cases("gfx1010", "gfx1012", rdnaFeatures)
+      // All gfx103x are the same for us
+      .StartsWith("gfx103", rdnaFeatures)
```
sjw36: Still not a fan of this string matching.. for gfx103* it will strcmp 10 times for every lookup.

Perhaps we should have a SmallDenseMap that can contains a "gfx<major>" that can be used when an exact match doesn't exist.

mlir/test/fusion/tosa-tp2-highlevel-fused.mlir
```diff
@@ -1,7 +1,7 @@
 // RUN: mlir-miopen-driver --host-pipeline highlevel %s | FileCheck %s
 
-// CHECK: miopen.conv2d(%{{.*}}, %{{.*}}, %{{.*}}) features =  none {arch = "gfx906", dilations = [1 : i32, 1 : i32], filter_layout = ["k", "c", "y", "x", "g"], input_layout = ["ni", "hi", "wi", "ci", "gi"], numCu = 64 : i32, output_layout = ["no", "ho", "wo", "ko", "go"], padding = [1 : i32, 1 : i32, 1 : i32, 1 : i32], strides = [1 : i32, 1 : i32]} : memref<64x128x3x3x1xf32>, memref<256x28x28x128x1xf32>, memref<256x28x28x64x1xf32>
-    
+// CHECK: miopen.conv2d(%{{.*}}, %{{.*}}, %{{.*}}) features =  none {arch = "gfx900", dilations = [1 : i32, 1 : i32], filter_layout = ["k", "c", "y", "x", "g"], input_layout = ["ni", "hi", "wi", "ci", "gi"], numCu = 64 : i32, output_layout = ["no", "ho", "wo", "ko", "go"], padding = [1 : i32, 1 : i32, 1 : i32, 1 : i32], strides = [1 : i32, 1 : i32]} : memref<64x128x3x3x1xf32>, memref<256x28x28x128x1xf32>, memref<256x28x28x64x1xf32>
```
sjw36: BTW, I would prefer not to see 'features = none', just like I don't see 'attributes {}'. Can you hide it when none.

mlir/test/mlir-miopen-lib/populate_bww.mlir
```diff
@@ -48,6 +48,6 @@
...
-// DRIVER3: miopen.conv2d_bwd_weight(%arg0, %arg1, %arg2, %arg3) features = xdlops {arch = "gfx908", dilations = [1 : i32, 1 : i32], filter_layout = ["g", "k", "c", "y", "x"], gemm_id = 2 : i32, input_layout = ["ni", "gi", "ci", "hi", "wi"], numCu = 120 : i32, output_layout = ["no", "go", "ko", "ho", "wo"], padding = [0 : i32, 0 : i32, 0 : i32, 0 : i32], strides = [1 : i32, 1 : i32]} : memref<1x1024x1024x1x1xf16>, memref<64x1x1024x14x14xf16>, memref<64x1x1024x14x14xf16>, memref<1x1024x1024x1x1xf32>
+// DRIVER3: miopen.conv2d_bwd_weight(%arg0, %arg1, %arg2, %arg3) features = xdlops|dot|atomic_add {arch = "gfx908", dilations = [1 : i32, 1 : i32], filter_layout = ["g", "k", "c", "y", "x"], gemm_id = 0 : i32, input_layout = ["ni", "gi", "ci", "hi", "wi"], numCu = 120 : i32, output_layout = ["no", "go", "ko", "ho", "wo"], padding = [0 : i32, 0 : i32, 0 : i32, 0 : i32], strides = [1 : i32, 1 : i32]} : memref<1x1024x1024x1x1xf16>, memref<64x1x1024x14x14xf16>, memref<64x1x1024x14x14xf16>, memref<1x1024x1024x1x1xf32>
```
sjw36: Also, not sure 'xdlops' is a good name, how about 'mfma' like 'dot'

sjw36 (changes requested):
I would like to see a singleton approach that creates a DB (StringMap) for lookup. Maybe even buildable at compile time with constexpr.

mlir/lib/Dialect/MIOpen/Generator/AmdArchDb.cpp
krzysz00: String matches hang around the compiler a lot, I would argue this isn't particularly unreasonable. And I just looked at the disassembly - this does not at all go to anything resembling a linear series of `strcmp` calls. Lots of `xor` and the like.

mlir/test/fusion/tosa-tp2-highlevel-fused.mlir
krzysz00: I don't think so, given that `features` is a required attribute and that `none` is how you print out its value when it's 0. And if we weren't doing a custom thing like that we'd have things like `features = 1 : i32` for having `mfma` on.

mlir/test/mlir-miopen-lib/populate_bww.mlir
krzysz00: Yeah, good point, I can tweak it.

mlir/lib/Dialect/MIOpen/Generator/AmdArchDb.cpp
krzysz00: That is, I don't think the complexity of trying to write out this out even more explicitly is worth it, especially when `-O3` is very good at its job.

sjw36: Sure, not surprised they have a full lexer-like parser/decoder. But this list will grow over time and the complexity will grow with it.

Maybe should use **major**-**minor** lookup, and where no **minor** matches, using the default **major**.

sjw36: Check latest comments on the ticket.

krzysz00: (Ok, I'll give you checking major and minor version separately to be safe)

krzysz00:
@zhanglx13 and @sjw36 Since there's now some ongoing work on dot products, I'd like to see this landed, even if it's not the full-fledged database Simon's looking for - it'll serve our purposes for knowing if the dot product instructions are there or not and having them enabled by default.

krzysz00:
In updates, it turns out that having all the fusion tests setting `arch = "gfx908"` was a bad call in the presence of default feature lists, because then the tests would fail if run on a gfx1030 because it'd generate an xdlops kernel.

So now much of the fusion tests use `arch = ""` so as to generate aggressively generic binaries

krzysz00:
Also, @sjw36, there's nothing about this PR that prevents us from evolving to your preferred approach later.

sjw36 (approved):
Yes, its good for now.

PR #759, opened 2022-09-07
Hot fix to the broadcast fusion

jungpark-mlir:
- Fix missing broadcast in mapping, now code actually matches to the new assertion
- Make e2e test to verify the result


sjw36 (approved):
Looks good.

sjw36:
BTW, there are many compile warnings in CopyOpt.cpp from the last commit.

jungpark-mlir:
@sjw36 
> BTW, there are many compile warnings in CopyOpt.cpp from the last commit.

Strange, I can see this one only, 
../mlir/lib/Dialect/MIOpen/Transforms/CopyOpt.cpp:103:14: warning: suggest parentheses around assignment used as truth value [-Wparentheses]
       } else if (csop = dyn_cast<memref::CollapseShapeOp>(use.getOwner())) {

PR #760, opened 2022-09-07
Improve non-tuned param ordering

manupak:
This commit improves the init param ordering
to be sorted based on padding amount.
i.e. lowest padding is preffered.

When in equal padding cases, the original
ordering will be respected.

Doing so, exposed some params should be
dynamically filtered out according to
XdlopsCodeSelection.

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/IR/XdlopsCodeSelection.h
+++ b/mlir/include/mlir/Dialect/Rock/IR/XdlopsCodeSelection.h
@@ -70,14 +70,13 @@ struct XdlopsCodeSelection {
   }
 
   static XdlopsCodeSelection get(mlir::Type dataType, int64_t MPerWave,
-                                 int64_t NPerWave, OpBuilder &b) {
+                                 int64_t NPerWave) {
     using amdgpu::MFMAPermB;
     constexpr int64_t waveSize = 64;
     // Determine which XDLOPS be used.
     int64_t mfmaNonKDim = 0, MRepeats = 0, NRepeats = 0, k = 0, blocksMfma = 0;
     SmallVector<MFMAParams, 2> imms;
-
-    if (dataType == b.getF32Type()) {
+    if (dataType.isF32()) {
       if (MPerWave == 128 && NPerWave == 64) {
         // instr = amdgpu::MFMAInstr::f32_32x32x1f32;
         mfmaNonKDim = 32;
@@ -189,7 +188,7 @@ struct XdlopsCodeSelection {
         dumpUnsupported(dataType, MPerWave, NPerWave);
         llvm_unreachable("Can't meaningfully continue xdlop generation");
       }
-    } else if (dataType == b.getF16Type()) {
+    } else if (dataType.isF16()) {
       if (MPerWave == 128 && NPerWave == 64) {
         // instr = amdgpu::MFMAInstr::f32_32x32x4f16;
         mfmaNonKDim = 32;
@@ -301,7 +300,7 @@ struct XdlopsCodeSelection {
         dumpUnsupported(dataType, MPerWave, NPerWave);
         llvm_unreachable("Can't meaningfully continue xdlop generation");
       }
-    } else if (dataType == b.getBF16Type()) {
+    } else if (dataType.isBF16()) {
       if (MPerWave == 128 && NPerWave == 64) {
         // instr = amdgpu::MFMAInstr::f32_32x32x2bf16;
         mfmaNonKDim = 32;
@@ -413,7 +412,7 @@ struct XdlopsCodeSelection {
         dumpUnsupported(dataType, MPerWave, NPerWave);
         llvm_unreachable("Can't meaningfully continue xdlop generation");
       }
-    } else if (dataType == b.getIntegerType(8)) {
+    } else if (dataType.isInteger(8)) {
       if (MPerWave == 32 && NPerWave == 32) {
         // instr = amdgpu::MFMAInstr::i32_32x32x8i8;
         mfmaNonKDim = 32;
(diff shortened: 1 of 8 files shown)


manupak:
Putting this to draft as this build upon https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/757 and blocked on merging the above.

manupak:
For the exclusive changes in this PR, please look at 355126df37bd98a10229fc64b33c50524a7a49fd

krzysz00:
Worth mentioning, you can set the base branch to be the other PR - that'll work fine

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
```diff
@@ -134,7 +134,77 @@ struct DerivedBlockGemmParams {
...
+
+  // Interface function to calculate padding amount of a given param set
+  virtual int64_t calculatePaddingAmount(const InitParamType &params,
+                                         const GemmSize &gemmSize) const = 0;
+
+  // This function will order the initParams used in a non-tuning flow to
+  // prioritize params that does not require padding to come first while
+  // otherwise maintaining the provided order.
+  std::vector<InitParamType>
+  orderInitParams(const std::vector<InitParamType> &initParams,
```
krzysz00: `const std::vector<T> &` is spelled `ArrayRef<T>`

mlir/include/mlir/Dialect/MIOpen/XdlopsCodeSelection.h
```diff
@@ -189,7 +188,7 @@ struct XdlopsCodeSelection {
         dumpUnsupported(dataType, MPerWave, NPerWave);
         llvm_unreachable("Can't meaningfully continue xdlop generation");
       }
-    } else if (dataType == b.getF16Type()) {
+    } else if (dataType.isF16()) {
```
krzysz00: Nice catches

mlir/lib/Dialect/MIOpen/Tuning/GridwiseGemmParams.cpp
```diff
@@ -1283,6 +1139,22 @@ LogicalResult PopulateParamsXDL::isValidGemm(const InitParamsXDL &param,
   return success();
 }
 
+int64_t
+PopulateParamsXDL::calculatePaddingAmount(const InitParamsXDL &params,
+                                          const GemmSize &gemmSize) const {
+  int64_t padM = params.gemmMPerBlock - gemmSize.gemmM % params.gemmMPerBlock;
```
krzysz00: That doesn't look right - I think you want padM to be 0 in the non-padding case. You probably want `mod_1_to_n` from the math utilities header here

manupak: Yes, you are right -- reused the existing padding calculating utils there and refactored it a bit.

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
manupak: Done

manupak:
> Worth mentioning, you can set the base branch to be the other PR - that'll work fine

Other PR branch is in my fork right ? this is a missing feature request for GH for a while now :sigh.
https://github.com/orgs/community/discussions/4477#discussioncomment-1039627

krzysz00 (commented):
Code-wise, looks excellent!

Can we get perf comparisons before merging?

manupak:
Sure, How would you generally do that (other than running a random assortment)? Any pointers are appreciated

krzysz00:
The way I usually do it is to run (after doing a `ninja check-rocmlir-build-only ci-performance-scripts`):

```
./bin/MIOpenDriver.py -b -o=some-name.csv
echo "description" >perf-run-date
[cd to a checkout of upstream]
ninja check-rocmlir-build-only ci-performance-scripts
./bin/MIOpenDriver.py -b -o=upstream.csv
echo "Upstream" >perf-run-date
[cd back]
python3 ./bin/perfRegressionReport.py [path-to-upstream].csv [path-to-here].csv output-file.html
```
then copy the file off, take a look at it, and post it here

krzysz00:
(You can, alternatively, move the results from your perf test along with the perf-run-date to another directory, then `git checkout miopen-dialect`, recompile, and run the script again)

manupak:
Noted with thanks! I ll have a go tomorrow

manupak:
https://gist.github.com/manupak/a1e679b89b1e05fe3db0a74b4879a266

Overall there is an increase there are few cases it drops..
It is worth noting the differences are in cases where they are believed to be generating a padded gemm.

manupak:
ok some worse numbers does not seem to reproduce now and running multiple measurements to see...

manupak:
I did three runs for the PR : 

1.) https://gist.github.com/manupak/887bcf5a3727f22ccfacff85b20603ac
2.) https://gist.github.com/manupak/1dcf61e381f472100f443bb2f8a52288
3.) https://gist.github.com/manupak/4156c2e8b9d09228d744a71ac1aee8b2

Generally looking at the changes only a single worse config stood out -- others more or less seemed like glitches.

| Direction | DataType | Layout | C |  H  |  W  |  K | Y | X | StrideH | StrideW | PaddingH | PaddingW | Unnamed: 0_new | MLIR TFlops (new) | Unnamed: 0_old | MLIR TFlops (old) | Current/Previous |
|:---------:|:--------:|:------:|:-:|:---:|:---:|:--:|:-:|:-:|:-------:|:-------:|:--------:|:--------:|:--------------:|:-----------------:|:--------------:|:-----------------:|:----------------:|
| fwd       | f16      | nhwc   | 3 | 230 | 230 | 64 | 7 | 7 |    2    |    2    |     0    |     0    |       68       |       11.75       |       68       |       21.11       |       0.56       |

It is interesting that it is also layout specific as the nchw layout seemed fine with config.

krzysz00:
Having looked, those are certainly substantial wins, and the fact that we now have problems with one fp16 config isn't anywhere near that big a deal in my book.

So congrats on your perf win!

krzysz00 (approved):

manupak:
Thanks for reviews @krzysz00  and working through with me with this.

PR #762, opened 2022-09-09
Define a targets attribute to save on hipGetDevice() calls

krzysz00:
- Rename -target in the various miopen tools to -chip and reuse -target for the chip name along with features
- Use rocm_agent_enumerator, with or without -name, to get the chip name with included features, and use that information to populate the name in the test harness
- Update the performance and parameter sweep scripts to pass in -target as well

Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/648 and
fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/643


zhanglx13:
I am trying to understand what is going on in this PR and here are my thoughts:
1. `target` is too broad and it includes everything about the underlying GPU, such as `triple`, `feature`, and `id`.
    - If you want to refer to `chip:feature`, you can use `arch`, as in other tools and `hipGetDeviceProperties().gcnArchName`.
 However, the `arch` option refers to slightly different things in different tools (may need some cleanup in the future). E.g.
      - `miopen-gen` has an option `arch` that is used to represent `chip`. But in `populateDefaults()`, `triple` is added to the string
      - `miopen-gen` has another option `conv-config`, in which `--arch` is used to set `triple:chip:feature`
      - `mlir-miopen-lib-test` has an option `args`, in which `--arch` is used to set `chip`.
    - Or you can use `target-id`, defined in the [AMDGPU Backend](https://llvm.org/docs/AMDGPUUsage.html#target-id), which refers to `chip:features` exactly.
2. Can we get rid of calling `hipGetDeviceProperties()` in `mlir-rocm-runner` in the same way as in `xmir-runner.cpp` [code](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/1252d5084784cb5c66c87c4770fc9e70b1743720/mlir/tools/xmir-runner/xmir-runner.cpp#L55)?
3. I think [MLIR#648](https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/648) is to remove the `target` option [here](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/1252d5084784cb5c66c87c4770fc9e70b1743720/mlir/tools/mlir-miopen-driver/mlir-miopen-driver.cpp#L84)

krzysz00:
@zhanglx13 Good point re `arch`  or `target-id`, I'll make a new PR that works with that
Although we have some places in our code that use `arch` = `chip`

One question I have for @sjw36, is the xmir `-targets` meant to be a list of `chip` or `chip:features` things?

sjw36:
> One question I have for @sjw36, is the xmir `-targets` meant to be a list of `chip` or `chip:features` things?

Seems like we will need the full spec since I guess compilation/execution can fail without the right features.

krzysz00:
Closing as mostly supersceded by #781 - the followup PR to clean up options to `rocmlir-driver` and `mlir-rocm-runner` will land soon

PR #763, opened 2022-09-09
[MIGraphX] Added extra include dirs to the exported library

zhanglx13:
Since MLIR exports headers from upstream (llvm-project), the directory for these headers should also be added to the exported library. Therefore, downstream users can include these headers as if they are from the llvm-project directly.

https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/640


mlir/tools/mlir-miopen-lib/CMakeLists.txt
```diff
@@ -86,53 +86,63 @@ if(BUILD_FAT_LIBROCKCOMPILER)
...
+
+    # Install lib${lib_name}.a to ${CMAKE_INSTALL_PREFIX}/lib/${package_name}/
     # and group it into the ${export-set}
     # The INTERFACE_INCLUDE_DIRECTORIES will be set as the property of the exported library
     # so that when comsumer libraries (MIOpen) link in lib${LIBRARY_NAME}, ${INCLUDE_DIR} is
     # automatically propagated to the consumer library's include_dirs
-    install(TARGETS ${target_name}
+    set(EXTRA_INCLUDES "")
+    if(ARG_EXTRA_INCLUDES)
+      set(EXTRA_INCLUDES ${ARG_EXTRA_INCLUDES})
```
jerryyin, line 139: Question: Is it the core change here, that MIGraphX needs to add `-DARG_EXTRA_INCLUDES`?

krzysz00: Might that be set by default?

krzysz00: ... ah, that's for allowing the user to specify even more includes, I think

zhanglx13: Yes. MIGraphX needs to add more dir into its includes. It is the same as
`target_include_directories(migraphx_gpu ARG_EXTRA_INCLUDES)`

zhanglx13: @krzysz00 By default, the `include/<packagename>` dir is added. Other dirs need to be explicitly specified.

krzysz00: Shouldn't those always be installed as part of our package?

That is, wouldn't it make more sense to update `mlir-miopen-lib` to include those migraphx headers unconditionally?

jerryyin: > Shouldn't those always be installed as part of our package?

I think I can see why this decision make sense: MIGraphX and MIOpen required headers are roughly independent. By adding this `extra includes` flag we excluded the migraphX header to be installed into systems that were targeting for MIOpen.

jerryyin (approved):

PR #764, opened 2022-09-13
[NFC] Use vector.transfer_read/transfer_write for in_bounds_load/store

krzysz00:
Since these ops now support using an in_bounds attribute to get rid of bounds checks, we might as well use them.


zhanglx13 (approved):
Just one question: why do you need to change the memref in the test from 2x2 to 4x4?

krzysz00:
So that we don't have any statically out of bounds reads

PR #765, opened 2022-09-13
Vectorize LDS writeback loop in non-xdlops

krzysz00:
(I tested perf, it's pretty much flat, maybe a few kernels got a boost but that could be measurement error)


jerryyin (approved):
LGTM, please help keep an eye on nightly/weekly as we are close to branching.

jungpark-mlir (approved):
Looks good.
I was also curious about the performance and it looks the backend finally schedules and vectorizes them pretty well.

krzysz00:
Oh, yeah, this is pretty much [NFC], by all accounts

krzysz00:
This is now `skip-ci` until #764 gets landed

PR #766, opened 2022-09-13
Fix build directories in MIOpenDriver.py

yiqian1:
Ignore the MLIR build directory when running the benchmarks for MIOpen as MLIR is not built.
In find_mlir_build_dir(), look for the path for miopen_gen instead of miopen_opt as miopen_opt may not be built here.

### DIFF
--- a/mlir/utils/jenkins/MIOpenDriver.py
+++ b/mlir/utils/jenkins/MIOpenDriver.py
@@ -47,17 +47,18 @@ def find_mlir_build_dir() -> str:
     Finds mlir build dir searching either WORKSPACE dir
     or home dir
     """
-    miopen_opt_path = None
+    miopen_gen_path = None
     candidate_paths = [
         # if the script is run from build dir
-        Path('./bin/miopen-opt'),
+        Path('./bin/miopen-gen'),
         # if the script is run from source
-        Path(__file__).parent.parent.parent.parent / 'build' / 'bin' / 'miopen-opt'
+        Path(__file__).parent.parent.parent.parent / 'build' / 'bin' / 'miopen-gen'
     ]
     for candidate_path in candidate_paths:
         if candidate_path.exists():
-            miopen_opt_path = candidate_path
-    if not miopen_opt_path:
+            miopen_gen_path = candidate_path
+    
+    if not miopen_gen_path:
         try:
             # Prioritize the search in the current repo first.
             search_root = str(subprocess.check_output(['git', 'rev-parse', '--show-toplevel']).decode().strip())
@@ -65,11 +66,15 @@ def find_mlir_build_dir() -> str:
             # Else look in the home or WORKSPACE directory
             search_root = os.environ.get('WORKSPACE', str(Path.home()))
             assert search_root, "Cant find WORKSPACE env arg or home directory"
-        miopen_opt_path = glob.glob(search_root + '/**/bin/miopen-opt', recursive=True)
-        assert len(miopen_opt_path) == 1, "Multiple paths found to contain */bin/miopen-opt"
-        miopen_opt_path = miopen_opt_path[0]
+        
+        miopen_gen_path = glob.glob(search_root + '/**/bin/miopen-gen', recursive=True)
+        if len(miopen_gen_path) == 0:
+            # MLIR miopen_gen not available
+            return None
+        assert len(miopen_gen_path) == 1, "Multiple paths found to contain */bin/miopen-gen"
+        miopen_gen_path = miopen_gen_path[0]
 
-    build_dir = Path(miopen_opt_path).parent.parent
+    build_dir = Path(miopen_gen_path).parent.parent
     return str(build_dir)
 
 
@@ -97,7 +102,7 @@ def find_miopen_build_dir() -> str:
         if len(miopen_driver_path) == 0:
             # MIOpen driver not available
             return None
-        assert len(miopen_driver_path) == 1, "Multiple paths found to contain */bin/miopen-opt"
+        assert len(miopen_driver_path) == 1, "Multiple paths found to contain */bin/MIOpenDriver"
         miopen_driver_path = miopen_driver_path[0]
 
     miopen_build_dir = Path(miopen_driver_path).parent.parent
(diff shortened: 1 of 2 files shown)


mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -30,7 +30,7 @@
 # Compiled regexp object used for extracting elapsed time from MIOpenDriver's output
 ELAPSED_TIME_RE = re.compile(r"Elapsed: (.*)ms")
 
-class Paths(NamedTuple):
```
krzysz00, line 33: I think the NamedTuple usage makes sense here, why remove it?

mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -97,35 +102,40 @@ def find_miopen_build_dir() -> str:
...
+        paths.libconv_validation_wrappers_path = str((mlir_lib_dir / 'libconv-validation-wrappers.so').resolve())
+        paths.libmlir_runtime_utils_path = str((llvm_lib_dir / 'libmlir_runner_utils.so').resolve())
 
     miopen_driver_path = None
     if miopen_build_dir_path:
         miopen_driver_bin_dir = Path(miopen_build_dir_path) / 'bin'
         miopen_driver_path = str((miopen_driver_bin_dir / 'MIOpenDriver').resolve())
+        paths.miopen_driver_path = miopen_driver_path
 
-    return Paths(
```
krzysz00, line 119: I'd keep this, but instead use some `... if mlir_bin_dir else None` in here

krzysz00 (commented):
I'm not sure we needed to drop `NamedTuple` here. @manupak for more thoughts

mlir/utils/jenkins/MIOpenDriver.py
yiqian1: Restored NamedTuple.

mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -97,35 +102,38 @@ def find_miopen_build_dir() -> str:
...
+    return Paths(miopen_gen_path = mlir_bin_dir + '/miopen-gen',
+        mlir_miopen_driver_path = mlir_bin_dir + '/mlir-miopen-driver',
         miopen_driver_path = miopen_driver_path,
-        rocm_runner_path = str((mlir_bin_dir / 'mlir-rocm-runner').resolve()),
-        libmlir_rocm_runtime_path =  str((llvm_lib_dir / 'libmlir_rocm_runtime.so').resolve()),
-        libconv_validation_wrappers_path = str((mlir_lib_dir / 'libconv-validation-wrappers.so').resolve()),
-        libmlir_runtime_utils_path = str((llvm_lib_dir / 'libmlir_runner_utils.so').resolve()),
-        configuration_file_path = str((configuration_dir / 'resnet50-miopen-configs').resolve()),
-    )
+        rocm_runner_path = mlir_bin_dir + '/mlir-rocm-runner',
```
krzysz00: It seems to me like the nice thing to have here would be the function
```
def append_maybe_path(p: Optional[Path], subpart: str) -> Optional[str]:
  if path:
    return str((Path(p) / subpart).resolve())
  else:
    return None
```
and so we don't have paths like `"/libmlir_rocm_runtime.so"` running around the `Paths` object

krzysz00 (commented):
I've got another thought on how to do this

mlir/utils/jenkins/MIOpenDriver.py
manupak: Agreed with the sentiment here and I'd also support having None rather than invalid paths running around.
I have some thoughts of some other structure that would suit the nature where mlir builds could also be optional as evident by this PR.

manupak (commented):
Thanks @yiqian1 ! Now I know that MLIR builds could also be optional.

With that regard my suggestion would be to make the Paths data structure a bit composable as follows by improving namedtuple to be a data class to support Optionals with defaults: 

```
# Here I am open to not needing this at this point of time as it is just one path
@dataclass
class MIOpenPaths:
    miopen_driver_path : str
 
@dataclass
class MLIRPaths:
    mlir_miopen_driver_path: str
    rocm_runner_path : str
    libmlir_rocm_runtime_path : str
    libconv_validation_wrappers_path : str
    libmlir_runtime_utils_path : str

@dataclass
class Paths:
    mlir_paths: Optional[MLIRPaths] = None
    # We could omit the data structure for now as it only contains a single path
    # and use the miopen driver path straight away
    miopen_paths: Optional[MIOpenPaths] = None 
    configuration_file_path : str
```

With the above changes, I think we can cleanly construct the Paths object based on the presence of the MLIR and/or MIOpen.

WDYT ?

yiqian1:
> Thanks @yiqian1 ! Now I know that MLIR builds could also be optional.
> 
> With that regard my suggestion would be to make the Paths data structure a bit composable as follows by improving namedtuple to be a data class to support Optionals with defaults:
> 
> ```
> # Here I am open to not needing this at this point of time as it is just one path
> @dataclass
> class MIOpenPaths:
>     miopen_driver_path : str
>  
> @dataclass
> class MLIRPaths:
>     mlir_miopen_driver_path: str
>     rocm_runner_path : str
>     libmlir_rocm_runtime_path : str
>     libconv_validation_wrappers_path : str
>     libmlir_runtime_utils_path : str
> 
> @dataclass
> class Paths:
>     mlir_paths: Optional[MLIRPaths] = None
>     # We could omit the data structure for now as it only contains a single path
>     # and use the miopen driver path straight away
>     miopen_paths: Optional[MIOpenPaths] = None 
>     configuration_file_path : str
> ```
> 
> With the above changes, I think we can cleanly construct the Paths object based on the presence of the MLIR and/or MIOpen.
> 
> WDYT ?

Nice. I'll make the change and try again.

mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -47,29 +47,34 @@ def find_mlir_build_dir() -> str:
...
             assert search_root, "Cant find WORKSPACE env arg or home directory"
-        miopen_opt_path = glob.glob(search_root + '/**/bin/miopen-opt', recursive=True)
-        assert len(miopen_opt_path) == 1, "Multiple paths found to contain */bin/miopen-opt"
-        miopen_opt_path = miopen_opt_path[0]
+        
+        miopen_gen_path = glob.glob(search_root + '/**/bin/miopen-gen', recursive=True)
+        if len(miopen_gen_path) == 0:
+            # MLIR miopen_gen not available
+            return None
+        assert len(miopen_gen_path) == 1, "Multiple paths found to contain */bin/miopen-gen"
```
manupak, line 80: To fix nightlies, instead of the assert (my bad!) lets also return None here as well -- so that the consumer can complain if mlir build is expected rather than prematurely asserting here.

manupak (commented):
I think this could fix the nightlies

krzysz00 (commented):
I think you'll need to update `parameterSweeps.py` as well for this

But this works for me pending Manupa's OK and that you've tested it

mlir/utils/jenkins/MIOpenDriver.py
yiqian1: It returns None when miopen_gen_path is not found. The assert is testing if we got multiple paths containing miopen-gen.

manupak: Ack

manupak (approved):
Thanks! LGTM as well!

PR #768, opened 2022-09-14
Replacing tuning decided vectorization into transformMap based one

jerryyin:
**Note: Earliest merge after ROCm 5.4 branch**

This PR includes the following:
 - In Gridwisegemm -> Blockwisegemm lowering, swapped the vectorization decision from tuning parameter derived ones to transformMap based one: `bestVectorization()`
   - Plumb the rest of non-coordinate transformed code just to make them work, eventually will rewrite and get rid of them
 - Removed a significant amount of tuning parameter derive process
   - Get rid of the following data types:  `DerivedParams`, `DerivedOutParams`

I believe part of the reason MIOpen do the vectorization derivation is due to their requirement of cannot have compile failure. Therefore they need to manually derive the parameter as much as possible, otherwise templatized computation may not be able to handle outliers. However, we are more flexible in terms that we can do additional steps of lowering to tell whether certain configs are applicable. This gives me more confidence to eventually remove affix tuning parameter pass.


mlir/lib/Dialect/MIOpen/Tuning/GridwiseGemmParams.cpp
```diff
@@ -884,48 +434,6 @@ LogicalResult PopulateParamsXDL::getKBlocks(ConvolutionContext &ctx,
...
-      math_util::integer_least_multiple(param.gemmMPerBlock, max_lds_align);
-  const auto b_block_space =
-      param.gemmKPerBlock *
-      math_util::integer_least_multiple(param.gemmNPerBlock, max_lds_align);
-
-  ldsSize = (a_block_space + b_block_space) * sizeof(float);
-
-  if (ldsSize > 64 * 1024) {
-    return failure();
-  }
```
jerryyin: Note, previously the assumption was:
 - we assume lds size must be counted as fp32 type
 - The upper bound of the size was statically computed as 64 * 1024

For now I'm bringing the same logic over such that this PR try its best to maintain equivalent coverage. I'd file an additional ticket to document/schedule this such that this is data type and device dependent.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -613,6 +614,12 @@ struct GridwiseGemmRewritePattern : public OpRewritePattern<GridwiseGemmOp> {
     LLVM_DEBUG(llvm::dbgs() << "a_block_space: " << a_block_space << "\n");
     LLVM_DEBUG(llvm::dbgs() << "b_block_space: " << b_block_space << "\n");
     LLVM_DEBUG(llvm::dbgs() << "double_block_space: " << block_space << "\n\n");
+
+    // TODO: adjust for data type and device
+    if (block_space * sizeof(float) > 64 * 1024)
```
jerryyin: Migrated over here.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1064,6 +1075,12 @@ struct GridwiseGemmV2RewritePattern
     LLVM_DEBUG(llvm::dbgs() << "b_block_space: " << b_block_space << "\n");
     LLVM_DEBUG(llvm::dbgs()
                << "total_block_space: " << total_block_space << "\n\n");
+
+    // TODO: adjust for data type and device
+    if (total_block_space * sizeof(float) > 64 * 1024)
```
jerryyin: Migrated over here.

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
```diff
@@ -148,45 +130,27 @@ class PopulateParams {
...
+                                DerivedBlockGemmParams &blockGemmDerivedParam,
+                                uint32_t &gridSize);
 
   LogicalResult populatePaddingKernelDerived(
       ConvolutionContext &ctx, const InitParamsNonXDL &validParams,
-      GemmSize &gemmSize, DerivedParams &gemmADerivedParam,
-      DerivedParams &gemmBDerivedParam,
-      DerivedBlockGemmParams &blockGemmDerivedParam,
-      DerivedOutParams &gemmCDerivedParam, uint32_t &gridSize);
+      GemmSize &gemmSize, DerivedBlockGemmParams &blockGemmDerivedParam,
```
krzysz00: Note to future us - see if we can drop the `DerivedBlockGemmParams` into a function that gets called in the lowering process - all those parameters are a static function of block size anyway.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -74,7 +74,7 @@ static void computeLoadStoreTypeInfo(OpBuilder &b,
 
   // In case KPack and vector load is used, and we vector load on GemmK
   // dimension (1), use the last dimension (GemmKPack) instead.
-  if ((loadLength > 1) && (kPack > 1) && (vectorDim == GemmK)) {
+  if ((loadLength > 1) && (kPack > 1) && (vectorDim == 1)) {
```
krzysz00: Would it hurt much to use the enum here?

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: Ok, so, deeper question: what is the actual limit here?

krzysz00: But more immediate note: please file a ticket for that TODO

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -613,6 +614,12 @@ struct GridwiseGemmRewritePattern : public OpRewritePattern<GridwiseGemmOp> {
     LLVM_DEBUG(llvm::dbgs() << "a_block_space: " << a_block_space << "\n");
     LLVM_DEBUG(llvm::dbgs() << "b_block_space: " << b_block_space << "\n");
     LLVM_DEBUG(llvm::dbgs() << "double_block_space: " << block_space << "\n\n");
+
+    // TODO: adjust for data type and device
+    if (block_space * sizeof(float) > 64 * 1024)
+      return failure();
+    else
```
krzysz00: Linter will yell at me for the else there, we could just make this a 
```
return success();
```

More interestingly, both `success()` and `failure()` take boolean arguments.
So you can `return failure(block_space > ...);` and that'll do the right thing

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1627,7 +1642,18 @@ struct GridwiseGemmV2RewritePattern
                               GemmABlockCopyThreadSliceLengths_GemmM};
     }
 
-    uint32_t blockwiseVectorDimA = matrix_a_source_vector_read_dim;
+    auto dimToIndex = [](GemmDimension dim) {
```
krzysz00: I think this is spelled `static_cast<uint32_t>(dim)`
and if it isn't, I'd rather see this as a `static` function and not a lambda

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1114,23 +1131,19 @@ struct GridwiseGemmV2RewritePattern
...
+    int64_t matrix_b_source_data_per_read = 0;
+    GemmDimension matrix_a_source_vector_read_dim;
+    GemmDimension matrix_b_source_vector_read_dim;
+
+    int64_t aCopyPerThread = (KPerBlock * MPerBlock) / blockSize;
+    int64_t bCopyPerThread = (KPerBlock * NPerBlock) / blockSize;
+    GemmDimension vectorTiebreaker =
+        (KPack > 1) ? GemmDimension::K : GemmDimension::MorN;
+    std::tie(matrix_a_source_vector_read_dim, matrix_a_source_data_per_read) =
+        bestVectorization(b, op.a(), aCopyPerThread, vectorTiebreaker);
```
krzysz00: I see one problem here: the xdlops path (if you look in `GemmToGridwise`) still applies a separate `kpack` dimension to the inputs. The code here doesn't account for that and so the K dimension won't be analyzed correctly.

Can you add back in `Value unkpack(Value matrix) {}` and pass the result of that to vectorization inference?

krzysz00 (changes requested):
Overall, very nice PR, I've mostly got style nits.

However, there is a big bug here, and that's to do with kPack. We need to bring in the `unkpack()` function that the non-xdlops version didn't need, and the vectorization analysis needs to be passed a G x K x M/N matrix

In minor questions, have you checked some of the cases (`-p -x2` mainly, possibly one of the E2E tests) to see if they haven't lost any vectorization?

Secondly, do we have perf?

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: Or, alternatively, move the function that applies kpack out of `--gemm-to-gridwise` into `--grideise-gemm-to-blockwise` and use it's result anywhere that isn't the vectorization stuff

jerryyin: Nice catch, thank you!

> Can you add back in Value unkpack(Value matrix) {} and pass the result of that to vectorization inference?

I hope I did the right thing in f5e59f4.

> Or, alternatively, move the function that applies kpack out of --gemm-to-gridwise into --grideise-gemm-to-blockwise

I figured that was part of the coordinate transforms stuff so just hold off doing that and will instead do it in next PR.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Yep, agreed that a static_cast is cleaner way to do it. Done in f5e59f4.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Actually linter is able to tell if there's a return for both if and else statement at the end of the return, so this should be fine.

Agreed though that a last catch all return is a better style. Addressed in f5e59f4

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: In the immediate next line there's this enum dimension computation based on size:

> vectorDim = sliceLengths.size() - 1;

I'm not sure I completely comprehend what that does so just left the whole thing to be unit32 type.

mlir/include/mlir/Dialect/MIOpen/Tuning/GridwiseGemmParams.h
jerryyin: Agreed

krzysz00: Can you file that ticket?

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: That's handling the kpack case

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: > But more immediate note: please file a ticket for that TODO

Absolutely.

> Ok, so, deeper question: what is the actual limit here?

If I understand it right, the gfx908 limit on lds should be 64k bytes. Therefore the actual limit should account for both data type and size. I guess the only reason we are still fine is due to:
 - Later arch has larger lds size
 - Among supported types, float is the largest 

and since we are bounded by lower bound, this is okay.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: My copy of clang-tidy whines about
```
if (blah) {
 ...
  return foo;
} else {
  ...
}
```

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: Yeah, that's next PR

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemm.cpp
```diff
@@ -135,15 +135,6 @@ static void maybeSetAttr(StringRef attr, Operation *from, Operation *to) {
   if (maybeValue)
     to->setAttr(attr, maybeValue);
 }
-void affixGemmAttributes(Operation *convOp, Operation *gop, OpBuilder &b) {
```
krzysz00: Is maybeSetAttr() dead code now?

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: If else also have a return then it shouldn't... Wired.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -262,6 +262,22 @@ static TransformingForOp createLdsStoreLoop(OpBuilder &b, Location loc,
...
+/// introduced in conv-to-gemm. kPack is a LDS-specific transformation and
+/// should not have been applied to the gridwise gemm argument. Future work will
+/// make the conversion stop doing that, and we'll be able to get rid of this.
+static Value unKpack(OpBuilder &b, Value matrix) {
+  ArrayRef<int64_t> shape = matrix.getType().cast<MemRefType>().getShape();
+  if (shape.size() != 4)
+    return matrix;
+  BottomUpTMBuilder unkpack(b, {"g", "k", "d", "kpack"}, shape,
+                            matrix.getLoc());
+  unkpack.passThrough({"g", "m"});
```
krzysz00: ... for some reason the comment I left didn't show up here, so yeah, spelling mistake and it's surprising that this doesn't explode.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: Or maybe if { ... return } else { ... return } is fine, I haven't looked

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1076,10 +1109,13 @@ struct GridwiseGemmV2RewritePattern
...
     Value zeroConstantOp = b.create<ConstantIndexOp>(loc, 0);
 
+    Value matA = unKpack(b, op.a());
+    Value matB = unKpack(b, op.b());
+
     // Obtain critical matrix dimensions.
     ArrayRef<int64_t> aShape, bShape, cShape;
-    aShape = op.a().getType().template cast<MemRefType>().getShape();
-    bShape = op.b().getType().template cast<MemRefType>().getShape();
+    aShape = matA.getType().template cast<MemRefType>().getShape();
```
krzysz00: Is this safe? Where else do we use aShape?

krzysz00 (commented):
A few notes, but it's mostly good

Wrt getting more substantial testing, I'm happy with a nightly run by hand and a perf report

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Addressed in 2bbf5c6. It's not... I was looking at `aShape` and `bShape` only previously and you reminded me that those will determine the `G, K, M, N` sizes which will be different without kpack. Reverted back to op.a and op.b and wired thing about dimension typo caused regression also went away.

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: I can dig into this a bit. It is surprising how this can surprisingly hide another bug as you mentioned [here](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/768#discussion_r974620575).

mlir/lib/Dialect/MIOpen/Transforms/ConvToGemm.cpp
jerryyin: It's now gone in 6da74d5. Nice catch!

mlir/lib/Dialect/MIOpen/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: It actually just reminded me that this scheduling can get complicated in a case like: (Note this may not be a problem today but I can see it is as we fuse more aggresively)
 - We have consumed pretty much all lds space in convolution due to the nature of tuning favors larger tiles in general.
 - In a fusion pass, it ended up also require some lds space

In each local scope thing are fine because we didn't exceed upper bound of lds. But it will yield a kernel that over allocate and fail at execution time. 

@sjw36 Does fusion do any LDS allocations? If it does we may need some global flag indicating how much space is left globally.

krzysz00: Fusion doesn't do LDS allocations

mlir/test/mlir-miopen-driver/tuning_xdlops.mlir
```diff
@@ -2,21 +2,13 @@
 
 // RUN: miopen-gen -p -x2 | mlir-miopen-driver -miopen-affix-params -miopen-conv-to-gemm | FileCheck %s --check-prefix=STEP1
 // RUN: miopen-gen -p -x2 | mlir-miopen-driver -miopen-affix-params -miopen-conv-to-gemm -miopen-gridwise-gemm-to-blockwise | FileCheck %s --check-prefix=STEP2
-// RUN: miopen-gen -p --fil_layout=kyxc --in_layout=nhwc --out_layout=nhwk -x2 | mlir-miopen-driver -miopen-affix-params -miopen-conv-to-gemm | FileCheck %s --check-prefix=NHWC
 
 // STEP1: mPerWave
 // STEP1: nPerWave
-// STEP1: matrix_c_data_per_copy = 4
-// STEP1: matrix_c_dest_vector_write_dim = 4
-// STEP1: matrix_c_source_vector_read_dim = 2
```
krzysz00: So if we want to keep the spirit of this test in, I'd go to `--gridwise-gemm-to-blockwise` and look for a `threadwise_copy_v2` with length `4`

krzysz00 (approved):
I leave testing to your discression, unless you want me to poke at it
This code seems quite fine
Approved

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1114,23 +1150,28 @@ struct GridwiseGemmV2RewritePattern
...
+    GemmDimension matrix_a_source_vector_read_dim;
+    GemmDimension matrix_b_source_vector_read_dim;
+
+    int64_t aCopyPerThread = (KPerBlock * KPack * MPerBlock) / blockSize;
+    int64_t bCopyPerThread = (KPerBlock * KPack * NPerBlock) / blockSize;
+    GemmDimension vectorTiebreaker =
+        (KPack > 1) ? GemmDimension::K : GemmDimension::MorN;
+    std::tie(matrix_a_source_vector_read_dim, matrix_a_source_data_per_read) =
+        bestVectorization(b, matA, aCopyPerThread, vectorTiebreaker);
+    if (matrix_a_source_vector_read_dim == GemmDimension::K) {
```
krzysz00, line 1205: This should be guarded by `kpack > 1`

krzysz00: (because otherwise you're turning off vectorization in the k dimmension for non-kpack and that's going to kill our perf)

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1114,23 +1150,28 @@ struct GridwiseGemmV2RewritePattern
...
+      matrix_a_source_data_per_read =
+          std::min(matrix_a_source_data_per_read, KPack);
+    }
+
+    std::tie(matrix_b_source_vector_read_dim, matrix_b_source_data_per_read) =
+        bestVectorization(b, matB, bCopyPerThread, vectorTiebreaker);
+    if (matrix_b_source_vector_read_dim == GemmDimension::K) {
```
krzysz00, line 1225: And similarly

krzysz00 (changes requested):
I'm leaving a blocker for the kpack stuff mentioned earlier today so this doesn't land during your vacation

krzysz00:
State of the PR?

krzysz00:
Nightly pass! http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-768/12/pipeline 

Also, preliminary looks at the perf numbers show an average gain of several flops

krzysz00 (approved):
Since the weeklies passed, approved!

(It'll be nice to get perf, especially since both this and Manupa's PR are likely to land today, but that's a separate issue)

PR #769, opened 2022-09-14
Revert "* encourage minor identity mappings in elementwise conversion"

jungpark-mlir:
This reverts commit 45ecde6aedeb2085abee1785abbd4189be7c7a71.

Fixes a failure in MIGraphX integration test. Caused by folding 1x1x1x1xF32 -> F32 scalar type and further coordinate manipulation doesn't work (especially in fusion) as it doesn't have any coordinate.


sjw36 (approved):
Looks good.

PR #772, opened 2022-09-16
Track alignment in vectorizer (for backport)

krzysz00:
Previously, if we had, for example, an image with width 4 padded to
width 6 (on the right) with a filter of 2, the vectorizer would infer
a vectorization length of 2 in the M dimension because, all else being
held constant, you could fetch the vectors (I[0], I[1]) (I[2],
I[3]), [faild bounds check -> (0, 0)] without partial oob.

However, suppose we had added 1 to the input coordinate to handle the
case where x = 1. Then we'd fetch the vectors (I[1], I[2]), (I[3],
I[4]), (0, 0). Note that I[4], which is reading into the next row and
should've been (I[3], 0) instead.

This commit resolves the issue by tracking alignment in the
vectorizer, and using that information to resolve the vectorization of
padding.

### DIFF
--- a/mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
+++ b/mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
@@ -326,7 +326,7 @@ propagateUnmergeVectorization(T &&dimAndLength,
         // by multiplying in the alignment of this dimension. This is not
         // the case when a held-constant dimension has interveined or a previous
         // dimension was not fully aligned to its coefficient
-        if (result->alignment == previousDimsStride)
+        if (result->alignment % previousDimsStride == 0)
           result->alignment *= upperInfo.alignment;
       } else {
         break;
@@ -476,7 +476,7 @@ propagateVectorizationInfo(TransformMapAttr map,
               coefficient ==
                   (ourResult->maxLength * ourResult->needsCoefficient)) {
             ourResult->maxLength *= upperLen;
-            if (coefficient == thisAlignment)
+            if (thisAlignment % coefficient == 0)
               ourResult->alignment *= thisAlignment;
           } else {
             break;
--- a/mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
+++ b/mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
@@ -439,7 +439,7 @@ propagateVectorizationInfo(TransformMapAttr map,
     // vectorization lengths of dimensions, stopping if the accumulated length
     // ends up not equal to theat dimension's coefficient or if the coefficient
     // isn't equal to the accumulated value times the smallest coefficient seen
-    // on a dimension being vectorized. d.
+    // on a dimension being vectorized.
     case TransformType::Embed: {
       // Since embed coefficients can go in any order, we need them sorted
       auto &&zip = llvm::zip(params, upperDims);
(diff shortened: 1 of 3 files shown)


krzysz00:
PR does fix the bug but is currently a draft on account of a minor change needed to Embed and Unmerge

krzysz00:
Update: a parameter sweep confirms this fixed the core bug, and this is ready for review, though I'm doing another parameter sweep now that I tweaked a few things just to be safe.

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
```diff
@@ -241,15 +241,23 @@ namespace {
...
 /// coefficient that the dimension needs to be multiplied by for the dimension
 /// to vectorize properly when it is embeded into a broader context.
+/// The alignment field tracks what we know about how the values in a particular
+/// dimension will be aligned, with the initial assiumption being that the
+/// input dimension is aligned to the maximum vectorization length. This
+/// information is important in ensuring that a pad() that would otherwise
+/// vectorize doesn't have reads that go partly into the padding.
 struct VectorizationInfo {
-  int64_t maxLength = 0, needsCoefficient = 0;
+  int64_t maxLength = 0, needsCoefficient = 0, alignment = 0;
```
jerryyin, line 250: I think the base case, if a user doesn't provide length and alignment should that be both 1 for length and alignment.

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
```diff
@@ -300,26 +308,40 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+        result = VectorizationInfo(1, upperInfo.needsCoefficient, 1);
       // Catch
       // 1. Dimensions merged out of order and then
       // 2. Previous dimensions having incomplete vector lengths
       if (upperInfo.needsCoefficient == previousDimsStride &&
           (result->maxLength * result->needsCoefficient) ==
               previousDimsStride) {
         result->maxLength *= upperInfo.maxLength;
+        if (!previousAlign.has_value())
+          previousAlign = upperInfo.alignment * previousDimsStride;
```
jerryyin, line 327: Question: Why shall the alignment multiply the `previousDimsStride`? Could you give an example?

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
```diff
@@ -401,7 +439,7 @@ propagateVectorizationInfo(TransformMapAttr map,
     // vectorization lengths of dimensions, stopping if the accumulated length
     // ends up not equal to theat dimension's coefficient or if the coefficient
     // isn't equal to the accumulated value times the smallest coefficient seen
-    // on a dimension being vectorized.
+    // on a dimension being vectorized. d.
```
jerryyin: Unintended edit?

jerryyin (approved):

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
krzysz00: These are struct defaults, no one will see them, if they show up we have a bug

mlir/lib/Dialect/MIOpen/utility/transformMapUtils.cpp
krzysz00: Well, if the current dimension is aligned to 4 (that is, its initial value is divisible by 4) and the stride up to that point has been 3, we now have a value divisible by 12.

Now, since this value gets set all the time, it's fundamentally true that `previousDimsStride` is 1 here, but it makes sense to leave it in for symmetry's sake.

PR #770, opened 2022-09-15
[hotfix] Disable conv when padding is an exact multiple of filter width

krzysz00:
At some point, we introduced a bug that causes convolutions such as the following one to fail

./bin/miopen-gen --operation conv2d  -t f32 --fil_layout kcyx --in_layout nchw --out_layout nkhw --batchsize 64 --in_channels 64 --in_h 64 --in_w 64 --out_channels 64 --fil_h 2 --fil_w 2 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h_l 0 --padding_h_r 0 --padding_w_l 0 --padding_w_r 2 -ph -pv | ./bin/mlir-miopen-driver -c | rocm-run 2>&1 | less

Since we're about to release, and since this isn't a particularly common usecase, disable convolutions that use such parameters so we can work out the bug later.

PR #773, opened 2022-09-16
Backport #772 to release/rocm-5.4

jerryyin:
Revert "[hotfix] Disable conv when padding is an exact multiple of filter width (#770)"

This reverts commit 63d561b75386d528acb78e87a6188474c6e4cd12.

Resolve incorrect results by handling alignment in the autovectorizer

Previously, if we had, for example, an image with width 4 padded to width 6 (on the right) with a filter of 2, the vectorizer would infer a vectorization length of 2 in the M dimension because, all else being held constant, you could fetch the vectors (I[0], I[1]) (I[2], I[3]), [faild bounds check -> (0, 0)] without partial oob.

However, suppose we had added 1 to the input coordinate to handle the case where x = 1. Then we'd fetch the vectors (I[1], I[2]), (I[3], I[4]), (0, 0). Note that I[4], which is reading into the next row and should've been (I[3], 0) instead.

This commit resolves the issue by tracking alignment in the vectorizer, and using that information to resolve the vectorization of padding.

Add a test, loosen up alignment logic a bit to make it pass in a reasonable way

Fix up logic for merge, unmerge, embed

### DIFF
--- a/mlir/lib/Dialect/MIOpen/Transforms/ConvToGemm.cpp
+++ b/mlir/lib/Dialect/MIOpen/Transforms/ConvToGemm.cpp
@@ -889,17 +889,6 @@ template <typename T> struct Conv2DRewritePattern : public OpRewritePattern<T> {
       break;
     }
 
-    // Hotfix: handle problems with the gemm logic not interecting well with
-    // even padding Note: xdlops doesn't use any of the suspect logic yet, and
-    // so we can put this fix after we move to atomic add.
-    int64_t x = filterTransform.startSize("x");
-    if ((x > 1) && ((leftPadW > 0 && leftPadW % x == 0) ||
-                    (rightPadW > 0 && rightPadW % x == 0))) {
-      return op.emitOpError(
-          "padding of input tensor in w dimension that's an exact multiple of "
-          "the filter width is temporarily disabled");
-    }
-
     TransformMapAttr filterTransformAttr = filterTransform.get();
     Value gemmFilter =
         b.create<TransformOp>(loc, op.filter(), filterTransformAttr);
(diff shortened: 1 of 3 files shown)


krzysz00 (approved):
Can we get the one last change (removing the `d.` edit) from the other PR?

jerryyin:
Release testing:
 - xdlops: http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-773/3/pipeline/21/
 - nonxdlops: http://ml-ci.amd.com:21096/blue/organizations/jenkins/MLIR%2Frelease-branch-testing/detail/release-branch-testing/40/pipeline

yiqian1 (approved):

jerryyin:
@krzysz00 
The existing condition is that nonxdlops release testing for this PR is passed whereas xdlops release testing is blocked by CI. I have tried kick off CI multiple times across the weekend but it either run into lockhart issues or blocked by limited instance of gfx908. I doubt the situation improve soon.

Do you think this PR will impact tuning in any way? I'm thinking of asking MIOpen to tune the 5.4 release MLIR package without this PR. Those configs that were blocked off are less likely to be production kernels, correct?

krzysz00:
@jerryyin Since this PR doesn't impact the xdlops path, I claim we can fake the xdlops testing and merge

Now, as to tuning ... I'd want to basically look at the configs MIOpen tunes to see if they use filter width equal to the padding (where filter width > 1)

jerryyin:
@krzysz00 Go ahead but please kick off a xdlops release testing on ROCm-5.4 branch. We need that record that there's a full pass in release testing.

@yiqian1 Please file the ticket to Tuna team to have them start tuning all solvers.

krzysz00:
@yiqian1 Let's not file that ticket until this PR lands?

yiqian1:
> @yiqian1 Let's not file that ticket until this PR lands?

Will file the ticket to Tuna team after this PR is in.

PR #775, opened 2022-09-21
[TESTING] Fixed a bug in checking xdlops presense

manupak:
This commit fixes a bug that checks the presense
of the xdlops where the search is not concat'ing
the string containing all the valid gpus with xdlops.


jungpark-mlir (approved):
LGTM

PR #777, opened 2022-09-21
Generalize CI testing for multiple chip features

zhanglx13:
This PR does the following: 
1. [Build and Test] CI extension
    - Instead of performing [sharedLib fixed tests], [sharedLib random tests], and [MIOpen test with librockCompiler] in parallel on gfx908||gfx90a or gfx906, the three tests are concatenated into a streamline and executed on different codepaths:
        - `mfma` codepath runs on gfx908||gfx90a GPUs on private CI
        - `navi21` codapath runs on gfx1030 GPUs on private CI
        - `vanilla` codapath runs on gfx906 GPUs on public CI
    - One special note about tests on navi21 codepath: **the number of lit workers needs to be limited to 8 on `ixt-sjc2-84`**, otherwise tests fail (here is an [example](http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-777/107/pipeline/141)). 
2. [Parameter Sweep] CI extension (weekly)
    - Enabled gfx1030
    - `parameterSweeps.py` now gets rid of `xdlops`. Instead, it picks perf_config according to the `codepath`, which can be obtained in the following two ways
        1.  command line option `--codepath` 
        2.  if `--codepath` is not provided or not supported, it is inferred from `arch` information, which is obtained from `rocm_agent_enumerator`.
3. [Tune MLIR Kernels] CI extension (weekly)
    - Enabled gfx1030
    - `miopen_validate.sh` now takes `-c|--codepath` to determine the MLIR solver name.
4. [MIOpen Resnet50 Config Test] CI extension
    - Reorganized the matrix structure so that each cell corresponds to a codepath and runs `miopen_validate.sh --test-all --no-tuning` with and w/o perfDB.
5. [Benchmark and performance report] CI extension (nightly)
    -  benchmark and performance stage is executed on all chips: gfx906, gfx908, gfx90a, and gfx1030. Only gfx906 tests run on public CI, the rest run on private CI.
    - `MIOpenDriver.py` now gets rid of `xdlops`. It uses `arch` and `chip` information everywhere to determine whether xdlops is enabled.
        - `MIOpenDriver.py` also accepts a command line option `--rocmlir_gen_flags <flags>`, which will be passed to `rocmlir-gen` directly to toggle each feature.
    - performance-related reports (`.csv`) are generated for each chip. Also, each chip will have its own performance report (`*.html`) tag in Jenkins.
    - In the report, `xdlops` column is replaced with `chip`.
6. Jenkins CI interface
    - Choice [codepath]
        - default: `mfma` and `navi21` codepath are executed on the private CI and `vanilla` codepath is executed on the public CI
        - mfma/navi21: only the selected codepath is executed
        - vanilla: the `vanilla` codepath is executed on the private CI 
    - Boolean [disableNavi21]: if selected, the `navi21` codepath/chip cells are disabled. This is useful to let PR pass in case navi21 nodes are not functioning well.

At this stage, we set `navi21` as an individual codepath. However, the generated kernel on `navi21` codepath is the same as the `vanilla` codepath. Later
1. when navi21 is fully brought up and optimizations are fully explored, we can move gfx1030 GPUs into the `vanilla` codepath
2. when navi3x GPUs are available for us to test, we can add a new codepath `wmma` for it.


zhanglx13:
Monitor of failed tests:
- ([build#7](http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-777/7/pipeline/81)) Navi21 fixed tests: 
    - failed test: `rocmlir-driver/e2e_for_pr/Resnet50_validation_nchw.mlir --check-prefix=CHECK_NCHW_2`
        - error: `Memory access fault by GPU node-2 (Agent handle: 0x561a119ee690) on address 0x7feb3d514000. Reason: Page not present or supervisor privilege`
- ([build#13](http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-777/13/pipeline/81)) Navi21 fixed tests
    - failed test: `fusion/e2e/tosa-to-rock-relu.e2e.mlir` 
        - error: `Warning: unknown chipset major revision, falling back to defaults: Memory access fault by GPU node-2 (Agent handle: 0x5600d925acc0) on address 0x7f0036e06000. Reason: Page not present or supervisor privilege.`
- ([build#15](http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-777/15/pipeline/81)) Navi21 fixed tests
    - failed test: `rocmlir-driver/conv2d_harness/conv2d_harness_ckyx_cnhw_knhw.mlir`
        - error: `Memory access fault by GPU node-2 (Agent handle: 0x5581c779c6d0) on address 0x7fd023cd7000. Reason: Page not present or supervisor privilege`
    - failed test: `rocmlir-driver/e2e_for_pr/Resnet50_validation_nhwc.mlir --check-prefix=CHECK_NHWC_1`
        - error: the difference between gpu and cpu results are too large, e.g. `cpu: 8.000000 gpu: -278.000000`
- ([build#17](http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-777/17/pipeline/79)) Navi21 fixed tests: A lot of tests have the error: `hipGetDeviceProperties() should never fail`
- ([build#18](http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-777/18/pipeline/79)) Same issue as in [build#17](http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-777/17/pipeline/79).
- ([build#21](http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-777/21/pipeline)) Same issue as in [build#17](http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-777/17/pipeline/79) + `Backing channel 'ixt-sjc2-84' is disconnected.`

krzysz00 (commented):
I'm not sure I agree with this

I don't think the feature toggle is needed

And I claim that, in general, we should go with `mlir && (rocm || navi21)` . Then, we should extend the bits that actually care about architecture (that is, the weekly tuning runs) to include navi21.

zhanglx13:
What I am doing here is to extend the current CI philosophy to include future gpus.
The current PR testing philosophy:
1. We choose `canXdlops` as one of the build parameters
2. Jenkins chooses the node accordingly
    - if `canXdlops` is chosen, pick `mlir && (gfx908 || gfx90a)`
    - if not, pick `mlir && rocm` or `mlir && gfx906`
3. In the build stage, the macro `ROCMLIR_DRIVER_XDLOPS_TEST_ENABLED` is set accordingly.
4. And the above macro is used to set the `%xdlops` variables in many E2E tests under `rocmlir-driver`.

So basically, we start from the feature (`canXdlops`), which is used by both Jenkins (for picking node) and our builds (for enabling/disabling `-x2` flag).
Therefore, when we have more features, it is cleaner to put all features in a group so that developers can pick what feature to test.
Here are my claims:
1. For PR CI testing, we care more about the features than the chip. E.g. for the `mfma` feature, we do not care whether 908 or 90a is chosen.
2. In some circumstances, such as new target bringup, developers need to test on a particular chip. This is added as another choice as one of the build parameters.

> Then, we should extend the bits that actually care about architecture (that is, the weekly tuning runs) to include navi21.

Agreed. But let's figure out the PR CI first.

krzysz00:
To summarize, from my perspective, "add Navi CI" means "add Navi nodes to the pool of CI nodes, and add one bit of testing to ensure Navi in particular works"

That is, "add Navi CI" is like "add MI-200 CI" from a while back. All this feature flag stuff is the wrong approach, IMO.

Our current approach to MI-200 support is that
1. Statistically, some of our jobs land on MI-200 and some on MI-100. We can have similar for Navi, and;
2. Tuning is explicitly run on both MI-100 and MI-200. We can do the same for Navi.

Now, the note here is that gfx900 doesn't support dot products, while gfx906 and gfx1030 do. I can see the argument for explicitly testing the dot and the non-dot path. If we want to explicitly check each of the distinct code paths, I propose that
1. We rename the canXdlops flag to `internalCi`, so that we can gate branches of the setup as appropriate
2. We explicitly have a matrix that splits over which feature we're testing: mfma, dot, and, in the future, wmma. Each matrix segment gets a corresponding set of nodes it can run on
3. Weekly CI's tuning targets all architectures by name
4. (The no-features case will require a build flag of some sort, perhaps, but that's future work, and also gfx900 isn't officially on the supported list so whatever)

krzysz00:
Ok, since I didn't see your moment before adding mine, having an optional "restrict to chip" widget isn't inherently unreasonable.

Also, PR CI and nightly CI and all live in the same file by design, so we can't just do PR CI here.

zhanglx13:
Actually, gfx908 and gfx90a both support the dot product instructions. And given that `dot` is similar to `fma` in terms of complexity, maybe it should not be a member of the feature set. What do you think @krzysz00 @jerryyin?
If so, we only need to take care of "big" features like `mfma` and `wmma` and they are mutually exclusive in terms of nodes.

Re "Add MI-200 CI", do you know anywhere that we care about the differences between MI-100 and MI-200 (except the f16 experiments that are affected by the flush-denorm-to-zero behaviors)? 
If not, then adding Navi CI should be different from adding MI-200 CI. And to me it looks more like adding a feature.
But if dot is not considered as a feature, adding Navi21 CI is not adding anything. Therefore, this PR is more like targeting the future Navi3x.

krzysz00:
See, the thing with the dot product is that there're some gfx101X that don't have it, so we do need to test both cases, I'd argue.

And so really what I'm saying is that we should generalize the CI so that feature testing isn't just xdlops/non-xdlops but one CI run for `mfma`, one for `dot`, (one for `wmma` once we're there) and so on.

Aka pull out `matrix {` again

zhanglx13:
Ok, let me clarify something.
1. We do not have gfx101x nodes on CI. Are we going to have some in the future? If not, do we still need to set up CI run for dot?
2. Are you saying for each PR, we need to test on all features? I was thinking about this as the changes in the PR can be relevant on all gpus. But for a PR that is targeting a particular feature, do we still need to test it against other features?
3. And for non-feature testing (currently the non-xdlops), it is running on the public CI so that we do not need to set it up in `Jenkinsfile` here, right?

krzysz00:
1. We don't have any 101X, and I don't know if we will. Even if we do they might be on the as-is list, and so the main reason we can't go for dot product everywhere is the gfx900 nodes in public CI
2. I claim that PR CI should run on all features, analogously to how we have xdlops and non-xdlops PR CI now. See also MIOpen's CI, which runs once per architecture
3. Nightly and weekly public CI jobs draw from `Jenkinsfile` not `Jenkinsfile.downstream`

Also, this PR's title should probably say"gfx10 features" in the interests of something something internal codenames something something

zhanglx13:
> so the main reason we can't go for dot product everywhere is the gfx900 nodes in public CI

Well, since this is the only target and chances are we may drop support for this gpu in the near future, can we push the burden of generating non-dot kernels to lowering passes? We have your GemmFeature list for lowering passes to query information about whether dot is supported.

> I claim that PR CI should run on all features, analogously to how we have xdlops and non-xdlops PR CI now. See also MIOpen's CI, which runs once per architecture

Make sense.

> this PR's title should probably say"gfx10 features"

I am planning to prepare for the `wmma` feature as much as I can. What about "multi features" since this PR tries to generalize the CI for multiple features?

krzysz00:
1. Oh, yeah, dot vs. non-dot is definitely only a matter of `rock.threadwise_gemm`. The way I'd see it is that if you have a dot product you load two f16/four i8 at a time and do the dot product, otherwise you load one and do an FMA (probably with an extension, but there might also be mixed-precision FMA hiding behind the intrinsic). The code'll be general enough and small enough that we probably don't need explicit testing of dot vs. non-dot
2. So, really, we don't need "has dot product support" as an axis in the CI. Everything we tune for has it, and we'll get some exercise for the non-dot code path when jobs land on gfx900 nodes. Although if we wanted to be very careful we could have a nightly run that specifically targets gfx900 (or that uses a currently-nonexistent option to force dot products off)
3. Yeah, something like "Generalize CI handling of chip features" might be a good name

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -81,46 +81,53 @@ void preMergeCheck() {
...
-        reportName: 'Performance report'
+        reportFiles: "${chip}_MLIR_Performance_Changes.html,${chip}_MLIR_vs_MIOpen.html",
+        reportName: "Performance report for ${chip}"
     ])
 
-    plot csvFileName: 'plot-nightly-perf-results-000001.csv',\
-        csvSeries: [[file: 'build/mlir_vs_miopen_perf_for_plot.csv', displayTableFlag: false]],\
+    plot csvFileName: "${chip}_plot-nightly-perf-results-000001.csv",\
+        csvSeries: [[file: "build/${chip}_mlir_vs_miopen_perf_for_plot.csv", displayTableFlag: false]],\
         title: 'Test performance summary',\
```
krzysz00: This'll need chip as well

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -130,10 +137,113 @@ void setHeartbeat() {
...
+        sh """
+            bash ${WORKSPACE}/mlir/utils/jenkins/miopen-tests/miopen_validate.sh --test-fwd --dtype int8 --tuning\
+            < ${WORKSPACE}/mlir/utils/jenkins/miopen-tests/resnet50-miopen-configs"""
+        sh 'ls -l MIOpenUserDB'
+    }
+}
+
+void check_fixedE2ETests(boolean nightly) {
+    buildProject('check-mlir check-rocmlir', """
+              -DROCMLIR_DRIVER_PR_E2E_TEST_ENABLED=${nightly ? '0' : '1'}
```
krzysz00: Now that I think of it ... why isn't this an unconditional '1'? Can you change that while you're here?

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -170,180 +288,118 @@ pipeline {
...
+                            anyOf {
+                                equals expected: "mfma", actual: params.codepath;
+                                allOf {
+                                    equals expected: "navi21", actual: params.codepath;
+                                    equals expected: false, actual: params.disableNavi21;
+                                }
+                                equals expected: "vanilla", actual: params.codepath;
+                            }
+                            equals expected: "${CODEPATH}", actual: params.codepath;
+                            equals expected: true, actual: params.canXdlops;
```
krzysz00, line 340: I see a problem here: we don't have any `gfx906` on private CI

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -153,6 +263,14 @@ pipeline {
                      description: 'Run the static library stage')
         booleanParam(name: 'perfTest', defaultValue: true,
                      description: 'Run the performance testing stage')
+
+        // choose the codepath for testing
+        choice(name: 'codepath',
+               choices: ['default', 'mfma', 'navi21', 'vanilla'],
```
krzysz00, line 277: Doesn't this need a `defaultValue: 'default'`

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -170,180 +288,118 @@ pipeline {
...
+                            sh 'rm build/CMakeCache.txt'
+                            buildMIOpenWithMLIR()
+                            testMIOpenDriver(/*testInt8=*/true, /*tuning=*/false, "selected-resnet50-miopen-configs")
+                            testMIOpenDriver(/*testInt8=*/true, /*tuning=*/true, "selected-resnet50-miopen-configs")
                         }
-                        stage("Build MIOpen with librockCompiler") {
-                            steps {
-                                buildMIOpenWithMLIR()
+                    }
+                    stage('Static Tests') {
```
krzysz00: Come to think of it, could we move this step after fixed e2e tests?

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -170,180 +288,118 @@ pipeline {
...
+                        when {
+                            allOf {
+                                equals expected: false, actual: params.nightly;
```
krzysz00: One bit of weirdness that could be possible: the static checks assume you have a compile_commands.json around, and so need the project to be configured - so if we try to cheat the CI (as we sometimes need to do) this'll fall over without any way to fix it. Since we do want these tests to run always, how about this:
- Split the shared library build and test into a build step (`check-mlir-build-only check-rocmlir-build-only`) and a test step (`check-mlir` `check-rocmlir`) and stick the static tests in between. You could even have them in the same `steps` block - I think Jenkins has a `when{}` at that level.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -170,180 +288,118 @@ pipeline {
...
-                        PATH="/opt/rocm/llvm/bin:$PATH"
-                        HOME="${WORKSPACE}"
-                    }
-                    stages {
-                        stage('Environment') {
-                            steps {
-                                showEnv()
-                            }
+                        steps {
+                            sh 'rm build/CMakeCache.txt'
```
krzysz00, line 389: A thing I'd like you to test, though maybe in a followup PR:
Instead of deleting the CMake cache to force a full rebuild, edit the top-level CMakeLists.txt to go to that `if (BUILD_FAT_LIBROCKCOMPILER)` and set all the `set()`s in the arms of that if to include the `FORCE` flag so that we re-set the various variables we're controlling. This may also involve moving the `set()` of the install prefix out of the `if()`

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -170,180 +288,118 @@ pipeline {
...
-                    environment {
-                        HOME="${WORKSPACE}"
+                }
+                environment {
+                    PATH="/opt/rocm/llvm/bin:$PATH"
+                    HOME="${WORKSPACE}"
+                }
+                stages {
+                    stage('Environment') {
+                        steps {
```
krzysz00, line 350: Do we already have an echo of which codepath it is or should we add it here?

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -505,224 +542,217 @@ pipeline {
...
-                // Note: add additional architectures here
-                dir ('MIOpen/build/MIOpenUserDB') {
-                    unstash name: "MLIR-PerfDB-${params.canXdlops ? 'gfx908' : 'gfx906'}"
-                    // This'll stop being a duplicate once Navi CI starts
-                    unstash name: "MLIR-PerfDB-${params.canXdlops ? 'gfx90a' : 'gfx906'}"
-                    sh 'date --utc +%Y-%m-%d >tuning-date'
-                }
-                archiveArtifacts artifacts: 'MIOpen/build/MIOpenUserDB/**',\
-                    onlyIfSuccessful: true
+                archivePerfDB()
```
krzysz00, line 545: SInce I can't attach to that line of code, I'll point out we don't have to say "xdlops" in the stage name here

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -505,224 +542,217 @@ pipeline {
...
-                                               projectName: "/MLIR/mlir-weekly",\
-                                               selector: lastSuccessful(),\
-                                               target: 'MIOpen/build/MIOpenUserDB'
-                        sh 'ls MIOpen/build/MIOpenUserDB'
-                        sh 'cat MIOpen/build/MIOpenUserDB/tuning-date'
+                agent {
+                    docker {
+                        image dockerImage()
+                        args dockerArgs()
+                        label "${CHIP}"
```
krzysz00: `mlir && ${CHIP}` so we don't steal nodes

mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -612,39 +619,34 @@ def main(args=None):
...
         else:
             # Will only reach here with more than 1 unspecified arguments
             # These are arguments are directly passed through to benchmarkMLIR
             if not parsed_args.mlir_build_dir:
                 raise RuntimeError("MLIR build dir was not provided/found")
-            df = pd.DataFrame([benchmarkMLIR(parsed_args.config, xdlops, paths, arch)])
-        df.to_csv(parsed_args.fileName)
+            df = pd.DataFrame([benchmarkMLIR(parsed_args.config, paths, arch)])
+        chip = GFX_CHIP_RE.search(arch).group(0)
+        df.to_csv(chip + '_' + parsed_args.fileName)
```
krzysz00: Minor comments:

I can't remember if we had a `--xdlops` option here, but, whether or not we did, can you add a `--rocmlir-gen-flags` option that then gets added in to the rocmlir-gen commands? I'd like to have it for by-hand experiments.

krzysz00: Also, instead of doing this, please adjust the defaults for `filename` to include `chip`. I want `-o` to mean `-o`, not some weird hybrid of `-o` plus this behavior

mlir/utils/jenkins/miopen-tests/miopen_validate.sh
```diff
@@ -177,7 +178,7 @@ function run_tests() {
     exit_status=$?
 
     rm "$TMPFILE"
-    echo -n "$DTYPE $LAYOUT XDLOPS=$XDLOPS DIR=$DIRECTION: "
+    echo -n "$DTYPE $LAYOUT CODEPATH=$CODEPATH DIR=$DIRECTION: "
```
krzysz00, line 181: Good refactor!

mlir/utils/jenkins/parameterSweeps.py
```diff
@@ -407,23 +407,32 @@ def main() -> bool:
         help="The build directory of MIOpen",
     )
     args = parser.parse_args()
-    options = Options(debug=args.debug, quiet=args.quiet, xdlops=args.xdlops,
+    options = Options(debug=args.debug, quiet=args.quiet,
         arch=','.join(getArch()), concurrent_tests=args.jobs)
     paths = MIOpenDriver.create_paths(args.mlir_build_dir, args.miopen_build_dir)
 
+    codepath = "none"
```
krzysz00: Can we get a command-line option to force the codepath?

(and probably a similar `--rocmlir-gen-flags` as in the main perf script?)

mlir/utils/jenkins/perfRegressionReport.py
```diff
@@ -60,12 +60,13 @@ def getPerfDate(statsPath: PurePath, default="???"):
         return default
 
 if __name__ == '__main__':
-    oldDataPath = PurePath(sys.argv[1]) if len(sys.argv) >= 2\
-        else PurePath('./', 'oldData/', reportUtils.PERF_REPORT_FILE)
-    newDataPath = PurePath(sys.argv[2]) if len(sys.argv) >= 3\
-        else PurePath('./', reportUtils.PERF_REPORT_FILE)
-    outputPath = PurePath(sys.argv[3]) if len(sys.argv) >= 4\
-        else PurePath('./', 'MLIR_Performance_Changes.html')
+    chip = sys.argv[1]
```
krzysz00, line 63: ... note to self, clean up the options parsing in here

krzysz00 (commented):
Fantastic refactoring! I think, soon after this, we'll be able to lose `Jenkinsfile.downstream` as well.

I've mostly got nitpicks and minor ordering complaints - my biggest notes are making sure static checks work in a CI cheat and adding some options to the perf scripts to allow flexibality when running by hand.

One observation I have is that this new setup will lead to our CI occupying one node for longer instead of two nodes for shorter in the PR case, hence me poking at whether we can avoid the full rebuilds in the shared/static switches. (That change I can leave to a future PR, but I wanted to flag it down)

@yiqian1, since you're the other CI person (also, I'll recommend trying to view this PR from `git show origin/navi-ci:mlir/utils/jenkins/Jenkinsfile` as the diff is a hot mess)

mlir/utils/jenkins/Jenkinsfile
zhanglx13: Setting it to 1 will run tests in the `e2e_for_pr` also in the nightly, which is a duplicate of PR CI. Do you want a double check of these tests?

mlir/utils/jenkins/Jenkinsfile
zhanglx13: I found some here: http://rocmhead.amd.com:8080/label/gfx906/ and `-hq-35` and `-hq-36` also have the `mlir` label.

mlir/utils/jenkins/Jenkinsfile
zhanglx13: No. The default is the first choice.

mlir/utils/jenkins/Jenkinsfile
krzysz00: Huh, good to know

mlir/utils/jenkins/Jenkinsfile
krzysz00: Eh, it's fine either way. We can probably keep things as they are.

mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -196,17 +198,24 @@ def __repr__(self):
...
+                dilationH={self.dilationH!r}, dilationW={self.dilationW!r}, group={self.group!r}, arch={self.arch!r})"""
 
-    def generateMlirDriverCommandLine(self):
+    def generateMlirDriverCommandLine(self, rocmlir_gen_flags):
         direction = {'fwd':'--operation conv2d',
                      'bwd':'--operation conv2d_bwd_data',
                      'wrw':'--operation conv2d_bwd_weight'}[self.direction]
 
+        feature_flags = rocmlir_gen_flags
+        if rocmlir_gen_flags == '':
```
krzysz00: I don't think we need this - I'd like architectural defaults to kick in if I don't give flags, and I don't want to create one more place we can forget to change the flags

krzysz00: Also, those flags should be `split()` and then appended to the end because the process launcher won't be doing any other lexing of the argument stream.

mlir/utils/jenkins/parameterSweeps.py
```diff
@@ -39,17 +39,26 @@ def __repr__(self):
...
+                group={self.group!r}, arch={self.arch!r}, perfConfig={self.perfConfig!r})"""
 
-    def generateMlirDriverCommandLine(self) -> Sequence[str]:
+    def generateMlirDriverCommandLine(self, codepath) -> Sequence[str]:
         direction = {'fwd': 'conv2d',
                      'bwd': 'conv2d_bwd_data',
                      'wrw':'conv2d_bwd_weight'}[self.direction]
 
+        feature_flags = ''
+        if codepath == 'mfma':
```
krzysz00: Same comment re flags as in the other instance...

mlir/utils/jenkins/Jenkinsfile
zhanglx13: I'll address this issue in a future PR.

krzysz00 (commented):
(apparently I didn't send some comments)

krzysz00:
Overall, I'm happy to let this in - just the one comment about flag handling. I just wanted to make sure it's tested and possibly to get @yiqian1 to get a second pair of eyes on it.

mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -595,10 +603,35 @@ def main(args=None):
...
+        help="rocmlir-gen flags to toggle each feature"
+    )
+
     parsed_args = parser.parse_args(args)
 
-    # Impose default behavior when no args have been passed
-    if len(args) == 0:
+    args_len = len(args)
+    # If rocmlir_gen_flags not provided, infer it from the chip
+    rocmlir_gen_flags = ''
```
krzysz00, line 617: Let me clarify:
1. rocmlir_gen_flags is not related to this args_len business: the point of args_len is to default to compaliring to MIOpen and MLIR if you call the script with no options
2. You must not infer default values for rocmlir_gen_flags. Don't do this switch-on-arch business. What you should do with rocmlir_gen_flags is split it on whitespace and append those split values to the end of the arguments to rocmlir_gen. rocmlir_gen should be inferring from `arch` by default, but I want rocmlir_gen_flags specifically to allow me to do run, on a gfx908, `./bin/MIOpenDriver.py -b --rocmlir_gen_flags="-mfma=off"` to get non-xdlops perf numbers

krzysz00 (changes requested):
Requesting changes to `rocmlir_gen_flags`

mlir/utils/jenkins/MIOpenDriver.py
zhanglx13: 1. So `MIOpenDriver.py --rocmlir_gen_flags="-mfma=off"` will never happen? which means when comparing MIOpen and MLIR you always want to infer the features from the arch?

krzysz00: No, it could happen, and it has the semantics "compare MIOpen and MLIR, when invoking `rocmlir-gen` add the flags "-mfma=off" to the command line"

It's a somewhat silly thing to ask for, and if someone actually wants to do that exact comparison, they'll probably need to patch in not setting the solver to the Xdlops version on the MIOpen side.

krzysz00: `--rocmlir_gen_flags` is the very simple option "Trust me, I want rocmlir-gen to get some extra arguments, just do it." This is by analogy to, say, `CFLAGS`

zhanglx13: If that is the case, `MIOpenDriver.py --rocmlir_gen_flags='-mfma=off'` will not set `batch_both` unless we rule out `--rocmlir_gen_flags` from `args_len`.

krzysz00: ... good point

krzysz00: `./bin/MIOpenDriver.py --rocmlir_gen_flags="-mfma=off"` is an error because you didn't specify a mode

krzysz00: ... but if you go and add `--batch_both` ... I'll assume you know what you want

zhanglx13: So do we take no mode as an error now or take it as `batch_both` as before?

mlir/utils/jenkins/parameterSweeps.py
```diff
@@ -407,23 +413,44 @@ def main() -> bool:
...
     )
     args = parser.parse_args()
-    options = Options(debug=args.debug, quiet=args.quiet, xdlops=args.xdlops,
-        arch=','.join(getArch()), concurrent_tests=args.jobs)
+
+    arch = ','.join(getArch())
+    supported_codepath = ['mfma', 'vanilla']
+    # If codepath not provided or not supported, infer it from the arch
+    codepath = args.codepath
+    rocmlir_gen_flags = ''
```
krzysz00: You've still got the weirdness for rocmlir_gen_flags in here too

krzysz00 (commented):
One more bit of cleanup on flags handling in parameter sweeps

mlir/utils/jenkins/parameterSweeps.py
zhanglx13: So you don't need the flag for `parameterSweeps.py` or you also want `rocmlir_gen_flags` as a command line option besides `codepath`?

krzysz00: Ah, I got thrown by the name.

The codepath-based method is fine, I don't think anyone will need any parameter sweeps weirder than that.

I'd go ahead and make `rocmlir_gen_flags` a list in this version so you don't have to `split()` it every time, but that's nitpicking and won't block the PR

krzysz00 (approved):
Approved, with the understanding that we won't land this until it's validated.

I've left a tiny nit on parameter sweeps, but nothing that blocks us

zhanglx13:
- Build of PR CI: [build#110](http://rocmhead.amd.com:8080/job/MLIR/job/mlir/job/PR-777/110/pipeline/)
- Build of weekly CI: [build#125](http://rocmhead.amd.com:8080/job/MLIR/job/mlir/job/PR-777/125/pipeline/): 
    - Tune MLIR stage failed, which happens on all arch and also in the [weekly CI](http://rocmhead.amd.com:8080/job/MLIR/job/mlir-weekly/97/)
    - Parameter sweep stage for gfx1030 is done on my dev machine (velocity-micro-1)
- Build of nightly CI: [build#108](http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-777/108/pipeline/)

yiqian1 (approved):

yiqian1:
LGTM. We may want to monitor the CI job execution time after merge.

zhanglx13:
I am merging this PR with navi21 codepath disabled. The addition of navi21 jenkins agents will be addressed in a following PR.

PR #778, opened 2022-09-23
Enable gfx1030 for MIOpenDriver

zhanglx13:
This PR updated the supported chips so that `ConvMlirIgemm*` solvers can be applicable for `MIOpenDriver` on Navi 21 machines.


zhanglx13:
The testing on Navi21 has passed in this [build](http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-777/16/pipeline).

krzysz00 (approved):
I'd go grepping for other mentions of gfx908 or such so as to make sure we didn't miss a spot, but there's no real reason not to approve this

mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
```diff
@@ -264,13 +264,19 @@ LogicalResult Conv2dGenerator::hasValidDimension() const {
...
+  // We support in between gfx900 to gfx908 and gfx1030 for nonxdlops algorithm
+  // For example, gfx803, gfx90c are unsupported now
   unsigned int chipHexNumber = 0;
   if (sscanf(config.chip.c_str(), "gfx%x", &chipHexNumber) != 1)
     return failure();
 
-  if ((chipHexNumber > 0x90a) || (chipHexNumber < 0x900))
+  constexpr size_t NUM_SUPPORTED_CHIPS = 5;
+  static const unsigned int supportedChips[NUM_SUPPORTED_CHIPS] = {
+      0x900, 0x906, 0x908, 0x90a, 0x1030};
```
jerryyin, line 275: 0x900 needs to be removed. I is dropped by ROCm overall and we should not add it back because it has no QA coverage.

krzysz00: I claim we should allow it with a warning in the interests of ROCm everywhere

PR #780, opened 2022-09-27
[DO NOT SQUASH] Move to using HIP config instead of HIP module in CMake

krzysz00:
To quote the external commit

[[mlir] Use hip's config mode to find libraries](https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/780/commits/953bf47f35f037ea0d050c07b1d11367e0e69067)
[953bf47](https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/780/commits/953bf47f35f037ea0d050c07b1d11367e0e69067)

Instead of using find_package(HIP) to find FindHIP.cmake, which
doesn't seem to be the preferred way to find HIP anymore, use
find_package(hip CONFIG) to find the HIP configuration. Give
preference to ${ROCM_PATH} over ${ROCM_PATH}/hip in order to handle
the fact that newer ROCm versions prefer the include path to use
${ROCM_PATH}/include/hip over ${ROCM_PATH}/hip/innclude/hip (the
latter throws up a bunch of deprecation warnings)

Then, instead of trying to manually find the host-side headers and
runtime library by hand, use the hip::host and hip::amdhip64 libraries
that the config module defines.

This makes the CMake config much less error-prone and brings it in
line with the recommended approach to finding HIP.

Fixes
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/658

### DIFF
--- a/external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt
+++ b/external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt
@@ -197,23 +197,8 @@ if(MLIR_ENABLE_ROCM_RUNNER)
       set(ROCM_PATH $ENV{ROCM_PATH} CACHE PATH "Path to which ROCm has been installed")
     endif()
   endif()
-  set(HIP_PATH "${ROCM_PATH}/hip" CACHE PATH "Path to which HIP has been installed")
-  set(CMAKE_MODULE_PATH "${HIP_PATH}/cmake" ${CMAKE_MODULE_PATH})
-  find_package(HIP)
-  if (NOT HIP_FOUND)
-    message(SEND_ERROR "Building mlir with ROCm support requires a working ROCm and HIP install")
-  else()
-    message(STATUS "ROCm HIP version: ${HIP_VERSION}")
-  endif()
-
-  # Locate HIP runtime library.
-  find_library(ROCM_RUNTIME_LIBRARY amdhip64
-               PATHS "${HIP_PATH}/lib")
-  if (NOT ROCM_RUNTIME_LIBRARY)
-    message(SEND_ERROR "Could not locate ROCm HIP runtime library")
-  else()
-    message(STATUS "ROCm HIP runtime lib: ${ROCM_RUNTIME_LIBRARY}")
-  endif()
+  list(APPEND CMAKE_PREFIX_PATH ${ROCM_PATH} "${ROCM_PATH}/hip")
+  find_package(hip REQUIRED)
 
   if (NOT DEFINED ROCM_TEST_CHIPSET)
     execute_process(COMMAND "${ROCM_PATH}/bin/rocm_agent_enumerator"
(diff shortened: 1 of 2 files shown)


external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt
```diff
@@ -208,7 +207,7 @@ if(MLIR_ENABLE_ROCM_RUNNER)
 
   # Locate HIP runtime library.
   find_library(ROCM_RUNTIME_LIBRARY amdhip64
-               PATHS "${HIP_PATH}/lib")
+               PATHS "${HIP_ROOT_DIR}/lib")
```
zhanglx13: This is changed in the current miopen-dialect branch. I cannot see when it happened since github blame does not work now.

zhanglx13 (approved):
LGTM

external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt
krzysz00: Ah, right - I changed the other way in the subtree update and now am moving it back

krzysz00:
(don't merge this, found an issue with it while poking at upstream)

krzysz00:
(I'm calling for another review since this is now basically a different PR)

external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt
```diff
@@ -197,23 +197,8 @@ if(MLIR_ENABLE_ROCM_RUNNER)
       set(ROCM_PATH $ENV{ROCM_PATH} CACHE PATH "Path to which ROCm has been installed")
     endif()
   endif()
-  set(HIP_PATH "${ROCM_PATH}/hip" CACHE PATH "Path to which HIP has been installed")
-  set(CMAKE_MODULE_PATH "${HIP_PATH}/cmake" ${CMAKE_MODULE_PATH})
-  find_package(HIP)
-  if (NOT HIP_FOUND)
-    message(SEND_ERROR "Building mlir with ROCm support requires a working ROCm and HIP install")
-  else()
-    message(STATUS "ROCm HIP version: ${HIP_VERSION}")
```
zhanglx13, line 206: You may still want to keep this message so that we can see the hip version.

Are you sure this is the find_package CONFIG mode? 
I did not find HIPConfig.cmake or HIP-config.cmake or something similar in the HIP project. But I did find FindHIP.cmake. I am really confused by the mode.
Or does ROCm provide hip configuration file?

krzysz00: You're looking for `hip-config.cmake` lowercase. It matters
You'll find it in `/opt/rocm/lib/cmake/hip`

And I can't tell if we get the HIP version from that mode - or at least what variable it sets once we've found HIP

krzysz00: Update: the variable I was looking for is `hip_VERSION`, and I'm adding a print for it internally, but not externally

zhanglx13 (approved):
Cool~

krzysz00:
(Today's very weird thing - the patch work upstream, but doesn't quite behave the extra utility library is present)

PR #782, opened 2022-09-27
[DO NOT SQUASH] Drop legacy memref utilities only used in in_warp_transpose test

krzysz00:
1. Get rid of some functions we don't need to be carrying around anymore in `external/` because they're only used in one test
2. Update said test to use `gpu.*` opserations


pcf000 (approved):
Looks okay from here.

PR #783, opened 2022-09-28
Changed the MIOpen branch to fix-kernel-count-MLIR

manupak:
This commit changes to use a branch in MIOpen that contains
fixes to correctly use get kernel count API.

Additionally, this PR is to verify backward compatibility of MIOpen
changes irrespective of #757

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -24,7 +24,7 @@ void buildMIOpen(String cmakeOpts) {
 }
 
 void getAndBuildMIOpen(String prefixOpt, String cmakeOpts) {
-    git branch: 'rename-MLIR', poll: false,\
+    git branch: 'fix-kernel-count-MLIR', poll: false,\
         url: 'https://github.com/ROCmSoftwarePlatform/MIOpen.git'
     buildMIOpen(cmakeOpts)
 }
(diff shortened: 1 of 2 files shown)


krzysz00:
So
1. This shouldn't be a separate PR, it should be in the same PR as the changes that require the move
2. Having looked at your MIOpen branch:
- In that branch, you shouldn't just change the API because we're dealing with two separate projects that have some minor backwards compatibility requirements.
- Instead, you should bump `MIIR_VERSION_FLAT` and then, in MIOpen, use a `#if` to pick between `MIIR_VERSION_FLAT >= 7` (your new one) and other options

krzysz00:
@jerryyin To make sure I'm stating how this needs to happen correctly

krzysz00:
(that being said, good find on how MIOpen isn't passing the tuning config to get kernel count!)

manupak:
Hi @krzysz00 ,

This PR is just meant to verify the change works without 757 (757 also has the change) and not to be merged.

Im still trying to get my head around what qualifies as a breaking change here : because in reality it is a just a string of arguments is what is passed through the getKernelCount API -- and this changes the contents of that this string.

However, I get the feeling that you are insisting that a structural changes to that string qualifies as a breaking change. Is that correct?

If so, basically based on the rocMLIR API version, we should/ should not pass the perf-config? (even though we know perf_config is ignored in getKernelCount API prior to 757)

krzysz00:
@manupak ... That's on me for losing the context of the patch. I thought, probably because I saw a `mlir::`, that your previous PR involved needing to change the API in `Miir.h`. Since it doesn't, and it just involves passing in the perf_config where we didn't previously, this isn't an API breaking change and I'm sorry for all the confusion.

And thanks for clearing up what this PR is for

manupak:
Ok this passed weekly CI as well -- so I am more confident now this is not a breaking change.

manupak:
I ll work on create a PR against MIOpen next...

jerryyin:
@manupak Does your MIOpen PR needs to land in ROCm 5.4? Since your padding PR refactoring landed after release branch I'm assuming that no problem exists with the ROCm 5.4, correct?

manupak:
@jerryyin yes that is correct -- no need to land in 5.4

PR #784, opened 2022-09-28
Clean up option handling of perf script

krzysz00:
1. Remove the single-dash -> double-dash hack
2. Explicitly use a positional argument to capture the case where you want to benchmark a specific config. This has the consequence that you will need to use -- to separate the config from other options
3. Make -o work again
4. Add --batch_bath in order to allow explicitly selecting the default behavior instead of relying on a lack of args (which would break with -o)
5. Update the Jenkinsfile to the new arguments

### DIFF
--- a/mlir/utils/jenkins/MIOpenDriver.py
+++ b/mlir/utils/jenkins/MIOpenDriver.py
@@ -535,7 +535,7 @@ def main(args=None):
         help="Only tune the MLIR kernels"
     )
     mutex_arg_group.add_argument(
-        "-b", "--batch",
+        "-b", "--batch_mlir",
         action="store_true",
         help="CSV batch benchmarking mode with MLIR"
     )
@@ -591,7 +591,7 @@ def main(args=None):
         if not parsed_args.miopen_build_dir:
             raise RuntimeError("MIOpen build dir was not provided/found where the test requires it")
 
-    if parsed_args.batch or parsed_args.batch_both:
+    if parsed_args.batch_mlir or parsed_args.batch_both:
         if not parsed_args.mlir_build_dir:
             raise RuntimeError("MLIR build dir was not provided/found")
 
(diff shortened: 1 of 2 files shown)


mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -583,14 +574,24 @@ def main(args=None):
         default=find_miopen_build_dir(),
         help="The build directory of MIOpen",
     )
-    parsed_args, _ = parser.parse_known_args(args)
+    parser.add_argument(
+        "config",
+        type=str,
+        nargs='*',
```
manupak, line 580: nice!

mlir/utils/jenkins/MIOpenDriver.py
```diff
@@ -549,15 +535,20 @@ def main(args=None):
         help="Only tune the MLIR kernels"
     )
     mutex_arg_group.add_argument(
-        "-b",
+        "-b", "--batch",
```
manupak: How about --batch_mlir here ? and do we need just the -b because we have three long arguments starting with b, I think if possible we should drop the -b.

manupak (commented):
Thanks for the further cleanup! Mostly looking good --  just a single comment.

mlir/utils/jenkins/MIOpenDriver.py
krzysz00: I got rid of the long arguments that start with `-b` in most cases
But `--batch_mlir` is good, I'll change it

I'd weakly prefer keeping it because `./bin/MIOpenDriver.py -b -o=blah.csv` is the most common manual use case in my experience

yiqian1 (approved):

mlir/utils/jenkins/MIOpenDriver.py
manupak: Ack.

manupak (approved):
LGTM!

PR #785, opened 2022-09-28
Added tests for mixr gemm (dot) fusions e2e checks

manupak:
This commit adds checks to evaluate which mixr gemm-based fusions are working and which are not. Future work is needed to see to fix failing cases.

### DIFF
--- a/mlir/test/fusion/e2e/mixr-gemm/mixr-gemm-bcast-add.e2e.mlir
+++ b/mlir/test/fusion/e2e/mixr-gemm/mixr-gemm-bcast-add.e2e.mlir
@@ -1,0 +1,15 @@
+// RUN: rocmlir-opt -migraphx-to-tosa %s | rocmlir-driver -host-pipeline highlevel | rocmlir-gen -ph -print-results -rand none - | rocmlir-driver -c  | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s
+
+module {
+  // CHECK:  [5,     5,     5],
+  // CHECK-NEXT: [5,     5,     5],
+  // CHECK-NEXT: [5,     5,     5],
+  // CHECK-NEXT: [5,     5,     5],
+  // CHECK-NEXT: [5,     5,     5]
+  func.func @dot_broadast_add(%arg0: tensor<1x5x4xf32>, %arg1: tensor<1x4x3xf32>, %arg2: tensor<3xf32>) -> tensor<1x5x3xf32> attributes{kernel, arch = ""} {
+    %0 = "migraphx.dot"(%arg0, %arg1) : (tensor<1x5x4xf32>, tensor<1x4x3xf32>) -> tensor<1x5x3xf32>
+    %1 = "migraphx.broadcast"(%arg2) {axis = 2:i64, out_lens= [1:i64, 5:i64, 3:i64] } : (tensor<3xf32>)-> tensor<1x5x3xf32>
+    %2 = "migraphx.add"(%0, %1) {} : (tensor<1x5x3xf32>, tensor<1x5x3xf32>)-> tensor<1x5x3xf32>
+    return %2 : tensor<1x5x3xf32>
+  }
+}
--- a/mlir/test/fusion/e2e/mixr-gemm/mixr-gemm-reshape-case2.e2e.mlir
+++ b/mlir/test/fusion/e2e/mixr-gemm/mixr-gemm-reshape-case2.e2e.mlir
@@ -1,0 +1,12 @@
+// RUN: rocmlir-opt -migraphx-to-tosa %s | rocmlir-driver -host-pipeline highlevel | rocmlir-gen -ph -print-results -rand none - | rocmlir-driver -c  | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s
+// XFAIL: *
+module {
+  // CHECK:  [4, 4, 4, 4, 4],
+  // CHECK-NEXT: [4, 4, 4, 4, 4],
+  // CHECK-NEXT: [4, 4, 4, 4, 4]
+  func.func @dot_reshape_2(%arg0: tensor<1x5x4xf32>, %arg1: tensor<1x4x3xf32>, %arg2: tensor<1x5x3xf32>) -> tensor<1x3x5xf32> attributes{kernel, arch = ""} {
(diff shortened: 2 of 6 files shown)


manupak:
closes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/666

manupak:
I have done minimal set of checks to see if the fusion with gemm works and found 1 reshape case it does not.
Let me know if you know any more interesting cases that we should check.

mlir/include/mlir/Dialect/MIGraphX/MIGraphXOps.td
```diff
@@ -282,7 +282,7 @@ def MIGraphX_FlattenOp :
 def MIGraphX_TransposeOp :
     MIGraphX_Op<"transpose">,
     Arguments<(ins AnyRankedTensor:$input,
-                   I64ArrayAttr:$dims
+                   ElementsAttr:$dims
```
krzysz00: Why this charge? You can go for `array<i64: v1, v2, ...>`?

mlir/test/fusion/e2e/mixr-gemm/mixr-gemm-reshape-case2.e2e.mlir
```diff
@@ -0,0 +1,12 @@
+// RUN: rocmlir-opt -migraphx-to-tosa %s | rocmlir-driver -host-pipeline highlevel | rocmlir-gen -ph -print-results -rand none - | rocmlir-driver -c  | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s
+// XFAIL: *
```
manupak, line 2: note to future task : this fails with : 
```
error: 'llvm.call' op 'malloc' does not reference a symbol in the current scope
```

mlir/include/mlir/Dialect/MIGraphX/MIGraphXOps.td
manupak: lemme check

manupak: error: expected attribute value
    %2 = "migraphx.transpose"(%0) {dims = array<i64: 0, 2, 1>} : (tensor<1x5x3xf32>)-> tensor<1x3x5xf32>

manupak: With I64ArrayAttr it only seems to accept dims = [0:i64, 2:i64, 1:i64]

manupak: However, that fails in TOSA conversion 

https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/8b9684b977f6773a7f30f1bcfa2fb344c16fdfa6/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.td#L41-L42

manupak: with 

```
/home/mkarunar/work/llvm-project-mlir/mlir/test/fusion/e2e/mixr-gemm/mixr-gemm-transpose.e2e.mlir:11:10: error: 'arith.constant' op value must be an integer, float, or elements attribute
    %2 = "migraphx.transpose"(%0) {dims = [0:i64, 2:i64, 1:i64]} : (tensor<1x5x3xf32>)-> tensor<1x3x5xf32>
         ^
/home/mkarunar/work/llvm-project-mlir/mlir/test/fusion/e2e/mixr-gemm/mixr-gemm-transpose.e2e.mlir:11:10: note: see current operation: %1 = "arith.constant"() {value = [0, 2, 1]} : () -> none
```

krzysz00: Ah, got it.

I figure our two options here are to either go for
- Using a NativeCodeCall in the rewrites to convert the array attribute to a tensor attribute (that's where I'm leaning); or
- Making that constraints a bit more specific then ElementsAttr - perhaps I64DenleElementsAttr or some such

mlir/test/fusion/e2e/mixr-gemm/mixr-gemm-reshape-case2.e2e.mlir
krzysz00: That's a fusion failure

mlir/include/mlir/Dialect/MIGraphX/MIGraphXOps.td
manupak: Cool! I ll have a quick look at 1) then...

manupak: Done.

jungpark-mlir (approved):
Looks good, thanks!

mlir/test/fusion/e2e/mixr-gemm/mixr-gemm-bcast-add.e2e.mlir
```diff
@@ -0,0 +1,15 @@
+// RUN: rocmlir-opt -migraphx-to-tosa %s | rocmlir-driver -host-pipeline highlevel | rocmlir-gen -ph -print-results -rand none - | rocmlir-driver -c  | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s
```
sjw36, line 1: You should be able to use the new verifier clone like this:

// RUN: rocmlir-opt -migraphx-to-tosa %s | rocmlir-driver -host-pipeline partition,highlevel | rocmlir-gen -ph -rand 1 -verifier clone -fut dot_broadcast_add__part_0 - | rocmlir-driver -c  | mlir-rocm-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %

but you will need to remove the kernel attribute as well.

Not sure how well this works thru mlir-rocm-runner

sjw36 (approved):
Looks good. But if you can convert the tests that would be better.

mlir/test/fusion/e2e/mixr-gemm/mixr-gemm-bcast-add.e2e.mlir
krzysz00: One thing I've been considering doing with adding gemm testing is adding the xmir runner pipeline as an option to `rocmlir-driver` so that we can use xmir stuff in tests without having to pay the compilation overhead inside the "runner"

(and I've also been contemplating making `-c` be "full" for kernel pipelines and stripping out most of mlir-rocm-runner's stuff)

krzysz00: (and I don't think we've had --verifier clone in the tests when this PR landed?)

PR #787, opened 2022-09-29
Move the cuwave related parameters out of tuning parameters attribute

krzysz00:
Instead of deriving the {m,n}{ThreadsPerCuwave,CuwavesPerBlock} attributes during tuning and packing them into the tuning parameters struct when they're not actually tuning parameters, use a function to compute those parameters as needed based on the block size. The new function is identical to the old logic in the tuning parameter pass.

This has the advantage of making the non-xdlops tuning parameters attribute exactly equivalent to the tuning parameter string we get from external clients (possibly mod ordering).

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/Tuning/BlockStructureParams.h
+++ b/mlir/include/mlir/Dialect/Rock/Tuning/BlockStructureParams.h
@@ -1,0 +1,43 @@
+//===- BlockStructureParams.h - Block structure derived parameters ----*-===//
+//
+// Part of the rocMLIR Project, under the Apache License v2.0 with LLVM
+// Exceptions. See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (c) 2022 Advanced Micro Devices INc.
+//===----------------------------------------------------------------------===//
+//
+// This file defines the BlockStructureParams structure, which describes how
+// the workitems of a workgroup are grouped into smaller parts based on.
+//
+//===----------------------------------------------------------------------===//
+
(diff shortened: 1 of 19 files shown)


krzysz00:
Nightly pass! http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-768/12/pipeline/788

mlir/include/mlir/Dialect/Rock/Tuning/BlockStructureParams.h
```diff
@@ -0,0 +1,43 @@
...
+/// This structure defines the following fields, which are all a function of
+/// the block size selected for a kernel
+/// - mThreadsPerCuwave: The length of the m dimension in the grid the 16
+/// threads
+///   of a CU wave are arranged into
+/// - nThreadsPerCuwave: The length of the n dimension of said thread grid
+/// - mCuwavesPerBlock: The length of the m dimension of the grid the CU waves
+///   composing each block is arranged in to
+/// - nCuwavesPerBlock: The n dimension of said grid of CU waves
+struct BlockStructureParams {
```
manupak: I'd prefer to keep the name to have an association to Gemm (as it did before in DerivedBlockGemmParams) . One suggestion might be GemmBlockStructureParams ?

manupak (commented):
Thanks for the changes and I really like the fact that we are unifying the tuning params.

I only had a naming suggestion.

mlir/include/mlir/Dialect/Rock/Tuning/BlockStructureParams.h
krzysz00: @manupak We good now?

manupak (approved):
LGTM!

PR #789, opened 2022-09-29
[DO NOT SQUASH] XMIR phase-3

sjw36:
MLIR#562: added async.launch conversion to CPU coroutine or call
* Changed AsyncToGPU to only convert if a GPU target is present
* Shortened partition naming
* enhanced Partition Pipeline with more TOSA optimizations
  * also moved AsyncLaunch into Partition Pipeline
  * CloneKernels now removes 'kernel' from host func
* updated XMIR pipeline

* added resnet50_blk.cpu.mlir test for async.launch on CPU (no-coroutines)

https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/562


mlir/lib/Dialect/Rock/Pipelines/XMIRPipelines.cpp
```diff
@@ -38,28 +38,40 @@ void xmir::buildModelPipeline(OpPassManager &pm,
...
   pm.addNestedPass<func::FuncOp>(createConvertLinalgToAffineLoopsPass());
   pm.addPass(createLowerAffinePass());
-  pm.addPass(createConvertSCFToCFPass());
-  if (options.cpuOnly) {
+  pm.addNestedPass<func::FuncOp>(createConvertSCFToCFPass());
+
+  // Target async.launch to cpu.coro or gpu.launch_func
+  pm.addPass(createConvertAsyncToGPUPass());
+  pm.addPass(createAsyncParallelForPass());
+  pm.addPass(createAsyncToAsyncRuntimePass(options.enableCoroutines));
```
sjw36: @krzysz00 if you have some time to look at coroutines, you can run xmir/e2e/resnet50_blk.cpu.mlir test but you would need set this parameter to true.

external/llvm-project/mlir/include/mlir/Dialect/Async/Passes.h
```diff
@@ -31,7 +31,8 @@ std::unique_ptr<Pass> createAsyncParallelForPass(bool asyncDispatch,
                                                  int32_t numWorkerThreads,
                                                  int32_t minTaskSize);
 
-std::unique_ptr<OperationPass<ModuleOp>> createAsyncToAsyncRuntimePass();
+std::unique_ptr<OperationPass<ModuleOp>>
+createAsyncToAsyncRuntimePass(bool enableCoroutines = true);
```
krzysz00: Instead of doing this, use the pass option struct that gets automatically generated now

krzysz00: You could also, probably, remove the `let constructor = ...` entirely and, if you wanted to keep back-compat, rename the tablegen definition to `AsyncToAsyncRuntimePass`

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -82,9 +83,8 @@ static Optional<DictionaryAttr> getGPUTarget(async::LaunchOp op) {
 //===----------------------------------------------------------------------===//
 
 namespace {
-class LaunchOpConversion : public OpConversionPattern<async::LaunchOp> {
-public:
-  using OpConversionPattern::OpConversionPattern;
+struct LaunchRewritePattern : public OpRewritePattern<async::LaunchOp> {
```
krzysz00, line 86: Why not a conversion pattern?

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -284,17 +284,21 @@ class LaunchOpConversion : public OpConversionPattern<async::LaunchOp> {
...
-  LogicalResult matchAndRewrite(async::AwaitOp op, OpAdaptor adaptor,
-                                ConversionPatternRewriter &rw) const override {
+  LogicalResult matchAndRewrite(async::AwaitOp op,
+                                PatternRewriter &rw) const override {
     auto tokenType = rw.getType<gpu::AsyncTokenType>();
-    rw.create<gpu::WaitOp>(op.getLoc(), tokenType, adaptor.getOperands());
-    rw.eraseOp(op);
+    Value input = op->getOperand(0);
+    if (input.getType() == tokenType) {
+      assert(op.getResultType() == None);
```
krzysz00, line 296: What's this assert for?

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -284,17 +284,21 @@ class LaunchOpConversion : public OpConversionPattern<async::LaunchOp> {
...
+  using OpRewritePattern<async::AwaitOp>::OpRewritePattern;
 
-  LogicalResult matchAndRewrite(async::AwaitOp op, OpAdaptor adaptor,
-                                ConversionPatternRewriter &rw) const override {
+  LogicalResult matchAndRewrite(async::AwaitOp op,
+                                PatternRewriter &rw) const override {
     auto tokenType = rw.getType<gpu::AsyncTokenType>();
-    rw.create<gpu::WaitOp>(op.getLoc(), tokenType, adaptor.getOperands());
-    rw.eraseOp(op);
+    Value input = op->getOperand(0);
```
krzysz00: Surely there's an accessor for the tokens field of the wait op?

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -312,30 +316,23 @@ void ConvertAsyncToGPUPass::runOnOperation() {
...
-  target.addDynamicallyLegalOp<async::LaunchOp>(
-      [&](async::LaunchOp op) { return !getGPUTarget(op).has_value(); });
-  // TODO(sjw): Make async.token universal
-  // target.addDynamicallyLegalOp<async::AwaitOp>([&](async::AwaitOp op) {
-  //     return true;
-  // });
+  {
+    // Convert async.await to gpu.wait if has gpu.tokens
+    RewritePatternSet patterns(ctx);
+    patterns.add<AwaitRewritePattern>(ctx);
```
krzysz00, line 332: Was it not possible to rely on the launch getting rewritten before the wait?

external/llvm-project/mlir/lib/Dialect/Async/Transforms/AsyncToAsyncRuntime.cpp
```diff
@@ -237,17 +239,66 @@ static Block *setupSetErrorBlock(CoroMachinery &coro) {
   return coro.setError;
 }
 
+/// Build coroutine logic in an outlined function.
```
krzysz00, line 242: This is a note to me to poke at this, work out how it differs from the handling for exec, and trying to figure out the bug

external/llvm-project/mlir/lib/Dialect/Async/Transforms/AsyncToAsyncRuntime.cpp
```diff
@@ -607,6 +703,34 @@ class AssertOpLowering : public OpConversionPattern<cf::AssertOp> {
   llvm::DenseMap<func::FuncOp, CoroMachinery> &outlinedFunctions;
 };
 
+//===----------------------------------------------------------------------===//
```
krzysz00, line 706: Why do we need this? Shouldn't `--canonicalize` do this?

external/llvm-project/mlir/lib/Dialect/Async/Transforms/AsyncToAsyncRuntime.cpp
```diff
@@ -825,6 +958,9 @@ void AsyncToAsyncRuntimePass::runOnOperation() {
   }
 }
 
-std::unique_ptr<OperationPass<ModuleOp>> mlir::createAsyncToAsyncRuntimePass() {
-  return std::make_unique<AsyncToAsyncRuntimePass>();
+std::unique_ptr<OperationPass<ModuleOp>>
+mlir::createAsyncToAsyncRuntimePass(bool enableCoroutines) {
```
krzysz00: Ok, yeah, make this function take the options struct

krzysz00 (commented):
I've got a few questions.

Also, re %chip, we don't have that substitution in our tests, last I checked. You'll need to go copy in the code from `external/` for it

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
sjw36: An async.launch doesn't necessarily convert to gpu.launch, only if it has a gpu target implementation.

The new conversion in async-to-async-runtime will convert any async.launch's that remain.

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
sjw36: async.await should never has a result when the operand is a token

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
sjw36: I think the pattern rewriter runs in reverse order within a block so it would hit the await ops first.

external/llvm-project/mlir/lib/Dialect/Async/Transforms/AsyncToAsyncRuntime.cpp
sjw36: maybe..

external/llvm-project/mlir/lib/Dialect/LLVMIR/IR/LLVMDialect.cpp
```diff
@@ -2805,15 +2805,31 @@ Value mlir::LLVM::createGlobalString(Location loc, OpBuilder &builder,
...
-  auto global = moduleBuilder.create<LLVM::GlobalOp>(
-      loc, type, /*isConstant=*/true, linkage, name,
-      builder.getStringAttr(value), /*alignment=*/0);
 
+  SmallString<64> uniqueName(name);
+
+  // Create the global at the entry of the module.
+  // Uniquify if necessary or return found equivalent.
+  LLVM::GlobalOp global = module.lookupSymbol<LLVM::GlobalOp>(uniqueName);
+  APInt i(32, 0);
```
krzysz00: I'm pretty sure this can be an `uint32_t` and then we can use `Twine(i)`

krzysz00 (commented):
This all seems reasonable, I've got the one nit.

Ideally, I'd like to see what the problem with coroutines is. Is it something that happens outside of a rocMLIR context? Because I'd expect people to want the coroutine option to work.

sjw36:
I have a simple example (https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/789/files#diff-bdf124cc08c68ab347fbf2c5d43fedc19ed43fbc96d0323993f57de115cf2d5e) that converts 2 async.launch's to coroutine and it runs clean. If the launch'd func's have conv2d and lowers the same way thru linalg to cpu, it fails in the LLVM backend because of token type.

external/llvm-project/mlir/lib/Dialect/LLVMIR/IR/LLVMDialect.cpp
sjw36: yeah, it is possible. I will probably follow it up in another PR if this passes CI.

sjw36:
Simplified testcase for coroutine in xmir-runner:

```
module {
  func.func private @resnet50__part_0() attributes {async.targets = []} {
    return
  }
  func.func private @resnet50__part_1() attributes {async.targets = []} {
    return
  }
  func.func @resnet50() {
    %token = async.launch @resnet50__part_0 () : () -> ()
    %token_0 = async.launch @resnet50__part_1 [%token] () : () -> ()
    async.await %token_0 : !async.token
    return
  }
  func.func @main() {
    call @resnet50() : () -> ()
    return
  }
}
```

krzysz00 (approved):
Approved, though if you could clean up the APInt nit while fixing the merge conflict that'd be nice

And we've definitely got some investigating to do on the coroutine issue

PR #792, opened 2022-10-03
[NFC] Remove raw accessors to head off issues at next LLVM update

krzysz00:
Also, take advantage of TypedValue<> in a few places.


manupak (approved):
I skimmed over the changes and I cant think of a reason how this can be successfully built if all callers aren't changed with the removal  of emitAccessorPrefix 

Am I missing something ? :D

krzysz00:
Yeah, I can't for the life of me think how this could break anything given that it compiles either

PR #793, opened 2022-10-04
Fixed reshape to collapse-expand fusion

manupak:
This commit fixes the fusion of reshape
that gets lowered to collapse and expand memref ops.

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/CopyOpt.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/CopyOpt.cpp
@@ -50,6 +50,76 @@ struct RockCopyOptPass
 template <typename T> struct MICORewritePattern : public OpRewritePattern<T> {
   using OpRewritePattern<T>::OpRewritePattern;
 
+  // This function finds the destination-passing style
+  // argument that the kernel will copy into it. It might
+  // encounter view-like ops : CollapseShapeOp or ExpandShapeOp
+  // where the reverse expression need to be built up for
+  // the users of the found destination arg.
+  LogicalResult findCopyDestArg(PatternRewriter &rewriter, Operation *op,
+                                Value &copyDestArg) const {
+    if (auto copyOp = dyn_cast<memref::CopyOp>(op)) {
+      return findCopyDestArg(rewriter, copyOp, copyDestArg);
+    }
+    if (auto collapseOp = dyn_cast<memref::CollapseShapeOp>(op)) {
+      return findCopyDestArg(rewriter, collapseOp, copyDestArg);
+    }
+    if (auto expandOp = dyn_cast<memref::ExpandShapeOp>(op)) {
+      return findCopyDestArg(rewriter, expandOp, copyDestArg);
+    }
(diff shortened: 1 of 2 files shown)


manupak:
closes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/668

mlir/lib/Dialect/Rock/Transforms/CopyOpt.cpp
```diff
@@ -50,6 +50,76 @@ struct RockCopyOptPass
...
+  // argument that the kernel will copy into it. It might
+  // encounter view-like ops : CollapseShapeOp or ExpandShapeOp
+  // where the reverse expression need to be built up for
+  // the users of the found destination arg.
+  LogicalResult findCopyDestArg(PatternRewriter &rewriter, Operation *op,
+                                Value &copyDestArg) const {
+    if (auto copyOp = dyn_cast<memref::CopyOp>(op)) {
+      return findCopyDestArg(rewriter, copyOp, copyDestArg);
+    }
+    if (auto collapseOp = dyn_cast<memref::CollapseShapeOp>(op)) {
```
krzysz00, line 63: As a general thought, would it make sense to have this just go through all `ViewLikeOpInterface`? 

... @jungpark-mlir since that might affect the rest of this file

manupak: @krzysz00 that is how I started on this work..., However, we need to build up the reverse expression from destination arg back to rock kernels -- therefore need to specialize based on collapse or expand anyway.

manupak: The trouble would be to figure out what is the reverse view creation op for every such op...

krzysz00: Ah, yeah, good point

manupak: I think given the current TOSA lowering flow, it felt to me we can rely on canonical form of memref Collapse/Expand ops in the post-bufferization phase. WDYT ?

jungpark-mlir: Just wondering about if there is any case leaving transpose ops out of the linalg.generic but hoping it's all fused before this.

jungpark-mlir (approved):
Looks good to me.

mlir/lib/Dialect/Rock/Transforms/CopyOpt.cpp
manupak: E.g. : https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/miopen-dialect/mlir/test/fusion/tosa-tp-highlevel.mlir.

I didn't see them at this level; so we should be good.

manupak:
Thanks folks for the review!

sjw36:
Great cleanup @manupak !

PR #794, opened 2022-10-04
Create RockGemmWrapperInterface and use it in a few spots.

krzysz00:
This is the first part of cleaning up the tensor_conv2d hack. We'll now have an interface for gemm-like things, which we can extend in the future (so that, for instance, AffixTuningParameters doesn't need to be a template).

The getGemmSize() method is a refactoring, but getOutArgument() is important for implementing bufferization generically over all the conv2d* and gemm operations in a followup commit.


mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -369,6 +369,85 @@ void RockDialect::initialize() {
...
+
+  int64_t y, x, ho, wo, hi, wi, k, c, n, g;
+  y = x = ho = wo = hi = wi = k = c = n = g = 0;
+
+  for (unsigned i = 0; i < filterLayoutAttr.size(); ++i) {
+    auto filterAttr = filterLayoutAttr.getValue()[i].cast<StringAttr>();
+    auto inputAttr = inputLayoutAttr.getValue()[i].cast<StringAttr>();
+    auto outputAttr = outputLayoutAttr.getValue()[i].cast<StringAttr>();
+
+    if (filterAttr.getValue() == "y") {
```
sjw36, line 398: switch?

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -369,6 +369,85 @@ void RockDialect::initialize() {
...
+      x = filterShape[i];
+    } else if (filterAttr.getValue() == "k") {
+      k = filterShape[i];
+    } else if (filterAttr.getValue() == "c") {
+      c = filterShape[i];
+    } else if (filterAttr.getValue() == "g") {
+      g = filterShape[i];
+    }
+
+    if (inputAttr.getValue() == "hi") {
```
sjw36, line 410: also switch

sjw36 (approved):
Nice cleanup.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
krzysz00: Can you switch/case on a string?
And I'm not sure StringSwitch will work because we've got variable assignments in here.

(I was copying existing code from `AffixTuningParams`)

sjw36: Just the leading char.

krzysz00: I'd say it's reasonable to check the exact value here

PR #795, opened 2022-10-05
Stop folding ranks with dim_size=1

jungpark-mlir:
Tosa broadcast requires the same ranked input arguments as a canonical form but this part of TosaToLinalg conversion code breaks the canonicalization and introduces additional reshape op at the early lifetime of linalg.generic. This is causing incremental mismatches between the argument shapes and difficulties in the further lowering path.

This patch keeps the original shape of the arguments by mapping broadcast dims to '0' in the affine map.
Next plan is, using the DropUnitDims pass to canonicalize the shapes after processing linalg.generic so that fusion code can handle that in cleaner fashion.

### DIFF
--- a/external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalg.cpp
+++ b/external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalg.cpp
@@ -635,6 +635,9 @@ elementwiseMatchAndRewriteHelper(Operation *operation,
         newShape.push_back(it.value());
         affineExprs.push_back(
             mlir::getAffineDimExpr(it.index(), rewriter.getContext()));
+      } else if (it.value() == 1) {
+        newShape.push_back(it.value());
+        affineExprs.push_back(getAffineConstantExpr(0, rewriter.getContext()));
       }
     }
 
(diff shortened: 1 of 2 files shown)


external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalg.cpp
```diff
@@ -635,6 +635,9 @@ elementwiseMatchAndRewriteHelper(Operation *operation,
         newShape.push_back(it.value());
         affineExprs.push_back(
             mlir::getAffineDimExpr(it.index(), rewriter.getContext()));
+      } else if (it.value() == 1) {
```
krzysz00, line 638: I'm not sure that sort of special-case fix should be happening? Didn't upstream point out a pass we could use that would clean some of this stuff up?

jungpark-mlir:
I'm fixing the tests.

sjw36:
This should be supported in rock-linalg-align since it could arise in other ways (i.e. new ops and transforms).

jungpark-mlir:
> This should be supported in rock-linalg-align since it could arise in other ways (i.e. new ops and transforms).

Yes, that's the Next plan above describes.
This change is orthogonal to that support, fixes current conversion which breaks the canonical form of Tosa broadcast.

krzysz00 (approved):
I think we can land this, but I'd like you to also land this on `miopen-dialect`.

And to have a plan for how to upstream or remove this patch later.

sjw36 (approved):
Looks good.

PR #796, opened 2022-10-06
Generalize bufferization, getting rid of rock.tensor_conv2d

krzysz00:
1. Change the RockGemmWrapperInterface to return a pointer, not a reference, to the out argument, allowing == to work and matching linalg's interfaces
2. Use the interface to implement bufferization for the ops that wrap around a rock.gemm (including rock.gemm itself).
3. Change conv2d* and gemm to take either tensors or memrefs as arguments, allowing for them to bufferize to themselves (note, this might get slightly funky with workspaces, but we won't be hitting that for quite a while, or ever)
4. Make the minor changes needed to make --conv-to-gemm and friends mostly work on tensors or memrefs as future-proofing - in any case, bufferization has to happen before --gemm-to-gridwise

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/IR/RockGemmWrapperInterface.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockGemmWrapperInterface.td
@@ -33,7 +33,7 @@ def RockGemmWrapperInterface : OpInterface<"RockGemmWrapperInterface"> {
           Return the OpOperand that corresponds to the operand argument
           that corresponds to the output result of the operation.
         }],
-        /*retType=*/"OpOperand &",
+        /*retType=*/"OpOperand *",
         /*methodName=*/"getOutArgument",
         /*args=*/(ins),
         /*methodBody=*/"",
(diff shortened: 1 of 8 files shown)


mlir/include/mlir/Dialect/Rock/IR/RockGemmWrapperInterface.td
```diff
@@ -33,7 +33,7 @@ def RockGemmWrapperInterface : OpInterface<"RockGemmWrapperInterface"> {
           Return the OpOperand that corresponds to the operand argument
           that corresponds to the output result of the operation.
         }],
-        /*retType=*/"OpOperand &",
+        /*retType=*/"OpOperand *",
```
sjw36, line 36: Can you just return an index? Returning pointers seems fragile.

sjw36 (changes requested):
Good cleanup. But can you avoid returning pointers. If not I will approve.

mlir/include/mlir/Dialect/Rock/IR/RockGemmWrapperInterface.td
krzysz00: I could, but then all the clients would need the much more verbose invocation `&op->getOpOperand(cop.getOutArgument())`, which still summons up the pointer.

I'm not too concerned about the fragility. If you go peek at `LinalgStructuredInterface` in `mlir/include/mlir/Dialect/Linalg/IR/LinalgInterfaces.td` , which is used in linalg bufferization, they return `OpOperand *` or SmallVector<OpOperand *>` all over the place. 

I'm pretty sure that, for a given `Operation`, its `OpOperand`s are always in the same place unless someone mutates the op in place (which, for bufferization, could only happen in `bufferize()`)

sjw36: Yeah, still don't like it but won't hold it up.

sjw36 (approved):
Approved.

mlir/include/mlir/Dialect/Rock/IR/RockGemmWrapperInterface.td
krzysz00: This is reasonable for the same reason passing `Value` by value is reasonable: in both cases, you're ultimately passing a pointer to something that's in a use-def list that's immutable. (OpOperand just doesn't have those nice wrappers around it)

PR #798, opened 2022-10-06
Revert Dockerfile change from MIOpen to Rock

sjw36:
I think this somehow got changed in rebranding.. But changing back to MIOpen still fails to build the docker image with:

> clang: error: unsupported option '--offload-arch=gfx900'

### DIFF
--- a/mlir/utils/jenkins/Dockerfile
+++ b/mlir/utils/jenkins/Dockerfile
@@ -120,11 +120,11 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
   apt-get clean && \
   rm -rf /var/lib/apt/lists/*
 
-# --------------------- Section 3: Rock dependencies ---------------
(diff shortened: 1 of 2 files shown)


sjw36:
@jerryyin any idea what is causing the new failure?

jerryyin (approved):
Nice catch.

As a FYI, this dockerfile change won't be captured and tested by this PR, (instead it will be effective only after this PR is merged and new docker built after that). Therefore the test result of this PR is irrelevant. I think you should just re-kick to make flaky test correct itself.

sjw36:
@jerryyin still doesn't build because of MIOpen build clang error.

jerryyin:
@sjw36 Took a look and it turns out that renaming break the auto docker build job. I'm on my way to fix that.

sjw36:
Thanks @jerryyin

PR #799, opened 2022-10-07
Update e2e padding perf_config test to use KPack=1

manupak:
This commit changes a missed test to use KPack=1
where a padding kernel is used until it is resolved.


krzysz00 (approved):
Approved, and I also approve cheating the CI on this one (that is, kicking off "CI runs" whene you uncheck both `sharedLib` and `staticLib` to disable all the tests)

PR #800, opened 2022-10-07
Align mixr conv2d lowering to TOSA to Linalg flow

manupak:
This commit changes to TOSA lowering to
make it runnable in a TOSA to Linalg flow.

Additionally this commit adds a multibroadcast
+ {mbcast,conv2d,add,relu} fusion test case.

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/687

### DIFF
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -16,6 +16,7 @@
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/MIGraphX/MIGraphXOps.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/Dialect/Tosa/IR/TosaOps.h"
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Transforms/DialectConversion.h"
@@ -61,12 +62,13 @@ class ConvConverter final
 public:
   using OpConversionPattern<migraphx::ConvolutionOp>::OpConversionPattern;
 
-  Value getZero(Location loc, Type elemType,
-                ConversionPatternRewriter &rewriter) const {
-    auto biasTy = RankedTensorType::get({1}, elemType);
-    auto zeroAttr =
-        DenseElementsAttr::get(biasTy, rewriter.getZeroAttr(elemType));
-    return rewriter.create<arith::ConstantOp>(loc, zeroAttr);
+  Value getZeroBias(Location loc, Type elemType, int64_t filterOutputChannels,
+                    ConversionPatternRewriter &rewriter) const {
+    auto biasTy = RankedTensorType::get({filterOutputChannels}, elemType);
+    auto arithZero =
+        rewriter.create<arith::ConstantOp>(loc, rewriter.getZeroAttr(elemType));
+    auto biasTensor = rewriter.create<tensor::SplatOp>(loc, biasTy, arithZero);
+    return biasTensor;
   }
 
   tosa::TransposeOp getRank4TransposeOp(Location loc, Value input,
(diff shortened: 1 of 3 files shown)


manupak:
I actually wanted a way to test this fusion case and ended up doing something else as well (to get references for the test).

Also I did not know a better way (easier) way to check against a CPU reference and this seems to work fine.

sjw36:
@manupak we just added a flow to fully validate fusions on gpu vs cpu flow. Unfortunately I accidentally reverted the one test that exercised the new flow.. @pcf000 is reverting it and adding a few more. Talk to him for an example.

The new XMIR flow goes from a TOSA model, so you will need to convert from MIXR to TOSA first.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -61,12 +62,13 @@ class ConvConverter final
 public:
   using OpConversionPattern<migraphx::ConvolutionOp>::OpConversionPattern;
 
-  Value getZero(Location loc, Type elemType,
-                ConversionPatternRewriter &rewriter) const {
-    auto biasTy = RankedTensorType::get({1}, elemType);
-    auto zeroAttr =
-        DenseElementsAttr::get(biasTy, rewriter.getZeroAttr(elemType));
-    return rewriter.create<arith::ConstantOp>(loc, zeroAttr);
+  Value getZeroBias(Location loc, Type elemType, int64_t filterOutputChannels,
```
jungpark-mlir, line 65: Is this change necessary? this is an WA to seal the gap between MIGraphX and tosa conv2d where MIGraphX doesn't get bias as an argument but always converts to broadcast_bias+bias tensor.
I'd like to have it as a minimal code in IR.

manupak: The current TOSA that is being created only works iff if its lowered to Rock dialect. i.e. migraphx to tosa to linalg flow is otherwise broken. (Not sure we care about that)

Alternatively we need to fix the TosaToLinalg but I feel we should do that upstream directly.

In conclusion, its not needed for the rock path as of today but it feels like we are creating TOSA that would work only for our flow. WDYT?

manupak:
@sjw36 sorry for the confusion with the title, I've change it to reflect more correctly the changes here.
It is not exercising a CPU flow rather a TOSA to Linalg conversion to obtain the references to verify the rock conversion + fusion is correct. The references also runs in the GPU (looking at the lowering).

I tried using the cloning functionality (which is essential what I've done : created a clone of the fut) but I understood that it does not support kernel generation, unless Im mistaken.

https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/758bae64f34c3d3b6acb673c6a46eaf08cc67725/mlir/tools/rocmlir-gen/rocmlir-gen.cpp#L2215-L2218.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jungpark-mlir: I see, 
I thought tosa.conv2d allows 1-element bias but maybe not, or just TosaToLinalg as you commented.
Just checked the spec and it says 'per output channel', so this makes better sense.

jungpark-mlir (approved):

manupak:
@pcf000 would you be able to have a look at the test ? I dont think there is a way to generate validation harness for kernel (rock) generation, but this is from my scan of that part of the code.

pcf000:
> I tried using the cloning functionality (which is essential what I've done : created a clone of the fut) but I understood that it does not support kernel generation, unless Im mistaken.

That's correct.  It depends on the structure created by the CloneKernels pass:  a kernel plus target-specific submodules duplicating the kernel plus a calling function.  I haven't looked at your test yet;  the question will be whether it fits that pattern, or if perhaps it's a pattern that we want to extend support to.

manupak:
Thanks @pcf000!

I did try and hit the error I have showed there. 
I agree with your viewpoint might be something that we might want extend in future (maybe first as GPU reference and then CPU reference)

But as for now, I feel this is sufficient to test whether fused kernel lowering is done successful ly and also produces the right output (trusting the upstream TOSA lowering via Linalg)

pcf000:
The test is similar to tests in fusion/e2e that I've made use "--verifier clone" in a pull request (with "-host-pipeline partition"), so it might be one we can add to later.  As it happens, though, mixr-bcast-add.e2e.mlir is a test that has troubles with that strategy and maybe this one will too.

PR #801, opened 2022-10-07
Rearrange utility scripts, add rocm-run and anonymize_regs.pl

krzysz00:
1. Move the tuning script to `mlir/utils/widgets/`
2. Add `rocm-run`, which finds `mlir-rocm-runner` and sets up the correct shared library arguments, so I don't have to keep passing it around to everyone
3. Add `anonymize-regs.pl`, a utility helpful when taking diffs over ISA to clean up noise from the register allocator.


jerryyin (approved):

zhanglx13 (approved):
1. I think it will be helpful if we can put some README here to introduce the usage of these tools, especially when we want to use them.
2. Can we or do we want to use `rocm-run` in out E2E tests so that when one of the external libraries change, we do not have to modify every test. (However, maybe it makes more sense to achieve this by using the auto-generation framework for E2E tests)

krzysz00:
I can say that we don't quite want to use the script in e2e tests, given that we want the tests to produce nice copy-paste-able output

PR #804, opened 2022-10-07
Make --rock-affix-params work for rock.gemm

krzysz00:
1. Make most of the tuning code take a RockGemmWrapperInterface instance and not a ConvContext
2. Extend RockGemmWrapperInterface to allow for that
3. Rename GemmContext to GemmSize and make it carry the g dimension as well because I can't believe I didn't do that
1. Create the .cpp file for the RockGemmWrapperInterface I'd forgotten about
5. Do all the reorginazation needed to make this work


mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -618,7 +549,8 @@ LogicalResult PopulateParamsXDL::obtainTuningParameters(
 }
 
 std::vector<InitParamsXDL>
-PopulateParamsXDL::getTuningParameters(ConvOpType dir, Type dataType) const {
+PopulateParamsXDL::getTuningParameters(Optional<ConvOpType> dir,
```
manupak, line 552: I think we can drop dir -- it is not being used.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -274,7 +204,8 @@ LogicalResult PopulateParams::obtainTuningParameters(
 }
 
 std::vector<InitParamsNonXDL>
-PopulateParams::getTuningParameters(ConvOpType dir, Type dataType) const {
+PopulateParams::getTuningParameters(Optional<ConvOpType> dir,
```
manupak, line 207: Here also, we can drop the dir as it is not being used.

mlir/lib/Dialect/Rock/Tuning/ConvContext.cpp
```diff
@@ -46,7 +47,7 @@ ConvolutionDims ConvolutionContext::getConvDims() {
 }
 
 ConvolutionContext mlir::rock::populateConvContext(Operation *op) {
-  ConvOpType opType = obtainConvDirection(op);
+  ConvOpType opType = *obtainConvDirection(op);
```
manupak, line 50: I think we dont need this opType as part of the ConvContext after this change.
My quick investigation suggests that only uses of populeConvContext remains in ConvToGemm (after this PR) where the caller knows the direction and even more I dont see ctx.opType is being used.

manupak (commented):
Changes looks good overall.

My only comment here stems from the fact we dont need to pass around a None coming from obtain obtainConvDirection(...).

To elaborate further, I think after the change populateConvContext gets limited to ConvToGemm (rightfully so...). In there, all the callers know the direction and I dont think ctx.opType is ever used. Therefore, I think we can safely remove opType from ConvContext which removes the need/scope for passing around a None in the case of rock.gemm.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
krzysz00: I left it in there because we might eventually have cases where backward data etc. have different sets of good heuristic parameters - we had those before, IIRC.

krzysz00:
@jerryyin Re having the convolution direction passed in to the "get heuristic parameter list" function ... do you think we should drop that parameter or leave it in just in case?

mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
```diff
@@ -357,14 +357,66 @@ Type Conv2dGenerator::getDataType(OpBuilder &builder) const {
...
+// we set gemmMExtra be 64 so (gemmM+gemmMExtra)%gemmMPerBlock=0.
+//
+// If padding is needed, returns a GemmSize containing the number of elements
+// needed to pad the M, N, and K dimensions (**not** the new gemm size).
+// Otherwise, returns None
+template <typename T>
+static Optional<GemmSize>
+calculatePaddingKernelSize(GemmSize gemmSize, ConvOpType dir, Type dataType,
+                           T populateParams) {
+  bool needExtraPad = false;
```
jerryyin, line 371: How are you able to move template definition into a cpp without running into linker issues? I'm a bit surprised that ... Unless this is used only inside of this cpp but not anywhere else. Is that the case?

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
manupak: How about we pass the Op directly instead as we are expanding this pass to rock.gemm now ?

krzysz00: I'd love to do that but `Conv2DGenerator` calls this for the backward weight padding kernel stuff

manupak: Hmmm.. I thought getTuningParameters is only called under affix params pass.

Just to clarify, my suggestion was to replace the dir with the Op (that already captures each convop corresponding to the direction as well as gemm op) here, assuming the clients of this function within affix params pass.

IIRC, there is a different function to obtain kernel counts, workspace, etc that also scrolls these things but it was not getTuningParameters

krzysz00: The call is in `calculatePaddingKernelSize()` in `Conv2dGenerator::needExtraPadBwdWeight`

If it weren't for that one call, I would've made this take the op just like the rest of the system

manupak: ah yes... I remember now.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
jerryyin: Looks like we can indeed drop the direction. But I'm catching up here... we used to initialize that heuristic parameters based different direction as well. I seem to notice that `orderInitPrams()` does its job on deciding based on amount of padding. Do we do any special ordering based on conv direction as well?

Thinking out loud: It might be that we used a consistent set of heuristic init parameters against all 3 conv directions, which is how you are able to simplify the logic here?

An orthogonal topic, it looks like xdlops and nonxdlops has some unevenness on int8, and I suspect we'd need to bring over a different set of tuning parameters on int8 nonxdlops as well. It may be a rocm x.x ticket.

jerryyin (approved):

mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
krzysz00: I moved it to the only file that used it

krzysz00:
@jerryyin We don't - and haven't for a while - done any selection based on forward/bwd data/bwd weight. We use the same set of heuristic parameters for every direction, and only switch sets based on data type.

We don't necessarily need to do that: we could have different parameter sets for different ops. The question for this PR is "do we keep that unused parameter (op direction) in the API to get the heuristic list?"

jerryyin:
> do we keep that unused parameter (op direction) in the API to get the heuristic list?

I don't think it matters. It is very easy to add the direction back, so we can remove the parameter and add it back later. 

Part of the reason, as I can see that we want to keep it is when we need different set of tuning parameter for different types, like int8.

I don't see this blocking this PR so approved it. I'll leave it to you to decide.

krzysz00:
Given that this is a rather important PR and that we're liable to refactor all this anyway, I'm fine keeping it as is

PR #806, opened 2022-10-10
Removing CK and its dependency

jerryyin:
Context: auto builds dockers continue to fail at CK building step:
 - Bumping the MIOpen branch back to develop as all pending PRs are merged
 - Explicitly removing miopen-CK backend as it depends on customized compiler versions

### DIFF
--- a/mlir/utils/jenkins/Dockerfile
+++ b/mlir/utils/jenkins/Dockerfile
@@ -125,6 +125,7 @@ WORKDIR /MIOpenDepsWork
 RUN wget https://raw.githubusercontent.com/ROCmSoftwarePlatform/MIOpen/develop/requirements.txt \
     && wget https://raw.githubusercontent.com/ROCmSoftwarePlatform/MIOpen/develop/install_deps.cmake \
     && sed -i -e '/rocMLIR/d' requirements.txt \
+    && sed -i -e '/composable_kernel/d' requirements.txt \
     && cmake -P install_deps.cmake --prefix /usr/local
 RUN echo $(git ls-remote https://github.com/ROCmSoftwarePlatform/MIOpen HEAD) > miopen-deps-commit-hash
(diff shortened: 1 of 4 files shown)


krzysz00 (approved):
LGTM

krzysz00:
@jerryyin @manupak - this PR is going out because Manupa's change that passes in perf config for kernel count landed, right?

jerryyin:
Yes, refer to https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1801

@manupak FYI, I have to do it in the PR to rule out the possibility of dependency mismatch between your branch and latest develop.

manupak:
Ah ok -- I ll prolly close this after : https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/805

jerryyin:
@manupak No I just reverted my change on bumping the pointer. Since dockerfile will not be built after a PR is merged, it looks like both our PRs are necessary, and mine has to be merge before yours so things can build for MIOpen.

zhanglx13:
So is CK removed in the image currently used by jenkins?
Re ck building failure for our docker image: is removing CK a temp workaround or we are good to test MIOpen without CK?

jerryyin:
@zhanglx13 Currently, no. I think an out-of-date version of CK has been included in existing docker for a long time. The updated CK has additional requirement on non-released compiler, if not handled well can backfire on us.

> is removing CK a temp workaround or we are good to test MIOpen without CK?

My opinion that we carrying a moving pointer of CK is too much of a maintenance effort (because that is completely orthogonal library). So:
 - We may never want to actively track ongoing development of CK (at least not till their integration are stable enough)
 - Only likely case CK still usefull is the case where we want perf number, but we may only want to track a previously released version of MIOpen + CK combination to reduce our effort in maintaining third party software.

PR #807, opened 2022-10-10
[DO NOT SQUASH] Propagate architecture name throughout the compilation process

krzysz00:
1. Rename a lot of "chip" to "arch" to indicate that it's the full
architecture name and not just the chipset
2. Set kernel.arch on the top-level module when generating a single
kernel for compilation
3. Look for kernel.arch when running backend pipelines to avoid system
device lookups
4. Remove the -triple, -target/-chip, and -features options in favor
of -arch
5. In general, when storing the architecture name somewhere,
canonicazize it to triple:chipset:[features how the backend expects
them]
6. Make the sanity tests set a concrete arch value to avoid weird
conflicts.

---

The system devices are no longer a StringMap of chip -> device,
because that's not sufficient to know if a device is usable. Instead,
they now have a vector of pointers into the device lists that were
produced by each detector.

And now there is a find() method that takes a partial architecture
name and reports if we have it on the system.

In addition, features are now stored for all targets as a
StringMap<bool> so we don't have to deal with the varying string
semantics as much.

In unresolved issues that are a @sjw36 @jungpark-mlir : in the MIGraphX flow, where should we be canonicalizing the architecture name?


pcf000 (approved):
Looks okay to me, but I do see a couple of places where it'll conflict with my current pull request (because I add some RUN lines that'll need to be updated).  Whichever one goes in second will need to update.

Second opinions always welcome.

external/llvm-project/mlir/include/mlir/ExecutionEngine/SystemDevices.h
```diff
@@ -28,15 +31,17 @@ struct SystemDevice {
...
+  llvm::StringMap<bool> features;
   uint32_t count = 1;
   llvm::StringMap<llvm::SmallString<8>> properties = {};
 };
 
 /// SystemDevices captures devices for the current system.
 /// - singleton - map references SystemDevice objects from device
 ///   singletons
-class SystemDevices : public llvm::StringMap<const SystemDevice &> {
+class SystemDevices {
```
sjw36, line 44: Hey, I saw that..

external/llvm-project/mlir/include/mlir/ExecutionEngine/SystemDevices.h
```diff
@@ -45,6 +50,9 @@ class SystemDevices : public llvm::StringMap<const SystemDevice &> {
     return s_systemDevices;
   }
 
+  FailureOr<const SystemDevice *const>
+  find(SystemDevice::Type type, llvm::StringRef partialArchName) const;
```
sjw36, line 54: The find interface is good, more restrictive but since it is not obvious that the object is always const, this is cleaner.

external/llvm-project/mlir/include/mlir/ExecutionEngine/SystemDevices.h
```diff
@@ -71,6 +79,8 @@ class SystemDevices : public llvm::StringMap<const SystemDevice &> {
   }
   // terminus
   void registerSystems(types<>) {}
+
+  llvm::SmallVector<const SystemDevice *, 0> devices;
```
sjw36, line 83: Should at least have static size=2 since there will always be a CPU and probably a GPU.

external/llvm-project/mlir/include/mlir/ExecutionEngine/SystemDevices.h
```diff
@@ -28,15 +31,17 @@ struct SystemDevice {
   enum class Type : int32_t { ECPU, EGPU, ENPU, EALT };
 
   Type type;
-  llvm::SmallString<32> chip;
+  llvm::SmallString<32> llvmTriple;
+  llvm::SmallString<8> chip;
+  llvm::StringMap<bool> features;
```
sjw36, line 36: Good generalization.

external/llvm-project/mlir/lib/ExecutionEngine/CpuSystemDetect.cpp
```diff
@@ -30,22 +30,18 @@ using namespace mlir;
...
-                   count,
-                   {{"llvm_triple", llvm::StringRef(targetTriple)}}};
+  SystemDevice dev{
+      SystemDevice::Type::ECPU, targetTriple, cpu, features, count, {}};
 
-  llvm::StringMap<bool> hostFeatures;
-  if (llvm::sys::getHostCPUFeatures(hostFeatures)) {
-    for (auto &f : hostFeatures) {
-      llvm::StringRef val(f.second ? "1" : "0");
-      dev.properties.insert({f.first().str(), val});
```
sjw36, line 47: I know we aren't using this but let's keep it for now.

external/llvm-project/mlir/lib/ExecutionEngine/SystemDevices.cpp
```diff
@@ -28,12 +33,67 @@ static const char *getDeviceTypeStr(SystemDevice::Type type) {
...
+    if (!feature.empty()) {
+      features.insert_or_assign(feature.drop_back(), feature.back() == '+');
+    }
+  }
+
+  for (const SystemDevice *const device : devices) {
+    bool matches = (type == device->type) &&
+                   (maybeTriple.empty() || maybeTriple == device->llvmTriple) &&
+                   (chip.empty() || chip == device->chip);
+    if (matches && !features.empty()) {
```
sjw36, line 67: What about when features.empty() but device->features is not empty.

mlir/lib/CAPI/Dialect/MIGraphX.cpp
```diff
@@ -131,6 +131,6 @@ MLIR_CAPI_EXPORTED void mlirMIGraphXAddBackendPipeline(MlirPassManager pm,
   opts.chip = chip;
   opts.features = features;
   opts.optLevel = 3;
-  opts.indexBitwidth = 64;
+  opts.indexBitwidth = 32;
```
sjw36, line 134: @jungpark-mlir please confirm this. I think MIGraphX expects 64 bit for index params.

external/llvm-project/mlir/lib/ExecutionEngine/RocmDeviceName.cpp
```diff
@@ -69,11 +67,22 @@ LogicalResult RocmDeviceName::parse(StringRef devName) {
...
-SmallString<256> RocmDeviceName::getFullName() const {
-  SmallString<256> fullName = chip.str();
-  if (features.size()) {
-    fullName += ":";
-    fullName += features;
+std::string RocmDeviceName::getFeaturesForBackend() const {
+  std::string result;
+  llvm::raw_string_ostream stream(result);
+  llvm::interleave(
+      features, stream,
```
sjw36: Is StringMap alphabetical sorted when iterated. May be required in canonical form. https://www.llvm.org/docs/AMDGPUUsage.html#amdgpu-target-id

external/llvm-project/mlir/lib/ExecutionEngine/SystemDevices.cpp
sjw36: Looks like when features are not set, the object supports 'any' value. So this does not need to be checked on the device.

sjw36 (changes requested):
Great cleanup. Just keep the cpu attributes and I will unblock.

external/llvm-project/mlir/include/mlir/ExecutionEngine/SystemDevices.h
krzysz00: I went for <0> because that's the common pattern for forcing a heap allocation in SmallVector

external/llvm-project/mlir/lib/ExecutionEngine/CpuSystemDetect.cpp
krzysz00: ... it's in the `features` attribute of the CPU device. So we haven't lost any info.

external/llvm-project/mlir/lib/ExecutionEngine/SystemDevices.cpp
krzysz00: Yeah. The check is "if I specified features, then the compiled binary must also have those features with those values"

mlir/lib/CAPI/Dialect/MIGraphX.cpp
krzysz00: Our kernel generation expects index = 32

krzysz00: and we use it everywhere else. Also, there are no index params

external/llvm-project/mlir/lib/ExecutionEngine/CpuSystemDetect.cpp
sjw36: Oh, how did I miss that.. makes perfect sense.

mlir/lib/CAPI/Dialect/MIGraphX.cpp
sjw36: what about for memref sizes and strides.

krzysz00: Bare pointer ABI

krzysz00: (and yes, we use "index = [u]int32_t" for those inside all our kernels

external/llvm-project/mlir/lib/ExecutionEngine/RocmDeviceName.cpp
krzysz00: Good catch, am fixing

mlir/lib/CAPI/Dialect/MIGraphX.cpp
sjw36: @jungpark-mlir please confirm that this change is correct.

krzysz00: (For extra context, https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1333/files, which has landed and is enabled, makes the only parameters to our functions pointers, so this change can't affect the API surface)

sjw36: Ok, good good.

sjw36: Do we detect when indexing may overflow on large conv2d? It should cause compile failure.

krzysz00: We do, but it's currently sitting in `Conv2DGenerator`. I noticed that this morning when Jungwook asked where the check was and have it as a ticket to move it somewhere else, probably on `GemmOp::verify()`

sjw36: Sounds good.

sjw36 (approved):
Approving with follow-up ticket to detect when index overflows.

PR #808, opened 2022-10-10
Add "--verifier clone" testing to many fusion/e2e tests

pcf000:
This is most of issue #683, I think.  Two tosa-to-rock tests aren't included because they don't work yet, and two mixr-gemm tests have different answers when partitioning and that's weird.


mlir/test/fusion/e2e/mixr-gemm/mixr-gemm-add.e2e.mlir
```diff
@@ -1,11 +1,20 @@
...
+// RUN: rocmlir-opt -migraphx-to-tosa %s | rocmlir-driver -host-pipeline partition,highlevel -targets gfx908,gfx90a | rocmlir-gen -ph -print-results -rand none -fut dot_add --verifier clone - | rocmlir-driver -host-pipeline xmodel -kernel-pipeline full -triple amdgcn-amd-amdhsa -targets gfx908,gfx90a | xmir-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CLONE
 module {
   // CHECK:  [5,     5,     5],
   // CHECK-NEXT: [5,     5,     5],
   // CHECK-NEXT: [5,     5,     5],
   // CHECK-NEXT: [5,     5,     5],
   // CHECK-NEXT: [5,     5,     5]
+
+  // CLONE: [1 1 1]
+  // CLONE-NEXT: Unranked Memref base
```
sjw36, line 11: Let's use '-rand 1 -rand_type float' for all these so we are sure it is working. And you can just check for '[1 1 1]' then.

sjw36 (changes requested):
See comment.

mlir/test/fusion/e2e/mixr-gemm/mixr-gemm-add.e2e.mlir
pcf000: Done, for all but tosa-to-rock-rsqrt.e2e.mlir.  For that one, the new random options caused all the result values to be NaN, I think because intermediate values were negative.  Left it with "-rand none".

krzysz00:
For the sqrt, it might be a good idea to go with `-rand fixed`

sjw36 (approved):
Looks good.

PR #809, opened 2022-10-11
Register MLIRContext CL Options

manupak:
This commit enhances rocmlir-driver and rocmlir-gen to support MLIRContext CL options.

### DIFF
--- a/mlir/test/rocmlir-driver/sanity.mlir
+++ b/mlir/test/rocmlir-driver/sanity.mlir
@@ -81,3 +81,7 @@
 // RUN: rocmlir-gen --arch gfx900 -mfma=off -atomic_add=off -dot=off -p -t i8 | rocmlir-opt -rock-affix-params -rock-conv-to-gemm -rock-gemm-to-gridwise -rock-gridwise-gemm-to-blockwise -rock-blockwise-gemm-to-threadwise -rock-threadwise-gemm-lowering -rock-sugar-to-loops -rock-clean-math -rock-loops-to-cf -convert-rock-to-gpu "-convert-gpu-to-rocdl=chipset=gfx900 index-bitwidth=32" | rocmlir-translate -gpu-module-to-rocdlir | opt -passes='default<O3>,strip' -S | llc -mcpu=gfx900
 // RUN: rocmlir-gen --arch gfx900 -mfma=off -atomic_add=off -dot=off -p -t i8 | rocmlir-driver -kernel-pipeline=gpu,rocdl --arch=gfx900 | rocmlir-translate -gpu-module-to-rocdlir | opt -passes='default<O3>,strip' -S | llc -mcpu=gfx900
 // RUN: rocmlir-gen --arch gfx900 -mfma=off -atomic_add=off -dot=off -p -t i8 --operation conv2d | rocmlir-driver -kernel-pipeline=gpu,rocdl --arch=gfx900 | rocmlir-translate -gpu-module-to-rocdlir | opt -passes='default<O3>,strip' -S | llc -mcpu=gfx900
+
+// Check for MLIRContext options
+
+// RUN: rocmlir-gen --arch gfx900 -mfma=off -atomic_add=off -dot=off -p -t i8 --operation conv2d --mlir-disable-threading | rocmlir-driver -kernel-pipeline=gpu,rocdl --arch=gfx900 --mlir-disable-threading | rocmlir-translate -gpu-module-to-rocdlir --mlir-disable-threading | opt -passes='default<O3>,strip' -S | llc -mcpu=gfx900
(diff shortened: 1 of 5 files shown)


krzysz00 (approved):

pcf000 (changes requested):
This is good;  I have a branch doing the same but not as thoroughly.  You could also do it to mlir-rocm-runner and xmir-runner, since they run passes before executing.  I find myself wanting --mlir-disable-threading there.

manupak:
@pcf000 I didn't see ParseCommandLineOptions are called there and inferred they dont need that because they are not technically compiler drivers. Am I wrong here ?

Update : they all call JitRunner -- do we want that there (as in before they call JitRunner) as well ?

pcf000:
I had put the registerMLIRContextCLOptions() call in main();  my only hesitation about putting it in JitRunnerMain() is that that's a main-line function instead of one of ours.

krzysz00:
I think the argument parsing should be in main-main

And I'd go looking at how `mlir-cpu-runner` does it

manupak:
https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/miopen-dialect/external/llvm-project/mlir/tools/mlir-cpu-runner/mlir-cpu-runner.cpp

Its very minimal there...

manupak:
I can put the registration before our runners invoke JitRunner -- I was not sure about the scope of what a runner should do in our case

krzysz00:
I agree with putting the registration right before we call JitRunner::main()

manupak:
@pcf000 I have added context CL options to them all. PTAL when you have some time...

mlir/test/rocmlir-driver/sanity.mlir
```diff
@@ -81,3 +81,7 @@
...
 // RUN: rocmlir-gen --arch %arch -mfma=off -atomic_add=off -dot=off -p -t i8 | rocmlir-driver -kernel-pipeline=gpu,rocdl -target=gfx900 | rocmlir-translate -gpu-module-to-rocdlir | opt -passes='default<O3>,strip' -S | llc -mcpu=gfx900
 // RUN: rocmlir-gen --arch %arch -mfma=off -atomic_add=off -dot=off -p -t i8 --operation conv2d | rocmlir-driver -kernel-pipeline=gpu,rocdl -target=gfx900 | rocmlir-translate -gpu-module-to-rocdlir | opt -passes='default<O3>,strip' -S | llc -mcpu=gfx900
+
+// Check for MLIRContext options
+
+// RUN: rocmlir-gen --arch %arch -mfma=off -atomic_add=off -dot=off -p --mlir-disable-threading | rocmlir-driver -kernel-pipeline=gpu,rocdl -target=gfx900 --mlir-disable-threading | rocmlir-translate -gpu-module-to-rocdlir --mlir-disable-threading | opt -passes='default<O3>,strip' -S | llc -mcpu=gfx900
```
krzysz00: I don't think we need to test this part of upstream's functionality - same goes for the other tests

manupak: I generally would like to see a test fail if functionality that we wanted removed... this would for example catch if MLIR context creation above the parsing of CLI parsing. Is there a downside of having a simple test ?

pcf000 (approved):

mlir/test/rocmlir-driver/sanity.mlir
krzysz00: I suppose there's not much downside, I'm just worried about the perf implications ... but it's only three tests, so it's probably fine.

krzysz00:
I'm going to skip-ci this until Manupa gets back, especially since I flubbed the merge conflict resolution

mlir/tools/rocmlir-driver/rocmlir-driver.cpp
```diff
@@ -370,9 +370,9 @@ int main(int argc, char **argv) {
 #endif
   MLIRContext context(registry);
   context
-      .loadDialect<xmodel::XModelDialect, rock::RockDialect, func::FuncDialect,
```
krzysz00: Did you mean to get rid of the `xmodel::` load?

krzysz00: Like, given that CI passed, it's probably fine and we can merge, but I thought I'd check

manupak: good catch; a rebasing artifact.

manupak: I've changed it now.. 

Putting on a TDD hat, I would agree with you but I should not be doing that in this PR :).

manupak:
taking this in now....

PR #810, opened 2022-10-11
Make TosaToRock use rock.gemm

krzysz00:
- Implement transpose handling for gemm
- Factor out handling of arch/num_cu/features
- Fix fold-transpose to not mention conv by name
- Improve copy-opt so that it handles the case of more than one
- transform before the gemm-wrapping op or no transforms (which is
- what came up in the gemm case)

### DIFF
--- a/mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
+++ b/mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
@@ -21,7 +21,9 @@
 #include "mlir/Dialect/Rock/utility/loweringUtils.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/Dialect/Tosa/IR/TosaOps.h"
+#include "mlir/IR/BuiltinAttributes.h"
 #include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/Diagnostics.h"
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Transforms/DialectConversion.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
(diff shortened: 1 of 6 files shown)


mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -76,31 +78,22 @@ static Value expandTensor(ConversionPatternRewriter &rw, Operation *op,
...
 
-  // expand tensors from rank 4 (NHWC) to rank 5 (NHWCG)
-  auto inputExp = expandTensor(rw, op, input);
-  auto filterExp = expandTensor(rw, op, filter);
-  auto outputExp = expandTensor(rw, op, output);
-
   // TODO(sjw): get these from options
-  StringRef chip = "";
+  StringAttr chip = StringAttr::get(op->getContext(), "");
   uint32_t num_cu = 64;
```
sjw36, line 88: When can we get rid of this? Is it not an 'arch' detail.

mlir/lib/Dialect/Rock/Transforms/CopyOpt.cpp
```diff
@@ -147,23 +152,45 @@ template <typename T> struct MICORewritePattern : public OpRewritePattern<T> {
...
-        // 1.1 Input of rock.transform
+        // 1.2 Input of rock.transform
         if (writer)
           return fail;
-        Value mrval = mrop;
-        // 1.1.0 Confirm output of rock.conv2d
+        rock::TransformOp originalMrop = mrop;
+        Value mrval = mrop.getResult();
+        // Dig through chains of transposes
+        while (mrval.hasOneUse() && (mrop = dyn_cast_or_null<rock::TransformOp>(
```
sjw36, line 176: Good, good.

sjw36 (approved):
Looks good!

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
krzysz00: The problem is that MIOpen passes it in. I suppose we could make it optional ...: but it's not a function of chipset, I think.

PR #811, opened 2022-10-13
Make rocmlir-gen less dependent on 5D values, remove most CPU helpers

krzysz00:
1. Change up the memref population logic to use 1D loops on collapsed tensors instead of using 5D loops, allowing us to use 3D tensors in those places in the future
2. Drop all the mcpuMemset functions by replacing them with loops that either fill with a fixed value (for -rand fixed) or call random{Integer,Float}Value() repeatedly (for -rand 1 and friends)
3. Change the memcpy emission to emit functions that know their size: hopefully this makes the JIT even happier, and it seemed like a reasonable thing to do since we can.
4. Remove certain rocm-runner integration tests that mirror upstream and were using cpu memset
5. Adjust host harness tests to replace the memset() calls with linalg.fill
6. Change the randomization system to just call srand() once, and not once per memref we're filling, improving randomization - this means that one tosa test that relied on a particular line appearing in the output has been adjusted to account for the different random values
6. Make the Verify*() function take a 1D input, allowing for gemm later
7. Adjust a lot of the rocmlir-gen tests to account for these changes, and use the fact that FileCheck lets you pass multiple prefixes to simplify many of these tests


mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -1429,60 +1535,67 @@ createCPUConvFunc(ModuleOp module,
...
+  if (srcType.hasStaticShape())
+    numElements = srcType.getNumElements();
+
+  if (numElements != -1)
+    funcNameStr << numElements;
+  else
+    funcNameStr << "any";
+
+  if ((numElements == -1 && !destType.hasStaticShape()) ||
+      numElements != destType.getNumElements())
```
zhanglx13, line 1565: Does `getNumElements()` returns -1 if the type does not have static shape?

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -1429,60 +1535,67 @@ createCPUConvFunc(ModuleOp module,
...
-
-  auto rSrcType = MemRefType::get({-1}, srcElemType);
-  auto rDstType = MemRefType::get({-1}, dstElemType);
+  mlir::Location loc = b.getUnknownLoc();
 
   // clang-format off
-  // func _memcpy_<srcElemType>_<dstElemType> (%arg0 : memref<?xf32>, %arg1 : memref<?xf16, %arg2 : index) {
-  //   scf.for %i0 = %c0 to %arg2 step %c1 {
+  // func _memcpy_<srcElemType>_<dstElemType>_<size> (%arg0 : memref<sizexf32>, %arg1 : memref<sizexf16>) {
+  //   %size = memref.dim %arg0(%c0)
```
zhanglx13, line 1576: where does `%c0` come from? Is it omitted because it is not important?

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -1429,60 +1535,67 @@ createCPUConvFunc(ModuleOp module,
...
+
+  mlir::Type srcElemType = srcType.getElementType();
+  mlir::Type dstElemType = destType.getElementType();
+  // memcpy_<srcElemType>_<dstElemType>_(srcSize|any)
   std::string funcName = "_memcpy_";
-  funcName += getTypeStr(srcElemType);
-  funcName += "_";
-  funcName += getTypeStr(dstElemType);
+  llvm::raw_string_ostream funcNameStr(funcName);
+  funcNameStr << srcElemType << "_" << dstElemType << "_";
```
zhanglx13, line 1553: What is the benefit of using `raw_string_ostream`?

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
zhanglx13: And why you want the `size` as part of the function name?

mlir/test/rocmlir-driver/misc_e2e/conv2d_harness_cpu_verifier_kyxc_nhwc_nhwk.mlir
```diff
@@ -11,34 +11,22 @@ module {
...
-
-    // populate initial values.
-    %c0_i16 = arith.constant 0 : i16
-    %c1_i16 = arith.constant 1 : i16
-    %c1_i16_0 = arith.constant 1 : i16
-    %c1_i32 = arith.constant 1 : i32
-    call @mcpuMemset5DFloatRandInt(%3, %c1_i16, %c1_i16_0, %c1_i32) : (memref<?x?x?x?x?xf32>, i16, i16, i32) -> ()
-    call @mcpuMemset5DFloatRandInt(%4, %c1_i16, %c1_i16_0, %c1_i32) : (memref<?x?x?x?x?xf32>, i16, i16, i32) -> ()
-    call @mcpuMemset5DFloatRandInt(%5, %c0_i16, %c0_i16, %c1_i32) : (memref<?x?x?x?x?xf32>, i16, i16, i32) -> ()
+    linalg.fill ins(%cst : f32) outs(%0 : memref<1x128x3x3x8xf32>)
```
zhanglx13, line 14: Does it fill in random numbers?

krzysz00: No, that's meant to be a fixed value. Note that the old calls to "Rand" had the same min and max value

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
krzysz00: Aesthetics, and possibly some performance gains from not using `std::string::operator+=`

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
krzysz00: c0 is omitted because it's not important, following the practice of the existing comments.

And I put the `size` in the name because I'm wanting to make things easy on `-O3` at the cost of code size - I had an attack of "huh, that'd be clever". I can make it more like the old form if you want

zhanglx13: So on `-O3`, the size will be inferred from the function name instead of `%size = memref.dim %arg0(%c0)`?
Well, I am trying to learn your "tricks" so I'd rather you keep it.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
krzysz00: Ok, so, my best guess - I might have to try this - is that if you pass in a `memref<*xT>` `getNumElements()` will, by way of `getShape()`, explode, but a `memref<?xT>` (for any number of `?`) will also explode by way of the computation going negative, which touches off an assert about integer overflow (or maybe something earlier)

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
krzysz00: Ok, so, two things of note:
1. As soon as we run `-canonicalize` on the IR (which happens at least once in our pipeline), that `memref.dim` gets replaced by the constant size from the type.
2. (Having gone and looked, it's not clear what optimization level we're using for our host code, it might be -O0 - I'll have to fix that in a future PR)
3. The idea here - which is partly broken by the fact that we stuck a type conversion in here so it might not work, but still - is that compilers are very good at spotting `memcpy()`, _especially_ `memcpy()` with a constant size, and will duly rewrite it to the most convoluted mess of highly-optimized assembly they can manage ... which is a function of the amount of memory being copied. So having separate functions with different constant sizes could give the host code bits a chance to do something very, very, clever

zhanglx13: I see. Thank you for the explanation.

zhanglx13 (approved):
Very solid refactor and cleanup!

Just out of curiosity. Do you know why we had those 4D* and 5D* cpu utility functions at the beginning instead of making everything 1D flat?

krzysz00:
I think it was because upstream did a lot of N-D functions back in the day so we copied them? But I'm not sure

PR #812, opened 2022-10-14
Rock tuning API

jungpark-mlir:
API to provide tuning space for the given Rock module.

Has:
- brute-force creation of conv2d Fwd tuning parameters
- test using C-API

Has not:
- consideration of layout/direction/data_type
- gemm tuning parameters
- problem definition of the given module that user can use as a DB key


jungpark-mlir:
FYI, output from the c-api test

Got tuning space,
Obtained perfconfig : "64,32,32,4,2,2"
quick set = 0, full set = 324
module {
  func.func @main(%arg0: memref<1x64x56x56xf32>, %arg1: memref<64x64x1x1xf32>, %arg2: memref<64xf32>, %arg3: memref<1x64x56x56xf32>) attributes {kernel} {
    %cst = arith.constant 0.000000e+00 : f32
    %cst_0 = arith.constant 3.40282347E+38 : f32
    %0 = memref.alloc() : memref<1x64x56x56xf32>
    %1 = rock.transform %arg0 by [#rock.transform_map<affine_map<(d0, d1, d2, d3, d4) -> (d0, d1, d2, d3)> by [<PassThrough ["dim0", "dim1", "dim2", "dim3"] at [0, 1, 2, 3] -> ["dim0", "dim1", "dim2", "dim3"] at [0, 1, 2, 3]>, <AddDim{1} ["g"] at [4] -> [] at []>] bounds = [1, 64, 56, 56, 1] -> [1, 64, 56, 56]>] : memref<1x64x56x56xf32> to memref<1x64x56x56x1xf32>
    %2 = rock.transform %arg1 by [#rock.transform_map<affine_map<(d0, d1, d2, d3, d4) -> (d0, d1, d2, d3)> by [<PassThrough ["dim0", "dim1", "dim2", "dim3"] at [0, 1, 2, 3] -> ["dim0", "dim1", "dim2", "dim3"] at [0, 1, 2, 3]>, <AddDim{1} ["g"] at [4] -> [] at []>] bounds = [64, 64, 1, 1, 1] -> [64, 64, 1, 1]>] : memref<64x64x1x1xf32> to memref<64x64x1x1x1xf32>
    %3 = rock.transform %0 by [#rock.transform_map<affine_map<(d0, d1, d2, d3, d4) -> (d0, d1, d2, d3)> by [<PassThrough ["dim0", "dim1", "dim2", "dim3"] at [0, 1, 2, 3] -> ["dim0", "dim1", "dim2", "dim3"] at [0, 1, 2, 3]>, <AddDim{1} ["g"] at [4] -> [] at []>] bounds = [1, 64, 56, 56, 1] -> [1, 64, 56, 56]>] : memref<1x64x56x56xf32> to memref<1x64x56x56x1xf32>
    rock.conv2d(%2, %1, %3) features =  none {arch = "", dilations = [1 : i32, 1 : i32], filter_layout = ["k", "c", "y", "x", "g"], input_layout = ["ni", "ci", "hi", "wi", "gi"], numCu = 64 : i32, output_layout = ["no", "ko", "ho", "wo", "go"], padding = [0 : i32, 0 : i32, 0 : i32, 0 : i32], perf_config = "64,32,32,4,2,2", strides = [1 : i32, 1 : i32]} : memref<64x64x1x1x1xf32>, memref<1x64x56x56x1xf32>, memref<1x64x56x56x1xf32>
    linalg.generic {indexing_maps = [affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>, affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%0 : memref<1x64x56x56xf32>) outs(%arg3 : memref<1x64x56x56xf32>) {
    ^bb0(%arg4: f32, %arg5: f32):
      %4 = arith.cmpf olt, %arg4, %cst : f32
      %5 = arith.select %4, %cst, %arg4 : f32
      %6 = arith.cmpf olt, %cst_0, %arg4 : f32
      %7 = arith.select %6, %cst_0, %5 : f32
      linalg.yield %7 : f32
    }
    return
  }
}
arg#0 (rank 4): <dim 0 : 1> <dim 1 : 64> <dim 2 : 56> <dim 3 : 56>
arg#1 (rank 4): <dim 0 : 64> <dim 1 : 64> <dim 2 : 1> <dim 3 : 1>
arg#2 (rank 1): <dim 0 : 64>
arg#3 (rank 4): <dim 0 : 1> <dim 1 : 64> <dim 2 : 56> <dim 3 : 56>
kernel name : main
block size : 64, grid size : 196
bin size : 9376
PASSED!

mlir/lib/CAPI/Dialect/Rock.cpp
```diff
@@ -0,0 +1,342 @@
...
+
+// Brute-force search in incremental order
+void createConv2DFwdTuningRangeBF(struct rock::TunableParams *newSpace,
+                                  rock::Conv2DOp convOp) {
+
+  auto bXdlops = convOp->getAttrOfType<BoolAttr>("xdlopsV2");
+  if (bXdlops != nullptr && bXdlops.getValue() == true) {
+    // XDLOPS
+    // M/block N/block K/block M/wave N/wave kPack aCopyMore bCopyMore
+    std::vector<std::vector<uint32_t>> tParams = {{4, 8, 16, 32, 64, 128},
```
sjw36: If these are always constant, let's make it a constexpr so it is only built at compile time.

mlir/include/mlir-c/Dialect/Rock.h
```diff
@@ -0,0 +1,64 @@
+//===-- mlir-c/Dialect/Rock.h - C API for Rock dialect --------*- C
+//-*-===//
```
krzysz00: Nit: could you remove some dashes to get this all onto one line

mlir/lib/CAPI/Dialect/Rock.cpp
```diff
@@ -0,0 +1,342 @@
...
+  int64_t leftPadW = ctx.getPaddingVal()[2];
+  int64_t rightPadH = ctx.getPaddingVal()[1];
+  int64_t rightPadW = ctx.getPaddingVal()[3];
+
+  int64_t dilationH = ctx.getDilationVal()[0];
+  int64_t dilationW = ctx.getDilationVal()[1];
+  int64_t strideH = ctx.getStrideVal()[0];
+  int64_t strideW = ctx.getStrideVal()[1];
+
+  std::vector<std::vector<uint32_t>> tParams = {{4, 8, 16, 32, 64, 128},
```
sjw36: Also constexpr.

mlir/lib/CAPI/Dialect/Rock.cpp
```diff
@@ -0,0 +1,342 @@
...
+                                                {0, 1}};
+  // std::vector<uint32_t> p7b = {1, 2, 4, 8};
+
+  // Each field of tuning parameter needs one set of coefficient + 1 constant.
+  // ((16+1)*8) number of quick set can be also set by MLR
+
+  // clang-format off
+  //                                         in_ch ;   in_h;   in_w;  fil_h;  fil_w; out_ch;  batch;  lPadH;  lPadW;  rPadH;  rPadW;   strH;   strW;   dilH;   dilW;  group;  const;
+  // From gfx90878_1.1.0.udb tuned at 9.Oct.2022  Weekly CI xdlops
+  std::vector<std::vector<float>> coeffs = {{ 0.0175,-1.1229,-1.1229,-12.153,-12.153,-0.0413, 0.0000,47.5203,47.5203,47.5203,47.5203,55.0025,55.0025, 0.0000, 0.0000, 0.0000, 108.5339},
```
sjw36: and constexpr.

sjw36 (approved):
LGTM. Just constexpr a few of things please.

mlir/include/mlir-c/Dialect/Rock.h
```diff
@@ -0,0 +1,64 @@
...
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM
```
krzysz00, line 3: Nit: part of the rocMLIR project, and add a `Copyright (c) 2022 Advanced Micro Devices Inc.` to the bottom of the header bit

mlir/include/mlir/Dialect/Rock/Tuning/RockTuningTypes.h
```diff
@@ -0,0 +1,41 @@
...
+
+#ifndef MLIR_DIALECT_ROCK_ROCKTUNINGTYPE_H
+#define MLIR_DIALECT_ROCK_ROCKTUNINGTYPE_H
+
+#include "mlir/Dialect/Rock/IR/Rock.h"
+
+namespace mlir {
+namespace rock {
+
+enum TunableOpType { TunableOpConv2dFwd, TunableOpGemm };
```
krzysz00: Nit: `enum class`

Also, let's not make a separate enum here. Could you please instead extend `ConvOpType` to `GemmWrapperOpType` and add `Gemm` as a case, or make a second enum `GemmWrapperOpType` that includes all the convolutions? Then, you can change the API in the tuning parameters struct to take this four-way enum - I'm wanting this for `miopen-gen` anyhow

mlir/lib/CAPI/Dialect/Rock.cpp
```diff
@@ -0,0 +1,342 @@
...
+DEFINE_C_API_PTR_METHODS(MlirRockTuningParam, mlir::rock::ParamEntry)
+// DEFINE_C_API_METHOD(MlirRockTuningKey, const void);
+
+using namespace mlir;
+
+// Brute-force search in incremental order
+void createConv2DFwdTuningRangeBF(struct rock::TunableParams *newSpace,
+                                  rock::Conv2DOp convOp) {
+
+  auto bXdlops = convOp->getAttrOfType<BoolAttr>("xdlopsV2");
```
krzysz00: Nope, you want .getFeatures(). `xdlopsv2` isn't an attribute anymore. Also, make this function take `rock::RockGemmWrapperInterface`

mlir/include/mlir/Dialect/Rock/Tuning/RockTuningTypes.h
```diff
@@ -0,0 +1,41 @@
...
+
+// Tuning parameter struct
+typedef std::vector<uint32_t> TuningParam;
```
krzysz00: Strong object: instead of doing this, please make this be `Attribute` ... or, really, here's the refactor I'd like to see: define an attribute interface, `RockTuningParamAttr`, and give it the methods you'd like - the main one is probably `toPerfConfig()`. That way, in the future, we'll be able to use the attributes directly and not have perf_config at all.

Also, having this be a pile of numbers loses important semantic information

mlir/lib/CAPI/Dialect/Rock.cpp
krzysz00: Big problem here: the range of the search space is dependent on the datatype, I'm pretty sure, mostly around kPack with int8. Also, you can have kPack = 2.

Third, why hardcode these constants when you can instead loop through them?

mlir/lib/CAPI/Dialect/Rock.cpp
```diff
@@ -0,0 +1,342 @@
...
+          }
+        }
+      }
+    }
+  }
+
+  newSpace->numHeuristicQuick = 0;
+}
+
+void createConv2DFwdTuningRangeMLR(struct rock::TunableParams *newSpace,
```
krzysz00: Secondly, this should take any tuneable thing

mlir/lib/CAPI/Dialect/Rock.cpp
```diff
@@ -0,0 +1,342 @@
...
+                                   rock::Conv2DOp convOp) {
+  // Initial point determined by Multi Linear Regression coefficients and
+  // iterate further on.
+
+  // These needs to have separate analysis for their own coefficient.
+  // For now, NCHW/F32/FWD only.
+  uint64_t layout = 0; // NCHW
```
krzysz00: Where do you set `layout`?

mlir/lib/CAPI/Dialect/Rock.cpp
krzysz00: This'll be architecture-dependent, make sure `arch` is a parameter somehow.

And probably include a script to generate these.

mlir/lib/CAPI/Dialect/Rock.cpp
```diff
@@ -0,0 +1,342 @@
...
+            }
+          }
+        }
+      }
+    }
+  }
+  newSpace->numHeuristicQuick = 1;
+}
+
+struct rock::TunableParams *createTunableParams(Operation *op) {
```
krzysz00: RockGemmWrapperInterface, not `Operation *`

mlir/lib/CAPI/Dialect/Rock.cpp
```diff
@@ -0,0 +1,342 @@
...
+    newSpace = nullptr; // not supported yet
+  }
+  return newSpace;
+}
+
+// Stringfy given param
+std::string getPerfConfigStr(std::vector<uint32_t> param) {
+  std::string result;
+  int i = 0;
+  for (; i < param.size() - 1; i++) {
```
krzysz00: Yeah, this should be a method on the `RockTuningParamsAttr` that I want to have, which'll take care of not including kPack in non-xdlops and adding the extra 0s in xdlops

krzysz00 (changes requested):
Strong design objections.

The tuning parameters should be represented as the MLIR-side attribute, not a vector of numbers.

This code should be using existing bits of MLIR's stuff, like `RockGemmWrapperInterface`, and adding relevant abstractions around xdlops/non-xdlops tuning parameter sets.

And also, I'd like to see most of the non-trivial code in `mlir/Dialect/Rock/` with the C API being a thin wrapper.

mlir/include/mlir-c/Dialect/Rock.h
jungpark-mlir: I can see that looks more like what the standard APACHE license template says, but different from the other part of MLIR project. Can you let me know what's the right policy for the license header?

krzysz00: That's a @mvermeulen question, really, and we're probably due for a mechanical PR that fixes up our header comments

I've taken to using "rocMLIR project" and putting an AMD copyright in new files

mlir/lib/CAPI/Dialect/Rock.cpp
```diff
@@ -0,0 +1,382 @@
...
+// DEFINE_C_API_METHOD(MlirRockTuningKey, const void);
+
+using namespace mlir;
+
+//////--------------- FIX! --------- Will be relocaetd  to
+/// rockTuningImpl.cpp----------///////////
+// vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
+
+// Brute-force search in incremental order
+void createGemmTuningRangeBF(struct rock::TunableParams *newSpace,
```
krzysz00: Can this be a C++ function that lives in the main codebase and then the C function is a very thin wrapper?

mlir/lib/CAPI/Dialect/Rock.cpp
```diff
@@ -0,0 +1,382 @@
...
+  return newSpace;
+}
+
+// ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+//////--------------- FIX! --------- Will be relocaetd  to
+/// rockTuningImpl.cpp----------///////////
+
+//////// This one to gemm wrapper interfasce
+// Stringfy given param
+std::string toPerfConfig(Attribute param) {
```
krzysz00: I'd like to see a `GemmParamsAttrInterface` for attributes, and for `toPerfConfig` to be a method on that interface. And then these vectors can be vectors of `GemmParamsAttrInterface`.

mlir/lib/CAPI/Dialect/Rock.cpp
```diff
@@ -0,0 +1,382 @@
...
+  std::string result;
```
krzysz00: Maybe a `SmallVectorImpl<char> &out`, and also `llvm::raw_svector_ostream`, so you can use `<<`

And probably you can go with `Twine`

krzysz00 (commented):
I've got more comments here, and some of them are "can we move a lot of this out of the C API and into the main code?"

mlir/lib/CAPI/Dialect/Rock.cpp
jungpark-mlir: Sure, I'm still working on it.

mlir/lib/CAPI/Dialect/Rock.cpp
jungpark-mlir: That's good point.

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
```diff
@@ -211,9 +211,10 @@ def Rock_GeneralGemmParamsAttr : Rock_Attr<"GeneralGemmParams", [RockTuningParam
 
   let extraClassDeclaration = [{
     std::string getPerfConfigStr(){
-      const Twine paramsConcat = Twine(getMPerBlock()) + ","
-      + Twine(getNPerBlock()) + ","
+      const Twine paramsConcat = Twine(64) + "," /* block size */
```
krzysz00: The perf config has a block size? Is that part of the tuning space?

mlir/include/mlir/Dialect/Rock/Tuning/RockTuning.h
```diff
@@ -0,0 +1,43 @@
...
+};
+
+// Total tuning space
+struct TunableParams {
+  std::vector<RockTuningParamAttrInterface> tuningRange;
+  KernelType primaryOpType;
+  int numHeuristicQuick;
+};
+
+TunableParams *createTunableParams(ModuleOp &mod);
```
krzysz00: Question: instead of using pointers, can we use some of MLIR's facilities to store the parameter and the space into the relevant MLIR context? 

(It's entirely fine if the answer is "no", I just want to check on the possibility)

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
krzysz00: And do we ignore it or use it?

mlir/lib/CAPI/Dialect/CMakeLists.txt
```diff
@@ -9,3 +9,13 @@ add_mlir_public_c_api_library(MLIRCAPIMIGraphX
   MLIRRockPipeline
   MLIRGPUTransforms
 )
+
+add_mlir_public_c_api_library(MLIRCAPIRock
```
krzysz00, line 13: Should this be here or in a subdirectory?

mlir/lib/CAPI/Dialect/Rock.cpp
```diff
@@ -0,0 +1,79 @@
...
+int mlirRockTuningGetNumParamsFull(MlirRockTuningSpace params) {
+  auto tuningSpace = unwrap(params);
+  return tuningSpace->tuningRange.size();
+}
+
+// Allocate parameter entry to hold string allocation and pass it more easily.
+MLIR_CAPI_EXPORTED MlirRockTuningParam
+mlirRockTuningCreateParamAt(MlirRockTuningSpace params, int pos) {
+  auto tuningSpace = unwrap(params);
+  rock::ParamEntry *param = new rock::ParamEntry();
```
krzysz00, line 48: Yeah, this is vaguely sus. So having the entire tuning space be an attribute might be the right call?

If nothing else, why not return a pointer into the vector?

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -0,0 +1,111 @@
...
+  struct TunableParams *newSpace;
+  newSpace = new TunableParams();
+
+  // create range and heuristic
+  bool bFound = false;
+  mod->walk([&](rock::RockGemmWrapperInterface op) {
+    if (!bFound) {
+      bFound = true;
+      createGemmTuningRangeBF(newSpace, op);
+      newSpace->primaryOpType = op.getKernelType();
```
krzysz00: You don't need `bFound`, `WalkResult::Interrupt()` is sufficient.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -0,0 +1,111 @@
...
+  return newSpace;
+}
+
+bool tuningSetParam(ModuleOp &mod, ParamEntry *paramEntry) {
+  bool bFound = false;
+  mod->walk([&](rock::RockGemmWrapperInterface op) {
+    if (!bFound) {
+      bFound = true;
+      auto ctx = op.getContext();
+      StringAttr attr = StringAttr::get(ctx, paramEntry->perfString);
```
krzysz00: Why not have the perf_config be a string attribute in the first place?

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -0,0 +1,111 @@
...
+      op->setAttr("perf_config", attr);
+    }
+  });
+  return bFound;
```
krzysz00: Same comment about WalkResult - you can check if you interrupted or not

krzysz00 (commented):
This is looking pretty good, just some minor comments

mlir/lib/CAPI/Dialect/CMakeLists.txt
krzysz00: (Looks like the practice is to put it here, resolving)

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
jungpark-mlir: Actually I was going to ask this but just forgot. Let me check with the others.

mlir/lib/CAPI/Dialect/Rock.cpp
jungpark-mlir: This is to have an own allocation of one params entry without tuning space. Later, I'd like to add another api creating ParamEntry from perfconfig string.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
jungpark-mlir: That's nice, thanks.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
jungpark-mlir: Right, that sounds better.
setParam is likely to be called more often than getParamStr from the users.

mlir/include/mlir/Dialect/Rock/Tuning/RockTuning.h
jungpark-mlir: I can't find anything useful, there is no enough mlir-c use-case unfortunately.
Using pointer just because the caller needs to provide the ownership of the allocation.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
jungpark-mlir: Uhm, just found there is a problem. A ParamEntry itself cannot have an mlir context to permanently stick with. So I can't create a StringAttr for a ParamEntry. ParamEntry will be created independently without tuning space (not yet available now).

krzysz00: I don't understand the last part of your comment

And yeah, given that we can make this a string

But, in the future, I'll want to not have `perf_config` strings at all except as part of MIOpen's interface, so we'll be able to get rid of this string

mlir/include/mlir/Dialect/Rock/Tuning/RockTuning.h
krzysz00: It's not about the C API, it's about MLIRContext having the ability to store things

mlir/lib/CAPI/Dialect/Rock.cpp
krzysz00: Ok, in which case, please add a comment indicating that this is a fresh allocation the caller is responsible for freeing

jungpark-mlir: https://mlir.llvm.org/docs/CAPI/#naming-convention-and-ownership-model
mlirXCreate : creates object and caller gets the ownership
and should be freed by - mlirXDestroy

krzysz00 (commented):
So the code looks good to me, as far as style/"am I willing to let this into our repo".

However, it looks possible that non-xdlops might force you to include blockSize as part of a tuning parameters entry, perhaps an `Optional<uint32_t>`

jungpark-mlir:
> So the code looks good to me, as far as style/"am I willing to let this into our repo".
> 
> However, it looks possible that non-xdlops might force you to include blockSize as part of a tuning parameters entry, perhaps an `Optional<uint32_t>`

Just made it not optional, tablegen doesn't recognize it as optional (because it's "optional <int>" ), it doesn't generates functions without blocksize argument. So caller always need to provide it.

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
```diff
@@ -200,6 +201,7 @@ def Rock_GeneralGemmParamsAttr : Rock_Attr<"GeneralGemmParams", [RockTuningParam
     - kpack: The number of values of k to pack contiguously into the shared buffer
   }];
   let parameters = (ins
+    "uint32_t":$blockSize,
```
krzysz00, line 204: Don't put this here. We already have blockSize as a separate attribute.

Alternatively, make this change for xdlops as well, and get rid of the blockSize attribute.

krzysz00: (Honestly, either one's fine, and putting block size in the big params struct might not be a bad cleanup ... just do it uniformly across xdlops and non-xdlops)

krzysz00 (commented):
Note wrt blockSize handling

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
jungpark-mlir: I can't easily agree to this.

blocksize is turned out to be a proper tuning parameter, needs to be there.
general/xdlops are already quite different ops, we don't need to have uniform tuning parameter over different ops.

Also, if we get blocksize into the xdlopsGemm's params set, that will break uniformity in tuning flow. 
Either user needs to provide correctly calculated blocksize in the tuning params or we need to ignore the given number.

For now, it only duplicates the value in op.attr and param since AffixTuningParameter

I'd suggest to revisit this problem when we remove perfconfig string in near future.

krzysz00: And I can't agree to that duplication, they might get out of sync

krzysz00: I am, however, willing to go for this:
- On the high-level ops (conv, gemm, etc., the ones that don't know if they're xdlops or not), rename `blockSize` to something like `derivedBlockSize` - and possibly do the same to `gridSize` (to `derivedGridSize`)
- In the non-xdlops flow, setting `derivedBlockSize` is an error (update `::verify`, either on the interfaces or on the ops themselves, to check this), and add `blockSize` to the tuning parameters like you did
- Add a `RockGemmWrapperInterface` method `Optional<uint32_t> getBlockSize()` that returns the blockSize from general tuning parameters if those are available and goes for `derivedBlockSize` otherwise.

krzysz00: (or maybe don't need the interface)
- Delete blockSize attribute on gridwise_gemm, blockwise_gemm, ...
- **Don't** delete or rename blockSize on gridwise_gemm_v2, blockwise_gemm_v2, ...

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
```diff
@@ -243,11 +247,14 @@ def Rock_XdlopsGemmParamsAttr : Rock_Attr<"XdlopsGemmParams", [RockTuningParamAt
...
   let extraClassDeclaration = [{
     std::string getPerfConfigStr(){
-      const Twine paramsConcat = Twine(getMPerBlock()) + ","
+      const Twine paramsConcat = 
+          Twine(getMPerBlock()) + ","
         + Twine(getNPerBlock()) + ","
         + Twine(getKPerBlock()) + ","
         + Twine(getMPerWave()) + ","
-        + Twine(getNPerWave()) + ",1,1";
+        + Twine(getNPerWave()) + ","
```
krzysz00, line 255: You have getNPerWave() twice

krzysz00: and "1,1 twice

mlir/include/mlir/Dialect/Rock/IR/RockGemmWrapperInterface.td
```diff
@@ -129,9 +129,32 @@ def RockGemmWrapperInterface : OpInterface<"RockGemmWrapperInterface"> {
...
+        /*desc=*/[{
+          Get blocksize if it is non-Xdlops and given in the gemmParam
+          Return 0 otherwise
+        }],
+        /*retType=*/"uint32_t",
+        /*methodName=*/"getBlockSize",
+        /*args=*/(ins),
+        /*methodBody=*/"",
+        /*defaultImplementation=*/[{
+          $_op->getDerivedBlockSize().value_or($_op->getParams().getBlockSize());
```
krzysz00: Nope.

mlir/include/mlir/Dialect/Rock/IR/RockGemmWrapperInterface.td
```diff
@@ -129,9 +129,32 @@ def RockGemmWrapperInterface : OpInterface<"RockGemmWrapperInterface"> {
         /*defaultImplementation=*/[{
           $_op->setAttr($_op.getParamsAttrName(), params);
         }]
+      >,
+    InterfaceMethod<
+        /*desc=*/[{
+          Get Optional blocksize
```
krzysz00: Get the value of the derived block size attribute if it is set.

mlir/include/mlir/Dialect/Rock/IR/RockGemmWrapperInterface.td
```diff
@@ -129,9 +129,32 @@ def RockGemmWrapperInterface : OpInterface<"RockGemmWrapperInterface"> {
...
+        }],
+        /*retType=*/"Optional<uint32_t>",
+        /*methodName=*/"getDerivedBlockSize",
+        /*args=*/(ins),
+        /*methodBody=*/"",
+        /*defaultImplementation=*/""
```
krzysz00, line 141: `$_op.getDerivedBlockSize()` or something to do with `getDerivedBlockSizeAttr()` and `getZExtValue()` if you can't do that.

mlir/include/mlir/Dialect/Rock/IR/RockGemmWrapperInterface.td
```diff
@@ -129,9 +129,32 @@ def RockGemmWrapperInterface : OpInterface<"RockGemmWrapperInterface"> {
...
+        }]
+      >,
+    InterfaceMethod<
+        /*desc=*/[{
+          Get blocksize if it is non-Xdlops and given in the gemmParam
```
krzysz00: Wrong semantics, wrong documentation.

`Optional<uint32_t> getBlockSize()` is the signature
The docs are
```
Return the block size this operation will execute with if it has been selected or None otherwise.
Prioritizes the block size defined in tuning parameters
```

mlir/include/mlir/Dialect/Rock/IR/RockGemmWrapperInterface.td
krzysz00: Something like
```
if (auto generalParams = $_op.getParamsAttr().dyn_cast_or_null<General...>()) {
  return generalParams.getBlockSize();
}
return $_op.getDerivedBlockSize();
```

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1164,7 +1164,7 @@ struct GridwiseGemmV2RewritePattern
     }
 
     // Obtain critical tuning parameters.
-    uint32_t blockSize = op.getBlockSize();
+    uint32_t blockSize = op.getDerivedBlockSize();
```
krzysz00: GridwiseGemmV2 should not have experienced this renaming, it's on the xdlops path

krzysz00 (changes requested):
1. I don't like the semantics of those interface methods
2. The validation on RockGemmWrapperInterface (or on its implementors, but you might as well put it on the interface) include something like
```
if ($_op.getDerivedBlockSize().has_value() && $_op.getParams().isa_and_nonnull<GeneralGridwiseParamsAttr>()) {
  return $_op.emitOpError("Derived block size attribute set when block size is present in tuning parameters");
```

mlir/test/CAPI/mixr_full.c
```diff
@@ -165,6 +166,20 @@ static bool constructAndTraverseIr(MlirContext ctx) {
...
+  MlirRockTuningParam obtainedParam =
+      mlirRockTuningCreateParamAt(tuningSpace, 0);
+  mlirRockTuningSetParam(module, obtainedParam);
+  MlirStringRef paramStr = mlirRockTuningGetParamStr(obtainedParam);
+  mlirRockTuningParamDestroy(obtainedParam);
+  mlirRockTuningSpaceDestroy(tuningSpace);
+
+  printf("Got tuning space,\n");
+  printf("Obtained perfconfig : \"%s\"\n", paramStr.data);
+  printf("quick set = %d, full set = %d\n", qNum, fNum);
```
jerryyin: This seems to be printing from a test. Is anywhere is expecting those test prints in unit tests?

mlir/include/mlir-c/Dialect/Rock.h
```diff
@@ -0,0 +1,62 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+MLIR_DECLARE_CAPI_DIALECT_REGISTRATION(Rock, rock);
+
+#define DEFINE_C_API_STRUCT(name, storage)                                     \
+  struct name {                                                                \
+    storage *ptr;                                                              \
+  };                                                                           \
```
jerryyin, line 27: Why wrap a raw pointer inside of a struct?

mlir/lib/CAPI/Dialect/Rock.cpp
```diff
@@ -0,0 +1,79 @@
...
+  auto mod = unwrap(module);
+  newParams = rock::createTunableParams(mod);
+  return wrap(newParams);
+}
+
+MLIR_CAPI_EXPORTED int
+mlirRockTuningGetNumParamsQuick(MlirRockTuningSpace params) {
+  auto tuningSpace = unwrap(params);
+  return tuningSpace->numHeuristicQuick;
+}
```
jerryyin, line 39: How is this API useful? Or more fundamentally, what does `numHeuristicQuick` field trying to do? I see it is set to 0 now which make me think this field shouldn't exist.

mlir/include/mlir-c/Dialect/Rock.h
```diff
@@ -0,0 +1,62 @@
+//===-- mlir-c/Dialect/Rock.h - C API for Rock dialect ---- C--===//
```
jerryyin, line 1: Since this is actually going to be exported, I strongly suggest to rename.

mlir/lib/CAPI/Dialect/Rock.cpp
```diff
@@ -0,0 +1,79 @@
...
+  auto tuningSpace = unwrap(params);
+  return tuningSpace->tuningRange.size();
+}
+
+// Allocate parameter entry to hold string allocation and pass it more easily.
+MLIR_CAPI_EXPORTED MlirRockTuningParam
+mlirRockTuningCreateParamAt(MlirRockTuningSpace params, int pos) {
+  auto tuningSpace = unwrap(params);
+  rock::ParamEntry *param = new rock::ParamEntry();
+  param->param = tuningSpace->tuningRange[pos];
```
jerryyin: This is not defensive enough. A client can pass in a `pos` larger than the size of range therefore crashing our plugin.

mlir/lib/CAPI/Dialect/Rock.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+MLIR_CAPI_EXPORTED
+int mlirRockTuningGetNumParamsFull(MlirRockTuningSpace params) {
+  auto tuningSpace = unwrap(params);
+  return tuningSpace->tuningRange.size();
+}
+
+// Allocate parameter entry to hold string allocation and pass it more easily.
+MLIR_CAPI_EXPORTED MlirRockTuningParam
+mlirRockTuningCreateParamAt(MlirRockTuningSpace params, int pos) {
```
jerryyin: Why make this opaque? Why not use a comma splited c string as the current tuning entry?

I think all 3 APIs can be consolidated into 1:
 - mlirRockTuningCreateParamAt
 - mlirRockTuningSetParam
 - mlirRockTuningGetParamStr

mlir/lib/CAPI/Dialect/Rock.cpp
```diff
@@ -0,0 +1,79 @@
...
+  auto tuningSpace = unwrap(params);
+  rock::ParamEntry *param = new rock::ParamEntry();
+  param->param = tuningSpace->tuningRange[pos];
+  param->perfString = param->param.getPerfConfigStr();
+  return wrap(param);
+}
+
+MLIR_CAPI_EXPORTED
+void mlirRockTuningParamDestroy(MlirRockTuningParam param) {
+  delete (unwrap(param));
```
jerryyin, line 54: Does `unwrap` help you do the nullptr check?

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -0,0 +1,99 @@
...
+        for (uint32_t gemmKPerBlock : ValidRangeXdlopsGemmParams[2]) {
+          for (uint32_t gemmMPerWave : ValidRangeXdlopsGemmParams[3]) {
+            for (uint32_t gemmNPerWave : ValidRangeXdlopsGemmParams[4]) {
+              for (uint32_t gemmKPack : ValidRangeXdlopsGemmParams[5]) {
+                XdlopsGemmParamsAttr gemmParams =
+                    b.getAttr<XdlopsGemmParamsAttr>(
+                        gemmKPerBlock, gemmMPerBlock, gemmNPerBlock, gemmKPack,
+                        gemmMPerWave, gemmNPerWave);
+                newSpace->tuningRange.push_back(
+                    gemmParams.cast<RockTuningParamAttrInterface>());
```
jerryyin, line 51: This seems extremely inefficient. Down sides:
 - Time complexity is very bad. The initialization time forced to be amount of tuning space.
 - Space complexity is very bad. We brute force duplicated a lot of information.

jerryyin (changes requested):
I still have some serious questions towards the way we use this set of API.

mlir/include/mlir-c/Dialect/Rock.h
krzysz00: `Dialect.cpp` for the C API relating to [Dialect] is standard convention upstream

mlir/lib/CAPI/Dialect/Rock.cpp
krzysz00: I strong disagree with having perf_config be a first-class citizen any more than necessary. We want to get rid of it, not entrench it further, IMO.

mlir/include/mlir-c/Dialect/Rock.h
jerryyin: But this is a very limited subset of Rock dialect, is `mlir-c` the best place to export it? I feel like this can be another orthogonal API like `Miir.h`. Eventually MIOpen is supposed to pull it in (one way or another) as well.

mlir/lib/CAPI/Dialect/Rock.cpp
jerryyin: Why not? It is how it is represented in the db, and a c-string is a very straightforward way to represent this information.

krzysz00: I'd like to see us constructing `[...]ParamsAttr` as soon as we can inside our code, so that perf_config becomes a very niche serialization format that only exists at the very edges of our API.

This would also mean getting rid of `InitParams[Non]XDL`

mlir/include/mlir-c/Dialect/Rock.h
jungpark-mlir: This is standard way to make the type opaque in mlir-c
https://github.com/llvm/llvm-project/blob/66b830889d9bc4bfc58aa1ea47d1b3a7cbee7c5c/mlir/include/mlir-c/IR.h#L31

mlir/lib/CAPI/Dialect/Rock.cpp
jungpark-mlir: This is assuming we can sort the tuning parameters in "likely to optimal" order with an analysis, and also can do another analysis to determine the number of tries until reach at the optimal number.
Currently it's set 0 because we don't have that analysis in current heuristic yet.

mlir/lib/CAPI/Dialect/Rock.cpp
jungpark-mlir: > Why make this opaque? Why not use a comma splited c string as the current tuning entry?
> 
> I think all 3 APIs can be consolidated into 1:
> 
> * mlirRockTuningCreateParamAt
> * mlirRockTuningSetParam
> * mlirRockTuningGetParamStr

We can have another API function to bind the first two but there is still missing API to get a param from string, that's the reason I split the first two.
The last one is needed when client wants to get serialized string to write the record back to the DB, so it's needed separately.

mlir/lib/CAPI/Dialect/Rock.cpp
jungpark-mlir: That's good point, we should cover that.

mlir/include/mlir-c/Dialect/Rock.h
jerryyin: Thanks for the pointer

mlir/lib/CAPI/Dialect/Rock.cpp
jungpark-mlir: Not really, but *Destroy function is just a wrapper to delete the given object, user should call it correctly.
As like delete() itself doesn't check the nullptr but leave it to crash when user double frees it.
That's the practice in mlir-c
https://github.com/llvm/llvm-project/blob/b172c7e1933b04f459b813a6c894d90ea65a9cd4/mlir/lib/CAPI/IR/IR.cpp

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
jungpark-mlir: I'm not trying to have something optimal this time because we don't know which way is the best at the moment.

Firstly, I'd like to store the parameters in the storage so that we can change the order at any time. This will help exploring quick tuning in the future and we can get something else once we figure out what to do.
With that, there is no faster way to get full tuning space, This is exactly the same we're currently doing in MIOpenDriver code.

Also, this is only a part of the internal implementation, we can change this in the future without changing the API itself. For example, we can embed the tuning parameter generator as a function in the tuning space struct under the same API.

mlir/lib/CAPI/Dialect/Rock.cpp
jerryyin: This sounds like a interesting idea, I think it does worth exploring. I have an question in the following use case:

The client want to do the tuning in the `quick` tuning range as well as using the `numHeuristicQuick` amount of times to do the tuning. However, it looks like `mlirRockTuningSetParam(tuningSpace, 0)` does not seem to care about which tuning space it is in. The results of this seems to indicate that the client is still working at the `full` tuning space, just less in numbers.

I think in order to achieve what you describe, you will need to differentiate between the two sorts of tuning space as well:
 - mlirRockTuningSpaceCreateQuick()
 - mlirRockTunignSpaceCreateFull()

mlir/test/CAPI/mixr_full.c
jungpark-mlir: Yeah, that should be done in the next step, I'm planning to expose some functionalities to a standalone tool or as rocmlir-gen features so that we can more easily include it in our infrastructure

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -0,0 +1,99 @@
...
+          }
+        }
+      }
+    }
+  }
+
+  newSpace->numHeuristicQuick = 0;
+}
+
+TunableParams *createTunableParams(ModuleOp &mod) {
```
jerryyin: Maybe rename it as `createTunableSpace()`? I misunderstood this as setting a single param at first sight.

mlir/lib/CAPI/Dialect/Rock.cpp
jerryyin: > I'd like to see us constructing [...]ParamsAttr as soon as we can inside our code

Okay agreed.

> The last one is needed when client wants to get serialized string to write the record back to the DB, so it's needed separately.

Now that I get a better sense of why you want to do this, can you modify the return type of the last function `mlirRockTuningGetParamStr` into a c-string? I see it less beneficial for client to use a `MlirStringRef` instead of having full control of the tuning parameters' life cycle. If they choose to store it and use it later, and if the reference is destroyed, they are run into trouble.

mlir/lib/CAPI/Dialect/Rock.cpp
```diff
@@ -0,0 +1,79 @@
...
+void mlirRockTuningSpaceDestroy(MlirRockTuningSpace params) {
+  delete (unwrap(params));
+}
+
+MLIR_CAPI_EXPORTED
+MlirStringRef mlirRockTuningGetParamStr(MlirRockTuningParam param) {
+  auto paramEntry = unwrap(param);
+  llvm::StringRef strRef = paramEntry->perfString;
+  return wrap(strRef);
+}
```
jerryyin: I'm marking this from previous comment. Return type should be a c-string, not a reference.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
jerryyin: You can *probably* ignore this after I have understood your intention in `Quick` tuning space which may be sorted and not linear, that's why we may have to expand this whole thing.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -0,0 +1,99 @@
...
+        newSpace->primaryOpType = op.getKernelType();
+        WalkResult::interrupt();
+      });
+
+  return newSpace;
+}
+
+bool tuningSetParam(ModuleOp &mod, ParamEntry *paramEntry) {
+  WalkResult setPrimary =
+      mod->walk([&](rock::RockGemmWrapperInterface op) -> WalkResult {
```
jerryyin, line 102: What will happen if there are more than one op that satisfy `rock::RockGemmWrapperInterface`? Say a module with 2 convolutions; a module with 2 gemms, or a module with a convolution and a gemm

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
jungpark-mlir: Thanks, and two 1's are required there

mlir/lib/CAPI/Dialect/Rock.cpp
jungpark-mlir: Yeah, that sounds better for the user to fully own the data.
I was just concerned about the allocation size but user can use strdup()

mlir/lib/CAPI/Dialect/Rock.cpp
jungpark-mlir: create tuning space always returns full range of sorted list. Let's say, full list has 100 entry and numHeuristicQuick=20. Client only needs to try 0~19 params in the tuning space.

krzysz00:
As a quick not, when we're adding this sort of header, we should `#define` a version, maybe `#define ROCMLIR_TUNING_VERSION 1` so we can change clients in ways that work with both older and newer commits of ours

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
```diff
@@ -243,11 +247,14 @@ def Rock_XdlopsGemmParamsAttr : Rock_Attr<"XdlopsGemmParams", [RockTuningParamAt
 
   let extraClassDeclaration = [{
     std::string getPerfConfigStr(){
-      const Twine paramsConcat = Twine(getMPerBlock()) + ","
+      const Twine paramsConcat = 
+          Twine(getMPerBlock()) + ","
         + Twine(getNPerBlock()) + ","
         + Twine(getKPerBlock()) + ","
+        + Twine(getKpack()) + ","
```
krzysz00: Note: I _think_ this is in the wrong order, and kPack comes after NPerWave? At least going off InitParamsXDL::visit()?

jungpark-mlir: You're right. I was confused with last but two but the last two was 1,1

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
jungpark-mlir: The first version only assumes we have one tunable operator in it. We'll consider multiple tunable ops in the future. For now, it just set the parameter to the  last tunable op  if it unexpectedly contains  multiple ops.

mlir/include/mlir-c/Dialect/Rock.h
jungpark-mlir: Enabling c api for a dialect opens the opportunity to register the dialect. With this, user can use generic mlir c api functions to create a module using Rock dialect IR.

In addition to that, we're adding tuning API functions this time and I believe we'll keep adding external APIs via c-api in the future.

jungpark-mlir: Had another look into the miir.
I think one day we need to migrate the implementation of miir*() functions under lib/Dialect/Rock and expose them via mlirRock*() here.

mlir/lib/Dialect/Rock/IR/RockGemmWrapperInterface.cpp
```diff
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-#include "mlir/Dialect/Rock/IR/RockGemmWrapperInterface.h"
+#include "mlir/Dialect/Rock/IR/Rock.h"
```
krzysz00, line 15: Why'd this need to change?

krzysz00 (changes requested):
I'm going to observe that we *still* don't validate both sides of the "no duplicate specification of the block size", and yes, I will be harping on that.

mlir/lib/Dialect/Rock/IR/RockGemmWrapperInterface.cpp
jungpark-mlir: Now we use GeneralGemmParamsAttr defined in RockAttr.td, only including RockGemWrapperInterface.h can't find it.
Rock.h includes header files in correct order.

krzysz00: Fair enough, makes sense

jungpark-mlir:
> 

Now, I have 2 strange verification failure, probably because something is missing related to that. Will try to fix it considering it.

jungpark-mlir:
> 1. I don't like the semantics of those interface methods
> 2. The validation on RockGemmWrapperInterface (or on its implementors, but you might as well put it on the interface) include something like
> 
> ```
> if ($_op.getDerivedBlockSize().has_value() && $_op.getParams().isa_and_nonnull<GeneralGridwiseParamsAttr>()) {
>   return $_op.emitOpError("Derived block size attribute set when block size is present in tuning parameters");
> ```

Added 2 in verify() for conv2D and gemm

mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
```diff
@@ -74,11 +74,12 @@ static void setUtilityKernelSizes(OpBuilder &b, Value arg, Operation *convOp,
...
   IntegerAttr blockSizeAttr = b.getI32IntegerAttr(blockSize);
   IntegerAttr gridSizeAttr = b.getI32IntegerAttr(gridSize);
-  convOp->setAttr("blockSize", blockSizeAttr);
+
+  // Tracking utility kernel block size separately.
+  convOp->setAttr("utilityBlockSize", blockSizeAttr);
   convOp->setAttr("gridSize", gridSizeAttr);
   convOp->setAttr("elems_per_thread", b.getIndexAttr(elemsPerThread));
 
-  funcOp->setAttr("block_size", blockSizeAttr);
```
krzysz00: Don't remove this. The utility kernel should have its own block size, fixed at 64, as previously written. In the alternative, the grid size needs to be scaled down.

It is an invariant that the block size on the function matches the block size on the ops.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -499,6 +499,14 @@ template <typename T> static LogicalResult verifyConvOp(T op) {
       isDisjointed("input_layout", "hi", "wi"))
     return op.emitError("Disjointed yx or hw!");
 
+  bool isXdlops = bitEnumContainsAll(
+      op->getAttr("features").template cast<GemmFeaturesAttr>().getValue(),
+      GemmFeatures::mfma);
+  if (op->hasAttr("derivedBlockSize") && !isXdlops) {
```
krzysz00: Instead of `hasAttr` by string, maybe use a higher-level call like `derivedBlockSizeAttr()` and checking for `none`

krzysz00: I'll also note that I'd prefer the different condition where the error is to have `derivedBlockSize` and a tuning parameter attribute that has a `GeneralGridwiseGemmParams`

jungpark-mlir: It's Operation* here, can't cast it either.

krzysz00: I'd argue for rewriting this to a combination of the conv interface and the gemm wrapper interface (especially since we can safely cast from the conv interface to the gemm interface without worrying much because we know the deal)

jungpark-mlir: I tried it first using 
RockConvInterface convInterface = cast<RockConvInterface>(op);
but compiler still treated convInterface as Operation* and threw error.

Anything  else I can try?

krzysz00: Make the function argument `RockConvInterface`. If you need the specific operation, you're looking for `getOperation()` on the interface

mlir/include/mlir/Dialect/Rock/IR/RockConvInterface.td
```diff
@@ -43,6 +43,13 @@ def RockConvInterface : OpInterface<"RockConvInterface"> {
         /*args=*/(ins),
         /*methodBody=*/"",
         /*defaultImplementation=*/[{ return $_op.getDilations(); }]>,
+    InterfaceMethod<
+        /*desc=*/[{Return features attr values.}],
+        /*retType=*/"GemmFeatures",
+        /*methodName=*/"getFeatures",
+        /*args=*/(ins),
+        /*methodBody=*/"",
+        /*defaultImplementation=*/"">,
```
krzysz00, line 52: I think you should actually set a default implementation here.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -98,7 +99,8 @@ def Rock_Conv2DBwdWeightOp : Rock_Conv2DOpBase<"conv2d_bwd_weight">
 {
   dag additionalArgs = (ins Optional<TensorOrMemRefRankOf<[F32], [5]>>:$workspace,
                             OptionalAttr<IndexAttr>:$kBlocks,
-                            I32Attr:$numCu);
+                            I32Attr:$numCu,
+                            OptionalAttr<I32Attr>:$utilityBlockSize);
```
krzysz00, line 103: Why are we doing this instead of setting `derivedBlockSize`?

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -663,6 +670,11 @@ LogicalResult GemmOp::verify() {
     return emitOpError("general kernels don't support non-set store methods");
   }
 
+  if (getDerivedBlockSize().has_value() && !isXdlops) {
```
krzysz00, line 678: Again, slightly not the right validation. `getDerivedBlockSize().has_value() && hasGeneralGemmParams`
where `hasGeneralGemmParams = getParams.isa_and_nonnull<General...ParamsAttr>()`

krzysz00 (commented):
A few more comments.

mlir/include/mlir/Dialect/Rock/IR/RockConvInterface.td
jungpark-mlir: 'features' is necessary for the conv ops and all of them should have their getFeatures(), we shouldn't have default here.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
jungpark-mlir: This is yet another block size only for the utility kernel. I want to make it sure they are not confused.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
jungpark-mlir: OK, just understood what you mean, it also can be forced to set. That makes better sense.

mlir/include/mlir/Dialect/Rock/IR/RockConvInterface.td
jungpark-mlir: Ideally, having default impl here makes better sense but getFeatures()  per  ops  are all automatically generated by the tablegen.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
krzysz00: I mean that the condition is not "we set derived block size on a non-xdlops kernel", it's "we set both the derived block size and included tuning parameters that contain a block size"

Because non-xdlops utility kernels will set the derived block size and delete the tuning parameters

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
krzysz00: I suppose ... we definitely need a ticket to define separate ops for the utility kernels

jungpark-mlir: > I suppose ... we definitely need a ticket to define separate ops for the utility kernels

https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/730

krzysz00:
I think, the changes to the verification condition aside, this is in a pretty good state. @jerryyin ?

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
jungpark-mlir: That reminds me I reverted deleting params attr but found correct place to delete it.

jungpark-mlir: Wait, this is still correct. An utility kernel itself can't be a gemmOp, it's lowered in `convToGemm`. Correct derivedBlockSize is used for the actual gemmOp, not utilityBlockSize.

krzysz00: I want the if statement to test what we actually care about, not a proxy for it.

The test is "is it the case that we have set both the derived block size and set a tuning parameters attribute that specified a block size". `isXdlops` has nothing to do with it.

jungpark-mlir: Right, that could be a possible error case, added it.
I wanted to check the case derived blocksize attr is set in non-xdlops path, which is a separate case.

krzysz00: Yeah, fair

krzysz00 (approved):
Seems reasonable to me at this point, approved

jungpark-mlir:
@jerryyin Could you please take a look? I've tried to address your comments in the code or answer.

mlir/lib/CAPI/Dialect/Rock.cpp
jerryyin: We can address this later and I can see you can re-arrange the sequence of tuning space to make this work.

mlir/include/mlir-c/Dialect/Rock.h
jerryyin: I think this is going to bring about some interesting discussion. Maybe we should have a meeting offline.

mlir/lib/CAPI/Dialect/Rock.cpp
jerryyin: If it is the standard practice then it is okay. It doesn't look like from the link, it is dereferencing pointer without a check.

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
```diff
@@ -211,9 +213,11 @@ def Rock_GeneralGemmParamsAttr : Rock_Attr<"GeneralGemmParams", [RockTuningParam
...
     std::string getPerfConfigStr(){
-      const Twine paramsConcat = Twine(getMPerBlock()) + ","
-      + Twine(getNPerBlock()) + ","
+      const Twine paramsConcat =
+        Twine(getBlockSize()) + ","
+      + Twine(getMPerBlock()) + ","
       + Twine(getNPerBlock()) + ","
+      + Twine(getKPerBlock()) + ","
       + Twine(getMPerThread()) + ","
       + Twine(getNPerThread());
```
jerryyin, line 222: Please adjust the sequence of this serialization so tuning parameters can be interchangeably used and compared.

https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/7898c00ad5c4aae24add7c276123ba3be01c6eff/mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h#L60-L65

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
```diff
@@ -243,11 +247,14 @@ def Rock_XdlopsGemmParamsAttr : Rock_Attr<"XdlopsGemmParams", [RockTuningParamAt
...
-      const Twine paramsConcat = Twine(getMPerBlock()) + ","
+      const Twine paramsConcat = 
+          Twine(getMPerBlock()) + ","
         + Twine(getNPerBlock()) + ","
         + Twine(getKPerBlock()) + ","
+        + Twine(getKpack()) + ","
         + Twine(getMPerWave()) + ","
-        + Twine(getNPerWave()) + ",1,1";
+        + Twine(getNPerWave()) + ","
+        + "1,1"; /* *ThreadCopyMore* */
```
jerryyin: Same with previous comment: Adjust the sequence to make them compatible.

https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/7898c00ad5c4aae24add7c276123ba3be01c6eff/mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h#L88-L96

mlir/test/CAPI/mixr_full.c
```diff
@@ -165,6 +166,24 @@ static bool constructAndTraverseIr(MlirContext ctx) {
...
+  MlirRockTuningParam tuningParam = mlirRockTuningParamCreate();
+  if (!mlirRockTuningParamGet(tuningSpace, 0, tuningParam)) {
+    printf("fails to obtain param\n");
+    return false;
+  }
+  mlirRockTuningSetParam(module, tuningParam);
+  char *paramStr = strdup(mlirRockTuningGetParamStr(tuningParam));
+  mlirRockTuningParamDestroy(tuningParam);
+  mlirRockTuningSpaceDestroy(tuningSpace);
+  printf("Obtained perfconfig : \"%s\"\n", paramStr);
```
jerryyin, line 184: I remember this was broken in the demo. I just want to double check that this is fixed?

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
jungpark-mlir: Both match with the visit?

mlir/test/CAPI/mixr_full.c
jungpark-mlir: Yeah, it's fixed now. I printed after freeing the cstring.

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
jerryyin: I misread on this one.

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
jerryyin: Please fix position of kPack

mlir/test/CAPI/mixr_full.c
jerryyin: Gotcha

jerryyin (approved):
Please fix the kpack position and this is ready to go. Thanks for you effort!

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
jungpark-mlir: Ah, this should be from bad merge, thanks.

PR #815, opened 2022-10-14
Use forceUnroll in lowering GridwiseGemmV2Op

yiqian1:
Enable a new tuning parameter to force unroll when creating TransformingFor loops in GridwiseGemmV2 lowering.

We create TransformingFor loop at other places, but the preliminary experiments showed turning forceUnroll on/off in GridwiseGemmToBlockwise.cpp had the most impact on performance.

MIOpen changes are [here](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1825).


mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
```diff
@@ -75,7 +75,7 @@ struct TransformingForRewritePattern
 
     bool useDiffs = op.useIndexDiffs().getValueOr(false);
     bool unroll = op.forceUnroll().getValueOr(false);
-
+llvm::errs()<<"\nSugarToLoops: useDiffs="<<useDiffs<<"; unroll="<<unroll<<"\n";
```
krzysz00: LLVM_DEBUG()

krzysz00 (commented):
I new see why you lost your perf wins 

First, those loops are only applicable in the non-xdlops case at the moment.

Second, in your experiments, you applied these parameters to _all_ `transforming_for` loops

To replicate your perf numbers, you'll need to go through and weave the tuning parameter in everywhere - every last `create<rock::TransformingFor>` in the setup.

You can probably get away without doing some of the really small ones, mainly the ones we create during fusion - but there's a bunch of these other loops: they're in blockwise gemm, in threadwise gemm, and so on and so forth.

yiqian1:
> First, those loops are only applicable in the non-xdlops case at the moment.

I'm pretty sure those loops are for xdlops. 

> Second, in your experiments, you applied these parameters to all transforming_for loops

In my manual experiments, I also tested those parameters separately for transformingFor loops in different files. Turning them off in thredwiseGemm actually made the performance worse. Those parameters when applied to global load loops or store loops made the most impact on performance (performance results similar to those when I applied them to all transformingFor loops). That's why I started with gridwise gemm.

krzysz00:
Ok, yeah, you did get the xdlops ones, and thanks for the info.

But also, it looks like your branch is from an old upstream, so your "perf loss" could just be the fact that Manupa and Jerry have gotten us a uniform +30% with their PRs. Try all this again after you've done a `git rebase miopen-dialect`? (or a `git rebase -i miopen-dialect` if you want to clean things up)

yiqian1:
Yeah, I didn't update my branch for quite some time. I'll rebase and try again.

krzysz00:
Other thought, because I noticed it when I initially wrote the issue down: make sure to also do the ThreadwiseCopyV2 loop

yiqian1:
> Other thought, because I noticed it when I initially wrote the issue down: make sure to also do the ThreadwiseCopyV2 loop

I don't see ThreadwiseCopyV2 using forceUnroll or index diff maps. Could you please let me know where I should look at?

krzysz00:
You should be looking near the bottom of GrihwineGemmToBlockwise, the loop that constructs a `ThreadwiseCopyV2Op`.

yiqian1:
> You should be looking near the bottom of GrihwineGemmToBlockwise, the loop that constructs a `ThreadwiseCopyV2Op`.

Yeah, I've done that.

yiqian1:
Here is my latest results of including these two tuning parameters:
MIOpen TFlops (with tuned MLIR) 33.5 vs 29.4 in our nightly;
MIOpen (with tuned MLIR)/MIOpen (with untuned MLIR) 1.39 vs 1.32 in our nightly.

http://rocmhead.amd.com:8080/job/MLIR/job/test-perf/49/Performance_20report/MLIR_vs_MIOpen.html

MIOpen tuning takes 17 hours on gfx908, http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Ftest-weeklyCI/detail/test-weeklyCI/28/pipeline/171
Our weekly spends 4 hours tuning on gfx908.

jerryyin:
@yiqian1 Are both parameters equally important in your experiment? How much do each of them contribute in the experiment?

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
krzysz00: Bump

mlir/utils/performance/perfRunner.py
```diff
@@ -400,7 +400,7 @@ def benchmarkExternal(cls, commandLine, paths: Paths, arch, envs=dict()):
         nanoSeconds = np.nan
         try:
             outs, errs = p1.communicate(timeout=300)
-            if len(errs) > 0:
+            if False and len(errs) > 0:
```
krzysz00: ??

mlir/utils/jenkins/MIOpen_tuned_fwd_vs_original_tuned.html
```diff
@@ -0,0 +1,1433 @@
+
```
krzysz00: Delete

krzysz00 (commented):
Can you add a test to make sure the tuning parameter you added does what you want?

Also, can you make sure the heuristic tuning parameters set forceUnroll = true, useIndexDiffs = true, ...?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -692,10 +692,10 @@ struct GridwiseGemmRewritePattern : public OpRewritePattern<GridwiseGemmOp> {
 
     TransformingForOp blockwiseLoadA =
         createGlobalLoadLoop(b, loc, wrappedA, aVectorGlobalMap, aCopyPerThread,
```
krzysz00, line 694: Why's this `true`? Shouldn't it be the tuning parameter?

krzysz00: ... ah, this is non-xdlops

krzysz00 (commented):
Can you check the `InitParamsXDL` over in affix-tuning-parameters to make sure that forceUnroll is defaulting to true, and add a test to catch that?

yiqian1:
> Can you add a test to make sure the tuning parameter you added does what you want?
> 

I need to think about how to add a simple test for this PR.

> Also, can you make sure the heuristic tuning parameters set forceUnroll = true, useIndexDiffs = true, ...?

krzysz00:
I missed the commit where you addressed that heuristic thing.

And the test will be usually updated "naturally" because there's tests for affix-tuning-parameters.

You can also have a test where you run `--rock-gridwise-gemm--to-blockwise` with the forceUnroll parameter either true or false and then `CHECK`/`CHECK-NOT` to see if the loops got `forceUnroll`

*Also*, your PR isn't setting forceUnroll on the GEMM writeback loop that's created in gridwise-gemm-to-blockwise?

krzysz00 (approved):
Ok, so, overall, LGTM and I'm happy landing this.

However,
1. Can just `CHECK` for a transforming_for with the right properties? I'd rather our tests not be sensitive to the exact number of loops we output in gridwise to blockwise
2. Have you confirmed that none of the existing tests (ex. the ones that test heuristic mode) break now that you've added this extra parameter to the xdlops parameters struct. That is, you might want to run `ninja check-rocmlir`

PR #816, opened 2022-10-14
Delete mlir-rocm-runner, moving its pipeline into rocmlir-driver

krzysz00:
- Add --host-pipeline=runner to rocmlir-driver, which does the [gpu and a bunch of stdlib dialects]->llvm steps to make host code and which compiles gpu kernels to HSA code objects
- Make -c mean "-kernel-pipeline full -host-pipeline runner" instead of "-kernel-pipeline gpu"
- This means that (with the exception of xmir-runner, which needed to do some runtime stuff to pick the target), we no longer have "runners" that do a bunch of non-trivial work on their input. An advantage of this is that benchmarking now doesn't include the time it takes to compile a GPU kernel (which it might have before)
- Adjust the xmir runner pipeline to include the arithmetic ops expansion pass that was needed for host code validation, and rearrange it some to unify it with the old mlir-rocm-runner
- Delete mlir-rocm-runner
- Adjust tests, mainly
  - Switching all instances of mlir-rocm-runner to mlir-cpu-runer -O2
  - Change the single-target verifier=clone tests to use the runner pipeline because they don't need the xmir functionality
  - Change the convolution harness tests to expect -c to produce a binary blob and not a gpu.module
  - Change various CPU e2e tests to include the relevant --convert-*-to-llvm invocations or to go through the runner pipeline
- Change the CI perf/parameter sweep scripts accordingly
- Change the rocm-run widget to pick up mlir-cpu-runner

This also means that we now no longer have a `hipGetDeviceProperties()` call lurking anywhere in our tests except the initial calls to `rocm_agent_enumerator` and the xmir stuff.


sjw36 (approved):
Good consolidation.

zhanglx13:
Are there any other benefits doing this besides benchmarking? 
I am inexperience in this area and maybe this is the reason that I find it weird to call `mlir-cpu-runner` to execute GPU kernels.

krzysz00:
`mlir-cpu-runner` isn't executing GPU kernels, exactly, it's executing host code, which then just so happens to launch some GPU kernels.

And the advantages to this is that now, all the IR transformation happens in one place, making it easier to hook in to. We no longer have "oh and then `mlir-rocm-runner` runs a whole bunch of passes *and* looks up which GPU you're running on"

mlir/test/e2e/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ set(ROCK_E2E_DEPENDS
   rocmlir-opt
   rocmlir-gen
   rocmlir-driver
-  mlir-rocm-runner
+  mlir-cpu-runner -O2
```
pcf000: Should the "-O2" be here?

krzysz00:
Upstream part has landed https://reviews.llvm.org/D136105

PR #817, opened 2022-10-14
Define a KernelOpType enum to prepare for gemm in rocmlir-gen

krzysz00:
This also clears up some stuff for tuning


jungpark-mlir (approved):
Looks good

PR #818, opened 2022-10-17
[HOTFIX] Relax the matrix size check on rock.gemm

krzysz00:
In some convolution configs, a tensor of size < 2 GB could be viewed as a matrix of size >= 2 GB because of padding, strides, or other convolution parameters. The previous applicability logic did not account for this possibility, but the configs worked anyway, and so the check may have been too strict.

Since we are currently much less likely to use signed math for large indices due to the integer range inference pass, and since our main uses of signed numbers are in mod() (which will have arguments generally smaller than the tensor size), it is, in practice, safe to proceed with matrix sizes below 4 GB.


jerryyin (approved):
LGTM, thank you for investigating into it.

PR #819, opened 2022-10-18
Enabling lld as a library invokation

jerryyin:
For the most part, this PR reverts #504. After this PR, MLIR won't be dependent on a LLD binary on a particular place of the OS. Key changes:
 - lld::CommonLinkerContext::destroy();
   - This is brough from COMGR and will destroy the linker state after linking
 - // cl::ResetAllOptionOccurrences();
   - This reverted the change from LLD to erase all cl::cmd options, causing `JitRunner` to malfunction
   - However, this is likely make the change un-upstreamable.

### DIFF
--- a/cmake/llvm-project.cmake
+++ b/cmake/llvm-project.cmake
@@ -9,7 +9,7 @@ set(MLIR_CMAKE_CONFIG_DIR
 set(MLIR_TABLEGEN_EXE mlir-tblgen)
 
 # LLVM settings
-set(LLVM_ENABLE_PROJECTS "mlir" CACHE STRING "List of default llvm targets")
+set(LLVM_ENABLE_PROJECTS "mlir;lld" CACHE STRING "List of default llvm targets")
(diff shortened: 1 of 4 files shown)


krzysz00 (approved):
Yeah, this looks fine here, but we probably can't sent it upstream 

... you might want to let the comgr folks know about the options thing for when they next do an upstream sync.

Though, come to think it ... I wonder if we could ask for the very specific thing we want (object to shared object) to be exposed as a library call in `lld` ...

jerryyin:
> ... you might want to let the comgr folks know about the options thing for when they next do an upstream sync.

That issue in particular has to do with command line options, it won't affect them. It affect us because our JitRunner rely on the command line options.

> Though, come to think it ... I wonder if we could ask for the very specific thing we want (object to shared object) to be exposed as a library call in lld ...

Yes, I think that worth the exploration if it get exposed in a more public way.

krzysz00:
To rephrase, I'd suggest going to the LLD folks and asking them if they'd be willing to add such an API

PR #822, opened 2022-10-19
Refactor pad/stride/dilation into a convParam attribute

giuseros:
- Create a ConvParam attributes to group conv-like parameters (for now just padding, strides and dilation)
- Create a RockConv interface to expose the the `rock.conv*` operations in a more uniform way
- Change the tests to use the new attribute


giuseros:
This fixes: https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/691

krzysz00 (commented):
So I'm going to be annoying here and suggest that padding, dilation, and strides should be three separate attributes, instead of defining a new struct. Also, I'd rather not put said new struct into the assembly format, just let it hang out in `attr-dict`.

(Although I don't have massively strong objections to doing it your way - I'd just rather not have an extra layer of indirection, especially since we can have

```
class I32ArrayLength<int n> : Confined<DenseI32ArrayAttr, ...>`

...
$padding: I32ArrayLength<4>
$strides:I32ArrayLength<2>
...
```

I tossed a struct together for tuning parameters beca

krzysz00:
I'll also note you can define a class for convolutions and use the `!con` operator and some template parameters to abstract out the Conv2D* parameter lists

giuseros:
Hi @krzysz00 , 
Thanks a lot for your comments! I also don't have any strong opinions,  so I moved things in the way you proposed. See if now it matches your preference

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -44,19 +45,28 @@ class TensorOrMemRefRankOf<list<Type> allowedTypes, list<int> ranks> :
              /*summary=*/"",
              "::mlir::ShapedType">;
 
+class I32ArrayLength<int n> : ConfinedAttr<I32ArrayAttr, [ArrayCount<n>]>;
 
-def Rock_Conv2DOp :
-    Rock_Op<"conv2d", [DeclareOpInterfaceMethods<RockGemmWrapperInterface>]>,
-    Arguments<(ins TensorOrMemRefRankOf<[F32, F16, BF16, I8], [5]>:$filter,
+class Rock_Conv2DOpBase<string mnemonic> :
```
krzysz00: Note: only forward convolution support I8.

Note 2: could that additional argument list be a template argument (of type `dag` to `Conv2DOpBase`?

krzysz00: (or, alternatively, explicitly ::verify() out int8 in non-forward convolutions

mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
```diff
@@ -858,6 +841,23 @@ LogicalResult Conv2dGenerator::genConvModule(ModuleOp &module, int kernel_id,
   attributes.push_back(builder.getNamedAttr(
       "features", builder.getAttr<GemmFeaturesAttr>(config.features)));
 
+  auto paddingArray = SmallVector<int32_t, 4>{
```
krzysz00: Is there any advantage to doing this style over `SmallVector<int32_t, 4> paddingArray = {...}`?

mlir/include/mlir/Dialect/Rock/IR/RockConvInterface.h
```diff
@@ -0,0 +1,27 @@
...
+
+#include "mlir/IR/OpDefinition.h"
+
+#include "mlir/Dialect/Rock/IR/RockTypes.h"
+
+#include "mlir/Dialect/Utils/StaticValueUtils.h"
+
+#include "mlir/Dialect/Rock/IR/RockConvInterface.h.inc"
+
+#endif // MLIR_DIALECT_ROCK_IR_ROCKCONVINTERFACE_H
```
krzysz00: Nit: Missing trailing newline

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -44,19 +45,28 @@ class TensorOrMemRefRankOf<list<Type> allowedTypes, list<int> ranks> :
              /*summary=*/"",
              "::mlir::ShapedType">;
 
+class I32ArrayLength<int n> : ConfinedAttr<I32ArrayAttr, [ArrayCount<n>]>;
```
krzysz00, line 48: Question: why not `DenseI32ArrayAttr`?

krzysz00 (commented):
I'm really liking this refactor, I'm mostly in the nitpick zone now.

Also, ... I think you forgot to add `staticValueUtils`

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
giuseros: I didn't have a reason, but now I found out that `ConfinedAttr` only works with  a subclass of `ArrayAttr` . So I cannot confine a dense array

krzysz00 (approved):
Nice refactor! I might end up fiddling with the ArrayAttr/DenseArrayAttr question in a future PR, but that'll probably be during some sort of more general cleanup

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
krzysz00: Huh, now we know. Cool!

PR #823, opened 2022-10-19
[mlir][AMDGPU] Update amdgpu.lds_barrier implementation

krzysz00:
On some architectures (currently gfx90a, gfx94*, and gfx10**), we can implement an LDS barrier using compiler intrinsics instead of inline assembly, improving optimization possibilities and decreasing the fragility of the underlying code.

Other AMDGPU chipsets continue to require inline assembly to implement this barrier, as, by the default, the LLVM backend will insert waits on global memory (s_waintcnt vmcnt(0)) before barriers in order to ensure memory watchpoints set by debuggers work correctly.

Use of amdgpu.lds_barrier, on these architectures, imposes a tradeoff between debugability and performance. The documentation, as well as the generated inline assembly, have been updated to explicitly call attention to this fact.

For chipsets that did not require the inline assembly hack, we move to the s.waitcnt and s.barrier intrinsics, which have been added to the ROCDL dialect. The magic constants used as an argument to the waitcnt intrinsic can be derived from llvm/lib/Target/AMDGPU/Utils/AMDGPUBaseInfo.cpp .

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
@@ -425,6 +425,11 @@ def AMDGPU_LDSBarrierOp : AMDGPU_Op<"lds_barrier"> {
     to complete before execution continues. Therefore, it should be used when
     operations on global memory can be issued far in advance of when their results
     are used (for example, by writing them to LDS).
+
+    WARNING: On architectures that do not support the BackOffBarrier feature,
+    (those which will implement this barrier by emitting inline assembly),
+    use of this operation will impede the usabiliity of memory watches (including
+    breakpoints set on variables) when debugging.
   }];
   let assemblyFormat = "attr-dict";
 }
--- a/external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
+++ b/external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
@@ -270,21 +270,55 @@ struct RawBufferOpLowering : public ConvertOpToLLVMPattern<GpuOp> {
 };
 
 struct LDSBarrierOpLowering : public ConvertOpToLLVMPattern<LDSBarrierOp> {
-  using ConvertOpToLLVMPattern<LDSBarrierOp>::ConvertOpToLLVMPattern;
+  LDSBarrierOpLowering(LLVMTypeConverter &converter, Chipset chipset)
+      : ConvertOpToLLVMPattern<LDSBarrierOp>(converter), chipset(chipset) {}
+
+  Chipset chipset;
 
   LogicalResult
   matchAndRewrite(LDSBarrierOp op, LDSBarrierOp::Adaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    auto asmDialectAttr = LLVM::AsmDialectAttr::get(rewriter.getContext(),
-                                                    LLVM::AsmDialect::AD_ATT);
-    const char *asmStr = "s_waitcnt lgkmcnt(0)\ns_barrier";
-    const char *constraints = "";
-    rewriter.replaceOpWithNewOp<LLVM::InlineAsmOp>(
-        op,
-        /*resultTypes=*/TypeRange(), /*operands=*/ValueRange(),
-        /*asm_string=*/asmStr, constraints, /*has_side_effects=*/true,
-        /*is_align_stack=*/false, /*asm_dialect=*/asmDialectAttr,
-        /*operand_attrs=*/ArrayAttr());
+    bool requiresInlineAsm =
+        chipset.majorVersion < 9 ||
+        (chipset.majorVersion == 9 && chipset.minorVersion < 0x0a) ||
+        (chipset.majorVersion == 11);
+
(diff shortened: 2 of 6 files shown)


external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -259,21 +263,34 @@ struct RawBufferOpLowering : public ConvertOpToLLVMPattern<GpuOp> {
 };
 
 struct LDSBarrierOpLowering : public ConvertOpToLLVMPattern<LDSBarrierOp> {
-  using ConvertOpToLLVMPattern<LDSBarrierOp>::ConvertOpToLLVMPattern;
+  LDSBarrierOpLowering(LLVMTypeConverter &converter, Chipset chipset)
```
jungpark-mlir, line 273: We tried this before but there is a long standing performance issue regarding this.
I don't think this has been fixed yet.

jungpark-mlir (changes requested):
Please hold this until we can confirm the known issue is fixed.

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
krzysz00: Wasn't that perf issue to do with `vmmcnt(0) & lgkmcnt(0)`?

krzysz00: But yeah, I can perf this

jungpark-mlir: The problem is backend could insert vmcnt(0) where only lgkmcnt(0) was given.

krzysz00: I looked, and it doesn't seem to be there. I can run perf to confirm, but still.

krzysz00: I'll note this isn't the barrier with a fence, this is the intrinsics that specifically cause s_waitcnt and s_barrier

jungpark-mlir: It used to happens on the first LDS wait in the main loop of the xdlops gemm.
The loop starts with a global load which is independent to the LDS data, deliberately try to overlap global/LDS latency.

krzysz00: Ok, yeah, the weirdness came back, I did check that exact spot. I'm going to go while an the compiler team

krzysz00:
To pass the perf data, which is wonky because 112 but doesn't show universal problems, 
[perf-no-inline-asm.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/9832453/perf-no-inline-asm.zip)

krzysz00:
(and actually shows perf gains from not having inline asm in there, but I'm not sure if that's because the machine was under load or what)

arsenm:
Should add a fixme that this still isn't really what we want. This should really be a fence with an attached set of address spaces

krzysz00:
This also isn't working due to extra vmmcnt(0) getting inserted, which there's an internal ticket for

jungpark-mlir:
> Should add a fixme that this still isn't really what we want. This should really be a fence with an attached set of address spaces

Do we have local address space fence now? It was discussed but not sure it's done already.
Agreed to use that if it's supported.

krzysz00:
See https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/701 for the pointer to the underlying compiler ticket

arsenm:
> > Should add a fixme that this still isn't really what we want. This should really be a fence with an attached set of address spaces
> 
> Do we have local address space fence now? It was discussed but not sure it's done already. Agreed to use that if it's supported.

No, that's the feature we need. The direct waitcnt usage is the workaround for not having it

krzysz00:
Ok, so, to update, further investigation reveals that the compiler will still slide in `vmcnt(0)`s on certain architectures (gfx908, gfx1100), and that there's no reason to think this won't be bad for perf.

I've raised the question of whether we could disable this behavior with the compiler team.

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
```diff
@@ -425,6 +425,11 @@ def AMDGPU_LDSBarrierOp : AMDGPU_Op<"lds_barrier"> {
     to complete before execution continues. Therefore, it should be used when
     operations on global memory can be issued far in advance of when their results
     are used (for example, by writing them to LDS).
+
+    WARNING: On architectures that do not supporth the BackOffBarrier feature,
```
skyreflectedinmirrors: 'supporth'

krzysz00:
@arghdos @arsenm Ping for compiler review

krzysz00:
Should we go ahead and land this?

sjw36 (approved):
Looks good. Let's land if it passes nightly and no perf implications.

PR #824, opened 2022-10-19
[DO NOT SQUASH] Formalize XModel attributes and pipelines

sjw36:
XMIR constructs have been ad-hoc and pipelines/passes were intermingled with unrelated Rock dialect.

This PR formalizes into an XModel dialect with:
* TargetObjectAttr - formal container for binarys
* KernelPackageAttr - formal spec for kernel launch spec
* Pipelines for graph,package,runner
   * graph - TOSA passes, including partition and target clones
   * package - Collect target implementations per kernel func
   * runner - Select target implementations and run
    
    https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/665


external/llvm-project/mlir/include/mlir/Dialect/XModel/IR/XModelAttrDefs.td
```diff
@@ -0,0 +1,112 @@
...
+        ```
+        where the <type> is one of 'ELF' or 'SPIRV'. And where <arch> 
+        specifies what target-arch the object is compiled for in the case
+        of an 'ELF' object. And where the <binary> is a string stored
+        representation of the actual binary.
+    }];
+
+    let parameters = (ins
+      StringRefParameter<"The object type">:$type,
+      StringRefParameter<"The object type">:$arch,
```
krzysz00: Duplicate description.

external/llvm-project/mlir/include/mlir/Dialect/XModel/IR/XModelAttrDefs.td
```diff
@@ -0,0 +1,112 @@
...
+        #xmodel.target_obj<<type> = <arch> {<attributes>} -> <binary>>
+        ```
+        where the <type> is one of 'ELF' or 'SPIRV'. And where <arch> 
+        specifies what target-arch the object is compiled for in the case
+        of an 'ELF' object. And where the <binary> is a string stored
+        representation of the actual binary.
+    }];
+
+    let parameters = (ins
+      StringRefParameter<"The object type">:$type,
```
krzysz00: Why not have this be a xmodel::TargetType?

external/llvm-project/mlir/include/mlir/Dialect/XModel/IR/XModelAttrDefs.td
```diff
@@ -0,0 +1,112 @@
...
+        #xmodel.kernel_pkg<'type' = 'target' : 'kernel_name' ['launch_dims'] {attributes} -> #xmodel.target_obj<>>
+        ```
+        where the params block may be omitted for transforms that do not take parameters.
+
+    }];
+
+    let parameters = (ins
+      AttrParameter<"::mlir::xmodel::TargetType", "The target IP type">:$type,
+      StringRefParameter<"The target name, redundant with target_object">:$target,
+      StringRefParameter<"The name of the function reference">:$kernel_name,
```
krzysz00, line 103: Couldn't/shouldn't this be a symbol reference?

external/llvm-project/mlir/include/mlir/Dialect/XModel/IR/XModelAttrDefs.td
```diff
@@ -0,0 +1,112 @@
...
+      ArrayRefParameter<"uint32_t", "The launch dimensions">:$launch_dims,
```
krzysz00, line 104: I'd suggest this should live on the `target_object` as an attribute for GPU ones

external/llvm-project/mlir/include/mlir/Dialect/XModel/Transforms/Passes.td
```diff
@@ -0,0 +1,38 @@
...
+
+def XModelAsyncGraphPass : Pass<"xmodel-async-graph", "func::FuncOp"> {
+  let summary = "convert kernel func call ops to async.launch ops with dependencies";
+  let dependentDialects = ["xmodel::XModelDialect", "async::AsyncDialect"];
+}
+
+def XModelSelectTargetsPass : Pass<"xmodel-select-targets", "func::FuncOp"> {
+  let summary = "Select runtime target for execution";
+  let options = [
+    ListOption<"targetTypes", "target-types", "std::string", "List of target types">,
```
krzysz00, line 33: Shame we can't make this enum-valued ... or can we?

external/llvm-project/mlir/include/mlir/Dialect/XModel/Transforms/Passes.td
```diff
@@ -0,0 +1,38 @@
...
+    ListOption<"targetChips", "chips", "std::string", "List of chip architectures">
```
krzysz00: "chip" or "architecture"? I lean heavily towards `arch` here.

And where do these names get canonicalized, if there is such a thing? 

That is, which of these are acceptable inputs here?
- `gfx908`
- `gfx908:sramecc+:xnack-`
- `amdgcn-amd-amdhsa:gfx908`
bearing in mind that a binary compiled for `gfx908` is acceptable to `gfx908:sramecc+:xnack-` but not the other way around, but that I can't say if "dropping features makes your binary more generic" is true everywhere

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
```diff
@@ -64,7 +64,7 @@ static Optional<DictionaryAttr> getGPUTarget(async::LaunchOp op) {
   if (!func.has_value() || func->getNumResults() != 0)
     return llvm::None;
 
-  auto attr = (*func)->template getAttrOfType<ArrayAttr>("async.targets");
+  auto attr = (*func)->template getAttrOfType<ArrayAttr>("xmodel.targets");
```
krzysz00, line 67: Silly question: would this pipeline make more sense as XmodelToGPU?

external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp
```diff
@@ -196,6 +196,12 @@ SerializeToHsacoPass::translateToLLVMIR(llvm::LLVMContext &llvmContext) {
     return ret;
   }
 
+  gpu::GPUModuleOp gpuMod = getOperation();
```
krzysz00: Note that I can't believe I missed until now: this should be in `serializeToBlob` because there's nothing AMD-specific about it.

Also, why do we need this at all?

external/llvm-project/mlir/lib/Dialect/XModel/IR/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+  MLIRXModelAttrDefsIncGen
+  )
+target_link_libraries(MLIRXModel
+  PUBLIC
+  MLIRGPUOps
+  MLIRIR
+  MLIRInferIntRangeInterface
+  MLIRDialect
+  MLIRAnalysis
+  MLIRExecutionEngine
```
krzysz00: Don't depend on ExecutionEngine, we'll get complained at.

ExecutionEngineUtils, I think, is the one that probably has what you want.

external/llvm-project/mlir/lib/Dialect/XModel/IR/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+  MLIRParser
+  MLIRROCDLDialect
```
krzysz00: Why do we depend on ROCDL?

external/llvm-project/mlir/lib/Dialect/XModel/IR/XModel.cpp
```diff
@@ -0,0 +1,264 @@
...
+//===---------------------------------------------------------
+mlir::Attribute TargetObjectAttr::parse(mlir::AsmParser &parser,
+                                        mlir::Type type) {
+  llvm::SMLoc startLoc = parser.getCurrentLocation();
+  if (parser.parseLess()) {
+    return {};
+  }
+
+  std::string typeName;
+  if (parser.parseKeywordOrString(&typeName)) {
```
krzysz00, line 52: I'd then run this through the string->enum function and error if that fails

external/llvm-project/mlir/lib/Dialect/XModel/IR/XModel.cpp
```diff
@@ -0,0 +1,264 @@
...
+  if (parser.parseOptionalAttrDict(attrList)) {
+    return {};
+  }
+  DictionaryAttr attrs = attrList.getDictionary(parser.getContext());
+
+  if (parser.parseArrow()) {
+    return {};
+  }
+
+  std::string binary;
```
krzysz00, line 82: Misc questions.: is the resource system that got poked around upstream actually landed? If so, is it possible that this could be either a string attribute or one of said resources? Is there a way to do that?

external/llvm-project/mlir/lib/Dialect/XModel/IR/XModel.cpp
```diff
@@ -0,0 +1,264 @@
...
+// #xmodel.target_kernel<'type' = 'arch' : 'entry_name' ['launch_dims']
+// {attributes} -> #xmodel.target_obj<>>
+//===---------------------------------------------------------
+void KernelPackageAttr::walkImmediateSubElements(
+    llvm::function_ref<void(mlir::Attribute)> walkAttrsFn,
+    llvm::function_ref<void(mlir::Type)> walkTypesFn) const {
+  walkAttrsFn(getObject());
+}
+Attribute
+KernelPackageAttr::replaceImmediateSubElements(ArrayRef<Attribute> replAttrs,
```
krzysz00, line 126: Why're we implementing these?

external/llvm-project/mlir/lib/Dialect/XModel/IR/XModel.cpp
```diff
@@ -0,0 +1,264 @@
...
+                                               ArrayRef<Type> replTypes) const {
+  assert(0);
```
krzysz00, line 128: llvm_unreachable("Some vaguely descriptive message about why no one will ever call this")

external/llvm-project/mlir/lib/Dialect/XModel/Transforms/AsyncGraph.cpp
```diff
@@ -1,4 +1,5 @@
-//===- AsyncLaunch.cpp - Convert kernel func call ops to async.launch -----===//
+//===- XModelAsyncGraph.cpp - Convert kernel func call ops to async.launch
+//-----===//
```
krzysz00: Can we shorten up the summary a bit to get rid of this ugly trailing dash situation? Maybe `Kernel calls to async.launch` with a longer description below?

external/llvm-project/mlir/lib/Dialect/XModel/Transforms/PackageTargets.cpp
```diff
@@ -69,15 +72,35 @@ struct RockApplyImplPass
...
+                auto funcName = attr.getLeafReference().getValue();
+                uint32_t gridSize =
+                    func->getAttrOfType<IntegerAttr>("grid_size").getInt();
+                uint32_t blockSize =
+                    func->getAttrOfType<IntegerAttr>("block_size").getInt();
+
+                auto chip = gpuMod->getAttr("xmodel.chip");
+                auto triple = gpuMod->getAttr("xmodel.triple");
+                auto features = gpuMod->getAttr("xmodel.features");
+                DictionaryAttr objAttrs =
```
krzysz00: Why are we storing these? Wouldn't `arch` be sufficient?

external/llvm-project/mlir/lib/Dialect/XModel/Transforms/SelectTargets.cpp
```diff
@@ -0,0 +1,85 @@
...
+    for (auto targetType : targetTypes) {
+      if (targetType == typeStr)
+        return true;
+    }
+    return false;
+  }
+
+  bool testChip(StringRef chip) const {
+    for (auto targetChip : targetChips) {
+      if (targetChip == chip)
```
krzysz00: I don't agree with `==` here. Again, `gfx908:sramecc+:xnack-` should match `gfx908` but prefer `gfx908:sramecc+:xnack-`

external/llvm-project/mlir/lib/Dialect/XModel/Transforms/SelectTargets.cpp
```diff
@@ -0,0 +1,85 @@
...
+        return true;
+    }
+    return false;
+  }
+
+  // Replaces synchronous call ops in the op's region with asynchronous ones and
+  // inserts the necessary synchronization (as async.await ops). Assumes
+  // sequential execution semantics and that no asynchronous ops yet.
+  void runOnOperation() override {
+    const char *targets_tag = "xmodel.targets";
```
krzysz00: constexpr llvm::StringLiteral `targetsTag("xmodel.targets");`

Especially since upstream style in mlir is `camelCase` and not `snake_case` (and not llvm's `HumpyCase`)

external/llvm-project/mlir/lib/Dialect/XModel/Transforms/SelectTargets.cpp
```diff
@@ -0,0 +1,85 @@
...
+    func::FuncOp func = getOperation();
+    OpBuilder b(func);
+    if (auto targets = func->getAttrOfType<ArrayAttr>(targets_tag)) {
+      xmodel::KernelPackageAttr targetKrn;
+      for (auto targetAttr : targets.getValue()) {
+        auto pkgAttr = targetAttr.cast<xmodel::KernelPackageAttr>();
+        auto type = pkgAttr.getType();
+        auto chip = pkgAttr.getTarget();
```
krzysz00: `arch` over `chip` and some fuzzy matching logic, I claim

external/llvm-project/mlir/lib/Dialect/XModel/Transforms/TargetKernels.cpp
```diff
@@ -1,4 +1,5 @@
-//===- CloneKernels.cpp ---------------------------------------------------===//
+//===- XModelTargetKernels.cpp
+//---------------------------------------------------===//
```
krzysz00: Drop some dashes to keep the header

external/llvm-project/mlir/lib/Dialect/XModel/Transforms/TargetKernels.cpp
```diff
@@ -15,39 +16,43 @@
...
-      RockCloneKernelsPass>::RockCloneKernelsPassBase;
+struct XModelTargetKernelsPass
+    : public xmodel::impl::XModelTargetKernelsPassBase<
+          XModelTargetKernelsPass> {
+  using xmodel::impl::XModelTargetKernelsPassBase<
+      XModelTargetKernelsPass>::XModelTargetKernelsPassBase;
+
   void runOnOperation() override {
-    const char *kernel = "kernel.module";
+    const char *kernel = "xmodel.module";
```
krzysz00: Might as well StringLiteral here too

mlir/tools/xmir-runner/xmir-runner.cpp
```diff
@@ -42,8 +42,8 @@ using namespace llvm;
 // CLI switch for launch-mode
 static cl::opt<std::string> targetType("target-type",
```
krzysz00, line 43: Can we make this enum-typed. `clEnumValN` is the magic incantation, I think

mlir/tools/xmir-runner/xmir-runner.cpp
```diff
@@ -69,43 +69,25 @@ static LogicalResult runMLIRPasses(ModuleOp m) {
...
-    }
-  });
+  SmallVector<std::string, 4> targetChips;
+  if (targetChip.getValue().empty()) {
+    auto devices = SystemDevices::get<CpuSystemDetect, RocmSystemDetect>();
+    auto device = devices.find(targetTypeEnum);
+    if (succeeded(device))
+      targetChips.push_back((*device)->chip.str().str());
+  } else {
+    targetChips.push_back(targetChip);
```
krzysz00: I'd like this to be named `targetArchs` and for `targetChip` to undergo canonicalization if possible

krzysz00 (commented):
Overall, I like the idea here, but I've got minor type quibbles and a more fundamental complaint/question about `chip` vs `arch` (and why we need to store triple/chip/features when we have `arch`)

external/llvm-project/mlir/include/mlir/Dialect/XModel/IR/XModelAttrDefs.td
sjw36: No, it is a symbol reference into the ELF binary.

external/llvm-project/mlir/include/mlir/Dialect/XModel/IR/XModelAttrDefs.td
sjw36: The target_obj may in the future contain more than 1 kernel. There could be multiple kernel_pkg's that reference a target_obj.

external/llvm-project/mlir/lib/Conversion/AsyncToGPU/AsyncToGPU.cpp
sjw36: Now that there is a reference to xmodel... makes sense.

external/llvm-project/mlir/lib/Dialect/XModel/IR/XModel.cpp
sjw36: Only way to make target_obj alias-able.. since it is an attribute referenced by an attribute.

external/llvm-project/mlir/lib/Dialect/XModel/Transforms/SelectTargets.cpp
sjw36: Yes, we need a compatibility check.

external/llvm-project/mlir/include/mlir/Dialect/XModel/IR/XModelAttrDefs.td
krzysz00: Ah.

Still feels weird to have something so GPU-specific in here.

external/llvm-project/mlir/include/mlir/Dialect/XModel/IR/XModelAttrDefs.td
krzysz00: Makes sense.

external/llvm-project/mlir/lib/Dialect/XModel/IR/XModel.cpp
sjw36: I don't see what you're referring to.

external/llvm-project/mlir/lib/Dialect/XModel/IR/XModel.cpp
sjw36: should never be called.

sjw36:
@krzysz00 most things addressed. I will hold off on use cl::opt enums for this PR.

external/llvm-project/mlir/lib/Dialect/XModel/IR/XModel.cpp
krzysz00: The main commit is, near as I can tell, https://reviews.llvm.org/D126446

external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp
krzysz00: ... was there something that changed here that didn't get committed?

external/llvm-project/mlir/lib/Dialect/XModel/IR/XModel.cpp
krzysz00: Hence `llvm_unreachable()` over an `assert(0)` that carries no information about why you'll never go here

krzysz00: or even `assert(0 && "This is a non-implementation to keep the linker happy and should never be called");`

external/llvm-project/mlir/lib/Dialect/XModel/Pipelines/Pipelines.cpp
```diff
@@ -15,35 +15,72 @@
...
+  pm.addPass(
+      tosa::createTosaPartition({{"tosa.conv2d", "tosa.depthwise_conv2d"}}));
+
+  // make async kernel launch's
+  /* rocmlir-opt --xmodel-async-graph
+   */
+  pm.addNestedPass<func::FuncOp>(createXModelAsyncGraphPass());
+
+  // clone 'kernel' funcs into __rock module
+  /* rocmlir-opt --xmodel-target-kernels
```
krzysz00: We can probably call this `mlir-opt` given that it's in external now? Also, does it still go to `__rock`?

external/llvm-project/mlir/lib/Dialect/XModel/Transforms/PackageTargets.cpp
krzysz00: Re-raising the question about why chip/target/features?

external/llvm-project/mlir/lib/Dialect/XModel/Transforms/SelectTargets.cpp
```diff
@@ -0,0 +1,95 @@
...
+      if (targetType == typeStr)
+        return true;
+    }
+    return false;
+  }
+
+  bool testArch(xmodel::TargetType type, StringRef arch) const {
+    if (!testType(type))
+      return false;
+    SystemDevice testDev{SystemDevice::Type::EGPU};
```
krzysz00, line 55: Why is this hardcoded to GPU?

external/llvm-project/mlir/lib/ExecutionEngine/SystemDevices.cpp
```diff
@@ -33,69 +33,110 @@ static const char *getDeviceTypeStr(SystemDevice::Type type) {
...
 }
 
-FailureOr<const SystemDevice *const>
-SystemDevices::find(SystemDevice::Type type,
-                    llvm::StringRef partialArchName) const {
-  llvm::SmallVector<StringRef, 3> parts;
+// SystemDevice::SystemDevice(Type _type, StringRef _triple, StringRef _chip, llvm::StringMap<bool> _features)
+//     : type(_type), llvmTriple(_triple), chip(_chip), features(_features) {}
 
+LogicalResult SystemDevice::parse(StringRef arch) {
```
krzysz00, line 39: Nice cleanup!

krzysz00 (commented):
Seems pretty reasonable to me, just a few small things.

And I agree that getting the ListOption/enum stuff can be a future PR.

external/llvm-project/mlir/lib/Dialect/XModel/IR/XModel.cpp
sjw36: I was building something like this, should work for this case.

external/llvm-project/mlir/lib/Dialect/XModel/Transforms/SelectTargets.cpp
sjw36: The check is basically only comparing arch vs arch strings. So the types just need to be the same.

external/llvm-project/mlir/lib/Dialect/XModel/IR/XModel.cpp
sjw36: Hmm, gonna punt for another PR on this as well.

external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp
sjw36: unused.. removed.

sjw36:
All green. @krzysz00 please re-review.

external/llvm-project/mlir/lib/Dialect/XModel/IR/XModel.cpp
krzysz00: I really want something more explanatory here than assert(0).

krzysz00 (approved):
I'm approving, but I'd like to see the assert(0) replaced with something more descriptive

And some followup PRs,, one among them being moving the launch sizes out of the target spec, or at least to somewhere where you won't set them if you're on a target that doesn't *have* launch dimensions.

PR #825, opened 2022-10-20
Rearrange non-xdlops memory operations, 12%-ish wrw perf boost

krzysz00:
Move the loads to the front of the loop, just like in the xdlops gemm. This leads to uniformity.

Most performance is flat or slightly improved, but we have somewhere around 10-18% performance improvement in wrw. Some forward-direction configs get worse, but it's not consistent which ones.

Here's a pair of performance runs from my Navi21 workstation, showing the wrw improvements.
[new-loop-order-perf.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/9833810/new-loop-order-perf.zip)

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
@@ -904,18 +904,6 @@ struct GridwiseGemmRewritePattern : public OpRewritePattern<GridwiseGemmOp> {
       PatternRewriter::InsertionGuard guard(b);
       b.setInsertionPointToStart(loopOp.getBody());
 
-      // LDS barrier.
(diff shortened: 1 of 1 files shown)


jerryyin:
I noticed you have two measurements from the file:
 - What's the difference between the two?
 - Are those on heuristic configs, not from retuned results?

This looks very good. Reason I asked is because I find a couple of regressions on fwd configs. It'd be perfect if after re-tuning, the regression ones should be replaced by a different config.

jerryyin (approved):
I just double checked with the existing MIOpen/CK implementation. All of them uniformly followed the pattern of `load` -> `gemm` -> `store` in the loop.

krzysz00:
These are heuristic results - and the two measurements are two independent runs of the benchmarks

krzysz00:
And this is really showing we need to refresh our heuristic config list

And to update: it was backward data configs that regressed, not forward ones

PR #826, opened 2022-10-20
Add repeat kernel launches to performance measurements for MLIR

krzysz00:


sjw36 (commented):
Does it loop around mgpuMemAlloc and mgpuMemcpy?

krzysz00:
No, this just loops around the kernel launch.

sjw36 (approved):
Looks good.

PR #827, opened 2022-10-21
Add a basic rocblas-benchmark-driver so we have something to compare against

krzysz00:


sjw36 (approved):
Nice utility. Can we do this for CK as well?

krzysz00:
Not knowing the CK API, I wouldn't know how possible that is. @giuseros for "do please note how much of a pain this would be while you're looking at CK anyway"

PR #829, opened 2022-10-24
xdlops gridwisegemm coordinate transform enablement

jerryyin:
This PR make xdlops gridwisegemm using coordinate transform as non-xdlops. It matched the implementation of xdlops from nonxdlops (in https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/718)

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
@@ -1344,6 +1344,11 @@ struct GridwiseGemmV2RewritePattern
         b.getIndexAttr(ldsBlockBOffset), mMyWaveOffsetA, mMyWaveOffsetB, arrayA,
         arrayB, regCAllocOp, op.getBlockSizeAttr(), op.getParamsAttr());
 
+    // Apparently, the canonicalizer doesn't get rid of empty loops without
+    // results properly, remove them ourselves.
+    if (nIterations <= 1)
+      b.eraseOp(loopOp);
+
     // -----
 
     // Matrix C write out logic.
--- a/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
@@ -80,8 +80,8 @@ static void computeLoadStoreTypeInfo(OpBuilder &b,
                                      int64_t kPack, Type elementType,
                                      Type &loadType, Type &intermediateType) {
 
-  // In case KPack and vector load is used, and we vector load on GemmK
-  // dimension (1), use the last dimension (GemmKPack) instead.
+  // In case kpack and vector load is used, and we vector load on GemmK
+  // dimension (1), use the last dimension (Gemmkpack) instead.
   if ((loadLength > 1) && (kPack > 1) && (vectorDim == 1)) {
     vectorDim = sliceLengths.size() - 1;
   }
@@ -381,7 +381,7 @@ static FailureOr<Value> wrapMatrixForGlobalLoad(
   return transformed;
 }
 
-/// Wraps the LDS buffer "buffer", which is K x D x KPack, into a
+/// Wraps the LDS buffer "buffer", which is K x D x kpack, into a
 /// tid x iter view.
 static FailureOr<Value> wrapLDSBufferForStore(OpBuilder &b, Location loc,
                                               Value buffer, StringRef dName,
@@ -591,7 +591,7 @@ struct GridwiseGemmRewritePattern : public OpRewritePattern<GridwiseGemmOp> {
   LogicalResult computeLDSBlockSizes(GridwiseGemmOp op, int64_t &a_block_space,
                                      int64_t &b_block_space,
                                      int64_t &block_space,
-                                     int64_t KPack = 1) const {
+                                     int64_t kpack = 1) const {
     GeneralGemmParamsAttr tuningParams = op.getParams();
     int64_t ThreadGemmAThreadCopySrcDataPerRead_M =
         tuningParams.getMPerThread();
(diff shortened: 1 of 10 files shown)


mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1167,508 +1167,138 @@ struct GridwiseGemmV2RewritePattern
...
+    int64_t nBlocks = N / nPerBlock;
+
+    int64_t kPerBlock = kpacksPerBlock * kpack;
+
+    int64_t aVectorLen = 0;
+    int64_t bVectorLen = 0;
+    GemmDimension aVectorDim;
+    GemmDimension bVectorDim;
+
+    SmallVector<StringRef, 3> bidGridOrder = {"g_block", "m_block", "n_block"};
```
krzysz00, line 1020: Potential bug source: the xdlops code expects "m_block" and "n_block" to be the other way around.

krzysz00 (commented):
I've got a guess

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: That was some amazing bug-finding abilities here. Thanks so much and this is it.

krzysz00: It's less "amazing bug-finding abilities" and more:
1. I ran into this weird quirk back when I was refactoring the output loops and
2. Remembered it would be coming up when I did the non-xdlops refactor, so I planned for it and
3. When I saw this line, it reminded me of the possibility of this bug and so I went off to check if this could be the problem

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -428,21 +229,23 @@ static FailureOr<Value> wrapLDSBufferForStore(OpBuilder &b, Location loc,
...
+  int64_t kpackPerThread = std::min(kPerThread, kpack);
+  int64_t kOuter = kPerThread / kpackPerThread;
+
   int64_t dataPerThread = kPerThread * dPerThread;
 
   TopDownTMBuilder splitIter(b, {"iter"}, {dataPerThread});
   if (vectorDim == GemmDimension::K)
-    splitIter.merge({dName, "k", "kpack"}, {0, 1, 2}, "iter",
-                    {dPerThread, kOuter, kpack});
+    splitIter.merge({dName, "k", "kpack_thread"}, {0, 1, 2}, "iter",
```
krzysz00, line 271: Nit: the rest of the names in here don't include the thread-ness, we could probably just call this "kpack"?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1167,508 +970,138 @@ struct GridwiseGemmV2RewritePattern
...
+        createGlobalLoadLoop(b, loc, wrappedA, aVectorGlobalMap, aCopyPerThread,
+                             aVectorLen, bid, tid);
+    TransformingForOp blockwiseLoadB =
+        createGlobalLoadLoop(b, loc, wrappedB, bVectorGlobalMap, bCopyPerThread,
+                             bVectorLen, bid, tid);
 
-    Value GemmBBlockCopySourceCoord_Y;
-    Value GemmBBlockCopySourceCoord_X;
+    // Obtain XDLOPS-related attributes.
+    int64_t MPerWave = tuningParams.getMPerWave();
```
krzysz00: Could we go ahead and call this `mPerWave`?

krzysz00: (and analogously - this is about me getting clang-tidy squiggles)

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1167,508 +970,138 @@ struct GridwiseGemmV2RewritePattern
...
-      GemmBBlockCopySourceCoord_Z = GemmBThreadDataIdBegin_Z;
-      GemmBBlockCopySourceCoord_Y = b.create<AddIOp>(
-          loc, n_block_data_on_global, GemmBThreadDataIdBegin_Y);
-      GemmBBlockCopySourceCoord_X = GemmBThreadDataIdBegin_X;
-    } else {
-      GemmBBlockCopySourceCoord_Y = GemmBThreadDataIdBegin_Y;
-      GemmBBlockCopySourceCoord_X = b.create<AddIOp>(
-          loc, n_block_data_on_global, GemmBThreadDataIdBegin_X);
-    }
+    auto MPerWaveConstantOp = b.create<ConstantIndexOp>(loc, MPerWave);
```
krzysz00: I'm guessing cleaning up some of this might be a future PR?

krzysz00 (approved):
LGTM, thanks for all the hard work on this, I have minor "while we're here" variable name cleanups, but nothing that prevents a merge.

Feel free to merge pending passing nightly and weekly

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: The reason why I added `kpack_thread` is to just give a reminder to who read the transform map that this dimension isn't kpack (and we don't guarantee the size of it is equal to kapck). I was having this assumption when I was doing such so having a hard time to realize why we need `kpackPerThread` abstraction. 

I think I get it from your side that the transformations already have tid in them so it is implied, but I'm just trying to make it a different term for differentiation.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Of course, I should. I'll do it in next commit.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Yes, blockwise level ops isn't using the layout information as it should, it will be a future PR for sure.

On a side topic, it looks like this PR CI starts to catch all sorts of corner cases again just like last time (in the vectorization PR). I'll try to touch less things functionally and throw a nightly run at this PR before merging it.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: Not a bad point, and I don't see any reason not to do it your way

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1016,19 +1016,41 @@ struct GridwiseGemmV2RewritePattern
...
-    int64_t bCopyKPerThread = bVectorDim == GemmDimension::K
-                                  ? bVectorLen
-                                  : bCopyPerThread / bVectorLen;
-    int64_t copyNPerThread = bVectorDim == GemmDimension::MorN
-                                 ? bVectorLen
-                                 : bCopyPerThread / bVectorLen;
+    int64_t aCopyKPerThread = 0;
+    int64_t copyMPerThread = 0;
+    if (aVectorDim == GemmDimension::K) {
+      aCopyKPerThread = std::min(aVectorLen, kPerBlock);
```
krzysz00: Huh. I guess it's not surprising you can hit this case - I guess we never had it with non-xdlops because of their tuning space.

Might be worth doing this to non-xdlops too?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1016,19 +1016,41 @@ struct GridwiseGemmV2RewritePattern
...
+      copyMPerThread = aCopyPerThread / aCopyKPerThread;
+    } else {
+      copyMPerThread = std::min(aVectorLen, mPerBlock);
+      aCopyKPerThread = aCopyPerThread / copyMPerThread;
+    }
+    if (aCopyKPerThread == 0 || copyMPerThread == 0) {
+      LLVM_DEBUG(llvm::dbgs() << "gemmA copy size too small.\n");
```
krzysz00: These could maybe be `return op.emitOpError()` and possibly include the actual values via `<<`?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Sounds good, will generalize in a subroutine so both xdlops/nonxdlops can use it.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Sounds good!

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Addressed in e22c5a4

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Addressed in ef3f98f23bd6657d79daf0fbbc5b73800c5723f6

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: Addressed in e22c5a42c92e7993cb604db3173f9cad7824ed91

PR #831, opened 2022-10-25
[CI][Navi21] Enable Navi21 codepath for PR and Nightly CI

zhanglx13:
Three navi workstations are added as jenkins agents. They have the label `gfx1030w && mlir`. 
This PR only enabled navi nodes for PR and Nightly CI only.

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -148,7 +148,9 @@ String getLabelFromCodepath(String codepath) {
     if (codepath == "mfma") {
         label = 'mlir && (gfx908 || gfx90a)'
     } else if (codepath == "navi21") {
-        label = 'navi_workstation'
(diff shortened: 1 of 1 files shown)


krzysz00 (commented):
Do you want this approved so you don't have to hound anyone for a review once it's working?

And, since we have one Navi node, how about only running Navi CI nightly and weekly?

zhanglx13:
Yeah, please approve so that I can merge it as long as the node is fully tested.
I'll test only the navi21 codepath for nightly and weekly so that other codepaths will not be too occupied.

krzysz00:
I more meant "don't run Navi21 tests on PRs, even after this lands, because we don't have enough nodes for that"

zhanglx13:
@krzysz00 So now we have three tested navi21 nodes (two workstations and one server node) that are working at the PR CI stage. And since the scope of this PR is changed to PR CI only, I think it is ready to merge.

krzysz00 (approved):
Seems reasonable to me, approved

PR #833, opened 2022-10-26
Use a smaller rebuild in CMake, saving CI time

krzysz00:
Instead of clearing the build cache to switch shared/static lib, set the CMake file up to FORCE the relevant settings (shared libraries, PR tests, and so on) ON or OFF when someone explicitly passes in -DBUILD_FAT_LIBROCKCOMPILER . Note that we introduce a variable that tracks when the forcing has taken place to ensure that we don't FORCE over options when we don't need to - that is, if we build static library and then build the normal build, we need to force the options back, but not in any other case, where we might've wanted to set those options manually.

I expect this will save on CI time.

Fixes
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/689

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,30 +52,45 @@ message(STATUS "ROCMLIR_BIN_DIR: ${ROCMLIR_BIN_DIR}")
 set(CMAKE_BUILD_RPATH "${ROCMLIR_LIB_DIR};${LLVM_EXTERNAL_LIB_DIR}")
 message(STATUS "CMAKE_BUILD_RPATH: ${CMAKE_BUILD_RPATH}")
 
+set(CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE STRING "")
+
+function(_maybe_force_set var val type docs force)
+  if (${force})
+    set(${var} ${val} CACHE ${type} "${docs}" FORCE)
+  else()
+    set(${var} ${val} CACHE ${type} "${docs}")
+  endif()
+endfunction()
+
+function(_set_librockcompiler_varying_bools value force)
+  _maybe_force_set(BUILD_SHARED_LIBS ${value} BOOL "" ${force})
+  _maybe_force_set(LLVM_BUILD_LLVM_DYLIB ${value} BOOL "" ${force})
+  _maybe_force_set(LLVM_BUILD_EXAMPLES ${value} BOOL "" ${force})
+  _maybe_force_set(MLIR_ENABLE_ROCM_RUNNER ${value} BOOL "" ${force})
+  _maybe_force_set(MLIR_INCLUDE_INTEGRATION_TESTS ${value} BOOL "" ${force})
+  _maybe_force_set(ROCMLIR_DRIVER_PR_E2E_TEST_ENABLED ${value}
+    BOOL "Enable build PR-triggered E2E tests for Rock driver" ${force})
+endfunction()
+
+# Set up defaults for not using librockcompiler either way
+_set_librockcompiler_varying_bools(ON OFF)
+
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -52,30 +52,45 @@ message(STATUS "ROCMLIR_BIN_DIR: ${ROCMLIR_BIN_DIR}")
 set(CMAKE_BUILD_RPATH "${ROCMLIR_LIB_DIR};${LLVM_EXTERNAL_LIB_DIR}")
 message(STATUS "CMAKE_BUILD_RPATH: ${CMAKE_BUILD_RPATH}")
 
+set(CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE STRING "")
```
zhanglx13, line 55: `CMAKE_INSTALL_PREFIX` is only needed when `BUILD_FAT_LIBROCKCOMPILER` is ON. Is it ok to define it on both cases?

CMakeLists.txt
```diff
@@ -52,30 +52,45 @@ message(STATUS "ROCMLIR_BIN_DIR: ${ROCMLIR_BIN_DIR}")
...
+
 if(BUILD_MIXR_TARGET)
   set(BUILD_FAT_LIBROCKCOMPILER ON CACHE BOOL "Build static librockCompiler")
   set(EXPORT_ALL_HEADERS ON CACHE BOOL "Export internal headers")
 endif()
 
 # Library type and linkage settings
-if( NOT DEFINED BUILD_FAT_LIBROCKCOMPILER )
+if(NOT DEFINED BUILD_FAT_LIBROCKCOMPILER)
   set(BUILD_FAT_LIBROCKCOMPILER OFF CACHE BOOL "Build fat librockCompiler to link into Rock driver")
```
zhanglx13, line 83: How about setting it FORCEly if it is not defined so that in Jenkinksfile, you do not have to explicitly define it as OFF.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -386,7 +386,6 @@ pipeline {
                             LD_LIBRARY_PATH="${WORKSPACE}/MIOpen/build/lib:$LD_LIBRARY_PATH:"
                         }
                         steps {
-                            sh 'rm build/CMakeCache.txt'
```
zhanglx13, line 389: From the CI, it seems that this does not reduce the number of objects needed to be built for the static library. Not sure about the other way around. You may want to try a nightly to see how it goes.

CMakeLists.txt
krzysz00: FORCE overrides settings from the command line

CMakeLists.txt
krzysz00: Probably fine

mlir/utils/jenkins/Jenkinsfile
krzysz00: What I think might be happening here is that some of those flags are changing the CXXFLAGS, which would be causing the mass recompile. I might want to

1. Change that if/elseif into a three-way if/elseif/else (use PR defaults, no forcing)
2. Check if we actually need to force all those flags. My very preliminary suspect here is LLVM_BUILD_EXAMPLES

CMakeLists.txt
zhanglx13: If it is set from the command line we won't get in the `if`, right?

krzysz00:
Looks like there's no obvious way for this to actually do something, so closing

CMakeLists.txt
krzysz00: No, we can, if you set the variable in a previous CMake run

PR #834, opened 2022-10-26
[DO NOT SQUASH] Fix XMIR silent failures when target selection fails

sjw36:
MLIR#718: Fixed XModel target selection
* error out when no target object found
* added negative test
    
MLIR#719: Added diagnostic messaging for mgpu routines
* Enable with env var 'XMIR_DIAG_LEVEL'
    
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/718
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/719


external/llvm-project/mlir/lib/Dialect/XModel/Transforms/SelectTargets.cpp
```diff
@@ -87,6 +89,11 @@ struct XModelSelectTargetsPass
       if (targetKrn)
         func->setAttr(targetsTag, b.getArrayAttr(targetKrn));
       else {
+        if (targetArchs.size()) {
+          auto msg = Twine("Target not found for '") + func.getName() + "'\n";
```
krzysz00: `func.emitError()` or `func.emitOpError()`, and never printing straight to `llvm::errs()`

external/llvm-project/mlir/lib/ExecutionEngine/RocmRuntimeWrappers.cpp
```diff
@@ -20,6 +20,16 @@
 
 #include "hip/hip_runtime.h"
 
+static bool isVerbose() {
+  static char *envval = getenv("XMIR_DIAG_LEVEL");
```
krzysz00: No outright `getenv()`, LLVM has a wrapper for that.

krzysz00: Second, don't call this XMIR-anything.

How's `ROCM_RUNTIME_WRAPPERS_TRACE`?

krzysz00: Third, ... this should probably work with a `#define LLVM_DEBUG_TYPE` and `LLVM_DEBUG`, that being the usual mechanism as opposed to setting random environment variables for tracing

mlir/test/xmir/e2e/resnet50_blk.fail.mlir
```diff
@@ -0,0 +1,45 @@
+// RUN: rocmlir-driver -host-pipeline partition,highlevel -targets gfx900 %s | rocmlir-gen -ph -print-results -rand_type float -rand 1 -fut resnet50 --verifier clone - | rocmlir-driver -host-pipeline xmodel -kernel-pipeline full | not xmir-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_c_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_async_runtime%shlibext --entry-point-result=void 2>&1 | FileCheck %s
```
krzysz00: Could we make this a much more targeted test? Run this through almost all of the pipeline and then just `RUN: xmir-runner %s`

krzysz00 (changes requested):
In terms of the core code changes, perfectly reasonable PR

In terms of how you're doing error reporting ... LLVM/MLIR have facilities for this and they should be used. Maybe not _necessarily_ in the runtime wrappers, but definitely in the main code.

And if you're checking for an error message, `rocmlir-opt` has a `-verify-diagnostics` option (by way of `mlir-opt`) that you can use to make more precise tests for errors

external/llvm-project/mlir/lib/ExecutionEngine/RocmRuntimeWrappers.cpp
sjw36: This is not llvm code.. those mechanisms are not available. I will use the std::getenv instead.

krzysz00: Point, the rocm runtime wrappers are a separate thing, and so using getenv() here is reasonable - I don't object to this one given the circumstances.

But you could also go and link in LLVM's `Support` component

sjw36: yeah, I would rather not do that.

mlir/test/xmir/e2e/resnet50_blk.fail.mlir
sjw36: Good idea.

external/llvm-project/mlir/lib/ExecutionEngine/RocmRuntimeWrappers.cpp
```diff
@@ -20,6 +20,16 @@
 
 #include "hip/hip_runtime.h"
 
+static bool isVerbose() {
+  static char *envval = std::getenv("ROCMLIR_RUNTIME_TRACE");
```
krzysz00, line 24: I'd nitpick that external code should probably go for something like `MLIR_ROCM_RUNTIME_TRACE` but that's a rename we can do while upstreaming this ... come to think it, this change should probably be upstreamed right after the PR lands

krzysz00 (approved):
LGTM, mod one minor naming nit that can be fixed while sending the upstream patch and reconciled in the next merge

external/llvm-project/mlir/lib/ExecutionEngine/RocmRuntimeWrappers.cpp
sjw36: That's good since it won't collide with any ROCM defines.

PR #837, opened 2022-11-01
Modify MIGraphX capi to use single arch string.

jungpark-mlir:


krzysz00 (commented):
I suspect we could leave the `BackendPipeline` call as is, but I don't have a strong opinion on that - it works either way.

In the MIGraphX code, do they have the full name and then split it? If so, we can save them splitting by changing this API. If they have the parts, then we should consume the parts.

However, the really importing thing in unrelated to any of this and involve the *front* fo the pipeline, which needs to be changed to not pass in a redundant `xdlopsv2` attribute and to pass the full arch name (or we need to have part of our code do the canonicalization, perhaps MIGraphXToTosa.

jungpark-mlir:
Exactly, I'm doing this to avoid split string in MIGraphX which now looks not needed.
MIGraphX side commit includes the change using full arch name in the front-end.

jungpark-mlir:
I mean the change in MIGraphX will follow this.

krzysz00 (changes requested):
Bump `MLIR_MIGRAPHX_DIALECT_API_VERSION` so we can update MIGraphX seamlessly

jungpark-mlir:
> Bump `MLIR_MIGRAPHX_DIALECT_API_VERSION` so we can update MIGraphX seamlessly

Done

krzysz00 (approved):
Approved on the condition that a MIGraphX PR that bumps our commit and accounts for this change is filed or will soon be

PR #838, opened 2022-11-01
[NFC] Rename threadwise_copy_v2 to global_store

krzysz00:
Change some names in AlignTiling.cpp to make it slightly clearer what's going on.

Also, amend the fusion flow from BufferLoad to GlobalLoad, which is one level higher up the stack (it can be one or more BufferLoad ops) so as to prevent any potential issues when the output loop vectorizes to a global_store that involves multiple buffer_store operations.

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
@@ -170,20 +170,20 @@ static Value makeBroadcast(PatternRewriter &b, MemRefType outType, Value inp,
   return inp;
 }
 
-static void insertCopyFromOtherArg(PatternRewriter &b, Location loc,
-                                   ThreadwiseCopyV2Op op, Value srcOp,
-                                   Value dest) {
+static void insertLoadFromOtherSource(PatternRewriter &b, Location loc,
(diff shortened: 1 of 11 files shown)


mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -423,13 +424,13 @@ LogicalResult MILARewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
   // 1. Trace input to threadwise_copy. Collect transforms (to be applied to
```
giuseros: threadwise_copy -> global_store ?

giuseros:
Hi @krzysz00 , I see that there is also something in `examples/prototypes` that still uses the old naming. Should those need to be updated as well?

krzysz00:
... we should probably delete `examples/prototypes/`. It's so massively out of date in terms of code structure that ... yeah.

And yeah, I'll fix the stray bit that didn't get caught by `sed`

giuseros:
LGTM, but I am not sure I have the rights to approve

krzysz00:
I think you do have the rights to approve - you're showing as "Member" same as me

sjw36 (approved):
Looks good.

jungpark-mlir (approved):
Yep, looks good.

PR #842, opened 2022-11-03
Add validations to ensure no gridwise gemm dimension is over 2 GB

krzysz00:
Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/705

Am leaving on hold until gridwise_gemm_v2 refactoring lands

### DIFF
--- a/mlir/test/Dialect/Rock/invalid.mlir
+++ b/mlir/test/Dialect/Rock/invalid.mlir
@@ -38,3 +38,104 @@ func.func @gemm_c_too_big(%a: memref<2048x2048x2048xf32>,
   : memref<2048x2048x2048xf32> = memref<2048x2048x2048xf32> * memref<2048x2048x2048xf32>
   func.return
 }
+
+// -----
+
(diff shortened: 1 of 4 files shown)


mlir/test/Dialect/Rock/invalid.mlir
```diff
@@ -38,3 +38,54 @@ func.func @gemm_c_too_big(%a: memref<2048x2048x2048xf32>,
   : memref<2048x2048x2048xf32> = memref<2048x2048x2048xf32> * memref<2048x2048x2048xf32>
   func.return
 }
+
```
jerryyin, line 41: Since you are updating the unit test, could you add one that test a dimension size to go over (std::numeric_limits<int32_t>::max()) and expect that error to happen?

jerryyin (approved):
Nice cleanup, thank you!

giuseros (commented):
LGTM other than Jerry's comment about adding an `int_max` validation test

PR #846, opened 2022-11-08
add resnet50 f32 fusion test cases

manupak:
This PR adds the resnet50 unique test cases.

Additionally, for this to work the GraphPipeline should
not call TosaOptionalDecompositions until we have 
support for tosa.fully_connected to rock.gemm.

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/724

### DIFF
--- a/mlir/test/fusion/e2e/resnet50/mixr-resnet-fusion-case-1.mlir
+++ b/mlir/test/fusion/e2e/resnet50/mixr-resnet-fusion-case-1.mlir
@@ -1,6 +1,15 @@
-// RUN: rocmlir-opt -migraphx-to-tosa %s | rocmlir-driver -host-pipeline partition,highlevel -targets %arch | FileCheck %s
+// RUN: rocmlir-opt -migraphx-to-tosa %s | rocmlir-driver -host-pipeline partition,highlevel -targets %arch | rocmlir-opt --rock-fold-transpose | FileCheck %s
 // CHECK: linalg.conv_2d
-// CHECK: rock.conv2d
+// CHECK: %[[ALLOC:.*]] = memref.alloc() {{.*}}
+// CHECK-DAG: %[[ARG0_TR:.*]] = rock.transform %[[ARG0:.*]]
+// CHECK-DAG: %[[ARG1_TR:.*]] = rock.transform %[[ARG1:.*]]
+// CHECK-DAG: %[[ALLOC_TR:.*]] = rock.transform %[[ALLOC]]
+// CHECK-DAG: rock.conv2d(%[[ARG1_TR]], %[[ARG0_TR]], %[[ALLOC_TR]])
+// CHECK-DAG: %[[LAIN0:.*]] = memref.collapse_shape %[[ALLOC]]
+// CHECK-DAG: %[[LAIN1:.*]] = memref.collapse_shape %[[ARG2:.*]]
+// CHECK-DAG: %[[LAIN2:.*]] = memref.collapse_shape %[[ARG3:.*]]
+// CHECK-DAG: linalg.generic {{.*}} ins(%[[LAIN0]],
+
 // RUN: rocmlir-opt -migraphx-to-tosa %s | rocmlir-driver -host-pipeline partition,highlevel -targets %arch | rocmlir-gen -ph -print-results -verifier clone -fut test - | rocmlir-driver -host-pipeline xmodel -kernel-pipeline full | xmir-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_c_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_async_runtime%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CLONE
 // CLONE: [1 1 1]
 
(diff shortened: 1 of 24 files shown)


manupak:
A gentle ping!

mlir/test/fusion/e2e/resnet50/mixr-resnet-fusion-case-1.mlir
```diff
@@ -0,0 +1,13 @@
+// RUN: rocmlir-opt -migraphx-to-tosa %s | rocmlir-driver -host-pipeline partition,highlevel -targets %arch | rocmlir-gen -ph -print-results -rand 4 -rand_type float -fut test --verifier clone - | rocmlir-driver -host-pipeline xmodel,runner -kernel-pipeline full -targets %arch | mlir-cpu-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_c_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CLONE
```
jungpark-mlir: I suppose we don't need `-host-pipeline partition` here because currently we rely on migraphx partitioning.

jungpark-mlir (approved):
Looks good to me, just check if we need partition

mlir/test/fusion/e2e/resnet50/mixr-resnet-fusion-case-1.mlir
manupak: Im simply being consistent with other mixr clone tests.  

https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/ca6015d47d9f73ace4c2eda3fc9059dfab27bdd4/mlir/test/fusion/e2e/mixr-gemm/mixr-gemm-add.e2e.mlir#L2

It also fails w/o partition maybe because we use clone verification as follows : 

 'RUN: at line 1';   /home/mkarunar/work/rocMLIR/build/bin/rocmlir-opt -migraphx-to-tosa /home/mkarunar/work/rocMLIR/mlir/test/fusion/e2e/resnet50/mixr-resnet-fusion-case-1.mlir | rocmlir-driver -host-pipeline highlevel -targets gfx908:sramecc+:xnack- | rocmlir-gen -ph -print-results -rand 4 -rand_type float -fut test --verifier clone - | rocmlir-driver -host-pipeline xmodel,runner -kernel-pipeline full -targets gfx908:sramecc+:xnack- | mlir-cpu-runner -O2 --shared-libs=/home/mkarunar/work/rocMLIR/build/external/llvm-project/llvm/lib/libmlir_rocm_runtime.so,/home/mkarunar/work/rocMLIR/build/lib/libconv-validation-wrappers.so,/home/mkarunar/work/rocMLIR/build/external/llvm-project/llvm/lib/libmlir_runner_utils.so,/home/mkarunar/work/rocMLIR/build/external/llvm-project/llvm/lib/libmlir_c_runner_utils.so --entry-point-result=void | /home/mkarunar/work/rocMLIR/build/external/llvm-project/llvm/bin/FileCheck /home/mkarunar/work/rocMLIR/mlir/test/fusion/e2e/resnet50/mixr-resnet-fusion-case-1.mlir --check-prefix=CLONE
--
Exit Code: 2

Command Output (stderr):
--
Warning: unknown architecture, falling back to defaults: 
error: failed to legalize operation 'builtin.unrealized_conversion_cast' that was explicitly marked illegal

manupak: PS I didnt mean to highlight the error its just copy-pasting error and markdown being overly smart here :)

jungpark-mlir: I wonder if there is any way to check if partitioning pass actually split the given func.
@sjw36 can it partition a func? can it run a module when it has more than one func in it?

sjw36: Partitioning only applies to non-kernel funcs and yes it will work for multiple funcs.

krzysz00:
@manupak Should this be getting merged? I've been avoiding pressing the merge button when the option was available (since merding a few PRs one right after another saves on CI runs)

manupak:
These tests are wrong! As the majority of the clone tests checked in that doesn't use xmir runnner.

They are just verifying linalg to linalg.
The proper command line (using xmir-runner among other things) for tests make these fail. We need some adjustment to threshold.

But the sad part is even with that error is not significantly different with and without your (over vectorization) + mine(unit dims) as per clone tests where as in migraphx is able to differentiate that there is a bug without those fixes.

So I ll put these to draft for now, so that it wont give the false impression that these are verified.
We need to figure out why verification infra in migraphx is able to catch this and while ours is not able to.

manupak:
Back to this one!

I ve amended the tests and added additional checks to ensure both rock and linalg are there.

mlir/test/fusion/e2e/resnet50/mixr-resnet-fusion-case-1.mlir
```diff
@@ -0,0 +1,15 @@
+// RUN: rocmlir-opt -migraphx-to-tosa %s | rocmlir-driver -host-pipeline partition,highlevel -targets %arch | FileCheck %s
+// CHECK: linalg.conv_2d
+// CHECK: rock.conv2d
```
sjw36: Let's go ahead and add `// CHECK: linalg.generic` after the check for `rock.conv2d` just to be sure a fusion did happen.

sjw36 (changes requested):
see inline comment.

mlir/test/fusion/e2e/resnet50/mixr-resnet-fusion-case-1.mlir
manupak: Right, I've added a bit elaborative CHECKS make sure its fused :)

manupak:
@sjw36 PTAL when you have sometime.

sjw36 (approved):
Looks good. Thanks

PR #847, opened 2022-11-08
Define a more specialized registration function

giuseros:
This is to tackle: https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/661 

What I did: 
- I updated the registration functions to only register dialects/passes we actually use
- I updated the `CMakeLists.txt` to reflect the new outlook

TODO:
- Rename `mlir` folder to `rocmlir`
- Search/replace include guards

@krzysz00 , please could you give it a look to see if this aligns with what you had in mind? 

I will hold on the TODOs to avoid making the review too messy


mlir/lib/CAPI/RegisterRocMLIR/CMakeLists.txt
```diff
@@ -1,12 +1,11 @@
...
-get_property(conversion_libs GLOBAL PROPERTY MLIR_CONVERSION_LIBS)
+get_property(rocmlir_dialect_libs GLOBAL PROPERTY ROCMLIR_DIALECT_LIBS)
+get_property(rocmlir_conversion_libs GLOBAL PROPERTY ROCMLIR_CONVERSION_LIBS)
 add_mlir_public_c_api_library(MLIRCAPIRegisterRocMLIR
   RegisterRocMLIR.cpp
 
   LINK_LIBS PUBLIC
-  ${dialect_libs}
-  ${conversion_libs}
+  ${rocmlir_dialect_libs}
```
krzysz00, line 8: I think this needs a `${rocmlir_conversion_libs}` too?

krzysz00 (commented):
Yeah, I think we can save the `mlir/` -> `rocmlir/` change for a different PR.

I think some of the conversion libs got dropped, which could cause issues with stuff like `--convert-tosa-to-rock`.

Have you observed a shorter build on `check-rocmlir-build-only` and on the static library case?

(This should show up in Jenkins)

krzysz00:
While you're here, can you go in to the test depends and remove mentions of `mlir-opt`, `mlir-translate`, and so on? Those will cause all of upstream to be build for tests, which we don't want, and, last I checked, we don't use them in the tests (you may wont to `grep` to be sure)

mlir/lib/CAPI/RegisterRocMLIR/CMakeLists.txt
giuseros: I thought it as well, but I think these get linked in by the dialects? I was not able to see any linking issues without the `conversion_libs` (and also CI seems to run fine)

giuseros:
To me, the time to build (both the library and the compiler) did not seem to improve by a lot (maybe CI will tell us otherwise).

The problem I found is that libraries like `libMLIRNVGPUDialect.{so,a}` still get built. This is because in upstream they include in the global list of dialects and some components (like the `JitRunner`) link against that list. So, even if we reduce the amount of dependencies in our infrastructure, we still depend on components (like the runner) that depend on libraries we will never use.

krzysz00:
How are the things like JitRunner being pulled into `libroccompiler` or the other externally-facing artifacts? (I can agree that `JitRunner` will show up in our tests, but that's nothing much we can do, but it'll still be good to have the refactor)

giuseros:
So the answer is that `RockOps` was depending on `MLIRJitRunner`. I removed it and a new set of dependencies arose. It all seems fine now. If I build the library (or no tests) we are down to 2500 files to build (~25% reduction), which is what we were expecting.

krzysz00:
Might be worth looking for other dependencies on upstream that we don't need

giuseros:
Yeah I agree, I tried to do that. I think/hope it's now down to the minimum

krzysz00 (commented):
Re conversion libraries, I think we should swap `add_mlir_conversion_library` for `add_rocmlir_conversion_library`

And I'd suggest going through the C API and removing our dependency on MLIRCAPIRegisterEverything, which might be part of what's pulling in all the conversion libs

krzysz00:
(and I worked out why no linker errors wrt conversion libs: `Pipelines` explicitly lists out its dependencies. We could probably make that slightly less error-prone by using a hypothetical ROCMLIR_CONVERSION_LIBS there)

giuseros:
I removed the `RegisterEverything` calls from the C api, and removed some dependencies from `Pipelines` (and around). I am a bit reluctant to add a fixed list of conversion libraries, because some dialects will link against things they don't really need. While now they are linking against the `MLIRRock*` transforms + additional stuff they really need

krzysz00 (commented):
Ok, so, to rephrase my comment. To cleanly handle all our tools (like `rocmlir-gen`, `rocmlir-opt`, etc.) the conversion libraries should be gathered into a variable somewhere that's rocmlir-specific. That could be `ROCMLIR_CONVERSION_LIBS`, or just adding to `ROCMLIR_DIALECT_LIBS`, or something - the point is, they shouldn't be going through `add_mlir_conversion_lib` because that'll leak into upstream.

Also, on the "removing dependencies" front, a big one (esp. because I've gotten burned by big compile times on this before). We don't need `::test::registerTestDialect()` or the like in the `rocmlir-*` tools, and so the test libraries don't need to include MLIRTestIR or MLIRTestPasses.

(I'd also think, for the sake of cleanliness, it might be good idea to have MLIRRockTestPasses add itself to the test libraries, for locality's sake, but that's not a huge thing)

Thanks for doing all the deep dives here!

mlir/include/mlir/InitRocMLIRDialects.h
```diff
@@ -14,21 +14,91 @@
...
 #include "mlir/IR/Dialect.h"
 
 namespace mlir {
 
 // Add all the MLIR dialects to the provided registry.
 inline void registerRocMLIRDialects(DialectRegistry &registry) {
   // clang-format off
-  registry.insert<rock::RockDialect,
-                  migraphx::MIGraphXDialect>();
+  registry.insert<
```
krzysz00: Minor comment - could this be factored into a `inline void registerUpstreamDialects()` just so it's easier to think to look there?

giuseros:
Ok, I tried to apply all the changes. Dev machine was down, so I couldn't fully run the tests locally, let's see what CI says :)

krzysz00 (approved):
LGTM given that everything passes

PR #849, opened 2022-11-08
Fix cases where too much vectorization was applied in fusion

krzysz00:
There exist some fusions (such as broadcasting a scalar) where we could get bad results where the vectorization of the gemm writeback is larger than the vectorization that the fusion argument (which is transformed to give it the same shape as the gemm output).

Address this bug by having the load loop for additional arguments be vectorized by a vectorization that's a function of the fusion-argument transforms and capped at the vectorization of the writeback. This will almost certainly lead to scalar loads.

Furthermore, resolve a long-standing bug in index diff updates that could lead to use-before-def but somehow didn't trigger until this bug fix.

@sjw36 Can you get me the test case that led you to find this issue so we can add it in here?

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
@@ -203,9 +203,7 @@ static void insertLoadFromOtherSource(PatternRewriter &b, Location loc,
   std::tie(source, sourceTransformsFromOp) = untransform(b, srcOp);
 
   int64_t copyLength = gemmStoreOp.getLength().getSExtValue();
-  Type typeToLoad = dest.getType().cast<MemRefType>().getElementType();
-  if (copyLength > 1)
-    typeToLoad = VectorType::get({copyLength}, typeToLoad);
+  Type destElemType = dest.getType().cast<MemRefType>().getElementType();
 
   ArrayAttr sourceLeftOob = gemmStoreOp.getLeftOobDims();
   ArrayAttr sourceRightOob = gemmStoreOp.getRightOobDims();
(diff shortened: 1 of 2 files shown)


mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -217,29 +215,50 @@ static void insertLoadFromOtherSource(PatternRewriter &b, Location loc,
...
+
+    Type typeToLoad = destElemType;
+    if (maxVectorLen > 1)
+      typeToLoad = VectorType::get(maxVectorLen, typeToLoad);
+
     auto copyLoop = b.create<TransformingForOp>(
-        loc, ArrayRef<ValueRange>{loadCoord},
-        ArrayRef<Attribute>{sourceTransformsFromOp}, /*bounds=*/consts,
-        /*strides=*/ArrayRef<int64_t>(consts), /*forceUnroll=*/true,
+        loc, ArrayRef<ValueRange>{loadCoord, zeroes},
```
jungpark-mlir, line 249: Was this entirely wrong or just the given vector width doesn't align?
I feel this solution lowers the transform in early stage.

Is there any way to handle it in lowering transformingFor?

krzysz00: This isn't lowering anything?

This just adds a second domain that gives us an actual loop index to store at.

sjw36:
@krzysz00 I have validated that the testcase passes with this change but I think it will need my changes as well. https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/727

sjw36 (approved):
Fixes bug. Thanks!

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
jungpark-mlir: Right, I see.

jungpark-mlir (approved):

PR #850, opened 2022-11-09
[mlir][AMDGPU] Remove buffer ops that are statically out of bounds

krzysz00:
When the bounds check attribute is true, the raw buffer load, store, and atomic operations have well-defined behavior (returning 0 for loads and ignoring stores) when the buffer access exceeds the bounds of the memory being accessed.

Because of how LLVM currently implements these buffer operations (as opaque intrinsics), the backend cannot optimize out this known behavior and eliminate the memory operations. Therefore, use MLIR's canonicalization system to eliminate these operations.

### DIFF
--- a/external/llvm-project/mlir/test/Dialect/AMDGPU/canonicalize.mlir
+++ b/external/llvm-project/mlir/test/Dialect/AMDGPU/canonicalize.mlir
@@ -1,0 +1,132 @@
+// RUN: mlir-opt %s -split-input-file -canonicalize  | FileCheck %s
+
+// CHECK-LABEL: func @known_oob_load
+func.func @known_oob_load(%arg0: memref<4xf32>) -> f32 {
+  // CHECK: %[[zero:.*]] = arith.constant 0.000000e+00 : f32
+  // CHECK: return %[[zero]]
+  %c4_i32 = arith.constant 4 : i32
+  %0 = amdgpu.raw_buffer_load {boundsCheck = true} %arg0[%c4_i32] : memref<4xf32>, i32 -> f32
+  func.return %0 : f32
+}
+
+// -----
+
+// CHECK-LABEL: func @known_oob_load_2d
+func.func @known_oob_load_2d(%arg0: memref<4x4xf32>) -> f32 {
+  // CHECK: %[[zero:.*]] = arith.constant 0.000000e+00 : f32
+  // CHECK: return %[[zero]]
+  %c0_i32 = arith.constant 0 : i32
+  %c4_i32 = arith.constant 4 : i32
+  %0 = amdgpu.raw_buffer_load {boundsCheck = true} %arg0[%c4_i32, %c0_i32] : memref<4x4xf32>, i32, i32 -> f32
+  func.return %0 : f32
+}
+
+// -----
+
+// CHECK-LABEL: func @known_oob_load_2d_on_last
+func.func @known_oob_load_2d_on_last(%arg0: memref<4x4xf32>) -> f32 {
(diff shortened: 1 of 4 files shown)


external/llvm-project/mlir/test/Dialect/AMDGPU/canonicalize.mlir
```diff
@@ -0,0 +1,132 @@
...
+  %c4_i32 = arith.constant 4 : i32
+  %0 = amdgpu.raw_buffer_load {boundsCheck = true} %arg0[%c4_i32] : memref<?xf32>, i32 -> f32
+  func.return %0 : f32
+}
+
+// -----
+
+// CHECK-LABEL: func @no_oob_check
+func.func @no_oob_check(%arg0: memref<4xf32>) -> f32 {
+  // CHECK: %[[loaded:.*]] = amdgpu.raw_buffer_load
```
jerryyin, line 95: [Question] In reality when this scenario happen (that we populate a raw buffer load without bounds check) it will load wrong values?

external/llvm-project/mlir/test/Dialect/AMDGPU/canonicalize.mlir
```diff
@@ -0,0 +1,132 @@
+// RUN: mlir-opt %s -split-input-file -canonicalize  | FileCheck %s
+
+// CHECK-LABEL: func @known_oob_load
+func.func @known_oob_load(%arg0: memref<4xf32>) -> f32 {
+  // CHECK: %[[zero:.*]] = arith.constant 0.000000e+00 : f32
+  // CHECK: return %[[zero]]
```
jerryyin, line 6: Nit: I think this applies to cases below as well. It'd be much robust if you use `CHECK-NOT` instead to explicitly reject the raw_buffer_load.

jerryyin (approved):
Nice work, you used a very robust way to implement the out of bound checks!

external/llvm-project/mlir/test/Dialect/AMDGPU/canonicalize.mlir
krzysz00: Depends on architecture.

On the MI-*, you can't turn bounds checks off, at least not with the buffer structure we're using, so it'll act as if we had called for bounds checks. Though maybe there's a chunk of the manual I missed that'll allow for the behavior.

On gfx10, you can turn bounds checks off completely, and if you do, I think there isn't any strict guarantee about what'll happen. Probably page fault or reading random garbage.

external/llvm-project/mlir/test/Dialect/AMDGPU/canonicalize.mlir
krzysz00: As in, CHECK-NOT in addition to the existing check or instead of it? 

I worded the test this way because I wanted to check the exact semantics of the rewrites

jerryyin: Yeah, I can see what you did make sense. I think this is just a style thing that doesn't impact the correctness of testing. Feel free to leave it.

My reasoning is that since we get rid of a instruction, we should not test whether zero is correctly returned (which should be covered in arith dialect).

krzysz00: My thoughts were that the function originally returns the value of the load, so I wanted to make sure that the return value got replaced by 0 like it should be.

krzysz00:
Landed upstream with style changes at https://reviews.llvm.org/D138146

PR #851, opened 2022-11-10
Add pass to fold equivalent buffer load operations

krzysz00:
Because, in our kernels, there is only one variable that refers to a given input buffer and is passed to the buffer load operations (that is, while a fusion arg might be expand_shape'd, we never use both the unexpanded and expanded memrefs), and because all buffer arguments are distinct pointers on pain of us glaring angrily at you, we can cheaply do an optimization that's harder prove in general: merging memory loads

that is, if we have
x = load A[i]
...
y = load A[i]

, we can replace all uses of y with x and delete the second load.

Long term, if LLVM knew that buffer loads were loads, we could rely on the infrastructure there to handle this more carefully. Or, equivalently, if MLIR had a better alias analysis, we could take advantage of it to be more cautions with it.

As it is, this code is in its own pass (to make it easier to check if that's the issue) and is safe given our current and future plans for our kernel generator, so I think we can risk putting something like this in.


jerryyin (approved):
Implementation looks good and this seems to be a safe thing to do.

> Long term, if LLVM knew that buffer loads were loads, we could rely on the infrastructure there to handle this more carefully. 

How did you find out that LLVM doesn't know buffer loads are loads? Mind add that comment to near the pass so we can delete it (Or make a decision about it) in the future?

krzysz00:
@jerryyin We've known this one for a while - LLVM represents buffer loads using an intrinsic fuction call, which means that they can't be reasoned about like `load` - all LLVM knows is "this reads some memory somewhere"

jerryyin:
@krzysz00 Oh yeah you are right... In other words, any abstract that force us to create a mapping in MLIR level (I think they are all intrinsic under the hood, right?) , LLVM cannot do good reasoning about it.

krzysz00:
The compiler folks have making the buffer loads look like actual loads on their todo list under "it'd be nice but we've got a lot else to do" which means that working on that might be me Q1 2023

PR #852, opened 2022-11-10
Bring up Ubuntu 20.04 + ROCm 5.3.2

jerryyin:


krzysz00 (approved):
LGTM. Have you tested this manually with something like

```
docker build -t ubuntu-20-04-test -f ./mlir/utils/jenkins/Dockerfile.rocm ./mlir/utils/jenkins
```
from the repo root?

jerryyin:
@krzysz00 I have tested through local checks as well as static lib generation manually.

> docker build -t ubuntu-20-04-test -f ./mlir/utils/jenkins/Dockerfile.rocm ./mlir/utils/jenkins

How does it work? You can just execute jenkinsfile by just referencing its location from `./mlir/utils/jenkins`?

krzysz00:
@jerryyin That's not executing Jenkins, that's just building the Docker container locally.

And then you give it a name so you can try to run stuff inside of it if you'd like

jerryyin:
@krzysz00 Okay I was confused because of that additional directory followed in it. After the `-f` option, you can get rid of that `Dockerfile.rocm` because it is gone now, just FYI.

PR #854, opened 2022-11-11
Fix recent broken format and let it check the incoming commit.

jungpark-mlir:


mlir/utils/jenkins/static-checks/premerge-checks.py
```diff
@@ -127,6 +127,6 @@ def run_clang_tidy(base_commit, ignore_config):
 
 
 if __name__ == '__main__':
-  if  not (   run_clang_format('HEAD~1','./mlir/utils/jenkins/static-checks/clang-format.ignore')
-          and run_clang_tidy('HEAD~1', './mlir/utils/jenkins/static-checks/clang-tidy.ignore') ):
+  if  not (   run_clang_format('HEAD^','./mlir/utils/jenkins/static-checks/clang-format.ignore')
```
krzysz00: Hold on a moment - we shouldn't be checking just the previous commit, we should be checking the differences between this branch and miopen-dialect tip.

Is that what this change does?

(That is, if I commit a badly-formatted change `a`, then a commit `b` with no format issues, will I get validation failure?)

jungpark-mlir: I think this script runs clang-format/tidy on the base_commit and check git diff to see if it made any change.
I'm not sure why it used HEAD~1 as a base_commit before.

krzysz00: I think the actual fix here might be to go upstream and grab whatever their premergeChecks.py looks like in case they fixed a bunch of bugs there too.

(IIRC the CI runs on the merge of `miopen-dialect` and the current branch, but I might be wrong about that)

jungpark-mlir: Yes, the title is not correct, it confused.

krzysz00: ???

Also, more fundamentally, can we split this into two PRs? I want to approve your clang-format fixes to unblock all the other PRs but I don't want to merge this change to the pre-merge script until we're confident it's correct.

... also, aren't `HEAD~` and `HEAD^` equivalent?

jungpark-mlir: Yeah, that's fair.
BTW it was `HEAD~1`

krzysz00: Which is the same as `HEAD~`

Now, having vaguely thought about this, if this code is getting the merge base between the branch and `miopen-dialect` or something like that, then it might not be correctly handling complex merge histories.

That is, if your badly-formatted change is in `A` on 
```
A
merge miopen-dialect
B
```

it might go through?

krzysz00 (approved):

krzysz00:
... actually, hold on, other possibility.

If your single commit `A` is immediately on tip, then the merge is a fast-forward and so the clang-format check does nothing

krzysz00:
(so we'd need Jenkins to `--no-ff` on PR CI if we want to keep doing what we're doing, assuming that's the problem)

(the test I'd go for is delibarately introducing a badly-formatted change right onto miopen-dialect, and seeing if that fails static checks or no)

jungpark-mlir:
I assumed clang-format scans all the code, sounds that's not true.

krzysz00:
No, `git clang-format` specifically doesn't scan the whole repo because that would take way too long. It only looks at files you changed, and probably only things you added.

jungpark-mlir:
Yeah, that looks complicated than I thought but still there is something possible.
I'll file a ticket.
(And will check MIGraphX CI to see if anything we can steal.)

jungpark-mlir:
upstream mlir doesn't have this, it has a script to run clang-tidy on all code which it says takes 10h.

jungpark-mlir:
Canceling per #855

PR #856, opened 2022-11-11
Make elementwise utility kernels their own operations

krzysz00:
Fixes
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/730

This gets rid of the gemm_id hack for making """conv2d*""" operations that were actually elementwise kernels (ex. zero init, copying workspace to filter).

- Define rock.zero_init_kernel and rock.converting_copy_kernel to represent the two elementwise kernels that we need for some backwards weight and backwards data cases
- Remove the gemm_id hack from --rock-conv-to-gemm and replace it with rewrite patterns for the elementwise kernels
- Use Conv2DGenerator to select the appropriate utility kernel ops when asked to generate the "convolution kernel" corresponding to one of the utility calls
- Fix up the utility kernel rewrite patterns to allow for kernels that don't vectorize properly - we probably haven't been hitting this usecase but we also have no solid reason to make that sort of thing (as we could get from, for example, bacwards data with a stride of 3) inappricable
- Change the "arch" attribute on Conv2DGenerator results to be the full arch name for consistency with everything else
- Update tests to match


krzysz00:
For example, one of the PR tests that hits wrw atomic add on xdlops now gives

```
 ./bin/rocmlir-gen --arch gfx908 -fil_layout=kcyx -in_layout=nchw -out_layout=nkhw -batchsize=64 -in_channels=128 -in_h=28 -in_w=28 -out_channels=512 -fil_h=1 -fil_w=1 --dilation_h=1 --dilation_w=1 --conv_stride_h=1 --conv_stride_w=1 --padding_h=0 --padding_w=0 -t f32
```

```
module attributes {xmodel.arch = "amdgcn-amd-amdhsa:gfx908"} {
  func.func @rock_conv2d_bwd_weight_gkcyx_ngchw_ngkhw_0(%arg0: memref<1x512x128x1x1xf32>, %arg1: memref<64x1x128x28x28xf32>, %arg2: memref<64x1x512x28x28xf32>) attributes {kernel = 0 : i32, xmodel.arch = "amdgcn-amd-amdhsa:gfx908"} {
    rock.zero_init_kernel %arg0 {arch = "amdgcn-amd-amdhsa:gfx908"} : memref<1x512x128x1x1xf32>
    return
  }
  func.func @rock_conv2d_bwd_weight_gkcyx_ngchw_ngkhw_1(%arg0: memref<1x512x128x1x1xf32>, %arg1: memref<64x1x128x28x28xf32>, %arg2: memref<64x1x512x28x28xf32>) attributes {kernel = 1 : i32, xmodel.arch = "amdgcn-amd-amdhsa:gfx908"} {
    rock.conv2d_bwd_weight(%arg0, %arg1, %arg2) features =  mfma|dot|atomic_add {arch = "amdgcn-amd-amdhsa:gfx908", dilations = [1 : i32, 1 : i32], filter_layout = ["g", "k", "c", "y", "x"], gemm_id = 1 : i32, input_layout = ["ni", "gi", "ci", "hi", "wi"], numCu = 64 : i32, output_layout = ["no", "go", "ko", "ho", "wo"], padding = [0 : i32, 0 : i32, 0 : i32, 0 : i32], strides = [1 : i32, 1 : i32]} : memref<1x512x128x1x1xf32>, memref<64x1x128x28x28xf32>, memref<64x1x512x28x28xf32>
    return
  }
}
```

mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
```diff
@@ -815,7 +816,7 @@ LogicalResult Conv2dGenerator::genConvModule(ModuleOp &module, int kernel_id,
 
   std::vector<NamedAttribute> attributes{
       builder.getNamedAttr("gemm_id", builder.getI32IntegerAttr(gemmId)),
```
giuseros: Can `gemm_id` be made a proper attribute, with a comment in the `td` file,  that explains what it does? I am not sure I fully understand what is used for

giuseros (commented):
Hi @krzysz00 , 
could you point me out where the `gemm_id` attribute is explained? I also left a comment about making it a proper attribute in a `td` file.

mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
krzysz00: ... good point, and, yes, I can go ahead and do that refactoring now.

gemm_id had two functions
- Marking which of the N different `rock.conv2d_bwd_data` operations a particular operation was so as to allow the backwards data algorithm (which is split into multiple kernels) to work correctly
- To mark which of several operations you were in in bwd weight and bwd data kernels so that we'd know which ones were utility kernels

Since this PR removes the hacky part of gemm_id, we can make the useful part (which bwd data op you're in) into an explicit attribute.

Good catch on the refactoring! I'll get on that

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -145,22 +145,33 @@ Type getResultType(Operation *convOp, Value outArg) {
   return nullptr;
 }
 
+static int64_t getUtilityVectorizationLen(ShapedType shape) {
```
giuseros: Could you explain a bit the logic here? Why `kMaxVectorOpLen=4`? Is this the same for every architecture?  If it is, is there any design doc you can cite? Does this mean every thread only computes a single element? What about more than one element per thread?

mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
giuseros: Thanks for the explanation @krzysz00 ! Now it is everything clearer :) !

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
krzysz00: MaxVectorOpLen is 4 words because there're no load ops on anything we target that's more than four words long. You'll see similar magic 4's in the global_load and global_store patterns.

Each threads computes `elemsPerThreads` elements at most, which is currently fixed to `512`

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -161,6 +168,68 @@ def Rock_GemmOp :
   }];
 }
 
+def Rock_ZeroInitKernelOp :
+    Rock_Op<"zero_init_kernel", []>,
+    Arguments<(ins AnyTensorOrMemRef:$buffer,
+                  StrAttr:$arch,
+                  OptionalAttr<I32Attr>:$blockSize,
+                  OptionalAttr<I32Attr>:$gridSize,
+                  OptionalAttr<IndexAttr>:$elemsPerThread)>,
```
jerryyin, line 177: Nit: I think `kUtilityKernelElemsPerThread` and `kUtilityKernelBlockSize` should be one of the tablegen defined constant, and can be applied here if no value is passed in. This avoids the many indirectly layers of the attribute passing. However, I realized that this is from a past implementation pattern so we don't need to address it in this PR.

mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
```diff
@@ -875,14 +883,41 @@ LogicalResult Conv2dGenerator::genConvModule(ModuleOp &module, int kernel_id,
     block->push_front(convOp);
   } break;
   case ConvOpType::BwdData: {
-    auto convOp = builder.create<Conv2DBwdDataOp>(
-        builder.getUnknownLoc(), ArrayRef<Type>{}, args, attributes);
-    block->push_front(convOp);
+    if (kernelId < 0) {
```
jerryyin, line 886: Can this happen? I thought around line 766 when rawKernelId comes in, it can no longer be less than 0?

mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
```diff
@@ -875,14 +883,41 @@ LogicalResult Conv2dGenerator::genConvModule(ModuleOp &module, int kernel_id,
...
+    }
   } break;
   case ConvOpType::BwdWeight: {
-    auto convOp = builder.create<Conv2DBwdWeightOp>(
-        builder.getUnknownLoc(), ArrayRef<Type>{}, args, attributes);
-    block->push_back(convOp);
+    bool hasUtilities = getBwdWeightKernelCount(builder) > 1;
+    if (hasUtilities && kernelId == 0) {
+      // If there is a workspace, zero-init it, otherwise fill the filter tensor
+      auto zeroInitOp = builder.create<ZeroInitKernelOp>(
```
jerryyin, line 903: It is nice cleanup to move this chunk of logic into conv2d generator where it should belong. Thanks!

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -145,22 +145,33 @@ Type getResultType(Operation *convOp, Value outArg) {
...
+  if (!kernelOp.getGridSize().has_value())
+    return kernelOp.emitOpError("grid size not defined for utility kernel");
+  if (!kernelOp.getElemsPerThread().has_value())
+    return kernelOp.emitOpError(
+        "elemsPerThread not defined fer utility kernel");
+  uint32_t blockSize = *kernelOp.getBlockSize();
+  int64_t elemsPerThread = kernelOp.getElemsPerThread()->getSExtValue();
   if (elemsPerThread % vectorLen != 0)
-    return convOp->emitOpError("Unevenly vectorized elementwise kernel");
+    return kernelOp.emitOpError("unevenly vectorized elementwise kernel");
```
jerryyin, line 176: I think this can be better handled when we populate the vectorization length. Make it into a for loop such that as long as vectorization length is not divisible, cut it to half until it is possible.

jerryyin (commented):
Very nice cleanup PR. I've only got some minor nits. We'd be ready to ship once those are sorted out.

mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
krzysz00: Yeah, the function that maps raw kernel IDs (0+) to kernelId still returns `-1` for the utility kernel

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
krzysz00: Well, the vectorization length getter will already ensure this invariant - here is another layer of sanity-checking

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
krzysz00: Tablegen defined constant?

Ah, as in DefalutValuedAttr

... yeah, that's a good future cleanup, someone ought file a ticket for it

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
jerryyin: If I read it right `getUtilityVectorizationLen()` will make its decision mostly on the type width, but didn't take `elemsPerThread` into account. On the other hand, `elemsPerThread` is a very flexible value because it get passed in directly as a op argument. If the externally passed in `elemsPerThread` becomes less than 4 word then this can be invalid. (Realistically if we take into account of `elemsPerThread` when making the vectorization decision we can make sure this never happen)

mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
jerryyin: Okay gotcha, I misread what `rawKernelId` is and turns out this is not an id but rather an index to reference from `kernelIds` (in backward data case only). This is somewhat confusing.

jerryyin (approved):
Since this PR has achieved good amount of cleanups and I'm not worried about some preexisting nits, approved.

mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
krzysz00: It used to be `kernel_id` and `gemm_id`, which was also not great naming. So ... honestly, if you've got better names, now's as good a time as any.

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
krzysz00: ... yep, I should stick `elemsPerThread` into this function. You're right.

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -145,11 +145,13 @@ Type getResultType(Operation *convOp, Value outArg) {
...
 
-static int64_t getUtilityVectorizationLen(ShapedType shape) {
+static int64_t getUtilityVectorizationLen(ShapedType shape,
+                                          int64_t elemsPerThread) {
   int64_t numElems = shape.getNumElements();
   constexpr int64_t kMaxVectorOpLen = 4; // words
   int64_t elemsPerWord = (32 / shape.getElementTypeBitWidth());
-  return math_util::gcd(numElems, kMaxVectorOpLen * elemsPerWord);
+  return math_util::gcd(math_util::gcd(numElems, elemsPerThread),
+                        kMaxVectorOpLen * elemsPerWord);
```
jerryyin, line 154: Yah, this is a better implementation than I suggested, thanks.

PR #858, opened 2022-11-14
Add missing pkg-config dependency

jerryyin:
This is left out by https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/852.

As it turns out `pkg-config` become necessary to build MIOpen. I have locally tested that MIOpen build with latest Dockerfile.

### DIFF
--- a/mlir/utils/jenkins/Dockerfile
+++ b/mlir/utils/jenkins/Dockerfile
@@ -68,15 +68,14 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
   libnuma-dev \
   software-properties-common
 
-RUN apt-add-repository ppa:deadsnakes/ppa && apt-get update \
-     && apt-get install -y --no-install-recommends \
-            python3.8 python3.8-dev python3.8-venv python3-psutil \
-            python3-pip python3-setuptools \
-     && update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.8 10 \
-     && apt upgrade \
-     && python3 -m pip install --upgrade pip \
-     && pip3 install pandas numpy scipy jinja2 tomli \
-     && pip3 install -r requirements.txt
+RUN apt-get update && apt-get install -y --no-install-recommends \
+           python3.8 python3.8-dev python3.8-venv python3-psutil \
+           python3-pip python3-setuptools \
+    && update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.8 10 \
+    && apt upgrade \
+    && python3 -m pip install --upgrade pip \
+    && pip3 install pandas numpy scipy jinja2 tomli \
+    && pip3 install -r requirements.txt
     # ^ pip install pandas... is needed to output performance tests and regenerate test suites
     # Has a copy of the requirements.txt install bit for the new python version
 
--- a/mlir/utils/jenkins/Dockerfile
+++ b/mlir/utils/jenkins/Dockerfile
@@ -76,14 +76,9 @@ RUN apt-add-repository ppa:deadsnakes/ppa && apt-get update \
      && apt upgrade \
      && python3 -m pip install --upgrade pip \
      && pip3 install pandas numpy scipy jinja2 tomli \
-     && pip3 install -r requirements.txt \
-     && sed -i -e 's#/usr/bin/python3#/usr/bin/python3.6#' /usr/bin/apt-add-repository /usr/bin/add-apt-repository
+     && pip3 install -r requirements.txt
     # ^ pip install pandas... is needed to output performance tests and regenerate test suites
     # Has a copy of the requirements.txt install bit for the new python version
-    # Note: the 'sed' above is a result of the python3.8 not being able to use the
-    # system copies of `gobject` and `apt_pkg`, which are build strictly for
-    # Python 3.6. It can be removed, along with the entire invocation of deadsnakes
-    # when we move to Ubuntu >= 20.04.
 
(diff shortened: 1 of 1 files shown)


krzysz00 (commented):
Since you got rid of the comment, can you also remove `apt-add-repository ppa:deadsnakes/ppa && ` above?

krzysz00 (approved):

mlir/utils/jenkins/Dockerfile
```diff
@@ -68,22 +68,16 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
-     && pip3 install pandas numpy scipy jinja2 tomli \
-     && pip3 install -r requirements.txt \
-     && sed -i -e 's#/usr/bin/python3#/usr/bin/python3.6#' /usr/bin/apt-add-repository /usr/bin/add-apt-repository
+RUN apt-get update && apt-get install -y --no-install-recommends \
+           python3.8 python3.8-dev python3.8-venv python3-psutil \
+           python3-pip python3-setuptools \
+    && update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.8 10 \
+    && apt upgrade \
+    && python3 -m pip install --upgrade pip \
+    && pip3 install pandas numpy scipy jinja2 tomli \
```
zhanglx13, line 77: When I tried to build another image on top of this one, it had the error `apt-add-repository: not found`. 
And it worked when I removed this chunk and only kept this line.
Can you give it a try and confirm?

krzysz00: ???

Maybe the `update-alternatives` line is the problem - I'd poke at this, but if the base Docker image works, you can paper over problems in the sub-images.

mlir/utils/jenkins/Dockerfile
```diff
@@ -68,22 +68,16 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
-            python3.8 python3.8-dev python3.8-venv python3-psutil \
-            python3-pip python3-setuptools \
-     && update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.8 10 \
-     && apt upgrade \
-     && python3 -m pip install --upgrade pip \
-     && pip3 install pandas numpy scipy jinja2 tomli \
-     && pip3 install -r requirements.txt \
-     && sed -i -e 's#/usr/bin/python3#/usr/bin/python3.6#' /usr/bin/apt-add-repository /usr/bin/add-apt-repository
+RUN apt-get update && apt-get install -y --no-install-recommends \
+           python3.8 python3.8-dev python3.8-venv python3-psutil \
```
zhanglx13, line 72: If you decide to remove this chunk, you also need to install `python3-venv`.

krzysz00: What do you mean by "this chunk"?

mlir/utils/jenkins/Dockerfile
zhanglx13: I think you are right. It requires `python3.10` in ubuntu 20.04 so 3.8 does not work anymore.

mlir/utils/jenkins/Dockerfile
zhanglx13: The whole `RUN` command.

mlir/utils/jenkins/Dockerfile
krzysz00: Wait, hold on, `apt-add-repositories` should be using system python, which, on 20.04 is 3.8?

krzysz00:
```
% /usr/bin/python3 --version
Python 3.8.10
```

zhanglx13: Ah. It should be the `sed` line that replaced python3 with python3.6.

krzysz00: That sed line is removed now - it was only there because in Ubuntu 18.04 we were manually upgrading to Python 3.8 but needed to keep the apt tools pointed at python3.6

mlir/utils/jenkins/Dockerfile
jerryyin: Do you know who is the consumer of all those dependencies? All in our scripts?

It is strange that we'd call out a particular 3.8 to be the python3 version we use. Do you think we should use python3 instead overall?

krzysz00: We needed to bump to 3.8 because our test scripts used `async`, which landed in 3.7 but Ubuntu 18.04 only had Python 3.6.

I'm pretty sure `-psutil` and maybe  `-venv` are from upstream.

We can probably drop the explicit install of Python.

The install of `pandas`, `numpy`, and so on are needed by the performance scripts

zhanglx13: `python3-venv` is required by MIOpen. Others are used by performance report generator. `tomli` is used to auto generate rock E2E test suits. Not sure about `jinja2`, which does appear in our codebase.

krzysz00: jinja2 is, IIRC, an indirect dependency of Pandas's HTML support but I'd need to poke Google there

jerryyin: @zhanglx13 In which stage does MIOpen need `python3-venv`, but not `python3.8-venv`?

krzysz00: `python3-venv` will be `python3.8-venv` on 20.04

PR #859, opened 2022-11-14
[mlir][SerializeToHsaco] Minimize dependencies of AMDGPU compilation

krzysz00:
The SerializeToHsaco uses functions from ExecutionEngineUtils to set up LLVM pass pipelines, but does not otherwise depend on the execution engine (except indirectly via a dependency on IPO). This commit removes the dependency on the execution engine to prevent unnecessarily compilations.


giuseros (approved):

jerryyin (approved):

krzysz00:
Submitted upstream at https://reviews.llvm.org/D138041

PR #860, opened 2022-11-15
Fix and begin testing mixed-layout convolutions

krzysz00:
Fixes
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/734

- Adds a transformation that runs before conv to gemm conversion that ensures that the filter and input agree on the order of what will be the K dimension (NHW/CYX) and that the input and output agree on the order of N, H, and W. Not having this pass causes incorrect results as the dimensions can be merged in the wrong order
- This pass uses the input tensor as the source of the true layout because I had to pick one and because, in the forward case, it has both gemmK and gemmN dimensions that need to match the output
- Fix the lowering_wrw_atomic_transform_nhwc test to actually be NHWC
- Add unit tests for the layout regularization transforms
- Add E2E tests that check various mixed-layout cases using a slightly larger version of the config that was the minimal reproducer for the earlier bug.
- As a result of said E2E tests, correct the assumption in forward and wrw kernels that N always comes before HW
- Note: the backwarrd data kernels are currently unaffected since they implicitly transform the kernel inputs to NCHW in the process of generating the implicit GEMM

I've manually run these new tests on gfx1030, but I'll want a nightly shared libs run to make sure I didn't miss anything.

Depends on PR #856 because we don't want the layout-fixer running on utility kernel ops.


jerryyin:
@krzysz00 Where do you see this PR going? Context is that we leave the mixed layout support viable internally because we are capable of doing that and want to be flexible. However, we explicitly banned it from MIOpen integration because we don't want client to try a layout of config that we haven't do enough of regression on. Whoever that needs this feature needs explicitly request what layout in particular and we need more emphasis on that layout in `MixedConvLayout.toml`.

 - Internally we need to blacklist layout that we have zero test coverage. It'd be easy to remove that blacklist later but it is important to remind that our testing coverage is limited.
 - If from now on we plan to add full support to mix layout this PR isn't enough. I'd request either a follow-up PR soon about the parameter sweep test that exercise all possible layouts as a third sweeping option.
 - if this is experimental only, we need a follow-up ticket to remember we support this but it is not completely tested.

---

BTW, good reasoning on forcing either cyx or nhw dimensions to align on the ordering. Coming to think of, that's indeed the only restriction we need to make sure gemm computations are aligned across 3 memrefs. But do you really see another rewrite pattern to be the right solution to it? I see this as a indication of bug in `ConvToGemm` pass because the merge ordering was put incorrectly without considering about that they can go out-of-sync between different memrefs.

krzysz00:
@jerryyin This PR arises from the linked bug, which went as follows
- ... -> tosa had an KCYX filter tensor and then transposes that made it KYXC to be acceptable to tosa.conv2d
- --canonicalize folded the transpose into the constant filter tensor
- Which made the layout GKYXC, NGCHW, NGKHW, which failed because of the merge ordering
- Therefore, since mixed layout happened "naturally", we either need to be throwing errors on it on our end or having some support for it
- I've elected to go for support

Re the solution:

The rewrite to fix up the mergeable dimensions isn't the only possible fix. Another would be to, in conv to gemm, track the ordering in one of the tensors and have the other ones reworked to match. The trouble here is that I'd've needed to reorder a bunch of things because, going off the bug above, I want to use the input tensor as the source of truth, but we generally transform the filter tensor first.

And also, the conv to gemm pass transforms are already kinda cluttered, I figured moving teh layout fix to its own rewrite pattern would be a nice bit of separation. I'm willing to explore different options: this is just what came to mind as a reasonable fix.

jerryyin:
> This PR arises from the linked bug, which went as follows

Okay sounds good. So I'll leave it to Tosa/MIGraphX side of people to judge for regression and coverage, as the requirement comes from there.

> The rewrite to fix up the mergeable dimensions isn't the only possible fix. 

I think in the long term, we may need to make more decision on which layout is the best to align on from perf angle. Thinking about this, abstracting it away into a pass (instead of convToGemm) might be a more robust way of doing it. We are aligned now. Thanks for the clarification.

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -306,6 +323,128 @@ struct ConvertingCopyKernelRewritePattern final
...
+    toLayoutIdxs.insert({pair.value(), pair.index()});
+
+  bool inOrder = true;
+  size_t prevIndex = 0;
+  for (StringAttr expected : expectedOrder) {
+    size_t thisIndex = toLayoutIdxs.find(expected)->getSecond();
+    if (thisIndex <
+        prevIndex) { // the values are not in the relative expected order
+      inOrder = false;
+    }
```
jerryyin, line 370: break here when it happens because there's no need to loop through the rest?

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -306,6 +323,128 @@ struct ConvertingCopyKernelRewritePattern final
...
+    // Defensively prevent possible weirdness that might arise when
+    // thisIndex < prevIndex.
+    prevIndex = std::max(thisIndex, prevIndex);
+  }
+  if (inOrder)
+    return failure();
+
```
jerryyin: Nit: If I'm to write this, I prefer to just get rid of this first loop but instead:
 - Use transform to convert the `toLayout` to `newToLayout`
   - Use index as internal lambda implementation details to the transform
 - Then compare whether `toLayout` is equivalent to `newToLayout` to decide whether it is inOrder.

jerryyin (approved):
Got some nits but this all looks mergeable to me. I'm on vacation tomorrow so want to give the approval so as not to block you. But I'd like a second reviewer on TOSA side to double check about the regression coverage.

mlir/include/mlir/Dialect/Rock/IR/RockConvInterface.td
```diff
@@ -22,6 +22,27 @@ def RockConvInterface : OpInterface<"RockConvInterface"> {
   let cppNamespace = "::mlir::rock";
 
   let methods = [
+    InterfaceMethod<
```
jerryyin, line 25: Nite: Rename those to + Shape, like `getFilter()` -> `getFilterShape()`

krzysz00: This isn't getting the shape? It's getting the actual value, which is a `TypedValue<ShapedType`

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
krzysz00: ... point

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
krzysz00: I was trying to have a fast (or at least fast-ish) path for the common case where we need to do nothing.

krzysz00: But yeah, making a transform unconditionally and seeing if it's the identity is also reasonable ... but a lot more work most of the time?

sjw36 (approved):
Looks good. I think this is the right approach using input as the standard. Thanks for quickly addressing this feature.

PR #861, opened 2022-11-16
[DO NOT SQUASH] [XMIR] fixed intermediate handling when shared across devices

sjw36:
* insert gpu.memcpy to/from device
* fixed async.await conversion (should be a synchronous gpu.wait)
* added gpu.is_async attribute on func to disable gpu-async-region

tosa-partition:
* Partition small and zero constants always
* And partition only the first transpose op

xmodel-graph-pipeline:
* Change XModel graph pipeline to disable leading ops in partitioning
          * and disable tosa-optional-decompositions

* Enable E2E testing for torch-mlir resnet18 model
      * add sub-set of resnet18 with multi-kernel

rock-linalg-align:
* fix for major identity
      * seen in torch-mlir ResNet50 model

https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/738


pcf000:
Is the rock-linalg-align patch in response to an "operand does not dominate use" error?  I happen to be seeing that with resnet18 via torch-mlir, and maybe you've already fixed it for me.

krzysz00:
An operand does not dominate use error - especially if it came out of --rock-sugar-to-loops - is something that got fixed last week.

krzysz00:
@sjw36 I suggest that, instead of disabling gpu-async-region, it be patched or reworked to work correctly on async GPU code. (More review to follow.)

external/llvm-project/mlir/lib/Conversion/XModelToGPU/XModelToGPU.cpp
```diff
@@ -91,45 +91,74 @@ struct LaunchRewritePattern : public OpRewritePattern<async::LaunchOp> {
...
+  }
+
+  Value moveMemory(OpBuilder b, async::LaunchOp launchOp, Value opr,
+                   uint32_t fidx, bool readAccess, bool writeAccess,
+                   llvm::SmallVector<Value> &copyBackOprs,
                    llvm::SmallVector<Value, 8> &asyncDeps) const {
     Location loc = opr.getLoc();
     auto tokenType = b.getType<gpu::AsyncTokenType>();
     auto oprAllocOp = opr.getDefiningOp<memref::AllocOp>();
+    auto bAlloc = b;
```
krzysz00, line 120: I think this could be replaced with an insertion point guard? I don't know how expensive the builder copy constructor is

external/llvm-project/mlir/lib/Conversion/XModelToGPU/XModelToGPU.cpp
```diff
@@ -140,6 +169,9 @@ struct LaunchRewritePattern : public OpRewritePattern<async::LaunchOp> {
     if (!kernelPkg.has_value())
       return rw.notifyMatchFailure(op, "no gpu target");
 
+    // Signal that func is already in gpu.async mode
+    caller->setAttr("gpu.is_async", UnitAttr::get(ctx));
```
krzysz00: Per comment on the PR itself ... why not fix up gpu-async-region?

external/llvm-project/mlir/lib/Conversion/XModelToGPU/XModelToGPU.cpp
```diff
@@ -91,45 +91,74 @@ struct LaunchRewritePattern : public OpRewritePattern<async::LaunchOp> {
...
+
     asyncDeps.push_back(dstToken);
     return dstMem;
   }
 
   LogicalResult matchAndRewrite(async::LaunchOp op,
                                 PatternRewriter &rw) const override {
     Location loc = op.getLoc();
-    auto module = op->getParentOfType<ModuleOp>();
+    auto caller = op->getParentOfType<func::FuncOp>();
```
krzysz00, line 165: What if none?

external/llvm-project/mlir/lib/Dialect/GPU/Transforms/AsyncRegionRewriter.cpp
```diff
@@ -339,13 +339,17 @@ struct GpuAsyncRegionPass::SingleTokenUseCallback {
 // inserts the necessary synchronization (as gpu.wait ops). Assumes sequential
 // execution semantics and that no GPU ops are asynchronous yet.
 void GpuAsyncRegionPass::runOnOperation() {
-  if (getOperation()->walk(ThreadTokenCallback(getContext())).wasInterrupted())
+  func::FuncOp func = getOperation();
+  if (func->hasAttr("gpu.is_async"))
```
krzysz00: What was the problem that caused us to need this hack? Can it be solved to make gpu-async-region have the property we want?

external/llvm-project/mlir/lib/Dialect/XModel/Pipelines/Pipelines.cpp
```diff
@@ -41,13 +41,22 @@ void xmodel::buildGraphPipeline(OpPassManager &pm,
...
+
+  // Disable decompositions, converts matmul to fully_connected
+  // pm.addNestedPass<func::FuncOp>(tosa::createTosaOptionalDecompositions());
+
+  // Canonicalizer applies Transpose ops to constant tensors
+  //  - disabled because it causes mixed layout in conv2d
+  // pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
+
+  tosa::TosaPartitionOptions opts;
+  opts.anchorOps = {"tosa.conv2d", "tosa.depthwise_conv2d"};
```
krzysz00, line 75: Why no "tosa.matmul"?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -140,22 +156,27 @@ static Value makeBroadcast(PatternRewriter &b, MemRefType outType, Value inp,
 
       inpType = inp.getType().template cast<MemRefType>();
       inpShape = inpType.getShape();
+    } else if (inpIdxMap.isMinorIdentityWithBroadcasting(&bcastDims)) {
+    } else if (isMajorIdentityWithBroadcasting(inpIdxMap)) {
```
krzysz00, line 160: Good catch on the new case.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -140,22 +156,27 @@ static Value makeBroadcast(PatternRewriter &b, MemRefType outType, Value inp,
...
+    LLVM_DEBUG(llvm::interleaveComma(bcastDims, llvm::dbgs()));
+    LLVM_DEBUG(llvm::dbgs() << "\n");
+
     // 1. insert a broadcast rock.transform
     SmallVector<uint32_t, 8> ptDims;
     SmallVector<int64_t, 8> bcastSizes;
     for (uint32_t dim = 0; dim < inpShape.size(); ++dim) {
-      if (std::find(bcastDims.begin(), bcastDims.end(), dim) !=
-          bcastDims.end()) {
+      if (llvm::is_contained(bcastDims, dim)) {
```
krzysz00, line 179: Nice cleanup!

krzysz00 (commented):
Main comment is wrt the `is_async` bit, other than that, seems broadly reasonable to me.

sjw36:
> @sjw36 I suggest that, instead of disabling gpu-async-region, it be patched or reworked to work correctly on async GPU code. (More review to follow.)

I tried that but the pass specifically assumes synchronous gpu ops that it will make async. We may have some of that from rocmlir-gen flow so we need this pass but the xmodel-to-gpu pass already has async ops (async.launch) which are converted to async gpu.launch ops.. the problem comes with the synchronous 'async.wait' which syncs the host side.

There could be a scenario where there are async and sync gpu.launch's in the same func, so I will revisit this.

external/llvm-project/mlir/lib/Conversion/XModelToGPU/XModelToGPU.cpp
sjw36: The memcpy needs to be inserted at 'b' location, whereas the alloc may be earlier.

external/llvm-project/mlir/lib/Conversion/XModelToGPU/XModelToGPU.cpp
sjw36: async.launch can only reside in a func.

external/llvm-project/mlir/lib/Conversion/XModelToGPU/XModelToGPU.cpp
krzysz00: Yes

So put the bits that need to go earlier inside an extra set of curly braces, summon up an insertion point guard, and then `setInsertionPoint()` on b? Or did I miss a use of `b` vs `bAlloc` that prevents that structure?

krzysz00:
We do have synchronous GPU ops that get made async in the rocmlir-gen flow

And, to make sure I understand the problem - is the issue async.wait getting converted into a gpu.wait? Because that doesn't seem like what the pass is doing?

It could be that the pass has some bugs around handling existing async operations - I know I sent a patch upstream for one such case a while back and kept it in here. https://reviews.llvm.org/D136105 , specifically. We may need some more like that?

external/llvm-project/mlir/lib/Conversion/XModelToGPU/XModelToGPU.cpp
krzysz00: Where's that invariant enforced?

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -115,9 +116,7 @@ bool isElementwiseOp(Operation *op) {
   // clang-format on
 }
 
-bool isFuseableOp(Operation *op) {
-  return isElementwiseOp(op) || isa<tosa::TransposeOp, tosa::ReshapeOp>(op);
-}
+bool isFuseableOp(Operation *op) { return isElementwiseOp(op); }
```
pcf000, line 119: Could just replace isFuseableOp with isElementwiseOp.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -480,21 +479,28 @@ bool TosaPartitionPass::isAnchorOp(Operation *op) {
...
+}
+
+bool TosaPartitionPass::isLeadingOp(Operation *op) const {
+  return isConstantZero(op) || isSmallishConstant(op) || isTransposeOp(op) ||
+         (!trailingOnly && isFuseableOp(op));
 }
 
-bool TosaPartitionPass::isTrailingOp(Operation *op) { return isFuseableOp(op); }
+bool TosaPartitionPass::isTrailingOp(Operation *op) const {
+  return isa<tosa::TransposeOp, tosa::ReshapeOp>(op) || isFuseableOp(op);
```
pcf000, line 492: Use isTransposeOp here.

mlir/test/xmir/mobilenetv1.mid.mlir
```diff
@@ -7,13 +7,8 @@ module {
 // CHECK: func.func @mobilenetv1(%arg0: memref<1x224x224x3xf32>, %arg1: memref<32x3x3x3xf32>, %arg2: memref<3x3x32x1xf32>, %arg3: memref<64x1x1x32xf32>, %arg4: memref<3x3x64x1xf32>, %arg5: memref<128x1x1x64xf32>, %arg6: memref<1x56x56x128xf32>) {
 // CHECK:   %[[TOKEN0:.*]] = async.launch @mobilenetv1__part_0 (%arg0, %arg1, %{{.*}}) : (memref<1x224x224x3xf32>, memref<32x3x3x3xf32>, memref<1x112x112x32xf32>)
 
-// CHECK:   %[[TOKEN1:.*]] = async.launch @mobilenetv1__part_1 [%[[TOKEN0]]] (%{{.*}}, %arg2, %{{.*}}) : (memref<1x112x112x32xf32>, memref<3x3x32x1xf32>, memref<12544x32xf32>)
+// CHECK-DIS:   %[[TOKEN1:.*]] = async.launch @mobilenetv1__part_1 (%{{.*}}, %arg2, %{{.*}}) : (memref<1x112x112x32xf32>, memref<3x3x32x1xf32>, memref<12544x32xf32>)
```
pcf000, line 10: Is this a way to disable a CHECK line?

external/llvm-project/mlir/lib/Conversion/XModelToGPU/XModelToGPU.cpp
```diff
@@ -91,45 +91,74 @@ struct LaunchRewritePattern : public OpRewritePattern<async::LaunchOp> {
...
-    } else {
-      if (readAccess) {
-        // else copy to device
-        auto memcpyToken = b.create<gpu::MemcpyOp>(
-            loc, tokenType, ValueRange{dstToken}, dstMem, opr);
-        dstToken = memcpyToken.getResult(0);
+
+    bool copyFlag = false;
+    auto makeCopy = [&]() {
+      if (!copyFlag) {
```
pcf000: I don't understand the copyFlag logic.  The idea is apparently to make sure that makeCopy() only makes a copy once, but I don't see any path in which makeCopy() is called more than once.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -140,22 +156,27 @@ static Value makeBroadcast(PatternRewriter &b, MemRefType outType, Value inp,
 
       inpType = inp.getType().template cast<MemRefType>();
       inpShape = inpType.getShape();
+    } else if (inpIdxMap.isMinorIdentityWithBroadcasting(&bcastDims)) {
```
pcf000, line 159: Perhaps put a comment in the empty condition to indicate that it's intentionally empty.

external/llvm-project/mlir/lib/Conversion/XModelToGPU/XModelToGPU.cpp
sjw36: This is fixed now. Updated gpu-async-region pass to more flexibly support already async'd gpu ops.

external/llvm-project/mlir/lib/Conversion/XModelToGPU/XModelToGPU.cpp
sjw36: Cleaned up. Yes, it was a left over from a previous process.

external/llvm-project/mlir/lib/Dialect/XModel/Pipelines/Pipelines.cpp
sjw36: Not in this PR.

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
sjw36: I like the naming for now.

mlir/test/xmir/mobilenetv1.mid.mlir
sjw36: yes.

sjw36:
@krzysz00 the hack is now appropriately handled I believe. Please re-review.

external/llvm-project/mlir/lib/Conversion/XModelToGPU/XModelToGPU.cpp
```diff
@@ -91,45 +91,79 @@ struct LaunchRewritePattern : public OpRewritePattern<async::LaunchOp> {
     return b.create<gpu::WaitOp>(loc, tokenType, deps).asyncToken();
   }
 
-  Value moveMemory(OpBuilder b, Value opr, uint32_t fidx, bool readAccess,
-                   bool writeAccess, llvm::SmallVector<Value> &copyBackOprs,
+  template <typename T>
+  bool isOnDevice(const T &oprUsers) const {
+    for (auto opUse : oprUsers) {
+      auto gpuLaunch = dyn_cast<gpu::LaunchFuncOp>(opUse);
```
krzysz00, line 97: This could be an `isa<>`?

external/llvm-project/mlir/lib/Conversion/XModelToGPU/XModelToGPU.cpp
```diff
@@ -91,45 +91,79 @@ struct LaunchRewritePattern : public OpRewritePattern<async::LaunchOp> {
...
+    return true;
+  }
+
+  Value moveMemory(OpBuilder b, async::LaunchOp launchOp, Value opr,
+                   uint32_t fidx, bool readAccess, bool writeAccess,
+                   llvm::SmallVector<Value> &copyBackOprs,
                    llvm::SmallVector<Value, 8> &asyncDeps) const {
+    if (auto gpuAllocOp = opr.getDefiningOp<gpu::AllocOp>()) {
+      // TEST: convergence or multi-input??
+      assert(isOnDevice(opr.getUsers()));
```
krzysz00, line 112: Is this something that must be true at this point? If not, maybe make this guy a `FailureOr<Value>` instead?

external/llvm-project/mlir/lib/Dialect/GPU/Transforms/AsyncRegionRewriter.cpp
```diff
@@ -141,7 +176,7 @@ struct GpuAsyncRegionPass::ThreadTokenCallback {
   // range starts with a `gpu.wait async` op, and ends with a `gpu.wait` op.
   // In between, each gpu::AsyncOpInterface depends on the current token and
   // produces the new one.
-  Value currentToken = {};
+  SmallVector<Value, 4> tokens;
```
krzysz00, line 179: I can't tell if this would make sense as a set or not. On the one hand, we're using it like a set. On the other hand, it's small enough that hash overhead becomes meaningful and also we want deterministic results between passes ... yeah, probably best to have this be a vector

external/llvm-project/mlir/test/Dialect/GPU/async-region.mlir
```diff
@@ -177,9 +177,9 @@ module attributes {gpu.container_module} {
     %t0 = gpu.wait async
     // CHECK: %[[t1:.*]] = gpu.wait async [%[[t0]]]
     %t1 = gpu.wait async [%t0]
-    // CHECK: %[[m:.*]], %[[t2:.*]] = gpu.alloc async [%[[t1]], %[[t0]]] ()
+    // CHECK: %[[m:.*]], %[[t2:.*]] = gpu.alloc async [%[[t0]]] ()
     %0 = gpu.alloc [%t0] () : memref<7xf32>
-    // CHECK: %[[t3:.*]] = gpu.dealloc async [%[[t2]]] %[[m]]
+    // CHECK: %[[t3:.*]] = gpu.dealloc async [%[[t2]], %[[t1]]] %[[m]]
```
krzysz00, line 182: How's the dependency on t1 end up down here? 

I take it that's because this is the first sync op after `t1` was created? If so, probably a good idea to add a comment.

krzysz00 (approved):
We're in the nitpicking zone, but I don't have any fundamental objections to this PR and would be OK with it landing as-is.

I think you should coordinate with @manupak to make sure that the major identity changes are still needed after his PR lands and to figure out who gets to eat the merge conflicts.

external/llvm-project/mlir/test/Dialect/GPU/async-region.mlir
sjw36: It picks up these dependencies because it has no deps.. 
The new change for partially async'd graph is:
* all ops get a token output except a gpu.wait
* only ops that have no dependencies get the current set of unused tokens

In this example, the dealloc is the first op that doesn't have deps and the alloc token is not used so it is added as well as the extra wait at the beginning.

krzysz00: Ah, cool

external/llvm-project/mlir/lib/Dialect/GPU/Transforms/AsyncRegionRewriter.cpp
sjw36: yes, and yes.

external/llvm-project/mlir/lib/Conversion/XModelToGPU/XModelToGPU.cpp
sjw36: It should always be true, I think. I think the tests now cover all cases.

krzysz00: Ok, cool, I just want to avoid crashing the program on things that you're likely to encounter in well-formed input someone might reasonably expect to work.

sjw36:
Looks like the major identity is still needed.

PR #862, opened 2022-11-16
Fix missing dependencies and wrong includes

giuseros:
cc @jungpark-mlir 

This should fix the error you were seeing when compiling MIGraphX. 

There was a missing dependency and the XModel pipelines were using `InitAllDialects.h` and `InitAllPasses.h`

Could you double check it works? If not, let's reiterate until we make it working.


krzysz00 (commented):
Do we need the TransformDialect?

giuseros:
Hi @krzysz00 , actually not :) Let's see what CI thinks about it (it compiles locally)

krzysz00 (approved):

PR #863, opened 2022-11-16
Remove a bunch of old applicability/tuning restrictions

krzysz00:
Going off of the comments and/or what they do, I can find no fundamental reason that the restrictions removed (or, in the case of kPack for various datatypes, loosened) needed to exist now that Jerry's refactoring of gridwise_gemm_v2 (and other previous coordinate transforms refactorings) need to exist.

This commit message will be updated with explanations for fixes for any bugs that this removal uncovers.

### DIFF
--- a/mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
+++ b/mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
@@ -28,6 +28,7 @@ llvm::raw_ostream &mlir::rock::operator<<(llvm::raw_ostream &os,
   case GemmDimension::MorN:
     return os << "GemmDimension::MorN";
   }
+  return os;
 }
 
 static int64_t obtainGridSize(const GemmSize &gemmSize,
--- a/mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
+++ b/mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
@@ -97,14 +97,6 @@ LogicalResult PopulateParams::calculateBlockGemmPerformanceParameters(
       (threadGemmNPerBlock % threadGemmNPerCluster == 0))
     return failure();
 
-  int64_t clusterMPerBlock = threadGemmMPerBlock / threadGemmMPerCluster;
-  int64_t clusterNPerBlock = threadGemmNPerBlock / threadGemmNPerCluster;
-
-  // inline asm only support clusterMPerBlock = 2 andclusterNPerBlock =
-  // 2
-  if (!(clusterMPerBlock == 2 && clusterNPerBlock == 2))
-    return failure();
-
   return success();
 }
 
@@ -119,12 +111,6 @@ LogicalResult PopulateParams::populateDerived(RockGemmWrapperInterface op,
     gemmSize.m += gemmExtraPad->m;
     gemmSize.k += gemmExtraPad->k;
     gemmSize.n += gemmExtraPad->n;
-  }
-
-  if (isa<Conv2DBwdDataOp>(op) &&
-      !(gemmSize.m % 32 == 0 && gemmSize.n % 32 == 0 && gemmSize.k % 4 == 0)) {
-    LLVM_DEBUG(llvm::dbgs() << "Invalid gemm sizes for backward data.\n");
-    return failure();
   }
 
   LogicalResult res = calculateBlockGemmPerformanceParameters(params, op);
@@ -376,38 +362,14 @@ PopulateParamsXDL::isValidBlockwiseGemmXDLOPS(const InitParamsXDL &param,
   if ((param.gemmNPerBlock % param.gemmNPerWave) != 0)
     return failure();
 
-  // TODO Remove. Note KPerBlock and KPack are independent tuning parameters.
-  // There's no need to check if they are divide exactly
-  if ((param.gemmKPerBlock % param.gemmKPack) != 0)
-    return failure();
-
   // Reject invalid KPACK values.
-  // For fp32: reject anything wider than 4.
-  // For fp16/bf16: reject anything narrower than 4, or greater than 8.
-  if (dataType.isF32() && param.gemmKPack > 4) {
-    LLVM_DEBUG(llvm::dbgs() << "Invalid KPACK tuning parameter: "
-                            << param.gemmKPack << "\n");
-    return failure();
-  }
-  if ((dataType.isF16() || dataType.isBF16()) && (param.gemmKPack != 1) &&
-      ((param.gemmKPack < 4) || (param.gemmKPack > 8))) {
-    LLVM_DEBUG(llvm::dbgs() << "Invalid KPACK tuning parameter: "
-                            << param.gemmKPack << "\n");
-    return failure();
-  }
-
-  // XXX FIXME: Temporarily reject KPACK=4 for fp32 backward weight
-  // convolution. It has been verified some configs would cause intermittent
-  // failures.
-  // TODO(whchung): Get to the bottom of this.
-  if ((param.gemmKPack == 4) && isa<Conv2DBwdWeightOp>(op) &&
-      dataType.isF32()) {
+  if (!XdlopsCodeSelection::get(dataType, param.gemmMPerWave,
+                                param.gemmNPerWave)
+           .isValid(param.gemmKPack, param.gemmKPerBlock)) {
     LLVM_DEBUG(llvm::dbgs()
-               << "Invalid config: fp32 XDLOPS backward weight convolution "
-                  "with KPACK=4\n");
+               << "Invalid kPack value " << param.gemmKPack << "\n");
     return failure();
   }
-
   return success();
 }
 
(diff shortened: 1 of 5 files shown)


krzysz00:
Preliminary results from comparing the PR tuning runs against Lixun's most recent PR:
- The non-xdlops tuning space is now substantially bigger
- But that does give us perf improvements (100-ish seconds best tuned results vs 120-ish seconds for the previous)
- xdlops didn't see much growth in tuning space, but it did see slight overall slowdowns, which is weird and probably attributable to machine effects
- I'll be running this through a full weekly, but I have some confidence that this is correct

krzysz00:
Behold, a passing weekly!

http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-863/12/pipeline

krzysz00:
And so ( @manupak , @yiqian1 @jerryyin , and possibly @whchung , @ [Jing Zhang, don't know his github] who put in some of these restrictions ):

At the cost of more tuning time, mainly in the non-xdlops case, this PR removes a bunch of tuning restrictions that are not needed to ensure the correctness of the code.

mlir/utils/widgets/tune_MLIR_kernels.sh
```diff
@@ -9,6 +9,7 @@ printUsage() {
     echo "  -c <\"config\">: Run MIOpenDriver on a single config"
     echo "                 Otherwise, on all configs in resnet50-miopen-configs"
     echo "  -u: turn off tuning (tuning is on be default)"
+    echo "  -b: don't run any tuning, just the compilation actions"
```
jerryyin, line 12: [Question] What's the use case of this option?

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -28,6 +28,7 @@ llvm::raw_ostream &mlir::rock::operator<<(llvm::raw_ostream &os,
   case GemmDimension::MorN:
     return os << "GemmDimension::MorN";
   }
+  return os;
```
jerryyin, line 31: Since you are already passing ostream as reference, it looks redundant to return it.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -85,25 +86,20 @@ LogicalResult PopulateParams::calculateBlockGemmPerformanceParameters(
...
-
-  int64_t clusterMPerBlock = threadGemmMPerBlock / threadGemmMPerCluster;
-  int64_t clusterNPerBlock = threadGemmNPerBlock / threadGemmNPerCluster;
-
-  // inline asm only support clusterMPerBlock = 2 andclusterNPerBlock =
-  // 2
-  if (!(clusterMPerBlock == 2 && clusterNPerBlock == 2))
+  int64_t threadGemmMPerCluster = param.gemmMPerThread *
+                                  derived.mThreadsPerCuwave *
+                                  derived.mCuwavesPerBlock;
```
jerryyin, line 91: Hmm so previously we always assume `gemmMPerThread == 1` and it actually doesn't break anything?

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -85,25 +86,20 @@ LogicalResult PopulateParams::calculateBlockGemmPerformanceParameters(
...
+  int64_t threadGemmNPerCluster = param.gemmNPerThread *
+                                  derived.nThreadsPerCuwave *
+                                  derived.nCuwavesPerBlock;
+
+  if ((param.gemmMPerBlock % threadGemmMPerCluster != 0) ||
+      (param.gemmNPerBlock % threadGemmNPerCluster != 0)) {
+    LLVM_DEBUG(
+        llvm::dbgs()
+        << "M per block or N per block aren't divisible by M/N per cluster\n");
     return failure();
```
jerryyin, line 101: I have no idea why there was the magic number of 2 and 2. It does seem to be a workaround from CK that we just bring it over regardless. (I'm talking about the side-by-side diff in the old code)

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -121,12 +117,6 @@ LogicalResult PopulateParams::populateDerived(RockGemmWrapperInterface op,
     gemmSize.n += gemmExtraPad->n;
   }
 
-  if (isa<Conv2DBwdDataOp>(op) &&
-      !(gemmSize.m % 32 == 0 && gemmSize.n % 32 == 0 && gemmSize.k % 4 == 0)) {
-    LLVM_DEBUG(llvm::dbgs() << "Invalid gemm sizes for backward data.\n");
-    return failure();
-  }
```
jerryyin, line 128: Again very strange we'd just disable this for no good reason. Thanks for chasing this down.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -376,38 +366,14 @@ PopulateParamsXDL::isValidBlockwiseGemmXDLOPS(const InitParamsXDL &param,
   if ((param.gemmNPerBlock % param.gemmNPerWave) != 0)
     return failure();
 
-  // TODO Remove. Note KPerBlock and KPack are independent tuning parameters.
-  // There's no need to check if they are divide exactly
-  if ((param.gemmKPerBlock % param.gemmKPack) != 0)
-    return failure();
```
jerryyin, line 382: This was documented here: https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/475.

According to Jack's comment, it looks like the problematic assumption is getting rid of in our gridwise->blockwise refactoring. Have you bumped into any additional breakages because of this.

Feel free to re-assign that ticket to you and close it once this PR is merged

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -376,38 +366,14 @@ PopulateParamsXDL::isValidBlockwiseGemmXDLOPS(const InitParamsXDL &param,
...
-  // failures.
-  // TODO(whchung): Get to the bottom of this.
-  if ((param.gemmKPack == 4) && isa<Conv2DBwdWeightOp>(op) &&
-      dataType.isF32()) {
+  if (!XdlopsCodeSelection::get(dataType, param.gemmMPerWave,
+                                param.gemmNPerWave)
+           .isValid(param.gemmKPack, param.gemmKPerBlock)) {
     LLVM_DEBUG(llvm::dbgs()
-               << "Invalid config: fp32 XDLOPS backward weight convolution "
-                  "with KPACK=4\n");
```
jerryyin, line 407: Good job!

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -376,38 +366,14 @@ PopulateParamsXDL::isValidBlockwiseGemmXDLOPS(const InitParamsXDL &param,
...
-    LLVM_DEBUG(llvm::dbgs() << "Invalid KPACK tuning parameter: "
-                            << param.gemmKPack << "\n");
-    return failure();
-  }
-  if ((dataType.isF16() || dataType.isBF16()) && (param.gemmKPack != 1) &&
-      ((param.gemmKPack < 4) || (param.gemmKPack > 8))) {
-    LLVM_DEBUG(llvm::dbgs() << "Invalid KPACK tuning parameter: "
-                            << param.gemmKPack << "\n");
-    return failure();
-  }
```
jerryyin, line 397: We may want to remember the rational of this and file ticket to do some performance studies.

The reason why this restriction is here may be due to two reasons:
 - Implicitly there may be a coupling in global load code that just load whatever kpack given to it in one go. I think you have fixed that
 - Implicitly maybe when the kpack is too wide for a type then tuning may just yield sub-optimal solutions and there's no need to keep trying those that are too wide.

So I'm more referring to the second bullet point (performance) than the 1st (correctness).

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -466,22 +422,6 @@ LogicalResult PopulateParamsXDL::populateDerived(RockGemmWrapperInterface op,
   return success();
 }
 
-LogicalResult PopulateParamsXDL::isValidGridGemmXdlops(GemmSize &gemmSize) {
```
jerryyin, line 469: I'm surprise such a filter existed before... Thanks for getting rid of it.

jerryyin (approved):
Nice job touching a chunk of code that no one remembers to touch for a long time. I have some questions but none of them is a blocker. My take-away for such kind of refactoring is that as long as tuning works fine, I'd take any PR that make this derive logic simpler.

mlir/utils/widgets/tune_MLIR_kernels.sh
krzysz00: It's for when I want to build MIOpen but don't want to actually run any configs (ex. I'll be running MIOpenDriver under debugging)

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
krzysz00: GCC was whining about non-void function potentially not returning a value, thought I'd fix

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
krzysz00: No, previously we'd divided by gemmMPerThread before doing the check. There could've been scenarios where it was broken.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
krzysz00: Yeah, idk either. Especially since the comment said something about inline ASM

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
krzysz00: I haven't seen breakages, I'll check the buggy config tomorrow before landing this.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
krzysz00: I figured these were here for correctness which is why I replaced them by the `isValid` check

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
krzysz00: Yep, it works now

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
krzysz00: And yeah, once Yi's tool happens, I'll be happy to do perf studies with it

PR #865, opened 2022-11-17
Breakdown E2E tests into single-test-per-file format

zhanglx13:
This PR:
1. Convert the following tests into toml configuration files.
    `/mlir/test/rocmlir-driver/auto_e2e/conv2d_host_validation_f32_fwd.mlir`
    `/mlir/test/rocmlir-driver/auto_e2e/conv2d_host_validation_f32_bwd.mlir`
    `/mlir/test/rocmlir-driver/auto_e2e/conv2d_host_validation_f16_fwd.mlir`
    `/mlir/test/rocmlir-driver/auto_e2e/conv2d_host_validation_f16_bwd.mlir`
3. Add a `single_mode` (`-s`) option to `generateE2ETest.py` so that each file only contains one test.
4. Enable `single_mode` for test suites in nightly CI.

The single-test-per-file format has the following benefits:
1. Now lit can run all tests instead of exiting a file immediately when it sees a failed test.
2. The error message is more compact, since it now shows THE error for THE test in the file.
3. The `qcm fence wait loop timeout expired` error is gone. Further explanation of this behavior is WIP.


mlir/test/e2e/conv2d_host_validation_f16_fwd.toml
```diff
@@ -0,0 +1,178 @@
+directory = "conv2d_host_validation_f16_fwd"
```
krzysz00: Comment: it looks like a lot of the tests are the same except for NHWC vs NCHW. Can you factor that out? It might lead to one or two more tests.

mlir/test/e2e/conv2d_host_validation_f32_bwd.toml
```diff
@@ -0,0 +1,49 @@
+directory = "conv2d_host_validation_f32_bwd"
```
krzysz00: Similar notes wrt data types.

mlir/test/e2e/generateE2ETest.py
```diff
@@ -76,6 +78,8 @@ def usage():
             output_path = arg
         elif opt == '-i':
             config_file = arg
+        elif opt == '-s':
+            single_mode = 1
```
krzysz00: `= True` : this ain't C

mlir/test/e2e/generateE2ETest.py
```diff
@@ -54,16 +55,17 @@ def generate_option_list(prefixes: dict, table: list, key1: str, key2: str):
...
 def usage():
-    print (f"usage: python3 {sys.argv[0]} [-h | -d <directory> | -i <configfile>]")
+    print (f"usage: python3 {sys.argv[0]} [-h | -d <directory> | -i <configfile> | -s]")
 
 # Main function
 
 if __name__ == '__main__':
     output_path = './'
     config_file = 'config.toml'
+    single_mode = 0
```
krzysz00: = False

krzysz00 (commented):
Minor style nits, file refactoring opportunities.

krzysz00:
Silly question: aren't the host validation tests always `-pv`?

zhanglx13:
@krzysz00 I factored out layouts and data types. Since in the previous test files, fwd tests have three combinations of latyout:
```
-fil_layout=kyxc -in_layout=nhwc -out_layout=nhwk
-fil_layout=kcyx -in_layout=nchw -out_layout=nkhw
-fil_layout=yxck -in_layout=nhwc -out_layout=nhwk
```
and bwd tests only have the first two, I divided the configs into *fwd.toml and *bwd.toml. 
Is there a limitation that bwd conv cannot have the third layout combination?

zhanglx13:
No. In the nightly CI fixed test stage, `-pv_with_gpu` is used.

zhanglx13:
I am not sure if there is a reason that we want to validate these tests with CPU. If not, we should probably remove the `host_validation` from the names.

krzysz00:
The host_validation tests are supposed to be running with CPU, I think, specifically to catch cases where we  have matching bugs in xdlops and non-xdlops.

Heck, there should be a nightly stage that uses `-pv` as opposed to `-pv_with_gpu`

krzysz00:
... yeah, it's the random tests

mlir/test/e2e/CMakeLists.txt
```diff
@@ -28,6 +36,12 @@ if (ROCK_E2E_TEST_ENABLED)
     PaddedGemmConfig.toml
   )
 endif()
+if (ROCMLIR_DRIVER_E2E_TEST_ENABLED)
+  list(APPEND CONFIGS
+    conv2d_host_validation_fwd.toml
```
krzysz00: ... don't do this

This goes in the same list as the other `ROCK_E2E_TEST_ENABLED` configs

krzysz00 (changes requested):
Ok, I have one nit about the CMake file (namely, this now gets folder into ROCK_E2E_TESTS_ENABLED) that I'm going to make a blocker.

Other than that, LGTM, and if you want to change the `host_validation` name to something like `regressions`, that'd be good.

(and ideally port the comments about where we got these tests from over)

krzysz00:
@zhanglx13 Re the backward data limitation ... probably because no one tried a `yxck` layout in those configs. There's no fundamental reason it can't work, but I don't think there's any demand for it so it's probably safe to not test for it.

mlir/test/e2e/CMakeLists.txt
zhanglx13: Why they cannot be on the same list?
Or you want these regression tests to be wrapped under a different test suite name?

krzysz00: The generation of the E2E tests from toml is controlled by `ROCK_E2E_TEST_ENABLED`, not `ROCMLIR_DRIVER_E2E_TEST_ENABLED`

krzysz00: That is, put these in that same list as `PaddedGemm.toml`

zhanglx13: I put them under `ROCMLIR_DRIVER_E2E_TEST_ENABLED` since they used to be controlled by this variable. But it turns out that `ROCMLIR_DRIVER_E2E_TEST_ENABLED` and `ROCK_E2E_TEST_ENABLED` always have the same value during our CIs. You are right that it makes more sense to put them under `ROCK_E2E_TEST_ENABLED`.

krzysz00: The TOML refactoring is causing us to move to that variable over time.

krzysz00 (approved):

PR #866, opened 2022-11-17
Add bert configs to CI

giuseros:
- I created a script to convert `rocblas-bench` input parameters to `perfRunner` parameters. 
- I added both the raw trace and the `perfRunner` ready trace into the commit
- I changed the rocblas driver to account for row vs column major formats
- I changed the `perfRunner.py` script so that if types or layout is already defined the script does not loop over them


mlir/utils/performance/bert-configs-raw
```diff
@@ -0,0 +1,15 @@
...
+./rocblas-bench -f gemm_ex --transposeA N --transposeB N -m 768 -n 16384 -k 768 --alpha 1 --a_type f32_r --lda 768 --b_type f32_r --ldb 768 --beta 0 --c_type f32_r --ldc 768 --d_type f32_r --ldd 768 --compute_type f32_r --algo 0 --solution_index 0 --flags 1
+./rocblas-bench -f gemm_strided_batched_ex --transposeA N --transposeB N -m 64 -n 256 -k 256 --alpha 1 --a_type f32_r --lda 64 --stride_a 16384 --b_type f32_r --ldb 256 --stride_b 65536 --beta 0 --c_type f32_r --ldc 64 --stride_c 16384 --d_type f32_r --ldd 64 --stride_d 16384 --batch_count 768 --compute_type f32_r --algo 0 --solution_index 0 --flags 1
+./rocblas-bench -f gemm_strided_batched_ex --transposeA T --transposeB N -m 256 -n 256 -k 64 --alpha 1 --a_type f32_r --lda 64 --stride_a 16384 --b_type f32_r --ldb 64 --stride_b 16384 --beta 0 --c_type f32_r --ldc 256 --stride_c 65536 --d_type f32_r --ldd 256 --stride_d 65536 --batch_count 768 --compute_type f32_r --algo 0 --solution_index 0 --flags 1
```
krzysz00: Nit: newline

mlir/utils/performance/bert-configs
```diff
@@ -0,0 +1,18 @@
+# This file has been generated with the following command:
+# convertRocBlasToPerfRunner.py -c bert-configs-raw --output-file bert-configs
+
+-transA false -transB false -g 64 -m 1024 -n 384 -k 1024 -t f32
```
krzysz00: Hold on, rocBLAS is column-major but these configs are row-major, so these need to be backwards

mlir/utils/performance/convertRocBlasToPerfRunner.py
```diff
@@ -0,0 +1,84 @@
...
+
+def stringify(config):
+    return ' '.join(key + " " + val for key, val in config.items())
+
+
+def convertToPerfRunner(rocblasIns):
+    perfRunnerIns = {}
+
+    # Default values (from rocblas-bench)
+    perfRunnerIns["-transA"] = "false"
```
krzysz00, line 16: These should be backwards

krzysz00 (changes requested):
I found one problem

However, it might be worth looking into MIGraphX to see if they're manually transposing to get things rocblas-shaped?

krzysz00 (approved):
This all looks good to me!

PR #868, opened 2022-11-18
Add support for canonical linalg fusion

manupak:
This commit enables the drop unit dims pass
to get the linalg generic op to a canonical form.
Then AlignTiling pass has been modified to fold
the views into indexing maps to add the proper
transforms.

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/688
closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/723

### DIFF
--- a/mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
+++ b/mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
@@ -28,6 +28,7 @@
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/Linalg/Passes.h"
 #include "mlir/Dialect/Linalg/Transforms/Transforms.h"
+#include "mlir/Dialect/MemRef/Transforms/Passes.h"
 
 #include "mlir/Conversion/RocMLIRPasses.h"
 #include "mlir/Dialect/Bufferization/Transforms/OneShotAnalysis.h"
@@ -61,6 +62,7 @@ void rock::buildBufferizePipeline(OpPassManager &pm,
   /* rocmlir-opt --linalg-fuse-elementwise-ops
    */
   pm.addNestedPass<func::FuncOp>(createLinalgElementwiseOpFusionPass());
+  pm.addNestedPass<func::FuncOp>(createLinalgFoldUnitExtentDimsPass());
 
   // for tosa control flow
   /* rocmlir-opt --tosa-to-scf --tosa-to-arith
@@ -81,6 +83,8 @@ void rock::buildBufferizePipeline(OpPassManager &pm,
 
   pm.addPass(createConvertTensorToLinalgPass());
   pm.addNestedPass<func::FuncOp>(createLinalgInitTensorToAllocTensorPass());
+  pm.addNestedPass<func::FuncOp>(createLinalgFoldUnitExtentDimsPass());
+  pm.addNestedPass<func::FuncOp>(memref::createFoldMemRefAliasOpsPass());
 
   bufferization::OneShotBufferizationOptions bufOpts;
   bufOpts.allowReturnAllocs = true;
--- a/mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
+++ b/mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
@@ -84,7 +84,6 @@ void rock::buildBufferizePipeline(OpPassManager &pm,
   pm.addPass(createConvertTensorToLinalgPass());
   pm.addNestedPass<func::FuncOp>(createLinalgInitTensorToAllocTensorPass());
   pm.addNestedPass<func::FuncOp>(createLinalgFoldUnitExtentDimsPass());
-  pm.addNestedPass<func::FuncOp>(memref::createFoldMemRefAliasOpsPass());
 
   bufferization::OneShotBufferizationOptions bufOpts;
   bufOpts.allowReturnAllocs = true;
--- a/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
@@ -59,6 +59,166 @@ struct RockLinalgAlignPass
   void runOnOperation() override;
 };
 
+// This rewrite will rewrite the linalg IO that has view like-ops surrounding
+// them to be consumed by the linalg operation itself adjusting the indexing
+// maps to faithfully represent them.
+struct InlineViewLikeOperandsLinalgRewritePattern
+    : public OpRewritePattern<linalg::GenericOp> {
+  using OpRewritePattern<linalg::GenericOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(linalg::GenericOp laGeneric,
+                                PatternRewriter &rewriter) const override;
+};
+
+// This function will take strides calculated using reassociation of dimensions
+// that are modified by view-like ops and recompute the affine expr from src
+// view to dest view.
+llvm::SmallVector<mlir::AffineExpr, 4>
+getDelinearizedAffineExpr(mlir::ArrayRef<int64_t> strides,
+                          mlir::ArrayRef<int64_t> shapes, Builder &b,
+                          unsigned int position) {
+  AffineExpr resultExpr = b.getAffineDimExpr(position);
+  int64_t rank = strides.size();
+  SmallVector<AffineExpr, 4> vectorOffsets(rank);
+  // If the rank is 1, expand or collapse shapes will just
+  // pass-through the dimensions.
+  if (rank == 1) {
+    vectorOffsets[0] = resultExpr;
+    return vectorOffsets;
+  }
+  for (unsigned i = 0; i < rank; i++) {
+    // If the current shape is 1 and the rank is non-zero,
+    // could only mean it is being broadcasted. Hence,
+    // putting zero.
+    if (shapes[i] == 1) {
+      vectorOffsets[i] = resultExpr * 0;
+    } else {
+      // Recording the vector offsets here.
+      vectorOffsets[i] = resultExpr;
+      // There is no point of putting a modulo if the size
+      // is equivalent to that.
+      if (i - 1 >= 0 && shapes[i] != strides[i - 1]) {
+        vectorOffsets[i] = vectorOffsets[i] % strides[i - 1];
+      }
+      vectorOffsets[i] = vectorOffsets[i].floorDiv(strides[i]);
+
+      // The resultExpr has to propagated anyway for
+      // other dimensions where the recording in the above
+      // will do the neccesary checks to remove the modulo
+      if (i - 1 >= 0) {
+        resultExpr = resultExpr % strides[i - 1];
+      }
+      resultExpr = resultExpr.floorDiv(strides[i]);
+    }
+  }
+  return vectorOffsets;
+}
+
+// This function will traverse the operands of the linalg.generic folding
+// the view like ops to the indexing maps and returning the ultimate
+// folded map as well as the root operand.
+LogicalResult
+foldViewLikeOperands(PatternRewriter &rewriter, Value op, AffineMap &foldedMap,
+                     Value &rootOp,
+                     SmallVectorImpl<Operation *> &toBeErasedViewLikeOps) {
+  if (memref::CollapseShapeOp collapseOp =
+          op.getDefiningOp<memref::CollapseShapeOp>()) {
+    SmallVector<AffineExpr, 4> resultExprs;
+    int iDimCount = 0;
+    for (SmallVector<int64_t, 2> groups :
+         collapseOp.getReassociationIndices()) {
+      assert(!groups.empty() && "association indices groups cannot be empty");
+      unsigned groupSize = groups.size();
+      SmallVector<int64_t> suffixProduct(groupSize);
+      // Calculate suffix product for all collapse op source dimension sizes.
+      suffixProduct[groupSize - 1] = 1;
+      for (unsigned i = groupSize - 1; i > 0; i--)
+        suffixProduct[i - 1] =
+            suffixProduct[i] * collapseOp.getSrcType().getDimSize(groups[i]);
+      // Derive the index values along all dimensions of the source
+      // corresponding to the index wrt to collapsed shape op output.
+      SmallVector<AffineExpr, 4> srcIndexExpr = getDelinearizedAffineExpr(
+          suffixProduct, collapseOp.getSrcType().cast<ShapedType>().getShape(),
+          rewriter, iDimCount++);
+      for (unsigned i = 0; i < groupSize; i++) {
+        resultExprs.push_back(srcIndexExpr[i]);
+      }
+    }
+    auto representativeMap = AffineMap::get(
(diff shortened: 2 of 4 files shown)


mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
```diff
@@ -81,6 +83,8 @@ void rock::buildBufferizePipeline(OpPassManager &pm,
 
   pm.addPass(createConvertTensorToLinalgPass());
   pm.addNestedPass<func::FuncOp>(createLinalgInitTensorToAllocTensorPass());
+  pm.addNestedPass<func::FuncOp>(createLinalgFoldUnitExtentDimsPass());
+  pm.addNestedPass<func::FuncOp>(memref::createFoldMemRefAliasOpsPass());
```
krzysz00: Huh, what's this pass do?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -59,6 +59,166 @@ struct RockLinalgAlignPass
...
+  using OpRewritePattern<linalg::GenericOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(linalg::GenericOp laGeneric,
+                                PatternRewriter &rewriter) const override;
+};
+
+// This function will take strides calculated using reassociation of dimensions
+// that are modified by view-like ops and recompute the affine expr from src
+// view to dest view.
+llvm::SmallVector<mlir::AffineExpr, 4>
```
krzysz00: Nit: You can take in `SmallVectorImpl<AffineExpr> &out` here and save on copies in the loop below.

Also, neither SmallVector nor AffineExpr need namespaces on them, I think

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -59,6 +59,166 @@ struct RockLinalgAlignPass
...
+// folded map as well as the root operand.
+LogicalResult
+foldViewLikeOperands(PatternRewriter &rewriter, Value op, AffineMap &foldedMap,
+                     Value &rootOp,
+                     SmallVectorImpl<Operation *> &toBeErasedViewLikeOps) {
+  if (memref::CollapseShapeOp collapseOp =
+          op.getDefiningOp<memref::CollapseShapeOp>()) {
+    SmallVector<AffineExpr, 4> resultExprs;
+    int iDimCount = 0;
+    for (SmallVector<int64_t, 2> groups :
```
krzysz00: Can you take this by const reference?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -59,6 +59,166 @@ struct RockLinalgAlignPass
...
+
+  LogicalResult matchAndRewrite(linalg::GenericOp laGeneric,
+                                PatternRewriter &rewriter) const override;
+};
+
+// This function will take strides calculated using reassociation of dimensions
+// that are modified by view-like ops and recompute the affine expr from src
+// view to dest view.
+llvm::SmallVector<mlir::AffineExpr, 4>
+getDelinearizedAffineExpr(mlir::ArrayRef<int64_t> strides,
```
krzysz00: Third nit: make these functions `static`

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -59,6 +59,166 @@ struct RockLinalgAlignPass
...
+    }
+    auto representativeMap = AffineMap::get(
+        /*numDims=*/collapseOp.getType().cast<ShapedType>().getShape().size(),
+        /*numSymbols=*/0, resultExprs, rewriter.getContext());
+    foldedMap = representativeMap.compose(foldedMap);
+    toBeErasedViewLikeOps.push_back(collapseOp);
+    return foldViewLikeOperands(rewriter, collapseOp.getViewSource(), foldedMap,
+                                rootOp, toBeErasedViewLikeOps);
+  }
+  if (memref::ExpandShapeOp expandShapeOp =
```
krzysz00: Question: given that both collapse and expand shape use the reassociation expressions concept, couldn't you generalize the collapse shape logic into a function that'll allow you to handle both? I'd rather us not have to trip over this in the future.

But if it's too much bother, no problem here.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -59,6 +59,166 @@ struct RockLinalgAlignPass
...
+               << "If needed, it would be similiar to add logic as above.\n");
+    return failure();
+  }
+  rootOp = op;
+  return success();
+}
+
+LogicalResult InlineViewLikeOperandsLinalgRewritePattern::matchAndRewrite(
+    linalg::GenericOp laGeneric, PatternRewriter &rewriter) const {
+  auto loc = laGeneric.getLoc();
```
krzysz00: `Location loc`

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -59,6 +59,166 @@ struct RockLinalgAlignPass
...
+    }
+    if (rootOp != output)
+      changedIOCount++;
+    newOutputs.push_back(rootOp);
+    newMaps.push_back(resMap);
+  }
+  if (changedIOCount == 0)
+    return failure();
+
+  SmallVector<StringRef> iteratorTypes = llvm::to_vector<4>(
```
krzysz00, line 229: I take it `getInputMutable()` and friends didn't come through here? Or, in the case of the affine maps, `setAttr`?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -435,16 +558,15 @@ LogicalResult MILARewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
   }
 
   SmallVector<AffineMap> idxMaps = laGeneric.getIndexingMapsArray();
-  // Output must be indexed by identity map for this to work
```
krzysz00, line 438: Huh. Is the thing you're doing with the inverses making this no longer the case?

If so, I still think there's a condition here. I saw `inversePermutation()` called in earlier code, so I'd like this if statement to check that the output index map is a permutation. And that it is equal to the map on the gemm output/corresponding la.generic input.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -435,16 +558,15 @@ LogicalResult MILARewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
...
   AffineMap outIdxMap = idxMaps.back();
-  if (!outIdxMap.isIdentity())
-    return failure();
 
   // 1. Trace input to global_store.
   // 1.1. Find the (implicit) gemm output
   GlobalStoreOp gemmStoreOp;
   for (auto pair : llvm::zip(idxMaps, laGeneric.inputs())) {
     AffineMap inpIdxMap = std::get<0>(pair);
+    auto invertOutIdxMap = inversePermutation(outIdxMap);
```
krzysz00: Hoist this out of the loop?

krzysz00 (commented):
Mostly nitpicking, this is a pretty solid cleanup.

I'm a bit hesitant about the removal of all the handling for expand_shape, but if the drop-unit-dims caused those to go away then it's *probably* fine.

(on that note, can you add tests to make sure fusion fails in the presence of expand_shape on the gemm result?)

Overall, nice work!

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Done

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Done

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Done. (Good thing I looked at this on a monday morning rather than a friday evening.. :) )

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Done

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Done

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Done

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
manupak: Ah not needed really.. this was a remnant from the earlier approach to the problem.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Well; it might work but figured out the builder was not the issue at that time.
I didn't see much difference in the two approaches..

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: So the check was too restrictive, IMO.

Lets see what we are doing here : 

We are trying to fuse a subsequent kernel to the lowest level vector operation.
So there are three coordinate spaces in the question : input space, operator space, and output space.

The Linalg generic define the indexing maps as operator space --> input space and operator space --> output space.
So the overly restrictive logic here was that we assume operator space --> output space (i.e. which is what "Output must be indexed by identity map for this to work" is supposed to mean, I think) is an identity --> so that we can treat : 

operator space --> input space == output space --> input space.

This is what I felt as overly restrictive and we dont need that.

So the generic way to support that is : 

We invert and compose : invert(op --> output) x (op --> input) == output --> input space which is what we are after. Therefore, we can properly ignore the variety of ways the linalg body could define the operation rather than being special cased here. All the indexing maps has to be invertible and that is guranteed by the construction of linalg generic for shape-to-loops hooks and vice-versa hooks to work.

Right, having said all that, I think if we still need conditions it should be in the form of iterator types and cyclic usage of output space within the op.

manupak (commented):
Thanks @krzysz00 for the super quick review.

I've addressed most comments (and have added back the expand_dims support in anycase as the code was now layered properly add that with ease -- though I still think it is unlikely we encounter them as before because the cases had to do with additional unit dims linalg passes add) and left some explanation for the two of the points.

let me know what you think.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -362,18 +516,21 @@ static Value reconfigureLAGeneric(PatternRewriter &b,
   auto regType = laIn.getType().template cast<MemRefType>();
   auto laOut = b.create<GpuAllocOp>(loc, regType);
 
+  AffineMap outIdxMap = idxMaps.back();
```
krzysz00, line 519: Random note: if  we don't have the check already, let's make sure somewhere ta the beginning that the `linalg.generic` has exactly one output.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -453,19 +596,31 @@ LogicalResult MILARewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
...
         return failure();
       }
-      if (!inpIdxMap.isIdentity()) {
-        LLVM_DEBUG(llvm::dbgs() << "Writeback input not read with identity map "
-                                   "- earlier lowering should've fixed\n");
+      SmallVector<unsigned> broadcastedDims;
+      // This is not to allow broadcasting but due to canonical linalg
+      // form if the unit dims can carry affine expr to be zero in the
+      // translation, hence there is a following check.
+      if (!outToInMap.isMinorIdentityWithBroadcasting(&broadcastedDims)) {
```
krzysz00, line 603: Ok, yeah, given the comment, this is the condition I'm interested in given your explanation above.
What we need is that the map output -> input be the identity, ideally, except that it some cases we'll get zeroes because of how broadcasts get squished down, so this is the condition we'd want.

My paranoia also says that we may also want to impose, in this particular case, the condition that
```
inv(op -> output) ; (op -> input) == inv(op -> input) ; op -> output
```
(where `;` is "and then"/composition), but I'm not sure the zeroes will work the same in both directions.

The underlying thing I want to make sure doesn't happen is the presence of non-trivial operations in the maps for the gemm result and the final output. So my gut is telling me that there's additional conditions on the maps for the writeback inputs and the final output.

That is, consider the following (which we might handle earlier but definitely should check for here)
```
%a = [gemm output]
%b = [elementwise add]
%c = [final result]
linalg.generic indexing_maps = [(d0, d1, d2) -> (d1, d2, d0), (d0, d1, d2) -> (d0, d1, d2), (d0, d1, d2) -> (d1, d2, d0)] ins(%a, %b) outs (%c) {
  bbN(%arg3, %arg4, %arg5) {
    %0 = add %arg3, %arg4
     linalg.yield %0
}
```

inv(output -> op) is `(d0, d1, d2) -> (d2, d0, d1)`, and so we have inv(output -> op) ; (op -> writeback) = identity

Similarly, looking at the bias tensor, the composition gives the map (d0, d1, d2) -> (d2, d0, d1)

Which ... might actually work, come to think of it

Because that says "to go from writeback coordinates to bias coordinates, permute (d0, d1, d2) -> (d2, d0, d1)"
which is equivalent to "to go from *bias* coordinates to *writeback* coordinates, permute the inverse of that, ie (d0, d1, d2) -> (d1, d2, d0)", which is ... correct.

But whether this'll work out correctly when all the maps are non-trivial permutations ... probably.

krzysz00 (approved):
Looks good to me - or at least I've convinced myself your logic is sound

jungpark-mlir (approved):
Look good to me, can you also test if this works with transpose?

krzysz00:
Speaking of, @manupak, as extra testing, go into `FoldTranspose.cpp` and remove the pass `FoldTransposingConvAccess` from the set of passes that run. If your code's correct, then tests should continue to pass.

(and, if my theory of the whole setup is correct, then removing that pass on `miopen-dialect` will cause tests to fail)

manupak:
> Look good to me, can you also test if this works with transpose?

As we discussed offline, mlir/test/fusion/tosa-to-rock-tp-add-*.mlir tests cover these already.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
sjw36: Yes, it was intentionally strict to handle the simple case and this is a much better generalization. Thanks @manupak !

sjw36 (approved):
Great generalization @manupak!

PR #869, opened 2022-11-18
conv2d_regression e2e tests failure with float random inputs

zhanglx13:
When using `-rand_type float` for conv2d regression tests, a lot of tests failed.
This PR resolves the issue by adjusting random range and thresholds.
- random range for float inputs are parameterized in `rocmlir-gen` as `-rand_min` and `-rand_max`.
   - The default float random range is kept as [-1,1]. 
   - For conv2d regression tests, the float random range is set to [1,3].
- `relDiff_threshold` is now used to set relative diff threshold for all data types except for f16, 
   whose relDiff threshold stays as 100.0f.
   - The default value of `relDiff_threshold` is 1e-6.
   - For fwd regression tests, `relDiff_threshold` is set to 1e-5.
   - For bwd regression tests, `relDiff_threshold` is set to 3.5e-5.
- `RMS_threshold` is set to 7e-4 for f16 bwd regression tests. 

This PR also adds a `-p_verify=summary` mode as a verification print option. This mode is similar to `-p_verify=failure` in that it prints debug info only when the test fails. The difference is that `-p_verify=summary` only prints the max values of each metric and the histogram of relDiff. It does not print the element wise information as `-p_verify=faliure` does.
The `summary` mode is set to be the default. We can have a better view of `check-rocmlir` when some tests fail, especially on jenkins.

This PR closes the following two tickets:
- https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/621
- https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/646

### DIFF
--- a/mlir/test/e2e/conv2d_regression_fwd.toml
+++ b/mlir/test/e2e/conv2d_regression_fwd.toml
@@ -1,6 +1,6 @@
 directory = "conv2d_regression_fwd"
 prefix = "rocmlir-gen"
-suffix = "--arch %arch %pv %random_data %rocmlir_gen_flags | rocmlir-driver -c | mlir-cpu-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix="
+suffix = "--arch %arch %pv %random_data %rocmlir_gen_flags -rand_min=1 -rand_max=3 -relDiff_threshold=0.00001 | rocmlir-driver -c | mlir-cpu-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix="
 
 [[axis]]
 name = "operation"
--- a/mlir/test/e2e/conv2d_regression_fwd.toml
+++ b/mlir/test/e2e/conv2d_regression_fwd.toml
@@ -16,6 +16,10 @@ name = "data type"
 values = ["f32", "f16"]
 prefix = "-t "
 
+[[axis]]
+name = "rand type"
+values = ["float"]
+prefix = "-rand_type "
 
 [[suite]]
 name = "config"
--- a/mlir/test/e2e/conv2d_regression_fwd.toml
+++ b/mlir/test/e2e/conv2d_regression_fwd.toml
@@ -1,6 +1,6 @@
 directory = "conv2d_regression_fwd"
 prefix = "rocmlir-gen"
-suffix = "--arch %arch %pv %random_data %rocmlir_gen_flags -rand_min=1 -rand_max=3 -relDiff_threshold=0.00001 | rocmlir-driver -c | mlir-cpu-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix="
+suffix = "-rand_type float -rand_min 1 -rand_max 3 -relDiff_threshold 0.00001 -arch %arch %pv %random_data %rocmlir_gen_flags | rocmlir-driver -c | mlir-cpu-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix="
 
 [[axis]]
 name = "operation"
@@ -16,14 +16,8 @@ name = "data type"
 values = ["f32", "f16"]
 prefix = "-t "
 
-[[axis]]
-name = "rand type"
-values = ["float"]
-prefix = "-rand_type "
-
 [[suite]]
 name = "config"
-
 
 [[suite.test]]
 config = "-p"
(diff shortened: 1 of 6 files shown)


krzysz00:
Is this perhaps a MI100/MI200 issue? Maybe related to denormal handling?

mlir/test/e2e/conv2d_regression_fwd.toml
```diff
@@ -16,6 +16,10 @@ name = "data type"
 values = ["f32", "f16"]
 prefix = "-t "
 
+[[axis]]
```
krzysz00: Move this into `suffix` as well?

mlir/test/e2e/conv2d_regression_fwd.toml
```diff
@@ -1,6 +1,6 @@
 directory = "conv2d_regression_fwd"
 prefix = "rocmlir-gen"
-suffix = "--arch %arch %pv %random_data %rocmlir_gen_flags | rocmlir-driver -c | mlir-cpu-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix="
+suffix = "--arch %arch %pv %random_data %rocmlir_gen_flags -rand_min=1 -rand_max=3 -relDiff_threshold=0.00001 | rocmlir-driver -c | mlir-cpu-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix="
```
krzysz00: Keep the `%blah` sections last, I'd suggest

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -485,15 +494,17 @@ static llvm::cl::opt<float>
 static llvm::cl::opt<float>
     relDiffThreshold("relDiff_threshold",
                      llvm::cl::desc("Threshold for relDiff metric"),
-                     llvm::cl::value_desc("error"), llvm::cl::init(100.0f));
+                     llvm::cl::value_desc("error"), llvm::cl::init(0.000001f));
 static llvm::cl::opt<std::string> printVerifyResults(
```
krzysz00: We could define an `enum class VerificationPrinting : char` and then `static_cast` it away instead of string switching?

krzysz00 (commented):
Minor structural commentary

krzysz00 (approved):
All this looks pretty reasonable to me - approved!

jerryyin:
It was on my list for review and when I finally have time to look at this, it is already merged. @zhanglx13 Thanks for the effort in this, good work in understanding and coming up with the fix for those annoying intermittent failed tests.

I still have some philosophical questions that I think would help us make decisions should we see similar kind of issues in the future.

- How did you come up with the right threshold you had? 3.5e-5 seemed pretty hard to come up with if it is a trial-and-error method and this make me think you might have a better way to pick this.
- What about the range? Why do you have the intuition to get rid of the negative ones in particular? Is this again a denorm related issue that MI200 use yet another way to handle the denorms?
- Do you see a more automated strategy in handling this? For example, use relative diff when and only when no denorm (or, when the range include both positive and negatives, which tend to cancel out each other) exists, because it is inherently flawed to on that scenario.

zhanglx13:
Ah, I thought you were busy with something else. I'll ping you in the future to make sure I get your feedback before merging PRs.

1. The verification function can print out max values for each metric. I just picked the threshold that is slightly larger than the max one.
2. I think the point is to get rid of inputs near zero. For fwd conv, small inputs lead to small outputs (do not necessarily fall into the denorm range), which is very likely to produce large relative element-wise errors. And for bwd convs, small inputs lead to very large outputs (do not necessarily overflow), which is very likely to produce large absolute element-wise errors. Therefore, getting rid of values around zero not only keeps us away from denorm values, but also help the results to be in a more "pretty" range.
What we learned previously about denorms was about f16, in which case small values usually mean denorm values. And we think f32 does not have such a problem. This PR found out that f32 can also have very large results due to very small inputs. And this is partly because we enabled more tests using the new framework.
3. I think what we concluded for f16 also applies to f32 inputs:
    - We trust RMS the most for any kind of inputs.
    - If random float inputs are desired, we need to ignore the element-wise metrics, or use a range that does not include [-1,1] for inputs.

jerryyin:
>  I'll ping you in the future to make sure I get your feedback before merging PRs.

No worries, and I don't mean to block merging. You should go ahead to merge if you are allowed to by github.

> I just picked the threshold that is slightly larger than the max one.

That is a nice trick, no wonder you added the summary section option for more generic use.

> If random float inputs are desired, we need to ignore the element-wise metrics, or use a range that does not include [-1,1] for inputs.

Should we create a ticket for this? I think the default behavior is not this yet, correct me if I'm wrong.

zhanglx13:
The majority ROCK E2E tests do not use float random inputs, and they just set the range as [-1,1]. Only the conv2d_regresesion_fwd/bwd.toml tests require some special settings. I think we don't need to use the same settings for all tests and that is the flexibility we can enjoy from this new framework. But let me know if you have other concerns.

jerryyin:
@zhanglx13 Sounds good. Though I think we should bake the following in as a warning:
 - range include both positive and negative
 - float random inputs

When someone choose to test the above, pop up a warning that relative test result should not be tested. Then redirect the developer to the toml settings you have come up with in the PR.

zhanglx13:
Adding such warnings is trivial. But the question is where should we add it?
- Do we add it in `rocmlir-gen`? But all it does is to generate the code, which does not necessarily get a chance to run. Is it a little weird for `rocmlir-gen` to generate warning messages related to runtime?
- Do we add it in the verification function? Well, in this case, it does not know random type and range.

jerryyin:
I think `rocmlir-gen` is a fair place for it. When a developer is trying to generate a kernel + test host sequence in certain combination, give them a warning at generation time. It is up to the generation harness to warn the developer that this generated test + kernel combination is likely to fail.

You are right that this shouldn't be coupled in verification function, all it does is blind execution of the parameters given to it.

PR #872, opened 2022-11-23
Fix expand shape support in AlignTiling

manupak:
This commit fixes a bug in passing the subset
of the shape to create affine maps to be folded.


jungpark-mlir (approved):
LGTM!

PR #873, opened 2022-11-23
Cleanup AlignTiling for better creation of transforms.

manupak:
This commit contains code that cleans up
AlignTiling pass to use a single place
to add transforms cater both transpose
and broadcasting.

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
@@ -36,6 +36,7 @@
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 
 #include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SmallSet.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/Support/Debug.h"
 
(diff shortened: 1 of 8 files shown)


manupak:
builds on top of https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/872, hence putting to draft.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -260,126 +261,118 @@ struct MILARewritePattern : public OpRewritePattern<linalg::GenericOp> {
...
-    unsigned resIdx = idxAndExpr.index();
-    AffineExpr expr = idxAndExpr.value();
-    if (auto dimExpr = expr.dyn_cast<AffineDimExpr>()) {
-      if (dimExpr.getPosition() != resIdx)
-        return false;
-    } else {
-      return false;
+  for (const auto &idxAndValue : llvm::enumerate(originalMap.getResults())) {
+    auto idx = idxAndValue.index();
+    auto resultExpr = idxAndValue.value();
```
krzysz00: Can we not `auto` this one? I'd actually like to know what the type is here

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -260,126 +261,118 @@ struct MILARewritePattern : public OpRewritePattern<linalg::GenericOp> {
...
+static void createPermutationForMinorIdentityWithBroadcast(
+    const AffineMap &originalMap, SmallVectorImpl<uint32_t> &perm) {
+  for (uint32_t i = 0; i < originalMap.getNumInputs(); ++i) {
+    perm.push_back(i);
+  }
+
+  llvm::SmallSet<uint32_t, 4> foundInputDims;
+  for (const auto &idxAndValue : llvm::enumerate(originalMap.getResults())) {
+    auto idx = idxAndValue.index();
+    auto resultExpr = idxAndValue.value();
```
krzysz00: Let's `Value` this as well?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -538,6 +517,21 @@ static GlobalStoreOp traceToGlobalStore(Value inp) {
   return allValidUses ? result : GlobalStoreOp();
 }
 
+static AffineMap findAliasingOutputIdxMap(linalg::GenericOp laGeneric,
```
krzysz00: Suggestion: look through the uses of `output`, pull the position of the argument in `laGeneric` from the corresponding use, then get the relevant `AffineMap`. Might save us an O(n) find.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -611,44 +659,42 @@ LogicalResult MILARewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
...
+    LLVM_DEBUG(llvm::dbgs()
+               << "Currently, multi-output fusion is not supported.\n");
+    return failure();
+  }
 
   // 1. Trace input to global_store.
   // 1.1. Find the (implicit) gemm output
   GlobalStoreOp gemmStoreOp;
+  Value laGenericInputLeadingToGlobalStore;
+  if (findGlobalStore(laGeneric, laGenericInputLeadingToGlobalStore,
```
krzysz00: Maybe `if(failed(...))` - I kinda like it better, but it could go either way

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -260,126 +261,118 @@ struct MILARewritePattern : public OpRewritePattern<linalg::GenericOp> {
...
-      TransformMapAttr permuteAttr = permuteBuilder.get();
-      Value ret = b.create<TransformOp>(loc, inp, permuteAttr);
-      AffineMap composed = permuteAttr.getMap().getAffineMap().compose(inpMap);
-      LLVM_DEBUG(llvm::dbgs() << "indexing = " << inpMap << " then transform "
-                              << permuteAttr.getMap().getAffineMap() << " is "
-                              << composed << "\n");
-      return {ret, composed};
+// This function will create a permutation that will permute the originalMap to
+// be a MinorIdentityWithBroadcast. This is used to add a permutation later in
+// the chain.
```
krzysz00, line 266: Could this function get a more detailed comment, perhaps with an example?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -260,126 +261,118 @@ struct MILARewritePattern : public OpRewritePattern<linalg::GenericOp> {
...
     }
-    BottomUpTMBuilder transform(b, inpShape, loc);
-    transform.passThrough(ptDims, ptDims);
-    transform.broadcast(bcastDims, bcastSizes);
+    if (isDimAdded) {
+      inp = b.create<TransformOp>(loc, inp, addDimtransform.get());
+    }
 
-    inp = b.create<TransformOp>(loc, inp, transform.get());
+    BottomUpTMBuilder permtransform(
```
krzysz00: Can you move all this into the if() below?

krzysz00 (approved):
Ok, overall, I think I'm happy with this, I'd mostly like to see more comments so future-us don't get confused by what this logic is meant to be doing

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Added an example -- I think it should suffice.

manupak (commented):
Thanks for the review! Ive addressed the comments

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Done

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Done

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: I removed this one as I was building up to support transforming store coordinates in the next PR which might be something we dont want to do for variety of reasons. Since we have the check, we can just use the ouput idx map.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Done

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Done

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Done

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -260,126 +261,123 @@ struct MILARewritePattern : public OpRewritePattern<linalg::GenericOp> {
...
+    SmallVector<uint32_t> perm;
+    createPermutationForMinorIdentityWithBroadcast(inpIdxMap, perm);
+    auto permMap = AffineMap::getPermutationMap(perm, b.getContext());
+    inpIdxMap = inpIdxMap.compose(permMap);
+    assert(
+        (inpIdxMap.isMinorIdentityWithBroadcasting(&bcastDims)) &&
+        "this is guranteed by createPermutationForMinorIdentityWithBroadcast");
+
+    BottomUpTMBuilder bcastTransform(b, inpShape, loc);
+    bool isBcastDone = false;
```
krzysz00: Nit: Maybe `hasBcast`? `isXDone` feels like an early-exit thing to me

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -260,126 +261,123 @@ struct MILARewritePattern : public OpRewritePattern<linalg::GenericOp> {
...
-                              << composed << "\n");
-      return {ret, composed};
+// This function will create a permutation that will permute the originalMap to
+// be a MinorIdentityWithBroadcast. This is used to add a permutation later in
+// the chain.
+// e.g. :
+// (d0, d1, d2, d4) -> (0, d1) was the map
+// This function will return a permutation : [0, 3, 2, 1] s.t.
+// apply it to the original map would result in
+// (d0, d4, d2, d1) -> (0, d1) in effect.
```
krzysz00, line 271: Ok, hold on, I'm slightly confused.

The permutation map is, as I understand it `(d0, d1, d2, d4) -> (d0, d4, d2, d1)`
Then if, after that, I compose in the broadcast, I get the final map `(0, d4)`?

... right, which is a minor identity with broadcast. 

And the reason we need this is because we're actually going to apply the _inverse_ of that permutation to the GEMM coordinates in order to get correct indexing for the extra argument. 

That is, a function later, with the transpose/broadcast handling, we can describe the three sections as follows

/// Broadcast those dimensions that the original linalg.generic map specifies are broadcast and collect their locations, accounting for the leading dimensions not represented in that map but which are present in the gemm coordinates

/// Then, add dimensions that are present in the writeback coordinates but are not present in the additional fusion argument with matching sizes. This, combined with the previous step, ensures that the view of the fusion argument has the same dimensions as the gemm output, though they are not necessarily in the same order.

/// Permute the dimensions of the fusion argument to match those of the gemm writeback by applying the inverse of the permutation that would have made the original indexing map into a minor identity with broadcast. The inverse of that permutation takes the gemm writeback coordinates and scatters them into positions that match the non-identity indexing pattern of the fusion argument.

krzysz00 (approved):
I'm approving again, but I've written some comments I'd like to see in the review comments. Feel free to adjust them until they make sense to you.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Yep! well explained.

So I just copied them 😄

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Done

PR #874, opened 2022-11-24
Adds support for transformed fusion output

manupak:
Currently, the AlignTiling had the assumption
that the operator being fused will have the same
shape as the gemm output buffer. However, this
    can change due to view like ops.

This commit adds support for tranform the
gemm store coordinates to match the shape of
the output of the fused operator should there
be a difference.

* Enabled matmul fusion in resnet50 that
   exercises the above case.

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/740
closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/692

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
@@ -36,6 +36,7 @@
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 
 #include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SmallSet.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/Support/Debug.h"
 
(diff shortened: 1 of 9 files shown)


manupak:
Builds on top of https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/873, hence putting to draft.

manupak:
Changes exlcusive to this PR : https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/874/commits/e6175bbe78cf930a98f3bb8a5136bf0d3b36c53d

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -587,6 +581,116 @@ static Value reconfigureLAGeneric(PatternRewriter &b,
...
+    return failure();
+  }
+  return success();
+}
+
+// This function will add a TransformingFor to translate original coordinates
+// of the gemmStore op to be transformed to match the aliasing shape of the
+// output. This is done by replacing the original gemmStoreOp with a new one
+// wrapped in a TransformingFor op.
+static void adjustStoreCoordsForDestArg(
```
krzysz00, line 645: Note: the OOB check I had there can potentially break in the presence of the Unfold operator (which currently comes up in the transforms for filter tensors and so won't be an issue here but...)

Also, for our consideration, whether adding the ability to add new domains to existing loops is a cleaner way to do some of this.

Will elaborate tomorrow afternoon

krzysz00:
More fundamental realization - transforming the gemm output coordinates could break the vectorization we inferred

manupak:
Closing this in favor of : https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/884

PR #875, opened 2022-11-25
Add a test for tosa matmul lowering

manupak:
This commit adds a test for tosa matmul lowering.
Currently this fails (at least for me) and let see what CI says.

If its not just me, we need to fix this before we can verify the fusion
is right (https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/692)


manupak:
Intention of the PR is not to merge but to check whether tosa matmul lowering is broken.

manupak:
This was just for testing the CI.

PR #876, opened 2022-11-25
Fix MIGraphX to Tosa to support 2D matmul

jungpark-mlir:
Turns out tosa always requires batch dimension.


manupak (approved):
LGTM!

krzysz00 (approved):
Extra LGTMs here

PR #879, opened 2022-11-28
MLIR#742: added e2e test for each fusion kernel produced by the torch…

sjw36:
…-mlir resnet50 model.

* disabled matmul fusion (resnet50_torch_part_53.mlir)

https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/742

### DIFF
--- a/mlir/test/xmir/e2e/resnet50_torch_part_0.mlir
+++ b/mlir/test/xmir/e2e/resnet50_torch_part_0.mlir
@@ -1,0 +1,20 @@
+// RUN: rocmlir-driver -host-pipeline partition,highlevel -targets %arch %s | rocmlir-gen -ph -print-results -rand 1 -rand_type float -fut resnet50_part_0 --verifier clone - | rocmlir-driver -host-pipeline xmodel -kernel-pipeline full | xmir-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_c_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_async_runtime%shlibext --entry-point-result=void | FileCheck %s
+// CHECK: RMS = {{.*}}e-08
+// CHECK: [1 1 0]
+module {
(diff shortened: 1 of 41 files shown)


mlir/test/xmir/e2e/resnet50_torch_part_0.mlir
```diff
@@ -0,0 +1,20 @@
+// RUN: rocmlir-driver -host-pipeline partition,highlevel -targets %arch %s | rocmlir-gen -ph -print-results -rand 1 -rand_type float -fut resnet50_part_0 --verifier clone - | rocmlir-driver -host-pipeline xmodel -kernel-pipeline full | xmir-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_c_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_async_runtime%shlibext --entry-point-result=void | FileCheck %s
```
krzysz00, line 1: Nit: Can we use `xmodel,runner` and `mlir-cpu-runner -O2` here instead? It'll simplify debugging.

sjw36: I prefer to keep the XMIR flow which follows the same steps. Debugging xmir-runner is generally equivalent to rocmlir-driver.

krzysz00: I prefer not to, as xmir-runner will both run a bunch of CPU-side passes we don't need and there's the confusion of having non-trivial compilation in "runner"

krzysz00 (changes requested):
I'll note that this'll run all of resnet50 on every PR and every time someone recompiles. Can you make these nightly-only?

sjw36:
> I'll note that this'll run all of resnet50 on every PR and every time someone recompiles. Can you make these nightly-only?

These are in the e2e directory. Doesn't that preclude them from PR testing.

krzysz00:
@sjw36 Sorry, didn't see your comment. The `mlir/test/xmir/e2e/` directory has the following `lit.local.cfg`

```
if not config.enable_rock_driver_pr_e2e_test or config.no_AMD_GPU or not config.arch_support_mfma:
  config.unsupported = True
```

So those e2e tests are only for PR testing at the moment

PR #881, opened 2022-11-29
Cleanup `test/rocmlir-driver/e2e_for_pr`

zhanglx13:
Tests in `e2e_for_pr` are converted into `test/e2e/PrResnet50.toml`.


krzysz00 (commented):
Some of the tests we had were only present for particular operations, can you add some [[exnclude]] blocks?

zhanglx13:
It's just the config to test the swizzles that has one operation. And I also added some [[exclude]] to only test the settings in the original file.

krzysz00 (approved):

okakarpa:
restest

krzysz00 (commented):
Hold on a moment ... weren't there `i8` and `bf16` tests?

zhanglx13:
There were no i8 tests and bf16.mlir actually contains a single test with f16.
Do you want to add them? It's trivial since we have the PrResnet50.toml.

krzysz00:
Ah, good to know. We can add them some other time.

PR #882, opened 2022-11-29
Remove Unfold transform

giuseros:
In this PR:
- Added an analysis pass to understand which elements of a merge will result in contiguous dimensions in the actual memory space
- Removed the injection of `Unfold` transform (which was injecting this information externally) in different parts of the codebase


giuseros:
Hi @krzysz00 , 
I tried to come up with something. Can we discuss what I am (surely) missing :) ? In particular, I see that `Embed`  doesn't work, but it seems more related to how `Embed` deals with an `Unfold` (i.e., I can correctly detect that the dimensions are contiguous, but somehow `Embed`  is not able to use the information, but again, I might be missing something)

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -345,9 +347,179 @@ propagateUnmergeVectorization(T &&dimAndLength,
   return result;
 }
 
+// Key data structures for vectorization analysis
+using MergePairKey = std::pair<TransformMapAttr, TransformAttr>;
+
+using DimToMergeMap =
+    llvm::DenseMap<int, std::pair<TransformMapAttr, TransformAttr>>;
```
krzysz00: Nit: `uint32_t` for dimension indices, `int64_t` for sizes

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -345,9 +347,179 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+
+using MergeToDimMap = llvm::DenseMap<MergePairKey, SmallVector<uint32_t>>;
+
+using ContiguousMergesMap =
+    llvm::DenseMap<MergePairKey, SmallVector<SmallVector<uint32_t>>>;
+
+/**
```
krzysz00: Nit: `///` comments because that's how upstream is doing it

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -345,9 +347,179 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+        for (auto d : lowerDims) {
+          dimToMerge[d] = std::make_pair(transformMap, transform);
+        }
+        break;
+      case TransformType::AddDim:
+        dummyDimensions.insert(upperDims[0]);
+        break;
+      case TransformType::Embed: {
+        SmallVector<int64_t> unmergeParams;
+        for (size_t i = 0; i < params.size() - 1; i++) {
```
krzysz00: Note: The parameters on `Embed{}` calls are not guaranteed to be sorted.

krzysz00: Second note: you'll need some sort of complexity here to determine where Embed{} acts like Unmerge{}. That is, at our most conservative, we can throw out any Embed{} whose sorted coefficients aren't of the form `1, l_1, l_1 l_2, ...`. If you want to be more clever, we can work out groups of such coefficients ... but maybe that's too much effort for now.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -345,9 +347,179 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+          unmergeParams.push_back(params[i + 1] / params[i]);
+        }
+        unmergeParams.push_back(params.back());
```
krzysz00: Missing a division in any case. But also, very annoyingly, that final coefficient doesn't tell you the length of the whole thing. You'll want to look into the sizes on the transform map, I think.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -345,9 +347,179 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+        scanForContiguousDimensions(
+            filterDummyDimensions(upperDims, dummyDimensions), unmergeParams,
+            dimToMerge, result);
+        break;
+      }
+
+      case TransformType::PassThrough:
```
krzysz00, line 526: These'll need a case to handle moving dimensions around? Or is that handled somewhere else?

krzysz00 (commented):
Some notes on Embed{}

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
giuseros: Yes, you are right. But I don't think the transform map would tell me either. For instance:
```
#transform_unmerge_3 = #rock.transform_map<affine_map<(d0, d1, d2, d3) -> (d3 + 2*d2 + 4*d1 + 128*d0)>
  by [<Embed{1, 2, 4, 128} ["Wo", "Ho", "K", "N"] at [3, 2, 1, 0] -> ["x"] at [0]>]
  bounds = [2, 2, 32, 128] -> [16384]>
```
The map is simply applying the coefficients to the dimensions (I think). Am I wrong? To sort this out, once I make sure that this is a merge-like transformation, I could either:
- add a `-1` in the `unmergeParams` and make the unmerge logic ignore the -1
- use the bounds to correctly detect the last parameter

Please, let me know if I am missing something.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
giuseros: I was originally basing everything on the values of the parameters, but as we discussed, these were not enough. So I added some logic to address dimension reshuffling and adding/deleting dimensions

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
giuseros: Also, independently by the last parameters I found that although I am grouping the dimensions in the right way, the vectorization of `Embed` still fails to vectorize properly. It seems to me that `Unfold+Embed` struggles to vectorize properly.

krzysz00: I'm suggesting using the bounds to work out the last parameter. It's not great, and we probably should move more Embed{} calls to Unmerge{}, but it should do.

giuseros: Perfect, let's try this way. Thanks @krzysz00 !

giuseros:
Ok, I think this is ready for review and testing. Few notes :
- Please, carefully check that I got the transformations right in the tests
- I enabled the analysis pass by not pushing the `Unfold` transformation (and any hack related to it) in the transformation pipeline from `ConvToGemm` . Is that the right thing to do?

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -346,9 +348,249 @@ propagateUnmergeVectorization(T &&dimAndLength,
   return result;
 }
 
+// Key data structures for vectorization analysis
+using MergePairKey = std::pair<TransformMapAttr, TransformAttr>;
+
+using DimToMergeMap =
```
krzysz00: For some of these, there might be an argument for SmallDenseMap (not sure I remember the spelling right)

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -346,9 +348,249 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+    }
+
+    for (size_t j = 0; j < mergeLowerDims.size(); j++) {
+      dimPosition[mergeLowerDims[j]] = j;
+    }
+  }
+
+  // Given a dimension this map tells us what is the parameter
+  // associated to that dimension
+  llvm::DenseMap<uint32_t, int64_t> paramMap;
```
krzysz00: Or for these cases, the VectorMap widget, perhaps, but that's kinda klunky ...

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -346,9 +348,249 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+// in the unmerge
+void scanForContiguousDimensions(
+    ArrayRef<uint32_t> unmergeDimsMaybe, ArrayRef<int64_t> unmergeParams,
+    DimRemapMap &dimRemap, DenseSet<uint32_t> &dimToDelete,
+    DimToMergeMap &dimToMerge, DenseMap<MergePairKey, MergeInfo> &mergeInfoMap,
+    ContiguousMergesMap &result) {
+
+  // Shrink the dimension space if some dimensions are not real
+  // DenseMap<size_t, size_t> realIndices;
+  SmallVector<uint32_t> unmergeDimsMaybeSorted(unmergeDimsMaybe);
```
krzysz00: Rename to `sortedUnmergeDims` since it's clear from context you're about to sort them.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -346,9 +348,249 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+  std::sort(unmergeDimsMaybeSorted.begin(), unmergeDimsMaybeSorted.begin());
+
+  SmallVector<uint32_t> unmergeDims;
```
krzysz00: This all could use some comments, I'm not sure why all this logic is here, but that might be because I'm reading from the top.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -346,9 +348,249 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+  llvm::DenseMap<uint32_t, int64_t> dimPosition;
+};
+
+// Given dimensions and parameters of an unmerge, scan the merges
+// to flag group of dimensions that are known to be contiguous
+// in the unmerge
+void scanForContiguousDimensions(
+    ArrayRef<uint32_t> unmergeDimsMaybe, ArrayRef<int64_t> unmergeParams,
+    DimRemapMap &dimRemap, DenseSet<uint32_t> &dimToDelete,
+    DimToMergeMap &dimToMerge, DenseMap<MergePairKey, MergeInfo> &mergeInfoMap,
```
krzysz00: Can we `const` things that aren't modified?

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -346,9 +348,249 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+    ContiguousMergesMap &result) {
+
+  // Shrink the dimension space if some dimensions are not real
+  // DenseMap<size_t, size_t> realIndices;
+  SmallVector<uint32_t> unmergeDimsMaybeSorted(unmergeDimsMaybe);
+  std::sort(unmergeDimsMaybeSorted.begin(), unmergeDimsMaybeSorted.begin());
+
+  SmallVector<uint32_t> unmergeDims;
+  size_t inc = 0;
+  for (auto dim : unmergeDimsMaybeSorted) {
```
krzysz00: I'm not convinced that sorting the arguments to unmerge is correct - especially since you'll need to permute the parameters to go with it.

The sorts are only needed for Embed{} to get the parameters in the right order.

That is, things could go wrong if you do `(y@0, x@1) <- Unmerge{4, 1}(w)`

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -346,9 +348,249 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+    SmallVector<uint32_t> outputDims;
+
+    // Keep adding dimensions until either:
+    // - they don't belong to the same merge anymore
+    // - their parameters don't match
+    // - they are unmerged in the wrong order
+    // - skip over singleton dimensions
+    for (size_t j = i; j < unmergeDims.size(); j++) {
+      // Protect against broadcasts
+      if (unmergeParams[j] == 1) {
```
krzysz00: Update the comment to `// Unit lengths don't affect the mergeability logic` or similar

But also, those unit dimensions should be included in the contiguous dimension set (`outputDims`, I think)

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -346,9 +348,249 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+        auto &&zip = llvm::zip(params, upperDims);
+        SmallVector<std::tuple<int64_t, uint32_t>> data(zip.begin(), zip.end());
+        std::sort(data.begin(), data.end());
+
+        // Verify that the Embed is an Unmerge operation and
+        // at the same time create the sorted unmerge params
+        bool maybeUnmerge = true;
+        SmallVector<int64_t> unmergeParams;
+        SmallVector<uint32_t> unmergeDims;
+        for (size_t i = 0; i < data.size() - 1; i++) {
```
krzysz00, line 491: Embed{0} goes to `dimsToDelete` and `Embed{parameter < 0}` breaks the isUnmerge

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -346,9 +348,249 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+      case TransformType::Broadcast:
+        // Map the lower dimensions to the dimensions originally
+        // used (i.e., upper dimensions)
+        for (auto pair : llvm::zip(upperDims, lowerDims)) {
+          auto u = std::get<0>(pair);
+          auto l = std::get<1>(pair);
+          dimensionMap[l] = u;
+        }
+        break;
+      case TransformType::Slice:
```
krzysz00: `Slice` goes up with `Pad`

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -346,9 +348,249 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+      case TransformType::Unfold: // TODO: remove Unfold
```
krzysz00: For this PR, make this be the Merge case. But, collect up all the Unfold{} transforms and assert they all got recorded as contiguous groups at the end.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -346,9 +348,249 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+  // associated to that dimension
+  llvm::DenseMap<uint32_t, int64_t> paramMap;
+  // Given a dimension, this map tells us what is its position
+  // in the dimension vector (e.g., if mergeDims = [0,2,3], dimPosition(2)=1)
+  llvm::DenseMap<uint32_t, int64_t> dimPosition;
+};
+
+// Given dimensions and parameters of an unmerge, scan the merges
+// to flag group of dimensions that are known to be contiguous
+// in the unmerge
```
krzysz00: I'm broadly suspicious of how intertwined and stateful this code is. Would it be possible to have a function that takes a single `TransformMapAttr` and some sort of `ContiguousMergesInfo` and then returns a _new_ `ContiguousMergesInfo` that reflects the effects of the transformations?

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -528,16 +770,20 @@ propagateVectorizationInfo(TransformMapAttr map,
...
       int64_t maxLen = input[upperDim]->maxLength;
       int64_t coeff = input[upperDim]->needsCoefficient;
+      int64_t align = input[upperDim]->alignment;
       int64_t stride = 1;
-      for (auto pair :
-           llvm::zip(llvm::reverse(lowerDims), llvm::reverse(params))) {
-        uint32_t lowerDim = std::get<0>(pair);
-        int64_t lowerLen = std::get<1>(pair);
+      size_t dim_pos = lowerDims.size() - 1;
+      for (auto g : llvm::reverse(groups)) {
```
krzysz00: I can now see why there's sorts() in the groups. However, the groups should not be sorted in integer order, they should be sorted by their position in the Merge parameters list.

Your code will do subtly wrong things for `d <- Merge{A, B, C}(x@2, y@1, z@0)` where `y` and `z` are contiguous.

mlir/test/Dialect/Rock/test_vectorization_inference.mlir
```diff
@@ -109,24 +173,24 @@ func.func @test_vectorization() {
...
   %18 = "get_length"() {transforms = [#transform_map7], in_dim = 0 : index, max_len = 32 : index} : () -> (memref<4x5x8xf32>)
   // Unless it was an illusion
-  // CHECK-NEXT: result = 32
+  //  CHECK-NEXT: result = 32
   %19 = "get_length"() {transforms = [#transform_map7, #transform_map8], in_dim = 0 : index, max_len = 32 : index} : () -> (memref<4x8xf32>)
   // or a broadcast
   // CHECK-NEXT: result = 32
   %20 = "get_length"() {transforms = [#transform_map7, #transform_map9], in_dim = 0 : index, max_len = 32 : index} : () -> (memref<4x1x8xf32>)
-
+
```
krzysz00: Nit: you added/removed a bunch of spaces, can you undo that? (The editor setting that automatically strips trailing spaces from lines is usually somewhere and may be a good call)

krzysz00 (commented):
Ok, so, overall comments are that those sort() calls are making me suspicious.
Heck, the structure of that unmerge check makes me suspicious.

I don't see any glaring correctness issues, but I see that I would have a hard time seeing them with how you've structured things. This isn't about you, necessarily, this is about me having a much easier time reasoning about analyses that have the structure of inference rules.

On that note, to pull out the skeleton of a test case I'm worried about

`Merge{4, 1}<["x"] at [0] -> ["a", "b"] at [1, 0]>`
`Unmerge{4, 1}<["a", "b"] at [1, 0] -> ["y"] at [0]>`
maxLength(0) = 8
That should vectorize to 8, I _think_ your code will instead segfault.

krzysz00:
As another note, you can store the groups in a union-find data structure (see LLVM, they have one). You can make those vectors of groups from the union find by looping backwards through the merge dimensions and collecting until the representative value changes.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
giuseros: This is done, but I think the `outputDims` (which I renamed to `groupCandidate`) should not contain the unit length. For instance if I have `Merge{8,3} at [0,2] , Broadcast{1} at[1]` we should ignore the second dimension (I think)

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
giuseros: I added a check for embed dimensions to be all `> 0`

giuseros:
Hi @krzysz00 , thank you so much for your comments! I restructured the code with the acquired knowledge about transformations and how an analysis pass should actually be structured. I think it should now be more readable. Things to check:
- Is the algorithm correct? 
- About the union-find DS you proposed (which I think is `EquivalenceClasses` in llvm), I don't see immediately how it should be used. I mean, the groups are built when I meet the Unmerge/Embed. When I meet those transforms I go check in the dimension table (`dimToMerge`)  if the dimensions belong to the same merge, and same params. If so, this represent a `groupCandidate`. If the position of these dimensions is sorted *in the merge*, then I add the `groupCandidate` to the group.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -346,9 +347,206 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+      i++;
+      continue;
+    }
+
+    SmallVector<uint32_t> groupCandidate;
+    SmallVector<size_t> dimPosition;
+    for (size_t j = i; j < upperDims.size(); j++) {
+
+      // Unit lengths don't affect the mergeability logic
+      if (params[j] == 1) {
```
krzysz00: I'd actually argue for always adding unit dimensions to the current group. That way, every input to the Unmerge{} is inside a group

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -346,9 +347,206 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+
+      auto keyJ = dimToMerge[upperDims[j]];
+      auto mergePairJ = std::make_pair(std::get<0>(keyJ), std::get<1>(keyJ));
+
+      if (!mergePairJ.second) {
+        break;
+      }
+      auto mergePairJParams = mergePairJ.second.getParams();
+      auto mergePairJDims = mergePairJ.second.getLowerDims();
+      size_t posJ = std::get<2>(keyJ);
```
krzysz00: Since we're dealing with these rather dense tuples ... would they make sense as structs instead?

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
krzysz00: But if you're only returning non-singleton groups, it works this way too.

The reason I suggested storing `contiguousGroups` as a `EquivalenceClasses` is because it makes the query logic inside vectorization easier.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -346,9 +347,206 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+      groupCandidate.push_back(mergePairJDims[posJ]);
+      dimPosition.push_back(posJ);
+    }
+
+    i += std::max(size_t(1), groupCandidate.size());
+
+    // Update the result with the current group for the mergePair key
+    if (groupCandidate.size() > 1 &&
+        std::is_sorted(dimPosition.begin(), dimPosition.end())) {
+      contiguousGroups[mergePairI].push_back(groupCandidate);
```
krzysz00: If we had a EquivalenceClasses, you could then start calling `join` on all the `groupCandidate` elements.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -346,9 +347,206 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+// This is an analysis pass to group the lower dimensions of a Merge
+// transformation into contiguous groups. E.g., if we have a Merge{8,8,3} [0] ->
+// [0,2,3] and we know that [2,3] are contiguous in the final representation, we
+// can split the merge group in
+// [[0] [2,3]]. This information will be used by the vectorizer. E.g., the
+// vectorizer can fulfill a vectorization by 4, since 8*3=24 is a multiple of 4.
+// In other words, every group of dimensions is treated as a single group
+ContiguousMergesMap findContiguousGroups(ArrayAttr transforms,
+                                         ArrayRef<int64_t> outputShape) {
+  // Transform table
```
krzysz00: Comment addition: Will be overwritten after processing each transform_map.

(That way we're all clear that this could've been factored out into a separate function but wasn't)

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -346,9 +347,206 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+  ContiguousMergesMap contiguousGroups;
+
+  for (TransformMapAttr transformMap :
+       transforms.getAsRange<TransformMapAttr>()) {
+    auto upperBounds = transformMap.getUpperBounds();
+
+    DimToMergeMap thisDimToMerge;
+
+    for (TransformAttr transform : transformMap.getOps()) {
+      auto transformType = transform.getType();
```
krzysz00: It might be good to use less auto here to improve readability.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -346,9 +347,206 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+        findCountiguousGroupsUnmerge(upperDims, params, dimToMerge,
+                                     contiguousGroups);
+        break;
+      }
+    }
+    dimToMerge = thisDimToMerge;
+  }
+
+  // Last global unmerge
+  SmallVector<uint32_t> sortedDims;
```
krzysz00: Couldn't this just be `[0, 1, 2, (outputShape.size() - 1]`?

krzysz00 (commented):
Ok, I'm much happier with this code, in that I can look at it and go "yeah, I don't see any obvious bugs". We're now in the minor style nitpicking zone.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
giuseros: Unfortunately not (I think), because of the AddDim. If we unmerge{0,1,2,3} but we previously had AddDim [1] it will fail.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
giuseros: So, the point is that the groups contain dimensions of the merges. I am simply signalling that in a Merge{a,b,c,d} the dimensions {c,d} are contiguous. This is why I am not inserting the singleton in the groups, since they belong to the `Unmerge`

giuseros:
Hi @krzysz00 , 
I applied most of your changes. Using the `EquivalenceClasses` I avoided to sort the groups, while I am simply querying for membership when I hit the `Merge` during `propagateVectorizationInfo`. Please, let me know what you think!

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -346,9 +348,219 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+      case TransformType::Unfold:
+      case TransformType::Merge:
+        for (size_t i = 0; i < lowerDims.size(); i++) {
+          thisDimToMerge[lowerDims[i]] = {transformMap, transform, i};
+        }
+        break;
+      case TransformType::AddDim:
+        break;
+      case TransformType::Embed: {
+        // TODO: we don't support stacked embed/embed. So we want to make sure
```
krzysz00: Can I ask what this case is about? I don't see the problem here - what complicates
```
(a, b) -> (a / 3, a % 3, b / 4, b % 4)
(a, b, c, d) -> (b + 3 * a, c + 4 * d)
```
from concluding that `a` and `b` get merged back contiguously?

krzysz00 (commented):
I'd like to understand that `//TODO` before we merge.

This is shaping up really nicely!

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
giuseros: This was only because I didn't add any test for this case. I did just now (and fixed a small issue with the upper bounds).

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -502,7 +498,7 @@ ContiguousMergesMap findContiguousGroups(ArrayAttr transforms,
         auto fasterParam = std::get<0>(data[0]);
         if (maybeUnmerge && fasterParam == 1 && upperBounds.size()) {
           uint32_t slowerDim = std::get<1>(data.back());
-          unmergeParams.push_back(upperBounds[0]);
+          unmergeParams.push_back(upperBounds[upperBoundIdx]);
```
krzysz00: ... Oh. I'll note that `upperBounds` is global. You want to be looking up `upperBounds[embed.getUpperDims()[slowerDim]`, if I understand your code correctly.

Test case:
- `(x) -> (x / 4, x % 4)`
- `(a, b) -> (b, a)` `// (x % 4, x / 4)`
- `<Embed{1, 4} ["a", "b"] at [0, 1] -> ["y"] at [0]>` aka `(a, b) -> (1 * a + 4 * b)`

krzysz00 (commented):
One potential issue here?

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -346,9 +348,209 @@ propagateUnmergeVectorization(T &&dimAndLength,
...
+      i++;
+      continue;
+    }
+
+    SmallVector<uint32_t> groupCandidate;
+    SmallVector<size_t> dimPosition;
+    for (size_t j = i; j < upperDims.size(); j++) {
+
+      // Unit lengths don't affect the mergeability logic
+      if (params[j] == 1) {
```
krzysz00: I'm still not quite happy about this one, unless it's not an issue in practice. Consider
`Merge{3, 1, 8}` , aka `(d0) -> (d0 / 8, 0, d0 % 8)`.

If we call those results `(a, b, c)`, then I'd expect both
`Unmerge{3, 8}(a, c)` and `Unmerge{3, 1, 8}(a, b, c)` to lead to a finding of continuity.

krzysz00: But also if we let `d` be a length-1 dimension, we might also want `Unmerge{3, 1, 1, 8}(a, b, d, c)` to work ... or maybe we don't because that's a silly pathological function no one should've been asking for.

krzysz00 (commented):
Note the second.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
giuseros: I see few issues with `upperBounds[embed.getUpperDims()[slowerDim]` (correct me if I am wrong) : 
- `Slowerdim` is not a position and `getUpperDims` is a vector, not a map. If I have `getUpperDims=[3,4,2,1]` and `slowerDIm` is 4, this will get out of bound
- You might have some dimensions that have been deleted by `AddDim` and then you can easily get out of bounds. 

This is why I created an index that "jumps" of the number of dimensions of the transformation we meet in the map

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
giuseros: I have the suspicious that those cases would work (although pathological they are):
- `Merge{3, 1, 8} at [a,b,c]`  and `Unmerge{3, 1, 8}(a, b, c)` would work for the code
- `Merge{3, 1, 8} at [a,b,c]`  and `Unmerge{3, 8}(a, b, c)` would work, because I don't check for the exact position. I only check that the paramters are the same and that the position are sorted in the same way of the merge. So it should work. 
-  `Merge{3, 1, 8} at [a,b,c]`  and `Unmerge{3, 1,1,8}(a, b, c)` should.... work? I think it is the same logic as before (parameters the same and position sorted in the same way)
Should I add a test and verify?

giuseros: `Unmerge{3, 1, 1, 8}(a, b, d, c)`  and `Unmerge{3,1,8}` would work, right? I am skipping over the singleton so for me that one is equivalent to `Unmerge{3,8}(a,c)` which would be ok. 
 
After thinking a bit about it, I think that also the case `Merge{3,1,8}` -> `Unmerge{3,8}` would work. You would iterate the dimensions of the `Unmerge`. And they would have the same paramters and the same order in the merge. This means the `Merge/Unmerge` pair will be a hit. 

What do you think?

giuseros: Ah, an idea could be that I don't track the dimensions in the merge that are broadcasts. I.e., when I meet `Merge{3,1,8} at [0,1,2]` I won't store `1` in my `dimToMerge` table. What do you think?

krzysz00: I think what you'd need to do is change the logic in the vectorizer to automatically add length-1 dimensions to the previous group as a special case.

krzysz00: (But hey, if it all works, it all works)

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
krzysz00: The point is that
1. The Embed takes dimension indexes (which map exactly to positions in `upperBounds`) as the upper dimensions
2. If you've sorted `zip(params, upperDims())`, then that second thing will be valid an index into `upperBounds`

Note that `upperBounds` is an attribute that exists on each transform_map.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
krzysz00: (Some of your comments disappeared, but I'd add those three examples as tests)

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
giuseros: Yes, now I see what you mean. I amended the code and added the test

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
giuseros: (sorry about deleted comments, I was basically thinking out loud )
Yes, you are right. If the `Merge` has a broadcast dimension we need to add it into the group. I did that, and I also  introduced a `isBroadcast` field in the dim table to track when different transformations introduce the broadcast.

giuseros:
Hi @krzysz00 , 
I applied your suggestion, and while doing it, I also realized that you were right about the last `Unfold`. I could simply feed `0, 1, ... numDims-1` as dimensions, without having to go through the dim table and sorting (in the previous version I couldn't, but now I can).

krzysz00 (approved):
I'm comfortable landing this and convinced that it is correct.

Good catch on the broadcast case!

Approved!

giuseros:
Thanks @krzysz00 , it is been super fun working on this!

krzysz00:
So we have it recorded, perf report from these changes on non-xdlops (mostly noise, some significant gains on wrw). 
[unfold-perf.html.txt](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10482748/unfold-perf.html.txt)

PR #885, opened 2022-12-01
MLIR#741: don't erase copyOp unless we are making the substitution

sjw36:
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/741


manupak (commented):
I see. The change makes sense to me.

Would it be possible to add a test case to ensure this is not regressed ?

krzysz00 (approved):
Seems reasonable, even though I'm not entirely sure what this fix does.

manupak:
Ok, following the conversation yesterday, I don't get how this fixes tosa matmul lowering. If this fixes that, that also means we will have a copy to global memory for just a single gemm operator?

(Deleted the rest, I keep forgetting that it was a XFAIL)

sjw36:
> Ok, following the conversation yesterday, I don't get how this fixes tosa matmul lowering. If this fixes that, that also means we will have a copy to global memory for just a single gemm operator?

The problem was on the host validation side where a memref.copy op was removed (CopyOpt.cpp:90) even though the copy substitution later fails. So the host results were wrong, but the rock.gemm results were correct.

manupak (approved):
Thanks @sjw36 for the explanations. The tests made it much clearer to me!

PR #887, opened 2022-12-01
draft for tuning table.

jungpark-mlir:
Here is a very draft of apis to store the tuning data table, I was thinking of.
I doubt  this work but tried to make it looking sensible. This is more like a imaginary code I hope to work.

Still no fancy idea how to serialize  the problem.

I'm fine with dropping all this code, but might be helpful for us to start the discussion tomorrow.


sjw36:
So we had a discussion today about tuna with Jehandad and he is going to setup a presentation. I am going to postpone the meeting until after that and will invite y'all to it.

PR #888, opened 2022-12-01
Use llvm.assume to encode bounds on block and grid sizes

krzysz00:
Since our rock kernels know their launch dimensions, we can bound the workgroup and workitem IDs from above by those dimensions. LLVM has an assume intrinsic that is, in effect, an assert() that doesn't actually generate code. The manual does warn that you shouldn't use this to annotate thing the optimizer could infer, but the launch dimensions of a kernel are not such a piece of information.

Performance measurements for Navi21 and comparisons of the generated assembly will be added as comments to the PR once I have them.

### DIFF
--- a/mlir/include/mlir/Conversion/RocMLIRPasses.td
+++ b/mlir/include/mlir/Conversion/RocMLIRPasses.td
@@ -32,7 +32,9 @@ def ConvertRockToGPUPass : Pass<"convert-rock-to-gpu", "ModuleOp"> {
     "amdgpu::AMDGPUDialect",
     "cf::ControlFlowDialect",
     "func::FuncDialect",
-    "gpu::GPUDialect"
+    "gpu::GPUDialect",
+    "arith::ArithmeticDialect",
+    "LLVM::LLVMDialect",
   ];
 }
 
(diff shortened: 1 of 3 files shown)


krzysz00:
Ok, so, on Navi21 (aka, the non-xdlops path), we see some (usually at most 1-2%) improvements to certain NHWC cases but some (at most 1-2% in the other direction) regressions in the NCHW case. This may be related to the loop unrolling stuff Yi's been poking at. We do see substantial improvements on certain kernels that we weren't doing well on before. Am waiting for MI-100 numbers to see if it's sensible to move forward here.

[assume-perf-gfx1030.html.txt](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10136114/assume-perf-gfx1030.html.txt)

krzysz00:
LDS writes de-vectorized over in non-xdlops with this change over in the non-xdlops case, I want to know why

krzysz00:
Discussion with backend folks reveals that putting `!range` data on the call sites for the intrinsics is a better idea. This'll be that instead.

krzysz00:
Supersceded and a bad idea

PR #890, opened 2022-12-03
Add a script to facilitate adding and searching configs from the ROCK E2E database

zhanglx13:
This PR adds a script, i.e. `rock_conv.sh`, to keep track of configs in the ROCK E2E framework. It can search configs, provided as *.mlir, *.toml, or command line options, from one or more *.toml files, which is considered as a conv2d config database. The script can also insert the config into the toml file if the toml file does not contain the config.

Use case: in the future when we want to add more configs into the ROCK E2E framework, this script can help us find out whether the configs are already included.


mlir/utils/widgets/rockE2E.sh
```diff
@@ -0,0 +1,432 @@
+#!/bin/bash
+
+declare -A configArr
+declare -a orders
+declare -A configVarToCL
+declare -A configCLToVar
+## Define conv parameters in order
+orders+=(operation)
```
krzysz00: You can just put these into one big `{operation filter)` type thing

krzysz00 (commented):
... didn't submit comment somehow

mlir/utils/widgets/rockE2E.sh
zhanglx13: Sorry I did not get what you mean. Can you elaborate?

krzysz00: Instead of using `+=`, you can write `orders=(operation filter)` (except listing all those options)

mlir/utils/widgets/rockE2E.sh
```diff
@@ -0,0 +1,433 @@
...
+    fi
+}
+
+
+PROG=${0##*/}
+PROG_DIR=$(cd ${0%/*} && pwd -P)
+ROCMLIR_DIR=$(cd ${PROG_DIR}; git rev-parse --show-toplevel || echo "${HOME}/rocMLIR/")
+CONVDB_DIR=${ROCMLIR_DIR}/mlir/test/e2e
+
+convDB="${CONVDB_DIR}/MixedConvLayouts.toml"
```
krzysz00: Nit: Why not just `for i in "${CONVDB_DIR}/*.toml"` and then, for instance, dropping the leading comma. This makes the script less error-prone in the future.

krzysz00 (commented):
I'm willing to approve this, since it's a script you found useful and you've presumably tested it.

That being said:
1. Can we get some usage examples at the top?
2. I'm poking removing the hardcoded list of DB names
3. Since it's not a bad call, can you hit this with `shellcheck`?

(But yeah, I've got no fundamental objections to useful utilities hanging around the edges of the repository)

mlir/utils/widgets/rockE2E.sh
zhanglx13: There are conv and gemm configurations in the directory and we may have other formats in the future. If we want to make it error-prune by using a loop, we need to have some rules about naming. Let me know what do you think.

krzysz00: Ok, wait, hold on, is this meant to be only for convolutions?

zhanglx13: For now, yes. 
It should easy to extended for Gemm. Or a similar script for Gemm.

krzysz00: I'll admit I don't entirely understand the point of this script

zhanglx13: When we want to expand the E2E tests, this script can help check if a given test is already in one of the toml configuration files.

krzysz00: Ah.

In which case, would it be possible to rig up said script to handle gemms as well so we don't have to maintain a second list of toml files people won't know to update?

krzysz00: (But that can be a followup PR)

krzysz00 (approved):
Seems reasonable, pending adding GEMM support in a future PR

zhanglx13:
For Gemm support, the parameters for a gemm config seems to be only the gemmsize, i.e. g,m,k,n. Do we care other parameters when deciding whether two tests are the "same"?

krzysz00:
We also have `transA` and `transB`, which may or may not be set on a test (along with the data type)

mlir/utils/widgets/rock_conv.sh
```diff
@@ -0,0 +1,435 @@
...
+usage() {
+    cat <<END_USAGE
+Usage: $PROG [options...]
+
+Options:
+    -h|--help                  Display this help message
+    --input=                   Input test file.
+                               Note: input files can be either mlir test files
+                               (.mlir) or toml configuration files
+    --database=                Config database file(s). If in search mode, multiple
```
jerryyin: Why not rename database to toml? Looks like this is pointing at only the `.toml` files.

mlir/utils/widgets/rock_conv.sh
```diff
@@ -0,0 +1,435 @@
...
+                               $PROG --input=myTest --database=db1,db2 --search
+                               If in insert mode, only one database file is allowed.
+                               Note:
+                               1. database files must be .toml files.
+                               2. if database file is located at /mlir/test/e2e,
+                                  then only the filename without extension is needed.
+                                  Otherwiese, full file path is required.
+                               3. If no database files are provided, all toml files
+                                  at /mlir/test/e2e will be used.
+    --config                   A single config (pure + accessories) used to be
```
jerryyin: What does `pure + accessories` mean?

mlir/utils/widgets/rock_conv.sh
```diff
@@ -0,0 +1,435 @@
...
+                               searched in the database file(s).
+    -s|--search                Search mode, i.e. search the given configs, either
```
jerryyin, line 107: So clarifications: 
 - confgs == rocMlir driver configs?
 - When searching from a input file that has many configs what would happen?

jerryyin (commented):
Looks like a nice tool to study about what configs are already being tested to avoid future duplicates. I'd abstain from reviewing the internals of the bash logic but reviewed mostly about the usage notes. 

On a side note, I think you can do a demo about the widgets script to the team about how they are useful. Those can be impactful daily scripts.

mlir/utils/widgets/rock_conv.sh
zhanglx13: This script is able to process both *.toml and *.mlir test files. But it seems that we do not want to consider *.mlir files as database. However, I feel like those *.toml files serve as a "database" and maybe in the future we will have other formats. What do you think?

mlir/utils/widgets/rock_conv.sh
zhanglx13: I've been always confused about what is a config. The command line inputs to `rocmlir-gen` to generate a conv kernel includes parameters that
1. describe the geometry of the problem, like c, h, w, b, g of input, filter, and output
2. direction of the conv, like fwd, bwd, and wrw
3. computer related parameters, like data types, random inputs, and arch

Here I mean the parameters in the first category (geometry parameters) define a pure config. And other parameters are accessories to a pure config. This is used to determine if two configs are the same, i.e. we only compare the pure part of the config.
I was also thinking the parameters in the first two categories (geometry + direction) define a config mathematically, i.e. everything we need to compute a conv problem by hand. But I do not find this definition useful.

mlir/utils/widgets/rock_conv.sh
zhanglx13: 1. yes
2. the script will search every config in the input file from the database and report whether it is found. This is like a batch mode.

mlir/utils/widgets/rock_conv.sh
jerryyin: I think it is a fair consideration but a precise naming will help with keeping everything up-to-date. If we come up with a new format and this has the toml option then we'd be able to easily see there needs to be a newer option. On the other hand, concept of database is vague in our end and can be confused with the one tuning can potentially use. Just my two cents here.

zhanglx13:
@jerryyin Re demo. I am thinking of writing some README like docs under the widget directory
And I am working on a summary of the E2E framework, which includes some example usage of this script.

mlir/utils/widgets/rock_conv.sh
jerryyin: Nice categorization, this is the first time I heard about the 3 categories and it make sense. But description here, `pure` + `accessories` seems to be misleading even after I understand about your definitions: It seems to imply `pure` or `geometry` side of configs must be complete, whereas you can pick and combine arbitrary number of `accessories` on top of `pure`. Is that the case? If so, you'll need to leave the definition inline here.

In reality my understanding is that your script will do the search regardless? So even if a user search for a config that has only the direction field (fwd), then the script will return all the `fwd` configs? If the script does not differentiate between the two concept then we should not put those here because it will be confusing to a user of the script.

mlir/utils/widgets/rock_conv.sh
jerryyin: Sounds good.

In other words: 
 - The source of search is either a `--config` or a `--input` (a `mlir` test, or a `toml` file). 
 - The target of search is either `-s` (a `mlir` test) or `--database` (a `toml` file).

For each option, could you clarify whether it is the source or target in its first description?

mlir/utils/widgets/rock_conv.sh
zhanglx13: It is indeed misleading. Thank you for pointing that out.
The configs in the user input (command line or a file) can contain any parameters. Let's call them user-configs. The script processes the parameters in the user-config in the following three 3 scenarios:
1. Respect the parameter setting. This is the case for parameters in the geometry category
2. Ignore the parameter setting. This is the case for parameters in the direction and computer related categories
3. For parameters in the geometry category that are not set in the user-config, use the default value defined in `rocmlir-gen.cpp`.

In this way, each user-config can be converted into a "canonicalized " format.
The same process is done for configs in the target *.toml files. As a result, the search operation is done as a string match.

I feel like I need to include the above in the description. But it might be too long and verbose. I'll try to add a short summary in the description.

mlir/utils/widgets/rock_conv.sh
zhanglx13: Actually we do not have *.mlir files as target of the search (the script can do that but users are not likely need to do that). I'll clarify the usage per our discussion.

PR #892, opened 2022-12-06
[DO NOT SQUASH] Set required work group sizes for GPUs to enable better optimizations

krzysz00:
**General summary:** Performance is slightly up (1-4%) on gfx1030 with almost no regressions, as shown in 
[assume-perf-v2-gfx1030.html.txt](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10159254/assume-perf-v2-gfx1030.html.txt) and gfx908 performance is basically flat per 
[assume-perf-v2-gfx908.html.txt](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10160274/assume-perf-v2-gfx908.html.txt)

Unlike #888, which will be closed in favor of this PR, this doesn't break the LDS write vectorization pass. Assembly-level changes on gfx1030 are limited, mainly consisting of more use of `mul_low_u16` instructions and other peephole changes that are correct under the correct assumption that no carry or overflow is possible. It's possible that there may be some untapped potential in hinting the grid size, but that's a broader question.

**The individual commits**

[[mlir][GPU] Add known_block_size and known_grid_size to gpu.func

In many cases, the the number of workgroups (the grid size) and the
number of workitems within each group (the block size) that a GPU
kernel will be launched with are known. For example, if gpu.launch is
called with constant block and grid sizes, we know that those are the
only possible sizes that will be used to launch that kernel. In other
cases, a custom code-generation pipeline that eventually produces GPU
kernels may know the launch dimensions of those kernels, or at least
may be able to provide an upper bound on them.

Other GPU programming systems, such as OpenCL, allow capturing such
information to enable compiler optimizations - see
reqd_work_group_size, but MLIR currently has no mechanism for doing so.

This set of attributes is the first step in enabling optimizations
based on the known launch dimensions of kernels. It extends the kernel
outline pass to set these bounds on kernels with constant launch
dimensions and extends integer range inference for GPU index
operations to account for the bounds when they are known.

Subsequent revisions will use this data when lowering GPU operations
to the ROCDL dialect.

@[krzysz00](https://github.com/ROCmSoftwarePlatform/rocMLIR/commits?author=krzysz00)
krzysz00 committed Dec 5, 2022

[mlir][ROCDL] Translate known block size attributes to ROCDL

1. When converting from the GPU dialect to the ROCDL dialect, if the
function that contains a gpu.thread_id or gpu.block_id op is annotated
with known_block_size, use that known_block_size to set a "range"
attribute on the corresponding rocdl intrinsic so that the LLVM
frontend can optimize based on that range information.
1b. When translating from the rocdl dialect to LLVM IR, use the
"range" attribute, if present, to set !range metadata on the relevant
function call.
2. When translating from gpu.func to llvm.func within gpu-to-rocdl,
copy the known_block_size attribute as rocdl.reqd_work_group_size to
enable further translations to set the corresponding metadata on the
LLVM IR function.
2b. Extend the ROCDL to LLVM IR translation to set the
!reqd_work_group_size metadata on LLVM functions when able and to also
use that information to set amdgpu-flat-work-group-size to the product
of those known sizes, thus preventing inconsistencies between the two
sets of metadata.

Also update tests and add functions to the ROCDL dialect to ensure
attribute names are used consistently.

Set gpu.known_block_size and gpu.known_grid_size on gpu.func ops

Not sure if anything in xmir needs to be changed here.

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
@@ -30,6 +30,12 @@ def ROCDL_Dialect : Dialect {
     /// Get the name of the attribute used to annotate external kernel
     /// functions.
     static StringRef getKernelFuncAttrName() { return "rocdl.kernel"; }
+    static constexpr ::llvm::StringLiteral getFlatWorkGroupSizeAttrName() {
+      return ::llvm::StringLiteral("rocdl.flat_work_group_size");
+    }
+    static constexpr ::llvm::StringLiteral getReqdWorkGroupSizeAttrName() {
+      return ::llvm::StringLiteral("rocdl.reqd_work_group_size");
+    }
   }];
 }
 
@@ -55,8 +61,9 @@ class ROCDL_SpecialRegisterOp<string mnemonic,
     list<Trait> traits = []> :
   ROCDL_Op<mnemonic, !listconcat(traits, [NoSideEffect])>,
   Results<(outs LLVM_Type:$res)>, Arguments<(ins)> {
-  string llvmBuilder = "$res = createIntrinsicCall(builder,"
-    # "llvm::Intrinsic::amdgcn_" # !subst(".","_", mnemonic) # ");";
+  string llvmBuilder = "$res = createIntrinsicCallWithRange(builder,"
+    # "llvm::Intrinsic::amdgcn_" # !subst(".","_", mnemonic)
+    # ", op->getAttrOfType<::mlir::DenseI32ArrayAttr>(\"range\"));";
   let assemblyFormat = "attr-dict `:` type($res)";
 }
 
--- a/external/llvm-project/mlir/lib/Conversion/GPUCommon/IndexIntrinsicsOpLowering.h
+++ b/external/llvm-project/mlir/lib/Conversion/GPUCommon/IndexIntrinsicsOpLowering.h
@@ -11,6 +11,7 @@
 #include "mlir/Conversion/LLVMCommon/Pattern.h"
 #include "mlir/Dialect/GPU/IR/GPUDialect.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "mlir/IR/BuiltinAttributes.h"
 
 namespace mlir {
 
@@ -23,11 +24,19 @@ template <typename Op, typename XOp, typename YOp, typename ZOp>
 struct GPUIndexIntrinsicOpLowering : public ConvertOpToLLVMPattern<Op> {
 private:
   unsigned indexBitwidth;
+  StringRef boundsAttrName;
 
 public:
   explicit GPUIndexIntrinsicOpLowering(LLVMTypeConverter &typeConverter)
       : ConvertOpToLLVMPattern<Op>(typeConverter),
-        indexBitwidth(typeConverter.getIndexTypeBitwidth()) {}
+        indexBitwidth(typeConverter.getIndexTypeBitwidth()),
+        boundsAttrName("") {}
+
+  explicit GPUIndexIntrinsicOpLowering(LLVMTypeConverter &typeConverter,
+                                       StringRef boundsAttrName)
+      : ConvertOpToLLVMPattern<Op>(typeConverter),
+        indexBitwidth(typeConverter.getIndexTypeBitwidth()),
+        boundsAttrName(boundsAttrName) {}
 
   // Convert the kernel arguments to an LLVM type, preserve the rest.
   LogicalResult
@@ -35,7 +44,7 @@ public:
                   ConversionPatternRewriter &rewriter) const override {
     auto loc = op->getLoc();
     MLIRContext *context = rewriter.getContext();
-    Value newOp;
+    Operation *newOp;
     switch (op.dimension()) {
     case gpu::Dimension::x:
       newOp = rewriter.create<XOp>(loc, IntegerType::get(context, 32));
@@ -48,15 +57,28 @@ public:
       break;
     }
 
-    if (indexBitwidth > 32) {
-      newOp = rewriter.create<LLVM::SExtOp>(
-          loc, IntegerType::get(context, indexBitwidth), newOp);
-    } else if (indexBitwidth < 32) {
-      newOp = rewriter.create<LLVM::TruncOp>(
-          loc, IntegerType::get(context, indexBitwidth), newOp);
+    Operation *function;
+    if (auto gpuFunc = op->template getParentOfType<gpu::GPUFuncOp>())
+      function = gpuFunc;
+    if (auto llvmFunc = op->template getParentOfType<LLVM::LLVMFuncOp>())
+      function = llvmFunc;
+    if (!boundsAttrName.empty() && function) {
+      if (auto attr = function->template getAttrOfType<DenseI32ArrayAttr>(
+              boundsAttrName)) {
+        int32_t maximum = attr[static_cast<uint32_t>(op.dimension())];
+        newOp->setAttr("range", rewriter.getDenseI32ArrayAttr({0, maximum}));
+      }
     }
 
-    rewriter.replaceOp(op, {newOp});
+    if (indexBitwidth > 32) {
+      newOp = rewriter.create<LLVM::SExtOp>(
+          loc, IntegerType::get(context, indexBitwidth), newOp->getResult(0));
+    } else if (indexBitwidth < 32) {
+      newOp = rewriter.create<LLVM::TruncOp>(
+          loc, IntegerType::get(context, indexBitwidth), newOp->getResult(0));
+    }
+
+    rewriter.replaceOp(op, newOp->getResults());
     return success();
   }
 };
--- a/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
+++ b/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
@@ -26,9 +26,11 @@
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/GPU/IR/GPUDialect.h"
 #include "mlir/Dialect/GPU/Transforms/Passes.h"
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
 #include "mlir/Dialect/LLVMIR/ROCDLDialect.h"
 #include "mlir/Dialect/Math/IR/Math.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
+#include "mlir/IR/BuiltinAttributes.h"
 #include "mlir/Pass/Pass.h"
 #include "mlir/Pass/PassManager.h"
 #include "mlir/Transforms/DialectConversion.h"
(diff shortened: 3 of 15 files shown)


krzysz00:
Update: there's some devectorization of LDS reads on some xdlops configs, I'll be looking into how to reproduce it in case there's perf impact / a compiler bug tomorrow

krzysz00:
Ok, so, the minor perf dips on some MFMA configurations come down to https://github.com/llvm/llvm-project/issues/59379 . Overall, these changes increase non-MFMA performance a bit and have negligible effect on MFMA performance ... and it's possible the linked issue will conveniently disappear if we refactor the blockwise MFMA gemm. So ... thoughts?

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
```diff
@@ -30,6 +30,12 @@ def ROCDL_Dialect : Dialect {
     /// Get the name of the attribute used to annotate external kernel
     /// functions.
     static StringRef getKernelFuncAttrName() { return "rocdl.kernel"; }
+    static constexpr ::llvm::StringLiteral getMaxFlatWorkGroupSizeAttrName() {
+      return ::llvm::StringLiteral("rocdl.max_flat_work_group_size");
+    }
```
jerryyin: I'm not entirely sure what this field is for

external/llvm-project/mlir/lib/Conversion/GPUCommon/IndexIntrinsicsOpLowering.h
```diff
@@ -48,15 +57,28 @@ struct GPUIndexIntrinsicOpLowering : public ConvertOpToLLVMPattern<Op> {
       break;
     }
 
+    Operation *function;
+    if (auto gpuFunc = op->template getParentOfType<gpu::GPUFuncOp>())
+      function = gpuFunc;
+    if (auto llvmFunc = op->template getParentOfType<LLVM::LLVMFuncOp>())
+      function = llvmFunc;
```
jerryyin, line 64: Is this necessary? I thought that the conversion source is guaranteed to be from gpu dialect therefore llvm side of things should exists yet.

giuseros:
I am trying to review this PR. Just to be sure: #888 will be closed? And this PR encompasses #888 + other changes?

krzysz00:
Yeah, #888 is now closed. The compiler folks told me this is a strictly better way to do the thing.

external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
```diff
@@ -148,6 +150,15 @@ struct LowerGpuOpsToROCDLOpsPass
     if (failed(applyPartialConversion(m, target, std::move(llvmPatterns))))
       signalPassFailure();
 
+    // Manually rewrite known block size attributes so the LLVMIR translation
+    // infrastructure can pick them up.
+    m.walk([](LLVM::LLVMFuncOp op) {
+      if (Attribute blockSizes =
+              op->removeAttr(gpu::GPUFuncOp::getKnownBlockSizeAttrName()))
+        op->setAttr(ROCDL::ROCDLDialect::getReqdWorkGroupSizeAttrName(),
```
jerryyin: So `KnownBlockSize` and `ReqdWorkGroupSize` are equivalent attributes but used across two abstraction layers? Why bother with the rename?

external/llvm-project/mlir/lib/Dialect/GPU/IR/InferIntRangeInterfaceImpls.cpp
```diff
@@ -23,40 +25,106 @@ static ConstantIntRanges getIndexRange(uint64_t umin, uint64_t umax) {
...
+  case Dimension::x:
+    return dims.x;
+  case Dimension::y:
+    return dims.y;
+  case Dimension::z:
+    return dims.z;
+  }
+}
+
+static uint64_t zext(uint32_t arg) { return static_cast<uint64_t>(arg); }
```
jerryyin, line 49: Why is it necessary to extend the type to uint64_t?

external/llvm-project/mlir/lib/Dialect/GPU/IR/GPUDialect.cpp
```diff
@@ -1046,6 +1047,27 @@ LogicalResult GPUFuncOp::verifyBody() {
...
+                                               StringRef attrName) {
+  auto maybeAttr = op->getAttr(attrName);
+  if (!maybeAttr)
+    return success();
+  auto array = maybeAttr.dyn_cast<DenseI32ArrayAttr>();
+  if (!array)
+    return op.emitOpError(attrName + " must be a dense i32 array");
+  if (array.size() != 3)
+    return op.emitOpError(attrName + " must contain exactly 3 elements");
+  return success();
```
jerryyin, line 1060: Add a check to make sure all sizes are positive?

external/llvm-project/mlir/lib/Target/LLVMIR/Dialect/ROCDL/ROCDLToLLVMIRTranslation.cpp
```diff
@@ -97,8 +116,36 @@ class ROCDLDialectLLVMIRTranslationInterface
...
+        llvm::Constant *constant = llvm::ConstantInt::get(i32, i);
+        metadata.push_back(llvm::ConstantAsMetadata::get(constant));
+      }
+      llvm::Function *llvmFunc =
+          moduleTranslation.lookupFunction(func.getName());
+      llvm::SmallString<8> llvmAttrValue;
+      llvm::raw_svector_ostream attrValueStream(llvmAttrValue);
+      attrValueStream << flatSize << "," << flatSize;
       llvmFunc->addFnAttr("amdgpu-flat-work-group-size", llvmAttrValue);
+
```
jerryyin: Typo? `amdgpu-flat-work-group-size` -> `amdgpu-max-flat-work-group-size`?

external/llvm-project/mlir/lib/Target/LLVMIR/Dialect/ROCDL/ROCDLToLLVMIRTranslation.cpp
```diff
@@ -97,8 +116,36 @@ class ROCDLDialectLLVMIRTranslationInterface
           moduleTranslation.lookupFunction(func.getName());
       llvm::SmallString<8> llvmAttrValue;
       llvm::raw_svector_ostream attrValueStream(llvmAttrValue);
-      attrValueStream << "1, " << value.getInt();
+      attrValueStream << "1," << value.getInt();
+      llvmFunc->addFnAttr("amdgpu-flat-work-group-size", llvmAttrValue);
+    }
+    // Set max-flat-work-group-size from known block sizes if set
```
jerryyin: [Question] Would you mind to explain a bit more of where you find this attribute, where it is defined and why it is necessary to set it now?

external/llvm-project/mlir/test/Dialect/GPU/int-range-interface.mlir
```diff
@@ -100,9 +106,9 @@ module attributes {gpu.container_module} {
       %global_id_y = gpu.global_id y
       %global_id_z = gpu.global_id z
 
-      // CHECK: test.reflect_bounds {smax = 9223372036854775807 : index, smin = 0 : index, umax = 9223372036854775807 : index, umin = 0 : index}
-      // CHECK: test.reflect_bounds {smax = 9223372036854775807 : index, smin = 0 : index, umax = 9223372036854775807 : index, umin = 0 : index}
-      // CHECK: test.reflect_bounds {smax = 9223372036854775807 : index, smin = 0 : index, umax = 9223372036854775807 : index, umin = 0 : index}
+      // CHECK: test.reflect_bounds {smax = 9223372036854775807 : index, smin = -9223372036854775808 : index, umax = -17179869180 : index, umin = 0 : index}
+      // CHECK: test.reflect_bounds {smax = 9223372036854775807 : index, smin = -9223372036854775808 : index, umax = -17179869180 : index, umin = 0 : index}
+      // CHECK: test.reflect_bounds {smax = 9223372036854775807 : index, smin = -9223372036854775808 : index, umax = -17179869180 : index, umin = 0 : index}
```
jerryyin, line 111: What is the deal with the negative number update for smin? I don't quite see the implementation you updated lead to this?

external/llvm-project/mlir/test/Dialect/GPU/int-range-interface.mlir
```diff
@@ -126,3 +132,86 @@ module attributes {gpu.container_module} {
...
+      %global_id_y0 = test.reflect_bounds %global_id_y
+      %global_id_z0 = test.reflect_bounds %global_id_z
+
+      %subgroup_size = gpu.subgroup_size : index
+      %lane_id = gpu.lane_id
+      %num_subgroups = gpu.num_subgroups : index
+      %subgroup_id = gpu.subgroup_id : index
+
+      // CHECK: test.reflect_bounds {smax = 128 : index, smin = 1 : index, umax = 128 : index, umin = 1 : index}
+      // CHECK: test.reflect_bounds {smax = 127 : index, smin = 0 : index, umax = 127 : index, umin = 0 : index}
```
jerryyin, line 205: [Question] Is it machine dependent? I actually don't know what is defined as a subgroup size.

jerryyin (commented):
I haven't touched any code in this domain for a long time and got equal amount of questions as review comments.

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
krzysz00: The TensorFlow people needed it to let them set `amdgpu-flat-work-group-size="1,1024` because of some compiler issue they had with 256

external/llvm-project/mlir/lib/Conversion/GPUCommon/IndexIntrinsicsOpLowering.h
krzysz00: I don't think there are hard guarantees we can rely on about the order that conversion patterns run in.

external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
krzysz00: The mlir-to-llvmir translator won't let the ROCDL translator use an attribute to modify a function unless it starts with `rocdl.`, so we have to do some sort of rename. This also makes it clearer what the attribute corresponds to in the LLVM IR, which uses `!reqd_work_group_size` for this annotation.

external/llvm-project/mlir/lib/Dialect/GPU/IR/GPUDialect.cpp
krzysz00: There're apparently people who'll allow `uint32_max` workgroups, at least going off the int range inference for GPU, so I don't think a positivity check is in order.

external/llvm-project/mlir/lib/Dialect/GPU/IR/InferIntRangeInterfaceImpls.cpp
krzysz00: We're dealing with `index` values, and I don't want any silent truncations.

external/llvm-project/mlir/lib/Target/LLVMIR/Dialect/ROCDL/ROCDLToLLVMIRTranslation.cpp
krzysz00: No, it's called flat-work-group-size, and it takes a range.

external/llvm-project/mlir/lib/Target/LLVMIR/Dialect/ROCDL/ROCDLToLLVMIRTranslation.cpp
krzysz00: ... that is a typo in the comment, though.

This attribute is something used an the AMDGPU backend. In at least one codepath, it's a fallbacak from when the `!reqd_work_group_size` isn't set, but I want to ensure we have consistency.

external/llvm-project/mlir/test/Dialect/GPU/int-range-interface.mlir
krzysz00: Changes to the implementation of the global ID, since it's bounded above by max(thread_id) * max(block_id), which is a more accurate bound than what I had previously (uint64_max)

external/llvm-project/mlir/test/Dialect/GPU/int-range-interface.mlir
krzysz00: We could find no machine with subgroups of size over 128 when GPU dialect range inference landed, this is just a copy of that. Good question!

giuseros (approved):
From my narrow understanding of this, it LGTM !

PR #893, opened 2022-12-06
Add warning when float random inputs have both positive and negative numbers

zhanglx13:
This PR prints a warning message when float random inputs within [-1, 1] are used for E2E tests. It suggests to use [1,3] as the range for float random inputs.

### DIFF
--- a/mlir/tools/rocmlir-gen/rocmlir-gen.cpp
+++ b/mlir/tools/rocmlir-gen/rocmlir-gen.cpp
@@ -1943,9 +1943,24 @@ static void emitPrintTensor(OpBuilder &b, Value var) {
   }
 }
 
+static void checkRandomInputsE2E() {
+  if (randomSeed != "none" && randomSeed != "fixed" &&
+      randomDataType == "float") {
+    int min = randMin.getValue();
+    int max = randMax.getValue();
+    if ((-1 <= min && min < 1) || (-1 < max && max <= 1)) {
+      llvm::outs() << "WARNING: E2E tests with float random inputs within ";
+      llvm::outs() << "[-1, 1] may fail\n";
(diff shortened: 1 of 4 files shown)


mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -1943,9 +1943,24 @@ static void emitPrintTensor(OpBuilder &b, Value var) {
   }
 }
 
+static void checkRandomInputsE2E() {
+  if (randomSeed != "none" && randomSeed != "fixed" &&
+      randomDataType == "float") {
+    int min = randMin.getValue();
+    int max = randMax.getValue();
+    if ((-1 <= min && min < 1) || (-1 < max && max <= 1)) {
```
jerryyin: How about `(min < 0 && max > 0)`? I think in general as long as there is a chance to cancel out each other, there's going to be denorm behavior.

krzysz00 (commented):
Seems like a reasonable fix to me. But we might want to change `rand_min` and `rand_max`'s defaults in the float case if we can?

zhanglx13:
So there are two groups of tests that require float random inputs:
- conv2d_regression_fwd/bwd.toml
- `fusion/` and `xmir/e2e/`
Using a range of [0,1] makes everyone happy after some changes to the threshold. So I'll make [0,1] as the default for float random inputs if it looks good to you, @krzysz00 @jerryyin.

@sjw36 Is it ok to use a random range [0,1] as inputs for tests like `resnet18_nlk3.mlir`. It can pass after changing the RMS threshold from `{{.*}e-08` to `{{.*}e-07`.

jerryyin (approved):

PR #894, opened 2022-12-07
Adding a Composable Kernel driver

giuseros:
This is adding a CK driver to see how it compares with MLIR on different archs. 
- Added the driver `cpp` file
- Added the possibility to invoke the driver from `perfRunner.py`
- Changed the `Jenkins` files

One caveat: `batched_gemm` is not supported on `gfx1030`, so we cannot rely on some of the comparisons

Putting as a draft/skip-ci for now, because I would like some feedback first


giuseros:
This now depends on #903

giuseros:
I am closing this PR so that I can start a fresh one

PR #895, opened 2022-12-07
[CI] Add MIGraphX Dockerfile

manupak:
This commit adds the dependencies of MIGraphX to
dockerfile.


krzysz00 (commented):
Is there a way to, say, get this off of MIGraphX instead of hard-coding the versions here?

manupak:
I got most of it from requirements file (where .txt files are used) but there are few that is lying around few shell script where I suppose we dont want to run the full script. So the handpicked ones are those.

mlir/utils/jenkins/Dockerfile
```diff
@@ -125,3 +125,16 @@ RUN wget https://raw.githubusercontent.com/ROCmSoftwarePlatform/MIOpen/develop/r
...
 RUN echo $(git ls-remote https://github.com/ROCmSoftwarePlatform/MIOpen HEAD) > miopen-deps-commit-hash
 WORKDIR /
+# --------------------- Section 4: MIGraphX dependencies ---------------
+WORKDIR /MIGraphXDeps
+RUN pip3 install setuptools wheel
+# Install MIGraphX python dependencies
+RUN pip3 install https://github.com/RadeonOpenCompute/rbuild/archive/master.tar.gz
+RUN pip3 install onnx==1.8.1 numpy==1.21.6 typing==3.7.4 pytest==6.0.1 packaging==16.8
+RUN pip3 install protobuf==3.20.0
+# Get the requirement files from the commit be70702dc80d59126d6b4011d6cc7913193db566
```
krzysz00: First, let's parametrize this like we do MIOpen.

Second, note that we already `pip install` some of these elsewhere in the Dockerfile.

Third, I don't like having tight version bounds like this somewhere where we won't think to touch them - can we drop the `==` bits? Unless we already had a bunch of those.

manupak: Parameterize which part and with reference to which part related to MIOpen?
(If you can point me to that in this file)

I can have a go which parts we can drop from pip installs...

manupak: Also noting pip installs are for verify function of MIGraphX - so I thought we didn't want a version skew..
Just curious, whats wrong with specific versions rather pip taking some version based on pip/python version from a CI perspective?

krzysz00: It's not that I'm opposed to pinning a version, it's that I don't want us to not bump the version because MIGraphX changed it and we didn't notice

krzysz00: Parametrize the MIGraphX commit. And I take that comment back because we just use MIOpen's develop at the time of Docker build

krzysz00: I still don't want to see commit IDs that're spread out into five different places though.

manupak: Ack about commit.

Re: pinned versions

Not sure by not pinning versions, we can avoid having to bump the versions. I mean if it does break with implicit versions, we have to bump anyway.

I ll try and see whats really necessary here

krzysz00: I'm not opposed to pinning versions, but I want to parse out the pinned versions from some canonical source instead of hardcoding them in our Dockerfile.

manupak:
Right, I managed to remove the adhoc python dependencies (which I previously collected from : https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/tools/install_prereqs.sh)

They are for testing and I now doubt we need them for scope of testing we do.
If we ever need them, I dont think us maintaining a parser to figure out the versions is the right solution.
Rather, I think we need to get MIGraphX to maintain them in a requirements file for their clients to use.

Also parameterized the commit to ENV which we could use in the Jenkins file.

mlir/utils/jenkins/Dockerfile
jerryyin: Eventually if there's similar amount of tests from MIGraphx we'd want to follow the convention from MIOpen and remove that pinned version. We should theoretically always need the TOT in each docker builds and manually bumped commit hash is a pain to manage by hand. I'm good with what you have now after the parameterized change.

jerryyin (approved):
LGTM

krzysz00:
I'd still much rather we use `develop` here so that dependency changes get picked up on our once a release-ish Docker build automatically and so that when we do the CI it tests our tip against MIGraphX tip

manupak:
I believe its a choice end of the day... 

I guess I have PTSD from my prior experience with dealing with projects with large number of dependencies where we resorted to do a once a release bump of versions (after testing rather than reacting to failures) at a single place. 

> so that dependency changes get picked up on our once a release-ish Docker build

I personally dont believe it is much of a burden to figure out the commit (as and when we rebuild the image) when we do that change the docker image vs just rebuilding the image. Well, if we are doing once a release docker image build then I would say tracking develop does not add much value (save some clicks to figure out the commit).

However, if you guys feels strongly we should just track 'develop' -- Im fine with that -- it might turn out fine as well.

jerryyin:
> if we are doing once a release docker image

Actually we do more frequent docker change as long as we have interface dependent features. Due to the project nature of MLIR now, we serve as a "plugin" to other project such as MIGraphX or MIOpen. Each time when we do the pointer bump it is almost guaranteed that we want to track the tip of the client at the time.

> I would say tracking develop does not add much value (save some clicks to figure out the commit)

I think automating this process is an important value, and its hard to guarantee people will remember to do it every time. On the other hand, forgetting to do it could break migraphx build (if we follow the same convention of MIOpen tests) and cause unnecessary trouble.

krzysz00:
The three of us should hop onto a call and hash this out?

manupak:
Maybe after the meeting ?

mlir/utils/jenkins/build-dockerfile.sh
```diff
@@ -20,25 +20,29 @@ start_clean_docker() {
 
 # Build the docker image and push it to the repository
 build_rocm_image() {
-  if docker build -t rocm-mlir -f ./mlir/utils/jenkins/Dockerfile ./mlir/utils/jenkins ; then
```
manupak, line 23: Is my bash knowledge failing me here or how did this ever work ? 
From what I know if a command succeeded, then this if clause is executed.

Just to be sure did some lookarounds as well : https://unix.stackexchange.com/questions/22726/how-to-conditionally-do-something-if-a-command-succeeded-or-failed

krzysz00: ... yeah, this is exactly backwards, good catch!

krzysz00 (approved):
LGTM and thanks for catching my trivial bug

manupak:
Do I have to run CI to get this merged ? (I skipped it because this PR doesn't affect anything)

manupak:
(nvm, I ll just run it)

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -65,6 +65,7 @@ String dockerArgs() {
 }
 
 String dockerImage() {
+    // If this is being changed please change Dockerfile.ci-migraphx's base image as well
```
jerryyin: If I read it right, now we have two set of images:
- rocm/mlir:rocm4.5-latest
- rocm/mlir.ci-migraphx:rocm4.5-latest

My questions are:
- Is the channel `rocm/mlir.ci-migraphx` created? I don't know if you start pushing to a non-existent channel, docker hub will create it.
- When will CI ever pull in the `rocm/mlir.ci-migraphx:rocm4.5-latest` images? I see we populate images but don't see a place that would consume those. Do you plan to work on that in follow-up PRs?

manupak: @jerryyin,

I think it will be created, if I am not misataken and Jenkins user has authority. If not I ll reach out.

Yea the following PR would pull them. I would need to see the logs in the PR branch see if everything is in place once this PR is landed.

manupak: ( I ll look into it tomorrow once Im back)

jerryyin: @manupak I don't know you are on vacation. Our team don't expect anyone on vacation to reply to github messages or emails. Otherwise thanks for the clarification and it sounds good. (I doubt jenkins user has the permission to create a new channel in dockerhub so just make sure you double check this is working as expected when you are back.)

manupak: @jerryyin I think you might be right about push event not creating a repository. (I think the "channel" vs tag confused me :) ).

So yes I need some help to get a repo created then.

manupak: (Sorry, I forgot to update my MS status)

mlir/utils/jenkins/build-dockerfile.sh
```diff
@@ -59,9 +63,9 @@ build_rocm_image() {
 main() {
   if [[ " ${ARGS[*]} " =~ " --force " ]] || \
       git diff --name-only HEAD^ HEAD | grep -q "Dockerfile"; then
-    echo "Start a new build"
     start_clean_docker
-    build_rocm_image
+    build_rocm_image ""
+    build_rocm_image ".ci-migraphx"
```
manupak: Note to self : Change this to agreed rocm/mlir-migraphx-ci before merge

manupak (commented):
Let me do a quick change

mlir/utils/jenkins/build-dockerfile.sh
manupak: Done

PR #896, opened 2022-12-07
[CI] Adding initial MIGraphX integration tests

manupak:
This commit adds a jenkins stage (default disabled) to check the MIGraphX intergration using Resnet50
model.

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/696

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -40,7 +40,7 @@ void buildMIGraphX(String cmakeOpts) {
 }
 
 void getAndBuildMIGraphX(String cmakeOpts) {
-    git branch: 'be70702dc80d59126d6b4011d6cc7913193db566', poll: false,\
+    git branch: params.MIGraphXBranch, poll: false,\
         url: 'https://github.com/ROCmSoftwarePlatform/AMDMIGraphX.git'
     buildMIGraphX(cmakeOpts)
 }
@@ -314,6 +314,11 @@ pipeline {
                      description: 'Run weekly-only jobs')
         string(name: 'MIOpenBranch', defaultValue: 'develop',
                description: 'The MIOpen branch to be used with the job')
+        // MIGRAPHX_COMMIT is defined in the Dockerfile
+        // So in order to update the commit, please change the Dockerfile
+        string(name: 'MIGraphXBranch', defaultValue: '${MIGRAPHX_COMMIT}',
+               description: 'The MIGraphX branch/commit to verify against.')
+
 
         // Each below control whether to run a individual stage from parallel run
         // They default to true or empty but deverloper can toggle them for debugging purpose
@@ -826,6 +831,7 @@ pipeline {
             when {
                 beforeAgent true;
                 equals expected: true, actual: params.checkMIGraphX;
+                equals expected: true, actual: params.nightly;
             }
             matrix {
                 axes {
@@ -858,15 +864,17 @@ pipeline {
                     }
                     stage("Install MIGraphX Dependencies") {
                         steps {
+                            // We are installing miopen-hip here (and not in Dockerfile) to avoid collision with MIOpen checks above
                             sh 'apt-get install miopen-hip'
-                            sh 'cget -p /MIGraphXDeps install ${WORKSPACE} -DBUILD_MIXR_TARGET=On -DLLVM_ENABLE_ZSTD=Off -DLLVM_ENABLE_THREADS=Off'
+                            // Package and install current checkout of rocMLIR as MIGraphX dependency.
+                            sh 'cget -p ${WORKSPACE}/MIGraphXDeps install ${WORKSPACE} -DBUILD_MIXR_TARGET=On -DLLVM_ENABLE_ZSTD=Off -DLLVM_ENABLE_THREADS=Off'
                         }
                     }
                     stage("Build MIGraphX with MLIR") {
                         steps {
                             sh 'rm -rf MIGraphX'
                             dir('MIGraphX') {
-                                getAndBuildMIGraphX("-DCMAKE_PREFIX_PATH=/MIGraphXDeps -DMIGRAPHX_ENABLE_MLIR=On -DCMAKE_CXX_COMPILER=/opt/rocm/llvm/bin/clang++")
+                                getAndBuildMIGraphX("-DCMAKE_PREFIX_PATH='/MIGraphXDeps;${WORKSPACE}/MIGraphXDeps' -DMIGRAPHX_ENABLE_MLIR=On -DCMAKE_CXX_COMPILER=/opt/rocm/llvm/bin/clang++")
                             }
                         }
                     }
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -29,6 +29,22 @@ void getAndBuildMIOpen(String prefixOpt, String cmakeOpts) {
     buildMIOpen(cmakeOpts)
 }
 
+void buildMIGraphX(String cmakeOpts) {
+    sh '[ ! -d build ] || rm -rf build'
+    cmakeBuild generator: 'Unix Makefiles',\
+        buildDir: 'build',\
+        buildType: 'Release',\
+        installation: 'InSearchPath',\
+        cmakeArgs: ${cmakeOpts}
+    sh 'cd build; make -j $(nproc)'
+}
+
+void getAndBuildMIGraphX(String cmakeOpts) {
+    git branch: 'be70702dc80d59126d6b4011d6cc7913193db566', poll: false,\
+        url: 'https://github.com/ROCmSoftwarePlatform/AMDMIGraphX.git'
+    buildMIGraphX(cmakeOpts)
+}
+
 void buildMIOpenWithMLIR() {
     buildProject('librockCompiler',
                     '-DBUILD_FAT_LIBROCKCOMPILER=ON')
@@ -307,6 +323,9 @@ pipeline {
                      description: 'Run the static library stage')
         booleanParam(name: 'perfTest', defaultValue: true,
                      description: 'Run the performance testing stage')
+        //TODO(@manupak): Switch the default to True after testing
+        booleanParam(name: 'checkMIGraphX', defaultValue: false,
+                     description: 'Run the MIGraphX integration tests')
 
         // choose the codepath for testing
         choice(name: 'codepath',
@@ -790,6 +809,73 @@ pipeline {
                                 sh 'python3 ./bin/perfRegressionReport.py ${CHIP} ./oldData/${CHIP}_mlir_vs_rocblas_perf.csv ./${CHIP}_mlir_vs_rocblas_perf.csv'
                             }
                             postProcessPerfRes("${CHIP}")
+                        }
+                    }
+                }
+                post {
+                    unsuccessful {
+                        rebootNode()
+                    }
+                    always {
+                        cleanWs()
+                    }
+                }
+            }
+        }
+        stage ("MIGraphX") {
+            when {
+                beforeAgent true;
+                equals expected: true, actual: params.checkMIGraphX;
+            }
+            matrix {
+                axes {
+                    axis {
+                        name 'CODEPATH'
+                        values 'mfma', 'vanilla', 'navi21'
+                    }
+                }
+                agent {
+                    docker {
+                        image dockerImage()
+                        args dockerArgs()
+                        label getLabelFromCodepath("${CODEPATH}")
+                        alwaysPull true
+                    }
+                }
+                when {
+                    beforeAgent true
+                    expression { return shouldRunFromCodepath("$CODEPATH") }
+                }
+                environment {
+                    HOME="${WORKSPACE}"
+                }
+                stages {
+                    stage('Environment') {
+                        steps {
+                            echo "codepath is ${CODEPATH}"
+                            showEnv()
+                        }
+                    }
+                    stage("Install MIGraphX Dependencies") {
+                        steps {
+                            sh 'apt-get install miopen-hip'
+                            sh 'cget -p /MIGraphXDeps install ${WORKSPACE} -DBUILD_MIXR_TARGET=On -DLLVM_ENABLE_ZSTD=Off -DLLVM_ENABLE_THREADS=Off'
+                        }
+                    }
+                    stage("Build MIGraphX with MLIR") {
+                        steps {
+                            sh 'rm -rf MIGraphX'
+                            dir('MIGraphX') {
+                                getAndBuildMIGraphX("-DCMAKE_PREFIX_PATH=/MIGraphXDeps -DMIGRAPHX_ENABLE_MLIR=On -DCMAKE_CXX_COMPILER=/opt/rocm/llvm/bin/clang++")
+                            }
+                        }
+                    }
+                    stage("Verify MIGraphX with MLIR") {
+                        steps {
+                            dir('MIGraphX/build') {
+                                sh 'wget https://github.com/onnx/models/raw/ba629906dd91872def671e70177c5544e0ea9e02/vision/classification/resnet/model/resnet50-v1-7.onnx'
+                                sh './bin/driver verify --onnx resnet50-v1-7.onnx'
+                            }
                         }
                     }
                 }
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -35,7 +35,7 @@ void buildMIGraphX(String cmakeOpts) {
         buildDir: 'build',\
         buildType: 'Release',\
         installation: 'InSearchPath',\
-        cmakeArgs: ${cmakeOpts}
+        cmakeArgs: "${cmakeOpts}"
     sh 'cd build; make -j $(nproc)'
 }
 
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -86,7 +86,7 @@ String dockerImage() {
 }
 
 String dockerImageCIMIGraphX() {
-    return 'rocm/mlir.ci-migraphx:rocm5.3-latest'
+    return 'rocm/mlir-migraphx-ci:rocm5.3-latest'
 }
 
 void preMergeCheck(String codepath) {
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -881,6 +881,10 @@ pipeline {
                     stage("Verify MIGraphX with MLIR") {
                         steps {
                             dir('MIGraphX/build') {
+                                // Verify MLIR unit tests
+                                sh 'make -j$(nproc) test_gpu_mlir'
+                                sh 'ctest -R test_gpu_mlir'
+                                // Verify ResNet50
                                 sh 'wget https://github.com/onnx/models/raw/ba629906dd91872def671e70177c5544e0ea9e02/vision/classification/resnet/model/resnet50-v1-7.onnx'
                                 sh './bin/driver verify --onnx resnet50-v1-7.onnx'
                             }
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -85,6 +85,10 @@ String dockerImage() {
     return 'rocm/mlir:rocm5.3-latest'
 }
 
+String dockerImageCIMIGraphX() {
+    return 'rocm/mlir.ci-migraphx:rocm5.3-latest'
+}
+
 void preMergeCheck(String codepath) {
     // Only do static check on mfma codepath during PR CI
     if ( (params.nightly == false) && (codepath == "mfma") ) {
@@ -316,7 +320,7 @@ pipeline {
                description: 'The MIOpen branch to be used with the job')
         // MIGRAPHX_COMMIT is defined in the Dockerfile
         // So in order to update the commit, please change the Dockerfile
-        string(name: 'MIGraphXBranch', defaultValue: '${MIGRAPHX_COMMIT}',
+        string(name: 'MIGraphXBranch', defaultValue: 'develop',
                description: 'The MIGraphX branch/commit to verify against.')
 
 
@@ -842,7 +846,7 @@ pipeline {
                 }
                 agent {
                     docker {
-                        image dockerImage()
+                        image dockerImageCIMIGraphX()
                         args dockerArgs()
                         label getLabelFromCodepath("${CODEPATH}")
                         alwaysPull true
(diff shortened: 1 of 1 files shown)


manupak:
depens on https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/895

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -29,6 +29,22 @@ void getAndBuildMIOpen(String prefixOpt, String cmakeOpts) {
...
+    cmakeBuild generator: 'Unix Makefiles',\
+        buildDir: 'build',\
+        buildType: 'Release',\
+        installation: 'InSearchPath',\
+        cmakeArgs: ${cmakeOpts}
+    sh 'cd build; make -j $(nproc)'
+}
+
+void getAndBuildMIGraphX(String cmakeOpts) {
+    git branch: 'be70702dc80d59126d6b4011d6cc7913193db566', poll: false,\
```
krzysz00: branch `develop`

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -306,6 +322,9 @@ pipeline {
                      description: 'Run the static library stage')
         booleanParam(name: 'perfTest', defaultValue: true,
                      description: 'Run the performance testing stage')
+        //TODO(@manupak): Switch the default to True after testing
```
krzysz00: I'd just add this to, say, `weekly`

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -793,5 +812,72 @@ pipeline {
...
+                    stage('Environment') {
+                        steps {
+                            echo "codepath is ${CODEPATH}"
+                            showEnv()
+                        }
+                    }
+                    stage("Install MIGraphX Dependencies") {
+                        steps {
+                            sh 'apt-get install miopen-hip'
+                            sh 'cget -p /MIGraphXDeps install ${WORKSPACE} -DBUILD_MIXR_TARGET=On -DLLVM_ENABLE_ZSTD=Off -DLLVM_ENABLE_THREADS=Off'
```
krzysz00: Let's not install into `/`, but go into `$WORKSPACE/MIGraphXDeps`

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -793,5 +812,72 @@ pipeline {
...
+                        steps {
+                            sh 'rm -rf MIGraphX'
+                            dir('MIGraphX') {
+                                getAndBuildMIGraphX("-DCMAKE_PREFIX_PATH=/MIGraphXDeps -DMIGRAPHX_ENABLE_MLIR=On -DCMAKE_CXX_COMPILER=/opt/rocm/llvm/bin/clang++")
+                            }
+                        }
+                    }
+                    stage("Verify MIGraphX with MLIR") {
+                        steps {
+                            dir('MIGraphX/build') {
```
krzysz00, line 884: Can we run the unit tests for MIGraphX as well?

mlir/utils/jenkins/Jenkinsfile
manupak: do we want a moving target ? I think that is fine for weekly/nightly to see when it brakes -- but I'd like to stop PRs getting merged that breaks that we know to work.

mlir/utils/jenkins/Jenkinsfile
manupak: well the dependencies are already there in the container. Similiar to MIOpenDeps. (it gets installed in the other PR dockerfile changes)

mlir/utils/jenkins/Jenkinsfile
manupak: We can but not all of them are sensitive our changes -- if anything we should only run test_gpu_mlir for which I dont know a way to carve that out.

manupak: (I can take look if there is something like LIT_FILTER)

mlir/utils/jenkins/Jenkinsfile
manupak: So it needs to be a location in the container which already contains the static dependencies installed.

mlir/utils/jenkins/Jenkinsfile
manupak: same concern as above but I'd like a some subset (that we can bikeshed) to run on a PR basis

mlir/utils/jenkins/Jenkinsfile
krzysz00: PR CI's already pretty long, and I don't want to add another build and a rather substantial validation sequence to it.

krzysz00: I'd like to see this running nightly or weekly and also a flag to allow us to force testing it on PRs where we're concerned.

krzysz00: And yes, the point is to have a moving target, though we can have a `migraphx-branch` parameter analogous to miopen-branch for when we've got a pending PR

mlir/utils/jenkins/Jenkinsfile
krzysz00: Yes, but then you shouldn't be running `cget` at all? Just run `cmake`, and to get the MLIR library in, copy it into a directory like how we do it for MIOpen.

mlir/utils/jenkins/Jenkinsfile
krzysz00: Fair enough. 

Also, do we know if the validation will for sure go through MLIR?

mlir/utils/jenkins/Jenkinsfile
manupak: MIGraphX uses more than rocmlir-lib exposes mixr CAPI s and whatnot. I never got that working. It uses more MLIR headers. From what Ive heard from @jungpark-mlir, this is the only way to make it work. If someone knows how to do that Im all ears..

krzysz00: First off, any `apt install` steps should be in the Dockerfile.

Second, whatever build flags are needed for the MIGraphX integration - I think it's to do with headers - you can just pass to our CMake, and then use our `install` into a prefix.

Third, this way doesn't actually test against whatever the CI pipeline is being run against, it tests against MiGraphX's commit, which is a useless test for us.

krzysz00: And fourth, use MIGraphX build flags to turn off, for example, the MIOpen dependcy, if you can.

mlir/utils/jenkins/Jenkinsfile
manupak: Whilst Im not a fan of moving parts in CI infra. i.e. you cant just checkout an early commit and run CI because u dont know what develop looked like then.

I can agree to parametrized branch where we can define it in the job what to use. 

But still, I would push to updating at agreed upon cadence to update commit rather than freely tracking a development branch

manupak: Re: PR CI, fair enough. Lets revisit if we have nightly or weekly breakages then?

manupak: (long is subjective.. I suppose but I like this project's faster CI ... So lets keep it that way)

mlir/utils/jenkins/Jenkinsfile
manupak: > First off, any apt install steps should be in the Dockerfile.

I think you are referring miopen-hip installation.
Yes, we can, I just kept it here to avoid any installation conflicts with whatever miopen installation happens before...
if that's not an issue I can move that.

> Second, whatever build flags are needed for the MIGraphX integration - I think it's to do with headers - you can just pass to our CMake, and then use our install into a prefix.

I can have a go tomorrow and see if this works.

> Third, this way doesn't actually test against whatever the CI pipeline is being run against, it tests against MiGraphX's commit, which is a useless test for us.

This is not correct (I think). cget uses the current folder (current rocMLIR checked out version) checked out in the workspace to package. -- may I ask how did you figure out it uses MiGraphX's commit or its relevance to what rocMLIR commit's being used.

> And fourth, use MIGraphX build flags to turn off, for example, the MIOpen dependcy, if you can.

I don't think this is possible as MIGraphX only offloads MLIR if there are fused kernels and standalone go to MIOpen -- so its not binary in that sense.

mlir/utils/jenkins/Jenkinsfile
manupak: -DMIGRAPHX_ENABLE_MLIR=On is the guarantee.

manupak: (Happy to know if there is an alternative way to make sure)

mlir/utils/jenkins/Jenkinsfile
krzysz00: The other thing here is that, even though we're running in a Docker container, we want all the things we install to be confined to $WORKSPACE. So that MIGraphX build will need a CMAKE_PREFIX_PATH that has two paths - one for the global dependencies, one for the MRIR we just installed.

... Come to think of it, in the Dockerfile, why install to /MIGraphXDeps at all, just go to /usr/local the same way the MIOpen dependencies do.

Given the flags you passed to `cget`, I expected it to be running in the MIGraphX context on initial reading, and so therefore to use MIGraphX's requirements files to get and build a MLIR. Heck, it has `get` in the name. But it's quite possible that this runs on $WORKSPACE and so is therefore fine.

And so that being said, use an install target of `${WORKSPACE}/MIGraphXDeps` not `/MIGraphXDeps`

krzysz00: Also, given the above, I'd like to see a very simple smoke test that runs one fused kernel through MLIR and verifies it so that we can disable out as much of MIGraphX (like that MIOpen dependency) as possible. I want this CI run to basically be "is the MIGraphX integration fundamentally broken (things don't compile/a simple example doesn't run) if you move to MLIR tip"

manupak: Only concern with /usr/local is a potential dependency collision with other clients. It felt compartmentalized to install it to a separate folder. I am still not getting the value of having to use under WORKSPACE dir vs usr/local vs /MIGraphXDeps. It is just installing the rocMLIR dependency to the where static dependencies of MIGraphX is installed.

Very simple smoke test you mention are the ones in test gpu mlir unit test of MIGraphX, I ll see if I can get that test to run, if not we d have to resort to a trimmed onnx model

krzysz00: Ok, so, we can have two directories, `/MIGraphXDeps`  and `$WORKSPACE/MIGraphXDeps`.

The reason for the workspace bit is that Jenkins jobs are designed to only touch subdirectories of WORKSPACE, and breaking that can lead to all sorts of weird things down the road, especially when it comes to, say, restarting build, asrchiving things, or maybe, say, we enable some sort of Docker container reuse or something.

The other advantage is that being WORKSPACE only allows us to run our CI jobs as a non-root user. I think that's not something we're currently doing, but we should.

manupak: Ok, that makes sense.

mlir/utils/jenkins/Jenkinsfile
manupak: I've added it 'when' of the stage following the pattern elsewhere but used nightly instead.

mlir/utils/jenkins/Jenkinsfile
manupak: Done

mlir/utils/jenkins/Jenkinsfile
manupak: OK, I've added the running mlir unit tests.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -297,6 +313,11 @@ pipeline {
                      description: 'Run weekly-only jobs')
         string(name: 'MIOpenBranch', defaultValue: 'develop',
                description: 'The MIOpen branch to be used with the job')
+        // MIGRAPHX_COMMIT is defined in the Dockerfile
+        // So in order to update the commit, please change the Dockerfile
```
krzysz00: I'm still going to insist that this should be develop

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -793,5 +817,79 @@ pipeline {
                 }
             }
         }
+        stage ("MIGraphX") {
+            when {
+                beforeAgent true;
+                equals expected: true, actual: params.checkMIGraphX;
```
krzysz00: Do these need an anyOf{} ? Not sure. Also, I don't think you defined the check MIGraphX thing

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -793,5 +817,79 @@ pipeline {
...
+                            }
+                        }
+                    }
+                    stage("Verify MIGraphX with MLIR") {
+                        steps {
+                            dir('MIGraphX/build') {
+                                // Verify MLIR unit tests
+                                sh 'make -j$(nproc) test_gpu_mlir'
+                                sh 'ctest -R test_gpu_mlir'
+                                // Verify ResNet50
```
krzysz00, line 888: If we remove this, can we not install MIOpen?

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -793,5 +817,79 @@ pipeline {
...
+                    stage('Environment') {
+                        steps {
+                            echo "codepath is ${CODEPATH}"
+                            showEnv()
+                        }
+                    }
+                    stage("Install MIGraphX Dependencies") {
+                        steps {
+                            // We are installing miopen-hip here (and not in Dockerfile) to avoid collision with MIOpen checks above
+                            sh 'apt-get install miopen-hip'
```
krzysz00: As an alternative, maybe use or create a build of MIOpen here and add it to $PATH? (Ex. If this were moved onto the same node as the perf test to save on MIOpen rebuilds)

krzysz00: _Or_ we could create a second Docker that inherits from the first but includes MIOpen and the MIGraphX dependencies

mlir/utils/jenkins/Jenkinsfile
manupak: I'd defer that to the image where dependencies are installed -- I insist we need to agree on what the image uses -- and lets not decide that here.

mlir/utils/jenkins/Jenkinsfile
manupak: No, MIOpen is a first class dependency which we cant disable atm.

mlir/utils/jenkins/Jenkinsfile
manupak: I can easily create a function including the logic in ""Build MIOpen with HIP" but then again I dont understand differential concern as that step (couple of lines above) also does a "make install" which installs miopen outside WORKSPACE.  Maybe the right thing is to do that to a prefix within WORKSPACE here and there as well.

Second docker image is not that bad of an idea -- but not sure we need it just yet.

mlir/utils/jenkins/Jenkinsfile
manupak: I think I defined it here : https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/896/files#diff-16af82a543140fc9970902655eb9815da207d857b849d253ab13c4c0f344aa6aR331 (line 331)

You mean not a AND condition but a OR condition ?

mlir/utils/jenkins/Jenkinsfile
manupak: > As an alternative, maybe use or create a build of MIOpen here and add it to $PATH? (Ex. If this were moved onto the same node as the perf test to save on MIOpen rebuilds)

apt-get install does not do a build anyway, right ?

mlir/utils/jenkins/Jenkinsfile
krzysz00: Yeah, an OR, so we can set this stage off on PRs if we want to

mlir/utils/jenkins/Jenkinsfile
krzysz00: I don't think the image and the CI being out of sync a bit on dependency versions is fine - we've been doing it for MIOpen with no issues

mlir/utils/jenkins/Jenkinsfile
krzysz00: Wait, the MIOpen with HIP does a make install? Who added that??

And another advantage to adding a second Docker image is that we don't have to `wget` Resnet50 every time

mlir/utils/jenkins/Jenkinsfile
manupak: Yea make sense! I ll change that to an OR then.

manupak:
@jerryyin @krzysz00 I think I captured our discussion and implemented it. PTAL when you have sometime.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -793,5 +818,78 @@ pipeline {
...
+                                getAndBuildMIGraphX("-DCMAKE_PREFIX_PATH='/MIGraphXDeps;${WORKSPACE}/MIGraphXDeps' -DMIGRAPHX_ENABLE_MLIR=On -DCMAKE_CXX_COMPILER=/opt/rocm/llvm/bin/clang++")
+                            }
+                        }
+                    }
+                    stage("Verify MIGraphX with MLIR") {
+                        steps {
+                            dir('MIGraphX/build') {
+                                // Verify MLIR unit tests
+                                sh 'make -j$(nproc) test_gpu_mlir'
+                                sh 'ctest -R test_gpu_mlir'
```
krzysz00, line 887: Aren't the make and ctest lines equivalent? (I don't know ctest, maybe the answer is no)

krzysz00 (approved):
LGTM aside from the possible double-test. Do make sure this actually successfully runs the tests you want before merging, though.

mlir/utils/jenkins/Jenkinsfile
manupak: The make will just build the test but not run them as the tests itself are a bit like cpp apps themselves.

manupak:
Thanks @krzysz00, the plan is to do a run with checkMIGraphX on the PR branch (pre-merge) and see the logs after docker changes get merged and images are created.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -793,5 +818,78 @@ pipeline {
                 }
             }
         }
+        stage ("MIGraphX") {
+            when {
+                beforeAgent true;
+                anyOf {
+                    equals expected: true, actual: params.checkMIGraphX;
+                    equals expected: true, actual: params.nightly;
```
jerryyin, line 836: My opinion is to run:
 - Minimal test using migraphx driver on PR CI (1 config per direction, per data type)
 - Resnet50 test in weekly CI

I don't think nightly CI is the right place for it (Unless we want to measure performance impact, on a config level). 

This is just so that there is 1:1 correspondence between MIOpen and MIGraphX integration.

manupak: We do run Resnet50 e2e tests (in tosa & migraphx dialects) in a nightly basis and they are not that slow to be pushed to be weekly (as compared to tuning). The e2e tests are more functional in nature that captures everything is correct (migraphx passes that happens before MLIR offloading and their compatibility of the MLIR offload done later in their pipeline) -- so unless there is a concern on time increase , I would prefer it to be nightly.

MIGraphx is inference only AFAIK, hence no directions. I feel extending the testing on MIGraphX is task on its own.
I do not intend extend the existing unit tests in the scope of the PR to keep this as basic minimal PR to get the CI bits hooked up. We can do follow up on MIGraphX side (that owns the unit tests for MLIR) to put more things there, if needed. The scope of this PR is just to make sure whatever tests and functionality that exists and tested are not broken by development in MLIR.

manupak: However, we could run the unit tests on the PR basis but our prior discussion was to MIGraphX testing on PR that we suspect to break things (e.g. fusion) using the Jenkins param.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -29,6 +29,22 @@ void getAndBuildMIOpen(String prefixOpt, String cmakeOpts) {
     buildMIOpen(cmakeOpts)
 }
 
+void buildMIGraphX(String cmakeOpts) {
+    sh '[ ! -d build ] || rm -rf build'
+    cmakeBuild generator: 'Unix Makefiles',\
+        buildDir: 'build',\
+        buildType: 'Release',\
+        installation: 'InSearchPath',\
+        cmakeArgs: ${cmakeOpts}
```
zhanglx13: You need to quote ${variables} for Jenkins to expand the variable.

zhanglx13 (changes requested):

mlir/utils/jenkins/Jenkinsfile
manupak: Oh! fixed now.

zhanglx13 (approved):

manupak:
@jerryyin,

From the CI run, it looks it does the things right : http://rocmhead.amd.com:8080/blue/rest/organizations/jenkins/pipelines/MLIR/pipelines/mlir/branches/PR-896/runs/25/log/?start=0

image

w.r.t.

image

As you can see it is quite fast -- hence I personally dont prefer the idea of putting in the weekly.

In the same time, if we want to move the unit tests to PR and then it is a delta time to run resnet as well because majority of the time (10 mins) is spent in building MIGraphX.

So finally, I think we have to resort to the following options : 

C1. We run these in the nightly and enable MIGraphX CI on PRs that we suspect might only be tested effectively (e.g. fusion) via MIGraphX (which is the current agreement with @krzysz00 and also the current status of this PR)
C2. We run the everything in the PR CI (because I dont see much value in saving < 1min on PR CI time given that we build MIGraphX to run the unit tests)

My vote is for C1 and move to C2 if we see nightly breakages.

(I'm probably be inclined to take this PR for now and do a change if we really want C2 which could be done incrementally on this one).

mlir/utils/jenkins/Jenkinsfile
jerryyin: Thanks for giving detailed reply on your opinions! I don't mean to block this PR with the comment. What you have on nightly with very limited impact on its runtime sounds reasonable. 

PR wise, speaking of my experience from MIOpen side, I think I still prefer some minimal coverage to make sure that things still build and work. There are a dozen times that the MIOpen minimal coverage PR testing captured a non-trivial kernel bug and help us to notice that early.

So to rephrase, I think what you have is good. We can have follow-ups on what make sense to track on PR.

manupak: No worries! :) 

I dont disagree with you; and more than happy to promote things to PR CI if things are falling between the cracks (i.e. seeing our nightlies fail because we are not enforcing checkMIGraphX option on relevant PRs as we thought we would be).

PR #897, opened 2022-12-07
Remove Tosa layerwise constant folding to avoid mixed layout in Rock

krzysz00:
We had this pass added early in our Tosa pipeline for reasons I'm not entirely clear on. The tosa to linalg passes will run this transformation, and those passes run after tosa-to-rock, and so therefore we don't lose the fact that constant tensors are NCHW. We also won't have to go through a lot of dramatics to un-transpose the constants.


krzysz00:
Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/745 in a "making the problem in that ticket not happen in the first place" way

sjw36 (approved):
This pass will run in tosa-to-linalg after partitioning so it will not affect gpu kernel code. 
Approved.

PR #900, opened 2022-12-08
[DO NOT MERGE] Move the second LDS barrier

jungpark-mlir:
I might be forgetting something but reviewed the barrier again and wonder this can be reschduled as this.

The point is, MFMA instruction itself doesn't really resolve or set any LDS/WG related dependency. Accumulator register is not shared across waves and only LDS-to-(acc)register data read needs to be fenced.

Local test looks fine but need more tests and performance numbers.

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
@@ -1282,7 +1282,8 @@ struct GridwiseGemmV2RewritePattern
               .getBeginOperandIndex(),
           iv);
 
-      // LDS barrier.
+      // LDS barrier : LDS_fence and WG_barrier to guarantee LDS updates are
+      // completed.
       b.create<LDSBarrierOp>(loc);
 
       // Emit blockwise GEMM.
@@ -1291,10 +1292,6 @@ struct GridwiseGemmV2RewritePattern
           b.getIndexAttr(ldsBlockBOffset), mMyWaveOffsetA, mMyWaveOffsetB,
           arrayA, arrayB, regCAllocOp, op.getBlockSizeAttr(),
           op.getParamsAttr());
-
-      // LDS barrier.
-      // This barrier prevents halo part of outputs having weird values.
-      b.create<LDSBarrierOp>(loc);
 
       // Emit blockwise stores
       BlockAndValueMapping storeAUpdates, storeBUpdates;
--- a/mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
@@ -554,6 +554,10 @@ struct BlockwiseGemmV2RewritePattern
       lklb.create<memref::StoreOp>(loc, valueB, bufferB, ValueRange{lkliv});
     }
 
+    // LDS barrier: LDS_fence and WG_barrier to ensure all data has been read
+    // out from LDS.
+    b.create<LDSBarrierOp>(loc);
+
     // Workload of either MPerWave and NPerWave that are larger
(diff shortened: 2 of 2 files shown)


mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1292,10 +1293,6 @@ struct GridwiseGemmV2RewritePattern
           arrayA, arrayB, regCAllocOp, op.getBlockSizeAttr(),
           op.getParamsAttr());
 
-      // LDS barrier.
-      // This barrier prevents halo part of outputs having weird values.
-      b.create<LDSBarrierOp>(loc);
```
krzysz00, line 1297: This needs to be a barrier, but not necessarily an LDS one. A straight `s_barrier` is sufficient, probably by way of the intrinsic. The pathological case is that wave N runs to storing data to LDS before wave M has had the chance to do any MFMAs

krzysz00: Or is that case handled by the move you made somehow

krzysz00 (changes requested):
Parking a blocker on this because I'm not convinced

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
jungpark-mlir: Yes, this change sets barrier after all waves got their data in the register to prevent that.

krzysz00: Ah ... yeah, that's actually a good point. Can you comment that over in gridwise gemm that the barrier is in blockwise gemm?

krzysz00: I'll be happy to land this, thanks for the explanation.

krzysz00: (but can you do this for non-xdlops too?)

jungpark-mlir: I'm not sure this also applies to non-xdlops, need to have a look into non-xdlops loop.

krzysz00: It's got the same high-level structure.

Also, can we get a by-hand perf run here?

jungpark-mlir: I'll have a look

jungpark-mlir: Unfortunately perf run proved this is not better. Mostly par but quite a few kernels show ~1% slower. 
Possibly because moving barrier before the mfma forced mfma to wait for all waves to finish reading from LDS and that's more than the benefits from the mfma overlapping with the later instructions.

Also experimented with simply removing that second LDS barrier, also doesn't give a better overall performance - 2 configs got 10% faster but 3 configs got 2 times slower.

Will close this, no point to have it. Was nice to refresh my memory around this.

krzysz00: Thanks for experimenting!

jerryyin: Yes, I think the both cases lds barrier guarantees LDS usage are done with current iteration before moving to the next. 

It is easier to maintain/understand this with both barriers emitted in the same lowering pass.

jungpark-mlir:
Close as discussed

PR #901, opened 2022-12-09
Using Tuning data

jungpark-mlir:


jungpark-mlir:
missing items in the problem definition
- convolution layout
- arch
- features

mlir/include/mlir/Dialect/Rock/Tuning/RockTuning.h
```diff
@@ -35,8 +35,20 @@ struct TunableParams {
 };
 
 TunableParams *createTunableParamSpace(ModuleOp &mod);
-
 bool tuningSetParam(ModuleOp &mod, ParamEntry *paramEntry);
+bool tuningSetStr(ModuleOp &mod, std::string perfConfig);
+
+struct TuningTable {
+  std::map<std::string, std::pair<std::string, float>> tuningMap;
```
krzysz00, line 42: `llvm::StringMap` perhaps, since we have it?

krzysz00 (changes requested):
I'm sticking a blocker on this because I'm not sure I like the hash-of-operation approach here - esp. since it'd be incompatible with short-term tuning data generation. And because I don't think this API will be useful to have around/I don't want to lock anyone into something like this.

I'd be wiling to land a version of this, but the thing as it stands isn't a good long-term plan and isn't compatible with what I'm planning to do to get tuning numbers.

(And yeah, I can unblock this if we really want it in by Wednesday night)

krzysz00:
Also, can we get a more substantial  description on this?

jungpark-mlir:
> I'm sticking a blocker on this because I'm not sure I like the hash-of-operation approach here - esp. since it'd be incompatible with short-term tuning data generation. And because I don't think this API will be useful to have around/I don't want to lock anyone into something like this.
> 
> 
> 
> I'd be wiling to land a version of this, but the thing as it stands isn't a good long-term plan and isn't compatible with what I'm planning to do to get tuning numbers.
> 
> 
> 
> (And yeah, I can unblock this if we really want it in by Wednesday night)

Whoops, I forgot to delete the hash function.
It doesn't use it, instead it uses the string extract of the operation.
I considered using llvm hashcode but found it's dynamic and non compatible in the later run.

Yeah, need more comments.

jungpark-mlir:
@krzysz00 BTW, as I commented last week (or tried), though the implementation is for the short-term, I suppose we'll need this level of abstraction layer in the future.
Especially, the DB key is not a part of this API but it's just current implementation. DB key will be a collection of the problem structure fields which are extracted from multiple operations in a fused kernel. The structure should be determined by us and abstracted under the API function (opaque to the client), that's what this API is trying to achieve. I presume this could be still helpful when MIGraphX enables rocMLIR tuning via Tuna.
Also we need to access this kind of functionality when we try to implement MLIR based execution driver. Not sure when but could  be a potential future plan.

Again, it doesn't lock anything around, instead this is suggesting an opaque API to abstract the low level structures so nothing is locked around by the API. For example, I tried to use the operation hash for the first time and than just changed to the combined string without changing the API itself (minor type change was needed though). We can move to a proper problem structure for the fused kernel in the future.
And this also abstracts the pointer to access the DB storage as a `TuningTable` which could make it easier to abstract the interface to apply the tuning parameters. This part needs more discussion with the client (potentially MIGraphX) and the DB service provider (Tuna) for the integration.

To be clear, I'm referring C-API functions as an API and considering everything in the cpp code as an changeable implementation.
Please let me know if you have a totally different plan in your mind which could be limited by this level of abstraction, nothing is fixed for the longer term.

mlir/include/mlir/Dialect/Rock/Tuning/RockTuning.h
jungpark-mlir: So, I tried to see how to use it and found this post
https://stackoverflow.com/questions/49031588/what-to-prefer-a-llvmstringmap-or-stdmap-during-code-generation-using-llvm
An answer points the stringmap is slower than std::map. Not sure it's still relevant but we can stick with std::map for this implementation.

jungpark-mlir:
I've added an option to rocmlir-gen to emit the string serialized problem, this string can be stored in the file as a table key field.

 ./bin/rocmlir-gen --emit-tuning-key -p --arch gfx900
rock.conv2d#memref<1x128x8x3x3xf32>#memref<128x1x8x32x32xf32>#[0 : i32, 0 : i32, 0 : i32, 0 : i32]#[1 : i32, 1 : i32]#[1 : i32, 1 : i32]#gkcyxnigicihiwinogokohowo#amdgcn-amd-amdhsa:gfx900#

krzysz00:
The serialization format is very different from what I'm using in my stuff. I'll be pushing a PR (even though I still haven't made the Jenkinsfile changes) to show what I'm going for.

jungpark-mlir:
> The serialization format is very different from what I'm using in my stuff. I'll be pushing a PR (even though I still haven't made the Jenkinsfile changes) to show what I'm going for.

Cool, hopefully I can easily change the format as you already have. I'll check it and make the change early tomorrow.

krzysz00:
See PR #903 for how I'm fixing to do it

krzysz00:
(If you want to split out the problem serializer into a separate PR, I'd be open to using it as the table key, though it does cost a process spawn)

jungpark-mlir:
@krzysz00 Basically the work is done as below,
```
./bin/rocmlir-gen --emit-tuning-key -p --arch gfx900
amdgcn-amd-amdhsa:gfx900        conv -F 1 -f NCHW -I NCHW -O NCHW -n 128 -c 8 -h 32 -w 32 -k 128 -y 3 -x 3 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -t f32

./bin/rocmlir-gen --arch gfx908 --operation gemm -p --emit-tuning-key
amdgcn-amd-amdhsa:gfx908        gemm -transA false -transB false -g 1 -m 1024 -n 512 -k 769 -t f32
```

I've referenced the order of the fields from `resnet50-miopen-configs` and `bert-configs`
One field I'm not sure about is `-m conv`, do I need to include that?

krzysz00:
I'll need to update the tuning runner/perf script to account for that `conv`/`gemm`, but that's not a big deal

jungpark-mlir:
@krzysz00 can you please approve this so I can merge it as soon as CI comes back normal?

krzysz00:
It looks like we'd need to make some changes to PR #903 to make this work. Or this PR would need to make changes after 903 to account for the fact that my 903 currently doesn't record if a config is a conv or a gemm.

If nothing else, we should coordinate our changes.

I also don't see the argument for landing this, unless we're planning to ship the tuning DBs I'll be generating and having MIGraphX use them.

jungpark-mlir:
I'll remove the operation type, it can be still distinguished by -F option.

We have no plan to ship this DB but MIGraphX could have an option to read a DB when user provides it. I'll develop it on the separate branch and not likely to be released in 5.5.

krzysz00:
No, wait, hold on, I've already changed up #903 to add the operation type 

(also, `-t` being at the back in `bert-configs` shouldn't've happened, `-t` goes in front (before the trans options) if we're looping through configs)

krzysz00 (commented):
Ok, I actually have a separate reason not to land this. The problem format for convolutions should include the G dimension because this is _our_ tuning infrastructure and we support more flexibility in the G dimension than MIOpen.

What we may want to do is to establish our own seralization format and convert the MIOpen command lines into it.

krzysz00 (approved):
Seems reasonable to me, I don't see any reason not to land this in 5.5 just in case we can rig up a client.

PR #902, opened 2022-12-09
Rework Xdlops selection process

jerryyin:
This PR completely reworked the existing xdlops and its group selection process. All xdlops selection related information population are moved to static one time initialization. In the runtime, the xdlops selection will be a simple query into the encoded map. This reduced the complexity of the original selection function to O(1). Additionally, I have decoupled the Xdlops selection into two logical classes:
 - MfmaInsn class encoded all information necessary as a single xdlops instructions
 - MfmaInsnGroup class encoded all information for a group of instruction to be selected

It is easier to review by commits:
 - 1st commit: Moved `XdlopsCodeSelection.h` into header + cpp
 - 2nd commit: Implemented the `MfmaInsn` and `MfmaInsnGroup` class
 - 3rd commit: Renamed `XdlopsCodeSelection` into `MfmaInsnGroup`

After this PR, I'd like to follow up and make the IR encode selected results so this doesn't have to happen more than once in the entire lowering process.


mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
```diff
@@ -0,0 +1,124 @@
...
+#include <set>
+
+#include "mlir/Dialect/AMDGPU/AMDGPUDialect.h"
+#include "llvm/ADT/SmallString.h"
+#include "llvm/ADT/StringMap.h"
+
+namespace mlir {
+namespace rock {
+
+enum class MfmaTypeId { Fp32TyId = 0, Fp16TyId, Bf16TyId, I8TyId };
```
krzysz00: Nit: can we put a `: uint32_t` here?

mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
```diff
@@ -0,0 +1,124 @@
...
+
+struct MfmaInsnInfo {
+  MfmaTypeId type;
+  int64_t mfmaNonKDim;
+  int64_t k;
+  int64_t blocksMfma;
+};
+
+struct MfmaInsnAttr {
```
krzysz00, line 34: NIt: If this'll be called `attr`, can we perhaps have it generated by tablegen/be an actual attribute? Or alternatively can we go for a different name?

krzysz00 (commented):
Hm ... I don't see any big issues here, but I do have a question.

If we were OK with having these MFMA properties live in a `MLIRContext` (which we have around all the time), would it be reasonable to make them actual attributes and then to pass the attribute struct around? Or to create a MLIR attribute way up in, say, affix-tuning-params that then gets bounced along?

But yeah, having tables instead of a big switch statement is a solid idea.

Also, while we're here, can we preemptively provide for passing in the chipset name? I'd like to lay the groundwork for "this only works on MI-200+" decisions and the like.

mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
```diff
@@ -0,0 +1,124 @@
...
+  MfmaInsnAttr attr;
+
+public:
+  static FailureOr<MfmaInsn> select(StringRef mfmaInsn);
+  MfmaInsn(const MfmaInsnAttr &mfmaInsnAttr);
+
+  MfmaInsnAttr getAttr();
+  Type getArgType(Type elementType);
+  VectorType getRetType(Type elementType);
+  bool isCohereantWithK(int64_t kPack, int64_t kPerBlock);
```
giuseros: Nit: did you mean Coherent (`isCoherentWithK`)?

mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
```diff
@@ -0,0 +1,124 @@
...
+//===----------------------------------------------------------------------===//
+
+// This file implements code selection logic for Mfma instructions.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef MLIR_MFMA_INSN_GROUP_H
+#define MLIR_MFMA_INSN_GROUP_H
+
+#include <set>
```
giuseros: Any reason why not using `llvm::DenseSet` or (if possible) `llvm::SmallDenseSet`?

mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
```diff
@@ -0,0 +1,124 @@
...
+struct MfmaInsnGroupAttr {
+  SmallString<16> insn;
+  int64_t MRepeats;
+  int64_t NRepeats;
+  SmallVector<MFMAParams, 2> imms;
+};
+
+class MfmaInsnGroup {
+private:
+  static const std::map<MfmaInsnGroupSelectKey, MfmaInsnGroupAttr>
```
giuseros: Why not a `llvm::denseMap` or `llvm::SmallDenseMap`?

giuseros (commented):
In my still narrow understanding of mfma this LGTM (only minor nits). I have two broader questions (probably stemming from me not understanding the code :) )
a)  You mention `runtime`. What do you mean exactly? The `mfma` selection happens at compile time, right?
b) What was the selection time upper bound in the original PR?

mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
jerryyin: For sure

mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
jerryyin: Hmm so which would you recommend? I do plan on to make it carried with instruction in the lowering process in the next release. In that aspect I think attribute is accurate. So the short answer here is that I'll address it in the next PR (basically the points you have mentioned from your summary comments).

mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
jerryyin: Nice catch!

mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
jerryyin: I think this is a left over and I never ended using std::set at all. Will get rid of this include.

mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
jerryyin: Because I'm too lazy to do this :-)

In this `MfmaInsnGroupSelectKey` type, since it is a customized type, using `DenseMap` required me to define a number of functions to instantiate the densemap including:
 - `getEmptyKey()`
 - `getTombstoneKey()`
 - `isEqual()`

I figured it was too much effort compared with just implementing a customized less than operator from requirement of `std::map`. Indeed all the trouble comes from the `MfmaTypeId` enum that it doesn't know how to handle it. I wonder if you have any suggestions over how to simplify this to fit into `llvm::DesnseMap`?

mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
krzysz00: Yeah, I'm not going to hold this up over naming complaints. Though `MfmaInsnInfo` is a good alternate name for this database if we want to not make it a `Attribute` subclass

mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
krzysz00: I'll note that you can autogenerate enums (even if they don't have a corresponding attribute type) with tablegen, it'll generate all the llvm::DenseMap boilerplate for you.

jerryyin: @krzysz00 Good point. I think I get a sense of how to hand craft that boilerplate. I'll leave the struct to tablegen conversion in next PR.

krzysz00 (approved):
Having skimmed this, it seems reasonable to me and it doesn't make any future refactorings any harder.

mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
jerryyin: That name was taken from a previous struct that indicate the manually typed in information. So I'll keep the naming but leave the attribute to tablegen in next PR.

jerryyin:
@giuseros Just saw your comments and want to give a response here:
a) Yes, I used confusing wordings here. I meant to say this reduce the compiler's runtime, or compilation time.
b) Because the selection was done at compilation time, we didn't care enough about its bound. But since new selection scheme is a cheap DenseMap query, it sure is much faster than the original if comparisons.

PR #903, opened 2022-12-12
GEMM tuning in CI

krzysz00:
1. Fix rocblas-driver to accept and ignore a perf_config argument, as the previous PR broke rocblas benchmarking
2. Amend the tuning runner to
  - Print progress reports during tuning to prevent CI timeouts
  - Write the tuning results out to a tab-separated value (TSV) file so future scripts can look at them.
  - Write the full tuning results to a file on a --debug
3. Amend the perf runner script to:
  - Optionally load a tuning database
  - Look up each (arch, commandLine) pair in said database, if provided, and apply the relevant perf config
  - On --batch_all (renamed from --batch_both), load the tuning DB if provided. If it is, benchmark untuned MLIR, tuned MLIR, and the external driver and compute all the comparisons. The behavior is unchanged when no tuning DB is provided
4. Amend the gemm report utilities to also print things like Tuned vs. Untuned if available
5. Change the Jenkinsfile to
  - Generate and stash a per-architecture tuning DB weekly
  - Use that tuning DB when doing perf runs for GEMM


krzysz00:
Weekly run for testing, skipping parameter sweeps http://rocmhead.amd.com:8080/job/MLIR/job/mlir/job/PR-903/2/

Nightly run for testing, perf tests only, to catch backwards compatibility issues http://rocmhead.amd.com:8080/job/MLIR/job/mlir/job/PR-903/3/

krzysz00:
New weekly run http://rocmhead.amd.com:8080/job/MLIR/job/mlir/job/PR-903/5/

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -765,7 +776,9 @@ pipeline {
                                 sh 'date --utc +%Y-%m-%d > perf-run-date'
                                 // Run MLIR perf benchmarks
                                 sh 'python3 ./bin/perfRunner.py'
-                                sh 'python3 ./bin/perfRunner.py --op=gemm --batch_both --configs_file=../mlir/utils/performance/bert-configs'
+                                sh """python3 ./bin/perfRunner.py --op=gemm --batch_all \
+--configs_file=../mlir/utils/performance/bert-configs \
+--tuning_db=../mlir_tuning_${CHIP}.tsv"""
```
zhanglx13: I think it is better to use a path relative to ${WORKSPACE}.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -723,6 +730,10 @@ pipeline {
                                 selector: lastSuccessful(),\
                                 target: 'MIOpen/build/MIOpenUserDB'
                                 sh 'ls MIOpen/build/MIOpenUserDB'
+                                // Move GEMM tuning DB out of the way if it exists
+                                catchError {
+                                    sh "mv MIOpen/build/MIOpenUserDB/*.tsv ${WORKSPACE}"
```
zhanglx13, line 763: Why do you need to move the db out of the directory?

zhanglx13 (commented):
One more question out of curiosity: why the data base uses .tsv instead of .csv?

mlir/utils/jenkins/Jenkinsfile
krzysz00: Because we'll be `rm`-ing that directory to get untuned MIOpen numbers

mlir/utils/jenkins/Jenkinsfile
krzysz00: Good point

krzysz00:
As to why tsv, perf_config contains commas and I didn't feel bothered to deal with escaping

mlir/utils/jenkins/Jenkinsfile
zhanglx13: From the doc, `catchError` will mark the build as `failure`, which will trigger node reboot in our nightly CI's. Is this the expected behavior of Gemm tuning?

yiqian1:
LGTM. Do you have nightly run for Gemm perfTest using the tuning DB?

yiqian1 (approved):

krzysz00:
@yiqian1 I've got a nightly run using the old weekly, but it's tricky to get nightly runs using the tuning data from the same PR.

I've been running the commands by hand (caught a bug in the perf regression report that way) using just about the same setup we'd see in CI. Results to be posted once I've got lockhart data

krzysz00:
Ok, so, here's the perf numbers - the `mlir_vs_rocblas.html` files in here are what you want to look at 
[gemm-tuning.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10220717/gemm-tuning.zip)

The takeaways are as follows:
- Overall, tuning gives us about 1.1x perf improvement on gfx1030 and gfx90a (**the gfx90a number has caveats, see below**) and a 1.17x (so about 1.2x) improvement on gfx908
- On gfx90a, tuning took us from 0.88x slower than rocBLAS on average to a factor of 0.98x, aka "on par" !!! And this is with the weirdness.
- The impacts of tuning are most felt on gfx908, where we go from barely beating rocblas on average (due to some very slow configs) to, at worst, being a factor of 0.85 slower to rocblas and being over 2x faster at best
- On gfx1030, a lot of the configs aren't too impacted by tuning, and there are some cases where the tuned version is slower. Despite this, we're still pullinng closer to rocBLAS on those cards. The strong possibility of an underlying performance issue caused by how we create non-xdlops kernels remains (ex. what if we started allowing for blockSize=32?)
- On gfx90a, things get _weird_.  On average, with tuning, we're now within 0.98x of rocBLAS, as mentioned above. However, if you observe the Tuned/Untuned ratios for each config, you'll notice that they're rather extreme - we both see massive improvements and substantial slowdowns. This isn't how this is supposed to work, obviously. I talked with @giuseros and we have a few theories
  - Since the tuning DB came from a different gfx90a node than the perf runs (which were done on `x1000c1s4b1n0` / the one I call `lockhart6`), it's *possible* that the number of CUs in the chips is different, causing different configs to be the performant ones
  - The structure of the tuning loop could be causing weird variance (ex. maybe the GPU needs to cool off)
  - There's a bug in the tuning API, and it's not enumerating all configs
- To check the number of CUs theory, I'm re-running the tuning generation on lockhart6. Once I've done that, I'll both compare the tuning databases and I'll re-run the perf analysis.

(Note that # of CUs isn't the only possible factor here, could be stuff like memory bandwidth, but, suffice to say, it's quite possible that chipset version isn't a unique enough key for tuning databases)

mlir/utils/jenkins/Jenkinsfile
krzysz00: I though the point of `catchError` was to _not_ mark the build as a failure, so that we don't reboot if the weekly CI doesn't have tuning info

krzysz00:
Another possibility: the MI-200 numbers were tuned with the block size changes, but run without them. The merge commit above should let us test that.

krzysz00:
Updates
1) I'm kicking off another tuning run that has fp16 in it at http://rocmhead.amd.com:8080/job/MLIR/job/mlir/job/PR-903/13/
2) Re-tuning on lockhart-6 *still* hits the same anomalies! Also, the winning perf configs are different. See 
[gfx90a-same-machine-tuning.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10222305/gfx90a-same-machine-tuning.zip) for more

krzysz00:
I found it! The emit tuning space option for xdlops isn't emitting 256 for the first parameter, at the very least!

krzysz00:
New weekly, please no node issues so we can get numbers and land this http://rocmhead.amd.com:8080/job/MLIR/job/mlir/job/PR-903/15

mlir/utils/performance/tuningRunner.py
```diff
@@ -13,60 +14,106 @@
...
+                nanoSeconds = float(time)
+
+            config.setPerfConfig(perfConfig)
+            entry = config.tableEntry(nanoSeconds)
+            allData.append(entry)
+            theseTFlops = entry['TFlops']
+            if not np.isnan(theseTFlops) and theseTFlops > maxTFlops:
+                maxTFlops = theseTFlops
+                winningConfig = perfConfig
+        verifyNs = verifyKernelWithPerfConfig(winningConfig, config, paths, rocmlir_gen_flags, debug)
```
giuseros: Why validating the winning config? Could we add the option of skipping the validation (which is timing out occasionally for big configurations)? 

cc: @sjw36 @jungpark-mlir

giuseros (changes requested):

giuseros:
Hi @krzysz00 , 
I rebased #894 on top of this PR and ran the performance analysis on gfx90a and I am attaching the comparison results. My summary is more succinct than yours :) : 
- Good performance on fp32 (sometimes beating CK for very irregular configurations)
- Less good performance on fp16 (we are still beating CK for irregular configs, but we lose more badly on "regular" configurations)
Please, have a look and let me know what you think!

[ck_performance_results.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10246931/ck_performance_results.zip)

mlir/utils/performance/tuningRunner.py
krzysz00: I'm validating the winner because
1. That's how MIOpen does it and;
2. Validation of that sort has caught bugs in convolution before

_That_ being said, it might be better to validate this against heuristic mode and not against CPU.

giuseros: Would it be a bad idea to add a flag like `--validation={none,CPU,heuristic}` ? In this way we can disable it when doing internal prototyping work and set it to `CPU` or `heuristic` in CI to catch bugs

krzysz00: No, that's actually a pretty solid idea. I'll get on that. And I'll need to extend `rocmlir-gen` to do the right thing on `-pv_with_gpu`.

krzysz00: (Or to add another flag for whether `-pv_with_gpu` strips perf_config or not)

krzysz00:
Good news: the tuning server does indeed make things run faster. Bad news, http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-903/30/pipeline shows a verification failure on Navi21 that I wasn't quite managing to reproduce initially. I'll `skip-ci` this for now and deal with it after the holidays

krzysz00:
http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-903/8/pipeline/378 is a passing GEMM tuning with a weird unstash failure. @giuseros @yiqian1 @jungpark-mlir can we get a re-review?

jungpark-mlir (approved):
LGTM

giuseros (approved):
LGTM

giuseros:
Any update about this? I ask because I would like to rebase/merge #894 which is dependent on this PR

krzysz00:
http://rocmhead:8080/job/MLIR/job/mlir/job/PR-903/18/ is a good run of this

PR #904, opened 2022-12-12
Fixing pytorch build by ignore conda build path

jerryyin:
To be reverted when corresponding fix is pushed to pytorch


krzysz00 (approved):
From looking at the documentation, I'm not sure that this'll actually filter out the include directory? But I'm not going to hold up the commit since it can't hurt and might help

jerryyin:
@krzysz00 The way it works it to ignore the path in all the `find_*()` commands, causing the files to be systematically ignored.

It was designed for almost the use case we run into, [here](https://cmake.org/cmake/help/latest/variable/CMAKE_IGNORE_PATH.html). Refer to the quote below.
> Ignoring search locations can be useful in cross-compiling environments where some system directories contain incompatible but possibly linkable libraries. For example, on cross-compiled cluster environments, this allows a user to ignore directories containing libraries meant for the front-end machine.

PR #905, opened 2022-12-14
Bugfixes found in the process of creating PR #903

krzysz00:
1. Add a perf_config option that's ignored to the rocblas driver so that rocblas benchmarks actually run
2. Change the xdlops tuning ranges to match MIOpen, including the different KPerBlock for int32, with the exception that we now also include kPack=1 in the xdlops range to prevent another "tuning made things slower" situation.


jungpark-mlir (approved):
LGTM

yiqian1 (approved):

PR #917, opened 2022-12-15
[hotfix] Disable not unrolling load/store loops

krzysz00:
This commit temporarily disables forceUnroll = false as a valid parameter setting for xdlops tuning, because in some cases it causes register allocator failures (we think).

Future work should be happening now to loosen this restriction.


jerryyin (approved):
For future reference, this forceunroll attribute seems to be causing non-trivial breakages around tuning. Refer to: http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir-weekly/detail/mlir-weekly/118/pipeline/291. Nearly every config is breaking on running out of registers. We'd need to run weekly when we re-enable this feature.

PR #919, opened 2022-12-16
[HOTFIX] set the forceUnroll to 1 in perf-config test

manupak:
This commit changes forceUnroll to align
with the current hotfix'd state.


zhanglx13 (approved):
Thanks for the quick fix. Since Krzysztof is not here today, I'll push the approve button.

PR #921, opened 2022-12-16
[Docker] Fix build-dockerfile.sh

manupak:
* Fixes the login return check
* replace pkill with removal of containers
* dummy change to dockerfile to trigger a build

I've already ran the changes to see if there are working by enabling the branch : http://ml-ci.amd.com:21096/job/MLIR/job/rocm-docker/819/


krzysz00 (approved):
LGTM! Although, since you're here, you could also make changes to the script trigger a rebuild?

manupak:
ok, I made it case insensitive

PR #922, opened 2022-12-16
MLIR#760: Fix bug in FoldTranspose for non-trivial expand_shape ops

sjw36:
MLIR#749: Added torch-mlir fusion tests
MLIR#766: tosa-to-rock added support for tosa.transpose int64 perms

Pipeline changes
- Added	tosa-to-tensor conversion pass (tosa.slice ops)
- minor	re-ordering

https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/760 https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/749 https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/766

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/FoldTranspose.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/FoldTranspose.cpp
@@ -58,52 +58,31 @@ struct InlineViewLikeOperandsLinalgRewritePattern
 // post-reassociation --> then, produce the set of resultant AffineExprs after
 // the reassociation.
 static void getDelinearizedAffineExpr(
-    ArrayRef<int64_t> originalStridesDimsBeingReassociated,
-    ArrayRef<int64_t> postReassociatedDimSizes, Builder &b,
-    unsigned int position, SmallVectorImpl<AffineExpr> &res) {
+    int64_t originalSize, ArrayRef<int64_t> postReassociatedDimSizes,
+    Builder &b, unsigned int position, SmallVectorImpl<AffineExpr> &res) {
   AffineExpr resultExpr = b.getAffineDimExpr(position);
-  int64_t rank = originalStridesDimsBeingReassociated.size();
-  // If the rank is 1, expand or collapse shapes will just
-  // pass-through the dimensions.
-  if (rank == 1) {
-    res[0] = resultExpr;
-    return;
-  }
-  for (int i = 0; i < rank; i++) {
+  int64_t rank = postReassociatedDimSizes.size();
+  for (int i = rank - 1; i >= 0; --i) {
     // If the postReassociatedDimSize is 1 and the rank is non-zero,
     // could only mean it is being broadcasted. Hence,
     // putting zero.
-    if (postReassociatedDimSizes[i] == 1) {
+    int64_t dim = postReassociatedDimSizes[i];
+    if (dim == 1) {
       res[i] = resultExpr * 0;
     } else {
       // Recording the vector offsets here.
       res[i] = resultExpr;
       // There is no point of putting a modulo if the size
       // is equivalent to that.
-      if (i - 1 >= 0 && postReassociatedDimSizes[i] !=
-                            originalStridesDimsBeingReassociated[i - 1]) {
-        res[i] = res[i] % originalStridesDimsBeingReassociated[i - 1];
-      }
-
-      if (postReassociatedDimSizes[i] >
-          originalStridesDimsBeingReassociated[i]) {
-        // We only need the floorDiv if the dimSize
-        // is larger than the stride
-        res[i] = res[i].floorDiv(originalStridesDimsBeingReassociated[i]);
-      } else if (postReassociatedDimSizes[i] <
-                 originalStridesDimsBeingReassociated[i]) {
-        // if the shape is smaller than the stride
-        // expr might as well be zero.
-        res[i] = res[i] * 0;
+      if (dim < originalSize) {
+        res[i] = res[i] % dim;
       }
 
       // The resultExpr has to propagated anyway for
       // other dimensions where the recording in the above
       // will do the neccesary checks to remove the modulo
-      if (i - 1 >= 0) {
-        resultExpr = resultExpr % originalStridesDimsBeingReassociated[i - 1];
-      }
-      resultExpr = resultExpr.floorDiv(originalStridesDimsBeingReassociated[i]);
+      resultExpr = resultExpr.floorDiv(dim);
+      originalSize /= dim;
     }
   }
   return;
@@ -111,6 +90,9 @@ static void getDelinearizedAffineExpr(
 
 // This function will create a affine map that represent the mapping
 // from higher rank memref type to lower rank memref type.
+// Eg:
+//   IN:  <12x12x32xf32> -> <12x384xf32>
+//   OUT: (d0, d1) -> (d0, d1 / 32, d1 % 32)
 static AffineMap createHigherToLowerRankViewAffineMap(
     PatternRewriter &rewriter,
     ArrayRef<ReassociationIndices> reassociationIndices,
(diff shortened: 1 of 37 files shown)


mlir/lib/Dialect/Rock/Transforms/FoldTranspose.cpp
```diff
@@ -376,7 +397,11 @@ struct FoldRockOutputTransforms : OpRewritePattern<linalg::GenericOp> {
         continue;
 
       AffineMap inpIdxMap = laGeneric.getTiedIndexingMap(&operand);
+      // (d0, d1, d2) -> (0, d1, d0 * 32 + d2)
+      // (d0, d1, d2) -> ??? (d0 % 1, d1, d0 / 32 + d2 % 32)
       auto invertInpIdxMap = inversePermutation(inpIdxMap);
+      if (!invertInpIdxMap)
+        continue;
```
manupak: Why is this needed ? (I also remember adding those special cases to zeroing out or dim size checks before module divs are added to keep it invertible).

Looking at the tests, I see you managed invert them but also then wonder if it would ever encounter this case.
I'd rather fail the pass here unless if there is a way to deal with them later which I doubt.

manupak: (Im more than happy to hear if the examples mentioned above could be inverted  by the existing affine infra :) )

mlir/lib/Dialect/Rock/Transforms/FoldTranspose.cpp
```diff
@@ -58,59 +58,41 @@ struct InlineViewLikeOperandsLinalgRewritePattern
 // post-reassociation --> then, produce the set of resultant AffineExprs after
 // the reassociation.
 static void getDelinearizedAffineExpr(
-    ArrayRef<int64_t> originalStridesDimsBeingReassociated,
```
manupak: Just a note on all the changes. This what I started with but ended up adding what's being removed here to keep the maps invertible IIRC. However, if you manage to get them inverted -- it is more than welcome.

mlir/test/xmir/e2e/bert-torch-tosa/bert_part_0.torch-tosa.mlir
```diff
@@ -0,0 +1,15 @@
...
+// CHECK-DISABLED: RMS = {{.*}}e-08
+// COM: CHECK: [1 1 1]
+// CHECK: bert_part_0__part_0
+module {
+  func.func @bert_part_0(%arg0: tensor<1x12x384xf32> {func.read_access}, %arg1: tensor<384x384xf32> {func.read_access}, %arg2: tensor<1x1x384xf32> {func.read_access}) -> (tensor<1x12x12x32xf32> {func.write_access}) {
+      %0 = "tosa.reshape"(%arg1) {new_shape = [1, 384, 384]} : (tensor<384x384xf32>) -> tensor<1x384x384xf32>
+      %1 = "tosa.matmul"(%arg0, %0) : (tensor<1x12x384xf32>, tensor<1x384x384xf32>) -> tensor<1x12x384xf32>
+      %2 = "tosa.add"(%1, %arg2) : (tensor<1x12x384xf32>, tensor<1x1x384xf32>) -> tensor<1x12x384xf32>
+      %3 = "tosa.reshape"(%2) {new_shape = [1, 12, 12, 32]} : (tensor<1x12x384xf32>) -> tensor<1x12x12x32xf32>
```
manupak, line 12: Can you make another narrower test of this case just for FoldTranspose pass ?

I think it is the most important functionality this change enables and I think we would need to assert this transform happens within the pass itself.

manupak (commented):
Thanks @sjw36 it is what I had in mind and more...

I was just positively surprised to know that affine infra is able invert the complex modulo expressions. If it works, I have no complaints.

However, I ve highlighted a test that I think we should enable as a test for the pass to make sure it is handled within the pass.

mlir/test/xmir/e2e/bert-torch-tosa/bert_part_0.torch-tosa.mlir
sjw36: This one does not work yet (see DISABLE above). See [762](https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/762).

This would be the case where we couldn't align the linalg.generic out with the in from the matmul.. Will work on that one next.

mlir/lib/Dialect/Rock/Transforms/FoldTranspose.cpp
sjw36: Yes, AffineMap inversion is very limited. So I dropped it and here do explicit low-to-high and high-to-low.

mlir/lib/Dialect/Rock/Transforms/FoldTranspose.cpp
sjw36: I added the continue to avoid the crash but yes, should be failure. It does not support the example listed in the comment (I think also 762).

mlir/lib/Dialect/Rock/Transforms/FoldTranspose.cpp
```diff
@@ -58,59 +58,41 @@ struct InlineViewLikeOperandsLinalgRewritePattern
...
-    return;
-  }
-  for (int i = 0; i < rank; i++) {
+  int64_t rank = postReassociatedDimSizes.size();
+  for (int i = rank - 1; i >= 0; --i) {
     // If the postReassociatedDimSize is 1 and the rank is non-zero,
     // could only mean it is being broadcasted. Hence,
     // putting zero.
-    if (postReassociatedDimSizes[i] == 1) {
+    int64_t dim = postReassociatedDimSizes[i];
```
krzysz00: Nit. `dim` feels like a dimension index. Maybe `dimSize` or `reassocSize` or something

mlir/test/xmir/e2e/bert-torch-tosa/bert_part_0.torch-tosa.mlir
krzysz00: Yeah, I'm on team "narrower tests" here. `--mlir-print-ir-after-all` may be helpful

krzysz00: Or at least having a separate test for the bug over in `Dialect/Rock`

krzysz00 (commented):
I come bearing naming nitpicks. It sounds like the PR is still not _quite_ working, going off the disabled tests?

sjw36:
> It sounds like the PR is still not _quite_ working, going off the disabled tests?

It is working for some cases, and I have follow-up tickets for the remaining issues. I will create a couple simple tests for the pass.

https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/762
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/763
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/764
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/765

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -675,16 +675,67 @@ Value mlir::rock::insertTransposeAndBroadcastTransforms(
     LLVM_DEBUG(llvm::dbgs() << "Reached makeBroadcast with map " << inpIdxMap
                             << " and diff = " << diff << "\n");
 
+    if (diff < 0) {
```
manupak, line 678: I think this is the only required change for the failing case. i.e. more dims in output.

manupak (commented):
In the spirit of getting the working bit landed, I would propose to to break this PR just to contain this and the test cases that otherwise fail w/o more dims in output case.

I think we can do the rest of the things in a follow up where we handle non-trivial expansion/collapse with modulo arithmetic. WDYT ?

PS : Maybe a seperate PR to add tests that are not related to any of these.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
sjw36: Yes, the FoldTranspose changes were a partial fix for a follow-up issue, but I am rethinking the solution. I have reverted FoldTranspose.cpp changes.

sjw36:
Everything should be addressed @manupak !

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -675,16 +675,67 @@ Value mlir::rock::insertTransposeAndBroadcastTransforms(
...
+          newInpIdxMap.setResult(newIdx++, resultExpr);
+          merges.push_back(mergeDims);
+          mergeDims.clear();
+        }
+      }
+      if (mergeDims.size())
+        merges.back().append(mergeDims);
+
+      TopDownTMBuilder collapseTransform(b, outShape, loc);
+      for (auto idxAndMerge : llvm::enumerate(merges)) {
```
manupak, line 703: nit : this feels like a good utility for transform map utils to create reassociation transforms that might be useful elsewhere w/o having to deal with names for axes.

manupak (approved):
LGTM!

as non-blocking further suggestions : 

* If we are adding the failing tests, I would prefer them to be XFAILs -- so they are explicit in our tests as needs fixing or we can add them once we fix them.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
manupak: So the suggestion was to lift this as a function here...

PR #928, opened 2022-12-29
Define the LLVM version for MLIR

### LINKED ISSUE #927: Undefined soname for MLIR libraries
Hi! I'm packaging rocMLIR for Arch Linux. The installed libraries have a generic "VERSION" suffix that makes it impossible to automatically check for ABI changes in the package or packages that depend on it. The problem is that `add_mlir_library` calls `llvm_add_library` that uses `LLVM_VERSION_MAJOR` and `LLVM_VERSION_SUFFIX` to set the soversion via `set_target_properties`. The file that defines these variables (in the external llvm-project) is included after the mlir folder. Thus the LLVM version is undefined and cmake defaults to VERSION. A possible solution is to copy the definitions of LLVM_VERSION_* to mlir/CMakeLists.txt.

tpkessler:
Resolves #927


mlir/CMakeLists.txt
```diff
@@ -1,3 +1,16 @@
+if(NOT DEFINED LLVM_VERSION_MAJOR)
+  set(LLVM_VERSION_MAJOR 16)
```
krzysz00, line 2: Since I don't have the linux packaging context, a question: Does this need to track the LLVM_VERSION field in `external/llvm-project/`? If not, would it make more sense for these values to track the ROCm version instead? That is, instead of `16.0.0`, would `5.5.0` (or, for the development branch, `6.0.0-git`, say) be better?

krzysz00: Or if we were going for ROCm versions, since each "minor" version is potentially ABI-breaking, would `55.0.0`, `60.0.0` and so on be better?

krzysz00 (commented):
I have a question but no fundamental objection to merging this. We'll just need to remember to bump this in the release process.

mlir/CMakeLists.txt
krzysz00: @tpkessler Your thoughts?

tpkessler: See my comment in the linked issue.

mlir/CMakeLists.txt
```diff
@@ -1,3 +1,16 @@
...
+endif()
+if(NOT DEFINED LLVM_VERSION_MINOR)
+  set(LLVM_VERSION_MINOR 0)
+endif()
+if(NOT DEFINED LLVM_VERSION_PATCH)
+  set(LLVM_VERSION_PATCH 0)
```
krzysz00, line 8: Ok, actual request, having stared at this all some. We should define the patch version to, for ROCm 5.5, `50500`, to match the amdgpu drivers' conventions for these things. That is, `[0]50500` for `5.5.0`)

krzysz00, line 8: Ok, actual request, having stared at this all some. We should define the patch version to, for ROCm 5.5, `50500`, to match the amdgpu drivers' conventions for these things. That is, `[0]50500` for `5.5.0`)

krzysz00:
This is now skip-ci pending confirmation that the patch actually works.

krzysz00:
I think this got fixed by our recent Windows build improvements, can you take another look?

jerryyin:
Please re-open if you are still interested

PR #929, opened 2023-01-03
Bump rocMLIR supported minor version up ROCm5.5

jerryyin:


zhanglx13 (approved):
Is there somewhere like a release note that we can document the changes in this new version?

jerryyin:
@zhanglx13 Good point. As far as I know, we can either added a text file that document this or use github release (and update when there's a new release). I don't think this is a requirement. I'd say if we want to do this we should ask @mvermeulen's opinion. The document for 1.0.0 is going to be tricky because one need to summarize our product entirely.

krzysz00 (approved):

mvermeulen:
For what it is worth, using the github release mechanism seems the most straightforward to me and using standard mechanism; agree will have to figure out what to describe as initial instance.

PR #931, opened 2023-01-06
Enabling 128x128 wave size

jerryyin:
Tested and can verify for 128x128 wave size. Sample config:

> --operation conv2d -t f32 --arch gfx90a:sramecc+:xnack- --fil_layout kcyx --in_layout nchw --out_layout nkhw --batchsize 8192 --in_channels 8192 --in_h 1 --in_w 1 --out_channels 8192 --fil_h 1 --fil_w 1 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h_l 0 --padding_h_r 0 --padding_w_l 0 --padding_w_r 0 -mfma=on -dot=on -atomic_add=on --perf_config 256,256,4,128,128,2,1,1

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
+++ b/mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
@@ -125,6 +125,7 @@ public:
                 const MfmaInsnGroupAttr &groupAttr);
   int64_t getMRepeats();
   int64_t getNRepeats();
+  static int64_t getLenPerMfmaGroup(int64_t lenPerWave);
   SmallVector<mlir::rock::MFMAParams, 2> getImms();
 
   MfmaInsnAttr getInsnAttr();
--- a/mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
+++ b/mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
@@ -396,8 +396,8 @@ FailureOr<MfmaInsnGroup> MfmaInsnGroup::select(mlir::Type elementType,
                           << "nPerWave: " << nPerWave << "\n");
 
   // Use 64x64 as base unit in large waves
-  int64_t mPerMfmaGroup = (mPerWave > 64) ? 64 : mPerWave;
-  int64_t nPerMfmaGroup = (nPerWave > 64) ? 64 : nPerWave;
+  int64_t mPerMfmaGroup = getLenPerMfmaGroup(mPerWave);
+  int64_t nPerMfmaGroup = getLenPerMfmaGroup(nPerWave);
 
   MfmaInsnGroupSelectKey key = {convertTypeToId(elementType), mPerMfmaGroup,
                                 nPerMfmaGroup};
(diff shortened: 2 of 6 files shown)


krzysz00 (commented):
Question: does this mean that we're no longer using the broadcast flags to handle larger MFMAs?

Can we get some configs that use the 128,64 and 64,128 perfConfigs and check their perf with this change?

Should we update the heuristic perf config list to include the 128x128 case?

(Code-wise, lgtm, but I have context questions so this isn't an approve)

jerryyin:
@krzysz00 

> Question: does this mean that we're no longer using the broadcast flags to handle larger MFMAs?

This is not the work I'm working together with Giuseppe. This just aim to enable larger wave size (such that large config can potentially benefit). I'd push this PR for CI to help me with the "how much better" question.

> Can we get some configs that use the 128,64 and 64,128 perfConfigs and check their perf with this change?

Good point and on my list for next step to do.

> Should we update the heuristic perf config list to include the 128x128 case?

It's already included, it's just that previously we always reject those. Now we let them through the flow. And in fact any size larger than 64 is a fair game now. If we extend the tuning range, they should be easily enabled too. We don't have to hard code the larger wave size as we used to.

krzysz00:
I'll note in the PR you explicitly got rid of the "128x64 uses several sets of broadcast flags" code. Which is why I asked.

MRepeat isn't sufficient, I think, you need to populate out the relevant broadcast flags.

jerryyin:
> MRepeat isn't sufficient, I think, you need to populate out the relevant broadcast flags.

Yes broadcast flags are necessary but if you cross compare between 128x64 or 64x128 for the existing definitions, you'll find those are just repetitive patterns based on 64x64. Now instead of manually defining those repetitive patterns, I populate those based on how many 64x64 does each wave has. It will reuse the broacast pattern flags based on the 64x64 definition.

krzysz00:
Huh, yeah, yeah they are. That answers that question

jerryyin:
@krzysz00 I can hardly find a config that can satisfy the applicability of between both 64, 128 and 128, 128. There are typical only one of them of valid dimension. However, this PR can indeed yield some concrete perf benefit (when the gemm is large). For example, the new perf report of gfx90a a wrw config for resnet50 becomes 14 times faster, refer to [here](http://rocmhead:8080/job/MLIR/job/mlir/job/PR-931/6/Performance_20report_20for_20gfx90a/):

<html>
<body>


Direction | DataType | Layout | C | H | W | K | Y | X | StrideH | StrideW | PaddingH | PaddingW
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
wrw | f32 | nchw | 64 | 56 | 56 | 64 | 3 | 3 | 1 | 1 | 1 | 1 


</body>
</html>

This is quite convincing as other configs stays strictly within 3% of the perf difference.

I think the summary here is that this change will benefit larger gemms. The cost of the performance is relatively cheap: try an additional entry where we are rejecting in the past.

krzysz00 (commented):
Ok, so, just to check, does this PR remove the 128x64 and 64x128 configs?

krzysz00:
Because those cases do have nonuniform broadcast parameters (`{1, 0, none}` and `{1, 1, none}`) and it looks like the 64x64 case just has `{0, 0, none}`

jerryyin:
> Ok, so, just to check, does this PR remove the 128x64 and 64x128 configs?

No this PR does not remove 64x128 and 128x64. It removed the hard coded values but those dimensions will continue to be supported. They will be supported based on 64x64.

> Because those cases do have nonuniform broadcast parameters ({1, 0, none} and {1, 1, none}) and it looks like the 64x64 case just has {0, 0, none}

You probably have misread it, but the 64x64 wave size does have the two immediates  you mentioned about: https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/f184a6132c07efb7747a0d9f19eb12e052f5c63a/mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp#L177-L181

krzysz00 (approved):
Having been pointed at the thing I missed, seems reasonable, approved.

PR #932, opened 2023-01-06
[DO NOT MERGE] Enable reduction for 64x64 wave

jerryyin:
This PR doesn't seek for merge, but just a means of communication. FYI, this enables the below:

> ninja rocmlir-driver && ./bin/rocmlir-gen -ph -operation gemm -t f32 --arch gfx90a:sramecc+:xnack- -g 64 -m 1024 -k 1024 -n 384 --transA --transB --kernel-repeats 5 --perf_config=128,128,4,64,64,8,1,1 -t f16 -pv | ./bin/rocmlir-driver -c | ../mlir/utils/widgets/rocm-run

However, the existing validation won't pass because (for existing implementation) reduction case doesn't support m/n repeat > 1.

@giuseros I think this will enable only the performance comparisons but additional work is needed to add support to reduction + repeats.


mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
```diff
@@ -237,10 +237,13 @@ auto getMfmaInsnGroupAttrMap = []() {
...
                       {1, 1, MFMAPermB::none}}}},
                    {{MfmaTypeId::Fp16TyId, 64, 64},
-                    {ROCDL::mfma_f32_32x32x4f16::getOperationName(),
-                     1,
-                     1,
-                     {{1, 0, MFMAPermB::none}, {1, 1, MFMAPermB::none}}}},
+                    {ROCDL::mfma_f32_32x32x8f16::getOperationName(),
+                     2,
+                     2,
+                     {{0, 0, MFMAPermB::none},
```
giuseros, line 243: This is what I was missing! Thanks Jerry, can I ask you why this needs to be like this?

giuseros:
Hi @jerryyin , I tried this out. It doesn't error anymore (thanks for the hunch) but we are still emitting too many `mfma`s. 
```
grep mfma mlir.s | wc -l 
32 
grep mfma ck.s | wc -l 
8
```
In theory each block is 256 threads, i.e., 4 wavefronts. Each wavefront computes `(2*32)x(2*32)=64x64` tiles. This means that 4 `mfma`s are needed per block to compute 128x128 tiles. Software pipelining will push that up to 8. Why do you think we get to 32? Is it because we are reducing in blocks of 4 instead of 8 (but even then I would expect 16 `mfma`s and not 32)?  
**EDIT**
Thinking more about this, is it possible this is due to unrolls? I.e., we are unrolling the `mRepeats` and `nRepeats` (so we go from `8 mfma` to `2x2x8 mfma`)?

krzysz00:
It could be `forceUnroll`, yeah. Or the loop tail. I'd go looking at the intermediates (after `-rock-threadwise-gemm-lowering`) to see how many `amdgpu.mfma` operations we have there.

giuseros:
Hi @krzysz00 , thanks for the comment! I was discussing this with Jerry, and what I see is that until we execute `llc` I see 8 `mfma`s. When I run `llc -mcpu=gfx90a` I see that the number of mfma increases from 8 to 32.

krzysz00:
Ah. Try `llc -O3` also, just out of curiosity. And make sure your `opt` is also O3.

(But yeah, the backend could be unrolling those loops)

giuseros:
I am already passing `-passes='default<03>,strip` to `opt` . I tried `-O3` to `llc` but unfortunately didn't make any difference. My doubt is if unrolling can hit performance so badly.

giuseros:
The plan with @jerryyin is to first make it validate (so we can be sure it is doing the right thing), and then trying to understand if the unrolling is the only difference with CK or if there are other differences compared with their (default) algorithm

jerryyin:
Yes, though my gut tell me that the xdlops selection difference here is probably a trivial factor and doesn't impact the performance as a major factor as we assumed. It is useful to have it enabled, but we'll have to look in a different direction on why we are 60% of the speed of CK.

krzysz00:
My gut is telling me that, given that we've got cases where we're almost even with CK and cases where we're way worse and that xdlops selection seems to dictate which case is which, this is a really important change.

jerryyin:
@giuseros 

> In theory each block is 256 threads, i.e., 4 wavefronts. Each wavefront computes (2*32)x(2*32)=64x64 tiles. This means that 4 mfmas are needed per block to compute 128x128 tiles. Software pipelining will push that up to 8. Why do you think we get to 32? Is it because we are reducing in blocks of 4 instead of 8 (but even then I would expect 16 mfmas and not 32)?

Upon further reviewing of the xdlops logic, you did not take into k dimension in the above reasoning. 

For this perf_config: 128,128,4,64,64,8,1,1. 
 1. block level tile: A[128, 4], B[4, 128]
     - Since kPack = 8, for A and B in K direction we actually have 4 * 8 = 32 elements
 2. This block level tile get split to be executed by 4 waves, each A and B [64, 32] elements
 3. Now, for this mWave = 64 and nWave = 64 wave level tile, each is executed by 2 x 2 = 4 xdlops with reduction on M/N dimension, each xdlops group handling a [32, 32] elements tile
 4. 1 xdlops can handle [32, 8] x [8, 32] tile, it took 4 iterations to loop through k dimension and finish execution for [32, 32] tile.

Aggregating all computation together:
 - Total number of xdlops/wave = 4(xdlops group) x 4(iterations) = 16 total xdlops
 - Since we have main loop + tail, in total we have 16 x 2 = 32 xdlops instructions

> Thinking more about this, is it possible this is due to unrolls? I.e., we are unrolling the mRepeats and nRepeats (so we go from 8 mfma to 2x2x8 mfma)?

Yes, it is due to unroll. If we unroll less aggresively you will see the inner loop (4 iterations), and get 8 total xdlops from IR.

@krzysz00 
> My gut is telling me that, given that we've got cases where we're almost even with CK and cases where we're way worse and that xdlops selection seems to dictate which case is which, this is a really important change.

It's possible. I think this conversation get diverged in two place. Refer to here: https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/768#issuecomment-1375488723. Since according to Giuseppe's existing experiment on a reduction with right number of xdlops, the performance didn't change that much, it make me to think it isn't use of reduction xdlops alone caused the perf gap.

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
jerryyin: This is because 64 x 64 tile is too large and has to split into 2 repeats in M dimension and 2 repeats in N dimension. For the total of 2 x 2 = 4 xdlops instructions, each need to come with a `{0, 0, none}` flag.

giuseros:
Closing this as a duplicate of  #931 (merged) + #941 (in review)

PR #933, opened 2023-01-10
[MIGraphXToTosa] Enable broadcast input to matmul when possible.

jungpark-mlir:
support converting
migraphx.dot(A, broadcast(B)) to tosa, where it only support implicit  broadcast for elementwise operators.

This only support the case when the batch dimension of B is broadcast,
e.g.
A : [g, m, k]
B : [1,  k, n]
broadcast(B) : [g, k, n]

by reshaping A -> [1, gxm, k]

### DIFF
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -317,7 +317,7 @@ public:
 
     // check batch dimension. Tosa matmul only allow a single dimension for it,
     // add reshape ops to flatten and restore the original dimension.
-    SmallVector<int64_t, 5> orgOutDims = outputTy.getShape();
+    SmallVector<int64_t, 5> orgOutDims(outputTy.getShape());
     RankedTensorType newOutType = RankedTensorType::get(orgOutDims, elementTy);
     size_t outRank = orgOutDims.size();
 
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -339,12 +339,12 @@ public:
 
       if (batchSizeA != batchSizeB) {
         // support when batchB dimension is broadcast
-        if (RankB == 3 && orgDimsB[0] == 1) {
+        if (rankB == 3 && orgDimsB[0] == 1) {
           // modify [g, m, k, n] to [1, g*m, k, n]
           batchSizeA = 1;
-          batchSize = 1;
+          batchSizeC = 1;
           orgDimsA[rankA - 2] *= batchSizeA;
-          orgOutDims[outRank - 2] *= batchSize;
+          orgOutDims[outRank - 2] *= batchSizeC;
         } else {
           // currently not supporting the other case, broadcast A could be
           // supported with an additional transpose.
@@ -355,7 +355,7 @@ public:
                              orgDimsA[outRank - 1]};
       int64_t newDimsB[3] = {batchSizeB, orgDimsB[outRank - 2],
                              orgDimsB[outRank - 1]};
-      int64_t newDimsOut[3] = {batchSize, orgOutDims[outRank - 2],
+      int64_t newDimsOut[3] = {batchSizeC, orgOutDims[outRank - 2],
                                orgOutDims[outRank - 1]};
       RankedTensorType newAType = RankedTensorType::get(newDimsA, elementTy);
       RankedTensorType newBType = RankedTensorType::get(newDimsB, elementTy);
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -322,8 +322,8 @@ public:
     size_t outRank = orgOutDims.size();
 
     if (outRank != 3) { // A, B, Out have the same rank. rank=2 assumes batch=1
-      ArrayRef<int64_t> orgDimsA = in_A.getType().cast<ShapedType>().getShape();
-      ArrayRef<int64_t> orgDimsB = in_B.getType().cast<ShapedType>().getShape();
+      SmallVector orgDimsA(in_A.getType().cast<ShapedType>().getShape());
+      SmallVector orgDimsB(in_B.getType().cast<ShapedType>().getShape());
       size_t rankA = orgDimsA.size();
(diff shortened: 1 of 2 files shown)


mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -317,23 +317,51 @@ class DotConverter final : public OpConversionPattern<migraphx::DotOp> {
 
     // check batch dimension. Tosa matmul only allow a single dimension for it,
     // add reshape ops to flatten and restore the original dimension.
-    ArrayRef<int64_t> orgOutDims = outputTy.getShape();
+    SmallVector<int64_t, 5> orgOutDims(outputTy.getShape());
```
manupak: Out of curiosity, why cant we get away with a ArrayRef here (as before)?

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -317,23 +317,51 @@ class DotConverter final : public OpConversionPattern<migraphx::DotOp> {
...
+        if (rankB == 3 && orgDimsB[0] == 1) {
+          // modify [g, m, k, n] to [1, g*m, k, n]
+          newDimsA[0] = 1;
+          newDimsA[1] *= batchSizeA;
+          newDimsOut[0] = 1;
+          newDimsOut[1] *= batchSizeC;
+        } else {
+          // currently not supporting the other case, broadcast A could be
+          // supported with an additional transpose.
+          return failure();
```
manupak: We should use op.emitError here as this is a unsupported case where the compilation should fail rather than a match failure.

mlir/test/Conversion/MIGraphXToTosa/mixr-to-tosa-ops.mlir
```diff
@@ -22,6 +22,17 @@ module  {
      return %0 : tensor<32x64xf32>
   }
 
+  // CHECK-LABEL: func.func @matmul_broadcast
+  func.func @matmul_broadcast(%arg0: tensor<64x64x2304xf16>, %arg1: tensor<64x64x768xf16>, %arg2: tensor<1x768x2304xf16>) -> tensor<64x64x2304xf16> attributes {arch = "gfx90a:sramecc+:xnack-", kernel = "mixr"} {
+    %0 = migraphx.multibroadcast(%arg2) {out_dyn_dims = [], out_lens = [64, 768, 2304]} : (tensor<1x768x2304xf16>) -> tensor<64x768x2304xf16>
+    // CHECK: [[RESHAPE1:%.+]] = tosa.reshape
```
manupak: I think it would be better have the shapes in the test to understand this change for future us...

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -317,23 +317,51 @@ class DotConverter final : public OpConversionPattern<migraphx::DotOp> {
...
+    SmallVector<int64_t, 5> orgDimsA(
+        in_A.getType().cast<ShapedType>().getShape());
+    SmallVector<int64_t, 5> orgDimsB(
+        in_B.getType().cast<ShapedType>().getShape());
+    size_t rankA = orgDimsA.size();
+    size_t rankB = orgDimsB.size();
+
+    // A, B, Out have the same rank. rank=2 assumes batch=1.
+    // Here handling special cases.
+    if (outRank != 3 || rankA != rankB ||
```
manupak, line 330: Is this meant to work for outRank > 4 ?

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -317,23 +317,51 @@ class DotConverter final : public OpConversionPattern<migraphx::DotOp> {
...
+
+      int64_t newDimsA[3] = {batchSizeA, orgDimsA[outRank - 2],
                              orgDimsA[outRank - 1]};
-      int64_t newDimsB[3] = {batchSize, orgDimsB[outRank - 2],
+      int64_t newDimsB[3] = {batchSizeB, orgDimsB[outRank - 2],
                              orgDimsB[outRank - 1]};
-      int64_t newDimsOut[3] = {batchSize, orgOutDims[outRank - 2],
+      int64_t newDimsOut[3] = {batchSizeC, orgOutDims[outRank - 2],
                                orgOutDims[outRank - 1]};
+      if (batchSizeA != batchSizeB) {
```
manupak, line 349: Would it possible add a test to check this case ?
(If I am not mistaken, the existing test checks for same batchsize)

manupak (commented):
few questions for my understanding and minor comments...

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
manupak: ah ignore... I just saw multibroadcast doing that.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jungpark-mlir: Right.
In the earlier version of this change tried to modify this array. Now it's not the case and I should revert this back, thanks.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jungpark-mlir: Yes, it works for that cases as well. i.e., it first flatten the batch dimension and potentially handle the broadcast when outRank > 4

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jungpark-mlir: done

mlir/test/Conversion/MIGraphXToTosa/mixr-to-tosa-ops.mlir
jungpark-mlir: done

manupak (approved):
LGTM.

Just an optional suggestion, if it was meant to work outRank > 4, a test would be nice to make sure otherwise erroring out would be better rather than breaking down somewhere later.. but however, I doubt that would be the case here, but... :)

jungpark-mlir:
Yeah, turns out the code didn't allow rank>4 actually. fixing now.

PR #934, opened 2023-01-10
Amend API to use tuned data

jungpark-mlir:
Found some change required while implementing the client side code.
- when the tuned data entry comes from file, it only has serialized form of the problem, not the original problem.
- modify the problem  serialization format to match with the string currently generated by the performance script.


krzysz00 (approved):
These seem like reasonable changes. Were you planning to backport these are is this a 6.0 thing?

jungpark-mlir:
No need to backport this to 5.4. practically.

PR #935, opened 2023-01-12
Add rock.threadwise_write_all to abstract writeback loop, help fusion

krzysz00:
In order to simplify an upcoming fusion refactoring, define rock.threadwise_write_all, which expands to a transforming_for loop that stores values from a memref of registers to a global buffer after peeling the rock.transform operations off of said buffer.

This operation is currently lowered within
-rock-gridwise-gemm-to-blockwise in order to not disturb the existing fusion code, but that should change in the future.

Also merge in upstream's clang-tidy config while I'm here because with ROCm 5.4.0 I was getting a bunch of complaints about const that weren't actually meant to be showing up.


sjw36 (approved):
Looks good. thx

PR #938, opened 2023-01-16
Add rock reduce operator

manupak:
Add rock.reduce (sum) operator

This commit adds support for lowering
tosa.reduce_sum operator to GPU using
the introduction of new rock.reduce operator.

It adds a simple heuristic to pick blockSize
and gridSize and uses a zero init op to initialize
the output.

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/772


mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -50,6 +50,12 @@ class TensorOrMemRefRankOf<list<Type> allowedTypes, list<int> ranks> :
              /*summary=*/"",
              "::mlir::ShapedType">;
 
+class TensorOrMemRefOf<list<Type> allowedTypes> :
```
krzysz00, line 53: ... hold on, didn't we already have this? ... Ok, since you added it, I guess not

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -50,6 +50,12 @@ class TensorOrMemRefRankOf<list<Type> allowedTypes, list<int> ranks> :
...
+  AnyTypeOf<[TensorOf<allowedTypes>,
+             MemRefOf<allowedTypes>],
+             /*summary=*/"",
```
krzysz00: Can you make the `summary` field meaningful? We'll want that for error messages to work.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -168,6 +174,30 @@ def Rock_GemmOp :
...
+def Rock_ReduceOp :
+  Rock_Op<"reduce", [AllElementTypesMatch<["in", "out"]>]>,
+  Arguments<(ins
+    Arg<TensorOrMemRefOf<[F32]>, "input", [MemRead]>:$in,
+    Arg<TensorOrMemRefOf<[F32]>, "output", [MemRead, MemWrite]>:$out,
+    ReduceMethodAttr:$reduceMethod,
+    IndexAttr:$axis,
+    I32Attr:$blockSize,
+    I32Attr:$gridSize,
+    OptionalAttr<BoolAttr>:$useLDS,
```
krzysz00: This is perhaps more nicely spelled `UnitAttr`

... also, note to past me, `OptionalAttr<UnitAttr> === UnitAttr` and so I should've not done that.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -168,6 +174,30 @@ def Rock_GemmOp :
...
+    OptionalAttr<BoolAttr>:$useDPP
+  )>,
+  Results<(outs Optional<TensorOf<[F32]>>:$result)> {
+  let summary = "Axes-wide reduction operation";
+  let hasVerifier = 1;
+  let assemblyFormat = [{
+    $reduceMethod `(` operands `)` attr-dict `:` `(` type($in) `,` type($out) `)` (`->` type($result)^)?
```
krzysz00: For the sake of my sanity, can we give this a nicer assembly format?

Perhaps `$reduceMethod $in into $out attr-dict type($in) into`type($out)`

Or `$out = $reduceMethod ( $in )`

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -300,6 +300,66 @@ class MatMulConverter final : public OpConversionPattern<tosa::MatMulOp> {
...
+    int64_t gridSize;
+    int64_t blockSize;
+    int64_t elemPerThread;
+    int64_t numElements =
+        adaptor.getInput().getType().cast<ShapedType>().getNumElements();
+    calculateLaunchDims(numElements,
+                        num_cu.has_value() ? (int64_t)num_cu.value() : 0,
+                        gridSize, blockSize, elemPerThread);
+
+    rw.create<rock::ZeroInitKernelOp>(
```
krzysz00: Please ensure that this goes into a separate `gpu.launch` from the `rock.reduce` call. Not doing that means race conditions.

mlir/lib/Conversion/TosaToRock/TosaToRockPass.cpp
```diff
@@ -53,7 +53,7 @@ struct TosaToRock : public impl::TosaToRockPassBase<TosaToRock> {
     target.addLegalDialect<rock::RockDialect, tosa::TosaDialect,
                            tensor::TensorDialect,
                            bufferization::BufferizationDialect>();
-    target.addIllegalOp<tosa::Conv2DOp, tosa::MatMulOp>();
+    target.addIllegalOp<tosa::Conv2DOp, tosa::MatMulOp, tosa::ReduceSumOp>();
```
krzysz00: Dynamically legal, perhaps? We don't want to actually fail on non-sum reductions?

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
```diff
@@ -0,0 +1,233 @@
...
+// reduced from arbitary dimension sizes to bid x tid x iter space.
+static Value createThreadView(Value redInput, int64_t blockSize,
+                              int64_t gridSize, Location loc,
+                              PatternRewriter &rewriter) {
+  auto totalThreads = gridSize * blockSize;
+  auto inpShape = redInput.getType().cast<ShapedType>();
+  auto elementCount = inpShape.getNumElements();
+  auto dataPerThread = (elementCount + (totalThreads - 1)) / totalThreads;
+
+  llvm::errs() << "totalThreads = " << totalThreads << "\n";
```
krzysz00: `LLVM_DEBUG(llvm::dbgs() <<`

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
```diff
@@ -0,0 +1,233 @@
...
+  llvm::errs() << "elementCount = " << elementCount << "\n";
+  llvm::errs() << "dataPerThread = " << dataPerThread << "\n";
+
+  SmallVector<SmallString<8>> topNames;
+  SmallVector<int64_t> topSizes;
+  int64_t topSizeProduct = 1;
+  SmallVector<unsigned int> topDims;
+  SmallVector<StringRef> topNameRefs;
+  for (auto dimAndSize : llvm::enumerate(inpShape.getShape())) {
```
krzysz00, line 77: There's already a three-argument constructor for these builders so you don't have to do this.

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
```diff
@@ -0,0 +1,233 @@
...
+    stMethod = StoreMethod::AtomicAdd;
+    return success();
+  }
+  return failure();
+}
+
+LogicalResult ReduceRewritePattern::matchAndRewrite(
+    ReduceOp op, ReduceOpAdaptor adaptor,
+    ConversionPatternRewriter &rewriter) const {
+  auto loc = op->getLoc();
```
krzysz00: `Location loc` and less use of `auto` generally.

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
```diff
@@ -0,0 +1,233 @@
...
+  uint64_t redAxis = adaptor.getAxisAttr().getInt();
+  auto gridSize = adaptor.getGridSizeAttr().getInt();
```
krzysz00: You can and should just call `op.getGridSize()` - `adaptor` is there specifically for values that might've already been rewritten by other conversion patterns.

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
```diff
@@ -0,0 +1,233 @@
...
+  auto blockSize = adaptor.getBlockSizeAttr().getInt();
+
+  // This pass doing this rewrite means the preceding
+  // fusion passes did not perform any fusion and the reduction
+  // op will become the anchor op. Therefore, setting the
+  // blockSize and gridSize attr of the func.
+  auto funcOp = op->getParentOfType<func::FuncOp>();
```
krzysz00: This sort of auto is fine

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
```diff
@@ -0,0 +1,233 @@
...
+  if (funcOp->hasAttr("block_size")) {
+    llvm::errs() << "blockSize = " << funcOp->getAttr("block_size") << "\n";
```
krzysz00: ... don't llvm::errs() here. The `op.emitError` or even `op.emitOpError` should be enough

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
```diff
@@ -0,0 +1,233 @@
...
+    return op.emitError("func op should not have a blockSize set for a "
```
krzysz00: Also, strong disagree. It's a kernel, give it a block size and grid size attribute just like the rest of them. This breaks the uniformity of `kernel`

krzysz00: That is, we set the attributes on the function at the time we determine them.

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
```diff
@@ -0,0 +1,233 @@
...
+      }
+    }
+    StoreMethod stMethod;
+    ReduceMethod rMethod = adaptor.getReduceMethod();
+    if (getStoreMethod(rMethod, stMethod).failed()) {
+      return op.emitError()
+             << "The Reduce Method" << getNameForReduceMethod(rMethod)
+             << " is not supported.!";
+    }
+    rewriter.create<BufferStoreOp>(
```
krzysz00: `GlobalStoreOp` to allow wider vectors

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
```diff
@@ -0,0 +1,233 @@
...
+  (void)runRegionDCE(rewriter, {*region});
+  return success();
+}
+
+void RockLowerReducePass::runOnOperation() {
+  MLIRContext *ctx = &getContext();
+  ConversionTarget target(*ctx);
+
+  target.addIllegalOp<rock::ReduceOp>();
+  target.addLegalOp<rock::TransformOp, rock::TransformingForOp,
```
krzysz00: Just `addLegalDialect rock::RockDialect`. Much less painful and won't really cause bugs.

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
```diff
@@ -0,0 +1,233 @@
...
+    }
+    rewriter.create<BufferStoreOp>(
+        loc, loadVal, adaptor.getOut(), std::get<0>(loadOobDims),
+        std::get<1>(loadOobDims), storeCoords,
+        StoreMethodAttr::get(rewriter.getContext(), stMethod),
+        /*offset=*/IntegerAttr());
+  }
+  auto region = op->getParentRegion();
+  rewriter.eraseOp(op);
+  (void)runRegionDCE(rewriter, {*region});
```
krzysz00: I don't think you need to do this?

krzysz00: (Or, if you do, maybe it should be happening up in gridwise-gemm-to-blockwise as well)

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
```diff
@@ -0,0 +1,233 @@
...
+      untransform(rewriter, threadViewVal);
+  auto loadOobDims = computeOobFromTransforms(rewriter, sourceTransformsFromOp);
+
+  auto threadViewShape = threadViewVal.getType().cast<ShapedType>().getShape();
+  SmallVector<int64_t> bounds(threadViewShape.size(), 1LL);
+  bounds[threadViewShape.size() - 1] =
+      threadViewShape[threadViewShape.size() - 1];
+  SmallVector<int64_t> strides(threadViewShape.size(), 1LL);
+
+  // Get current workgroup ID.
```
krzysz00, line 143: High-level suggestion, instead of using `(bid, tid, i) -> flatten()` use `(bid, i, tid)` so that we get well-behaved coalescing loads.

krzysz00 (changes requested):
This here is about 70% aesthetics complaints and 30% not introducing footguns.

I'd like to see a test that the memset kernel gets partitioned away from the reduce kernel.

manupak:
Planning to add more tests...

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
manupak: yea -- pending cleanups I missed.

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
manupak: same here -- consequence me commiting to dev branch and moving servers. I m planning to remove these. Apologies about the noise.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
manupak: Right!

Would it better than a global sync here ?

krzysz00: ... there's no such thing as a global sync

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
manupak: There are few dead-end transforms that created to get the maps for transformingFor.
Using this to remove them. 

I think gridwise-gemm-to-blockwise is calling a canonicalize as a cleanup pass (which confused me at first looking at mlir-print-ir-after-all to see canonicalize pass appearing to do so much lowering) -- somehow getting cleaned up there?

krzysz00: Yeah, canonicalize does a DCE.

Also, you
1. Don't exactly _need_ to DCE those
2. Can pass an array of `TransformMap` attributes (or an `ArrayAttr` of them) as the third argument to `untransform()`

mlir/lib/Conversion/TosaToRock/TosaToRockPass.cpp
manupak: Not sure I follow, non-sum reductions are not affected ?

krzysz00: If an op is illegal, and you fail to rewrite it, that's a pass failure.

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
manupak: Thanks for the info. I wasn't getting the reason for the mixed usage of the two.

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
manupak: Hmmm.... currently these are set in AffixTuningParams.
Do you mean to move it there ?

manupak: Or for this case, we determine them at creation (TosaToRock) -- maybe there then ?

krzysz00: The ugly thing is that there's no function there. 

What I'd do is to the copy over in affix-tuning-params.

krzysz00: Yeah, I mean this copy from the attributes to the surrounding function should happen, say, in affix-tuning-params, not in this lowering pattern.

mlir/lib/Conversion/TosaToRock/TosaToRockPass.cpp
manupak: yea and the illegal op specifically is tosa::ReduceSumOp.. so what is the relevance to non-sum reductions ?

krzysz00: ... my not reading correctly and thinking that you were working on a general `tosa.reduce`

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
manupak: I see.

krzysz00: (I mean, there _is_, and it has the form "allocate global memory, have each block inspect the total number of blocks we've been launched with, and do a fetch_add followed by a atomic spin wait for all the blocks to reach this point")

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: Yes, I could change that -- I actually did muse over this and was torn between the two.
I thought I should actually explain why I went this format.  It was mainly because in the tensor format, it would appear as: 

$reduceMethod $in into $out attr-dict type($in) intotype($out)` --> type($result) which I found less clearer compared to standard DPS style with optional returns (which what the proposed is). However, the proposed is inconsistent with other rock ops fancier assembly format -- which Im not sure that Im a fan of -- I do prefer simpler (input operands, output dps operands) : op input types --> op output types.

I think I would vote for consistency and move to your suggestion now but thought should explain for the record.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
jungpark-mlir: Strictly saying, reduce op is reading the output buffer (read by atomic op).
Theoretically, execution layer should evaluate the dataflow analysis and set the dependency between the launch.
- dependency can be represented in async dialect but not sure we can lower to it.
- Assuming there is a layer performs the dataflow analysis, not sure if it handles WAW dependency from the given IR order, otherwise we may want to get the output buffer as an input argument as well.
@sjw do we have an execution-level pipeline which can evaluate dataflow among multiple kernels and run them?

krzysz00: The thing is that creating a kernel operation (like the zero_init_kernel) doesn't actually create a kernel launch.

It's perfectly valid to, say, write
```
func.func @f(%a, %b, %c, %d, %e) attributes {kernel} {
    rock.gemm %c = %a * %b
    rock.gemm %e = %c * %d
}
```
but this will not ensure synchronization on `%c.

Similarly
```
func.func @f(%a, %b, %c) attributes {kernel} {
   rock.zero_init_kernel %c
  rock.gemm %c = %a * %b
}
launch @f
```

is absolutely not equivalent to
```
func.func @f(%a, %b, %c) attributes {kernel} {
  rock.zero_init_kernel %c
}
func.func @g(%a, %b, %c) attributes {kernel} {
  rock.gemm %c = %a * %b
}
launch @f 
launch @g 
```

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
krzysz00: Entirely fair on the DPS situation. I wrote the `rock.*` assembly formats on the principle that they mostly appear in `memref` form.

And, even with DPS, you still get `out`/`return` redundancies.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
manupak: OK, I went back to the drawing board for this one.. 

From a kernel generation PoV what we want here is : 

```llvm
module @__xmodule_amdgcn_amd_amdhsa_gfx90a_srameccY_xnackY attributes {xmodel.arch = "amdgcn-amd-amdhsa:gfx90a:sramecc+:xnack+", xmodel.module} {
    func.func private @test_reduce__part_0_0(%arg0: memref<2x3x1xf32>) attributes {kernel, original_func = @test_reduce__part_0, launch_func = @launch_test_reduce__part_0} {
      rock.zero_init_kernel %arg0 {arch = "amdgcn-amd-amdhsa:gfx90a:sramecc+:xnack+", blockSize = 256 : i32, elemsPerThread = 1 : index, gridSize = 1 : i32} : memref<2x3x1xf32>
      return
    }
    func.func private @test_reduce__part_0_1(%arg0: memref<2x3x40xf32>, %arg1: memref<2x3x1xf32>) attributes {kernel, original_func = @test_reduce__part_0, launch_func = @launch_test_reduce__part_0} {
      rock.reduce  sum(%arg0, %arg1) {axis = 2 : index, blockSize = 256 : i32, gridSize = 1 : i32} : (memref<2x3x40xf32>, memref<2x3x1xf32>)
      return
    }
  }
}
```

Right ? 

Let me zoom out and share my thoughts on a possibility of a partitioning flow that gives access to rewriting the launches per op.

Currently we do the partitioning as follows : 

```llvm
module {
  func.func private @test_reduce__part_0(%arg0: tensor<2x3x40xf32> {func.read_access}) -> (tensor<2x3x1xf32> {func.write_access}) {
    %0 = "tosa.reduce_sum"(%arg0) {axis = 2 : i64} : (tensor<2x3x40xf32>) -> tensor<2x3x1xf32>
    return %0 : tensor<2x3x1xf32>
  }
  func.func @test_reduce(%arg0: tensor<2x3x40xf32>) -> tensor<2x3x1xf32> {
    %token, %results = async.launch @test_reduce__part_0 (%arg0) : (tensor<2x3x40xf32>) -> tensor<2x3x1xf32>
    async.await %token : !async.token
    return %results : tensor<2x3x1xf32>
  }
  module @__xmodule_amdgcn_amd_amdhsa_gfx90a_srameccY_xnackY attributes {xmodel.arch = "amdgcn-amd-amdhsa:gfx90a:sramecc+:xnack+", xmodel.module} {
    func.func private @test_reduce__part_0(%arg0: tensor<2x3x40xf32> {func.read_access}) -> (tensor<2x3x1xf32> {func.write_access}) attributes {kernel, original_func = @test_reduce__part_0} {
      %0 = "tosa.reduce_sum"(%arg0) {axis = 2 : i64} : (tensor<2x3x40xf32>) -> tensor<2x3x1xf32>
      return %0 : tensor<2x3x1xf32>
    }
  }
}
```

What if we do the partitioning as follows : 

```llvm
module {
  func.func private @test_reduce__part_0(%arg0: tensor<2x3x40xf32> {func.read_access}) -> (tensor<2x3x1xf32> {func.write_access}) {
    %0 = "tosa.reduce_sum"(%arg0) {axis = 2 : i64} : (tensor<2x3x40xf32>) -> tensor<2x3x1xf32>
    return %0 : tensor<2x3x1xf32>
  }
  func.func private @launch_test_reduce__part_0(%arg0: tensor<2x3x40xf32> {func.read_access}) -> (tensor<2x3x1xf32> {func.write_access}) {
    %token, %results = async.launch @test_reduce__part_0 (%arg0) : (tensor<2x3x40xf32>) -> tensor<2x3x1xf32>
    async.await %token : !async.token
    return %results : tensor<2x3x1xf32>
  }
  func.func @test_reduce(%arg0: tensor<2x3x40xf32>) -> tensor<2x3x1xf32> {
    %0 = call @launch_test_reduce__part_0(%arg0) : (tensor<2x3x40xf32>) -> tensor<2x3x1xf32>
    return %0 : tensor<2x3x1xf32>
  }
  module @__xmodule_amdgcn_amd_amdhsa_gfx90a_srameccY_xnackY attributes {xmodel.arch = "amdgcn-amd-amdhsa:gfx90a:sramecc+:xnack+", xmodel.module} {
    func.func private @test_reduce__part_0(%arg0: tensor<2x3x40xf32> {func.read_access}) -> (tensor<2x3x1xf32> {func.write_access}) attributes {kernel, original_func = @test_reduce__part_0, launch_func = @launch_test_reduce__part_0} {
      %0 = "tosa.reduce_sum"(%arg0) {axis = 2 : i64} : (tensor<2x3x40xf32>) -> tensor<2x3x1xf32>
      return %0 : tensor<2x3x1xf32>
    }
  }
}
```

that could get lowered to as follows using the launch_func attr that would allows us to rewrite the launches as well :

```llvm
module {
  func.func private @test_reduce__part_0(%arg0: memref<2x3x40xf32> {func.read_access}, %arg1: memref<2x3x1xf32> {func.write_access}) {
    %cst = arith.constant 0.000000e+00 : f32
    %0 = memref.alloc() {alignment = 128 : i64} : memref<2x3xf32>
    linalg.fill ins(%cst : f32) outs(%0 : memref<2x3xf32>)
    linalg.generic {indexing_maps = [#map0, #map1], iterator_types = ["parallel", "parallel", "reduction"]} ins(%arg0 : memref<2x3x40xf32>) outs(%0 : memref<2x3xf32>) {
    ^bb0(%arg2: f32, %arg3: f32):
      %2 = arith.addf %arg2, %arg3 : f32
      linalg.yield %2 : f32
    }
    %1 = memref.expand_shape %0 [[0], [1, 2]] : memref<2x3xf32> into memref<2x3x1xf32>
    memref.copy %1, %arg1 : memref<2x3x1xf32> to memref<2x3x1xf32>
    return
  }
  func.func private @launch_test_reduce__part_0(%arg0: memref<2x3x40xf32> {func.read_access}, %arg1: memref<2x3x1xf32> {func.write_access}) {
    %token0, %results = async.launch @test_reduce__part_0_0 (%arg0) : (memref<2x3x40xf32) -> ()
    async.await %token0 : !async.token
    %token1, %results = async.launch @test_reduce__part_0_1 (%arg0, %arg1) : (memref<2x3x40xf32, memref<2x3x1xf32>) -> ()
    async.await %token1 : !async.token
    return %results : tensor<2x3x1xf32>
  }
  func.func @test_reduce(%arg0: memref<2x3x40xf32>, %arg1: memref<2x3x1xf32>) {
    %0 = call @launch_test_reduce__part_0(%arg0,  %arg1) : (memref<2x3x40xf32>, memref<2x3x1xf32>) -> ()
    return %0 : tensor<2x3x1xf32>
  }
  module @__xmodule_amdgcn_amd_amdhsa_gfx90a_srameccY_xnackY attributes {xmodel.arch = "amdgcn-amd-amdhsa:gfx90a:sramecc+:xnack+", xmodel.module} {
    func.func private @test_reduce__part_0_0(%arg0: memref<2x3x1xf32>) attributes {kernel, original_func = @test_reduce__part_0, launch_func = @launch_test_reduce__part_0} {
      rock.zero_init_kernel %arg0 {arch = "amdgcn-amd-amdhsa:gfx90a:sramecc+:xnack+", blockSize = 256 : i32, elemsPerThread = 1 : index, gridSize = 1 : i32} : memref<2x3x1xf32>
      return
    }
    func.func private @test_reduce__part_0_1(%arg0: memref<2x3x40xf32>, %arg1: memref<2x3x1xf32>) attributes {kernel, original_func = @test_reduce__part_0, launch_func = @launch_test_reduce__part_0} {
      rock.reduce  sum(%arg0, %arg1) {axis = 2 : index, blockSize = 256 : i32, gridSize = 1 : i32} : (memref<2x3x40xf32>, memref<2x3x1xf32>)
      return
    }
  }
}
```
I think it would not stop us serving the existing clients if they dont want to call the launch_func -- we could just export metadata about the kernels in the way they want to call the kernels.

cc : @sjw36 too.

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
manupak: Can you eloborate what you mean here ? Is there another way to do this or are you saying the builder of the op can capture this ?

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
manupak: Done!

manupak (commented):
I think I've addressed all the comments (except for 1 that I dont fully follow) and pushed out the tosa to rock conversions until we have a way to handle output init or graph of kernels.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: Done!

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: Done!

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: Done!

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
manupak: Pushed out tosa to rock conversions for now...

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
manupak: cleaned.

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
manupak: done

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
manupak: done

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
manupak: done

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
manupak: pushed tosa to rock conversions.

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
manupak: done

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
manupak: done

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
manupak: Thanks for the tip on untransform ( I didn't know that) and done!

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -1514,6 +1514,33 @@ void WorkitemIdOp::inferResultRanges(ArrayRef<ConstantIntRanges> argRanges,
...
   setResultRanges(getResult(), getIdRange("block_size", getOperation()));
 }
+
+//===-----------------------------------------------------===//
+// ReduceSumOp
+//===-----------------------------------------------------===//
+
+LogicalResult ReduceOp::verify() {
+  ReduceOp &op = *this;
+  auto axis = op.getAxis();
```
krzysz00: Why `&op` instead of just calling the methods directly?

Also, can we get less `auto`?

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
krzysz00: `TransformMapBuilder` has a constructor that takes three arguments with no names, just a shape.

And if you need the names, you can call `getLowerName(i)` or possibly `getLowerNames()` to get the names - I'd look at the `BottomUpTMBuilder` to find the methods.

krzysz00 (commented):
This all seems reasonable, I'm glad we have tests.

I've added an explanation of what I was thinking with the transform map building - if you want to ignore it, I'll approve.

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
manupak: OK, well I had to migrate a protected one to be public; I think it should be okay ? (as we have getEndNames anyway).

manupak (commented):
All comments addressed now!

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
manupak: @sjw36: FYI, now we can use this API elsewhere we used to create names for TrMaps when the dims are dynamic.

krzysz00: I don't think you want `getEndNames()`, you might want `getStartNames()`

krzysz00: And the reason `getStartNames()` was protected was because I didn't want people trying to modify that list and break class invariants.

We could add `getStartNameRefs()` populating a `SmallVectorImpl<StringRef>&`

manupak: Thats what I have added here ?

```
void TransformMapBuilder::getStartNames(SmallVectorImpl<StringRef> &names) {
```

mlir/lib/Dialect/Rock/IR/TransformMapBuilder.cpp
```diff
@@ -204,6 +204,13 @@ void TransformMapBuilder::getEndNames(SmallVectorImpl<StringRef> &names) {
   }
 }
 
+void TransformMapBuilder::getStartNames(SmallVectorImpl<StringRef> &names) {
```
manupak, line 207: I think we are saying the same thing..

manupak (commented):
@krzysz00 would you be able to take another look?

mlir/lib/Dialect/Rock/IR/TransformMapBuilder.cpp
```diff
@@ -631,7 +638,7 @@ void BottomUpTMBuilder::broadcast(ArrayRef<uint32_t> endDims,
   for (auto tuple : llvm::zip(endDims, endSizes)) {
     uint32_t dim = std::get<0>(tuple);
     int64_t size = std::get<1>(tuple);
-    auto &name = getStartNames()[dim];
+    auto name = startName(dim);
```
krzysz00, line 641: Hopefully there're no weird dangling pointers here, but it should be fine.

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
```diff
@@ -0,0 +1,212 @@
...
+  for (auto dimAndSize : llvm::enumerate(inpShape.getShape())) {
+    size_t dim = dimAndSize.index();
+    int64_t dimSize = dimAndSize.value();
+    topSizes.push_back(dimSize);
+    topSizeProduct *= dimSize;
+    topDims.push_back(dim);
+  }
+
+  BottomUpTMBuilder threadsToInpTensor(rewriter, topSizes, loc);
+  SmallVector<StringRef, 4> topNameRefs;
```
krzysz00: For clarity, this should be `bottonNameRefs` - you've got the thing backwards. The names you're getting are the names of the output dimensions.

krzysz00 (approved):
Minor naming quibble aside, lgtm!

krzysz00:
Sorry about the holdup, I didn't re-read the changes and thought you'd just moved the method from `protected` to `public` without changes.

mlir/lib/Dialect/Rock/Transforms/LowerRockReduce.cpp
manupak: done!

manupak:
Thanks @krzysz00 for the reviews!

PR #939, opened 2023-01-16
Clone memref.global ops into GPU modules in convert-rock-to-gpu

krzysz00:
Fixes
https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/764

### DIFF
--- a/mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
+++ b/mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
@@ -185,21 +185,28 @@ void LowerRockOpsToGPUPass::runOnOperation() {
     b.create<cf::BranchOp>(loc, clonedFuncEntry);
 
     // Clone in global constants
-    WalkResult result =
-        funcBody.walk([&b, &gpuMod, &gpuModuleSymbolTable](
-                          memref::GetGlobalOp op) -> WalkResult {
-          SymbolRefAttr globalSym = op.getNameAttr();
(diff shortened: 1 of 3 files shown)


mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
```diff
@@ -181,6 +184,24 @@ void LowerRockOpsToGPUPass::runOnOperation() {
...
+          SymbolRefAttr globalSym = op.getNameAttr();
+          auto toClone = dyn_cast_or_null<memref::GlobalOp>(
+              SymbolTable::lookupNearestSymbolFrom(op, globalSym));
+          if (!toClone)
+            return WalkResult::interrupt();
+          OpBuilder::InsertionGuard guard(b);
+          Operation *cloned = toClone.clone();
+          // There probably shouldn't be any renames, but let's be careful.
+          StringAttr newNameAttr = gpuModuleSymbolTable.insert(cloned);
+          op.setNameAttr(FlatSymbolRefAttr::get(newNameAttr));
```
sjw36: But now you will have to update references if the name changed. Probably best to assert that it didn't change.

sjw36 (commented):
Just 1 caveat.

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
krzysz00: Ok, I've now tried to do something about that (and the possibility that we might clone the same constant twice)

sjw36 (approved):
LGTM.

PR #941, opened 2023-01-18
Add support for m/n repeats in the reduction case

giuseros:
This is fixing the case where we select a 64x64 `MPerWavexNPerWave` tile and we want it to be implemented as a `32x32` Xdlops tile with `<mRepeats,nRepeats> = <2,2>`. 

If performance are ok, we can enable this to be the default. 

If they are not, we should disable this (making `getMaximalLenPerWave` return always `64`) and work on merging the load/store loops (see https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/781) [like they do in CK](https://github.com/ROCmSoftwarePlatform/composable_kernel/blob/1a0b0e7bec7415a5604da4aec73161ad4b3a8feb/include/ck/tensor_operation/gpu/block/blockwise_gemm_xdlops.hpp#L325) and then reevaluate performance.


mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
```diff
@@ -395,24 +395,22 @@ FailureOr<MfmaInsnGroup> MfmaInsnGroup::select(mlir::Type elementType,
...
-    if (mPerWave > 64) {
-      groupAttr.mRepeats = mPerWave / 64;
-    }
 
-    if (nPerWave > 64) {
-      groupAttr.nRepeats = nPerWave / 64;
-    }
+    // Override the repeat information in case this is for larger wave
+    groupAttr.mRepeats = mPerWave / maximalLen;
+    groupAttr.nRepeats = nPerWave / maximalLen;
```
jerryyin: Instead of inventing more concept like `maximalLen` you can use an existing parameter and it would be able to generalize:

`groupAttr.mRepeats = mPerWave / (mfmaInsnAttr.mfmaNonKDim * mfmaInsnAttr.blocksMfma);`

Since 
 - the reduction ones' `blocksMfma` is always 1, it will be only a fragment of 64
 - the non-reduction ones' `blocksMfma * mfmaNonKDim` must be 64

This can make your repeats deduction logic to scale to all xdlops instructions.

giuseros: Yes, actually this was a bug. I can simply use `m/nPerMfmaGroup`. However, I still need the concept of `maximalLen`, because it will be used to pick the right `m/nPerMfmaGroup` (and hence the right `mfma`). Let me clarify. 

Suppose we have `MPerWave x NPerWave = <64,64>` . 

In our xdlops DB, we have these entries:
1. <64,64> -> `mfma32x32x4f16` -> broadcast
2. <32,32> -> `mfma32x32x8f16` -> reduction

If `maximalLen=64`, this means we select entry `1`  - which is wrong. 

If I change the `<64,64>` entry in our DB to use a reduction and set `m/n repeats = <2,2>` I will cover a `<128x128>` tile size - still wrong. 

By setting `maximalLen=32` I have exactly what I want (and it is coherent with our infra). When I select `fp16` I want the maximal length to be `32x32` (which uses reduction) : anything bigger will be "composed" by `32x32` tiles.

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
```diff
@@ -395,24 +395,22 @@ FailureOr<MfmaInsnGroup> MfmaInsnGroup::select(mlir::Type elementType,
                           << "mPerWave: " << mPerWave << "\n"
                           << "nPerWave: " << nPerWave << "\n");
 
-  // Use 64x64 as base unit in large waves
-  int64_t mPerMfmaGroup = getLenPerMfmaGroup(mPerWave);
-  int64_t nPerMfmaGroup = getLenPerMfmaGroup(nPerWave);
+  // Decision on how we want to "compose" tiles: there is a maximal legnth,
+  // after which we want to iterate over each tile.
+  int64_t maximalLen = getMaximalLenPerWave(elementType);
```
krzysz00: I'd like a different name here. Maximal length of what?

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
jerryyin: In my equation:
 - mfma32x32x4f16: mfmaNonKDim * blocksMfma = 64
 - mfma32x32x8f16: mfmaNonKDim * blocksMfma = 32
 - mfma16x16x16f16: mfmaNonKDim * blocksMfma = 16

On the other hand, your hard-coded number won't work for the 3rd bullet point and would cause wrong repeat number to be computed.

giuseros: Yes, I agree on the bug. So if you see the latest change it is:
```
    groupAttr.mRepeats = mPerWave / mPerMfmaGroup;
    groupAttr.nRepeats = nPerWave / nPerMfmaGroup;
```
I think that `mPerMfmaGroup==mfmaNonKDim * blocksMfma `, but anyway if `mPerWave < maximalLen` then `mPerMfmaGroup=mPerWave` and `groupAttr.mRepeats=1`. Otherwise, it computes the right `mRepeats`

giuseros: So, ok, after a chat I think we clarified everything. Basically `mPerWaves` and `nPerWaves` will act as keys for our database. Everything else needs to be derived from the `mfma` instruction we select. 

The only caveat is that for broadcasts, your formula could get `mRepeats` or `nRepeats` to zero (e.g., `32x64` or `64x32`). I simply did a `std::max(1, formula)` to cover those cases.

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
```diff
@@ -406,20 +406,18 @@ FailureOr<MfmaInsnGroup> MfmaInsnGroup::select(mlir::Type elementType,
...
     auto maybeInsn = MfmaInsn::select(groupAttr.insn);
     if (failed(maybeInsn)) {
       LLVM_DEBUG(llvm::dbgs()
                  << "Unsupported instruction: " << groupAttr.insn << "\n");
       return failure();
     }
+
+    auto mfmaInsnAttr = (*maybeInsn).getAttr();
+    groupAttr.mRepeats = std::max(int64_t(1), mPerWave / (mfmaInsnAttr.mfmaNonKDim * mfmaInsnAttr.blocksMfma));
+    groupAttr.nRepeats = std::max(int64_t(1), nPerWave / (mfmaInsnAttr.mfmaNonKDim * mfmaInsnAttr.blocksMfma));
```
jerryyin: Nit: create a variable and give it a proper name for the multiplication result

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
```diff
@@ -205,10 +205,10 @@ auto getMfmaInsnGroupAttrMap = []() {
                      {{0, 0, MFMAPermB::none}}}},
 
                    {{MfmaTypeId::Fp16TyId, 64, 64},
-                    {ROCDL::mfma_f32_32x32x4f16::getOperationName(),
+                    {ROCDL::mfma_f32_32x32x8f16::getOperationName(),
                      1,
                      1,
```
jerryyin: I'd like m/nRepeats field gone (but initialized to 1, 1 by default). Looks like we are already no longer using it, and keeping this initialized per mfma group will make it confusing.

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
```diff
@@ -205,10 +205,10 @@ auto getMfmaInsnGroupAttrMap = []() {
                      {{0, 0, MFMAPermB::none}}}},
 
                    {{MfmaTypeId::Fp16TyId, 64, 64},
-                    {ROCDL::mfma_f32_32x32x4f16::getOperationName(),
+                    {ROCDL::mfma_f32_32x32x8f16::getOperationName(),
```
jerryyin: For the next step, we need to use those reduction instructions much more aggressively. Could you:
1. Give a 1 to 1 comparison for the CK comparison selected xdlops and see the perf change. (I remember you mentioned there's a certain percentage of improvement so just want to make sure.)
2. Amend the change the majority of the instruction selection to reduction xdlops. Once this is done, we'd need to do a full convolution + gemm tuning, and see the impact of performance in both cases.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -501,8 +502,16 @@ struct BlockwiseGemmV2RewritePattern
...
+      //   for (index_t mr = 0; mr < mRepeats; mr++)
+      //   {
+      //     a[mr*KPerThread + k_i] = p_a_wave[(k_i * num_input_blks + blk_id) *
+      //     M + blk_td + mr*num_input_blks];
+      //   }
+      //   for (index_t nr = 0; nr < nRepeats; nr++)
+      //   {
+      //     b[nr*KPerThread + k_i] = p_b_wave[(k_i * num_input_blks + blk_id) *
+      //     N + blk_td + nr*num_input_blks];
+      //   }
```
jerryyin: I'd prefer:
 - Repeat loop stay outside, and K loop stay inside
 - Once above is done, fuse it with the non-reduction case for the outer and inner loop. We only do the code generation differently inside the inner loop, based on reduction/non-reduction.

jerryyin (changes requested):

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
giuseros: So something like:
```
for m=0:mRepeat:
   for n=0:nRepeat:
      for k=0:KPerThread:
```

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
giuseros: To make sure I understand. I will:
- remove those fields from the xdlops DB
- set the default values for these to 1,1 (when I compute them)

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
giuseros: Just to be clear, the tables are  exactly the same. Check [our table ](https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/941/files/149b3329c1078f0d15272aea537dc1c609ee46db#diff-beeb368616f48d4988170968b32377f1187a7fc55962da48a57bb9c8104b2970 )vs [their table](https://github.com/ROCmSoftwarePlatform/composable_kernel/blob/1a0b0e7bec7415a5604da4aec73161ad4b3a8feb/include/ck/tensor_operation/gpu/warp/xdlops_gemm.hpp). 

The main difference is that they specify `MPerXdl` and `mRepeats` while we specify directly `mPerWave` . So they can have that table and decide to "compose" `mPerWave = 64 = 32*2` from their tuning loop .

To mirror this, without adding more parameters, we need to change the table to hardcode their tuning choices. I guess I can do that for now, but maybe we can discuss this to understand how we want to move on

jerryyin: I don't see an inherent difference here. They look the same to me. Why do you think we need to change our table to mirror their solution?
 - They specify mPerXdlops + mRepeats, and with multiplication, they know mPerWave
 - We specify mPerWave + mPerXdlops, and with a division, we know mRepeats

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
giuseros: Working on this, I realized that the loop I have now is like this:
```
for k = 0:KPerThread:
    for m = 0:mRepeat:
        load A[k,m]
    for n = 0:nRepeat:
        load B[k,n]
``` 

If we want to have the K loop inside, we will have:
```
for m = 0:mRepeat:
    for k = 0:KPerThread:
        load A[k,m]

for n = 0:nRepeat:
    for k = 0:KPerThread:
        load B[k,n]
``` 

So two separate loops with the K loop repeated twice. I think this was originally the reason why I kept the K loop outside. What version should I prefer? Btw, I still think that these things should be parametrized and exposed as transformations instead of hard-coding a choice

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
giuseros: I am referring to this point:

> Amend the change the majority of the instruction selection to reduction xdlops. Once this is done, we'd need to do a full convolution + gemm tuning, and see the impact of performance in both cases.

If we amend the change to use more reduction xdlops, we will have to change our table, which will diverge from the one CK uses.

giuseros: Also, when I say "we specify", I mean "the tuner specifies". At tuning time they provide `mRepeats` and `mPerXdl` while our  tuner only specifies `mPerWave` (and we deduce `mPerXdl` and `mRepeats`)

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: I think whether you want `m/n` or `k` being the inner loop depends on the layout of LDS, which means it's a function of `kpack`.

Or possibly this is another case for `getMaxVectorization()` to determine the loop order.

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
krzysz00: @jerryyin I think the point here is that CK has an extra tuning parameter that they use to distinguish `MPerWave = 64 <=> MPerXdlops = 64, MRepeats = 1` from `MPerWave = 64 <=> MPerXdlops = 32, MRepeats = 2`. We currently can't distinguish those with our tuning parameters.

krzysz00 (commented):
Hopefully this puts my comments through

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
giuseros: Looking at non-reduction cases, I think I get what @jerryyin meant. He wants the code to look exactly like non-reduction, where they use the second version of the loop (i.e., the two separate loops). Is that right @jerryyin ? Btw, why is it a function of KPack? Still looking at the non-reduction case, I don't see much different in the logic they treat `KPack==1` and `KPack>1` cases, except for a multiplication by `KPack` itself when it's `>1`.

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
giuseros: This is outdated, I removed it in favour of the equation @jerryyin is suggesting.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: I mention kPack because it affects the LDS layout, which is `[k, m, kPack]`. If a thread needs to load multiple `k` values, say `A[0, 0]` and `A[0, 1]` (logical m/k coordinates), it matters if those are `buffer[0]` and `buffer[1]` (as they would be if `kPack >= 2`) or if they are `buffer[0]` and `buffer[M]`, as they would be if `kPack = 1`.

giuseros: Well, yes, but if `KPack  > 1`  I think we always deal with vectors of KPack and those vectors are always a stride distant (M or N). Basically (IIUC) the unit of load is always a stride apart in the LDS buffer

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
jerryyin: @giuseros Did they actually used this flexibility? I believe the answer is no because in the limited tuning parameter selections that we have looked at together, they'd not compose the same wave size in two ways. They always only have the reduction combination.

@krzysz00 There's the background that CK has removed all broadcast cases and replaced them all-together with reduction cases. In your example, when MPerWave = 64 there's only 1 sane reduction xdlops, which is to use `32x32x8fp16`. In this case we can deduce the repeat to be 2 correctly.

I think the intent here is to test and get comparison data. If we can verify the conclusion as CK does, that reduction is at least as good as broadcast, then there's no reason to change our implementation for now.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
jerryyin: > for k = 0:KPerThread:
    for m = 0:mRepeat:
        load A[k,m]
    for n = 0:nRepeat:
        load B[k,n]

@giuseros Please avoid this. This will make you run into the same backend bug as we have discussed yesterday. Have two loops and use lambda in the inner loop.

jerryyin: @krzysz00 
> I mention kPack because it affects the LDS layout

Those current loop computation will consider about kPack but hardcoded multiplication from base offset with kPack. I think the biggest obstacle we fuse those loop together is blocked by backend bug.

> Or possibly this is another case for getMaxVectorization() to determine the loop order.

Yes, once we move kPack to be its own dimension, using max vectorization is a meaningful thing to try.

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
giuseros: Re do they use the flexibility? Somehow yes. Because they can have our exact table but they are still able to specify (during tuning) only the reductions. If we want to do that,  we will have to change our table and remove the broadcasts. But this means that we will loose even the opportunity to try them. 

And this maybe touches also your second point: they didn't remove the broadcasts: they only tune for reductions, but all the broadcasts are still there in the table. So if they ever want them, they can always try them (while if we change our table, we won't be able to do this)

jerryyin: @giuseros 
> Somehow yes. Because they can have our exact table but they are still able to specify (during tuning) only the reductions.

I don't think they and us are tuning based on the same range. In the tuning range they have, I remember they wouldn't pick non-reduction xdlops. If I'm wrong, please refence the line where they have this sort of flexibility.

> they didn't remove the broadcasts: they only tune for reductions

I think their blockwise copy only have reduction flavored the logic because it help to simplify the implementation.

Look, taking one step back, I do think we need data to convince each other. Find me a case where broadcast alternative beat reduction and we can mimic what they do. I'm leaning towards not to do it as for now because tuning and re-tuning is such a pain, and I want to avoid it if possible.

giuseros:
@jerryyin , @krzysz00 , 
I reordered the loading loops and forced 64x64 (both for `fp16` and `fp32`) to use a reduction. 

What I am attaching is the perf comparison using the same tuning db. Long story short there are few small wins and 1 small loss. 

There is also a moderate win on `i8` for a `mWave x nWaves = 32x32`, which is a configuration I think I am not touching (but I might be wrong). 

Next, I will try to use only reductions and create a new tuning DB. 

[reduction64x64.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10536928/reduction64x64.zip)

jerryyin:
@giuseros The performance looks even. I think what this indicate is that switching to reduction doesn't degrade performance even for the cases where broadcast is winning. Next up we need to do tuning to see what the overall impact is.

Note: i8 are already using reduction and there's no way touching fp16 can impact that. So that means if the perf diff <= 12% should all be viewed as normal deviations.

krzysz00:
I'd also test dropping swizzles if you're focusing on Gemm

giuseros:
@jerryyin , @krzysz00 
The last commit contains moving everything to reductions. I will first make the tests pass (this is hopefully but unlikely a `nop`). Please, have a look to see if there is anything obviously wrong

@krzysz00 
I would try to remove swizzles in a separate PR after this is merged (so that we don't mix changes)

giuseros:
Hi guys, 
These are the last performance reports with all reductions + retuning. 

Overall, I agree with @jerryyin : performance look even. What do you think? 

[reductionAll.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10547089/reductionAll.zip)

mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
```diff
@@ -110,6 +110,14 @@ struct MfmaInsnGroupAttr {
...
   int64_t nRepeats;
   SmallVector<MFMAParams, 2> imms;
+  // Reduction constructor
+  MfmaInsnGroupAttr(const SmallString<16> &insn)
+      : insn{insn}, mRepeats{1}, nRepeats{1},
+        imms{{{0, 0, amdgpu::MFMAPermB::none}}} {}
+  // Broadcast constructor
+  MfmaInsnGroupAttr(const SmallString<16> &insn,
+                    const SmallVector<MFMAParams, 2> &imms)
+      : insn{insn}, mRepeats{1}, nRepeats{1}, imms{imms} {}
```
jerryyin: I don't like the idea of pre-populating with wrong m/nRepeats and then correct them. We can go with either path:
 - Populate the correct repeat at the initialization time
 - Get rid of m/nRepeat as member variable and compute them when caller wants to `getMRepeat()` or `getNRepeat()`. This is in consideration that the computation overhead is minimal.

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
```diff
@@ -406,20 +323,20 @@ FailureOr<MfmaInsnGroup> MfmaInsnGroup::select(mlir::Type elementType,
...
       LLVM_DEBUG(llvm::dbgs()
                  << "Unsupported instruction: " << groupAttr.insn << "\n");
       return failure();
     }
+
+    auto mfmaInsnAttr = (*maybeInsn).getAttr();
+    // mnPerXdl is how many row/columns a single Xdlops instruction will compute
+    int64_t mnPerXdl = (mfmaInsnAttr.mfmaNonKDim * mfmaInsnAttr.blocksMfma);
+    groupAttr.mRepeats = std::max(int64_t(1), mPerWave / mnPerXdl);
+    groupAttr.nRepeats = std::max(int64_t(1), nPerWave / mnPerXdl);
```
jerryyin: Following the previous message, we can push this into either:
 - The `get*` function
 - The static initialization

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -411,157 +410,133 @@ struct BlockwiseGemmV2RewritePattern
...
-                                    ValueRange{destOffsetB});
+    auto KPerThreadConstantOp = b.create<ConstantIndexOp>(loc, KPerThread);
+
+    // store bufferA logic.
+    // for(index_t m_i = 0; n_i < mRepeats; ++n_i)
+    //   for(index_t k_i      = 0; k_i < K; ++k_i)
+    //     if (!Reduction) a[k_i + m_i * KPerThread] = p_a_wave[k_i * N + laneId
+    //     +  mPerMfmaGroup* m_i]; if (Reduction)  a[k_i + m_i * KPerThread] =
+    //     p_a_wave[(k_i * num_input_blks + blk_id) * N + blk_td +
+    //     m_i*num_input_blks];
```
jerryyin: Could you fix the alignment for comments? It is hard to read now.

Also, I'd like to keep the original branch of reduction/non-reduction such that the two code path are populated separately. I agree that you have made it more succinct by combining them together, but having to branch multiple times based on `IsKReduction` had made readability of the code dropped significantly. It is hard to reason about the offset computation that should stay decoupled between the two path.

My original comment was more meant the following:

```cpp
auto ldsToRegisterCopy = [&](len_i, len, k_i, K, reg, lds) {
    dstOffset = k_i + len_i * K;
    if (IsKReduction) {
       srcOffset = k_i * len + laneId + mPerMfmaGroup * len_i;
    } else {
       srcOffset = (k_i * num_input_blks + blk_id) * len + blk_td + len_i * num_input_blks;
    }
    reg[dstOffset] = lds[srcOffset];
}

for m:
   for k:
      ldsToRegisterCopy(m_i, M, k_i, K, regA, ldsA);

for n
   for k:
      ldsToRegisterCopy(n_i, N, k_i, K, regB, ldsB);
```

This way, it is easy to see that the only difference between kReduction and non-kReduction is the source offset computation. And it will make it easier for us to move this logic to coordinate transform later.

mlir/test/Dialect/Rock/lowering_blockwise_gemm_v2.mlir
```diff
@@ -2,7 +2,7 @@
 
 func.func @rock_blockwise_gemm_v2_two_results(%matrix : memref<1024xf32, 3>,
                                                 %bufferA : memref<2xvector<2xf32>, 5>, %bufferB : memref<2xvector<2xf32>, 5>,
-                                                %matrixC : memref<2xvector<32xf32>, 5>) {
+                                                %matrixC : memref<4xvector<16xf32>, 5>) {
```
jerryyin, line 5: Why change this?

mlir/test/Dialect/Rock/lowering_xdlops_gemm_v2.mlir
```diff
@@ -1,21 +1,21 @@
 // RUN: rocmlir-opt -rock-threadwise-gemm-lowering %s | FileCheck %s
-#transform_map0 = #rock.transform_map<affine_map<(d0, d1) -> (d0 * 8 + d1)> by [<Embed{8, 1} ["m", "k"] at [0, 1] -> ["raw"] at [0]>] bounds = [1, 8] -> [8]>
-#transform_map1 = #rock.transform_map<affine_map<(d0, d1) -> (d0 * 8 + d1)> by [<Embed{8, 1} ["n", "k"] at [0, 1] -> ["raw"] at [0]>] bounds = [1, 8] -> [8]>
-#transform_map2 = #rock.transform_map<affine_map<(d0, d1, d2) -> (d0 * 2 + d1 * 2 + d2)> by [<Embed{2, 2, 1} ["m", "n", "v"] at [0, 1, 2] -> ["raw"] at [0]>] bounds = [1, 1, 2] -> [2]>
+#transform_map0 = #rock.transform_map<affine_map<(d0, d1) -> (d0 * 2 + d1)> by [<Embed{2, 1} ["m", "k"] at [0, 1] -> ["raw"] at [0]>] bounds = [2, 2] -> [4]>
+#transform_map1 = #rock.transform_map<affine_map<(d0, d1) -> (d0 * 2 + d1)> by [<Embed{2, 1} ["n", "k"] at [0, 1] -> ["raw"] at [0]>] bounds = [1, 2] -> [2]>
+#transform_map2 = #rock.transform_map<affine_map<(d0, d1, d2) -> (d0 * 2 + d1 + d2)> by [<Embed{2, 1, 1} ["m", "n", "v"] at [0, 1, 2] -> ["raw"] at [0]>] bounds = [2, 1, 1] -> [2]>
```
jerryyin, line 4: Why change this unit test? I don't see any problem with the way it was.

jerryyin:
Based on the review, I believe the mRepeat and nRepeat will continue to be 1 for the int8 case (You did not add new entries to support a different mPerWave and nPerWave size for int8.), therefore, this code change should not impact the execution of int8 at all. If you disagree, please go ahead to do a ISA dump and compare the new/previous IR/ISA for int8 config. I think  you will see the exact same IR/ISA between and after this change.

krzysz00:
I think that the unit test needs to change because the reduction has a different k_base than the broadcast?

Also, that refactoring that you called out _could_ be causing the code structure changes that are giving int8 a performance improvement.

jerryyin:
> I think that the unit test needs to change because the reduction has a different k_base than the broadcast?

But he didn't change the interface? The signature continue to be matrixA: `m x k`, matrixB: `n x k`, matrixC: `m x n x v`. I think my point is that with reduction it consumes more k's. But a number of @giuseros change make the size actually smaller. It doesn't quite add up to me.

> Also, that refactoring that you called out could be causing the code structure changes that are giving int8 a performance improvement.

With m/nRepeat guaranteed to be zero (len_i becomes 0) because there will only be one repeat, the equation:

` srcOffset = (k_i * num_input_blks + blk_id) * len + blk_td + len_i * num_input_blks;`

Will become:

` srcOffset = (k_i * num_input_blks + blk_id) * len + blk_td;`

This offset computes the same offset as before the refactoring.

krzysz00:
@jerryyin It's worth remembering that `xdlops_gemm_v2` (last I checked) takes vectors of some width (I think either `kpack` or `k_base`) as input in the internal buffers. So changing the xdlop from broadcast to reduction could change which values are needed.

Also, the broadcast may have been loading more values, so there were more values in registers.

mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
giuseros: The first is not possible, because 128 tiles are not in the table, but the user can still select them (and we would use the 64 entr, but mRepeats needs to be 128/32=4 instead of 64/32=2). I can get on with the second method you propose (I also quite like it)

mlir/test/Dialect/Rock/lowering_blockwise_gemm_v2.mlir
giuseros: Because we are using 64x64 mPerWave and nPerWave. This means mRepeats=2 and nRpeats=2, which means that this will result in a  `2*2*vector<16xf32>` output. The `2*2` concerns the repeats and the `vector<16xf32>` concerns the fact that we are doing a reduction and not a broadcast

mlir/test/Dialect/Rock/lowering_xdlops_gemm_v2.mlir
giuseros: The reasoning is similar to the previous reply. We are using reductions now, so for instance in the first example: 
- `A` will be `mRepeats*KPerThreadsxf32` and needs to be transformed in `mRepeatsxKPerThreadxf32`, i.e., `2*2xf32` to `2x2xf32`
- Same reasoning for `B`, but `nRepeats` is one, since `nPerWave=32`
- `C` will be `mRepeats*nRepeats*nvectorsxf32` and need to be transformed to `mRepeats 
 x nRepeats x nvectors x f32`, I.e., from `2*2*1xf32` to `2x2x1xf32`

The same holds for the other example.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
giuseros: Yes, it is fine, I was wondering between the two. As long as we have the loop common to the two cases, it is fine for me

mlir/test/Dialect/Rock/lowering_blockwise_gemm_v2.mlir
jerryyin: Fair enough, thanks for the explanation. I think I confused your on-going work with mine. I just took a look at the gridwisegemm logic and confirmed result memref type is `mRepeats * nRepeats * imms.size()`.

mlir/test/Dialect/Rock/lowering_xdlops_gemm_v2.mlir
jerryyin: Thanks for the explanation. LGTM!

jerryyin:
@krzysz00 Yes, I was reminded that the reduction case would require more repeats in general to handle a size larger than reduction itself. So this all make sense now.

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
```diff
@@ -159,153 +159,70 @@ auto getMfmaInsnGroupAttrMap = []() {
   static llvm::DenseMap<MfmaInsnGroupSelectKey, MfmaInsnGroupAttr,
                         MfmaInsnGroupSelectKeyInfo>
       groupAttrMap{{{MfmaTypeId::Fp32TyId, 64, 64},
-                    {ROCDL::mfma_f32_32x32x1f32::getOperationName(),
-                     1,
-                     1,
-                     {{1, 0, MFMAPermB::none}, {1, 1, MFMAPermB::none}}}},
+                    {ROCDL::mfma_f32_32x32x2f32::getOperationName()}},
```
jerryyin, line 162: This is not a review comment but throwing some thoughts here.

I think compared with CK we have a chance to shorten this list. (But this may moderately impact performance). For example, I think their wave size selection only has "basic units nubmers" like: `64x64`, `32x32`, `16x16` and `4x4`. Then we are pretty much covered with the same support range. On the other hand, our choice of "unevenness" seems to leave more choice. For example for a block size of 64x32: 
 - We can either work it out using 2 waves of 32x32
 - Or, use one wave of 64x32, but `mRepeat = 2`

Now, thinking aloud, it looks to me that keeping those entries are still necessary because different topology will yield different solutions.

mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
jerryyin: I still think the first approach is doable because at the initialization time we already know the wave size and the selected instruction. So it is equivalent to the computation you carried on below for each entry. But I agree that it may not worth the effort to move such trivial computation into the static initialization. Good choice on the second alternative~

giuseros:
@jerryyin I applied the changes we agreed. As per the performance comparison, I printed the IR immediately after the BlockwiseToThreadwise and you were right, they are exactly the same. I also reran that particular experiment, and indeed the difference is much smaller.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -405,170 +403,124 @@ struct BlockwiseGemmV2RewritePattern
...
-              loc,
-              lklb.create<AddIOp>(
-                  loc,
-                  lklb.create<MulIOp>(loc, lkliv,
-                                      inputSpansPerMfmaInConstantOp),
-                  blk_id),
-              MConstantOp),
-          kBaseA);
-
+    auto KPerThreadConstantOp = b.create<ConstantIndexOp>(loc, KPerThread);
```
krzysz00: Can we get `Value` here, not `auto`? Or do we care? (Not that big a deal)

krzysz00 (commented):
Seems broadly reasonable to me.

giuseros:
I gave another go to performance: now fluctuations are reduced, but overall I still a bit of it, especially for int8 . However, I guess we can conclude they are pretty much even in terms of performance. 

[reductionAll2.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10571724/reductionAll2.zip)

krzysz00:
I'll note that the performance changes that show up as statistically significant are mostly in cases where tuning picked a different perfConfig, so I think there's an argument to be made here that you're seeing the result of bad luck during tuning.

But, overall, a small bump is still a bump, I'll take it.

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
giuseros: > - We can either work it out using 2 waves of 32x32
> - Or, use one wave of 64x32, but mRepeat = 2

We should be able to achieve that now, but it means having the big table in c++. 

This is what I was saying when I was talking about including `mRepeats`/`nRepeats` in the number of parameters of the tuner. The tuner should decide that, while our table should contain only the unit-tiles (i.e., the building blocks).  

Just to be clear, I am not suggesting to make the search space bigger (unless we benefit from this), but I think the pruning of the search space should be the tuner prerogative and not hardcoded in C++

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
giuseros: It's not a big deal, but I agree on the principle that auto should be used only when the type is too verbose. I replaced a bunch of `auto` with `Value` so it's more readable .

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -394,181 +393,134 @@ struct BlockwiseGemmV2RewritePattern
...
+    //   for(index_t k_i= 0; k_i < K; ++k_i)
+    //
+    // Note: p_a_wave need to be offseted by waveOffsetA.
+    Value inputSpanLenConstantOp = b.create<ConstantIndexOp>(loc, inputSpanLen);
+    Value inputSpansPerMfmaInConstantOp =
+        b.create<ConstantIndexOp>(loc, inputSpansPerMfmaIn);
+
+    Value blk_id = b.create<DivUIOp>(loc, laneId, inputSpanLenConstantOp);
+    Value blk_td = b.create<RemUIOp>(loc, laneId, inputSpanLenConstantOp);
+
```
jerryyin: Move those in the reduction condition to scope them better.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -394,181 +393,134 @@ struct BlockwiseGemmV2RewritePattern
...
-              MConstantOp),
-          kBaseA);
-
+    Value KPerThreadConstantOp = b.create<ConstantIndexOp>(loc, KPerThread);
+
+    // store bufferA logic.
+    // for(index_t m_i = 0; n_i < mRepeats; ++n_i)
+    //   for(index_t k_i= 0; k_i < K; ++k_i)
+    //
+    // Note: p_a_wave need to be offseted by waveOffsetA.
```
jerryyin: Move it to below where we populate the loop

jerryyin (approved):
Last couple of nits and this is ready to merge. Approving it now, merge it after you have addressed those. Overall thanks for your work and exploration in the area!

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
giuseros: I did this, but a thought occurred to me. If we bring the `blk_id` and `blk_td` inside the `lambda` we are duplicating these. I.e., we will have:
```
blkid = computeblkid()
blktd = computeblktd()
for m:
    for k:

blkid = computeblkid()
blktd = computeblktd()
for n:
    for k:
```
Is this ok? I am worried that the compiler might not be so smart to get rid of the duplication

giuseros: cc also @krzysz00

krzysz00: I _think_ those operations have `NoSideEffects`/`Pure`, so when we run CSE, which we eventually will, they'll get merged

giuseros: Yes they do (just checked) :) Thanks @krzysz00 !

PR #942, opened 2023-01-18
Enabling looped load/execution for xdlops gemm

jerryyin:
This PR convert the flat load/execution loop of blockwise and xdlops gemm from

```cpp
for m = 0 : mRepeat
  for i = 0 : K 
    a = load(A) 

for n = 0 : nRepeat
  for i = 0 : K
    b = load(B) 

for m = 0 : mRepeat
  for n = 0 : nRepeat
    for i = 0 : K 
      c = a * b 
```

to

```cpp
for m = 0 : mRepeat
  for n = 0 : nRepeat
    for k = 0 : K   
      a = load(A)
      b = load(B)
      c = a * b
```

This reduces the register to allocate at one go and increase the amount of parallelism between load and execution.

Next step of after PR, per discussion with @giuseros is to make the loop pipelined in a follow-up pass.

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
@@ -416,7 +416,10 @@ struct BlockwiseGemmV2RewritePattern
     auto nOffset = olnb.create<AddIOp>(
         loc, bBase, olnb.create<MulIOp>(loc, nPerMfmaGroupConstantOp, olniv));
 
-    auto innerLoopK = olnb.create<AffineForOp>(loc, 0, KPerThread);
+    int64_t kStep = 1;
+    if (KPack == 1)
+      kStep = k_base;
+    auto innerLoopK = olnb.create<AffineForOp>(loc, 0, KPerThread, kStep);
     auto ilkb = ConversionPatternRewriter::atBlockBegin(innerLoopK.getBody(),
                                                         olmb.getListener());
     auto ilkiv = innerLoopK.getInductionVar();
@@ -485,11 +488,11 @@ struct BlockwiseGemmV2RewritePattern
         Value valueA = ilkbase.create<InBoundsLoadOp>(
             loc, bufferAElementType, op.getMatrixA(), sourceOffsetA);
         ilkbase.create<memref::StoreOp>(loc, valueA, bufferA,
-                                        ValueRange{zeroConstantOp});
+                                        ValueRange{ilkbaseiv});
         Value valueB = ilkbase.create<InBoundsLoadOp>(
             loc, bufferBElementType, op.getMatrixB(), sourceOffsetB);
         ilkbase.create<memref::StoreOp>(loc, valueB, bufferB,
-                                        ValueRange{zeroConstantOp});
+                                        ValueRange{ilkbaseiv});
       }
     } else {
       // const index_t blk_id = laneId / mfma_type.num_threads_blk;
@@ -567,11 +570,11 @@ struct BlockwiseGemmV2RewritePattern
         Value valueA = ilkbase.create<InBoundsLoadOp>(
             loc, bufferAElementType, op.getMatrixA(), sourceOffsetA);
         ilkbase.create<memref::StoreOp>(loc, valueA, bufferA,
-                                        ValueRange{zeroConstantOp});
+                                        ValueRange{ilkbaseiv});
         Value valueB = ilkbase.create<InBoundsLoadOp>(
             loc, bufferBElementType, op.getMatrixB(), sourceOffsetB);
         ilkbase.create<memref::StoreOp>(loc, valueB, bufferB,
-                                        ValueRange{zeroConstantOp});
+                                        ValueRange{ilkbaseiv});
       }
     }
 
--- a/mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
@@ -373,15 +373,6 @@ struct BlockwiseGemmV2RewritePattern
           "this should never happen");
     }
 
-    // const index_t laneId = get_thread_local_1d_id() % mfma_type.wave_size;
-    // FloatA a[KPerThread * mRepeats];
-    // FloatB b[KPerThread * nRepeats];
-    // constexpr index_t KRepeats = KPack / mfma_type.k_base;
-    // auto pa = reinterpret_cast<const data_type*>(&a);
-    // auto pb = reinterpret_cast<const data_type*>(&b);
-    // constexpr index_t AStride = KPerThread * KRepeats;
-    // constexpr index_t BStride = KPerThread * KRepeats;
-
     auto tid = b.create<WorkitemIdOp>(loc, b.getIndexType());
     constexpr int64_t waveSize = 64;
     auto laneId =
@@ -396,7 +387,6 @@ struct BlockwiseGemmV2RewritePattern
 
     auto MConstantOp = b.create<ConstantIndexOp>(loc, M);
     auto NConstantOp = b.create<ConstantIndexOp>(loc, N);
-    auto KConstantOp = b.create<ConstantIndexOp>(loc, K);
 
     auto mPerMfmaGroupConstantOp =
         b.create<ConstantIndexOp>(loc, mPerMfmaGroup);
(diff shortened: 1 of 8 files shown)


giuseros:
Just a comment on the headline section. I think what we want is:
``` 
for m = 0 : mRepeat
 a = load(A)
  for n = 0 : nRepeat
   b = load(B)
    for k = 0 : K   
      c = a * b
```
I didn't look at the code yet, but surely this should be the final result we are after (we want to still load `a` and `b` once and use them through the loop - [like here](https://github.com/ROCmSoftwarePlatform/composable_kernel/blob/1a0b0e7bec7415a5604da4aec73161ad4b3a8feb/include/ck/tensor_operation/gpu/block/blockwise_gemm_xdlops.hpp#L316))

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -412,163 +402,183 @@ struct BlockwiseGemmV2RewritePattern
...
           kBaseB);
 
-      if (KPack > 1)
-        sourceOffsetB = lklb.create<MulIOp>(
-            loc, sourceOffsetB, lklb.create<ConstantIndexOp>(loc, KPack));
-
-      Value valueB = lklb.create<InBoundsLoadOp>(
-          loc, bufferBElementType, op.getMatrixB(), sourceOffsetB);
-      lklb.create<memref::StoreOp>(loc, valueB, bufferB, ValueRange{lkliv});
+      if (KPack > 1) {
```
krzysz00, line 542: Am I missing something, or is this code basically repeated twice?

mlir/lib/Dialect/Rock/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -203,60 +202,24 @@ struct XdlopsGemmV2RewritePattern : public OpConversionPattern<XdlopsGemmV2Op> {
...
     int64_t k_base = mfmaAttr.k_base;
 
-    bool IsKReduction = (blocksInOutRegs == 1) && (inputSpansPerMfmaIn > 1);
-
-    Value matrixA = adaptor.getMatrixA();
-    Value matrixB = adaptor.getMatrixB();
-    Value matrixC = adaptor.getMatrixC();
-    auto matrixAType = matrixA.getType().cast<MemRefType>();
-    auto matrixBType = matrixB.getType().cast<MemRefType>();
+    Value bufferA = adaptor.getMatrixA();
```
krzysz00, line 212: If you `TypedValue<MemRefType>` here, you don't need the `cast<>()` later down.

mlir/lib/Dialect/Rock/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -279,66 +242,32 @@ struct XdlopsGemmV2RewritePattern : public OpConversionPattern<XdlopsGemmV2Op> {
 
     auto generateMfmaOnKDim = [&](Value &regAOffset, Value &regBOffset,
                                   Value &regCOffset) {
+      // Doesn't work for now
```
krzysz00, line 245: What doesn't work about it?

krzysz00 (commented):
Ok, so, if I'm understanding the setup correctly, this is moving the MRepeats / NRepeats loop out into blockwise_gemm?

Also, why remove the transforming_for based computations of the locations to load from? We put them there to stop emitting a bunch of explicit multiplies and adds?

Also, while we're here, do we want to loosen up the requirements on kPack (aka the type checks on blockwise_gemm_?v2)? We could just pass xdlops_gemm a MxK / NxK (or KxM / KxN) view of LDS that will implicitly lower down to kPack when we load k_base things from it?

jerryyin:
@giuseros You think that we should load the entire K dimension for each repeat at one go? I'd more or less be worried about us running out of registers when we have a huge number of kPerThread * kPack size. (Maybe that's necessary overhead.) 

I think from the discussion perspective, we can control follow-up about how many K to load each time. Desirably we want to load just the right amount to hide delay between lds-> register load and xdlops execution.

mlir/lib/Dialect/Rock/Transforms/ThreadwiseGemmLowering.cpp
jerryyin: Sorry that's leftover comment... Will get rid of this comment.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
jerryyin: It somewhat repeated twice between the kpack > 1 and kpack = 1 conditions because we have to do differently between the two types. I was reminded about this in the overdue ticket that uniform kpack into a dimension as well. But this PR is geared towards performance (and evaluation). I'd very well like to decouple the refactoring ticket to a different PR.

mlir/lib/Dialect/Rock/Transforms/ThreadwiseGemmLowering.cpp
jerryyin: Yes, this is leftover that should be get rid of, will remove.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: I'd like to use the fact that we're doing a rewrite for performance as a reason to also do the refactoring so we can't keep putting off the refactoring for forever.

jerryyin: Right at the moment I don't seek for merging this PR until I got clear experiment feedback on performance improvements. I think I agree with you in general, that refactoring is good, but the question is about when.

As you can see in the past experience, the refactoring PRs (transform-maps in threadwisegemm) that went through ended up getting discarded because we'd no longer need a coordinate transform in the inner k loop. I'd argue that putting off the pure refactoring PRs has a benefit (in my efficiency) when we fast prototype.

Talking about this PR, I may end up discarding it if experiment shows a performance degrade. If there's a perf improvement, and we decide to keep it, I'd still leaning towards doing refactoring in a follow-up PR, just not to pack all changes together. We should create ticket and assign to next milestone if you have concerns on the when problem.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -412,163 +402,183 @@ struct BlockwiseGemmV2RewritePattern
 
     int64_t KPerThread = IsKReduction ? K / inputSpansPerMfmaIn : K;
 
+    auto outerLoopM = b.create<AffineForOp>(loc, 0, mRepeats);
+    auto olmb = ConversionPatternRewriter::atBlockBegin(outerLoopM.getBody(),
+                                                        b.getListener());
+    auto olmiv = outerLoopM.getInductionVar();
+    auto mOffset = olmb.create<AddIOp>(
+        loc, aBase, olmb.create<MulIOp>(loc, mPerMfmaGroupConstantOp, olmiv));
```
giuseros, line 410: I don't think this will work for reductions as well (and since it is outside the `if`, I think you meant for it to work with reductions). I guess to unify this, we should use the same trick you mentioned in #941 , i.e., using `mfmaInsnAttr.mfmaNonKDim * mfmaInsnAttr.blocksMfma`

giuseros (changes requested):

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -412,163 +402,183 @@ struct BlockwiseGemmV2RewritePattern
...
+
+    auto outerLoopN = olmb.create<AffineForOp>(loc, 0, nRepeats);
```
giuseros, line 412: Also, is it possible to make these `AffineForOp` to be `transforming_for`? I know the code was already in that way, but I am wondering if there is any technicality I am missing on why we can't use `transforming_for`. It would be beneficial also because it does have unrolls,etc..

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -412,163 +402,183 @@ struct BlockwiseGemmV2RewritePattern
...
-      auto blk_id = b.create<DivUIOp>(loc, laneId, inputSpanLenConstantOp);
-      auto blk_td = b.create<RemUIOp>(loc, laneId, inputSpanLenConstantOp);
+          ilkb.create<ConstantIndexOp>(loc, inputSpanLen);
+      auto blk_id = ilkb.create<DivUIOp>(loc, laneId, inputSpanLenConstantOp);
+      auto blk_td = ilkb.create<RemUIOp>(loc, laneId, inputSpanLenConstantOp);
 
-      Value kBaseA = b.create<AddIOp>(loc, aBase, blk_td);
-      Value kBaseB = b.create<AddIOp>(loc, bBase, blk_td);
+      Value kBaseA = ilkb.create<AddIOp>(loc, aBase, blk_td);
+      Value kBaseB = ilkb.create<AddIOp>(loc, bBase, blk_td);
```
giuseros, line 506: This should be `mOffset` instead of `aBase` and `nOffset` instead of `bBase`. If you make this change, this'll work also for the reduction case.

giuseros (changes requested):

giuseros:
So @jerryyin , I made it work this PR also for reductions (it's two straightforward changes). But I  see the performance 2%/3% worse. This holds using broadcast or reductions. Let's play with this a bit more to have the loop scheduled in a la Composable Kernels and see where this brings performance(?). Just to be clear, this is what I am running:
```
rocmlir-gen -ph -operation gemm -t f32 --arch gfx90a:sramecc+:xnack-  -g 64 -m 1024 -k 1024 -n 384  --transA --transB --kernel-repeats 5 --perf_config=128,128,4,64,64,8,1,1 -t f16 -ph  | rocmlir-driver -c  | ${ROCMLIR_PATH}/mlir/utils/widgets/rocm-run
```

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
jerryyin: I'm not sure I get your question. I believe this `mOffset` is only used in the non-reduction branch below.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
jerryyin: Agreed, once your PR merges, we can add this additional offset from `mOffset` to ensure the added case will be enabled.

I will point out that as the code base in `develop` branch today, leaving this as should not break any pre-existing config (because we wouldn't select reduction case with multiple repeats).

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
jerryyin: This all needs to be converted to transforming for and coordinate transforms. Right now it stays at a polished prototype level for performance evaluation.

jerryyin:
Closing as this is superseded by #976

PR #943, opened 2023-01-18
Reenable swizzles using if and not select

krzysz00:
Will check perf implications on a Lockhart before this lands.


sjw36 (commented):
Good to check perf. Thread conditional code is generally poor perf compared to select.

krzysz00:
Closed because we're not doing swizzles anyway

PR #944, opened 2023-01-18
Add rock.threadwise_read_into for a cleaner fusion flow

krzysz00:
Also fix up some assembly formats and add side effect annotations (thanks to Manupa for reminding me we have those in his reduce PR).


sjw36 (approved):
Looks good. Thanks.

PR #945, opened 2023-01-19
Fixing redundant archive by moving to the reports folder

jerryyin:
The jenkins HTML publisher plugin has the side effect of archiving too much (undocumented from its [reference page](https://www.jenkins.io/doc/pipeline/steps/htmlpublisher/)). This entire build folder was archived, causing disk running out of space.

After a couple of attempt to fix it, I figured that the easy way is to give it a folder where no redundant build files exists. @okakarpa confirmed in a test run that this will no longer store the useless intermediate files and make nightly archive size fall to reasonable size.

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -131,7 +131,7 @@ void postProcessPerfRes(String chip) {
         allowMissing: false,
         alwaysLinkToLastBuild: false,
         keepAll: true,
-        reportDir: 'build',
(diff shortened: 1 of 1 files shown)


yiqian1 (approved):

yiqian1:
Nice. I've noticed that publishHTML took quite a long time to finish and was wondering why.

jerryyin:
@yiqian1 Yes, how could anyone knew they'd choose to archive everything in the path... It's just bad plugin design choice.

PR #946, opened 2023-01-23
[DO NOT SQUASH] Remove certain CMake crimes around CMAKE_PREFIX_PATH

krzysz00:
HIP needs /opt/rocm in CMAKE_PREFIX_PATH, setting that globally is bad practice, compromise by temporarily setting it just for finding HIP.


jungpark-mlir (approved):
Makes a sense to me.

PR #947, opened 2023-01-24
Adding a Composable Kernel driver

giuseros:
This is adding a Composable Kernels benchmark driver that should ease our comparison with this library. 

- Added the main C++ driver
- Updated the `JenkinsFile` to build `Composable Kernels` (CK)
- Updated the `perfRunner.py` to add the possibility to run against `CK`


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -790,6 +808,20 @@ pipeline {
                             }
                         }
                     }
+                    stage("Build Composable Kernels") {
```
manupak, line 814: If you look above, this runs when a nighlty param is set.
So in Jenkins, you would need to re-run the job with nighlty enabled to test this..

You can make sure looking at the logs to see whether the stage has executed.

mlir/utils/performance/common/common_utils.h
```diff
@@ -0,0 +1,152 @@
...
+  case DataType::F16:
+    return llvm::APFloat::S_IEEEhalf;
+  case DataType::BF16:
+    return llvm::APFloat::S_BFloat;
+  case DataType::I8:
+    assert(0 && "Can't have i8 floats");
+  }
+}
+
+static void *allocAndFill(DataType dataType, size_t byteSize, bool isOut) {
```
manupak: I think it would be good to have docs for common utils: here and other functions.

mlir/utils/performance/perfRunner.py
```diff
@@ -653,10 +698,16 @@ def benchmarkMLIR(commandLine, confClass, paths: Paths, arch, tuningDb: MaybeTun
...
+<<<<<<< HEAD
 #Generate MLIR vs. MIOpen or rocBLAS performance results
 def generatePerformanceResults(configs, confClass, paths: Paths, arch, tuningDb: MaybeTuningDb, rocmlir_gen_flags):
     # Never pass tuning DB to this run
     mlir_df = pd.DataFrame(benchmarkMLIR(testVector.split(sep=' '), confClass, paths, arch, None, rocmlir_gen_flags)
+=======
+#Generate MLIR vs. MIOpen or rocBLAS or CK performance results
+def generatePerformanceResults(configs, confClass, paths: Paths, arch, rocmlir_gen_flags):
+    mlir_df = pd.DataFrame(benchmarkMLIR(testVector.split(sep=' '), confClass, paths, arch, rocmlir_gen_flags)
+>>>>>>> 0ca6a3cf3ad1... Adding a Composable Kernel driver
```
manupak: Fix

manupak (commented):
I just did a quick skim (prolly leave it to someone familiar with this part of the code for a detailed look).

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -790,6 +808,20 @@ pipeline {
...
+                        steps {
+                            sh 'rm -rf composable_kernel'
+                            dir('composable_kernel') {
+                                getAndBuildCK('''
+                                    -DGPU_TARGETS="gfx1030;gfx908;gfx90a"
```
krzysz00: Perhaps `-DCMAKE_BUILD_TYPE=Release` unless that's already being set above and doesn't imply `-O3` for some reason

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -23,6 +23,18 @@ void buildMIOpen(String cmakeOpts) {
     sh 'cd build; make -j $(nproc) MIOpenDriver'
 }
 
+void buildCK(String cmakeOpts) {
+    sh '[ ! -d build ] || rm -rf build'
+    cmakeBuild generator: 'Unix Makefiles',\
+        buildDir: 'build',\
+        buildType: 'Release',\
+        installation: 'InSearchPath',\
+        cmakeArgs: """-DCMAKE_CXX_COMPILER=/opt/rocm/bin/hipcc
```
krzysz00: Why `hipcc`? `hipcc` is deprecated.

mlir/utils/performance/ck-benchmark-driver/ck-benchmark-driver.cpp
```diff
@@ -0,0 +1,224 @@
+//===- ck-benchmark-driver.cpp - Composable Kernel benchmark driver
+//--------===//
```
krzysz00: I'd shorten to `CK benchmark driver` and add a longer comment down below so that you don't get header comments spread onto two lines. Or just remove some `-` from the second part.

mlir/utils/performance/common/common_utils.h
krzysz00: ... also, why are we putting a bunch of code into header files? Can we make a `benchmarkUtils.cpp` that both the rocblas and ck benchmark drivers link to? (Like, obviously the `#define` needs to be in a header but still)

Or at least if you're going to do this sort of thing, `.cpp.inc` would make it clearer what the deal is.

mlir/utils/performance/perfRunner.py
```diff
@@ -766,7 +814,7 @@ def foundExternalTool(paths: Paths, opType: Operation):
     if opType == Operation.CONV and not paths.miopen_driver_path:
         return False
     if opType == Operation.GEMM and \
-            (not paths.mlir_paths or not paths.mlir_paths.rocblas_benchmark_driver_path):
+            (not paths.mlir_paths and (not paths.mlir_paths.rocblas_benchmark_driver_path or not paths.mlir_paths.ck_benchmark_driver_path)):
```
krzysz00: Maybe instead pass in `Optional[GEMMLibrary]` here to make it a more precise check.

krzysz00 (commented):
Overall, seems reasonable, just a few nits.

But also, can you make the "build CK" stage not break the build if it fails? I don't want our CI getting held up by breakage in CK.

mlir/utils/jenkins/Jenkinsfile
giuseros: Corrected it. I thought that hipcc was the same as clang these days (e.g., ., if I do `hipcc --help` I get the `clang` help)

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -23,6 +23,18 @@ void buildMIOpen(String cmakeOpts) {
     sh 'cd build; make -j $(nproc) MIOpenDriver'
 }
 
+void buildCK(String cmakeOpts) {
+    sh '[ ! -d build ] || rm -rf build'
+    cmakeBuild generator: 'Unix Makefiles',\
+        buildDir: 'build',\
+        buildType: 'Release',\
+        installation: 'InSearchPath',\
+        cmakeArgs: """-DCMAKE_CXX_COMPILER=/opt/rocm/llvm/bin/clang++
```
krzysz00, line 32: I'd add CMAKE_C_COMPILER too just to be safe

mlir/utils/performance/common/benchmarkUtils.h
```diff
@@ -0,0 +1,95 @@
...
+// Allocate and fill an area of memory
+void *allocAndFill(DataType dataType, size_t byteSize, bool isOut);
+
+// Return sizeof(dataType) in bytes
+size_t getByteSize(DataType dataType, size_t elems, bool isOut);
+
+// Allocate a device buffer and copy the date from host memory
+void *getGpuBuffer(const void *hostMem, size_t byteSize);
+
+#endif // MLIR_UTILS_PERFORMANCE_COMMON_BENCHMARKUTILS_H
```
krzysz00: Nit: missing trailing newline

mlir/utils/performance/ck-benchmark-driver/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+  set(LLVM_LINK_COMPONENTS
+    Support)
+  
+  set(LIBS ${composable_kernel_LIBRARY})
+
+  set(EXCLUDE_FROM_ALL TRUE)
+  add_llvm_executable(ck-benchmark-driver
+
+    ck-benchmark-driver.cpp
+    ../common/benchmarkUtils.cpp
```
krzysz00: Nit: could I get `benchmarkUtils.cpp` to be its own library? That way we're not compiling it twice.

krzysz00 (commented):
We're now in the nitpick zone!

giuseros:
I fixed some bugs and applied the changes. The hard part for me is to make this pass in our CI

giuseros:
Since CK is using RTTI features, in order to have an executable that links against LLVM and CK we need to build with `LLVM_ENABLE_RTTI=ON`. I added that and also a bit of bash code to only build (and install) the library (by default CK would install everything, meaning up to 1hr of compilation time)

giuseros:
After a chat with @krzysz00 , we think it is better to create a common set of utilities that don't call into LLVMSupport so that we are also future proof if we ever want to try on other libraries that might use RTTI

giuseros:
Hi @krzysz00 , I made all the benchmark code LLVM independent. Please, let me know what you think!

mlir/utils/performance/common/benchmarkUtils.cpp
```diff
@@ -0,0 +1,254 @@
...
+  case DataType::BF16:
+    for (auto flt : patternFlt) {
+      ushort bf16flt = float_to_bfloat16(flt);
+      auto *p = reinterpret_cast<unsigned char const *>(&bf16flt);
+      res.push_back(p[0]);
+      res.push_back(p[1]);
+    }
+    break;
+  case DataType::I8:
+    for (auto i : patternFlt) {
```
krzysz00: `patternInt`, I'm pretty sure.

mlir/utils/performance/common/benchmarkUtils.cpp
```diff
@@ -0,0 +1,254 @@
...
+    std::string arg = argv[i];
+    if (arg == "-g") {
+      res.gemmG = atoi(argv[++i]);
+    } else if (arg == "-m") {
+      res.gemmM = atoi(argv[++i]);
+    } else if (arg == "-k") {
+      res.gemmK = atoi(argv[++i]);
+    } else if (arg == "-n") {
+      res.gemmN = atoi(argv[++i]);
+    } else if (arg == "--transposeA") {
```
krzysz00: These are usually passed as `--transposeA=true` or `--transposeA=false` or the like, I think.

mlir/utils/performance/common/benchmarkUtils.cpp
```diff
@@ -0,0 +1,254 @@
...
+      res.transposeA = true;
+    } else if (arg == "--transposeB") {
+      res.transposeB = true;
+    } else if (arg == "-t") {
+      res.dataType = parseDataType(argv[++i]);
+    } else if (arg == "--perf-config" || arg == "--arch") {
+      i++;
+    } else {
```
krzysz00, line 210: Can we get error reporting for unknown arguments?

krzysz00 (commented):
I'm trusting your float conversion functions.

But I like this setup - I'm now nitpicking a bunch.

I'd also add a comment explaining why we don't use `APFloat` and company.

giuseros:
I applied your changes, and also pointed out where the conversion math came from :)

mlir/utils/performance/ck-benchmark-driver/ck-benchmark-driver.cpp
```diff
@@ -0,0 +1,232 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+// Include common utility functions
+#include "../common/benchmarkUtils.h"
+
+// CK includes
+#include "ck/ck.hpp"
+// CK __noinline__ definition clashes with the one in libc
+#undef __noinline__
```
krzysz00, line 21: I don't think we need this anymore.

krzysz00 (approved):
This looks pretty good to me, aside from one last note

mlir/utils/performance/ck-benchmark-driver/ck-benchmark-driver.cpp
giuseros: If I remove that and recompile, I get errors like:
```
..../include/c++/12/bits/shared_ptr_base.h:196:22: error: use of undeclared identifier 'noinline'; did you mean 'inline'?
```
Apparently CK is defining its own `__noinline__` but the attribute `noinline` is not present in libc (this was my interpretation of the issue).

krzysz00: Huh, weird. Could be a question of what C++ standard is being used.

Given that, I'd keep the `#undef` and just make confused noises about it.

Maybe ask the CK people if they know what's up

PR #949, opened 2023-01-25
Adds support for rock.reduce (sum) fusion

manupak:
This commits adds support for AlignTiling pass to
fold tailing rock.reduce ops to gemm-like ops.

* Added unit tests
* Added integration tests.

closes: https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/773


krzysz00:
I think this might want to wait on @sjw36's rework?

manupak:
Do we know the timeline for that ?

It is just a tailing rewrite of reduce op into the global store (note that not yet extending support of inbetween -- i.e. between the linalg block or store loop and the reduce) -- view-like ops or transposes). So I thought the "rework" is done prior to this to handle view-like/transposes affecting the linalg block.

(I ll do a seperate ticket to do the same to support reduce as well once the "rework" is done)

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -444,10 +452,63 @@ static bool isUnfusedKernelStore(rock::GlobalStoreOp store) {
...
+  Type elementType = globalStoreOp.getSource().getType().getElementType();
+  Value zero = rewriter.create<arith::ConstantIndexOp>(loc, 0);
+  int64_t reductionAxis = reduceOp.getAxisAttr().getInt();
+  int64_t inpRank = reduceOp.getIn().getType().cast<ShapedType>().getRank();
+  if (elementCount != 1 && reductionAxis == inpRank - 1) {
+    // if the reduction dimension is the last dimension and its a vector.
+    OpBuilder::InsertionGuard guard(rewriter);
+    rewriter.setInsertionPoint(globalStoreOp);
+    VectorType loadedElType = VectorType::get({elementCount}, elementType);
+    Value loadVal = rewriter.create<vector::LoadOp>(
```
krzysz00: I don't think this is correct? Consider the possibility that the vector length is 4 and the output is a NxKx2x2 tensor. Adding each thread's values together will lead to too much reduction.

manupak: right! yes -- vector length exceeding product of lowest dimSizes.. I ll add a test and a fix.

krzysz00: I wouldn't waste too much time. In the fusion refactor (which is a few days out, hopefully - or at least it feels like it's going places) - the vectorization length won't be decided yet when we get to align_tiling.

manupak:
OK, I've rebased.

Yes, the new system is much easier :)

manupak:
@krzysz00 when you have some time, I'd appreciate another look here...

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -373,11 +385,57 @@ static bool isUnfusedKernelStore(ThreadwiseWriteAllOp store) {
...
+  }
+  TransformMapAttr trAttr = dropReductionDim.get();
+  {
+    OpBuilder::InsertionGuard guard(rewriter);
+    rewriter.setInsertionPoint(threadwiseWriteOp);
+    redIn = rewriter.create<TransformOp>(loc, redIn, trAttr);
+    threadwiseWriteOp.getDestMutable().assign(reduceOp.getOut());
+    threadwiseWriteOp.setStoreMethodAttr(
+        StoreMethodAttr::get(rewriter.getContext(), StoreMethod::AtomicAdd));
+    auto extraViews = extractVector(threadwiseWriteOp.getExtraViews());
```
krzysz00: I'm not sure this is correct. `extraViews` are applied before the transformation chain on the `Value`. You may want to transform `reduceOp.getOut()`

mlir/test/fusion/e2e/reductions/gfx908/lit.local.cfg
```diff
@@ -0,0 +1,2 @@
+if not "gfx908" in config.arch:
```
krzysz00: Why gfx908? Why not use the `enable_xdlops_e2e_tests` or however it's spelled?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -373,11 +385,57 @@ static bool isUnfusedKernelStore(ThreadwiseWriteAllOp store) {
...
+  return llvm::to_vector<4>(arrayAttr.getAsRange<Attribute>());
+}
+
+LogicalResult
+ReduceRewritePattern::matchAndRewrite(rock::ReduceOp reduceOp,
+                                      PatternRewriter &rewriter) const {
+  Location loc = reduceOp.getLoc();
+  ThreadwiseWriteAllOp threadwiseWriteOp =
+      traceToThreadwiseWrite(reduceOp.getIn());
+  if (reduceOp.getReduceMethod() != ReduceMethod::Sum) {
```
krzysz00, line 543: I think this is a `matchFailure()`

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -373,11 +385,57 @@ static bool isUnfusedKernelStore(ThreadwiseWriteAllOp store) {
...
+    return reduceOp.emitError("We only support sum reductions.!");
+  }
+  if (!threadwiseWriteOp) {
+    LLVM_DEBUG(llvm::dbgs() << "rock.reduce input is not leading to "
+                               "ThreadwiseWriteAllOp of the previous op.");
+    return failure();
+  }
+  int64_t reductionAxis = reduceOp.getAxisAttr().getInt();
```
krzysz00, line 567: So as to avoid fusing multiple reductions, check to make sure the store method is either `set` or `atomic_add` up here.

mlir/test/fusion/e2e/reductions/gfx90a/rock-gemm-reduce-case2.mlir
```diff
@@ -0,0 +1,47 @@
+// This test checks a case where reduction is happening in a middle dimension
+
+// RUN: cat %s | rocmlir-gen -ph -print-results -fut test_reduce -verifier clone - | rocmlir-driver -host-pipeline xmodel -kernel-pipeline full | xmir-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_c_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CLONE
```
krzysz00: Useless use of `cat`, either use `<` or pass in `%s` as an argument to rocmlir-gen

mlir/test/fusion/e2e/reductions/gfx90a/rock-gemm-reduce-case1.mlir
```diff
@@ -0,0 +1,47 @@
+// This test checks a case where reduction is happening in fastest changing dimension where vectorization is present
```
krzysz00: Suggestion: Use `-host-pipeline=xmodel,runner` and then `mlir-cpu-runner`. 

But also, to make this not hardcode a target name, use the arch name we have floating around the tests and do a `sed -e 's/###GPU_ARCH###/%name_of_the_variable/g` on the input file before passing it to `rocmlir-gen`. This'll also let you set `xmodel.arch`

mlir/test/fusion/e2e/reductions/gfx90a/rock-gemm-add-reduce.mlir
```diff
@@ -0,0 +1,62 @@
+// This test checks a case where reduction is happening in a middle dimension
+
+// RUN: cat %s | rocmlir-gen -ph -print-results -fut test_reduce -verifier clone - | rocmlir-driver -host-pipeline xmodel -kernel-pipeline full | xmir-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_c_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CLONE
```
krzysz00: See comments on test below.

mlir/test/fusion/e2e/reductions/gfx908/lit.local.cfg
krzysz00: Ok, yeah, instead of duplicating the tests, `sed` in the architecture name.

krzysz00 (commented):
I've got a correctness concern and I'm iffy about how you've handled the tests

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Can you elaborate a bit more on the difference?
i.e. taking an extraView on output memref vs transforming output memref and passing that in..

krzysz00: Strictly speaking, there's no difference. Threadwise_write_all writes to `outputCoords := viewsOnArgument(extraViews(bid, tid, iter))`

Conceptually, threadwise_write_all always takes an argument that's in GEMM coordinates, and I want to keep it that way.

manupak:
Thanks for sed based suggestion on injecting arch. I had to use cat because rocmlir-gen complained it cant accept two positional args ('-') to pipe in but I can use <

Im just a bit hesitant of not using xmir-runner because of past problems with clone verification flow. If you can point me at a working example that uses 'partition' flow on tosa based test (i.e. arch not put in) AND clone verification, I may give it a try.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: done.!

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: I think failing to match the pattern(rewrite) doesn't mean some other rewriter is going to take care of that in this case. it will just make the error more obscure...

because at this stage this failure means compilation failure. I think we should fail here if it comes here because we are already committed to fusion by now.

manupak (commented):
Thanks for the reviews @krzysz00! I think I've addressed all the comments.

manupak:
This is now rebased and good for another look....
@krzysz00 @sjw36 when you got a minute.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
krzysz00: Makes sense

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -521,11 +528,74 @@ static bool isUnfusedKernelStore(ThreadwiseWriteAllOp store) {
...
+    threadwiseWriteOp = static_cast<ThreadwiseWriteAllOp>(
+        rewriter.clone(*threadwiseWriteOp.getOperation()));
+  }
+  int64_t reductionAxis = reduceOp.getAxisAttr().getInt();
+  TypedValue<ShapedType> redIn = reduceOp.getIn();
+  ArrayRef<int64_t> reduceInShape = redIn.getType().getShape();
+  TopDownTMBuilder dropReductionDim(rewriter, reduceInShape, loc);
+  for (uint32_t i = 0; i < reduceInShape.size(); ++i) {
+    if (i == reductionAxis) {
+      dropReductionDim.constDim("reduction_dim", i, 0, 1);
```
krzysz00: Is `constDim` the right operation here? I'd think it might be `broadcast`?

The point of this map, as far as I understand it, is to map `(..., dR, ...) -> (... 0, ...)`, aka, it's the map where `dR` was broadcast.

krzysz00 (commented):
Overall, the code here seems reasonable.

Are we already set up to hit this pattern?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: It is a fine line...

Id agree with broadcast if it was a load.

However, in a reduction store, we just force dR to be zero -- so constDim made slightly bit more sense..

krzysz00: A broadcast does force the dimension to 0

I think a broadcast is more semantically correct and might help the vectorizer out.

manupak: Ok, I ll buy the vectorizer argument if thats the case.

krzysz00: Basically, the contiguous-dimensions checker understands broadcasts, but doesn't understand not using an input and then defining a 0 - it won't know the relationship between the argument and the result.

It'd take a somewhat contrived scenario for this to actually come up - mainly a reduction of length 1 - but still.

manupak: ok done!

manupak:
@krzysz00 I've changed it to broadcast now. PTAL

krzysz00 (approved):
Seems reasonable to me

PR #951, opened 2023-01-26
Add ALLOW-RETRIES: 2 to e2e tests to account for driver problems

krzysz00:
Sometimes, a perfectly correct kernel will throw an intermittent page fault because the hardware is just having a bad day or something. If you run the same kernel again, it'll often work. Until we've fixed the underlying driver trouble, only mark e2e tests as failed if they fail twice.


zhanglx13:
I did not find any information about `allow retries` but only one mention in the lit doc. So I have two questions:
1. Is it `ALLOW-RETRIES` or `ALLOW_RETRIES` or they both work?
2. For test files with more than one `RUN:`, do you need one `ALLOW_RETRIES` for each or you just need one for the whole file?

krzysz00:
1. Good catch on the spelling
2. Since all `//RUN` lines are collected into one script, and since I checked the source code, it looks like you only need one `ALLOW_RETRIES`, and, in fact, you can't have more than one.

zhanglx13 (approved):
LGTM.
Maybe you want to try several runs of nightly CIs to see if this kills the seg fault issue.

PR #952, opened 2023-01-26
Add ConstDim operator

krzysz00:
This operator takes no inputs (no dimensions in the view) and adds constant-valued dimensions into the output (underlying tensor/next view).

The operator is parameterized by {constant_value, dimension_length, constant_value, dimension_length, ...}.

The point of all this is to allow us to nicely represent certain maps, like "() -> (0, 0, 0, 0)", that you find in various fusion scenarios.

A commit that uses this may or may not be upcoming.

### DIFF
--- a/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
+++ b/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
@@ -56,8 +56,9 @@ static void propagateTransformOob(TransformMapAttr transformMap,
     }
     case TransformType::PassThrough:
     case TransformType::Slice:
-    case TransformType::AddDim: {
-      // Zip ends at end of shortes array, allowing addDim here
+    case TransformType::AddDim:
+    case TransformType::ConstDim: {
+      // Zip ends at end of shortes array, allowing addDim and constDim here
       for (auto pair : llvm::zip(upperDims, lowerDims)) {
         uint32_t upper = std::get<0>(pair);
         uint32_t lower = std::get<1>(pair);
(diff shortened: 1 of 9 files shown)


mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -467,7 +468,11 @@ ContiguousMergesMap findContiguousGroups(ArrayAttr transforms,
           thisDimToMerge[lowerDims[i]] = {transformMap, transform, i, false};
         }
         break;
+      // AddDim drops dimensions down a hole, while ConstDim conjures them
+      // from nowhere. In either case, there is no merge that can be associated
+      // with them.
       case TransformType::AddDim:
+      case TransformType::ConstDim:
```
giuseros, line 475: Shouldn't that also behave like a broadcast? I mean what happens if I have:
`Merge{8,3} -> ConstDim{1,5} -> Unmerge{8,5,3}` 
Now the Unmerge "thinks" that dim 0 and 2 are separate, while it's only an illusion

mlir/test/Dialect/Rock/index_diff_update.mlir
```diff
@@ -27,6 +27,11 @@
         #rock.transform<PassThrough ["y"] at [1] -> ["y"] at [0]>]
     bounds = [16, 4] -> [4]>
 
+#transform_map6 = #rock.transform_map<affine_map<(d0) -> (d0, 1)>
+    by [<PassThrough ["x"] at [0] -> ["x"] at [0]>,
+        <ConstDim{1, 8} [] at [] -> ["y"] at [1]>]
```
giuseros, line 32: Does `ConstDim{1,8}` mean add a constant dimension at 1 with size 8?

krzysz00: Nope. It means "add constant dimension with value 1 and length 8"

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
giuseros: After a chat we agreed that:
- If we add a unit dimension, the case is covered already
- If we add a non-unit dimension it should not work (and indeed it won't work)

mlir/test/Dialect/Rock/index_diff_update.mlir
giuseros: Yes, that makes sense. Thanks!

giuseros (approved):
LGTM!

PR #955, opened 2023-01-28
Fix address-sanitiser errors and a compile warning

pcf000:
When debugging something of mine, I built with the address sanitiser (-DLLVM_USE_SANITIZER=Address) and stumbled across several errors, which are fixed here.  Also fixed one compiler warning because it bugged me.

Still one other compiler warning, and still one ASAN error in TosaPartition.cpp that I'm working on eliminating while making the pass generic.

### DIFF
--- a/mlir/include/mlir-c/Dialect/Rock.h
+++ b/mlir/include/mlir-c/Dialect/Rock.h
@@ -88,8 +88,9 @@ void mlirRockTuningTableDestroy(MlirRockTuningTable table);
 // the best performing tuning parameter to simplify the underlying
 // implementation, which can be revisited in the future.
 MLIR_CAPI_EXPORTED
-bool mlirRockTuningUpdateTable(MlirRockTuningTable perfTable, char *probCStr,
-                               char *perfCStr, float time);
+bool mlirRockTuningUpdateTable(MlirRockTuningTable perfTable,
+                               const char *probCStr, const char *perfCStr,
+                               float time);
 
 // Search the tuning table and get the stored best value for the given problem.
 // The definition of the tuning problem is internally described and opaque to
--- a/mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
@@ -223,14 +223,13 @@ def Rock_GeneralGemmParamsAttr : Rock_Attr<"GeneralGemmParams", [RockTuningParam
 
   let extraClassDeclaration = [{
     void getPerfConfigStr(std::string &perfStr){
-      const Twine paramsConcat =
-        Twine(getBlockSize()) + ","
+      perfStr =
+       (Twine(getBlockSize()) + ","
       + Twine(getMPerBlock()) + ","
       + Twine(getNPerBlock()) + ","
       + Twine(getKPerBlock()) + ","
       + Twine(getMPerThread()) + ","
-      + Twine(getNPerThread());
-      perfStr = paramsConcat.str();
+      + Twine(getNPerThread())).str();
     }
   }];
 
@@ -258,16 +257,16 @@ def Rock_XdlopsGemmParamsAttr : Rock_Attr<"XdlopsGemmParams", [RockTuningParamAt
 
   let extraClassDeclaration = [{
     void getPerfConfigStr(std::string &perfStr){
-      const Twine paramsConcat = 
-          Twine(getMPerBlock()) + ","
+      perfStr =
+         (Twine(getMPerBlock()) + ","
         + Twine(getNPerBlock()) + ","
         + Twine(getKPerBlock()) + ","
         + Twine(getMPerWave()) + ","
         + Twine(getNPerWave()) + ","
         + Twine(getKpack()) + ","
         + Twine(getForceUnroll()) + ","
-        + "1"; /* *ThreadCopyMore* */
-      perfStr = paramsConcat.str();
+          + "1") /* *ThreadCopyMore* */
+        .str();
     }
   }];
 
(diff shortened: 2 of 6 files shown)


mlir/include/mlir-c/Dialect/Rock.h
```diff
@@ -88,8 +88,9 @@ void mlirRockTuningTableDestroy(MlirRockTuningTable table);
 // the best performing tuning parameter to simplify the underlying
 // implementation, which can be revisited in the future.
 MLIR_CAPI_EXPORTED
-bool mlirRockTuningUpdateTable(MlirRockTuningTable perfTable, char *probCStr,
-                               char *perfCStr, float time);
+bool mlirRockTuningUpdateTable(MlirRockTuningTable perfTable,
+                               const char *probCStr, const char *perfCStr,
```
pcf000, line 92: Compiler warning came from passing a const-char-* to a char-\*.  This function passes it to another const-char-\* anyway, just add the const here.

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
```diff
@@ -223,14 +223,13 @@ def Rock_GeneralGemmParamsAttr : Rock_Attr<"GeneralGemmParams", [RockTuningParam
...
-      const Twine paramsConcat =
-        Twine(getBlockSize()) + ","
-      + Twine(getMPerBlock()) + ","
-      + Twine(getNPerBlock()) + ","
-      + Twine(getKPerBlock()) + ","
-      + Twine(getMPerThread()) + ","
-      + Twine(getNPerThread());
-      perfStr = paramsConcat.str();
+      perfStr =
+        std::to_string(getBlockSize()) + ","
```
pcf000: The Twines were making temps on the stack and retaining references to them in the result that was passed to perfStr.  I'm not sure exactly how, but the comments do say that Twine is intended for temporary use.  I made them all std::string because I know how that behaves.

mlir/include/mlir/Dialect/Rock/IR/TransformMapBuilder.h
```diff
@@ -44,8 +44,8 @@ class TransformMapBuilder {
   void getEndNames(SmallVectorImpl<StringRef> &names);
   void getStartNames(SmallVectorImpl<StringRef> &names);
 
-  SmallString<8> startName(uint32_t dim);
-  SmallString<8> endName(uint32_t dim);
+  StringRef startName(uint32_t dim);
```
pcf000, line 47: BottomUpTMBuilder::broadcast() called startName() to acquire a string which it saved in a vector of StringRef.  The call constructed a temp SmallString for the return value, which went away when the scope was exited, leaving the StringRef dangling.  It appears to me that it didn't really want a string copy, just a pointer into TransformMapBuilder's startNames collection.

mlir/test/CAPI/mixr_full.c
```diff
@@ -215,7 +215,7 @@ static bool constructAndTraverseIr(MlirContext ctx) {
   // ranks, dimensions of each arguments and kernel name.
   int argSize = 0;
   mlirGetKernelInfo(module, &argSize, NULL);
-  void *argInfo = malloc(argSize);
+  void *argInfo = malloc(argSize+1);
```
pcf000: Printing the kernel name was getting a heap overflow, because there was no zero terminator on the string in argInfo and the trailing data happened to be nonzero.  Let's allocate one extra byte and force a zero in there.

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
krzysz00: That should be "all those twines in parens" then (`to_str()`)

Or changing that `std::string&` to `SmallVectorImpl<char>&` and `to_vector()` on `Twine`.

`Twine` is important to prevent all the `std::string` stuffles.

mlir/include/mlir/Dialect/Rock/IR/TransformMapBuilder.h
krzysz00: Yeah, reasonable

krzysz00 (commented):
I have comments on what should be done to the twine.

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
pcf000: Did go with an "all Twines in parens" then .str() strategy, but also changed the function to return a value instead of update a reference parameter

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
```diff
@@ -222,15 +222,14 @@ def Rock_GeneralGemmParamsAttr : Rock_Attr<"GeneralGemmParams", [RockTuningParam
   );
 
   let extraClassDeclaration = [{
-    void getPerfConfigStr(std::string &perfStr){
-      const Twine paramsConcat =
+      std::string getPerfConfigStr(){
```
krzysz00: I think the reference parameters were used to allow for populating the perf config repeatedly without reallocation. Perhaps changing from `std::string &` to `SmallVectorImpl<char> &` would also be worth considering.

In any case, @jungpark-mlir .

krzysz00: (But if we don't hear anything soon I'm willing to approve)

pcf000: Changed it back to something similar to the original, but with all the Twines inside one set of parens and fed to a .str() call.  Apparently they manage to deallocate themselves between the assignment and the use in the original code.

krzysz00 (approved):
LGTM!

PR #958, opened 2023-01-30
Adding script to convert from hsaco to StringAttribute

jerryyin:
It took me some effort to figure out what the actual text format of our elf binary is. I rewrite the serialization functions into a standalone python program. With this script, the developer can easily manipulate the populated IR or ISA and then re-inject it back to the same host code and run using the jit runner.

This script take two arguments:
 - -i where you specify the hsaco file
 - Optional -o where by default it will generate a file ended with `.attr`
 - In executing this script, it will also print out the type of the string attribute such that a developer can conveniently swap the type from the printed mlir global type

### DIFF
--- a/mlir/utils/widgets/hsacoToStringAttr.py
+++ b/mlir/utils/widgets/hsacoToStringAttr.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python3
+import argparse
+import os
+
+# https://github.com/llvm/llvm-project/blob/dc37dc824aabbbe3d029519f43f0b348dcad7027/llvm/include/llvm/ADT/StringExtras.h#L125-L128
+def isPrint(C: str) -> bool:
+    UC = ord(C)
+    return 0x20 <= UC <= 0x7E
+
+# https://github.com/llvm/llvm-project/blob/dc37dc824aabbbe3d029519f43f0b348dcad7027/llvm/lib/Support/StringExtras.cpp#L62-L71
+def printEscapedString(data, out):
+    for val in data:
+        c = chr(val)
+        if c == '\\':
+            out.write('\\' + c)
+        elif isPrint(c) and c != '"':
+            out.write(c)
+        else:
+            out.write('\\' + hex(ord(c) >> 4)[2:] + hex(ord(c) & 0x0F)[2:])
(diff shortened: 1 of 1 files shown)


mlir/utils/widgets/hsacoToStringAttr.py
```diff
@@ -0,0 +1,48 @@
...
+    parser.add_argument("-i", help="Input hsaco kernel file", required=True)
+    parser.add_argument("-o", help="Output kernel text file", default=None)
+
+    args = parser.parse_args()
+    return args
+
+def main(args):
+    if (args.o == None):
+        args.o = args.i.rsplit('.', maxsplit=1)[0] + ".attr"
+    print("Converting from " + args.i + " to " + args.o)
```
krzysz00: I'd lose this automated message - it'll clutter up pipelines if people want to do this in a script.

krzysz00 (approved):
Nit, but approved

mlir/utils/widgets/hsacoToStringAttr.py
jerryyin: Fair enough, let me get rid of this and use a comment instead.

krzysz00:
As a further thought, could you have this do a thing where if you give it enough surrounding context, it'll print out something you can easily copy-paste into MLIR? Instead of having to pull the type name in separately?

krzysz00:
(the symbol name might be enough there?)

jerryyin:
>  it'll print out something you can easily copy-paste into MLIR? Instead of having to pull the type name in separately?

Unfortunately the type name show up in more than one place so it needs the entire IR context fed to it (and sed swap the type name).

For example:

> %33 = llvm.mlir.addressof @rock_gemm_module_gpubin_cst : !llvm.ptr<array<5888 x i8>>
    %34 = llvm.getelementptr %33[0, 0] : (!llvm.ptr<array<5888 x i8>>) -> !llvm.ptr<i8>

You are talking about the following use case?

Piping the jit host code + assembly as argument, then it populate the updated jit host code that can be used in jit runner directly?

krzysz00:
Right, the type shows up in a few places. Never mind my thought, then. The script is pretty useful as is.

PR #960, opened 2023-02-01
Improved handling of out of bounds checks to account for more complex scenarios

krzysz00:
(Note to merger: take this description as part of the commit message, remove this line)

The existing code for out of bounds checks failed in certain cases, namely things that reduce to

```mlir
func.func @example(..., %arg2: memref<15xf32>) {
   %c1 = rock.transform %arg2 by [<Unmerge{5, 3} ["m", "n"] at [0, 1] -> ["raw"] at [0]>] bounds = [5, 3] -> 15
  %c = rock.transform %c1 by [<AddDim>{1} ["g"] at [0]>, <Pad{0, 27} ["mPad"] at [1] -> ["m"] at [0]>, <Pad{0, 29} ["nPad"] at [2] -> ["n"] at [1]>] bounds = [1, 32, 32] -> [5, 3]>
  ... store [g, m, n] to %c
```

In this case, the bounds checker would just add a test for `raw >= 15`, which would miss the invalid coordinates `[0, 0, 4]` for example.

Since we've now found examples where this behavior happens,  we need to actually fix it.

The fix is as follows
- Remove the out of bounds inference utility and, as a consequence, `leftOobDims` and `rightOobDims` from the buffer-related load and store ops
- Add a `%valid` parameter to the buffer ops that indicates if a load/store should go through or if it should be sent out of bounds
- Add validity as an implicit output of each iteration domain in a `rock.transforming_for`. When the transforming_for is being lowered, the validity parameter will be set to false if, during the computation, part of the computed coordinates become invalid as an address (ex. if they fall within padding or if they go out of the bounds on an `Embed`)
- If the buffer ops get told something is invalid, they will change the load/store address to just outside the range of the memref, causing the hardware to return 0/drop the write.
- The buffer ops have grown `oobIsOverflow` to allow the elementwise kernels to stay straightforward because they don't need the full bounds check logic and we don't want to burden hte compiler with it.

(PR status: Unit and e2e tests pass, I have initial perf numbers, checking for the possibility of flukes.)


krzysz00:
Ok, so, perf 
[new-oob-perf.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10562670/new-oob-perf.zip). On average, perf impacts are flat or a slight worsening (well within the noise threshold), with the worst perf impacts (to the extent there are such) falling onto backward data.

There were a few 2x decreases in that data, but I haven't yet confirmed that those weren't benchmarking error.

Examination of a backwards data ISA didn't show too many changes other than rearrangement - or at least nothing that could cause a big perf drop. Then again, I didn't look at the exact cases that might have problems.

krzysz00:
Update: those big dips may have been benchmarking flukes, since the rerun gave the comprable

```
152,bwd,f32,gfx90a,kyxc,nhwc,nhwk,256,512,28,28,1024,1,1,1,1,2,2,0,0,,14.39305822648652
153,bwd,f32,gfx90a,kyxc,nhwc,nhwk,256,512,28,28,128,1,1,1,1,1,1,0,0,,16.631562545361664
154,bwd,f32,gfx90a,kyxc,nhwc,nhwk,256,512,28,28,256,1,1,1,1,1,1,0,0,,24.28515416975539
```

krzysz00:
There's also a version of this I cooked up that uses `scf.if` to try to contain computations to only where they're valid. Compared to this branch, it seems to be about even, with a few improvements on the benchmarks that are more dominated by address computations. I'll leave that for a future PR. (quick ISA checks show that even though we're emitting an `if`, it gets compiled out to various conditional operators).

jerryyin (commented):
I think I'll have to abstain from this review as it is a lot of changes in a section of code I'm not familiar about. Thinking about this, it'd be a great topic if you can tell more about the implementation details of out of bounds handling in a design meeting.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -856,9 +699,87 @@ int64_t mlir::rock::getMaxVectorization(ArrayAttr transforms, uint32_t dim,
...
+        result |= (params[i] != 0 || params[i + 1] != 0);
+      }
+    } else if (type == TransformType::Embed) {
+      result |= embedCanBeInvalid(map, op);
+    }
+  }
+  return result;
+}
+
+Value mlir::rock::updateValidityAfter(OpBuilder &b, Location loc,
```
manupak, line 737: I think docs for these functions would be useful

manupak (commented):
I did have a look and reflects the discussion on the channel broadly.
(Sorry, I m not very familiar with all the code here to reason about some changes)

One thing (maybe you already have that I cant find, if so please point me at it), is I'd like to see tests where two dimensional padding scenario occurs.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
krzysz00: They're in the header

krzysz00:
Re tests for the case that triggered all this: I'm not sure it's particularly straightforward to write such a test, but we will get one by way of the fusion refactor.

sjw36 (approved):
Nice simplification.

PR #961, opened 2023-02-01
Enable BF16 E2E tests

yiqian1:
These bf16 e2e tests have been run with -pv and -pv_with_gpu on lockhart6 and all passed.

### DIFF
--- a/mlir/test/e2e/Resnext101Config.toml
+++ b/mlir/test/e2e/Resnext101Config.toml
@@ -13,7 +13,7 @@ values = ["-fil_layout=gkcyx -in_layout=ngchw -out_layout=ngkhw", "-fil_layout=g
 
 [[axis]]
 name = "data type"
-values = ["f32"]
+values = ["f32", "bf16"]
 prefix = "-t "
 
 ## Resnext101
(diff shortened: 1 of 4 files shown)


mlir/test/e2e/Resnext101Config.toml
```diff
@@ -13,7 +13,7 @@ values = ["-fil_layout=gkcyx -in_layout=ngchw -out_layout=ngkhw", "-fil_layout=g
 
 [[axis]]
 name = "data type"
-values = ["f32"]
+values = ["f32", "bf16"]
```
krzysz00: I'm not sure we want to enable here since this is f32-only unlike the rest of them?

krzysz00 (commented):
Do we want to get the PR tests as well or?

yiqian1:
> Do we want to get the PR tests as well or?

Probably not, as bf16 is not critical at this moment? We're only testing f32 and f16 for PR now.

mlir/test/e2e/Resnext101Config.toml
yiqian1: > I'm not sure we want to enable here since this is f32-only unlike the rest of them?

Removed

krzysz00 (approved):

PR #963, opened 2023-02-06
Fixing bugs coming from PR #941

giuseros:
This PR is fixing 2 bugs stemming from the introduction of reductions in PR #941 . 

1. The first bug is trivial, and it is mostly a typo in the non-reduction code (which is now not much tested as it used to be). This is the simple fix in `mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp`

3. The second bug is nastier. The problem is that, using reductions, the "universal" configuration we used to compute padding (when the `perf_config` was not passed) is not suitable anymore for reductions, because it has `KPerBlock=4 and KPack==1`. This meant that we were padding in the wrong way, which resulted in failures. My solution was to add a KPack set to 4 for the `UniversalParameters` in the XDL case.

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
@@ -41,14 +41,16 @@ struct InitParams {
   int64_t gemmMPerBlock;
   int64_t gemmNPerBlock;
   int64_t gemmKPerBlock;
+  int64_t gemmKPack;
 };
 
 struct InitParamsNonXDL : InitParams, Serializable<InitParamsNonXDL> {
   constexpr InitParamsNonXDL(uint32_t bSize, int64_t mPerBlock,
                              int64_t nPerBlock, int64_t kPerBlock,
                              int64_t mPerThread, int64_t nPerThread)
-      : InitParams{mPerBlock, nPerBlock, kPerBlock}, gemmMPerThread(mPerThread),
-        gemmNPerThread(nPerThread), blockSize(bSize) {}
+      : InitParams{mPerBlock, nPerBlock, kPerBlock, int64_t(1)},
+        gemmMPerThread(mPerThread), gemmNPerThread(nPerThread),
+        blockSize(bSize) {}
   int64_t gemmMPerThread;
   int64_t gemmNPerThread;
   uint32_t blockSize;
@@ -71,8 +73,8 @@ struct InitParamsXDL : InitParams, Serializable<InitParamsXDL> {
                           int64_t kPerBlock, int64_t mPerWave, int64_t nPerWave,
                           int64_t kPack, bool aThreadCopyMoreGemmK,
                           bool bThreadCopyMoreGemmKPack)
-      : InitParams{mPerBlock, nPerBlock, kPerBlock}, gemmMPerWave(mPerWave),
-        gemmNPerWave(nPerWave), gemmKPack(kPack),
+      : InitParams{mPerBlock, nPerBlock, kPerBlock, kPack},
+        gemmMPerWave(mPerWave), gemmNPerWave(nPerWave),
         gemmAThreadCopyMoreGemmK(aThreadCopyMoreGemmK),
         gemmBThreadCopyMoreGemmKPack(bThreadCopyMoreGemmKPack) {}
 
@@ -81,7 +83,6 @@ struct InitParamsXDL : InitParams, Serializable<InitParamsXDL> {
 
   int64_t gemmMPerWave;
   int64_t gemmNPerWave;
-  int64_t gemmKPack;
   bool gemmAThreadCopyMoreGemmK;
   bool gemmBThreadCopyMoreGemmKPack;
 
--- a/mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
@@ -185,7 +185,7 @@ private:
   // if can't select config from above , use this config to do
   // padding kernel for example , GemmK/block is 16 , if your gemmK is  13 , we
   // add more 3 gemmk
-  static const InitParams universalParameters;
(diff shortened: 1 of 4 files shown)


mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
```diff
@@ -41,14 +41,16 @@ struct InitParams {
   int64_t gemmMPerBlock;
   int64_t gemmNPerBlock;
   int64_t gemmKPerBlock;
+  int64_t gemmKPack;
```
jerryyin: @krzysz00 I think kPack wasn't a tuning parameter in nonxdlops kernel? So this might have to be reverted.

mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
```diff
@@ -351,45 +351,24 @@ template <typename T>
...
-    gemmKRemain = gemmSize.k % extraParams.gemmKPerBlock;
-    if (gemmKRemain != 0)
-      gemmKExtra = extraParams.gemmKPerBlock - gemmKRemain;
-
-    // llvm::errs() << "gemmMExtra: " << gemmMExtra << "gemmNExtra: " <<
-    // gemmNExtra << "gemmKExtra: " << gemmKExtra << "\n";
+    auto extraParams = populateParams.getUniversalParameters();
+    return calculatePadding(
+        extraParams.gemmKPerBlock, extraParams.gemmMPerBlock,
+        extraParams.gemmNPerBlock, gemmSize, extraParams.gemmKPack);
```
jerryyin: @manupak for review

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -420,7 +420,7 @@ struct BlockwiseGemmV2RewritePattern
         sourceOffset = mnb.create<AddIOp>(
             loc, sourceOffset, mnb.create<MulIOp>(loc, mnPerMfmaGroup, mn_i));
         sourceOffset = kb.create<AddIOp>(loc, sourceOffset,
-                                         mnb.create<MulIOp>(loc, MN, mn_i));
+                                         kb.create<MulIOp>(loc, MN, k_i));
```
jerryyin, line 423: Nice catch!

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -263,7 +263,7 @@ PopulateParamsXDL::initParametersForwardI8[
 };
 // clang-format on
 
-const InitParams PopulateParamsXDL::universalParameters = {32, 64, 4};
+const InitParams PopulateParamsXDL::universalParameters = {32, 64, 4, 4};
```
jerryyin: I'm not sure if adding kpack by default to be 4 on xdlops is fair. Also + @manupak / @krzysz00 for their opinions.

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
krzysz00: Yeah, kPack isn't a tuning parameter in non-xdlops, and we can't just add it like this.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
krzysz00: Yeah, I'm not sure about that either.

krzysz00 (changes requested):
Adding a hold because of the field in the tuning parameters struct that doesn't match the perf_config string.

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
```diff
@@ -203,6 +207,8 @@ class PopulateParams : public BasePopulateParams<InitParamsNonXDL> {
 
   const InitParams &getUniversalParameters() const;
```
krzysz00: Do we know why this is returning an `InitParams` and not one of its subtypes? That would get rid of the need for `getUniversalKPack()`

giuseros:
Hi @krzysz00 I tried to apply what we have discussed and also submitted a ticket to remove `universalParameters` from our code: https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/800

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -493,6 +496,15 @@ const InitParams &PopulateParamsXDL::getUniversalParameters() const {
   return universalParameters;
 }
 
+int64_t PopulateParamsXDL::getUniversalKPack() const {
+  // TODO(grossini): this is because we are using universal parameters
```
krzysz00: Hold on, can you make sure that the heuristic parameters list doesn't have the invalid config?

krzysz00 (commented):
I still want to make sure our heuristic parameters list doesn't have the bad configuration.

krzysz00 (approved):
Well, this at least fixes the immediate problem in a way that doesn't break anything, so let's go with it.

Please file a ticket for refactoring the "will this need padding" logic to use the same system as in the affix-parameters pass, possibly by reworking those functions to take the relevant components of a `RockGemmWrapperInterface`.

This could also be solved by doing the multi-kernel refactoring so that we don't need to make these decisions until after we've built the conv2d

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
giuseros: Unfortunately, we cannot remove the invalid config from the list, because it is valid for `fp32`. It is only for `fp16` that `32, 64, 4, 32, 64, 1` becomes invalid

PR #964, opened 2023-02-06
[DO NOT SQUASH] MLIR#760: Refactor fusion flow

sjw36:
* In tosa-to-rock insert rock.transform ops for all expand/collapse_shape ops
      * insert rock.transforms for transpose ops
      * insert rock.transforms to transpose rock.conv2d ops vs tosa.conv2d
    * Added rock-regularize pass to make linalg.generic regular for fusion
      * also pushes rock.transforms to memref.alloc writer chain (inverted)
    * Added rock-transform-to-memref for unfolded rock.transform ops
      * change back to memref.expand/collapse_shape ops
    * Moved rock.threadwise_read_into/write_all op conversion to rock-blockwise-gemm-to-threadwise
    * Removed unneeded passes
      * rock-fold-transform: refactored in tosa-regularize
      * rock-copy-opt: memref.copy converted in rock-linalg-align
    * enabled e2e for MLIR-762 bug fixed
    * cleanup other tests
    
    https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/760
    https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/762


mlir/include/mlir/Dialect/Rock/Passes.h
```diff
@@ -27,10 +27,11 @@ namespace rock {
 #define GEN_PASS_DECL_ROCKCLEANMATHPASS
 #define GEN_PASS_DECL_ROCKCONVTOGEMMPASS
 #define GEN_PASS_DECL_ROCKCOPYOPTPASS
```
krzysz00: Isn't copy-opt going away? Shouldn't we remove this?

mlir/include/mlir/Dialect/Rock/Passes.td
```diff
@@ -82,6 +77,11 @@ def RockBufferLoadMergePass : Pass<"rock-buffer-load-merge", "::mlir::func::Func
   let dependentDialects = ["::mlir::amdgpu::AMDGPUDialect"];
 }
 
+def RockTransformToMemrefPass : Pass<"rock-transform-to-memref", "::mlir::func::FuncOp"> {
```
krzysz00, line 86: Why do we have this?

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
```diff
@@ -104,60 +104,60 @@ void rock::buildBufferizePipeline(OpPassManager &pm,
...
-    pm.addPass(rock::createRockLoopsToCfPass());
+    pm.addNestedPass<func::FuncOp>(
+        rock::createRockBlockwiseGemmToThreadwisePass());
+    pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
+    pm.addNestedPass<func::FuncOp>(
+        rock::createRockThreadwiseGemmLoweringPass());
+    pm.addNestedPass<func::FuncOp>(rock::createRockSugarToLoopsPass());
+    pm.addNestedPass<func::FuncOp>(rock::createRockCleanMathPass());
+    pm.addNestedPass<func::FuncOp>(rock::createRockBufferLoadMergePass());
+    pm.addNestedPass<func::FuncOp>(rock::createRockTransformToMemrefPass());
```
krzysz00, line 157: What's this meant to target? Shouldn't all uses of `transform` been removed by now?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -61,136 +62,110 @@ struct RockLinalgAlignPass
...
-          b.create<GlobalLoadOp>(loc, typeToLoad, source, isValid,
-                                 copyLoop.getLowerCoords(/*domain=*/0));
-      b.create<InBoundsStoreOp>(loc, loaded, dest,
-                                copyLoop.getLowerCoords(/*domain=*/1)[lastDim]);
+static void moveTransformsBefore(PatternRewriter &b, Value val) {
+  if (auto defOp = val.getDefiningOp()) {
+    if (auto rtop = dyn_cast<rock::TransformOp>(defOp)) {
+      moveTransformsBefore(b, rtop.getOperand());
+
+      defOp->remove();
```
krzysz00, line 85: `b.erase(defOp)` so the listeners can fire

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -61,136 +62,110 @@ struct RockLinalgAlignPass
...
+
+  return nOut;
+}
+
+static Value applyTransforms(PatternRewriter &b, ThreadwiseWriteAllOp storeOp,
+                             Value src) {
   // 0. capture the memref containing the outputs being written
   Location loc = storeOp.getLoc();
-  Value gemmOuts = storeOp.getSource();
+  Value gemmOuts = storeOp.getOperand(0);
```
krzysz00: Why this change? It's clearer to use the names of the arguments

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -61,136 +62,110 @@ struct RockLinalgAlignPass
...
+      for (auto &use : result.getUses()) {
+        return traceToRealOp(use.getOwner());
+      }
+    }
+  }
+  return op;
 }
 
-static GlobalStoreOp traceToGlobalStore(Value inp) {
+static rock::ThreadwiseWriteAllOp traceToThreadwiseWrite(Value inp) {
```
krzysz00: I don't think you need the `rock::` here.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -346,115 +274,137 @@ LogicalResult MILARewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
...
+  Value laOutRegs = reconfigureLAGeneric(b, laGeneric, fusionRegs, gemmStoreOp);
+  // 2.2.0. Move the generic before the write-back. This'll put all
+  // the copy loops for other inputs before the generic due to insertion
+  // order.
+  laGeneric->moveBefore(gemmStoreOp);
+
+  gemmOut.replaceAllUsesWith(fusionRegs);
+
+  // 2.3. Replace rock.threadwise_write_all inputs with la.generic result vgprs
+  gemmStoreOp.setOperand(0, laOutRegs);
```
krzysz00: `.getSourceMutable() = `

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -346,115 +274,137 @@ LogicalResult MILARewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
...
+
+  out = transformAccordingly(b, out, gemmStoreOp.getOperand(1));
+  gemmStoreOp.setOperand(1, out);
```
krzysz00: Similarly on not using the integer numbers here.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -346,115 +274,137 @@ LogicalResult MILARewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
...
-                                           : WalkResult::advance();
-      });
-  if (verifyAllStores.wasInterrupted())
-    signalPassFailure();
+  {
+    RewritePatternSet patterns(ctx);
+    patterns.add<LAGenericRewritePattern>(ctx);
+    patterns.add<MemcpyRewritePattern>(ctx);
+    GreedyRewriteConfig config;
+    config.useTopDownTraversal = true;
```
krzysz00, line 379: Huh, and now I know

mlir/lib/Dialect/Rock/Transforms/TransformToMemref.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+//===----------------------------------------------------------------------===//
+// TransformOp conversion to MemRef
+//===----------------------------------------------------------------------===//
+struct TransformRewritePattern : public OpRewritePattern<TransformOp> {
+  using OpRewritePattern<TransformOp>::OpRewritePattern;
+  LogicalResult matchAndRewrite(TransformOp op,
+                                PatternRewriter &b) const override {
+    Location loc = op.getLoc();
+    auto srcType = op.getOperand().getType().cast<ShapedType>();
```
krzysz00: Does this end up needing the cast? I could've sworn that if you use more specific methods you'll get a `TypedValue<ShapedType>`

mlir/lib/Dialect/Rock/Transforms/TransformToMemref.cpp
```diff
@@ -0,0 +1,132 @@
...
+    auto resType = op.getResult().getType().cast<ShapedType>();
+
+    bool expanded = resType.getShape().size() > srcType.getShape().size();
+    SmallVector<ReassociationIndices> merges(
+        expanded ? srcType.getShape().size() : resType.getShape().size());
+
+    // only converts simple expand/collapse form
+    for (auto tattr : op.getTransform().getOps()) {
+      auto inDims = expanded ? tattr.getLowerDims() : tattr.getUpperDims();
```
krzysz00: Can we go ahead and call these the `ArrayRef<int64_t>` that they are?

mlir/lib/Dialect/Rock/Transforms/TransformToMemref.cpp
```diff
@@ -0,0 +1,132 @@
...
+          assert(merges[inDim].empty());
+          merges[inDim].push_back(std::get<1>(pair));
+        }
+        break;
+      case rock::TransformType::Pad:
+      case rock::TransformType::Slice:
+      case rock::TransformType::Embed:
+      case rock::TransformType::Broadcast: // Unsupported
+      case rock::TransformType::AddDim:
+      case rock::TransformType::ConstDim:
```
krzysz00, line 100: AddDim and ConstDim aren't failure cases - they come up when going to/from a scalar memref like `memref<f32>`. That should be handled here.

krzysz00 (commented):
In good news, all this seems mostly fine! I've got some minor comments and a few questions.

Also, are there tests for the transform->expand/collapse_shape pass? It feels like we could use some.

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
sjw36: Unfortunately no. There is at least one in the init kernel, to flatten to 1 dimension.

mlir/lib/Dialect/Rock/Transforms/TransformToMemref.cpp
sjw36: Agreed, but we only see this case for collapse_shape in the init kernel.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
sjw36: Not erasing the defOp, just moving it to the current insertion point.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
sjw36: It is no longer a rock.global_store, don't remember when I changed it but getSource wasn't available at the time.

mlir/include/mlir/Dialect/Rock/Passes.td
sjw36: see comment below.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: I think you could use moveAfter/moveBefore for that...

external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalgPass.cpp
```diff
@@ -55,6 +55,7 @@ struct TosaToLinalg : public impl::TosaToLinalgBase<TosaToLinalg> {
     target.addLegalOp<tosa::IfOp>();
     target.addLegalOp<tosa::ConstOp>();
     target.addLegalOp<tosa::WhileOp>();
+    target.addLegalOp<tosa::ReshapeOp>();
```
manupak, line 58: Why this change ? 
Im worried that this is not upstreamable...

external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalg.cpp
```diff
@@ -815,112 +816,12 @@ static LogicalResult reduceMatchAndRewriteHelper(Operation *op, uint64_t axis,
...
 
-  rewriter.replaceOpWithNewOp<tosa::ReshapeOp>(op, resultTy,
-                                               linalgOp.getResults());
+  rewriter.replaceOpWithNewOp<tosa::ReshapeOp>(
+      op, resultTy, linalgOp.getResults()[0],
+      rewriter.getI64ArrayAttr(resultTy.getShape()));
   return success();
 }
 
-static bool findIntermediateShape(ArrayRef<int64_t> lhsShape,
```
manupak, line 823: Is this part of upstream merge ? or why are we doing this here

mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
```diff
@@ -74,6 +75,22 @@ Value insertTransposeAndBroadcastTransforms(OpBuilder &b,
                                             ArrayRef<int64_t> outShape,
                                             Value inp, AffineMap inpIdxMap);
 
+// This function will take an input TransformMapAttr and invert the
+// shapes and transforms.
+TransformMapAttr invertTransformMap(OpBuilder &b,
```
manupak, line 80: nice addition!

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -65,16 +65,14 @@ class ConvConverter final
   Value getZeroBias(Location loc, Type elemType, int64_t filterOutputChannels,
                     ConversionPatternRewriter &rewriter) const {
     auto biasTy = RankedTensorType::get({filterOutputChannels}, elemType);
-    auto arithZero =
-        rewriter.create<arith::ConstantOp>(loc, rewriter.getZeroAttr(elemType));
-    auto biasTensor = rewriter.create<tensor::SplatOp>(loc, biasTy, arithZero);
-    return biasTensor;
+    return rewriter.create<arith::ConstantOp>(loc,
```
manupak, line 68: Can you verify the TOSA produced here is verifiable via a non-rock flow ?
Historically, the change done here only worked in a tosa->rock based lowering only... (maybe that has changed upstream).

If we still want to do this change, I'd like to see a clone test involving a migraphx conv2d test (which exercises non-rock based lowering for tosa).

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -217,9 +219,13 @@ class ConvConverter final : public OpConversionPattern<tosa::Conv2DOp> {
     FailureOr<rock::Conv2DOp> rockConv = makeRockConv2D(
         rw, op, input, inputLayout, filter, filterLayout, output, outputLayout,
         op.getPad(), op.getStride(), op.getDilation());
+
     if (failed(rockConv))
       return failure();
 
+    // disable layout changes since will be transforms
+    (*rockConv)->setAttr("has_relayout_do_not_unfold", rw.getUnitAttr());
```
manupak: Can we declare this in a common place ? 

Also who is the consumer and purpose of this attr ?

mlir/lib/Conversion/TosaToRock/TosaToRockPass.cpp
```diff
@@ -39,21 +40,25 @@ struct TosaToRock : public impl::TosaToRockPassBase<TosaToRock> {
     if (!func->hasAttr("kernel")) {
       return;
     }
+    // expand tosa.reshape to tensor.collapse_shape and tensor.expand_shape
+    // within kernel functions.
+    OpPassManager pm("func.func");
+    pm.addPass(mlir::tosa::createTosaToTensor());
```
manupak: Why do we have to run the pass here ? (and not in the kernel pipeline)

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -61,136 +62,110 @@ struct RockLinalgAlignPass
   void runOnOperation() override;
 };
 
-struct MILARewritePattern : public OpRewritePattern<linalg::GenericOp> {
+struct LAGenericRewritePattern : public OpRewritePattern<linalg::GenericOp> {
```
manupak, line 65: Thanks for this! I never knew what MILA stands for :)

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -61,136 +62,110 @@ struct RockLinalgAlignPass
...
-  //    buffer reads)
-  return makeTransformingCopyLoop(b, gemmStoreOp, ret);
+  // 2.2. load into registers
+  b.create<ThreadwiseReadIntoOp>(loc, src, alloc, storeOp.getExtraViews(),
+                                 storeOp.getForceUnroll(),
+                                 storeOp.getUseIndexDiffs());
+  return alloc;
+}
+
+static Operation *traceToRealOp(Operation *op) {
```
manupak: nit: I'd prefer a bit more specific name such as traceToNonViewUser / traceToNonTrUser or similiar

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -61,136 +62,110 @@ struct RockLinalgAlignPass
...
+  if (auto transform = dyn_cast<rock::TransformOp>(op)) {
+    Value result = transform.getResult();
+    if (result.hasOneUse()) {
```
manupak: I think this works the on the assumption that transform chain does not have divergence.
Therefore, we need to handle the case when divergence is there and for that if its not needed, lets error out here...

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -61,136 +62,110 @@ struct RockLinalgAlignPass
...
     if (isa<memref::DeallocOp>(use)) {
       // ignore
       continue;
     }
-    if (isa<linalg::GenericOp>(use)) {
+    if (auto lgop = dyn_cast<linalg::GenericOp>(use)) {
       // reader
-    } else if (auto store = dyn_cast<GlobalStoreOp>(use)) {
+      if (!llvm::is_contained(lgop.inputs(), inp)) {
+        allValidUses = false;
```
manupak: Since this is a refactor,

nit : lets get rid of this boolean and nullptr returns and switch destination passing style.
For that, I'd suggest lets make the function return a LogicalResult with a reference as a arg (Operation&) to be filled. I personally feels its more cleaner that way.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -346,115 +274,137 @@ LogicalResult MILARewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
...
-        LLVM_DEBUG(llvm::dbgs() << outToInMap << "\n");
-        LLVM_DEBUG(llvm::dbgs() << "^ is not a isProjectedPermutation from "
-                                   "output coords to fusion input\n");
-        return failure();
-      }
+  for (auto idxMap : laGeneric.getIndexingMapsArray()) {
+    if (!idxMap.isIdentity()) {
+      LLVM_DEBUG(llvm::dbgs() << idxMap << "\n");
+      LLVM_DEBUG(llvm::dbgs() << "^ is not an identity map\n");
+      return failure();
```
manupak: So if I understand correctly, now we create the views (via transforms) prior to here, therefore Linalg block access those views. Thus, enabling make all indexing maps identity (which is really cool!).

So I'd think this should be a place we should error out (i.e. op.emitError) and not a matching failure anymore as we dont have anything beyond this pass to handle them.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: I think we should be able to access the operand by name still (the corresponding one for ThreadWiseWrite)

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -258,64 +226,24 @@ static Value reconfigureLAGeneric(PatternRewriter &b,
...
-      LLVM_DEBUG(llvm::interleaveComma(permutedDims, llvm::dbgs()));
-      LLVM_DEBUG(llvm::dbgs() << "\n");
-
-      for (auto bDim : permutedDims) {
-        if (inpShape[bDim] != 1) {
-          LLVM_DEBUG(llvm::dbgs()
-                     << "The store input cannot be a real broacast.\n");
-          return failure();
-        };
+static Value findThreadwiseWrite(linalg::GenericOp laGeneric,
```
manupak, line 226: nit : I'd like this to return LogicalResult rather than a null in the case of failure...

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: same concern here and before as well.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -346,115 +274,137 @@ LogicalResult MILARewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
...
+
+    // 1. replace memref.copy with rock.threadwise_write_all
+    trg = transformAccordingly(b, trg, gemmStoreOp->getOperand(1));
+    gemmStoreOp->setOperand(1, trg);
+
+    if (auto outAlloc = trg.getDefiningOp<memref::AllocOp>())
+      outAlloc->moveBefore(gemmStoreOp);
+
+    b.eraseOp(copy);
     return success();
```
manupak, line 356: Here also, shouldn't this be a pass failure and not a match failure if there memref::Copy is to be found at this stage ?

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
```diff
@@ -0,0 +1,323 @@
...
+        }
+      }
+    }
+    return lres;
+  }
+};
+
+////////////////////////////////////////////////////////////////////////////////
+////////////////////////////////////////////////////////////////////////////////
+struct RockRegularizePass
```
manupak, line 286: This is awesome!

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -974,3 +990,191 @@ Value mlir::rock::insertTransposeAndBroadcastTransforms(
   }
   return inp;
 }
+
+TransformMapAttr mlir::rock::invertTransformMap(
```
manupak, line 994: I would rather make this a property of the TransformMapAttr to define its own inversion (optionally ofc -- where its possible to do so)
Any reason why we went on to create a utility function ?

external/llvm-project/mlir/lib/Conversion/TosaToTensor/TosaToTensor.cpp
```diff
@@ -15,21 +15,247 @@
...
 using namespace mlir;
 using namespace tosa;
 
+static bool findIntermediateShape(ArrayRef<int64_t> lhsShape,
+                                  ArrayRef<int64_t> rhsShape,
+                                  SmallVector<int64_t> &intermediateShape,
+                                  bool isDynamic) {
+  if (isDynamic) {
+    // TODO (natashaknk): Make dynamic intermediate shape not always be rank-1
+    intermediateShape = {-1};
```
jungpark-mlir, line 29: fyi, mlir::ShapedType::kDynamic is newly introduced in upstream, I'll update it later.

manupak (changes requested):
Thanks for the refactor and the effort went into this.
I have nits, questions and actual suggestions for few changes.

I'd insist on having specific tests for the new pass (rock regularize) -- hence the hold.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
manupak: (This was historically tested before we had rock.gemm in the migraphx.dot tests but I guess now they are not)

external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalgPass.cpp
jungpark-mlir: This change migrates reshape rewriting to TosaToTensor which looks making better sense because it converts to 'tensor' dialect op.

sjw36: Yeah, the conversions seemed out of place. Thanks @krzysz00 for contributing, can you add these to the upstream list.

external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalg.cpp
sjw36: see above.

external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalgPass.cpp
manupak: I see. agreed that its better... but do we require that ? 
I'd be inclined to fix this in upstream first as we are not maintaining these unless we rely on this.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
sjw36: seems to work, I am adding the test.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
sjw36: this attribute is going away soon.. @krzysz00 is there a ticket?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
sjw36: yep, MI artifact..

external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalgPass.cpp
krzysz00: We are relying on being able to call just the reshape-> collapse/expand shape patterns, yes

krzysz00: That is, the fact that those patterns were in tosa-to-linalg and not tosa-to-tensor is a problem for our rewrites. I intend to upstream this move once it lands here.

mlir/lib/Dialect/Rock/Transforms/TransformToMemref.cpp
krzysz00: But why leave a perfectly obvious bit of code out? It might make people think there's some deep reason we didn't implement it, especially given the AddDim/ConstDim logic over in collapse/expand_shape to transforms

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
krzysz00: There should be a getSource(). Now, as to why the change, it's possible you wrote this code to be generic over `Operation *` at some point.

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
krzysz00: Ah, which then gets caught up be regularize, which we then have to put back ... yeah, makes sense. I'd add this to the documentation of the pass.

external/llvm-project/mlir/lib/Conversion/TosaToTensor/TosaToTensor.cpp
krzysz00: This'll get picked up by trying to port the patch over?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
krzysz00: You'd be looking for `FailureOr<Value>` but also, null as the failure value for Value/Attribute/... is reasonably accepted

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
krzysz00: @giuseros would know the timeline.

The point of this is that current `conv-to-gemm` *mostly* uses the dimension attributes (indicating whether something is a NCHW, NHWC, and so on) as a guide to interpreting the input memrefs, which could be transformed from whatever underlying layout we'd like. 

 However, it assumes, when looking at the filter tensor, that the described dimensions match the underlying layout for the purposes of determining whether the C [cat] Y [cat] X  merge is an `Unfold` or not. `Unfold` is a `Merge` with the additional assertion that the dimensions it produces will be stored contiguously in the underlying memory. This was an assertion we could confidently make back when we were more of a CK clone, but we can't really determine that sort of thing anymore.

@giuseros did the work needed to make the vectorization analyzer infer unfold-like properties. He's still got the followup PR to remove `Unfold` as a concept, and to change up our one `Unfold`-based optimization into something more general.

For now, this attribute is the magic incantation to make the conv-to-gemm pass never declare the filter tensor merge to be `Unfold`, even if it looks like it should be.

external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalgPass.cpp
manupak: Thanks for the clarifications...
Cool! seems fine if we require it... (worried it was a valid but unrelated change)

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
krzysz00: I  think that'll get caught by the "are there still `memref.alloc`s in the function check.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
krzysz00: Inverses are only really needed for this one rewrite, so having them walking around all the time is kinda silly, I'd think.

krzysz00:
[AMD Official Use Only - General]

I mean, this could be split into a second PR and then have this one be a dependency of that one. Might make history a bit cleaner at the cast of requiring `rebase -i` usage at worst.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
sjw36: I will clean it up a bit but I prefer return-by-value (like shared_ptr) vs pass-by-reference. It is more clean from usage:
**RBV:**
`if (auto op = traceTo(use)) {
`
- uses move semantics
- MLIR ops are shared_ptr-like
- clear ownership

**PBR:**
```
Operation value;
if (succeeded(traceTo(use, value)) {
```

- not clear from call that ownership is passed
- requires more scoping
- should 'value' be tested as well, to ensure traceTo set it???

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: I'd prefer FailureOr\<Value\> over null -- It seems easy to miss to check the null, IMHO where as in missing failure check gives out a compiler warning.

Anyway its nit picking (and subjective)...

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
krzysz00: `Operation` and friends usually get returned by value in MLIR from what I've seen.

sjw36:
> I mean, this could be split into a second PR and then have this one be a dependency of that one. Might make history a bit cleaner at the cast of requiring `rebase -i` usage at worst.

Could be pulled out but if I don't squash it would be a unique commit.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
manupak: Ah just realizing TransformTypes are just enums. I was thinking they were classes which can hold the "property" that we can nicely attribute the inverse to.

manupak: (So its structurally not possible do what I propose, but I'd be cautious about proliferation of utility functions where each type could independently define its inverse)

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: It is stylistic -- there are pros and cons for either style.
One could use the return channel as way of error propagation and the caller holds the ownership of the output memory (if its not smart-pointer'ish)

a middle ground might be FailureOrValue (as @krzysz00 suggested somewhere else).

but anyway it would be good to agree on a style (which I thought was more or less DPS oriented).
I think Operation is bit smart-pointy (I might be wrong -- Im saying this by just looking at existence of explicit deep copies etc) but in-case if we are returning a container it might not work well (e.g. SmallVectorImpl) -- right ?
Hence, due to asymmetry of  things used as outputs in LLVM/MLIR, I usually preferred DPS as reference counting is not always present but like to hear what folks think...

manupak: (Also its a nit -- feel free to ignore if you think its not applicable :) )

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: I think so too.. but why do we need to do that there since we have it here ?

manupak:
> > I mean, this could be split into a second PR and then have this one be a dependency of that one. Might make history a bit cleaner at the cast of requiring `rebase -i` usage at worst.
> 
> Could be pulled out but if I don't squash it would be a unique commit.

Advantage might be both commits will be tested in the two PR case -- Im fine either way as Im convinced now that this change is required here.

sjw36:
I believe all feedback has been addressed. Please confirm.

manupak:
I think there are a still a couple left... and I think we need some tests around the newly added pass.
Could you resolve the conversations that you think you addressed -- so that we can see whats left maybe ?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
sjw36: Caught by alloc check at the end but it is kind of clumsy..

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -300,141 +306,106 @@ class MatMulConverter final : public OpConversionPattern<tosa::MatMulOp> {
...
+    Operation *cval = v.getDefiningOp();
+    if (isa<arith::ConstantOp>(cval) || isa<tosa::ConstOp>(cval)) {
       auto cattr = cval->getAttr("value").cast<DenseElementsAttr>();
       auto vals = cattr.tryGetValues<int32_t>();
       if (succeeded(vals))
         return SmallVector<int32_t>(*vals);
       auto vals64 = cattr.tryGetValues<int64_t>();
-      if (succeeded(vals64))
-        return SmallVector<int32_t>(*vals64);
+      assert(succeeded(vals64));
```
krzysz00: This can be phrased as `*vals64`.

sjw36:
I added a negative test for -rock-regularize to catch cases that should fail, all other tests will exercise this pass.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -300,141 +306,106 @@ class MatMulConverter final : public OpConversionPattern<tosa::MatMulOp> {
...
+      return SmallVector<int32_t>(*vals64);
     }
     // May be bufferization cast
     //  but this is no longer a bufferization pass, so assert
     assert(0);
```
krzysz00: Can we drop this section of the function?

Also, I'd argue for this entire thing being a "fail the pass" function, not a "crash the program" one.

That is, adding non-constant transpose indices isn't a case where we've violated our own invariants, but one where the programmer passed in input this pass can't handle. So we should fail gracefully.

mlir/lib/Conversion/TosaToRock/TosaToRockPass.cpp
krzysz00: So that it only runs on `kernel` functions.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
sjw36: For this case, it is not a failure to not find the ThreadwiseWriteAll, and I prefer to return a null object for simple conditional test and usage.

krzysz00 (commented):
Tiny comments, I don't think there's anything blocking in here

krzysz00:
Re tests, I agree with @manupak's point that adding a test that specifically confirms (even for some small case) that the pass we added does what we say it does is useful in increasing our test coverage, and while that "this pass is exercised by other passes" is a true statement, it's better to also have specific test coverage for each pass.

(Also, I'm going to go back and poke that the transform->expand/collapse pass should handle AddDim and ConstDim)

mlir/lib/Conversion/TosaToRock/TosaToRockPass.cpp
sjw36: Yes, this can be moved to the pipeline.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
sjw36: done

mlir/lib/Conversion/TosaToRock/TosaToRockPass.cpp
krzysz00: I disagree with moving it to the pipeline, as I want this to only run on functions that are being `tosa-to-rock`d

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
sjw36: done

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
sjw36: There was a big discussion about this form in MLIR, lots of opinions (can't find it, maybe it was on discord). There was an Optional<> wrapper that was decidedly redundant since the Operation,Value, etc containers are all smart-ptr like. Here again, a null object return is valid.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
sjw36: Unfortunately the Rewriter doesn't have this method, and it has the current insertion point.

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
sjw36: added.

mlir/include/mlir/Dialect/Rock/Passes.h
sjw36: gone.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
sjw36: gone

mlir/lib/Dialect/Rock/Transforms/TransformToMemref.cpp
sjw36: then I will need a test..

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
sjw36: yep

sjw36:
> Re tests, I agree with @manupak's point that adding a test that specifically confirms (even for some small case) that the pass we added does what we say it does is useful in increasing our test coverage, and while that "this pass is exercised by other passes" is a true statement, it's better to also have specific test coverage for each pass.

Ok, but just 1.. I don't like alot of tests expecting a rigid pattern which breaks every time something gets tweaked.. :)

> 
> (Also, I'm going to go back and poke that the transform->expand/collapse pass should handle AddDim and ConstDim)

As I recall, AddDim can be used to Broadcast as well right? Seems to complicate this conversion that doesn't get exercised.. It is a Conversion pass so it will fail if a rock.transform remains.

mlir/lib/Conversion/TosaToRock/TosaToRockPass.cpp
sjw36: It was already in the pipeline, just not early enough. It is a normal conversion for tosa lowering (needed for the tosa.slice op).

krzysz00:
The point of tests that expect a rigid pattern that break when you tweak something is that they let you know that your changed did something, and so adjusting the tests serves as documentation of the changes.

(See also LLVM proper which even has an "update the test to match my code changes" script)

`AddDim` and `ConstDim` get produced by the logic for expand_shape and collapse_shape, and this pass is meant to be the inverse of that logic. You'll be perfectly within your rights to fail on `AddDim{N}` where `N != 1` for example, but this pass should be the exact inverse of the utility functions that convert `memref.expand_shape` and `memref.collapse_shape` to coordinate transformations.

sjw36:
> `AddDim` and `ConstDim` get produced by the logic for expand_shape and collapse_shape, and this pass is meant to be the inverse of that logic. You'll be perfectly within your rights to fail on `AddDim{N}` where `N != 1` for example, but this pass should be the exact inverse of the utility functions that convert `memref.expand_shape` and `memref.collapse_shape` to coordinate transformations.

The conversion from memref.expand/collapse never creates AddDim or ConstDim because the reassocs cover all dims. I'm going to punt and if this case comes up it will fail the pass and can be addressed then.

mlir/lib/Conversion/TosaToRock/TosaToRockPass.cpp
manupak: Few thoughts here : 

* I generally dont prefer passes buried within the passes.
* I thought tosa-to-rock was run on partitioned functions in the pipeline anyway.. but I can see the point of running the passes outside the driver.
* It also feels like we need a tensor-to-rock pass (maybe not right now)
   * it will also fit nicely with proposal of seperating tosa-to-tensor bits out of tosa-to-linalg and we will not have this somewhat weird depedency that tosa to needed to be converted to tensor before it can be converted to rock.

mlir/test/Dialect/Rock/regularize.mlir
```diff
@@ -0,0 +1,44 @@
+// RUN: rocmlir-opt --rock-regularize %s | FileCheck %s
```
manupak, line 1: Thanks for adding this!

manupak (approved):
Thanks for adding the unit tests for the pass -- lifting the hold from my side.
The changes look good to me now... (I ll let you guys finish other conversations)

(Agreeing with all the points of @krzysz00)

I see them as necessary (but not sufficient) tests
I would always advocate for *some* unit tests for the passes to improve comprehension of what they are supposed to do and means to promote modularity/composability of each pass going forward.  It will also aid us in future refactoring/NFCs.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: I am yet to find returns of llvm containers -- i.e. SmallVectorImpl -- to be smart-ptr like. They are not valid once the stack frame is destroyed (is what I find) -- (which is of course not applicable here)

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -217,9 +219,13 @@ class ConvConverter final : public OpConversionPattern<tosa::Conv2DOp> {
     FailureOr<rock::Conv2DOp> rockConv = makeRockConv2D(
         rw, op, input, inputLayout, filter, filterLayout, output, outputLayout,
         op.getPad(), op.getStride(), op.getDilation());
+
     if (failed(rockConv))
       return failure();
 
+    // disable layout changes since will be transforms
```
jungpark-mlir: Hang on,
So now, it doesn't do high-level layout optimization and just fuse transposes as transforms?
For example, MIGraphX always has NCHW layout and tosa only uses NHWC, so the conversion always includes transforms on both input/output.
This changes memory access  pattern of all LD/ST and probably kill the performance.
Even though somehow the transforms can be well optimized at the end, we don't have a scheme to tune the kernel with the  fused transforms at the moment.

I'm concerning this can significantly degrade the performance of current MIGraphX integration and would suggest to hold this part of change until we can evaluate the performance.

jungpark-mlir (changes requested):
Performance concerning on the part of this change.

mlir/test/Dialect/Rock/regularize.mlir
sjw36: I agree with unit testing as well, and yes it is lacking. Just a bit frustrated with the e2e resnet kernel tests that also checked for a rigid IR, and most of these were very similar so they didn't add much new coverage.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
sjw36: @krzysz00 believes this should work out as before since it untransforms to determine actual layouts. Is there a way to generate the perf from a PR?

krzysz00: Ok, so, all this does (if we're talking about the magic attribute) is stop conv-to-gemm from marking the indices as `Unfold`, which is a funky optimization that would break if you change which dimensions are contiguous in memory that'll be going away soon.

_However_, on to your broader point, this does have the possibility to kill perf, and @sjw36, you may want to write up a "pull transposes into the convolution layout" pass again. The reason for that is that, while untransform() will get you to your underlying out, conv-to-gemm uses the layout specified on the convolution (and not any underlying layout, which can't really be interpreted) to determine how to flatten out the dimensions.

For example, if I tell `rock.conv2d` that I have a KCYX filter, then conv-to-gemm will generate `gemmK <- Merge(C, Y, X)`. *However,*, if I say I have a `KYXC` tensor, we will instead have the transform `gemmK <- Merge(Y, X, C)`, on the assumption that you're generally calling `rock.conv2d` with a tensor whose layout is basically like what you passed as an argument to `rock.conv2d`. Therefore, having the transpose just hanging out in the transform stack, while not fatal to correctness, could be bad for perf

krzysz00:
@sjw36 Agreed on the punt in the transforms->expand/collapse pass, but, in that case, let's make sure it's documented in the code why that pass is there and why those cases weren't addressed.

mlir/lib/Conversion/TosaToRock/TosaToRockPass.cpp
krzysz00: Whereas I don't see the problem with nesting in a preprocessing pass that's only needed at this stage for tosa-to-rock conversion, and having it only run on the functions we're running tosa-to-rock on.

In the general case, tosa-to-tensor has to run _**after** tosa-to-linalg, because tosa-to-linalg can create a `tosa.reshape` now. However, we need those rewrites now - and only for functions targetting the GPU - in order to get the transforms thing working.

And since the only way to conditionally apply passes is to nest them, we need to nest them.

@manupak @sjw36  for "put it back"

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
```diff
@@ -48,6 +48,11 @@ void rock::buildBufferizePipeline(OpPassManager &pm,
                                   const rock::BufferizeOptions &options) {
   bool noRock = options.disableRock;
 
+  // for tosa reshape ops
+  /* rocmlir-opt --tosa-to-tensor
+   */
+  pm.addNestedPass<func::FuncOp>(tosa::createTosaToTensor());
```
krzysz00: To put it another way, I don't want to break any other tosa pass's assumptions and only run the rewrites in contexts where we _know_ we're about to do `tosa-to-rock`, which means the attribute check, which means nested passes.

sjw36: All the clone tests run this pipeline on host code, all are passing. Furthermore, tosa-to-linalg should only care about tosa ops, not the converted memref ops.

krzysz00: Except there are tosa optimizations that we run or might want to run while lowering the host code that would expect a `reshape` instead of the tensor operations.

manupak:
@sjw36 @krzysz00 @jungpark-mlir 

Re: double transposes killing perf,

Another option is we could rewrite the {tosa.transpose + tosa.conv2d} pattern in rock.conv2d with appropriate layout because it rather sounds like we are working around a limitation of TOSA here.

manupak:
Also prior to merging we need to enable checkMIGraphX jenkins param for this PR.

sjw36:
> Another option is we could rewrite the {tosa.transpose + tosa.conv2d} pattern in rock.conv2d with appropriate layout because it rather sounds like we are working around a limitation of TOSA here.

I don't think it is a limitation since transposes can be used to handle all layouts. Instructions in general have specified requirements of data layout or specified system-wide. I prefer the simplicity of tosa.conv2d with specified layout, and using other ops to change views appropriately. Using string properties is kind of krufty (even linalg opted for 2 different ops with explicit layouts instead of properties).

manupak:
> > Another option is we could rewrite the {tosa.transpose + tosa.conv2d} pattern in rock.conv2d with appropriate layout because it rather sounds like we are working around a limitation of TOSA here.
> 
> I don't think it is a limitation since transposes can be used to handle all layouts. Instructions in general have specified requirements of data layout or specified system-wide. I prefer the simplicity of tosa.conv2d with specified layout, and using other ops to change views appropriately. Using string properties is kind of krufty (even linalg opted for 2 different ops with explicit layouts instead of properties).

Hmmm, I would generally agree with you in most cases for simplicity .. but I think in this case its not just the views because the layout could affect traversal in which conv2d should be doing. Thus, it seems to me like the layout is important as a property of the conv2d to decide that. If we have to look at the surrounding ops to figure out how the conv2d traversal should be done , it feels conv2d op definition is not self contained.

I guess even with your argument it still hold the case that we should look at conv2d + transposes together when creating rock.conv2d ?

sjw36:
Yes, agreed it has to be folded into the logic somewhere but I would argue that should happen when lowering to read/write logic since that is where it is needed. Anywhere actual layout is needed, it can be deduced from tracing to func args.

manupak:
> Anywhere actual layout is needed, it can be deduced from tracing to func args.

I think this is where I slightly disagree. Whilst, we can do this, it makes the op loosely defined, IMHO -- Hence the limitation of the tosa.conv2d as it is too specific and is not able to accept a hint

TLDR;

Just to clarify my thoughts (only for conversation sake and not as something required of this PR)  and happy to be corrected.

The traversal/parallel_compute of points in the operator space is constrained using data dependencies and this is case where too much freedom is available in deciding how to do that. Whilst we generally infer the traversal/parallel_compute of points in the operator space using the data layout (usually of input tensors), from correctnes PoV, I feel its a bit independent of that. What we generally do is we use the layout of input tensors as a proxy to direct the op undergo a form of traversal/parallel_compute of the points in the operator space. However, having the ability to supply a hint to the op (the layout) will aid to do a better/performant lowering and the lowering will be contained within conv2d op.

Now, the transposes/reshapes just maps this space to input space in a different way and in-fact we can use that to help define the traversal of the op it self.

(acknowledging a discussion with @jungpark-mlir today morning)
This becomes important in tuning and I think we need a tuned points for different traversal/parallel_compute patterns (inferred by layouts)

jungpark-mlir (approved):
I was only concerning lack of the solution right now, seems more discussion is needed to determine the future direction.
Approving since folding transpose is reiterated.

krzysz00:
From a correctness POV, just translating Tosa conv2d into an NHWC convolution will work. But, from a performance POV, the implicit GEMM algorithm does actually depend on the layout of the convolution arguments in memory for performance - that is, if we have an NCHW tensor, we should construct the K dimension of the gemm as [C, H, W], while, for example, if we had a WCH tensor, we should use a [W, C, H] K layout. If the order of concatenations in the implicit gemm construction doesn't match the tensor layout, there's a problem.

Sometimes, this is unavoidable, like in a mixed-layout convolution, where one of the tensors has to be read wrong for the sake of consistency.

However, in the tosa case, this isn't unavoidable, and conv-to-gemm should be able to get the real layout, or close to it.

Currently, we handle this by having the conv2d operator specify the layouts of its inputs and then having the implicit gemm construction assume that that layout is bascially correct when making the choice of how to construct the gemm dimensions.

In principle - but that's a future PR - we could, similarly to how we have a vectorization widget, have a "tell me what order these dimensions go in" widget, so that conv-to-gemm could use to make decisions about how to order gemm dimensions. However, that would only really be needed to work around Tosa's convolution being NHWC when we could get most of the way there by folding the `tosa.transpose` ops in ... which is probably something Tosa already does since Linalg has a whole lot of `conv2d` flavors?

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
```diff
@@ -186,12 +187,12 @@ def Rock_TransformMapAttr : Rock_Attr<"TransformMap"> {
     AttrParameter<"AffineMapAttr", "The composed map describing the `ops`">:$map,
     // Thes aren't strictly needed and can probably be removed if we're willing to
     // pay an ergonomics penalty.
-    ArrayRefParameter<"int64_t", "The shape of the transform input">:$upperBounds,
-    ArrayRefParameter<"int64_t", "The shape of the transform output">:$lowerBounds
+    DefaultValuedParameter<"DenseI64ArrayAttr", "$_builder.getDenseI64ArrayAttr({})">:$upperBounds,
```
krzysz00, line 190: Why does need to be `DenseI64ArrayAttr` and not `ArrayRef<int64_t>`

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
```diff
@@ -186,12 +187,12 @@ def Rock_TransformMapAttr : Rock_Attr<"TransformMap"> {
...
+    DefaultValuedParameter<"DenseI64ArrayAttr", "$_builder.getDenseI64ArrayAttr({})">:$lowerBounds
     );
 
     let assemblyFormat = [{
-        `<` $map `by` ` ` `[` $ops `]` `bounds` `=` `[` $upperBounds `]` `->` `[` $lowerBounds `]` `>`
+                             `<` $map `by` ` ` `[` $ops `]` `bounds` `=` ($upperBounds^) : (`[` `]`)? `->` $lowerBounds `>`
```
krzysz00: This could be \`[\` $upperBounds \`]\` and an ArrayRef parameter still, maybe?

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
```diff
@@ -186,12 +187,12 @@ def Rock_TransformMapAttr : Rock_Attr<"TransformMap"> {
...
     }];
```
krzysz00, line 196: Let's do the same syntax for lower bounds, those can be the empty array too

mlir/include/mlir/Dialect/Rock/Passes.td
```diff
@@ -11,9 +11,10 @@
 
 include "mlir/Pass/PassBase.td"
 
-def RockFoldTransposePass : Pass<"rock-fold-transpose", "::mlir::func::FuncOp"> {
-  let summary = "fold transposes to enable fusion with convolution and GEMM";
-  let dependentDialects = ["rock::RockDialect", "linalg::LinalgDialect", "memref::MemRefDialect"];
+def RockViewToTransformPass : Pass<"rock-view-to-transform", "::mlir::func::FuncOp"> {
+  let summary = "expand convolution into coordinate transformations and gemm";
```
krzysz00: ... that's the descriptoin of `rock-conv-to-gemm`. Can you please explain in more detail what this pass does?

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
krzysz00: I'm insisting again that we put this into `tosa-to-rock`

krzysz00: ... Or this view-to-transforms pass

mlir/lib/Dialect/Rock/Transforms/ViewToTransform.cpp
```diff
@@ -0,0 +1,165 @@
...
+    if (!func->hasAttr("kernel")) {
+      return;
+    }
+    MLIRContext *ctx = &getContext();
+    // Split patterns into two stages by bufferization
+    RewritePatternSet patterns(ctx);
+    ConversionTarget target(*ctx);
+
+    target.addLegalDialect<rock::RockDialect,
+                           bufferization::BufferizationDialect>();
```
krzysz00: Why is the bufferization dialect legal here

krzysz00 (commented):
Ok, so, having talked through the things that made me go (!!!), I'm not unhappy with this PR.

And so I'm willing for it to land the way it is if time requires, but, ideally
1. All the new passes should receive high-level descriptions and are otherwise documented so we know what they are. Most importantly **the summary on the `view-to-transforms` pass needs to be accurate**
2. We follow up by upstreaming the tosa-to-tensor split
3. That early tosa-to-tensor gets moved to only running on functions that are getting a `tosa-to-rock`

sjw36:
> Also prior to merging we need to enable checkMIGraphX jenkins param for this PR.

How do you enable this? Not seeing an option in parameters.. I don't see nightly anymore either, did it change?

krzysz00:
> > Also prior to merging we need to enable checkMIGraphX jenkins param for this PR.
> 
> How do you enable this? Not seeing an option in parameters.. I don't see nightly anymore either, did it change?

Are you looking in private or public CI? Public CI doesn't have the nightly toggle or the MIGraphX one, all this gets done on `rocmhead`

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
sjw36: Added this folding back in.

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
sjw36: Moving to the rock only section.

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
sjw36: Empty ArrayRef not supported. Only Attr's can support the optional value in the assemblyFormat field.

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
sjw36: see comment above.

mlir/include/mlir/Dialect/Rock/Passes.td
krzysz00: I don't see this line of code updated with a new description? Or is that a "resolved" in the "I've done this in my local copy and will push it soon" sense?

PR #965, opened 2023-02-08
Replace Unfold with a merge collapse rewrite

krzysz00:
Unfold was an old CK/MIOpen trick for not outputting merge carry logic in places that didn't need it. They handled this by having a coordinate transformation that was like Merge but also assorted that its arguments would be contiguous in the underlying buffer.

Since the project has moved to more general convolution and gemm generation, making that assertion has become difficult, and there've been hacks (the "has_relayout_do_not_unfold" flag) to prevent bugs that would arise from sticking an Unfold{} in that wasn't true. This was getting unsustainable.

@giuseros , in PR #882 / d4e14e155c , put together a newer system for handling the case where we'd break a dimension apart and then put it back together, which removed the need for Unfold and also improved performance by making the Unfold mechanism more general.

However, Unfold itself was not removed by that PR, and neither was the "no carry checks" optimization that it enabled. This PR does both.

In order to recover (and probably improve on) the performance we were getting from Unfold, we introduce collapseContiguousMerges(), a function that takes an array of transformations and rewrites them into functions that are equivalent **for the purposes of indexing some given output shape**. This function will, for example, rewrite the view

"gemmM"@0 <- "k"@0
"gemmK"@1 <- Merge{C, Y, X}("c"@1, "y"@2, "x"@3)

indexing into a [K, C, Y, X] space, into

"gemmM"@0 <- "k"@0
"gemmK"@1 <- Merge{1, 1, CYX}("c"@1, "y"@2, "x"@3)

Equivalently, we replace the map
(m, k) -> (m, k / YX, (k % YX) / X, k % X)
with
(m, k) -> (m, 0, 0, k)

and it will map
(m, n) -> (n / HW, m, (n % HW) / W, n % W)
to
(m, n) -> (n / HW, m, 0, n % HW)

As we do this, we go ahead and introduce a simplification in the affine map building process for `transform_map`s. Namely, we detect the case where we were about to emit (dX mod N) floordiv N and replace that with 0 ourselves, since we know that, while, in general, that is the function that maps positive dX to 0 and negative ones to -1, ... we don't care, because negative indices will lead to the out of bounds detector being triggered and so are sources of undefined behavior when they appear in a transform_map input.

Similarly, if we're about to emit (dN / S) when converting a Merge to an affine map, where S is the stride of the whole input dimension, we instead emit 0, as 0 <= dN < S is assumed to be true of the input to a Merge{}.

These two changes allow us to generalize and simplify the anti-carry check code in SugarToLoops. Since contiguous merges have been collapsed, we will no longer perform the work to ensure accurate carries in cases where they're not needed, such as between the Y and X coordinates of a load from a KCYX tensor. We can also omit putting a carry on a unit dimension, where it'll be promptly passed to the next dimension.

This means that our index diff mechanism generates less code, improving performance.

Performance results will be attached once I have them, but I can verify that the assembly generated for kernels has changed.


krzysz00:
I can report that performance is just about flat
[no-unfolds-perf-conv-gfx90a.html.txt](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10691000/no-unfolds-perf-conv-gfx90a.html.txt)

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -321,16 +321,6 @@ struct ConvertingCopyKernelRewritePattern final
...
-/// In most cases, we can treat the CYX or YXC parts of the filter dimension
-/// as a single unit for the purposes of vectorization, coordinate computations,
-/// and so on. However, if the filter layout is created by transposing the
-/// underlying memory in order to match another tensor, we cannot do this
-/// and so must flag when we have performed such a relayout.
-///
-/// This is a nominally temporary hack that should be removed once we can more
-/// precisely handle Unfold{} by static analysis instead of by heuristic.
-/// As such, it'll probably presist for quite some time.
-constexpr llvm::StringLiteral dontUnfoldAttrName("has_relayout_do_not_unfold");
```
manupak, line 336: Re: fusion refactor, 
ah ha! we had this defined and explained but buried here... 
anyways, doesn't matter as it goes away!

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -699,6 +685,63 @@ int64_t mlir::rock::getMaxVectorization(ArrayAttr transforms, uint32_t dim,
   return result;
 }
 
+ArrayAttr mlir::rock::collapseContiguousMerges(ArrayAttr transforms,
+                                               ArrayRef<int64_t> outputShape) {
+  auto contigousMerges = findContiguousGroups(transforms, outputShape);
```
manupak: nit :  maybe not use auto here.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -699,6 +685,63 @@ int64_t mlir::rock::getMaxVectorization(ArrayAttr transforms, uint32_t dim,
...
+  for (auto map : transforms.getAsRange<TransformMapAttr>()) {
+    bool changed = false;
+    SmallVector<TransformAttr> ops;
+    ops.reserve(map.getOps().size());
+    for (TransformAttr op : map.getOps()) {
+      if (op.getType() != TransformType::Merge) {
+        ops.push_back(op);
+        continue;
+      }
+      auto mergeData = contigousMerges.find({map, op});
```
manupak, line 708: here too.

giuseros (approved):
LGTM (it's nice how can I now understand most of it!)

manupak (commented):
Broadly looks good! (and learnt about this unfold)

few nits..

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
krzysz00: This one's an iterator type, I think it's justified. The previous auto isn't have fixed it.

manupak (approved):
LGTM!

PR #966, opened 2023-02-09
Update the Int8 verifier to handle overflow

yiqian1:
- Implement CPU validation functions for int8 conv and gemm in int64_t in order to detect overflow.
- Change the int8 verifier to use integer comparisons instead of float comparisons.
- Print integer results as integers instead of floats.

### DIFF
--- a/mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
+++ b/mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
@@ -563,13 +563,13 @@ extern "C" void mcpuVerifyInt32(int32_t *gpuAllocated, int32_t *gpuAligned,
   for (int64_t i = 0; i < valSize; ++i) {
     if (gpuAligned[i] != valAligned[i]) {
       if (printDebug)
-        printf("gpu=%d val=%d\n", gpuAligned[i], valAligned[i]);
-      printf("[%d %d %d]\n", 0, 0, 0);
+        printf("%ld: gpu=%d val=%d\n", i, gpuAligned[i], valAligned[i]);
+      printf("[-1]");
       return;
     }
   }
 
-  printf("[%d %d %d]\n", 1, 1, 1);
+  printf("[1 1 1]\n");
   return;
 }
 
@@ -587,12 +587,12 @@ extern "C" void mcpuVerifyInt32Int64(int32_t *gpuAllocated, int32_t *gpuAligned,
         printf("gpu results overflow \n");
     } else if (gpuAligned[i] != valAligned[i]) {
       if (printDebug)
-        printf("gpu=%d val=%ld\n", gpuAligned[i], valAligned[i]);
-      printf("[%d %d %d]\n", 0, 0, 0);
+        printf("%ld: gpu=%d val=%ld\n", i, gpuAligned[i], valAligned[i]);
+      printf("[-1]\n");
       return;
     }
   }
 
-  printf("[%d %d %d]\n", 1, 1, 1);
+  printf("[1 1 1]\n");
   return;
 }
--- a/mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
+++ b/mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
@@ -425,6 +425,13 @@ void printDebugVerifyResults(int64_t dataSize, float maxAbsDiff,
   }
 }
 
+enum class PrintOption : char {
+  Always = 3,  // always print debug info
+  Failure = 2, // print elem-wise diff + summary only if the test fails
+  Summary = 1, // print summary info only if the test fails
+  Off = 0      // do not print debug info
+};
+
 template <typename T>
 void mcpuVerify(T *gpuResults, T *validationResults, int64_t dataSize,
                 float thr_RMS, float thr_absDiff, float thr_relDiff,
@@ -465,12 +472,6 @@ void mcpuVerify(T *gpuResults, T *validationResults, int64_t dataSize,
   constexpr size_t NUM_BUCKETS = NUM_BOUNDARIES + 3;
   int hist_relDiff[NUM_BUCKETS] = {0};
   // Obtain print debug info option
-  enum class PrintOption : char {
-    Always = 3,  // always print debug info
-    Failure = 2, // print elem-wise diff + summary only if the test fails
-    Summary = 1, // print summary info only if the test fails
-    Off = 0      // do not print debug info
-  };
   PrintOption print_option = static_cast<PrintOption>(printDebug);
 
   for (int64_t i = 0; i < dataSize; ++i) {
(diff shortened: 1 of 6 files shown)


mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
```diff
@@ -558,9 +558,41 @@ extern "C" void mcpuVerifyInt32(int32_t *gpuAllocated, int32_t *gpuAligned,
                                 int64_t gpuStride, int32_t *valAllocated,
                                 int32_t *valAligned, int64_t valOffset,
                                 int64_t valSize, int64_t valStride,
-                                float thr_RMS, float thr_absDiff,
-                                float thr_relDiff, char printDebug) {
+                                char printDebug) {
+  assert(gpuSize == valSize);
+  for (int64_t i = 0; i < valSize; ++i) {
+    if (gpuAligned[i] != valAligned[i]) {
+      if (printDebug)
```
krzysz00: Please implement the full range of debug flags here

mlir/test/rocmlir-driver/populate_host_convert.mlir
```diff
@@ -1,13 +1,13 @@
 // RUN: rocmlir-gen --arch %arch -p -ph -pr | FileCheck %s --check-prefix=F32
 // RUN: rocmlir-gen --arch %arch -p -ph -pr -t f16 | FileCheck %s --check-prefixes=F16,CHECK
 // RUN: rocmlir-gen --arch %arch -p -ph -pr -t bf16 | FileCheck %s --check-prefixes=BF16,CHECK
-// RUN: rocmlir-gen --arch %arch -p -ph -pr -t i8 | FileCheck %s --check-prefixes=I8,CHECK
```
krzysz00, line 4: Why is `CHECK` removed here?

krzysz00 (commented):
Some nitpicks about the tests, but also mainly wanting to know why all the debug options aren't implemented

mlir/test/rocmlir-driver/populate_host_convert.mlir
yiqian1: It's used to check that a conversion function exists. We used to convert i32 to f32 and print. I've changed it to directly print out i32. So there shouldn't be a conversion function, just like f32.

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
yiqian1: The debug information is quite simple, so I chose to print them all as long as printDebug is on. Do you prefer different debug information for different debug options? What would you suggest?

zhanglx13: I am not sure if we want to print all the diff's during CI if some int tests fail. It will make the output of lit very confusing.

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
```diff
@@ -558,9 +558,41 @@ extern "C" void mcpuVerifyInt32(int32_t *gpuAllocated, int32_t *gpuAligned,
...
+        printf("gpu=%d val=%d\n", gpuAligned[i], valAligned[i]);
+      printf("[%d %d %d]\n", 0, 0, 0);
```
zhanglx13: Since we are not using the metrics and 0 means the diff is larger than the threshold, maybe put -1 here to indicate this is an integer test failure.

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
yiqian1: Agree. The debug information for float types don't really make sense for integers.

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
yiqian1: OK. I'll make it just print -1.

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
krzysz00: Let me rephrase: we have four debug options: `always`, `failure`, `summary`, and `off`.

All of these options can be meaningfully implemented for the int8 verifier.

`summary` would, for example, be a count of incorrect elements and possibly the largest difference or something like that.

mlir/test/rocmlir-driver/populate_host_convert.mlir
krzysz00: Ah, right, now I'm seeing how that test is structured. Makes sense.

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
krzysz00: @zhanglx13 I disagree. We can think of integers as still using the metrics, except all the metric values are set to `0`.

krzysz00: So it should still be `[0, 0, 0]` here

zhanglx13: Yeah, that's a good interpretation.
I was thinking [0,0,0] means the difference is larger than the threshold. So in the float case, we usually go and check to see if we need to set a larger threshold or there is really a bug. For integer case, [0,0,0] means there is a bug. If we keep using [0,0,0], we need to keep in mind that they mean different things for float and integers.

krzysz00: They ... don't? At least not how I'm thinking of it. An integer computation is one where we don't have acceptable rounding losses at all, so the max RMS error = the max absolute difference = the max relative difference = 0.

zhanglx13: I was saying [0,0,0] has different meanings for int and float.
For int, it means the test fails because of a bug.
For float, it does not always indicate a bug. It usually requires further investigation to identify the issue.
Such information is desired when we see the results from lit. Either it is embedded in the verification result ([-1,-1,-1] means a bug for the int test) or we need to check the data type manually for the test (which is not in the output of lit since the test does not have datatype in their filenames).
But I don't think this makes a big difference since we will examine the failed tests anyway. So I'm ok with either method.

krzysz00: Ah.

I'd want to stick with [0, 0, 0] for consistency.

And I'd at least argue that float tests that aren't `[1, 1, 1]` at least deserve to be looked at very carefully.

zhanglx13: Sure. @yiqian1 you can keep it unless you have a different opinion on this.

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
yiqian1: We could implement that way, but how useful would those information be? I mean I wouldn't care much about the largest difference or the number of incorrect elements. They may help debugging, but collecting those information in CI sounds wasting time for me.

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
yiqian1: No, I don't have a strong opinion on this.

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
krzysz00: This is important to match CI, which, IIRC, calls for `debug=summary` by default

krzysz00: More importantly, when I'm testing, and I set one of those four values for the debug output flag, I expect them to work as specified.

yiqian1: OK, if you insist. For 'failure' I'll print out the correct value, incorrect value and difference, and 'summary' is the number of incorrect elements and the largest difference. Anything else?

krzysz00: Ok, so, the behaviors are this
- On `always`, or for `failure`, or `summary` when the test isn't successful, print some sort of summary (what you suggested about largest difference  + number of failures is good)
- With `always` or `failure`, print out the individual failing elements.

krzysz00: That's what the float verifier does

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
```diff
@@ -553,14 +554,78 @@ extern "C" void mcpuVerifyFloat(float *gpuAllocated, float *gpuAligned,
...
+  int64_t overflow_count = 0; // the number of overflow elements
+  int64_t maxAbsDiff = 0;
+
+  PrintOption print_option = static_cast<PrintOption>(printDebug);
+  for (int64_t i = 0; i < dataSize; ++i) {
+    int64_t valNum = static_cast<int64_t>(valAligned[i]);
+    int32_t gpuNum = gpuAligned[i];
+    if (valNum > INT32_MAX || valNum < INT32_MIN) {
+      overflow_count++;
+      if (print_option == PrintOption::Always)
```
krzysz00, line 570: This is `Always || Failure`, I'd think?

krzysz00 (approved):
One minor bit of weirdness around the if statements aside, this all seems pretty good to me, approved.

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
yiqian1: I'd say overflow indicates bad test vectors for the GPU but not a failure

PR #967, opened 2023-02-09
Remove swizzles from lowering pipeline

krzysz00:
Per https://github.com/ROCmSoftwarePlatform/rocMLIR/discussions/957, swizzles are bad for performance on MI-200, especially when
dealing with GEMMs (which don't have a lot of address calculations to
hide the data movement in), and never really gave us much to begin
with.

They were probably OK on MI-100 because of the agpr->vgpr movements.

They also make delaying vectorization analysis until after fusion
complicated.

We're keeping the `in_warp_transpose` op, though, it's a good example and might be a useful testcase for `amdgpu.dpp`.

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -1141,7 +1141,9 @@ def Rock_XdlopsGemmV2Op:
   let hasVerifier = 1;
 }
 
-// TODO(kdrewnia): upstream these
+/// ONLY NEEDEF FOR in_warp_transpose BELOW. DEAD BUT LEFT IN FOR FUTURE
+/// USE, JUST LIKE in_warp_transpose ITSELF.
+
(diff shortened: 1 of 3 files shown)


jerryyin:
Do you think it's worth doing the other way: Keep the op and the unit test, but remove it from the lowering pass? Regardless the performance implication, I find this overall feature to be well designed. The way you made the unit test work can also be set as examples for other thread level ops too. There's a value in keep it functional and it is not a whole lot of maintenance burden to us.

manupak:
I kind of share the same sentiment. Can we do a warning/deprecation notice (or something) instead ? 
So folks will not actually use it as a base to build other features.. (well unless there is a compelling case..)

krzysz00:
@jerryyin ... yeah, you've got a point. I can scope this down to just the lowering pipeline changes.

In addition, while `gpu.warp_swizzle` is kinda special-purpose, having this op and its unit test around will be useful for when we want to make a `amdgpu.dpp`

jerryyin (approved):

PR #968, opened 2023-02-10
Exclude ck driver build to WA CI failure

jungpark-mlir:
Current CK code keep failing to build and cannot find any stable branch.
exclude building ck driver until we find a solution for it.


krzysz00 (approved):
Approved on the basis that this is a high-importantce hotfix.

As a followup, can you change the CI file move building the rocBLAS and CK drivers into a separate step or steps, with the difference being that if those steps fail, it'll allow  the CI to continue?

PR #969, opened 2023-02-13
Adding quantization unit test and fix fold transpose pass

jerryyin:
I'm adding the tentative quantization unit test and fixing the high level pipeline that doesn't converge for those examples.

Background context: MIGraphX and CK has different implementation for quantization and I don't want to wait till they get uniformed. So I'm posting this review to clear the roadblock from our end first.
 - MIGraphX implementation: [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/op/quantizelinear.hpp#L69-L74)
 - CK implementation: [here](https://github.com/ROCmSoftwarePlatform/composable_kernel/blob/ad541ad6b9de9b0579d5254f82e9d5b86103d309/include/ck/tensor_operation/gpu/element/quantization_operation.hpp#L66-L73)

Once the standard implementation is uniformed, I'll delete the one that doesn't match.

Regarding on `FoldTranspose` fix. Our Tosa pipeline cannot handle `Tosa.cast` operator because it will continually amend the `linalg.generic` input since it is non-trivial transform. Then when it tried to push the transform map to `rock.transform`, it always ended yielding a same `rock.alloc()` + `rock.transform()`, making the fold transpose pass not able to converge after ten runs. I added an additional check and delete the redundant `rock.alloc()` when this happens such that it can converge to finish running through the high level pipelines.

Next step: work through the kernel pipeline and fix any issues there.

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/FoldTranspose.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/FoldTranspose.cpp
@@ -390,6 +390,7 @@ struct FoldRockOutputTransforms : OpRewritePattern<linalg::GenericOp> {
     // Actually do the rewrites, if any
     SmallVector<AffineMap> idxMaps = laGeneric.getIndexingMapsArray();
     SmallVector<Value> inputs = laGeneric.inputs();
+    bool hasReplaced = false;
     for (auto &tuple : toReplace) {
       unsigned opIndex = std::get<0>(tuple);
       Value opValue = std::get<1>(tuple);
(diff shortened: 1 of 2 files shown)


krzysz00:
@jerryyin I like the test, but the FoldTranspose pass is, I think, going away soon.

I can approve this anyway.

jerryyin:
@krzysz00 Thanks. I'm less worried about the actual implementation of the fold transpose pass but more care about the pass of this test. I'd say merge this PR so that:
 - I can unblock and proceed with kernel side investigation
 - And we make sure the test continue to pass once your refactor goes in

krzysz00 (approved):
Approved

sjw36 (approved):
Yes, let's go ahead and merge it. I can take care of the fusion refactor implications.

PR #974, opened 2023-02-15
Fix missing CK driver nightly failure

giuseros:
CK is still under heavy development and might happen that CK doesn't get build. We already disabled the driver, but we also need to disable later stages of the performance pipeline to ensure that problems in CK don't affect benchmark report generation

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -58,10 +58,19 @@ void getAndBuildMIGraphX(String cmakeOpts) {
     buildMIGraphX(cmakeOpts)
 }
 
+CKBuilt = false
 void getAndBuildCK(String cmakeOpts) {
     git branch: params.CKBranch, poll: false,\
         url: 'https://github.com/ROCmSoftwarePlatform/composable_kernel.git'
     buildCK(cmakeOpts)
+}
+
+String getGEMMBenchmarkDrivers(){
+    if (CKBuilt == true){
+        return "rocblas-benchmark-driver ck-benchmark-driver"
+    } else {
+        return "rocblas-benchmark-driver"
+    }
 }
 
 void buildMIOpenWithMLIR() {
@@ -815,7 +824,7 @@ pipeline {
                         steps {
                             sh 'rm -rf composable_kernel'
                             dir('composable_kernel') {
-                                catchError (buildResult: null) {
+                                try {
                                     getAndBuildCK('''
                                         -DGPU_TARGETS="gfx1030;gfx908;gfx90a"
                                         -DCMAKE_CXX_FLAGS="-O3"
@@ -824,6 +833,10 @@ pipeline {
                                         -DCMAKE_BUILD_TYPE=Release
                                         ''')
                                     sh 'cd build/library; make install; cp ../composable_kernelConfig*.cmake ${WORKSPACE}/composable_kernel/build/CKInstallDir/lib/cmake/composable_kernel'
+                                    CKBuilt = true
+                                } catch (err) {
+                                    currentBuild.result = null
+                                    CKBuilt = false
                                 }
                             }
                         }
@@ -833,7 +846,7 @@ pipeline {
                             // Clean up build settings to disable static library and allow
                             // ROCm testing
                             sh 'rm build/CMakeCache.txt'
-                            buildProject('check-rocmlir-build-only ci-performance-scripts rocblas-benchmark-driver',
+                            buildProject("check-rocmlir-build-only ci-performance-scripts ${getGEMMBenchmarkDrivers()}",
                                          '''-DCMAKE_PREFIX_PATH=${WORKSPACE}/composable_kernel/build/CKInstallDir
                                             -DCMAKE_CXX_COMPILER=/opt/rocm/llvm/bin/clang++
                                             -DCMAKE_C_COMPILER=/opt/rocm/llvm/bin/clang''')
@@ -848,7 +861,7 @@ pipeline {
                                 sh """python3 ./bin/perfRunner.py --op=gemm --batch_all \
 					--configs_file=${WORKSPACE}/mlir/utils/performance/bert-configs \
 					--tuning_db=${WORKSPACE}/mlir_tuning_${CHIP}.tsv"""
-                            catchError (buildResult: null) { // Might error if the CK driver has not been built
+                            if (CKBuilt == true) {
                                     sh """python3 ./bin/perfRunner.py --op=gemm --batch_all \
                         --configs_file=${WORKSPACE}/mlir/utils/performance/bert-configs \
                         --tuning_db=${WORKSPACE}/mlir_tuning_${CHIP}.tsv --external-gemm-library CK"""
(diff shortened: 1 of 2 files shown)


jungpark-mlir (approved):
LGTM

giuseros:
Hi @jungpark-mlir , I came up with another idea. We can set a flag to signal if CK has been built and then conditionally build the `ck-benchmark-driver` and (still conditionally) run all the performance scripts. In this way, whenever we get a functioning upstream CK the driver will get built and performance will get compared. What do you think?

jungpark-mlir:
> Hi @jungpark-mlir , I came up with another idea. We can set a flag to signal if CK has been built and then conditionally build the `ck-benchmark-driver` and (still conditionally) run all the performance scripts. In this way, whenever we get a functioning upstream CK the driver will get built and performance will get compared. What do you think?

Yeah, that sounds nice.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -848,9 +863,13 @@ pipeline {
                                 sh """python3 ./bin/perfRunner.py --op=gemm --batch_all \
 					--configs_file=${WORKSPACE}/mlir/utils/performance/bert-configs \
 					--tuning_db=${WORKSPACE}/mlir_tuning_${CHIP}.tsv"""
-                                sh """python3 ./bin/perfRunner.py --op=gemm --batch_all \
-                    --configs_file=${WORKSPACE}/mlir/utils/performance/bert-configs \
-                    --tuning_db=${WORKSPACE}/mlir_tuning_${CHIP}.tsv --external-gemm-library CK"""
+                            script{
+                                if (CKBuilt == true) {
```
manupak: This does not fit well with the stage description : "Test MLIR vs MIOpen", either we have to change the stage name or move this to a different stage such as Test MLIR vs CK.

nit : in that way, we could even use when clause to guard the stage as well..

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -833,7 +848,7 @@ pipeline {
                             // Clean up build settings to disable static library and allow
                             // ROCm testing
                             sh 'rm build/CMakeCache.txt'
-                            buildProject('check-rocmlir-build-only ci-performance-scripts rocblas-benchmark-driver',
+                            buildProject("check-rocmlir-build-only ci-performance-scripts ${getGEMMBenchmarkDrivers()}",
```
manupak: Out of curiosity, does copying the script requires that CK needs to be built ? (assuming all this does is copy the script apart from building ? )

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -815,15 +824,21 @@ pipeline {
...
+                                script{
+                                    try {
+                                        getAndBuildCK('''
+                                            -DGPU_TARGETS="gfx1030;gfx908;gfx90a"
+                                            -DCMAKE_CXX_FLAGS="-O3"
+                                            -DCMAKE_PREFIX_PATH="/opt/rocm"
+                                            -DCMAKE_INSTALL_PREFIX=${WORKSPACE}/composable_kernel/build/CKInstallDir
+                                            -DCMAKE_BUILD_TYPE=Release
+                                            ''')
+                                        sh 'cd build/library; make install; cp ../composable_kernelConfig*.cmake ${WORKSPACE}/composable_kernel/build/CKInstallDir/lib/cmake/composable_kernel'
```
manupak: Can we some how echo the commit here ? 

(Whilst I generally dont like ambiguous dependencies (i.e. CI not defining the commit to build), I felt the general consensus was to be able to check the latest -- thus, it would be great to know the commit of the last successful build)

manupak (commented):
It almost feels like we really want the CK build and test optionally -- i.e. "we'd like to know the numbers if it can build the latest one but if it cant, then we dont want it to interfere the job"

If thats the philosophy, I'd argue that we should make it a separate stage to build, test and measure rather than trickling down conditionals to various stages.

giuseros:
Yes, this is the philosophy: try to build CK and compare against it, don't fail if you can't. 

I guess the point is that there are a lot of things in common with the other perf stages. Build MLIR, build the driver (which shares the same set of utilities), copy the perf scripts and run are all common. The only difference is about what external library to build (and compare against). 

The most time consuming part is to build MLIR: if we think it's fine to build it twice, then it is just a matter of aesthetic, and we can move to a separate stage.

manupak:
> Yes, this is the philosophy: try to build CK and compare against it, don't fail if you can't.
> 
> I guess the point is that there are a lot of things in common with the other perf stages. Build MLIR, build the driver (which shares the same set of utilities), copy the perf scripts and run are all common. The only difference is about what external library to build (and compare against).
> 
> The most time consuming part is to build MLIR: if we think it's fine to build it twice, then it is just a matter of aesthetic, and we can move to a separate stage.

I didn't mean to rebuild MLIR and in fact I was asking the question can we install perf script regardless of the CK build -- because historically it just meant copying the script from utils/* to build/. Then we can move the CK build and measure to different _optional_ stage.

Something like : 

stage("Build MLIR") ...
< other stages that compares with MIOpen >
stage("Build and compare CK")

giuseros:
I followed your advice and the `Jenkinsfile` looks cleaner indeed ! Thanks @manupak !

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -870,7 +879,6 @@ pipeline {
                                 sh 'ls -l'
                                 sh 'python3 ./bin/createPerformanceReports.py ${CHIP}'
                                 sh 'python3 ./bin/createGemmPerformanceReports.py ${CHIP}'
-                                sh 'python3 ./bin/createGemmPerformanceReports.py ${CHIP} CK'
```
krzysz00, line 873: Instead of this, can you go for `|| true` to let this fail silently?

manupak: Should it fail silently if CK was built ?

giuseros: I can, but it looks a bit like a bash hack. What's the drawback in having the CK stage to write its own report?

krzysz00: ... me not noticing that it's doing that x.x

manupak (approved):
LGTM!

A future note : once this is stable and going well. I think we should limit the catch error clause to omit the perfRunner.py as I believe it is completely within our control and any of our PRs should not break that.

giuseros:
I restarted the build and also fixed a +/- 1 bug I realised was there in the drivers.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -848,20 +840,37 @@ pipeline {
...
 					--configs_file=${WORKSPACE}/mlir/utils/performance/bert-configs \
 					--tuning_db=${WORKSPACE}/mlir_tuning_${CHIP}.tsv"""
-                                sh """python3 ./bin/perfRunner.py --op=gemm --batch_all \
-                    --configs_file=${WORKSPACE}/mlir/utils/performance/bert-configs \
-                    --tuning_db=${WORKSPACE}/mlir_tuning_${CHIP}.tsv --external-gemm-library CK"""
                             }
                         }
                     }
-                    stage("Copy earlier performance results") {
+                    stage("Test MLIR vs CK") {
```
jerryyin, line 848: I'd like a `when` clause with 

> equals expected: true, actual: params.checkCK;

This will help CK to be less disruptive in our CI: we can turn it off in jenkins build parameters as soon as it starts to block our CI.

giuseros: Yes, good idea!

PR #977, opened 2023-02-17
LLVM subtree update

jungpark-mlir:


krzysz00 (commented):
Ok, since I'm doing this by running a `git diff` on the command line, these notes will not be conveniently attached to line numbers
- The various changes to `NoMemoryEffects` could/should be the broader `Pure` trait instead, since things like transform aren't just "not affecting memory". (`Pure` also means the operation can be speculated - which it can - and may gain other meanings in the future)
- Nit: `mlir/include/mlir/InitRocMLIRDialects.h` has a trailing space after `ArithDialect`
- Nice cleanup of the I64 array attributes in MIGraphXToTosa
- Your change to LogicalResult BufferLoadOp::verify() is correct but not the clearest. I'd test for the attribute itself being null (aka the default memory space), or that it is the gpu global memory. We don't, for example, want to let `#spirv.cross_function` through
- Over in BlockwiseGemmToThreadwise, `gpu::AddressSpaceAttr::get` can be spelled `b.getAttr<gpu::AddressSpaceAttr>(gpu::GPUDialect::getPrivateAddressSpace())` to avoid having to pull out the context
- (and same for other places)
- It looks like a bunch of `amdgpu.raw_buffer_load` and `amdgpu.raw_buffer_store` tests got affected by the changes to the printer not printing the default value of attributes anymore. On top of removing the `{boundsCheck = true}` that was no longer present in the `CHECK` lines, please 
  - Remove it from the test itself to minimize confusion
  - Add a test case where `boundsCheck = false` and make sure the attribute persists to all the relevant tests
 - In `mlir/tools/rocmlir-gen/rocmlir-gen.cpp`, change the `llvm::Optional` to `std::optional`
 - And in `mlir/lib/Conversion/TosaToRock/TosaToRock.cpp`
 - And so on - basically, `git diff develop mlir` and start looking for `Optional<` that didn't get removed
 - And now looking at the external diff: There might be some trivial reconciliation to do in terms of files that are present in our copy but not in upstream (LLVM docs, some clang tests, etc.) but that's a nit
 - Re the conversions we moved from TosaToLinalg to TosaToTensor, can yo make sure that the code in TosaToTensor matches what is currently (/at the time you took the merge commit) in TosaToLinalg
   - ... ok, yeah, that needs to happen, since, for instance, the current TosaToLinalg reshape stuff uses pattern benefits but the old code that we put in to TosaToTensor doesnt't
 - I'm broadly trusting the external changes
 - Please create a ticket to move to the new fold API
 
Ok, so, overall, thanks for slogging through all this. Here's my feedback. A lot of it is style.

sjw36:
Let's get this reviewed and passing and committed asap. Thanks

krzysz00:
@sjw36 I've reviewed, see above

jungpark-mlir:
Thanks @krzysz00 good points.

Mostly covered now.
- Two tickets opened for memory effect and new fold
- Optional gone except those in RockOps.td tablegen doesn't accept std::optional in some cases.
- buffer load/store tests are tested in AMDGPU dialect, which unfortunately doesn't include checking boundsCheck=false is preserved now. maybe another ticket if we really need to test it.

Yeah, TosaToLinalg and TosaToTensor is a problem, I'll have another look but also need a help from @sjw36

jungpark-mlir:
> * Your change to LogicalResult BufferLoadOp::verify() is correct but not the clearest. I'd test for the attribute itself being null (aka the default memory space), or that it is the gpu global memory. We don't, for example, want to let `#spirv.cross_function` through

@krzysz00 I'm not clear what do you mean how to change it
current logic 
``` 
if (memSpaceValueAttr != nullptr &&
      memSpaceValueAttr.getValue() != gpu::AddressSpace::Global)
    return emitOpError("Destination memref must live in global memory");
```

only allows when it's either `null` or `global`

external/llvm-project/mlir/lib/Conversion/TosaToTensor/TosaToTensor.cpp
```diff
@@ -256,9 +256,8 @@ class SliceOpConverter : public OpConversionPattern<tosa::SliceOp> {
                   ConversionPatternRewriter &rewriter) const final {
     Location loc = sliceOp.getLoc();
     Value input = adaptor.getInput();
-    SmallVector<int64_t> strides;
+    SmallVector<int64_t> strides sizes;
     auto starts = sliceOp.getStart();
-    auto sizes = sliceOp.getSize();
     strides.resize(sliceOp.getType().template cast<ShapedType>().getRank(), 1);
```
jerryyin: sizes used to get initialized before the for loop. This doesn't look like equivalent change before & after.

mlir/tools/xmir-runner/xmir-runner.cpp
```diff
@@ -50,7 +50,7 @@ namespace test {
 void registerTestDialect(DialectRegistry &);
 } // namespace test
 
-static LogicalResult runMLIRPasses(ModuleOp m) {
+static LogicalResult runMLIRPasses(Operation *m, JitRunnerOptions &options) {
```
jerryyin: Leaving a comment since I didn't see `runMLIRPasses()` call sites update at this commit.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -1301,7 +1301,7 @@ LogicalResult BufferLoadOp::verify() {
     return emitOpError("Expected " + Twine(nDims) + " coordinates for load");
   auto memSpaceValueAttr =
       sourceType.getMemorySpace().dyn_cast_or_null<gpu::AddressSpaceAttr>();
-  if (memSpaceValueAttr == nullptr ||
+  if (memSpaceValueAttr == nullptr &&
```
jerryyin: Those `||` to `&&` changes in this commit look suspicious, when it is nullptr, it doesn't make sense to `getValue()` from it

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -1301,7 +1301,7 @@ LogicalResult BufferLoadOp::verify() {
     return emitOpError("Expected " + Twine(nDims) + " coordinates for load");
   auto memSpaceValueAttr =
       sourceType.getMemorySpace().dyn_cast_or_null<gpu::AddressSpaceAttr>();
-  if (memSpaceValueAttr == nullptr &&
+  if (memSpaceValueAttr != nullptr &&
```
jerryyin: Okay, immediately bumped into this commit that resolved my questions. But I assume we also need to nullptr check beforehand? That we disallow none address space?

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -1301,8 +1301,8 @@ LogicalResult BufferLoadOp::verify() {
     return emitOpError("Expected " + Twine(nDims) + " coordinates for load");
   auto memSpaceValueAttr =
       sourceType.getMemorySpace().dyn_cast_or_null<gpu::AddressSpaceAttr>();
-  if (memSpaceValueAttr != nullptr &&
-      memSpaceValueAttr.getValue() != gpu::AddressSpace::Global)
+  if (!(memSpaceValueAttr != nullptr &&
+        memSpaceValueAttr.getValue() == gpu::AddressSpace::Global))
```
jerryyin: Ahh you fixed it here

jerryyin (commented):
Tried my best to left commit based comments. A number of them are cancelling each other. Not too worried as long as we have full pass in tests. I'd prefer passing record of a weekly test before merging.

jerryyin:
@jungpark-mlir If you keep bumping into the tidy issue, try disable it and file a ticket to address later.

https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/615bd027d06b8fed41256e2b803bd61ba66fe6fc/mlir/utils/jenkins/static-checks/premerge-checks.py#L131

mlir/tools/xmir-runner/xmir-runner.cpp
krzysz00: I'm assuming that runMLIRPasses() takes a default argument here

krzysz00:
@jungpark-mlir 

1. Re the buffer ops, I didn't mean the tests in the upstream dialect. I meant the tests in ` mlir/test/Dialect/Rock/buffer-load-merge.mlir`
2. Re `tosa-to-tensor`, since I'm the one who made that split, I can help. The core of it is this: take upstream's copy of `TosaToLinalg`, take the rewrites that create `ExpandShape` and `CollapseShape`, and move them into `TosaToTensor`, along with the code that added those rewrites to the `TosaToLinalg` rewrite set. The simplest way may be to copy in `TosaToLinalg.cpp` and `TosaToTensor.cpp` from upstream and then make the move again 
(though making sure to add any comments I may have added while doing that refactoring)
4. Re the validation logic, which @jerryyin also commented on, we should only accept the op if either
  - The underlying attribute (*before* `dyn_cast_or_null`) is `nullptr` or
  - The attribute, if you `dyn_cast_or_null<gpu::AddressSpaceAttr>` it, is equal to the global address space
  - (The reason we're checking the underlying attribute is that, for example, dyn_cast_or_null<gpu::AddressSpaceAttr>(TransformMapAttr::get(...)) == nullptr` but should be rejected

jungpark-mlir:
> The underlying attribute (before dyn_cast_or_null) is nullptr or
> The attribute, if you dyn_cast_or_null<gpu::AddressSpaceAttr> it, is equal to the global address space

You mean above two cases are fine but
fail if it becomes nullptr after dyn_cast_or_null?

jungpark-mlir:
>Re the buffer ops, I didn't mean the tests in the upstream dialect. I meant the tests in  mlir/test/Dialect/Rock/buffer-load-merge.mlir

Right, just understand and adding a couple of tests

external/llvm-project/mlir/lib/Conversion/TosaToTensor/TosaToTensor.cpp
jungpark-mlir: That's dead code now.

mlir/tools/xmir-runner/xmir-runner.cpp
jungpark-mlir:
```
tools/xmir-runner/xmir-runner.cpp:113:  jitRunnerConfig.mlirTransformer = runMLIRPasses;
lib/ExecutionEngine/JitRunner.cpp:370:    if (failed(config.mlirTransformer(m.get(), runnerOptions)))
```

Just searched in the code and I guess it's called this way, unfortunately changes are spread across the commits.

krzysz00: I don't know if we need to do much here - if it compiles, it compiles

krzysz00:
> > The underlying attribute (before dyn_cast_or_null) is nullptr or
> > The attribute, if you dyn_cast_or_nullgpu::AddressSpaceAttr it, is equal to the global address space
> 
> You mean above two cases are fine but fail if it becomes nullptr after dyn_cast_or_null?

Yep, that's what I mean

jungpark-mlir:
TosaToLinalg ->Tensor merge is redone,
Copy the upstream and redo the move as  suggested, should be clean now.

krzysz00 (approved):
I think you got all my comments, approved.

PR #978, opened 2023-02-17
Adding features to most Rock ops

manupak:
In order to pick the correct atomics in the
global store, we need to know sub-arch features
present. Currently the codebase doesn't plumb
this coherently.

This commit adds the plumbing of the features
for most rock ops that ends up using some sort
of global store.

* adds atomic_fmax_f32 feature

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/814


mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -201,7 +202,7 @@ def Rock_ReduceOp :
 def Rock_ZeroInitKernelOp :
     Rock_Op<"zero_init_kernel", []>,
     Arguments<(ins AnyTensorOrMemRef:$buffer,
-                  StrAttr:$arch,
```
krzysz00, line 204: Don't remove `arch` here - it may still be useful

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -201,7 +202,7 @@ def Rock_ReduceOp :
...
+                  Rock_GemmFeaturesAttr:$features,
```
krzysz00, line 205: Also, I don't think this kernel needs the features, unless you're adding them to the gemm wrapper interface?

krzysz00 (changes requested):
Don't remove the `arch` field - we still want it around

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: ok, so two points here : 
* We dont use it as of now -- nothing breaks if I remove it.
* Should we be really passing the string target-triple ? especially given that we have the bag of arch subfeatures (i.e. GemmFeatures) that could be passed instead. I really think we should not be putting in code in lowering : 
```
if arch == "gfxXXX": 
  do something
else if arch == "gfxYYY":
  do something else
```

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: Probably you are right... 

I just added features to every kernel that ends up using a buffer store which needs to know the features because from a layering perspective how buffer store is lowered kept independent of the kernel. However, since we know this kernel always want the buffer store lowered in a certain way that does not require features, we might as well put features::none for the buffer store.

manupak (commented):
Thanks @krzysz00 for quick reviews.

I ll probably remove the features from the zero init kernel but I think we should not be passing the string target triple around.

manupak:
Ok, I've removed features from utility kernels.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
krzysz00: We will be putting that in when making decisions about which xdlops are available

manupak: ok, please help me understand this : 

right now If I do a grep for "arch" or "Arch" in the Rock/ : 

```
manupa@manupa-System-Product-Name:~/work/rocMLIR/mlir/lib/Dialect/Rock$ grep "arch" . -r
./Generator/AmdArchDb.cpp:AmdArchInfo mlir::rock::lookupArchInfo(StringRef arch) {
./Generator/AmdArchDb.cpp:  std::tie(firstPart, remainingParts) = arch.split(':');
./Generator/AmdArchDb.cpp:  llvm::errs() << "Warning: unknown architecture, falling back to defaults: "
./Generator/AmdArchDb.cpp:               << arch << "\n";
./Generator/Conv2dGenerator.cpp:    const std::string &arch, const std::string &chip, const std::string &triple,
./Generator/Conv2dGenerator.cpp:    : config{arch,
./Generator/Conv2dGenerator.cpp:  std::string arch;
./Generator/Conv2dGenerator.cpp:  strToStr("arch", arch);
./Generator/Conv2dGenerator.cpp:  if (failed(splitter.parse(arch))) {
./Generator/Conv2dGenerator.cpp:  // Canonicalize architecture name
./Generator/Conv2dGenerator.cpp:  arch = canonicalArch.str();
./Generator/Conv2dGenerator.cpp:  config.arch = arch;
./Generator/Conv2dGenerator.cpp:  AmdArchInfo archInfo = lookupArchInfo(splitter.getChip());
./Generator/Conv2dGenerator.cpp:  config.features = archInfo.defaultFeatures;
./Generator/Conv2dGenerator.cpp:  StringAttr archStrAttr = builder.getStringAttr(config.arch);
./Generator/Conv2dGenerator.cpp:  NamedAttribute archAttr = builder.getNamedAttr("xmodel.arch", archStrAttr);
./Generator/Conv2dGenerator.cpp:      archAttr};
./Generator/Conv2dGenerator.cpp:    module->setAttr(archAttr.getName(), archAttr.getValue());
./Generator/Conv2dGenerator.cpp:      builder.getNamedAttr("arch", archStrAttr),
./Tuning/ConvContext.cpp:  auto archVal = op->getAttrOfType<StringAttr>("arch").getValue();
./Tuning/ConvContext.cpp:  return {archVal,     numCu,      opType, dimIndexAndSize, strideVal,
./Tuning/RockTuningImpl.cpp:// Brute-force search in incremental order
./Tuning/RockTuningImpl.cpp:  auto search = perfTable->tuningMap.find(problem);
./Tuning/RockTuningImpl.cpp:  if (search != perfTable->tuningMap.end()) {
./Tuning/RockTuningImpl.cpp:  auto search = perfTable->tuningMap.find(problem);
./Tuning/RockTuningImpl.cpp:  if (search != perfTable->tuningMap.end()) {
manupa@manupa-System-Product-Name:~/work/rocMLIR/mlir/lib/Dialect/Rock$ grep "Arch" . -r
./Generator/CMakeLists.txt:  AmdArchDb.cpp
./Generator/AmdArchDb.cpp://===- AmdArchDb.cpp - Dtabase of AMD GPU features ------------------===//
./Generator/AmdArchDb.cpp:#include "mlir/Dialect/Rock/Generator/AmdArchDb.h"
./Generator/AmdArchDb.cpp:static constexpr AmdArchInfo gcnInfo(GemmFeatures::none, /*waveSize=*/64),
./Generator/AmdArchDb.cpp:AmdArchInfo mlir::rock::lookupArchInfo(StringRef arch) {
./Generator/AmdArchDb.cpp:    return llvm::StringSwitch<AmdArchInfo>(minor)
./Generator/AmdArchDb.cpp:        .Case("06", AmdArchInfo(GemmFeatures::dot, /*waveSize=*/64))
./Generator/AmdArchDb.cpp:    return llvm::StringSwitch<AmdArchInfo>(minor)
./Generator/Conv2dGenerator.cpp:#include "mlir/Dialect/Rock/Generator/AmdArchDb.h"
./Generator/Conv2dGenerator.cpp:  SmallString<64> canonicalArch;
./Generator/Conv2dGenerator.cpp:  splitter.getFullName(canonicalArch);
./Generator/Conv2dGenerator.cpp:  arch = canonicalArch.str();
./Generator/Conv2dGenerator.cpp:  AmdArchInfo archInfo = lookupArchInfo(splitter.getChip());
./Tuning/RockTuningImpl.cpp:  problemOS << gemmIF.getArch() << tab;
./Transforms/ConvToGemm.cpp:      /*cTransposed=*/nullptr, op.getArchAttr(), op.getNumCuAttr(),
./Transforms/ConvToGemm.cpp:      /*cTransposed=*/nullptr, op.getArchAttr(), op.getNumCuAttr(),
./Transforms/ConvToGemm.cpp:                     /*cTransposed=*/nullptr, op.getArchAttr(),
```

"arch" only seems to be present in generators and rock::GemmOp. So generators parse the arch and generate features and use them throughout the lowering (rightfully so). rock::GemmOp only needed it because GridwiseGemm needed it and so on -- but none of the lower level ops seems never really used it. 

Are you talking about an upcoming feature ?

krzysz00: Yeah, I'm talking about the upcoming need to, for example, not select fp8 when we're not on MI-300, or, more usefully, the ability to select the MI-200+ xdlops when we're on something newer than MI-100

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
krzysz00: ... Actually, good point on adding the features to the elementwise kernels, you might as well do it that way.

krzysz00:
Sorry, I missed the part where you had a good point about adding features to the utility kernels.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: Hmmm, why cannot we resolve the target-triple to features beforehand for those (a change in AmdArchDB)? -- rather than having to pass around arch string to decide in the lowering.

(Roughly being analogous to target features of AMDGPU backend of LLVM)

manupak: My concern is AMDGPU arch versions will add and drop features (as we are seeing with fmax for e.g.) and I personally feel best way to control the impact on the our lowering is to resolve the arch string to features early on (as we do with GemmFeatures) and only use those in the lowering. WDYT ?

krzysz00: It's not an unreasonable decision, though I do think things like the xdlops database might want to be keyed by arch.

(You'll note, for example, that the `amdgpu` dialect has `chipset` and does a lot of `chipset.minorVersion >= 0x0a` type stuff)

Though ... yeah, having stuff like "bf16_mfma" and "fp8" gated by features is also sensible, and we could key the xdlops selector by feature instead.

... I'll approve dropping `arch` as an attribute, yeah.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -191,7 +192,7 @@ def Rock_ReduceOp :
   let summary = "Axes-wide reduction operation";
   let hasVerifier = 1;
   let assemblyFormat = [{
-    $reduceMethod $in `into` $out attr-dict `:` type($in) `into` type($out) (`->` type($result)^)?
+    $reduceMethod $in `into` $out `features` `=` $features attr-dict `:` type($in) `into` type($out) (`->` type($result)^)?
```
krzysz00, line 195: ... hold on, since we're here, can you do me a favor and see what happens if we don't explicitly mention `$features`? I wonder how ugly the syntax will get. That's probably a future PR though

krzysz00 (approved):
Approved, you've got a point. Might be a good idea to put the features on the elementwise kernels though.

manupak:
Thanks @krzysz00 for reviews! I ll take a look at dropping features as a special attr if that what you meant a bit later -- i just have lot of open PRs that I'd like to close out first :) that have a bit of dependency chain. Also, I've reverted and added back features to utility kernel.

krzysz00:
@manupak I more meant seeing if we can git rid of some of the special syntax now that there've been some improvements to the parser/printer generator upstream. For example, we might be able to get ride of the parentheses around `storeMethod`. Which is something I may take up.

PR #979, opened 2023-02-17
MLIR#808: fixed rock-linalg-align for mixed types

sjw36:
* added I8 support for rock.threadwise_write_all
* updated fusion/quantization.mlir test to run kernel pipeline

https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/808

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
@@ -221,8 +221,8 @@ static void reconfigureLAGeneric(PatternRewriter &b,
 
   // 2.3. Reset iterator types
   SmallVector<StringAttr, 5> lgIterArr(regRank, b.getStringAttr("parallel"));
-  laGeneric.setIteratorTypesAttr(b.getArrayAttr(ArrayRef<Attribute>(
-      lgIterArr.begin(), lgIterArr.end())));
+  laGeneric.setIteratorTypesAttr(
+      b.getArrayAttr(ArrayRef<Attribute>(lgIterArr.begin(), lgIterArr.end())));
 }
 
 static Value findThreadwiseWrite(linalg::GenericOp laGeneric,
@@ -273,8 +273,7 @@ LAGenericRewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
     return failure();
 
   auto outType = out.getType().cast<ShapedType>();
-  auto inpType =
-      laGenericInputLeadingToGemmStore.getType().cast<ShapedType>();
+  auto inpType = laGenericInputLeadingToGemmStore.getType().cast<ShapedType>();
   if (outType.getShape() != inpType.getShape()) {
     return laGeneric.emitError("input and output types must match");
   }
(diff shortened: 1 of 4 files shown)


mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -96,24 +96,29 @@ static Value transformAccordingly(PatternRewriter &b, Value nOut, Value refOp) {
...
   }
 
   return nOut;
 }
 
+Value makeRegs(PatternRewriter &b, MemRefType::Builder &mrb, Location loc,
+               Type srcType) {
+  auto srcMemType = srcType.cast<MemRefType>();
+  // 1. create a second allocation of the same type to hold loaded elements
+  return b.create<GpuAllocOp>(loc, static_cast<MemRefType>(mrb.setElementType(
```
krzysz00, line 110: Address space needs to be set here.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -183,44 +188,40 @@ static ThreadwiseWriteAllOp traceToThreadwiseWrite(Value inp) {
...
       newInput = applyTransforms(b, twWriteOp, inp);
     }
     newInputs.push_back(newInput);
-    laGenericAMaps.push_back(AffineMap::getMultiDimIdentityMap(
-        newInput.getType().template cast<MemRefType>().getRank(), ctx));
-  }
 
-  laGenericAMaps.push_back(
-      AffineMap::getMultiDimIdentityMap(regType.getRank(), ctx));
+    auto inpRank = newInput.getType().template cast<ShapedType>().getRank();
```
krzysz00: No need to say `template` here

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -271,10 +272,10 @@ LAGenericRewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
   if (!laGenericInputLeadingToGemmStore)
     return failure();
 
-  auto actualLAGenericOut = laGeneric.getOutputOperand(0);
-  if (laGenericInputLeadingToGemmStore.getType() !=
-      actualLAGenericOut->get().getType()) {
-    // TODO(sjw): fix for mixed types
+  auto outType = out.getType().template cast<ShapedType>();
+  auto inpType =
+      laGenericInputLeadingToGemmStore.getType().template cast<ShapedType>();
```
krzysz00: Same nit note on the template bit

krzysz00 (changes requested):
You're creating `GpuAlloc` ops without specifying the private address space. This can cause issues.

Also, you need to also update GlobalStore, BufferLoad, ... to list `I8`

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
sjw36: It gets that from the MemRefType::Builder which is initialized by the register type for the blockwise_gemm output.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
sjw36: yes, copied code.. it is ugly.

sjw36:
> Also, you need to also update GlobalStore, BufferLoad, ... to list `I8`

Updated rock.global_store. buffer_load already had I8.

jerryyin (approved):
Nice work enabling the alignment of memref of different type but same shape. LGTM

krzysz00 (approved):
Looks good to me

PR #982, opened 2023-02-17
[DO NOT SQUASH] MLIR#807: add tosa.cast to fusable elementwise ops in tosa-partition

sjw36:
MLIR#812: fixed bug in tosa-to-linalg-named for quantized bias add
* added test
    
    https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/812

MLIR#813: fuse all subsequent transpose/reshape ops in tosa-partition
* also make tosa-partition run bottom-up (should use a pass instead)
* changed tosa-partition=trailing-only to default true
* updated tosa-partition tests
MLIR#807: add tosa.cast to fusable elementwise ops in tosa-partition
    
    https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/807
    https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/813

* fixed rock.transform_map for empty lowerBounds
* updated test for quantization kernel in partitioning

### DIFF
--- a/external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
+++ b/external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
@@ -63,7 +63,8 @@ public:
   bool isLeadingOp(Operation *op) const;
   bool isTrailingOp(Operation *op) const;
   StringRef partitionTag() const;
-  void traceInputs(Operation *op, SetVector<Operation *> &predecessors,
+  void traceInputs(Operation *op, Operation *ignoreOp,
+                   SetVector<Operation *> &predecessors,
                    SetVector<Value> &inputNodes);
   void runOnOperation() override;
 };
@@ -77,7 +78,8 @@ bool isElementwiseOp(Operation *op) {
   return op->hasTrait<OpTrait::Elementwise>() ||
          op->hasTrait<OpTrait::ResultsBroadcastableShape>() ||
          // clang-format off
-    isa<tosa::ClampOp,
+    isa<tosa::CastOp,
+        tosa::ClampOp,
         tosa::SigmoidOp,
         tosa::TanhOp,
 // ResultsBroadcastableShape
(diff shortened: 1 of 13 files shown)


mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
```diff
@@ -192,7 +192,7 @@ def Rock_TransformMapAttr : Rock_Attr<"TransformMap"> {
     );
 
     let assemblyFormat = [{
-        `<` $map `by` ` ` `[` $ops `]` `bounds` `=` ($upperBounds^) : (`[` `]`)? `->` $lowerBounds `>`
+        `<` $map `by` ` ` `[` $ops `]` `bounds` `=` ($upperBounds^) : (`[` `]`)? `->` ($lowerBounds^) : (`[` `]`)? `>`
```
manupak, line 199: Why this change ?

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -562,7 +567,7 @@ void TosaPartitionPass::runOnOperation() {
 
       // Grab a useful set of leading ops, like we do for trailing.
       SetVector<Operation *> leadingOps;
-      traceInputs(anchorOp, leadingOps, inputNodes);
+      traceInputs(anchorOp, anchorOp, leadingOps, inputNodes);
```
manupak, line 569: Why anchor op as ignore op here ?

manupak (commented):
Changes broadly looks good!
just two questions..

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
sjw36: I found with new partitioning where an empty lower bounds happens. It's like a memref.collapse_shape : memref<1x1x1xf32> into memref<f32>

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
sjw36: Since this routine is now used below for walking up inputs to trailing elementwise ops, the ignoreOp was added so we don't recurse back up the anchor op path. But here it is mostly a placeholder (vs nullptr) since we want to walk back every input of the anchor op.

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
manupak: Ack!

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
manupak: Ack!

manupak (approved):
LGTM!

pcf000 (approved):
A few things I did differently in my generic-partition branch;  we'll harmonise later.

PR #983, opened 2023-02-20
Limit the amount of vectorization

giuseros:
This PR is a stab to start improving GEMM performance.
- **Limit vectorization to the hardware vector length**. Vectorizing more than 128 bits is detrimental for performance, since we ask too much from a thread. 
- **Extend the tilings we use for `int8`**. Limiting the tiles to `32,32` (m/n repeats = 1) jointly with limiting the vectorization means we ask too few from the warps. Since we can now support bigger tiles, I added them and also added a case for KPack=16. 

This patch improve performance from 2% to 90% across the board

The improvement is attached. In any test I did, I always saw this sort of improvement. 

[results.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10785464/results.zip)

### DIFF
--- a/mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
+++ b/mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
@@ -218,6 +218,8 @@ auto getMfmaInsnGroupAttrMap = []() {
                     {ROCDL::mfma_f32_32x32x4bf16::getOperationName()}},
                    {{MfmaTypeId::Bf16TyId, 16, 16},
                     {ROCDL::mfma_f32_16x16x8bf16::getOperationName()}},
+                   {{MfmaTypeId::I8TyId, 64, 64},
+                    {ROCDL::mfma_i32_32x32x8i8::getOperationName()}},
                    {{MfmaTypeId::I8TyId, 32, 32},
                     {ROCDL::mfma_i32_32x32x8i8::getOperationName()}},
                    {{MfmaTypeId::I8TyId, 16, 16},
--- a/mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
+++ b/mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
@@ -158,6 +158,7 @@ auto getMfmaInsnGroupAttrMap = []() {
   using amdgpu::MFMAPermB;
   static llvm::DenseMap<MfmaInsnGroupSelectKey, MfmaInsnGroupAttr,
                         MfmaInsnGroupSelectKeyInfo>
+      // f32
       groupAttrMap{{{MfmaTypeId::Fp32TyId, 64, 64},
                     {ROCDL::mfma_f32_32x32x2f32::getOperationName()}},
                    {{MfmaTypeId::Fp32TyId, 64, 32},
@@ -178,6 +179,7 @@ auto getMfmaInsnGroupAttrMap = []() {
                     {ROCDL::mfma_f32_32x32x2f32::getOperationName()}},
                    {{MfmaTypeId::Fp32TyId, 16, 16},
                     {ROCDL::mfma_f32_16x16x4f32::getOperationName()}},
+                   // f16
                    {{MfmaTypeId::Fp16TyId, 64, 64},
                     {ROCDL::mfma_f32_32x32x8f16::getOperationName()}},
                    {{MfmaTypeId::Fp16TyId, 64, 32},
@@ -198,6 +200,7 @@ auto getMfmaInsnGroupAttrMap = []() {
                     {ROCDL::mfma_f32_32x32x8f16::getOperationName()}},
                    {{MfmaTypeId::Fp16TyId, 16, 16},
                     {ROCDL::mfma_f32_16x16x16f16::getOperationName()}},
+                   // bf16
                    {{MfmaTypeId::Bf16TyId, 64, 64},
                     {ROCDL::mfma_f32_32x32x4bf16::getOperationName()}},
                    {{MfmaTypeId::Bf16TyId, 64, 32},
@@ -218,10 +221,19 @@ auto getMfmaInsnGroupAttrMap = []() {
                     {ROCDL::mfma_f32_32x32x4bf16::getOperationName()}},
                    {{MfmaTypeId::Bf16TyId, 16, 16},
                     {ROCDL::mfma_f32_16x16x8bf16::getOperationName()}},
+                   // Int8
                    {{MfmaTypeId::I8TyId, 64, 64},
+                    {ROCDL::mfma_i32_32x32x8i8::getOperationName()}},
+                   {{MfmaTypeId::I8TyId, 64, 32},
+                    {ROCDL::mfma_i32_32x32x8i8::getOperationName()}},
+                   {{MfmaTypeId::I8TyId, 32, 64},
                     {ROCDL::mfma_i32_32x32x8i8::getOperationName()}},
                    {{MfmaTypeId::I8TyId, 32, 32},
                     {ROCDL::mfma_i32_32x32x8i8::getOperationName()}},
+                   {{MfmaTypeId::I8TyId, 64, 16},
+                    {ROCDL::mfma_i32_16x16x16i8::getOperationName()}},
+                   {{MfmaTypeId::I8TyId, 16, 64},
+                    {ROCDL::mfma_i32_16x16x16i8::getOperationName()}},
                    {{MfmaTypeId::I8TyId, 16, 16},
                     {ROCDL::mfma_i32_16x16x16i8::getOperationName()}}};
 
--- a/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
@@ -91,7 +91,7 @@ static Type obtainAccumulatorType(OpBuilder &b, Type &elementType,
 static std::pair<GemmDimension, int64_t>
 bestVectorization(OpBuilder &b, Value matrix, int64_t dataPerThread,
                   GemmDimension tiebreaker, int64_t kPerBlock,
-                  int64_t dPerBlock) {
+                  int64_t dPerBlock, Type elementType) {
   Value tensor;
   ArrayAttr transforms;
   std::tie(tensor, transforms) = untransform(b, matrix);
@@ -104,6 +104,15 @@ bestVectorization(OpBuilder &b, Value matrix, int64_t dataPerThread,
   int64_t dVectorLen = getMaxVectorization(
       transforms, static_cast<uint32_t>(GemmDimension::MorN),
       math_util::gcd(dataPerThread, dPerBlock), tensorShape);
+
+  // Vectorizing more than the physical vector length (128 bits) might
+  // be harmful for coalescence and other metrics. Let's limit the maximum
+  // amount of data to load to the maximum vector length. This means a
+  // warp will issue, if possible, a global_load_dwordx4 instruction
+  const int64_t maxVectorLenBits = 128;
+  auto bwidth = elementType.getIntOrFloatBitWidth();
+  kVectorLen = std::min(maxVectorLenBits / bwidth, kVectorLen);
+  dVectorLen = std::min(maxVectorLenBits / bwidth, dVectorLen);
 
   if (kVectorLen > dVectorLen)
     return {GemmDimension::K, kVectorLen};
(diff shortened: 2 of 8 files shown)


mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
```diff
@@ -218,6 +218,8 @@ auto getMfmaInsnGroupAttrMap = []() {
                     {ROCDL::mfma_f32_32x32x4bf16::getOperationName()}},
                    {{MfmaTypeId::Bf16TyId, 16, 16},
                     {ROCDL::mfma_f32_16x16x8bf16::getOperationName()}},
+                   {{MfmaTypeId::I8TyId, 64, 64},
+                    {ROCDL::mfma_i32_32x32x8i8::getOperationName()}},
```
jerryyin, line 226: I'm surprised I have missed this xdlops in the past. Just double check: This is available in all gfx908+ architectures, right?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -969,6 +969,15 @@ struct GridwiseGemmV2RewritePattern
...
         b, matB, bCopyPerThread, vectorTiebreaker, kPerBlock, nPerBlock);
 
+    // Vectorizing more than the physical vector length (128 bits) might
+    // be harmful for coalescence and other metrics. Let's limit the maximum
+    // amount of data to load to the maximum vector length. This means a
+    // warp will issue, if possible, a global_load_dwordx4 instruction
+    const int64_t maxVectorLenBits = 128;
+    auto bwidth = elementType.getIntOrFloatBitWidth();
+    aVectorLen = std::min(maxVectorLenBits / bwidth, aVectorLen);
+    bVectorLen = std::min(maxVectorLenBits / bwidth, bVectorLen);
```
jerryyin: I more think this should be moved inside `bestVectorization()` because we use this routine in a number of cases. So you might get even more perf improvements in this PR.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -299,6 +299,8 @@ PopulateParamsXDL::isValidBlockwiseGemmXDLOPS(const InitParamsXDL &param,
     // clang-format off
     validWaveGemmSize = {
       std::make_tuple(32, 32, 2),
+      std::make_tuple(64, 64, 2),
+      std::make_tuple(128, 128, 2),
```
jerryyin: Good job, thanks!

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -42,7 +42,7 @@ void createGemmTuningRangeBF(struct TunableParams *newSpace,
       {8, 16, 32},
       {4, 8, 16, 32, 64, 128},
       {4, 8, 16, 32, 64, 128},
-      {1, 4, 8},
+      {1, 4, 8, 16},
```
jerryyin, line 45: I'm giving you a heads up that too wide kPack might cause issues in conv side because we may run out of registers. So I think you may want to wait for my PR to merge and then run this through convolution tuning of the CI to make sure not regression will show up.

jerryyin (commented):
Excellent PR, glad you observed the opportunity in perf improvement. Only minor nits here and very close to merge once couple of questions are addressed.

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
giuseros: Yes, this was already present in our database. What was missing was using tiles bigger then `32x32` so to let the warps iterate over more tiles (i.e., repeats > 1). I am not sure that was needed, since we were using large vectorization. But now that I limited the vectorization is sensible to extend this, so that a single warp can do more

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
giuseros: Yes, this was an idea that crossed my mind, but the issue is that I only know that there is the `global_load_dwordx4` in the ISA. I am not sure if there is the a similar instruction for LDS loads and if the maximum length of LDS vectors is still 128 bits.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
giuseros: No problem! Actually, I was thinking: now that we have reduction fully supported, should we remove having different sets for int8 and floating points? I mean we can try the same exact configurations on all the data types.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
giuseros: This is only the tuning though. I mean, if in convolution 16 is too big, KPack 4 or 8 will be selected. I am only suggesting to the tuner to try an additional  kPack  for int8. Actually, I also see that the search space is similar for int8 and floating points. Should we unify that as well?

jerryyin: > I am only suggesting to the tuner to try an additional kPack for int8. 

If it is only for our end, then I'm good. However, if there's a perf benefit from the extended kPack range, it'd be bad if MIOpen tuning can't take advantage of it. The thing in MIOpen end is, we have a requirement that once the applicability check pass, kernel run shouldn't fail.

This is the kind of case where the applicability check will indeed pass because ~~there's no way to tell if registers are used up. However, when it fail, it fail in generation of the kernel.~~ (BTW I think there's a way to tell in advance because we have encoded architecture and CU so we can have that information. But it's going to be whole another feature to work with)

>  I also see that the search space is similar for int8 and floating points. Should we unify that as well?

You can give it a shot but I suspect it will run into exact issue as I mentioned above. A longer K dimension might cause non-int8 type to file after we report ourselves to be applicable, it will end to be an exception on MIOpen end. That's the main reason I want the change to be tested through MIOpen.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
jerryyin: Seems fair, yes we should give it a shot and see how much perf impact it is.

Though if there's not a lot then we should fall back to existing narrow pattern because then it filtered out a lot of tuning space.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: My impression is that 128 is the upper limit of hardware vectorization but I can be wrong.

@krzysz00 Opinion on LDS vectorization size?

krzysz00: There is a 128-bit LDS vector load, yeah

(You may not see it because the compiler uses the 2 64-bit loads version when the LDS access is unassigned)

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
krzysz00: Re kPack - if we're limiting vectorization to 128 bits, if there are no xdlops that take more than 128 bits worth of k per thread, we _might_ want to considered limiting kPack in tuning selection to increase tuning performance?

Or, alternatively, report inapplicable perf configs when kPack > 128 / sizeInBits(dtype) &&  kPack != k_per_lane(selected xdlops)}

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
giuseros: I investigated this yesterday, and yes, the register structure is such that 128 can be seen as a hardware constant: more details in https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/731

I am attaching also the performance report for moving the vectorization limit into `bestVectorization`.  There is a small improvement (I think) but it's pretty much the same. 

[new_results.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10802579/new_results.zip)

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
giuseros: I will tune again and see if there is any improvement

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
giuseros: Can we have a chat about this? I am still unclear how MIOpen uses our things and this might be a good occasion for me to learn that

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
giuseros: So yes, few int8 scenarios got a small improvement (7 to 11 %). 

I am attaching the performance comparison here (please, ignore f32 and f16, since they are exactly the same)

[enlarge_int8_space.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10804246/enlarge_int8_space.zip)

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
krzysz00: Sure - let's bring Jerry in on it though, since he can find the MIOpen code faster than I can

krzysz00: (... alternative thought, we - and MIOpen - may want to limit the kPack value to what's applicable for a given datatype)

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: I'd go ahead and make the move, and that'll make `getMaxVectorization()` take a data type argument like it logically ought to

giuseros:
Hi @krzysz00 , @jerryyin 
This is the results of convolution. The majority is green, but there are also few suspicious reds. Should I pursue that further? 

The methodology I followed is:
- Run `tune_MLIR_kernels.sh` (this PR vs `origin/develop`). I enhanced the MIOpen branch with the additional value for KPack. 
- Run `./bin/perfRunner --miopen_use_tuned_mlir`
- Run `./perfRegressionReport gfx90a old.csv new.csv conv.html`

Is there anything I am missing? Or any way I can do this better?

[conv.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10840344/conv.zip)

krzysz00:
Per discussions in chat, I think this is a solid perf improvement and we should merge.

giuseros:
Hi @krzysz00 , @jerryyin ,
Update on this. The deeps (and also the very high peaks) that I found during perf testing were due to instability of the `MIOpenDriver`: if I ran the driver multiple times, I would get sporadic high times, like @krzysz00 suggested (this was not reproducible using `rocmlir-gen`). 

I would postpone the tuner optimization (reducing the search space, or partitioning the search space on a data type basis). So in this PR I:
- Limit the amount of vectorization (I moved the limit to `bestVectorization`)
- Increase the tilings supported by xdlops for int8 (I also added irregular tiles, like 64x32 or 128x64 ) like we do for f32.

krzysz00:
@giuseros The instability does appear during GEMM tuning/when using `rocmlir-gen` and friends.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -105,6 +105,15 @@ bestVectorization(OpBuilder &b, Value matrix, int64_t dataPerThread,
       transforms, static_cast<uint32_t>(GemmDimension::MorN),
       math_util::gcd(dataPerThread, dPerBlock), tensorShape);
 
+  // Vectorizing more than the physical vector length (128 bits) might
+  // be harmful for coalescence and other metrics. Let's limit the maximum
+  // amount of data to load to the maximum vector length. This means a
+  // warp will issue, if possible, a global_load_dwordx4 instruction
+  const int64_t maxVectorLenBits = 128;
+  auto bwidth = elementType.getIntOrFloatBitWidth();
+  kVectorLen = std::min(maxVectorLenBits / bwidth, kVectorLen);
```
krzysz00: One note: gcd, not min. A best vectorization of N means that you can vectorize with any `d` that divides `N`, not any `d < N`

krzysz00 (commented):
Do you want to push the vectorization limit even further, down into the vectorization inference itself? 

Also, if you're going to put this limit in, make sure to get the `threadwise_write_all` and `threadwise_read_into` patterns in BlockwiseGemmToThreadwise as well

jerryyin:
> The deeps (and also the very high peaks) that I found during perf testing were due to instability of the MIOpenDriver: if I ran the driver multiple times, I would get sporadic high times

Actually MIOpenDriver would run a command multiple times, getting rid of highest and lowest and take average out the only intermediate performance. The sporadic high times might be normal because everything is warmed up. I suspect, on the other hand, our profiling take everything too conservatively, and only get data from a cold start.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
giuseros: Yes, you are right, good catch!

giuseros:
Hi @krzysz00 , I added a new function `getMaxVectorizationForDatatype` which invokes `getMaxVectorization` and limits the length for the passed data type. I also updated all uses of `getMaxVectorization()` with this newly added function.

krzysz00 (approved):
Looks good to me here

jerryyin (approved):

PR #984, opened 2023-02-20
Improve default performance by adding up-to-date default configs

giuseros:
In PR https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/983 we extended the tiling values supported for `int8` (which were need since we were limiting the vectorization). 

This PR takes a step forward to update the default tuning configs using the ones that were found fastest on the bert database ( [I used the results obtained from PR #983](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10786432/results.zip)). 

Since now everything is a reduction, I thought also of unifying `int8` vs non-`int8` default configs. 

I am marking this dependent on #983 so that I can test default performance (i.e., `--perf_conifg=`) once #983 has been merged.

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
@@ -216,7 +216,7 @@ public:
 
 class PopulateParamsXDL : public BasePopulateParams<InitParamsXDL> {
 private:
-  static constexpr size_t nInitParameters = 4;
+  static constexpr size_t nInitParameters = 5;
   // Initial tuning parameters for forward convolution and backward
   // convolution.
   static const InitParamsXDL initParameters[nInitParameters];
--- a/mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
@@ -216,14 +216,10 @@ public:
 
 class PopulateParamsXDL : public BasePopulateParams<InitParamsXDL> {
 private:
-  static constexpr size_t nInitParameters = 9;
+  static constexpr size_t nInitParameters = 15;
   // Initial tuning parameters for forward convolution and backward
   // convolution.
   static const InitParamsXDL initParameters[nInitParameters];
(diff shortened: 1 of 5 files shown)


giuseros:
This fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/799

krzysz00 (commented):
As a sanity check, are these in order? That is, the tiebreaker for configs that use the same padding is their order in the heuristic list, so we want the better ones to come first.

krzysz00 (approved):
I don't see any reason not to merge this, assuming we have perf numbers

giuseros:
Hi @krzysz00 , 
With a single set of default tuning params is hard to improve all the performance. I split our table in three fp32/fp16/int8 . Fp16 settings will be used also for bf16 (there is a swtich on the bit width). 

With this, default perf jumped by 25% (there are some regression, but it's hard making everybody happy). 

[default.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10910246/default.zip)

krzysz00:
I'd say this is a good argument for merging this

I'd be interested - if you have the time to investigate this - in what the old vs. new perf configs were on the cases where we had that 10% regression.

giuseros:
So there are three configs that got worse
1. `-t f16  -transB true -transA true  -g 64 -m 1024 -k 4096 -n 384`   
     `{128,128,4,64,64,4,1,1} -> {128,128,4,64,64,8,1,1}`
2. `-t f32 --transA --transB  -g 1 -m 2 -k 768 -n 16384` 
    `{4,64,16,4,64,1,1,1,} -> {32,64,4,32,64,4,1,1,}`
3. `-t i8 --transB  -g 1024 -m 384 -k 64 -n 384` 
    `{64,64,8,32,32,8,1,1,} -> {128,128,8,64,64,8,1,1}`

Unfortunately for cases 1 and 3 there isn't much I can do. If make those cases use the original config, then also many other cases will use the same config, and perf will get worse. 

For case 2, I added back the `{32,64,4,32,64,4,1,1,}` as the last config, and it seems that it went fine. New performance are attached: [default2.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10920488/default2.zip)

PR #986, opened 2023-02-22
Fix premerge-checks to lint against develop

manupak:
This commit fixes the base commit being the commit just before 1 to develop branch. In that way, the whole PR is linted.

### DIFF
--- a/mlir/utils/jenkins/static-checks/premerge-checks.py
+++ b/mlir/utils/jenkins/static-checks/premerge-checks.py
@@ -17,21 +17,21 @@ Example usage:
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 import os
+import sys
 import re
 import subprocess
 from typing import Tuple, Optional
 import pathspec
 import unidiff
+import argparse
 
 
 def get_diff(base_commit) -> Tuple[bool, str]:
-  r = subprocess.run(f'/opt/rocm/llvm/bin/git-clang-format {base_commit}', shell=True)
-  if r.returncode != 0:
-    r = subprocess.run(f'git checkout -- .', stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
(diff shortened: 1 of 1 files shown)


mlir/utils/jenkins/static-checks/premerge-checks.py
```diff
@@ -17,21 +17,21 @@
...
 
 
 def get_diff(base_commit) -> Tuple[bool, str]:
-  r = subprocess.run(f'/opt/rocm/llvm/bin/git-clang-format {base_commit}', shell=True)
-  if r.returncode != 0:
-    r = subprocess.run(f'git checkout -- .', stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
+  diff_run = subprocess.run(f'/opt/rocm/llvm/bin/git-clang-format --diff {base_commit}', shell=True)
+  if diff_run.returncode != 0:
     return False, ''
-  diff_run = subprocess.run(f'git diff -U0 --no-prefix --exit-code', stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
```
krzysz00, line 32: I think the point of this diff call is to show the changes that come up when you run clang-format. We might want to keep it, but it's not urgent given our workflow.

krzysz00 (approved):
Approved

mlir/utils/jenkins/static-checks/premerge-checks.py
manupak: Thats spelled git clang-format --diff and its still maintainined in the same way and one does not need to do the changes and revert

manupak: (See my new diff_run var)

krzysz00: Ah, cool. in which case, we're good

PR #987, opened 2023-02-23
[CI] Ignore third-party files in clang-tidy

manupak:
This commit adds a filter to filter out clang-tidy warning originate outside our repo.

Summary of changes : 
* We had been using opt/rocm/bin/git-clang-format with clang-format-10 and this PR changes that to be opt/rocm/bin/git-clang-format with opt/rocm/bin/clang-format (aka clang-format-15'ish)
* Adds a third-party files ignore pattern to clang-tidy checks
* Adds a switch enable/disable linting of external/ folder from Jenkins -- only to be used with upstream merges.

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -121,7 +121,12 @@ void preMergeCheck(String codepath) {
           ln -s build/compile_commands.json compile_commands.json
         fi
         '''
-        sh 'python3 ./mlir/utils/jenkins/static-checks/premerge-checks.py'
+        if (params.ignoreExternalLinting == true) {
+            sh 'python3 ./mlir/utils/jenkins/static-checks/premerge-checks.py --ignore-external'
+        }
+        else {
+            sh 'python3 ./mlir/utils/jenkins/static-checks/premerge-checks.py'
+        }
     } else {
         echo "Static Test step skipped"
     }
@@ -363,6 +368,10 @@ pipeline {
         // option to disable navi21 cells in case nodes are offline
         booleanParam(name: 'disableNavi21', defaultValue: false,
                      description: 'Disable testing on Navi21')
+
+        // option only to be used with upstream merges
+        booleanParam(name: 'ignoreExternalLinting', defaultValue: false,
+                     description: 'Ignore linting on files in external/')
     }
     stages {
         stage("Set System Property") {
--- a/mlir/utils/jenkins/static-checks/premerge-checks.py
+++ b/mlir/utils/jenkins/static-checks/premerge-checks.py
@@ -39,7 +39,11 @@ def get_diff(base_commit) -> Tuple[bool, str]:
   print(diff)
   return diff
 
-def run_clang_format(base_commit, ignore_config):
+def check_external_file(filename: str) -> bool:
+  regex = f'^external/'
+  return re.search(regex, filename)
+
+def run_clang_format(base_commit, ignore_config, ignore_external_files: bool = False):
   """Apply clang-format and return if no issues were found.
   Extracted from https://github.com/google/llvm-premerge-checks/blob/master/scripts/clang_format_report.py"""
 
@@ -58,6 +62,9 @@ def run_clang_format(base_commit, ignore_config):
     patched_file_tgt = patched_file.target_file[2:]
     if ignore.match_file(patched_file_src) or ignore.match_file(patched_file_tgt):
       continue
+    if ignore_external_files:
+      if check_external_file(patched_file_src) or check_external_file(patched_file_tgt):
+        continue
     hunk: unidiff.Hunk
     for hunk in patched_file:
       success = False
@@ -90,7 +97,7 @@ def check_third_party_file(filename: str) -> bool:
   return not re.search(regex, filename)
 
 
-def run_clang_tidy(base_commit, ignore_config): 
+def run_clang_tidy(base_commit, ignore_config, ignore_external_files: bool = False):
   """Apply clang-tidy and return if no issues were found.
   Extracted from https://github.com/google/llvm-premerge-checks/blob/master/scripts/clang_tidy_report.py"""
 
@@ -129,8 +136,12 @@ def run_clang_tidy(base_commit, ignore_config):
           print('{} is ignored by pattern and no comment will be added'.format(file_name))
           continue
         if check_third_party_file(file_name):
-          print('{} is ignored as its a third-party file and no comment will be added'.format(file_name))
+          print(f'{file_name} is ignored as its a third-party file and no comment will be added')
           continue
+        if ignore_external_files:
+          if check_external_file(file_name):
+            print(f'{file_name} is ignored as its a external (upstream) file and no comment will be added')
+            continue
         if severity == 'warning':
           warn_count += 1
         if severity == 'error':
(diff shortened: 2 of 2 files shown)


jerryyin (approved):
Look good. Did you manually tested this with @jungpark-mlir 's upstream sync PR? I think that's a good use case for this script.

giuseros (approved):
Nice! LGTM!

manupak:
> Look good. Did you manually tested this with @jungpark-mlir 's upstream sync PR? I think that's a good use case for this script.

@jerryyin good suggestion!

I attempted this. So to get the linters to run : 
I needed to add  : 
```
/opt/rocm/llvm/bin/git-clang-format --binary /opt/rocm/llvm/bin/clang-format
```

Now it 'runs' successfully but I see required formatting changes all over in external/ subtree update PR.

Having thought about a bit, I feel its meaningless to format external/. I mean only argument for format external/ is that we want to make sure the downstream changes to external/ are properly formatted but the way auto formatters work it is not exactly additive (i.e. it can also affect unrelated thing if it captured in the hunk).

Though I must admit, I was bit puzzled at certain changes that seemed to be impossible to have been caused by us.  
E.g. : 

```
diff --git a/external/llvm-project/llvm/lib/Support/regcomp.c b/external/llvm-project/llvm/lib/Support/regcomp.c
index 06aeb93b6a90..2c70c3bba726 100644
--- a/external/llvm-project/llvm/lib/Support/regcomp.c
+++ b/external/llvm-project/llvm/lib/Support/regcomp.c
@@ -527,8 +527,9 @@ p_ere_exp(struct parse *p)
                        assert(p->pbegin[backrefnum] != 0);
                         assert(OP(p->strip[p->pbegin[backrefnum]]) == OLPAREN);
                         assert(OP(p->strip[p->pend[backrefnum]]) == ORPAREN);
-                        (void) dupl(p, p->pbegin[backrefnum]+1, p->pend[backrefnum]);
-                       EMIT(O_BACK, backrefnum);
+                        (void)dupl(p, p->pbegin[backrefnum] + 1,
+                                   p->pend[backrefnum]);
+                        EMIT(O_BACK, backrefnum);
```

I am biasing towards the conclusion that we should omit external/ in the 'rocMLIR' linter because ultimately all external/ changes will be linted and landed upsteam in the way upstream want it (well! ideally :) ).

I ll leave it you guys to decide. Signing out for today!

jerryyin:
@manupak Thanks for trying it out!

> Now it 'runs' successfully but I see required formatting changes all over in external/ subtree update PR.

I think part of the reason is that we used the rocm alternative of clang-format but not the upstream version? I was assuming that we had always been using the upstream clang-format so that it can give a consistent result between us and the external folder. But turns out that my assumption is wrong. (I'm still leaning towards using the similar toolchain with upstream, that now come as a separate topic).

> I am biasing towards the conclusion that we should omit external/ in the 'rocMLIR'

I agree with you that doing the way it is, it is meaningless to go back and re-format the external folder according to our version of clang-format and its rules. Have you already made the update to drop the external folder comparison? Once its done this PR is ready to land.

krzysz00:
I'm going to disagree with this PR except maybe as a hotfix

We should be using whichever `clang-format` upstream runs on its code, however we make that happen, and if we make a change to external, it should pass `clang-format` checks.

krzysz00:
And I'll note that until @manupak's recent RP, we didn't use to use the `git-clang-format` from `/opt/rocm`, we used to use the one in `$PATH`, which was the `clang-10` one.

krzysz00:
So perhaps the fix is to revert to `/usr/bin/git-clang-format` instead

jerryyin:
> we used to use the one in $PATH, which was the clang-10 one.

@krzysz00 Okay thanks for the reminder, we are on the same page to be consistent with upstream.

krzysz00:
Speaking of, @manupak , why'd you change the clang-format executable?

manupak:
@krzysz00 if you mean this https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/986/files#diff-758e3adf6a842db25fe9444a5a8536c8e044842daf657c53b6ef990156d1758aL28

I didn't change git-clang-format executable

manupak:
> And I'll note that until @manupak's recent RP, we didn't use to use the `git-clang-format` from `/opt/rocm`, we used to use the one in `$PATH`, which was the `clang-10` one.

Im a bit surprised by this...

krzysz00:
Well, I was wrong then

Would you be willing to limit the `external/` to file that aren't in `external/llvm-project/mlir`, since that's a directory we do change and do want to keep an eye on?

manupak:
> I'm going to disagree with this PR except maybe as a hotfix
> 
> We should be using whichever `clang-format` upstream runs on its code, however we make that happen, and if we make a change to external, it should pass `clang-format` checks.

All this PR changes is to ignore third party clang tidy warnings and make opt/rocm/clang-format with opt/rocm/git-clang-format (the latter is not a change of this or previous PR).

Whilst I agree, to have identical clang-format to upstream, we currently run opt/rocm/git-clang-format with clang-format-10 that is not only inconsistent with each other but also with upstream.

We can leave the clang-format invocation at that if ya'll think its better (unchanged by this or previous PR) but I think ignoring 3rd party clang tidy warnings should be done, IMHO.

krzysz00:
Ok, hold on, before and after your previous PR (and this PR, if it changes anything), which actual `clang-format` binary was being used? The one in `/opt/rocm/llvm/bin` or the one in `/usr/bin`?

krzysz00:
But yeah, ignoring third-party clang-tidy is fine, I'd say.

krzysz00:
I thought it was ignoring clang-format for external/ as well

manupak:
Before the PRs, git-clang-format defaults git default of clangFormat.binary which runs the system default (i.e. on the container) whicj clang-format-10.

Previous PR didn't change that behavior.

This PR changes the binary to be rocm/bin/clang-format (a.k.a clang-format-15'ish)

manupak:
clang-format-10 is not going to work with upstream merge, so I went with the best we have in the container: rocm/bin/clang-format (clang-format-15) but thats also doesn't work exactly (runs but suggests changes to upstreamed code) -- maybe due to its not what exactly upstream uses?

So I can keep the clang-format invocation the same for this PR and deal with it in the upstream merge thats is where this will be a problem OR another PR to update our image to match upstream clang-format.

WDYT?

manupak:
I ll also note a possible reason of us trying to lint upstream code 'might' not work:

Upstream also updates the lint config throughout the history. What we are trying to do is take two snapshots (between subtree merges) and try to lint everything based on the linter version + config of the last snapshot. However, upstream merged those commits based on the linter+lint config available when it was merged.

This may or may not be the case but we can only know by trying the upstream linter on the subtree update PR. I feel we ll have to come to a decision if it turns out to be the case. Happy to take a call tomorrow if you guys feel like it.

krzysz00:
Looks like upstream uses, assuming https://github.com/google/llvm-premerge-checks/blob/main/containers/buildkite-premerge-debian/Dockerfile is the correct file, clang-format-15

So, long-term, we should probably update our Dockerfile to bump to clang-format-15 as released.

For now, we can go with /opt/rocm/llvm/bin/clang-format

Now, when you say changes to the upstream code, does that mean changes to code that we've modified, or changes to things we're merging in that we didn't touch? If it's the second one, then we should go ahead and land this "ignore external/" bit.

krzysz00:
... @manupak Yeah, I can see it. Code that got committed with config 1 passed lint, config 2 would have failed it, but lint checks only run on code that's changed. So excluding external/ from clang-format may be the right call.

Or, more meaningfully, providing a way to force clang-format to ignore external/ for upstreme merges in particular, so that when we commit code to `external/`, it gets linted, but that'd be a followup PR

manupak:
> Or, more meaningfully, providing a way to force clang-format to ignore external/ for upstreme merges in particular, so that when we commit code to `external/`, it gets linted, but that'd be a followup PR

@krzysz00 ok, I've added an option all the way to jenkins to switch this -- so its upto the PR owner. PTAL.

manupak:
(Also to reduce confusion to others, I've added a summary at the top)

krzysz00 (approved):

krzysz00:
Ok, now I understand the PR and that it's doing a few different things, all of which makes sense. Approved.

manupak:
Thanks folks for reviews! 
hopefully should aid in upstream merge and future PRs

PR #988, opened 2023-02-27
Update global load/LDS store thread grouping to improve coalescing

krzysz00:
Per @giuseros, grouping threads of a workgroup into a D x K group unconditionally causes failures to coalesce when vectorizing in the D dimension. Further consideration revealed that this grouping is not required: the only requirement is that the threads be grouped in the same way for global load and LDS store.

This commit fixes this unnecessary specificity.

A future followup will update the tiebreak logic to account for what GEMM dimension has actual unit stride.


giuseros:
Hi @krzysz00 , I had the same exact PR ready to be pushed, but was waiting because I prefer to merge first #983  and then test this in isolation on top of that. 

Also there are other things I wanted to investigate about how we use vectorization logic before merging this change (connected to few CK investigations). Do you mind closing this so that I can carry on with my investigations? Thanks!

PR #989, opened 2023-02-28
Adding a simple script to parse the MIOpen tuning DB

giuseros:
This is a script I used in #983 to parse the MIOpen tuning DB and run configurations with `rocmlir-gen`. 

All the script does, for now, is to print out a series of `rocmlir-gen` commands with the appropriate convolution + perf configuration (parsed from the MIOpen tuning db). 

In the future, we could extend `./perfRunner` to use this script and run conv workloads without going through MIOpen.

### DIFF
--- a/mlir/utils/performance/parseMIOpenTuningDB.py
+++ b/mlir/utils/performance/parseMIOpenTuningDB.py
@@ -1,4 +1,5 @@
 import argparse
+import getopt
 import sqlite3
 import sys
 
@@ -7,6 +8,7 @@ MLIR_FILTER_LAYOUTS = {"NCHW": "kcyx", "NHWC": "kyxc"}
 MLIR_OUTPUT_LAYOUTS = {"NCHW": "nkhw", "NHWC": "nhwk"}
 MLIR_OPS = {"F": "conv2d", "W": "conv2d_bwd_weight", "B": "conv2d_bwd_data"}
 MLIR_DATA_TYPE = {"FP16": "f16", "FP32": "f32", "INT8INT8INT32": "i8"}
+
 
 # Convert a key,value pair from the tuning DB to a "--option value" parameter for rocmlir-gen
 def convertToRocmgenParam(key, val):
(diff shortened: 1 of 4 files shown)


mlir/utils/performance/parseMIOpenTuningDB.py
```diff
@@ -0,0 +1,113 @@
...
+        description="This scripts parse the MIOpen performance DB  and generates rocmlir-gen calls",
+    )
+
+    parser.add_argument("--tuning-db", type=str, help="Path to the tuning db")
+    parser.add_argument(
+        "--print-tables",
+        action="store_true",
+        help="Print the tables in the tuning database",
+    )
+    parser.add_argument("--arch", type=str, help="Architecture we tuned for")
```
krzysz00: Do we want to make this a required positional argument? Or to read it out of the database? Or ... I'm not unhappy with this, just suggesting alternatives

krzysz00 (commented):
Can we go ahead and `JOIN` on the perf_config table?

(I'll note, that the `-1` perf configs mean "use heuristic mode")

mlir/utils/performance/parseMIOpenTuningDB.py
```diff
@@ -0,0 +1,113 @@
...
+
+def selectAllTables(cursor):
+    sql_query_get_all_tables = """SELECT name FROM sqlite_master WHERE type='table';"""
+    cursor.execute(sql_query_get_all_tables)
+    tables = cursor.fetchall()
+    return tables
+
+
+def selectAllConfigs(cursor):
+    sql_query_print_config = """SELECT * FROM config"""
```
jerryyin: This will seriously limit the applicability of this script and make a normal db run into performance issues.

mlir/utils/performance/parseMIOpenTuningDB.py
```diff
@@ -0,0 +1,113 @@
...
+    cursor.execute(sql_query_print_config)
+    configs = cursor.fetchall()
+    keys = [description[0] for description in cursor.description]
+    return configs, keys
+
+
+def selectPerfConfig(cursor, perf_config_id):
+    sql_query_select_perf_conifg = (
+        f"""SELECT * FROM perf_db WHERE config={perf_config_id}"""
```
jerryyin: Same as above.

jerryyin (changes requested):
I see what you are trying to do with the script, and it make sense if it is strictly in this context that you populate rocmlir command for every entry in the db. 

However, I think another way to do it will make the script much more useful:
 - Single entry mode:  it poke the perfdb and try to `SELECT * FROM perf_db INNER JOIN config ON perf_db.config = config.id WHERE ((in_channels = 64) AND ...)`
 - Multi-entry mode: It read in the config from the third party text file, like the ones where we defined resnet50 configs, line by line.
   - Then for each line, it use the single entry mode to grab their perf config string and populate the rocmlir command 

By doing it this way, we are allowed to load perf config from perfdb regardless of how many entries the db has. Rather, it allows us to query arbitrary list of config from the config list.

mlir/utils/performance/parseMIOpenTuningDB.py
giuseros: You mean reading from the database path? I was not able to find the arch anywhere else

krzysz00: I'd've thought it might be in the table somewhere with the rest of the conv config.

But we could get it out of the path, since those are standardized.

Or pass it in.

What I was suggesting is that, if it's going to be an argument, make the script be `./parseMIOpenDB.py [arch] [path]` so you can't leave `[arch]` off by accident.

krzysz00:
@jerryyin I'd suggest that, even if we're doing a perf config lookup, we may want to print out the whole `rocmlir-gen` command.

giuseros:
Hi @jerryyin , I can take that approach, the only issue is here:

> Single entry mode: it poke the perfdb and try to SELECT * FROM perf_db INNER JOIN config ON perf_db.config = config.id WHERE ((in_channels = 64) AND ...)

In our txt file a lot of parameters are missing like layout, data type and direction of the convolution. So this will (I believe) return multiple entries in the database. Should I print them all?

jerryyin:
@krzysz00 Agreed

@giuseros Good point, I think when fields like layout/type are missing, you can:
 - supply them as arguments to the script
 - When arguments are empty, populate a default type/layout

I think we rarely want to do every comparison all at once. Unless we do, then it'd be fairly easy to extend this script with a "printAll" option later.

krzysz00:
@jerryyin On the other hand, I think it will be useful for analysis to have a "dump the MIOpen perf DB in a form that's acceptable input to rocmlir-gen" option.

jerryyin:
@krzysz00 That should be equivalent to supplying the resnet50 configs to this script. (And the only use case for this scenario is that this db must be populated only by our CI's tuning.)

If you are talking about a local test scenario to compare a limited subset of configs, it should be fairly trivial to manually write a config file. Since you will have to run them anyways.

krzysz00:
You can manually run the CI tuning commands locally to get that same perf DB, and having a one-click way to extract the perf DB from such a local run and from CI so you can `diff` them is good

Passing in the resnet50 configs would probably do it.

giuseros:
Hi @jerryyin , @krzysz00 ,
I made few changes that should make everyone happy :) 
- We can now pass a `--miopen_config` or `--miopen_configs_file`
- We can force a given direction/layout/datatype
- For incomplete configurations we select all the records in the DB that apply

In this way:

- `python3 ./bin/parseMIOpenTuningDB.py --tuning-db=../MIOpen/build/MIOpenUserDB/gfx90a6e_1.1.0.udb` parses all the tuning DB
- `python3 ./bin/parseMIOpenTuningDB.py --tuning-db=../MIOpen/build/MIOpenUserDB/gfx90a6e_1.1.0.udb --miopen_config="-n 256 -c 1024..."` parses the given config (if it is incomplete, more records will be parsed)
- `python3 ./bin/parseMIOpenTuningDB.py --tuning-db=../MIOpen/build/MIOpenUserDB/gfx90a6e_1.1.0.udb --miopen_configs_file="-n 256 -c 1024..."` parses the given config file (if the records are incomplete, more records will be parsed)

So if we want to parse a single config, it needs to be specific, so for instance:
``` 
$ python3 ./bin/parseMIOpenTuningDB.py --tuning-db=../MIOpen/build/MIOpenUserDB/gfx90a6e_1.1.0.udb --miopen_config="-n 256 -c 1024 -H 14 -W 14 -k 2048 -y 1 -x 1 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -t 1" --layout=NHWC --direction F --data_type=FP32` 
```
generates
```
rocmlir-gen -ph --arch gfx90a --perf_config=128,128,4,64,64,4,1,1 --in_layout nhwc --out_layout nhwk --fil_layout kyxc  -t f32 --operation conv2d --in_channels 1024 --in_h 14 --in_w 14 --fil_h 1 --fil_w 1 --out_channels 2048 --batchsize 256 --padding_h 0 --padding_w 0 --conv_stride_h 2 --conv_stride_w 2 --dilation_h 1 --dilation_w 1
```

jerryyin (approved):
Good job!

Last non blocking nits:
 - This script should be moved under `mlir/utils/widgets`. We put all convenient script not directly invoked by CI to there so things get uncluttered.
 - You should split up the MIOpen to MLIR command translation into its own python script, invokable by the tuning db parsing script. Parsing a MIOpen config and convert to MLIR corresponding config has long been a pain point and you have actually addressed it together with db query functionality. If you can split it up then more audience will have a chance to use your contributions here.

krzysz00:
I'll note that the MIOpen->MLIR translation comes up a bunch in the CI scripts and could maybe be factored out from there, or the infrastructure from there could be used.

giuseros:
Re reusing the MIOpen->MLIR translation, I thought about it, but the issue is that I am not doing MIOpen -> MLIR. 

What I am doing is MIOpen -> TuningDB and then TuningDB -> MLIR. 

It is true that the loop's structure in  MIOpen -> TuningDB is basically copied from `perfRunner`, but the content of the loop is different from f the one in `perfRunner`, and I struggled to see how to unify them

PR #990, opened 2023-02-28
Adding migraphx quantize linear op and its TOSA lowering

jerryyin:
I added the quantize linear op lowering implemented according to most recent meeting with MIGraphX and CK.

```cpp
__host__ __device__ constexpr void
operator()(int8_t& y, const int32_t& x, const int32_t& bias) const
{
    float y_fp32 = ck::type_convert<float>(x + bias);
    y_fp32 = math::clamp(round(requantScale_ * y_fp32), -128.f, 127.f);
    y      = ck::type_convert<int8_t>(y_fp32);
}

```

This sample implementation match perfectly with the original prototype:
 - The int32 to fp32 type cast is aligned with tosa.cast
 - The fp32 to int32 type cast includes: rounding, clamp and type convert, also aligned with tosa.cast

### DIFF
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -18,6 +18,7 @@
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/Dialect/Tosa/IR/TosaOps.h"
+#include "mlir/Dialect/Tosa/Utils/QuantUtils.h"
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Transforms/DialectConversion.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
@@ -54,6 +55,19 @@ static TosaOp createOpAndInfer(mlir::PatternRewriter &rewriter,
   Type newOutTy = RankedTensorType::get({returnShape[0].getDims()}, elemType);
   auto result = op->getResult(0);
   result.setType(newOutTy);
+  return op;
+}
+
+static tosa::CastOp createCastOp(mlir::PatternRewriter &rewriter,
+                                 mlir::Location loc, Type resElementType,
+                                 Value input) {
+  ShapedType inputType = input.getType().cast<ShapedType>();
+  auto elementType = inputType.getElementType();
+  Type resType = inputType.cloneWith({}, resElementType);
+
+  auto op = rewriter.create<tosa::CastOp>(loc, elementType, input);
+  auto result = op->getResult(0);
+  result.setType(resType);
   return op;
 }
 
@@ -149,6 +163,12 @@ public:
       cop->setAttr("xdlopsV2", attr);
     if (auto attr = op->getAttrOfType<StringAttr>("perf_config"))
       cop->setAttr("perf_config", attr);
+
+    if (elementTy.isInteger(8)) {
+      auto quantAttr = rewriter.getAttr<tosa::ConvOpQuantizationAttr>(
+          /*inputZp =*/0, /*weightZp =*/0);
+      cop->setAttr("quantization_info", quantAttr);
+    }
 
     // transpose the output back to NCHW so that it can match following
     // operators.
@@ -478,11 +498,47 @@ public:
   }
 };
 
+class QuantizeLinearConverter final
+    : public OpConversionPattern<migraphx::QuantizeLinearOp> {
+public:
+  using OpConversionPattern<migraphx::QuantizeLinearOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(migraphx::QuantizeLinearOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const final {
+    auto input = op.getOperand(0);
+    auto scale = op.getOperand(1);
+    ShapedType inputType = input.getType().cast<ShapedType>();
+    auto elementType = inputType.getElementType();
+    Location loc = op->getLoc();
+
+    assert(!elementType.isInteger(8) &&
+           "quantlinear op Only supporting int8 elementType now");
+
+    Value shifted = input;
+    if (op.getNumOperands() == 3) {
+      auto bias = op.getOperand(2);
+      shifted = createOpAndInfer<tosa::AddOp>(rewriter, loc, elementType, input,
+                                              bias);
+    }
+
+    Type intermediateElementType = rewriter.getF32Type();
+    Value upCast =
+        createCastOp(rewriter, loc, intermediateElementType, shifted);
+    Value scaled = createOpAndInfer<tosa::MulOp>(
+        rewriter, loc, intermediateElementType, upCast, scale, /*shift=*/0);
+
+    Type destElementType = rewriter.getIntegerType(8);
+    Value downCast = createCastOp(rewriter, loc, destElementType, scaled);
+    rewriter.replaceOp(op, {downCast});
+    return success();
+  }
+};
 } // namespace
 
 void migraphx::populateMIGraphXToTosaConversionPatterns(
     MLIRContext *context, RewritePatternSet &patterns) {
   patterns.add<ConvConverter, BroadcastConverter, MultiBroadcastConverter,
                ReshapeConverter, SoftmaxConverter, DotConverter,
-               ReduceMeanConverter>(context);
+               ReduceMeanConverter, QuantizeLinearConverter>(context);
 }
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -41,7 +41,7 @@ static bool isBroadcastable(Operation *op, Operation *operand) {
 
 template <typename TosaOp, typename... Args>
 static TosaOp createOpAndInfer(PatternRewriter &rewriter, Location loc,
-                               Type elemType, Args &&... args) {
+                               Type elemType, Args &&...args) {
   auto op =
       rewriter.create<TosaOp>(loc, UnrankedTensorType::get(elemType), args...);
   InferShapedTypeOpInterface shapeInterface =
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -515,11 +515,6 @@ public:
     auto elementType = inputType.getElementType();
     Location loc = op->getLoc();
 
-    if (!elementType.isInteger(32)) {
-      return rewriter.notifyMatchFailure(
-          loc, "quantlinear op Only supporting int32 elementType now");
-    }
-
     Value shifted = input;
     if (auto bias = op.getBias()) {
       shifted = createOpAndInfer<tosa::AddOp>(rewriter, loc, elementType, input,
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -60,12 +60,9 @@ static TosaOp createOpAndInfer(PatternRewriter &rewriter, Location loc,
 static tosa::CastOp createCastOp(PatternRewriter &rewriter, Location loc,
                                  Type resElementType, Value input) {
   ShapedType inputType = input.getType().cast<ShapedType>();
-  auto elementType = inputType.getElementType();
   Type resType = inputType.cloneWith({}, resElementType);
 
-  auto op = rewriter.create<tosa::CastOp>(loc, elementType, input);
-  auto result = op->getResult(0);
-  result.setType(resType);
+  auto op = rewriter.create<tosa::CastOp>(loc, resType, input);
   return op;
 }
 
@@ -162,6 +159,14 @@ public:
     if (auto attr = op->getAttrOfType<StringAttr>("perf_config"))
       cop->setAttr("perf_config", attr);
 
+    // Note: For TOSA convolution, a non-float type is considered as a
+    // quantized convolution. For quantized convolution, it is required
+    // to carry the "quantization_info" as attribute. Adding this
+    // attribute help us populate the correct TOSA IR.
+    //
+    // When we add support to quantized types and TOSA.rescale Op, we
+    // should make the quantized attribute to accept actual zero point
+    // values from intput and filter.
     if (elementTy.isInteger(8)) {
       auto quantAttr = rewriter.getAttr<tosa::ConvOpQuantizationAttr>(
           /*inputZp =*/0, /*weightZp =*/0);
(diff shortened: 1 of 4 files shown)


mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -150,6 +164,12 @@ class ConvConverter final
     if (auto attr = op->getAttrOfType<StringAttr>("perf_config"))
       cop->setAttr("perf_config", attr);
 
+    if (elementTy.isInteger(8)) {
+      auto quantAttr = rewriter.getAttr<tosa::ConvOpQuantizationAttr>(
+          /*inputZp =*/0, /*weightZp =*/0);
+      cop->setAttr("quantization_info", quantAttr);
+    }
+
```
jerryyin, line 175: FYI: I ended up figure out that this attribute is the minimal change I need to make int8 convolution valid IR in TOSA. It required an attribute to tell what zero point for intput and weight is. Since we don't do any quantization related compute in middle of the convolution, add the attribute with hardcoded zero as zero point to comply with TOSA standard.

krzysz00: Could we handle things by meaningfully setting this attribute and then generating the `linalg.generic`s in `TosaToRock`? Or would that, for example, only work for uniform quantizations?

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -57,6 +58,19 @@ static TosaOp createOpAndInfer(mlir::PatternRewriter &rewriter,
   return op;
 }
 
+static tosa::CastOp createCastOp(mlir::PatternRewriter &rewriter,
```
krzysz00: I know it became house style in this file somehow, but can we lose the leading `mlir::`s here?

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -478,11 +498,47 @@ class ReduceMeanConverter final
...
 
+class QuantizeLinearConverter final
+    : public OpConversionPattern<migraphx::QuantizeLinearOp> {
+public:
+  using OpConversionPattern<migraphx::QuantizeLinearOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(migraphx::QuantizeLinearOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const final {
+    auto input = op.getOperand(0);
```
krzysz00: Nit: use actual names, not integer indices. Those are way too error-prone

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -478,11 +498,47 @@ class ReduceMeanConverter final
...
+    ShapedType inputType = input.getType().cast<ShapedType>();
+    auto elementType = inputType.getElementType();
+    Location loc = op->getLoc();
+
+    assert(!elementType.isInteger(8) &&
+           "quantlinear op Only supporting int8 elementType now");
+
+    Value shifted = input;
+    if (op.getNumOperands() == 3) {
+      auto bias = op.getOperand(2);
```
krzysz00: Probably better written as `getBias()` and matching on the optional

krzysz00 (commented):
Style nitpicks!

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jerryyin: @krzysz00 Maybe, but only where a `TOSA.rescale` is already provided. That was the intended use case of quantization provided by TOSA. (So that tosa.rescale can decide is parameters according to input/weight quantization factors)

In the meeting with MIGraphX, they explicitly mention that will not give us any information related with intput/weight tensor and only pass us with output's quantized information. So this philosophy goes against TOSA. But since the required semantic was against `TOSA.rescale` anyways, we have little work-around but to match the behavior with what MIGraphX has asked for.

krzysz00: Got it, seems reasonable to me

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jerryyin: Good advice!

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jerryyin: Yes, that's very reasonable.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jerryyin: Agreed, I think my memory of operand usage still stays a couple years back when the only way to do it is to reference the index. Thanks for reminding me about it.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -57,6 +57,18 @@ static TosaOp createOpAndInfer(mlir::PatternRewriter &rewriter,
   return op;
 }
 
+static tosa::CastOp createCastOp(PatternRewriter &rewriter, Location loc,
+                                 Type resElementType, Value input) {
+  ShapedType inputType = input.getType().cast<ShapedType>();
```
krzysz00, line 62: Minor question: does this need the `cast`? `getInput()` may well be a `TypedValue<ShapedType>` already.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -57,6 +57,18 @@ static TosaOp createOpAndInfer(mlir::PatternRewriter &rewriter,
...
+  auto elementType = inputType.getElementType();
+  Type resType = inputType.cloneWith({}, resElementType);
+
+  auto op = rewriter.create<tosa::CastOp>(loc, elementType, input);
```
krzysz00: Is there a form of this builder that takes the result type? The fact that you're calling `setType()` feels off.

krzysz00: (but maybe that's just how Tosa stuff goes)

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
krzysz00: Maybe add a comment explaining what and why the workaround is?

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -478,11 +496,46 @@ class ReduceMeanConverter final
...
+    : public OpConversionPattern<migraphx::QuantizeLinearOp> {
+public:
+  using OpConversionPattern<migraphx::QuantizeLinearOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(migraphx::QuantizeLinearOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const final {
+    auto input = op.getInput();
+    auto scale = op.getScale();
+    ShapedType inputType = input.getType().cast<ShapedType>();
```
krzysz00, line 514: Same note re `cast`

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -478,11 +496,46 @@ class ReduceMeanConverter final
...
+    auto elementType = inputType.getElementType();
+    Location loc = op->getLoc();
+
+    assert(!elementType.isInteger(8) &&
```
krzysz00: Instead of an assert(), `notifyMatchFailure()`. Passing in bad input is not assertion-worthy unless such input is 100% our fault.

krzysz00 (approved):
Approved pending the nits (most notably the `assert()`)

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jerryyin: Unfortunately `input` is a Value, `getType()` will only return the most generic `mlir::Type`. There has to be a type cast there to grab the element type.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jerryyin: You caught my bug, nice! I was trying to match with the generate op creation function in a few lines above. It is only after this comment that I realize the CastOp's creation needs to take the result type, not input element type.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jerryyin: It's probably 100% our fault (or migraphx fault) to ask a type beyond what they tell us to. But I agree assert is probably too strong for this type of issue.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jerryyin: Done in [8aed021](https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/990/commits/8aed02156d5720249f58330a42c5e7fc8a8c919b)

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jerryyin: Done in 8aed021

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jerryyin: Done in 8aed021

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -478,11 +501,48 @@ class ReduceMeanConverter final
...
+  LogicalResult
+  matchAndRewrite(migraphx::QuantizeLinearOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const final {
+    auto input = op.getInput();
+    auto scale = op.getScale();
+    ShapedType inputType = input.getType().cast<ShapedType>();
+    auto elementType = inputType.getElementType();
+    Location loc = op->getLoc();
+
+    if (!elementType.isInteger(32)) {
```
manupak: nit : alternatively we could make this illegal by migraphx op definition -- so that they will get caught migraphx verifiers rather than in the conversion process. It does not matter here so much as there is no higher level dialect but if that wasn't the case it would be useful as it get reflect in docs and other places.

Again its a nit, feel free to ignore if you want.

E.g. : 
https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/bb18479bf625e6d36b9eb0496a7a5ed061044e5e/mlir/include/mlir/Dialect/Rock/IR/RockOps.td#L181-L182

manupak (changes requested):
There might be an issue in lowering quantize_linear from migraphx.

So the current pass pipeline of migraphx looks as follows : 

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/a63ee2e0a81aa3a59ff65d7aa67b522993977274/src/targets/gpu/target.cpp#L103-L161

So it does the quantization canonicalization at their end currently before mlir-based fusion passes are called and it does something like : 

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/a63ee2e0a81aa3a59ff65d7aa67b522993977274/src/rewrite_quantization.cpp#L35-L70.

So based on the current way the pipeline is organized, I fear we may never see a quantizelinear at our end.

Apologies If I wasn't clear before, this is what I meant in the question : "who owns the quant canonicalization ?"  

If we really want to handle the quant canonicalization , there are few options I could see : 
A1. we could put a gate on the migraphx pass on rewrite_quantization if MLIR is enabled.
A2. we could re-order the migraphx passes.
A3. we just let them deal with canonicalization and lower the ops that they overwrite on quantize_linear.

I think for A1 and A2 choices we need to evaluate perf and functional impacts.
Out of the two, A1 might be prefferable.

Im just putting a hold here until we know a resolution because it might mean we dont need to support the lowering of quantize_linear but might need something like migraphx.round (A3)

jerryyin:
> So based on the current way the pipeline is organized, I fear we may never see a quantizelinear at our end.
> A1. we could put a gate on the migraphx pass on rewrite_quantization if MLIR is enabled.
> A2. we could re-order the migraphx passes.
> A3. we just let them deal with canonicalization and lower the ops that they overwrite on quantize_linear.

@manupak Thanks for reminding me about your concerns, I have considered about those before coming up with the quantize linear op. I've got a couple of discussions with @sjw36 and Alan Tuner on MIGraphX end. The conclusion of next step is more or less aligned with what you have. Appreciate you for taking the additional proof-read on migraphX end!

 - For A1, if we choose with the current implementation then it means to macro out the `rewrite_quantization` pass. I mentioned the possibility of this to Alan and he is okay with it.
 - For A2, I'm hesitant to do this because I don't have enough background on existing pass orderings
 - For A3, It seems to be the more natural way to go but there are more mismatches in base ops compared to TOSA. The details please see https://github.com/ROCmSoftwarePlatform/rocMLIR/discussions/981 (Sorry should have send it to you before this code review)

The purpose of this step is to get everything going and tested before migraphX and CK properly sort out theirs. This allows me to start before that meeting with CK and MIGraphX and start to roll out the implementation as soon as possible. (A3, for example, has a implicit dependency on how migraphx decides to implement the kernel. They may drag some additional op or data type in that's not as what I had planned in my proposal.) To give you the full picture of my plan. For the next round of TODOs:
 - End to end testing from migraphX dialect to kernel
 - Come up with the PR on migraphX end to implement A1. (I will have to coordinate with you if you plan to make mlir compilation a runtime switch and get rid of the macro)
 - Come back to take a second look at A3, but that means we'll have to implement maintain a chunk of Op that is somewhat similar yet duplicated compared to what TOSA has. So this step only act as a backup should MIGraphX change their mind and reject my PR.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jerryyin: Good point! will do

manupak (approved):
Ok, if A1 is direction of travel, this PR make sense! (removing the hold).

Just something to be wary of :

I'd also note that not running rewrite_quantization might have an implication/change on how we want fuse_pointwise (that is also run before fuse_mlir on MIGraphX end) because currently we rely on that capture all the pointwise ops that follows a gemm/conv2d. Thus, not running rewrite_quantization might not fuse all the ops (breaking where quantize_linear appears ?). This affect the fuse_mlir pass as it would only offload immediate fused pointwise block following the gemm/conv2d -- if that makes sense.

manupak:
(We might want to fuse_pointwise to consider quantize_linear as a pointwise op too because I think right now that pass never sees a quantize_linear -- I might be wrong)

jerryyin:
> (We might want to fuse_pointwise to consider quantize_linear as a pointwise op too because I think right now that pass never sees a quantize_linear -- I might be wrong). because currently we rely on that capture all the pointwise ops that follows a gemm/conv2d.

Good point and thanks for the pointer, just become aware of `fuse_pointwise` pass in your comments. It sounds very reasonable that we'd need to capture it to make it fus-able.

PR #991, opened 2023-02-28
Make threadwise_write_all a writer in regularization

krzysz00:
In the tuning loop, and probably in the MIOpen context, the applicability section of the kernel pipeline (GridwiseGemmToBlockwise and before) may be run multiple times. The first run stops at GridwiseGemmToBlockwise, and errors at or before this point are deemed to be the fault of the tuning config.

Then, when the compilation pipeline is run, the steps that were run as part of the applicability test are run again. What this means is that Regularize may receive as its input the result of
GridwiseGemmToBlockwise. Since that IR was already reguralized, the pass should be a noop.

However, because threadwise_write_all wasn't on the list of writers, the regularizer would treat it as a reader and go looking for another writer, not find one, and leave the IR in some very weird state. This patch fixes the problem, allowing us to tune kernels with fusions in them.

Thanks to @yiqian1  for substantial initial debugging effort.


sjw36 (approved):
Good find!

PR #992, opened 2023-03-02
Adds multiple output fusion support

manupak:
This commit adds support selectively clone
threadswise write all depending on whether
the intermediaries are copied to output buffers
( > 1).

I also needed to change AsyncGraph pass as it creates
awaits for each output although they could align
to a single token (i.e multiple outputs from a single kernel)

I also needed to change rocmlir-gen test harness creation
as it assumed that there will be only one output in the 
kernel.

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/803

### DIFF
--- a/external/llvm-project/mlir/lib/Dialect/XModel/Transforms/AsyncGraph.cpp
+++ b/external/llvm-project/mlir/lib/Dialect/XModel/Transforms/AsyncGraph.cpp
@@ -73,8 +73,10 @@ class XModelAsyncGraphPass
     // Insert host sync before operation that reads an async::value
     for (auto operand : op->getOperands()) {
       if (auto itoken = res2tokens.lookup(operand)) {
-        createWaitOp(op, itoken);
-        currentTokens.erase(itoken);
+        if (currentTokens.contains(itoken)) {
+          createWaitOp(op, itoken);
+          currentTokens.erase(itoken);
+        }
       }
     }
     // Insert host synchronization before terminator or op with side effects.
--- a/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
@@ -91,16 +91,8 @@ static void moveTransformsBefore(PatternRewriter &b, Value val) {
   }
 }
 
-static Value transformAccordingly(PatternRewriter &b, Value nOut, Value refOp) {
-  if (auto rtop = dyn_cast<TransformOp>(refOp.getDefiningOp())) {
-    // 0. recurse and apply from the begining of the transform chain
-    nOut = transformAccordingly(b, nOut, rtop.getOperand());
-
-    // 1. apply identical transforms to other inputs
-    nOut = b.create<TransformOp>(rtop.getLoc(), nOut, rtop.getTransform());
-  }
-
-  return nOut;
+static SmallVector<Attribute> extractVector(ArrayAttr arrayAttr) {
+  return llvm::to_vector<4>(arrayAttr.getAsRange<Attribute>());
 }
 
 Value makeRegs(PatternRewriter &b, MemRefType::Builder &mrb, Location loc,
@@ -111,8 +103,9 @@ Value makeRegs(PatternRewriter &b, MemRefType::Builder &mrb, Location loc,
                                        srcMemType.getElementType())));
 }
 
-static Value applyTransforms(PatternRewriter &b, ThreadwiseWriteAllOp storeOp,
-                             Value src) {
+static Value
+applyTransforms(PatternRewriter &b, ThreadwiseWriteAllOp storeOp, Value src,
+                const SmallVector<TransformMapAttr> &relativeViewsOnStore) {
   // 0. capture the memref containing the outputs being written
   Location loc = storeOp.getLoc();
   Value gemmOut = storeOp.getSource();
@@ -129,57 +122,127 @@ static Value applyTransforms(PatternRewriter &b, ThreadwiseWriteAllOp storeOp,
   moveTransformsBefore(b, src);
 
   // 2.1. apply transform chain from output
-  src = transformAccordingly(b, src, storeOp.getDest());
+  auto extraViews = extractVector(storeOp.getExtraViews());
+  extraViews.insert(extraViews.end(), relativeViewsOnStore.begin(),
+                    relativeViewsOnStore.end());
 
   // 2.2. load into registers
-  b.create<ThreadwiseReadIntoOp>(loc, src, alloc, storeOp.getExtraViews(),
+  b.create<ThreadwiseReadIntoOp>(loc, src, alloc, b.getArrayAttr(extraViews),
                                  storeOp.getForceUnroll(),
                                  storeOp.getUseIndexDiffs());
   return alloc;
 }
 
-static Operation *traceToNonViewOp(Operation *op) {
+static void traceToNonViewUsers(Operation *op,
+                                SmallVectorImpl<Operation *> &nonViewOps) {
   if (auto transform = dyn_cast<TransformOp>(op)) {
     Value result = transform.getResult();
-    // TODO(sjw): fix when divergence is encountered
-    assert(result.hasOneUse());
     for (auto &use : result.getUses()) {
-      return traceToNonViewOp(use.getOwner());
+      traceToNonViewUsers(use.getOwner(), nonViewOps);
     }
+  } else {
+    nonViewOps.push_back(op);
+  }
+}
+
+static Operation *traceToNonViewDef(Operation *op,
+                                    SmallVectorImpl<TransformMapAttr> &views,
+                                    Operation *terminator) {
+  if (op == terminator) {
+    return op;
+  }
+  if (auto transform = dyn_cast<TransformOp>(op)) {
+    views.push_back(transform.getTransformAttr());
+    Operation *nonViewDef = traceToNonViewDef(
+        transform.getViewSource().getDefiningOp(), views, terminator);
+    return nonViewDef;
   }
   return op;
 }
 
-static ThreadwiseWriteAllOp traceToThreadwiseWrite(Value inp) {
+static Operation *traceToNonViewDef(Operation *op) {
+  if (auto transform = dyn_cast<TransformOp>(op)) {
+    return traceToNonViewDef(transform.getViewSource().getDefiningOp());
+  } else {
+    return op;
+  }
+}
+
+// This function checks the results of the current operation
+// is copied to a block argument output, hence needs to be
+// preserved.
+static bool isLeadingToBlockArg(Operation *op) {
+  Operation *nonViewDef = traceToNonViewDef(op);
+  for (Value result : nonViewDef->getResults()) {
+    for (Operation *use : result.getUsers()) {
+      SmallVector<Operation *, 4> nonViewUsers;
+      traceToNonViewUsers(use, nonViewUsers);
+      for (Operation *nonViewUser : nonViewUsers) {
+        if (memref::CopyOp memCopy = dyn_cast<memref::CopyOp>(nonViewUser)) {
+          if (memCopy.getTarget().isa<BlockArgument>()) {
+            return true;
+          }
+        }
+      }
+    }
+  }
+  return false;
+}
+
+static ThreadwiseWriteAllOp traceToThreadwiseWrite(
+    Value inp, SmallVectorImpl<TransformMapAttr> &inpToTWWriteAllViews) {
   // 1. Validate that the only uses of the linalg.generic input are the one
   // generic and a copy operation or transform.
   ThreadwiseWriteAllOp result;
   for (Operation *use : inp.getUsers()) {
-    use = traceToNonViewOp(use);
-    if (isa<memref::DeallocOp>(use)) {
-      // ignore
-      continue;
-    }
-    if (auto lgop = dyn_cast<linalg::GenericOp>(use)) {
-      // reader
-      if (!llvm::is_contained(lgop.getInputs(), inp)) {
+    SmallVector<Operation *, 4> nonViewUsers;
+    traceToNonViewUsers(use, nonViewUsers);
+    for (Operation *nonViewUse : nonViewUsers) {
+      if (isa<memref::DeallocOp>(nonViewUse)) {
+        // ignore
+        continue;
+      }
+      if (auto lgop = dyn_cast<linalg::GenericOp>(nonViewUse)) {
+        // We dont check for linalg.generic readers
+        // here because they could both be reading and writing
+        // to intermediary buffer that could be copied as an
+        // additional kernel output.
+      } else if (auto reduceOp = dyn_cast<ReduceOp>(nonViewUse)) {
+        // reader
+        if (inp != reduceOp.getIn()) {
+          return ThreadwiseWriteAllOp();
+        }
+      } else if (auto memcpy = dyn_cast<memref::CopyOp>(nonViewUse)) {
+        // reader
+        if (memcpy.getOperand(0) != inp) {
+          return ThreadwiseWriteAllOp();
+        }
+      } else if (auto store = dyn_cast<ThreadwiseWriteAllOp>(nonViewUse)) {
+        // Threadwise copy that is already unttransformed (new style)
+        if (result) {
+          LLVM_DEBUG(llvm::dbgs() << "Multiple global stores somehow\n");
+          // It is possible to contain multiple threadwise write all ops
+          // in the presense of multiple outputs in the kernel. When tracing
+          // we just pick the latest ThreadwiseWriteAllOp.
+          if (result->isBeforeInBlock(store)) {
+            SmallVector<TransformMapAttr> views_;
+            traceToNonViewDef(store.getDest().getDefiningOp(), views_,
+                              inp.getDefiningOp());
+            result = store;
+            inpToTWWriteAllViews.insert(inpToTWWriteAllViews.begin(),
+                                        views_.begin(), views_.end());
+          }
+        } else {
+          SmallVector<TransformMapAttr> views_;
+          traceToNonViewDef(store.getDest().getDefiningOp(), views_,
+                            inp.getDefiningOp());
+          result = store;
+          inpToTWWriteAllViews.insert(inpToTWWriteAllViews.begin(),
+                                      views_.begin(), views_.end());
+        }
+      } else {
         return ThreadwiseWriteAllOp();
       }
-    } else if (auto memcpy = dyn_cast<memref::CopyOp>(use)) {
-      // reader
-      if (memcpy.getOperand(0) != inp) {
-        return ThreadwiseWriteAllOp();
-      }
-    } else if (auto store = dyn_cast<ThreadwiseWriteAllOp>(use)) {
-      // Threadwise copy that is already unttransformed (new style)
-      if (result) {
-        LLVM_DEBUG(llvm::dbgs()
-                   << "Multiple global stores somehow, no fusion\n");
-        return ThreadwiseWriteAllOp();
-      }
-      result = store;
-    } else {
-      return ThreadwiseWriteAllOp();
     }
   }
 
@@ -189,20 +252,21 @@ static ThreadwiseWriteAllOp traceToThreadwiseWrite(Value inp) {
 }
 
 // Returns the value of the buffer that's meant to be the new writeback.
-static void reconfigureLAGeneric(PatternRewriter &b,
-                                 linalg::GenericOp laGeneric, Value inRegs,
-                                 Value outRegs,
-                                 ThreadwiseWriteAllOp twWriteOp) {
+static void reconfigureLAGeneric(
+    PatternRewriter &b, linalg::GenericOp laGeneric, Value inRegs,
+    Value outRegs, ThreadwiseWriteAllOp twWriteOp,
+    const SmallVector<TransformMapAttr> &relativeViewsOnStore) {
   SmallVector<Value, 5> newInputs;
   SmallVector<AffineMap, 5> lgAMaps;
 
   for (auto inp : laGeneric.getInputs()) {
     Value newInput;
-    if (traceToThreadwiseWrite(inp)) {
+    SmallVector<TransformMapAttr> views_;
+    if (traceToThreadwiseWrite(inp, views_)) {
       newInput = inRegs;
     } else {
       // 2.1.1. Align tiling of other inputs
-      newInput = applyTransforms(b, twWriteOp, inp);
+      newInput = applyTransforms(b, twWriteOp, inp, relativeViewsOnStore);
     }
     newInputs.push_back(newInput);
 
@@ -227,14 +291,14 @@ static void reconfigureLAGeneric(PatternRewriter &b,
   laGeneric.setIteratorTypesAttr(ArrayAttr::get(ctx, iteratorTypes));
 }
 
-static Value findThreadwiseWrite(linalg::GenericOp laGeneric,
-                                 ThreadwiseWriteAllOp &twWriteOp) {
+static Value
+findThreadwiseWrite(linalg::GenericOp laGeneric,
+                    ThreadwiseWriteAllOp &twWriteOp,
+                    SmallVector<TransformMapAttr> &laInToOutViews) {
   for (auto input : laGeneric.getInputs()) {
-    if (auto allocOp = input.getDefiningOp<memref::AllocOp>()) {
-      if (auto twop = traceToThreadwiseWrite(input)) {
-        twWriteOp = twop;
-        return input;
-      }
+    if (auto twop = traceToThreadwiseWrite(input, laInToOutViews)) {
+      twWriteOp = twop;
+      return input;
     }
   }
 
@@ -269,8 +333,9 @@ LAGenericRewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
   // 1. Trace input to global_store.
   // 1.1. Find the (implicit) gemm output
   ThreadwiseWriteAllOp gemmStoreOp;
+  SmallVector<TransformMapAttr> laInToOutViews;
   Value laGenericInputLeadingToGemmStore =
-      findThreadwiseWrite(laGeneric, gemmStoreOp);
+      findThreadwiseWrite(laGeneric, gemmStoreOp, laInToOutViews);
   if (!laGenericInputLeadingToGemmStore)
     return failure();
 
@@ -289,6 +354,11 @@ LAGenericRewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
   if (!gemmStoreOp) {
     LLVM_DEBUG(llvm::dbgs() << "Align tiling: couldn't find writeback\n");
     return failure();
+  }
+
+  if (isLeadingToBlockArg(gemmStoreOp.getDest().getDefiningOp())) {
+    gemmStoreOp =
+        static_cast<ThreadwiseWriteAllOp>(b.clone(*gemmStoreOp.getOperation()));
   }
 
   // 2. Apply if input found
@@ -312,18 +382,21 @@ LAGenericRewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
   MemRefType::Builder mrb(gemmOutType);
   auto laOutRegs = makeRegs(b, mrb, loc, out.getType());
   // 2.2. Tile linalg.generic with vgpr as input, return output vgprs
-  reconfigureLAGeneric(b, laGeneric, fusionRegs, laOutRegs, gemmStoreOp);
+  reconfigureLAGeneric(b, laGeneric, fusionRegs, laOutRegs, gemmStoreOp,
+                       laInToOutViews);
   // 2.2.0. Move the generic before the write-back. This'll put all
   // the copy loops for other inputs before the generic due to insertion
   // order.
-  laGeneric->moveBefore(gemmStoreOp);
+  gemmStoreOp->moveAfter(laGeneric);
 
   gemmOut.replaceAllUsesWith(fusionRegs);
 
   // 2.3. Replace rock.threadwise_write_all inputs with la.generic result vgprs
   gemmStoreOp.getSourceMutable().assign(laOutRegs);
-
-  out = transformAccordingly(b, out, gemmStoreOp.getDest());
+  auto extraViews = extractVector(gemmStoreOp.getExtraViews());
+  extraViews.insert(extraViews.end(), laInToOutViews.begin(),
+                    laInToOutViews.end());
+  gemmStoreOp.setExtraViewsAttr(b.getArrayAttr(extraViews));
   gemmStoreOp.getDestMutable().assign(out);
 
   if (auto outAlloc = out.getDefiningOp<memref::AllocOp>()) {
@@ -336,28 +409,36 @@ LAGenericRewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
 LogicalResult MemcpyRewritePattern::matchAndRewrite(memref::CopyOp copy,
                                                     PatternRewriter &b) const {
   auto src = copy.getSource();
-  auto trg = copy.getTarget();
+  auto target = copy.getTarget();
 
   Operation *gemmStoreOp = nullptr;
+  SmallVector<TransformMapAttr> views;
   if (auto allocOp = src.getDefiningOp<memref::AllocOp>()) {
-    if (auto twop = traceToThreadwiseWrite(src)) {
+    if (auto twop = traceToThreadwiseWrite(src, views)) {
       gemmStoreOp = twop;
     }
   }
 
-  if (gemmStoreOp && isa<ThreadwiseWriteAllOp>(gemmStoreOp)) {
-    PatternRewriter::InsertionGuard guard(b);
-    b.setInsertionPoint(gemmStoreOp);
+  if (gemmStoreOp) {
+    if (ThreadwiseWriteAllOp twWriteAllOp =
+            dyn_cast<ThreadwiseWriteAllOp>(gemmStoreOp)) {
+      PatternRewriter::InsertionGuard guard(b);
+      b.setInsertionPoint(twWriteAllOp);
 
-    // 1. replace memref.copy with rock.threadwise_write_all
-    trg = transformAccordingly(b, trg, gemmStoreOp->getOperand(1));
-    gemmStoreOp->setOperand(1, trg);
+      // 1. replace memref.copy with rock.threadwise_write_all
+      auto extraViews = extractVector(twWriteAllOp.getExtraViews());
+      extraViews.insert(extraViews.end(), views.begin(), views.end());
+      twWriteAllOp.setExtraViewsAttr(b.getArrayAttr(extraViews));
+      twWriteAllOp.getDestMutable().assign(target);
+      // twWriteAllOp->moveAfter(target.getDefiningOp());
 
-    if (auto outAlloc = trg.getDefiningOp<memref::AllocOp>())
-      outAlloc->moveBefore(gemmStoreOp);
+      if (auto outAlloc = target.getDefiningOp<memref::AllocOp>()) {
+        outAlloc->moveBefore(gemmStoreOp);
+      }
 
-    b.eraseOp(copy);
-    return success();
+      b.eraseOp(copy);
+      return success();
+    }
   }
 
   return failure();
--- a/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
@@ -32,7 +32,9 @@
 #include "mlir/Dialect/Utils/ReshapeOpsUtils.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
 #include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/Dominance.h"
 #include "mlir/IR/PatternMatch.h"
+#include "mlir/Interfaces/SideEffectInterfaces.h"
 #include "mlir/Pass/PassManager.h"
 #include "mlir/Transforms/DialectConversion.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
@@ -147,6 +149,42 @@ static void traceToNonViewUsers(Operation *op,
   }
 }
 
+// This function traces to non view readers.
+// TODO(@manupak): we should really be implementing
+// MemoryEffectOpInterface to ops to make this simpler
+// to identify readers.
+static LogicalResult
+traceToNonViewReaders(Operation *op, Value parentVal,
+                      SmallVectorImpl<Operation *> &nonViewReaders) {
+  if (auto transform = dyn_cast<TransformOp>(op)) {
+    Value result = transform.getResult();
+    for (auto &use : result.getUses()) {
+      return traceToNonViewReaders(use.getOwner(), result, nonViewReaders);
+    }
+  } else {
+    if (ThreadwiseWriteAllOp twWriteAllOp =
+            dyn_cast<ThreadwiseWriteAllOp>(op)) {
+      if (twWriteAllOp.getSource() == parentVal) {
+        nonViewReaders.push_back(op);
+      }
+    } else if (MemoryEffectOpInterface memEffectOp =
+                   dyn_cast<MemoryEffectOpInterface>(op)) {
+      if (hasEffect<MemoryEffects::Read>(memEffectOp, parentVal)) {
+        nonViewReaders.push_back(op);
+      }
+    } else if (CopyOpInterface copyOp = dyn_cast<CopyOpInterface>(op)) {
+      if (copyOp.getSource() == parentVal) {
+        nonViewReaders.push_back(op);
+      }
+    } else {
+      return op->emitError() << "Found an unsupported operator that needs to "
+                                "be added reader checks \n"
+                             << op;
+    }
+  }
+  return success();
+}
+
 static Operation *traceToNonViewDef(Operation *op,
                                     SmallVectorImpl<TransformMapAttr> &views,
                                     Operation *terminator) {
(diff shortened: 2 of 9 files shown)


external/llvm-project/mlir/lib/Dialect/XModel/Transforms/AsyncGraph.cpp
```diff
@@ -73,8 +73,10 @@ class XModelAsyncGraphPass
     // Insert host sync before operation that reads an async::value
     for (auto operand : op->getOperands()) {
       if (auto itoken = res2tokens.lookup(operand)) {
-        createWaitOp(op, itoken);
-        currentTokens.erase(itoken);
+        if (currentTokens.contains(itoken)) {
```
krzysz00, line 76: What's this change for?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -111,8 +103,9 @@ Value makeRegs(PatternRewriter &b, MemRefType::Builder &mrb, Location loc,
                                        srcMemType.getElementType())));
 }
 
-static Value applyTransforms(PatternRewriter &b, ThreadwiseWriteAllOp storeOp,
-                             Value src) {
+static Value
+applyTransforms(PatternRewriter &b, ThreadwiseWriteAllOp storeOp, Value src,
+                const SmallVector<TransformMapAttr> &relativeViewsOnStore) {
```
krzysz00: `const SmallVector<T> &` is spelled `ArrayRef<T>`

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -129,57 +122,127 @@ static Value applyTransforms(PatternRewriter &b, ThreadwiseWriteAllOp storeOp,
...
 
   // 2.1. apply transform chain from output
-  src = transformAccordingly(b, src, storeOp.getDest());
+  auto extraViews = extractVector(storeOp.getExtraViews());
+  extraViews.insert(extraViews.end(), relativeViewsOnStore.begin(),
+                    relativeViewsOnStore.end());
 
   // 2.2. load into registers
-  b.create<ThreadwiseReadIntoOp>(loc, src, alloc, storeOp.getExtraViews(),
+  b.create<ThreadwiseReadIntoOp>(loc, src, alloc, b.getArrayAttr(extraViews),
```
krzysz00: Can we not do this? `extraViews` should be the TID->GEMM maps? But it's not _wrong_.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -129,57 +122,127 @@ static Value applyTransforms(PatternRewriter &b, ThreadwiseWriteAllOp storeOp,
...
+        // to intermediary buffer that could be copied as an
+        // additional kernel output.
+      } else if (auto reduceOp = dyn_cast<ReduceOp>(nonViewUse)) {
+        // reader
+        if (inp != reduceOp.getIn()) {
+          return ThreadwiseWriteAllOp();
+        }
+      } else if (auto memcpy = dyn_cast<memref::CopyOp>(nonViewUse)) {
+        // reader
+        if (memcpy.getOperand(0) != inp) {
```
krzysz00: Isn't this `getSource()` or something?

krzysz00 (commented):
Overall thoughts: would it be possible to not use `extraViews` as expansively as you are doing?

I'll have to read the changes more carefully, but because you also went for refactoring into `extraViews` it gets hard to tell.

external/llvm-project/mlir/lib/Dialect/XModel/Transforms/AsyncGraph.cpp
manupak: Because there seemed to be an assumption token --> output operand to be 1:1. 
w/o this it will end up creating two awaits on the same token.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: ah yes... my bad

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: yes we could... I found it easier to read the IR this way. Any downside ? 
(I just needed this to speed up my debugging while developing this... but can use TransformOps if that effort is worthwhile spending..)

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Yea.... I would use that -- _well_ but its not a change in this PR... :)

krzysz00: Might as well clean it up while you're here.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
krzysz00: I think it's easier to conceptually understand when there's a sequence of `rock.transform` calls? 

This is more because of the theory that we're always aligning the writes out to GEMM space, so `extraViews` is a minimal thing that `gridwise_gemm_to_blockwise` uses to translate from `{bid, tid, i)` to GEMM coordinates, and then the transforms on the value take you the rest of the way there.

It's not a hard requirement, but ... I kinda like it the way we had it before.

external/llvm-project/mlir/lib/Dialect/XModel/Transforms/AsyncGraph.cpp
krzysz00: Ah

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: I ll have a go tomorrow to see but we might wanna codify an op verifier if thats desired to reduce ambiguity in the future

manupak: (Though I still think its easier to read it left to right compared to interleaved transform chains in fusion scenarios -- so its subjective)

manupak:
OK, I've changed from appending extraViews to transforming the output. @krzysz00 PTAL when you have some time...

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: now done!

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: done!

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -189,20 +254,21 @@ static ThreadwiseWriteAllOp traceToThreadwiseWrite(Value inp) {
...
 
 // Returns the value of the buffer that's meant to be the new writeback.
-static void reconfigureLAGeneric(PatternRewriter &b,
-                                 linalg::GenericOp laGeneric, Value inRegs,
-                                 Value outRegs,
-                                 ThreadwiseWriteAllOp twWriteOp) {
+static void reconfigureLAGeneric(
+    PatternRewriter &b, linalg::GenericOp laGeneric, Value inRegs,
+    Value outRegs, ThreadwiseWriteAllOp twWriteOp,
+    const SmallVector<TransformMapAttr> &relativeViewsOnStore) {
```
krzysz00: Same minor note re ArrayRef

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -189,20 +254,21 @@ static ThreadwiseWriteAllOp traceToThreadwiseWrite(Value inp) {
...
   SmallVector<Value, 5> newInputs;
   SmallVector<AffineMap, 5> lgAMaps;
 
   for (auto inp : laGeneric.getInputs()) {
     Value newInput;
-    if (traceToThreadwiseWrite(inp)) {
+    SmallVector<TransformMapAttr> views_;
```
krzysz00: Why the trailing underscore?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -291,6 +358,11 @@ LAGenericRewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
     return failure();
   }
 
+  if (isLeadingToBlockArg(gemmStoreOp.getDest().getDefiningOp())) {
```
krzysz00: I think this needs some comments or an example (or renames). From what I can tell, this code will always clone the `threadwise_write_all`

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -138,48 +135,116 @@ static Value applyTransforms(PatternRewriter &b, ThreadwiseWriteAllOp storeOp,
...
+    return op;
+  }
+}
+
+// This function checks the results of the current operation
+// is copied to a block argument output, hence needs to be
+// preserved.
+static bool isLeadingToBlockArg(Operation *op) {
+  Operation *nonViewDef = traceToNonViewDef(op);
+  for (Value result : nonViewDef->getResults()) {
```
krzysz00, line 220: Might as well preemptively add the case where `result` is itself a `BlockArgument`?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -138,48 +135,116 @@ static Value applyTransforms(PatternRewriter &b, ThreadwiseWriteAllOp storeOp,
...
+        if (inp != reduceOp.getIn()) {
+          return ThreadwiseWriteAllOp();
+        }
+      } else if (auto memcpy = dyn_cast<memref::CopyOp>(nonViewUse)) {
+        // reader
+        if (memcpy.getSource() != inp) {
+          return ThreadwiseWriteAllOp();
+        }
+      } else if (auto store = dyn_cast<ThreadwiseWriteAllOp>(nonViewUse)) {
+        // Threadwise copy that is already unttransformed (new style)
```
krzysz00, line 270: What's this mean?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -138,48 +135,116 @@ static Value applyTransforms(PatternRewriter &b, ThreadwiseWriteAllOp storeOp,
...
+        if (result) {
+          LLVM_DEBUG(llvm::dbgs() << "Multiple global stores somehow\n");
+          // It is possible to contain multiple threadwise write all ops
+          // in the presense of multiple outputs in the kernel. When tracing
+          // we just pick the latest ThreadwiseWriteAllOp.
```
krzysz00: Is this sound? I'm not quite convinced this is correct.

krzysz00 (commented):
I'm a bit unsure about whether these changes are correct

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Im not sure either... it was there.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Sure -- but dont we have copies always in fusion cases ? now that we dont have copy-opt

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Ack

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: ah need to change that.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: > From what I can tell, this code will always clone the threadwise_write_all

Why would you think that ? 

It is meant clone if the gemmStoreOp dest is leading to output buffer -- which is not always case. In the single output fusion case, prior to aligntiling the output of gemmStoreOp dest is a memref that is an input to linalg block. Hence, we dont need to clone that case because it is not a real output of the kernel.

However, if that output was mem copied to an output buffer, then we need to clone (instead of modifying the gemmStoreOp).

Do you want me to put a description on the code ?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
krzysz00: Yeah but suppose, for example, that I've decided to omit that trailing copy, either because someone actually wrote a general copy opt in the future or because, for example, I'm manually crafting the `la.generic`s myself

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: If what you are saying is true -- we'd have three threadwise_write_all ops (1 original and 2 for two linalg blocks)

krzysz00: Ok, wait, hold on, I thought `gemmStoreOp` was the `threadwise_write_all`

krzysz00: Because I thought we had an infinite loop here: find the store op, clone it, find that, clone that, and so on

manupak: (I ll add unit tests once we agree on the changes)

krzysz00: The thing is, the threadwise_write_all gets modified to point at the generic output after fusion. So I could see this code leading to having extra threadwise_write_all's

manupak: > The thing is, the threadwise_write_all gets modified to point at the generic output after fusion. So I could see this code leading to having extra threadwise_write_all's 

No it get modifed to point to internal memref alloca that get copy optimized later if it is actually an output

manupak: (These internal memref allocas are useful as they are bit SSA'ish)

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: done!

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: Having a second look, we shouldn't encounter such....
It was only because we were cloning a bit early. fixed that now.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
krzysz00: "the generic output" means the output of the linalg.generic. And then the rewrite for copies come is and the threadwise_write_all gets updated again to the output of the memref.copy, and so on

krzysz00: I'm basically wanting to make sure this code is clear on future reading. Currently, it wasn't. Maybe it's variable naming.

manupak: > Because I thought we had an infinite loop here: find the store op, clone it, find that, clone that, and so on

In any case, it wouldn't be an infinite loop; it would overclone max per linalg block (per rewrite).

But to your point, yes Im relying on the internal alloc s and their copies distinguish which are outputs and which are internal based on the fact what gets copied out to output memrefs because it is the state of IR we receive today. 

I do acknowledge the point if a better bufferization pass will eliminate those internal allocas -- we might need to bit more analysis to identify outputs vs internal states but IMO we should not run that until later -- if possible.

krzysz00: And, I'll note, that in the single argument case, the output leads to a block argument too? And so it would be cloned?

manupak: In the single output case, the linalg block will have two memref : input and output.

The rock op will be writing to the input memref of linalg (which is not copied, hence not cloned)
(Prior to the rewrite of linalg -- I mean)

manupak: (I've added a description now -- PTAL)

manupak: I ll also add unit test as promised before, once we agree on the changes -- which would also make this clearer....

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -389,8 +378,16 @@ LAGenericRewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
   // 2.2.0. Move the generic before the write-back. This'll put all
   // the copy loops for other inputs before the generic due to insertion
   // order.
-  gemmStoreOp->moveAfter(laGeneric);
 
+  // Before being copy-optimized the output of prior op (could be gemm-like rock
+  // op or linalg block) will be written to the internal memref alloca. If that
+  // is not copied as an output, it is not an actual output. Therefore, we dont
+  // need clone the gemmStore.
+  if (isLeadingToBlockArg(gemmStoreOp.getDest().getDefiningOp())) {
```
krzysz00: Ok, I see what's going on here, now that you've restructured things. Can I ask for a different name for this function? Perhaps `leadsToUnwrittenBlockArg`

krzysz00 (commented):
I'm happy with the changes overall, and with the most recent commit, I understand what's going on here and can reduce the confusion to "can we get a different name for that function"?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: done!

krzysz00 (commented):
Ok, I'm still very confused by the logic, especially since, having been renamed, the unwritten block argument check doesn't seem to do what the name suggests.

I don't understand why the clone() isn't triggered in the case where there's only one output.

I would have expected the logic to be that, if there is no path down the writer/reader chains from the current fusable thing's output to an output that appears within a threadwise_write_all, you need to create a new threadwise_write_all .

That is, if I have 
```
%t1 = alloc()
%t2 = alloc()
%t3 = alloc()
threadwise_write_all %c -> f(%t1)
la.generic %t1 -> g(%t2)
la.generic %t2 -> h(%t3)
memref.copy %t3 -> %argN
```
this should not create extra copies, because you can chose the writer/reader chain from %t2 to the %t3 that shows up in the copies.

However, for 
```
%t1 = alloc()
%t2 = alloc()
%t3 = alloc()
threadwise_write_all %c -> f(%t1)
la.generic %t1 -> g(%t2)
la.generic %t1 -> h(%t3)
memref.copy %t2 -> %argM
memref.copy %t3 -> %argN
```
you do need to clone an extra writer.

I'm confused about what algorithm you're trying to implement for determining when to clone.

manupak:
> That is, if I have
> 
> ```
> %t1 = alloc()
> %t2 = alloc()
> %t3 = alloc()
> threadwise_write_all %c -> f(%t1)
> la.generic %t1 -> g(%t2)
> la.generic %t2 -> h(%t3)
> memref.copy %t3 -> %argN
> ```

Here when rewrite la.generic 1, it checks threadwise_write_all %c -> f(%t1) whether t1 is copied out.  its not it will not clone and modify to be threadwise_write_all %cc -> f(%t2).

Then when we rewrite  la.generic 2, threadwise_write_all %cc -> f(%t2) and t2 is not copied either -->  it will not clone and modify to be threadwise_write_all %ccc -> f(%t3).

no la.generics are left -- only 1 threadwise_write_all remains.

> However, for
> 
> ```
> %t1 = alloc()
> %t2 = alloc()
> %t3 = alloc()
> threadwise_write_all %c -> f(%t1)
> la.generic %t1 -> g(%t2)
> la.generic %t1 -> h(%t3)
> memref.copy %t2 -> %argM
> memref.copy %t3 -> %argN
> ```

Here when rewrite la.generic 1, it checks threadwise_write_all %c -> f(%t1) whether t1 is copied out.  its not it will not clone and modify to be threadwise_write_all %cc -> f(%t2).

Then when we rewrite  la.generic 2, threadwise_write_all %cc -> f(%t2) and IS copied  -->  it WILL clone to be threadwise_write_all %ccc -> f(%t3).

no la.generics are left -- 2 threadwise_write_all s remains.

krzysz00:
Ok, but I still don't get how the checks work. The arguments to the check for if you need to clone is the threadwise_write_all's destination ...

krzysz00:
Furthermore, how about

```
%t1 = alloc()
%t2 = alloc()
%t3 = alloc()
la.generic %t1 -> %t2
la.generic %t2 -> %t3
memref.copy %t2, %argM
memref.copy %t3, %argN
```

krzysz00:
Basically, the way this code is written makes it really hard for me to convince myself that it's correct.

krzysz00:
That may or may not be a function of the actual logic, it *could* be naming.

manupak:
I dont know how to make it more simple.. 
(The only logic there is to trace from the input to linalg block to writer of that memref that is also a tw_write_all and check whether that input is being copied out ? )

In each rewrite, it looks at linalg's input and to which tw_write_all it traces to. When the rewrite progresses  tw_write_all destination gets updated -- cloned or not.

In both your scenarios, it just need to check whether the dest of tw_write_all ( = is also input memref of linalg being rewritten) it traces to is being copied out.

> Furthermore, how about
> 
> ```
> %t1 = alloc()
> %t2 = alloc()
> %t3 = alloc()
> la.generic %t1 -> %t2
> la.generic %t2 -> %t3
> memref.copy %t2, %argM
> memref.copy %t3, %argN
> ```

What is the fusion here ? just pure linalg fusion ? I'd expect an original tw_write_all to be present for gemm-like op

krzysz00:
Re my last testcase, Yeah, I left off the write_all to %t1, my bad

The way you explained it doesn't make sense. _Every_ input is being copied out eventually.

How does the following algorithm sound:
- In `RockRegularize`, if a writer has multiple readers - excluding copies directly into a temporary buffer (to prevent recursion) - add temporary buffers and `memref.copy` ops to distinguish the uses, leaving one of them alone.
- During fusion, when you have multiple readers from the output of what you just fused (where all but one will be a `memref.copy` to a temporary), clone the threadwise_write_all for each additional reader.
  - Alternatively, you could phrase this as "the fusion rule for `memref.copy` is to clone the `threadwise_write_all`, unless the output of the copy is a block argument) ... yeah, that'd be simpler

manupak:
I don't find that more simpler than this approach...
If not basically the same thing with just more rework. 

I m struggling to understand why is it so complex to follow logic : the input memref of linalg being rewritten, if its copied out (hence needs to be preserved), then tw_write_all needs to be cloned.

Why?

because the tw_write_all is doing the right thing as long as input memref is concerned and we need another tw_write_all to write to the output memref of current linalg. 

Then we progressively keep rewriting the linalgs..

krzysz00:
> if it's being copied out
_All_ memrefs are being copied out.

krzysz00:
So what you're saying is that you always need to clone the threadwise_write_all

sjw36:
Is there a reason why we cannot just make memref.copy convert to tw_write_all somehow? I know that would be a major rewrite but now we create a tw_write_all that writes to a local memref, then we have to move it with all these gymnastics. Not relevant for this PR.

manupak:
I'd prefer this approach because it can constructively rewrite the linalg blocks producing correct IR between rewrites.

All memrefs are not copied out. I don't see that happening, you can see that if you look at IR before linalg align. Local memrefs will be there between boundaries of linalg blocks.

If you guys think this is wrong, please provide me with a counter example, I m more than happy to add that as a test case.

@sjw36 in effect thats what is being done -- except memref copy is eliminated in a different rewrite. We could merge memcpy rewrite pattern to linalg patern -- then itd be as you say.

krzysz00:
Ok, so, the thing is, when I say "copied out" and see functions that trace to a block argument, I'm considering the full tree here - eventually, everything will reach a write to a block argument.

I don't understand what, in your current solution, prevents the logic for "this result has multiple readers" from being applied in the case where there is only one reader. Can you point me to what about your code ensures your changes have no effect for the one-output case?

And @sjw36, that was my proposal, where we regularize the tree to add copies to local temporaries and have the fusion rule for memref.copy clone the threadwise_write_all when those are encountered.

krzysz00:
The problem is that I don't understand the algorithm at a high level well enough to provide such a counterexample.

The current fusion flow, as far as I'm concerned, is this:

Moving top-down, consider each fusable operation (a la.generic, a memref.copy, and so on). Find which of its inputs is connected to a buffer that is being written to by threadwise_write_all. 

(Then, for all other inputs, preform loads from global memory (via transforms) that will fetch values that will line up with with the values in the registers that are being written threadwise)

Then, fuse, by replacing the inputs with their in-register versions and the output with fresh registers, and swapping the fusable operation and the threadwise_write_all.

Finally, replace the temporary written by `threadwise_write_all` with the memory result the fusable operation targetted originally, eliminating the temporary (and extending the transformation stack, potentially).

As a special case, when the fusable operation is `memref.copy`, you don't need to allocate any new registers, just do that operation swap, since `memref.copy` doesn't affect the written values.

---

My proposed change is that we have the regularization step to make the multiple uses of a temporary distinct. Then, we amend the case for `memref.copy` to say that, when the target of the `memref.copy` is a temporary buffer, we clone the relevant threadwise_write_all and only update what the clone points to.

manupak:
Maybe we -- I -- are not thinking straight now -- it is quite late for me now.

@krzysz00 I dont see what you describing happening in single output cases for that we have many in the our codebase.
I ll try to put a long description on Monday. 

Again simplest anwser I could give is : 
Single output cases, there are two memrefs involved : 

```
%t1 = alloc()
%t2 = alloc()
threadwise_write_all %c -> %t1
la.generic %t1 ->  %t2
memref.copy %t2 -> %argN
```

There is at least two alloc s involved before memcpy optimizations, ALWAYS. When we trace from la.generic, %t1 is not copied out. Hence not cloned. But if we had : 

```
%t1 = alloc()
%t2 = alloc()
threadwise_write_all %c -> %t1
la.generic %t1 ->  %t2
memref.copy %t1 -> %argM
memref.copy %t2 -> %argN
```

then it is cloned-- rightfully so. 

Good night!

krzysz00:
I started writing this before your latest reply - and yeah, let's resume Monday. Looks like we're dealing with almost the same example.

I want to understand, concretely, how what happens in the fusion pass to the IR:
```mlir
%t1 = alloc()
%t2 = alloc()
%viewT1 = [... a bunch of transforms ...](%t1)
threadwise_write_all %c -> %viewT1
la.generic ins(%t1) outs(%t2) {...}
memref.copy %t2 -> %argN
```

differs from what happens for the IR
```mlir
%t1 = alloc()
%t2 = alloc()
%viewT1 = [... a bunch of transforms ...](%t1)
threadwise_write_all %c -> %viewT1
la.generic ins(%t1) outs(%t2) {...}
memref.copy %t2 -> %argM
memref.copy %t2 -> %argN
```
When is the decision to create the extra `threadwise_write_all` taken? What are the conditions for determining there needs to be a clone?

(Under my proposal, the above two-output code would have been regularized to
```mlir
%t1 = alloc()
%t2 = alloc()
%viewT1 = [... a bunch of transforms ...](%t1)
threadwise_write_all %c -> %viewT1
la.generic ins(%t1) outs(%t2) {...}
%t_distinguish = alloc()
memref.copy %t2 -> %distinguish
memref.copy %t2 -> %argM
memref.copy %distinguish -> %argN
```
and the fusion rules would turn `memref.copy %t2 -> %distinguish` into a clone of the `threadwise_write_all`)

And re your example, consider also
```mlir
%t1 ..., %tN = alloc()
// Transforms elided for clarity
%c = threadwise_write_all %c -> %t1
la.generic ins(%t1), outs(%t2) { ... f ... }
la.generic ins(%t2) outs (%t3) { ... g ... }
la.generic ins(%t2) outs (%t4) { ... h ... }
memref.copy %t3 -> %argM
memref.copy %t4 -> %argN
```
where you need to clone the threadwise_write_all at the first linalg.generic, but neither of the uses of `%t2` is, in your terms, copied out.

I also find the fact that we're using ordering in the IR as a way to pick out which threadwise_write_all to treat as the source values. It feels really fragile.)

manupak:
(Last message for the day,  I promise)

I dont see the need to regularize as such... unnecessary coupling between two passes.

> differs from what happens for the IR
> 
> ```mlir
> %t1 = alloc()
> %t2 = alloc()
> %viewT1 = [... a bunch of transforms ...](%t1)
> threadwise_write_all %c -> %viewT1
> la.generic ins(%t1) outs(%t2) {...}
> memref.copy %t2 -> %argM
> memref.copy %t2 -> %argN
> ```

I think you mean : 

 ```mlir
 %t1 = alloc()
 %t2 = alloc()
 %viewT1 = [... a bunch of transforms ...](%t1)
 threadwise_write_all %c -> %viewT1
 la.generic ins(%t1) outs(%t2) {...}
 memref.copy %t1 -> %argM
 memref.copy %t2 -> %argN
 ```

> When is the decision to create the extra threadwise_write_all taken? What are the conditions for determining there needs to be a clone?

It is taken when la.generic input %t1 is copied out or not....
if %t1 copied as in your example then clone if not dont as in my first example of my previous message.

> I also find the fact that we're using ordering in the IR as a way to pick out which threadwise_write_all to treat as the source values. It feels really fragile.)

Where is the order being used ? if you are talking about multiple tw_write_all being found -- I just removed it. We dont rely on the order because each op writes to unique memrefs.

manupak:
I've done few more changes to generalize cloning criteria...

Let me try another shot in explaining things..

## What is different from Friday ? 

Previuosly, this PR was checking the input memref of Linalg::Generic, that leads to a tw_write_all, is being copied out as the cloning criteria because it was sufficient for our lowering flow.

Now, I've generalized that even more...
Ultimately, what needs to be checked is in 'any' op (linalg or mem copy) rewrite in this fusion pass, is that whether input operand that leads to tw_write_all has the current op being rewritten as the unique reader. In an analogous world of SSA IR, this is similar as dominance check. I've done those changes now and please have a look whether that is more understandable...

## My understanding on the current lowering flow 

We start from a tensor graph which is in SSA form.

E.g. : 

```mlir
    func.func @dot_add(%arg0: tensor<1x5x4xf32>, %arg1: tensor<1x4x3xf32>, %arg2: tensor<1x5x3xf32>, %arg3: tensor<1x5x3xf32>) -> tensor<1x5x3xf32> attributes {arch = "", kernel} {
    %0 = "tosa.reshape"(%arg0) {new_shape = array<i64: 1, 5, 4>} : (tensor<1x5x4xf32>) -> tensor<1x5x4xf32>
    %1 = "tosa.reshape"(%arg1) {new_shape = array<i64: 1, 4, 3>} : (tensor<1x4x3xf32>) -> tensor<1x4x3xf32>
    %2 = "tosa.matmul"(%0, %1) : (tensor<1x5x4xf32>, tensor<1x4x3xf32>) -> tensor<1x5x3xf32>
    %3 = "tosa.reshape"(%2) {new_shape = array<i64: 1, 5, 3>} : (tensor<1x5x3xf32>) -> tensor<1x5x3xf32>
    %4 = "tosa.add"(%3, %arg2) : (tensor<1x5x3xf32>, tensor<1x5x3xf32>) -> tensor<1x5x3xf32>
    %5 = "tosa.add"(%4, %arg3) : (tensor<1x5x3xf32>, tensor<1x5x3xf32>) -> tensor<1x5x3xf32>
    return %5 : tensor<1x5x3xf32>
  }
```

Then before linalg elementwise fusion it would look as follows : 

```mlir
// -----// IR Dump Before LinalgElementwiseOpFusion (linalg-fuse-elementwise-ops) //----- //
func.func @dot_add(%arg0: tensor<1x5x4xf32>, %arg1: tensor<1x4x3xf32>, %arg2: tensor<1x5x3xf32>, %arg3: tensor<1x5x3xf32>) -> tensor<1x5x3xf32> attributes {arch = "", kernel} {
  %0 = bufferization.alloc_tensor() : tensor<1x5x3xf32>
  %1 = rock.gemm %0 = %arg0 * %arg1 features =  none storeMethod =  set {arch = ""} : tensor<1x5x3xf32> = tensor<1x5x4xf32> * tensor<1x4x3xf32> -> tensor<1x5x3xf32>
  %2 = tensor.empty() : tensor<1x5x3xf32>
  %3 = linalg.generic {indexing_maps = [affine_map<(d0, d1, d2) -> (d0, d1, d2)>, affine_map<(d0, d1, d2) -> (d0, d1, d2)>, affine_map<(d0, d1, d2) -> (d0, d1, d2)>], iterator_types = ["parallel", "parallel", "parallel"]} ins(%1, %arg2 : tensor<1x5x3xf32>, tensor<1x5x3xf32>) outs(%2 : tensor<1x5x3xf32>) {
  ^bb0(%in: f32, %in_0: f32, %out: f32):
    %6 = arith.addf %in, %in_0 : f32
    linalg.yield %6 : f32
  } -> tensor<1x5x3xf32>
  %4 = tensor.empty() : tensor<1x5x3xf32>
  %5 = linalg.generic {indexing_maps = [affine_map<(d0, d1, d2) -> (d0, d1, d2)>, affine_map<(d0, d1, d2) -> (d0, d1, d2)>, affine_map<(d0, d1, d2) -> (d0, d1, d2)>], iterator_types = ["parallel", "parallel", "parallel"]} ins(%3, %arg3 : tensor<1x5x3xf32>, tensor<1x5x3xf32>) outs(%4 : tensor<1x5x3xf32>) {
  ^bb0(%in: f32, %in_0: f32, %out: f32):
    %6 = arith.addf %in, %in_0 : f32
    linalg.yield %6 : f32
  } -> tensor<1x5x3xf32>
  return %5 : tensor<1x5x3xf32>
}
```

Note that there would always be  tensor.empty() created per tosa operator -- unless they get eliminated by linalg elementwise fusion pass and none of this buffers will be copied out of the function.

Then after the highlevel pipeline is executed it would be something like : 

```mlir
func.func @dot_add(%arg0: memref<1x5x4xf32>, %arg1: memref<1x4x3xf32>, %arg2: memref<1x5x3xf32>, %arg3: memref<1x5x3xf32>, %arg4: memref<1x5x3xf32>) attributes {arch = "", kernel} {
    %alloc = memref.alloc() {alignment = 64 : i64} : memref<1x5x3xf32>
    rock.gemm %alloc = %arg0 * %arg1 features =  none storeMethod =  set {arch = ""} : memref<1x5x3xf32> = memref<1x5x4xf32> * memref<1x4x3xf32>
    %0 = rock.transform %alloc by #transform_map : memref<1x5x3xf32> to memref<5x3xf32>
    %1 = rock.transform %arg2 by #transform_map : memref<1x5x3xf32> to memref<5x3xf32>
    %2 = rock.transform %arg3 by #transform_map : memref<1x5x3xf32> to memref<5x3xf32>
    %alloc_0 = memref.alloc() {alignment = 64 : i64} : memref<5x3xf32>
    linalg.generic {indexing_maps = [#map1, #map1, #map1, #map1], iterator_types = ["parallel", "parallel"]} ins(%0, %1, %2 : memref<5x3xf32>, memref<5x3xf32>, memref<5x3xf32>) outs(%alloc_0 : memref<5x3xf32>) {
    ^bb0(%in: f32, %in_1: f32, %in_2: f32, %out: f32):
      %4 = arith.addf %in, %in_1 : f32
      %5 = arith.addf %4, %in_2 : f32
      linalg.yield %5 : f32
    }
    %3 = rock.transform %alloc_0 by #transform_map1 : memref<5x3xf32> to memref<1x5x3xf32>
    memref.copy %3, %arg4 : memref<1x5x3xf32> to memref<1x5x3xf32>
    return
  }
```

Now lets a take look at pre align tiling IR : 

```mlir
#map = affine_map<(d0, d1) -> (0, d0, d1)>
#map1 = affine_map<(d0, d1) -> (d0, d1)>
#map2 = affine_map<(d0, d1, d2) -> (d0 * 5 + d1, d2)>
#transform_map = #rock.transform_map<#map by [<Merge{1, 5} ["dim0"] at [0] -> ["col0", "col1"] at [0, 1]>, <PassThrough ["dim1"] at [1] -> ["dim1"] at [2]>] bounds = [5, 3] -> [1, 5, 3]>
#transform_map1 = #rock.transform_map<#map2 by [<Unmerge{1, 5} ["exp0", "exp1"] at [0, 1] -> ["dim0"] at [0]>, <PassThrough ["dim1"] at [2] -> ["dim1"] at [1]>] bounds = [1, 5, 3] -> [5, 3]>
module {
  func.func @dot_add(%arg0: memref<1x5x4xf32>, %arg1: memref<1x4x3xf32>, %arg2: memref<1x5x3xf32>, %arg3: memref<1x5x3xf32>, %arg4: memref<1x5x3xf32>) attributes {arch = "", kernel} {
    %alloc = memref.alloc() {alignment = 64 : i64} : memref<1x5x3xf32>
    rock.gemm %alloc = %arg0 * %arg1 features =  none storeMethod =  set {arch = ""} : memref<1x5x3xf32> = memref<1x5x4xf32> * memref<1x4x3xf32>
    %0 = rock.transform %alloc by #transform_map : memref<1x5x3xf32> to memref<5x3xf32>
    %1 = rock.transform %arg2 by #transform_map : memref<1x5x3xf32> to memref<5x3xf32>
    %2 = rock.transform %arg3 by #transform_map : memref<1x5x3xf32> to memref<5x3xf32>
    %alloc_0 = memref.alloc() {alignment = 64 : i64} : memref<5x3xf32>
    linalg.generic {indexing_maps = [#map1, #map1, #map1, #map1], iterator_types = ["parallel", "parallel"]} ins(%0, %1, %2 : memref<5x3xf32>, memref<5x3xf32>, memref<5x3xf32>) outs(%alloc_0 : memref<5x3xf32>) {
    ^bb0(%in: f32, %in_1: f32, %in_2: f32, %out: f32):
      %4 = arith.addf %in, %in_1 : f32
      %5 = arith.addf %4, %in_2 : f32
      linalg.yield %5 : f32
    }
    %3 = rock.transform %alloc_0 by #transform_map1 : memref<5x3xf32> to memref<1x5x3xf32>
    memref.copy %3, %arg4 : memref<1x5x3xf32> to memref<1x5x3xf32>
    return
  }
}
```

So in out fusion flow, when we do two main types of rewrites : 
F1. linalg generic rewrite
F2. memref.copy rewrite

In either case, we trace the input of those operations -- that is a memref -- to its writer who is tw_write_all operation. Then we had been just mutate destination of tw_write_all to be the output memref of the current operation being rewritten (F1 or F2).

IMO, that is not 100% correct. 
Why ? 
We should be really checking whether that input memref has any more 'readers' before doing that. If it has more readers, then we should be really cloning tw_write_all  operation and then mutate the clone.

So we encounter this non-unique reader scenario in multiple outputs and therefore we should be cloning and doing the rewrite for both F1 and F2.

## Why I dont like 'distinguish' redundant copy approach ?

> (Under my proposal, the above two-output code would have been regularized to
> 
> ```mlir
> %t1 = alloc()
> %t2 = alloc()
> %viewT1 = [... a bunch of transforms ...](%t1)
> threadwise_write_all %c -> %viewT1
> la.generic ins(%t1) outs(%t2) {...}
> %t_distinguish = alloc()
> memref.copy %t2 -> %distinguish
> memref.copy %t2 -> %argM
> memref.copy %distinguish -> %argN
> ```

I dont like this solution because "memref.copy %t2 -> %distinguish" is just a redundant copy and it doesn't express or lower any more information than it originally had. Moreover, it also promotes fragility in the align tiling pass that it always require the IR to be in this form and it wont work for the original (which is Isomorphic) form. Therefore, I disagree with approach. 

Also, Im not a fan of putting more things in regularize that promotes contrived coupling between the passes in the lowering pipeline  -- I think we should rather introduce more ops and use canonicalization with strict op verifiers. Thus, it reduces ambiguity in the IR (e.g. extraViews, view balancing from readers to writers, etc).

krzysz00:
@manupak I did specifically mean an example that copies `%t2` twice there.

Also

> I dont see the need to regularize as such... unnecessary coupling between two passes.

Those passes are coupled already: regularize establishes the conditions needed for align-tiling to be able to work. The _point_ of regularize is to transform the program into a semantically equivalent program that's easier to handle in align-tiling.

krzysz00 (approved):
Ok, now that you've restructured things, I can at least understand the algorithm - during fusion, we check to see if there are other readers: if there are, we work on a clone of the `threadwise_write_all` for all but the last reader so as to not break anything.

And with the said restructuring, I have no correctness concerns, approved.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -138,48 +137,151 @@ static Value applyTransforms(PatternRewriter &b, ThreadwiseWriteAllOp storeOp,
...
+}
+
+// This function traces to non view readers.
+// TODO(@manupak): we should really be implementing
+// MemoryEffectOpInterface to ops to make this simpler
+// to identify readers.
+static LogicalResult
+traceToNonViewReaders(Operation *op, Value parentVal,
+                      SmallVectorImpl<Operation *> &nonViewReaders) {
   if (auto transform = dyn_cast<TransformOp>(op)) {
```
krzysz00, line 159: I'll just comment here that this `transform` care shouldn't be firing - the regularization pass ought to eliminate transforms on readers of things written by, say, a threadwise_write_all or la.generic.

PR #993, opened 2023-03-02
Decouple kpack and xdlops private buffers

krzysz00:
Storing vectors of length kpack in the buffers passed to the threadwise gemm is awkward and isn't what threadwise gemm actually needs. However, to avoid blowups during MemToReg, we probably still want the values in those arrays to be vectors. Therefore, change blockwise gemm to construct vectors of the argument type of the underlying MFMA and to pass those into xdlops_gemm_v2.

This should not have any correctness effect, and is extremely likely to have performance impact.

It also enables, for example, xpack=8 for f32, in case that ends up being useful.

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
@@ -326,10 +326,8 @@ struct BlockwiseGemmV2RewritePattern
            "LDS buffer segment for A is kpack-aligned");
     assert(ldsOffsetB % KPack == 0 &&
            "LDS buffer segment for B is kpack-aligned");
-    auto dataType = adaptor.getMatrixA()
-                        .getType()
-                        .template cast<MemRefType>()
-                        .getElementType();
+    auto dataType =
+        adaptor.getMatrixA().getType().cast<MemRefType>().getElementType();
 
     // The address calculations into the LDS buffer assume that the buffer
     // has type vector<KPack x T>. Then, we convert that into an address
@@ -442,14 +440,52 @@ struct BlockwiseGemmV2RewritePattern
         sourceOffset = kb.create<MulIOp>(
             loc, sourceOffset, kb.create<ConstantIndexOp>(loc, KPack));
 
-      Value destOffset = k_i;
-
-      // Emit load/store
+      Type loadType = vectorTypeOrSelf(dataType, KPack);
+      Value value =
+          kb.create<InBoundsLoadOp>(loc, loadType, ldsOrig, sourceOffset);
       auto bufferType = regDest.getType().cast<MemRefType>();
       Type bufferElementType = bufferType.getElementType();
-      Value value = kb.create<InBoundsLoadOp>(loc, bufferElementType, ldsOrig,
-                                              sourceOffset);
-      kb.create<memref::StoreOp>(loc, value, regDest, ValueRange{destOffset});
+
+      // We're loading in units of kPack, but storing in units of k_base.
+      if (KPack == k_base) {
+        Value destOffset = k_i;
+        kb.create<memref::StoreOp>(loc, value, regDest, ValueRange{destOffset});
+      } else if (KPack > k_base) {
+        int64_t numStores = KPack / k_base;
+        Value baseDestOffset = kb.createOrFold<arith::MulIOp>(
+            loc, k_i, kb.createOrFold<arith::ConstantIndexOp>(loc, numStores));
+        for (int64_t i = 0; i < numStores; ++i) {
+          Value sliceStart =
+              kb.createOrFold<arith::ConstantIndexOp>(loc, k_base * i);
+          Value slice = kb.create<ExtractSliceOp>(loc, bufferElementType, value,
+                                                  sliceStart);
+          Value destOffset = kb.createOrFold<arith::AddIOp>(
+              loc, baseDestOffset,
+              kb.createOrFold<arith::ConstantIndexOp>(loc, i));
+          kb.create<memref::StoreOp>(loc, slice, regDest,
+                                     ValueRange{destOffset});
+        }
+      } else if (KPack < k_base) {
+        // Here we are gathering loaded values into vectors for passing into
+        // MFMAs.
+        Value destValsPerKpack =
+            kb.createOrFold<arith::ConstantIndexOp>(loc, k_base / KPack);
+        // This is fine, since the inputs to MFMAs are contiguous in the k
+        // dimension.
+        Value destOffset =
+            kb.createOrFold<arith::DivUIOp>(loc, k_i, destValsPerKpack);
+        Value destVecPart =
+            kb.createOrFold<arith::RemUIOp>(loc, k_i, destValsPerKpack);
+        Value destSlicePos = kb.createOrFold<arith::MulIOp>(
+            loc, destVecPart,
+            b.createOrFold<arith::ConstantIndexOp>(loc, KPack));
+        Value destVec = kb.create<memref::LoadOp>(
+            loc, bufferElementType, regDest, ValueRange{destOffset});
+        Value newDestVec = kb.create<InsertSliceOp>(
+            loc, bufferElementType, value, destVec, destSlicePos);
+        kb.create<memref::StoreOp>(loc, newDestVec, regDest,
+                                   ValueRange{destOffset});
+      }
     };
 
     auto ldsToRegisterCopyKdim =
--- a/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
@@ -1176,6 +1176,7 @@ struct GridwiseGemmV2RewritePattern
     int64_t rowGroupsPerBlock = mfmaAttr.rowGroupsPerBlock;
     int64_t inputSpanLen = mfmaAttr.inputSpanLen;
     int64_t inputSpansPerMfmaIn = mfmaAttr.inputSpansPerMfmaIn;
+    int64_t k_base = mfmaAttr.k_base;
     int64_t blocksInOutRegs = mfmaAttr.blocksInOutRegs;
 
     int64_t blocksPerRepeat = (mPerRepeat * nPerRepeat) / (m * n);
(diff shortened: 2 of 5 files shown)


mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -442,14 +440,52 @@ struct BlockwiseGemmV2RewritePattern
...
+      Type loadType = vectorTypeOrSelf(dataType, KPack);
+      Value value =
+          kb.create<InBoundsLoadOp>(loc, loadType, ldsOrig, sourceOffset);
       auto bufferType = regDest.getType().cast<MemRefType>();
       Type bufferElementType = bufferType.getElementType();
-      Value value = kb.create<InBoundsLoadOp>(loc, bufferElementType, ldsOrig,
-                                              sourceOffset);
-      kb.create<memref::StoreOp>(loc, value, regDest, ValueRange{destOffset});
+
+      // We're loading in units of kPack, but storing in units of k_base.
```
jerryyin, line 449: So you are effectively vectorizing the load/store in lds -> register transfer? That's good motivation!

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -442,14 +440,52 @@ struct BlockwiseGemmV2RewritePattern
...
+              kb.createOrFold<arith::ConstantIndexOp>(loc, k_base * i);
+          Value slice = kb.create<ExtractSliceOp>(loc, bufferElementType, value,
+                                                  sliceStart);
+          Value destOffset = kb.createOrFold<arith::AddIOp>(
+              loc, baseDestOffset,
+              kb.createOrFold<arith::ConstantIndexOp>(loc, i));
+          kb.create<memref::StoreOp>(loc, slice, regDest,
+                                     ValueRange{destOffset});
+        }
+      } else if (KPack < k_base) {
```
jerryyin, line 468: ~~Implementation wise looks good though. I don't think this will happen, maybe just add an assert here. (Because we are not packing enough, it takes multiple iterations).~~

Coming to think about it, I think you got a point in having this enabled. Although we'll have to take multiple iterations to fill up a k_base, but bank conflict in loading lds might get better. It's worth taking smaller kPack cases to experiment to see if that can actually improve performance.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1202,29 +1203,17 @@ struct GridwiseGemmV2RewritePattern
...
-          MemRefType::get({arrayBSize}, VectorType::get({kpack}, elementType),
-                          AffineMap{}, privateMemoryAddressSpace);
-    } else {
-      arrayAType = MemRefType::get({arrayASize}, elementType, AffineMap{},
-                                   privateMemoryAddressSpace);
-      arrayBType = MemRefType::get({arrayBSize}, elementType, AffineMap{},
-                                   privateMemoryAddressSpace);
-    }
+    arrayAType = MemRefType::get({inputBufferSize}, mfmaGroup.getArgType(),
+                                 AffineMap{}, privateMemoryAddressSpace);
```
jerryyin, line 1214: I'm a little surprised that you transitioned from a unit of kpack to a unit of k_base. I thought we'd go with unit of 1 and let backend consolidate where it make sense. At the same time, it make sense to pick the unit of k_base because that's eventually what our xdlops will be able to take advantage of.

mlir/lib/Dialect/Rock/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -248,88 +244,22 @@ struct XdlopsGemmV2RewritePattern : public OpConversionPattern<XdlopsGemmV2Op> {
...
+      {
+        OpBuilder::InsertionGuard guard(b);
+        b.setInsertionPointToStart(mfmaLoop.getBody());
+        Value coord = mfmaLoop.getLowerCoords(/*domain=*/0)[0];
+
+        Value regIdxA = b.create<arith::AddIOp>(loc, coord, regAOffset);
+        Value regIdxB = b.create<arith::AddIOp>(loc, coord, regBOffset);
+        Value argA = b.create<memref::LoadOp>(loc, argType, bufferA, regIdxA);
+        Value argB = b.create<memref::LoadOp>(loc, argType, bufferB, regIdxB);
+        populateMfma(b, argA, argB, regCOffset);
```
jerryyin: This used to be the part that I hate most to accommodate the incompatibility of kpack and kbase when they don't fit with each other. Now that you made the unit to be k_base, the loop is just natural. Good job!

jerryyin (approved):
I'm impressed, this would take me longer to implement. Thank you!

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: No, yeah, I was remembering why we ended up with buffers of, say, `2xvector<32xf32>` instead of `64xf32` for matrix C.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: We already had this vectorized, I think

krzysz00: ... also, the alluded-to "bug" was actually the gpu on my favorite Lockhart being down

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: kPack < k_base is also known as kPack = 1 and a k reduction

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
jerryyin: So you are saying although previously we only load and store 1 element, the backend will consolidate the load/store a vector?

krzysz00: The comment above about the "bug" was in reference to a previous commit - I put it in the wrong place.

Now, as to this comment, blockwise gemm already loaded vectors of length kPack from LDS.

jerryyin: Ah right the unit used to be kPack.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
jerryyin: k reduction case usually has a kPack >= k_base > 1. 
 - kPack = 1 < k_base was previous supported
 - (kPack > 1) but < k_base is the case that you have just newly enabled in this PR

krzysz00: No, I'll note than none of this code actually depends on kpack != 1

So this retains support for kpack = 1, k_base > 1, but also adds support for the other option, which is still rejected by tuning because there's not really a good reason to have it.

jerryyin: Yah no I think we were on the same page and my previous comment was exactly talking about it, that you have supported more scenarios than before.

> but also adds support for the other option, which is still rejected by tuning because there's not really a good reason to have it.

Part of my implication here is to re-examine assumption again because it looks to me that by enabling this additional setup where we don't pack enough, there's a chance that it can improve certain cases/data type as well. I think your work might actually improve some int8 performance if we enable some narrower tuning. What that means is that we can experiment to see if we run a case where kPack < k_base, if performance can become better than the kPack > k_base.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1202,29 +1203,17 @@ struct GridwiseGemmV2RewritePattern
...
+    arrayBType = MemRefType::get({inputBufferSize}, mfmaGroup.getArgType(),
```
giuseros, line 1215: Why do you need two distinct `arrayAType` and `arrayBType`? Could you have a unique `arrayType`?

mlir/test/Dialect/Rock/lowering_xdlops_gemm_v2.mlir
```diff
@@ -1,15 +1,20 @@
...
+  // CHECK-SAME: bounds [2]
+  // CHECK: [[a:%.+]] = memref.load [[ABuf]]
+  // CHECK: [[b:%.+]] = memref.load [[BBuf]]
+  // CHECK: [[c:%.+]] = memref.load [[CBuf]]
+  // CHECK: amdgpu.mfma [[a]] * [[b]] + [[c]] {{.*}} : f32, vector<16xf32>
+  %c0 = arith.constant 0 : index
   rock.xdlops_gemm_v2 %matrixC += %matrixA[%c0] * %matrixB[%c0] {
     params = #rock.xdlops_gemm_params<
-       kPerBlock = 8,
+       kPerBlock = 4,
```
giuseros, line 17: Why do you have to change this?

giuseros (approved):
I also like this change a lot :) Very minor comments, and then if you could run a performance comparison before/after it would be great!

mlir/test/Dialect/Rock/lowering_xdlops_gemm_v2.mlir
krzysz00: I had to change this so the test would actually run correctly

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: Yeah, I could, and it's not an unreasonable cleanup.

However, once FP8 is happening, we'll want separate A and B types, so ... might as well not clean it up.

krzysz00:
@giuseros I did a perf run. The changes are all fractions of a % in random directions. For all I know (I might dump an assembly output) this PR is actually NFC

PR #996, opened 2023-03-03
Improve memory coalescing and maximize LDS vectorization

giuseros:
This is trying to achieve two things:
- Improve coalescence when the global matrix we are reading from is contiguous in the K dimension (instead of distributing the threads along M/N first and then K we do the other way round)
- Improve LDS vectorization by always maximizing the vectorization when storing in LDS 

This is the perf comparison against the latest `main/develop`:
[coalescing_lds_3.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10885021/coalescing_lds_3.zip)

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
@@ -82,16 +82,15 @@ static Type obtainAccumulatorType(OpBuilder &b, Type &elementType,
   return accumulatorType;
 }
 
-/// Given a G x K x D matrix and the block tuning parameters and how much data
-/// each thread will load,
-/// return the dimension in which the load of this matrix from global memory
-/// should be vectorized and the length of that vector load. Also takes
-/// `tiebreaker`, the vectorization dimension to be used when both choises are
-/// equal, which should be the vectorization dimension of the store to LDS.
+/// Given a copy layout <copyDPerThread, copyKPerThread>, come up with the best
+/// vectorization strategy for the layout. For instance, if the layout is <D,K>
+/// = <2,16> and K is contiguous, we will vectorize by 16 along K and we will
+/// loop over the other dimension
 static std::pair<GemmDimension, int64_t>
-bestVectorization(OpBuilder &b, Value matrix, int64_t dataPerThread,
-                  GemmDimension tiebreaker, int64_t kPerBlock,
-                  int64_t dPerBlock, Type elementType) {
+bestGlobalVectorization(OpBuilder &b, Value matrix, int64_t copyDPerThread,
+                        int64_t copyKPerThread, GemmDimension tiebreaker,
+                        int64_t kPerBlock, int64_t dPerBlock,
+                        Type elementType) {
   Value tensor;
   ArrayAttr transforms;
   std::tie(tensor, transforms) = untransform(b, matrix);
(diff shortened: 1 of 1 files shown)


mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -138,6 +138,24 @@ computeCopyPerThread(GemmDimension dim, int64_t vectorLen,
   return std::make_pair(copyKPerThread, copyDPerThread);
 }
 
+static FailureOr<std::pair<int64_t, int64_t>>
+computeCopyPerThreadV2(int64_t vectorLen, int64_t copyPerThread,
```
krzysz00: Why `V2`? You can just replace the old method with the new one?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -973,41 +997,53 @@ struct GridwiseGemmV2RewritePattern
...
-               << "gridSize: " << gridSize << "\n"
-               << "blockSize: " << blockSize << "\n"
-               << "aCopyPerThread: " << aCopyPerThread << "\n"
-               << "bCopyPerThread: " << bCopyPerThread << "\n"
-               << "aVectorDim: " << aVectorDim << "\n"
-               << "aVectorLen: " << aVectorLen << "\n"
-               << "bVectorDim: " << bVectorDim << "\n"
-               << "bVectorLen: " << bVectorLen << "\n"
-               << "vectorTiebreaker: " << vectorTiebreaker << "\n");
+    // When we use XDLops we always have a [K0, M/N, K1] LDS buffer
```
krzysz00: Use our names - [KOpter, M/N, kpack]

krzysz00 (commented):
I don't think this is it.

What I think should be happening here is that you choose the vectorization dimension (K or M) based on whether kPack is 1 or not.

Then, you vectorize the global load along that dimension.

And this way, you can also do the right thing for non-xdlops, which uses a [K, M] layout at the moment.

giuseros:
Hi @krzysz00 ,  
> What I think should be happening here is that you choose the vectorization dimension (K or M) based on whether kPack is 1 or not. Then, you vectorize the global load along that dimension.

I can select `copyKPerThread` or `copyDPerThread` based on Kpack like you propose, that's fine (I  did want only target xdlops). 

But why forcing a vectorization? What I want is to read from Global `maxVlen` elements along K (or M for Kpack 1) and `dataToRead/maxVlen` on the other dimension. Because I know that I will want to write `maxVlen` vectors into LDS. 

Then the vectorizer can do what it wants, as long as a single thread ends up with `maxVlen` elements ready to be written into LDS.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
giuseros: I wanted to explicitly target only xdlops, but I can extend to non-xdlops as well.

giuseros:
> Then, you vectorize the global load along that dimension.

To reiterate, before signing off, if my matrix is `int8` `KxM` I  want each thread to read 16 elements along K (one by one, since they are strided) and the rest along M. 

This is because those 16 elements will form a vector that will be easily stored into LDS. So the number of elements to read along M (with the possibility of vectorizing them as global loads) will be <1,2,4> but never 16.

krzysz00:
If we want a vectorized write to LDS, this is either vectorization in KPack if KPack > 1, or vectorization in M/N otherwise? That is, the global vectorization is the maximum vectorization you can achieve under the constraints of the LDS's vectorization?

krzysz00:
... Ok, I see what you're getting at. You're using the LDS format to generate constraints for the global vectorizer, namely what the K/D split has to be.

However, hold on a minute, what's wrong with writing 256 bits to LDS with a pair of ds_write_b128 instructions?

krzysz00:
(Also, can we go ahead and do the same optimization for the non-xdlops path?)

giuseros:
> However, hold on a minute, what's wrong with writing 256 bits to LDS with a pair of ds_write_b128 instructions?

If I understand correctly your suggestion, there are different reasons I wouldn't make this the default behaviour:
- When K is not the main dimension we would stop any possibility of vectorizing in global memory (because this would mean reading 2*128 bits along K)
- When K is not the main dimension, reading along K means reading strided values(by M or N) and this might have impact on caching
- When K is the main dimension we fall in the problem of distributing workitems too far apart (because each thread has now to read 2*128 bits) and we would have coalescing issues
- On a minor note, CK does never try this. They always vectorize (in LDS) by 128 bits and vectorize in global the best they can

I am not against the idea of investigating the possibility of trying different global/LDS layouts and letting the tuner decide the best one, but I would leave the one CK uses as default.

krzysz00:
Ok, the workitems too far apart thing makes some sense, yeah.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -138,6 +138,24 @@ computeCopyPerThread(GemmDimension dim, int64_t vectorLen,
   return std::make_pair(copyKPerThread, copyDPerThread);
 }
 
+static FailureOr<std::pair<int64_t, int64_t>>
+computeCopyPerThreadV2(int64_t vectorLen, int64_t copyPerThread,
+                       int64_t kPerBlock, int64_t dPerBlock, Location loc) {
+  int64_t copyKPerThread = std::min(vectorLen, copyPerThread);
```
krzysz00: Ok, to rephrase my comment, I don't think it's good to always have the k dimension be the vectorization dimension. The optimal vectorization for LDS is either K or M, depending on whethere kPack == 1 or not

Unless I've really misunderstood your changes

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -973,41 +997,53 @@ struct GridwiseGemmV2RewritePattern
...
+
+    // If we are vectorizing along dimension M or N, we cannot vectorize
+    // more than copyMPerThread
+    if (aVectorDim == GemmDimension::MorN) {
+      aVectorLen = std::min(aVectorLen, copyMPerThread);
+    }
+    if (bVectorDim == GemmDimension::MorN) {
+      bVectorLen = std::min(bVectorLen, copyNPerThread);
+    }
+
```
jerryyin: For understanding, it looks like those lines are key for this PR? 
 - Effectively, you want to limit the vectorization on M or N dimension when we are trying to make the decision to do a global -> vector load
 - The reasoning behind all this is due to you know lds layout is [K, M, kPack], kPack being the fastest changing dimension
   - Since kPack is always the fastest changing dimension, we prioritize maximize gemmK vectorization length over gemmM or gemmN

Please point out if any of my summary is wrong, just so that we are on the same page with your motivations.

------

My main question on this reasoning process is: Why would global load guard by the layout of lds store?
   - On a second thought, I think we should do global load only considering about global layout. Because the overhead of global load is a lot higher than LDS operations
   - That means that we shouldn't do global load based on vectorization length(existing logic). Rather, we should do global load based on fastest changing dimension. If global layout is [gemmM(k), gemmK(cyx)], then we should always try to both load and maximize the vectorization on gemmK dimension. 
   - Similarly, when the global layout is [k, y, x, c], again we should prioritize on gemmK dimension because the load stride between the gemmK dimension can be minimized.
   - Only when we have a layout that is [c, y, x, k] should we prioritize vectorization on gemmM/N dimension

Following this thought, I'd argue that both `aVectorDim` and `bVectorDim` should be `GemmDimension::K`, because we care about [k, c, y, x] or [k, y, x, c] layout a lot more than other layouts.

krzysz00: ... the fastest changing dimension will vectorize best, since you can't vectorize a strided dimension?

Though that does argue for a better tiebreak, namely that we need to extend the vectorizer to report out (since it already computes) what the mimimum stride in the vectorization was. That is, if we see "vectorization of 1, smallest stride in requested dimension = 1" and "vectorization of 1, smallest stride in requested dimension = H*W", we should pick the former.

But also also, it _feels_ like what the CK people found is that, since our global memory loads are software pipelined across a big block of computations, while our LDS operations aren't, having less efficient global loads (less vectorized) is less of a penalty than writing to LDS in an unoptimized way.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: > The optimal vectorization for LDS is either K or M, depending on whethere kPack == 1 or not

Agreed, so we should statically pick the optimal dimension (based on kPack) rather than make it based on potential vectorization length.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
jerryyin: > it feels like what the CK people found is that, since our global memory loads are software pipelined across a big block of computations, while our LDS operations aren't

Seems to be a reasonable point, the global load indeed get pipeline so penalty might be smaller

> That is, if we see "vectorization of 1, smallest stride in requested dimension = 1" and "vectorization of 1, smallest stride in requested dimension = H*W", we should pick the former.

Agreed.

giuseros: Maybe we can have a chat and talk about this? I am free in 1h30

giuseros:
I applied your changes, and it got quite more general, thanks guys! Basically now `computeCopyPerThread` decides the reading layout `<d,k>`  of each thread (from global)  and `bestGlobalVectorization` vectorizes it. I didn't retune but re-running `perfRunner` I got similar performance. I might retune just to see if something went even better.

krzysz00 (approved):
This looks pretty sensible to me, approved pending perf numbers

krzysz00:
... I just saw you run perfRunner, I take back the "pending". This is flat-out approved.

giuseros:
Just out of curiosity I ran the tuner again and the result is attached. The new algorithm improved things things even more (especially for `fp16` and `int8`). 

Merging this in 

[coalescing_lds_7.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/10931131/coalescing_lds_7.zip)

PR #997, opened 2023-03-06
Get rid of universal parameters

giuseros:
This is a refactor to get rid of the `universalParameters` used by the `Conv2DGenerator` to decide workspace size during wrw convolution. 

Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/800

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
@@ -257,7 +257,7 @@ private:
   // add more 3 gemmk.
   uint32_t obtainBlockSize(const InitParamsXDL &params, int64_t waveSize);
 
-  LogicalResult getKBlocks(int64_t convGroupSize, const GemmSize &gemmSize,
+  LogicalResult getKBlocks(int64_t batchSize, const GemmSize &gemmSize,
                            const InitParamsXDL &params, int64_t &gemmKBlocks,
                            uint32_t numCu);
 
--- a/mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
@@ -43,6 +43,30 @@ struct InitParams {
   int64_t gemmKPerBlock;
 };
 
+/// Store information useful for populating perf configurations
+struct PopulateParamsInfo {
+  GemmSize gemmSize;
+  GemmFeatures gemmFeatures;
+  Type inputType;
+  KernelType kernelType;
+  int64_t batchSize;
+  uint32_t numCu;
+
+  PopulateParamsInfo(GemmSize gemmSize, GemmFeatures gemmFeatures,
+                     Type inputType, KernelType kernelType)
+      : gemmSize(gemmSize), gemmFeatures(gemmFeatures), inputType(inputType),
+        kernelType(kernelType) {}
+
+  PopulateParamsInfo(GemmSize gemmSize, GemmFeatures gemmFeatures,
+                     Type inputType, KernelType kernelType, int64_t batchSize,
+                     uint32_t numCu)
+      : gemmSize(gemmSize), gemmFeatures(gemmFeatures), inputType(inputType),
+        kernelType(kernelType), batchSize(batchSize), numCu(numCu) {}
+
+  /// Extract the relevant information from a RockGemmWrapperInterface operation
+  static PopulateParamsInfo fromOp(RockGemmWrapperInterface op);
+};
+
 struct InitParamsNonXDL : InitParams, Serializable<InitParamsNonXDL> {
   constexpr InitParamsNonXDL(uint32_t bSize, int64_t mPerBlock,
                              int64_t nPerBlock, int64_t kPerBlock,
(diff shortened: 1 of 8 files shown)


mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
```diff
@@ -230,19 +255,17 @@ class PopulateParamsXDL : public BasePopulateParams<InitParamsXDL> {
   // if can't select config from above , use this config to do
   // padding kernel for example , GEMMK/block is 16 , if your gemmK is  13 , we
   // add more 3 gemmk.
-  static const InitParamsXDL universalParameters;
-
   uint32_t obtainBlockSize(const InitParamsXDL &params, int64_t waveSize);
 
-  LogicalResult getKBlocks(Conv2DBwdWeightOp op, const GemmSize &gemmSize,
-                           const InitParamsXDL &params, int64_t &gemmKBlocks);
+  LogicalResult getKBlocks(int64_t convGroupSize, const GemmSize &gemmSize,
```
krzysz00: This should be named `convBatchSize`?

krzysz00 (approved):
Looks like a very nice refactor to me! I'd wait for @manupak to take a look, but I don't see any real issues here

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
giuseros: I called `batchSize` to be coherent with the rest.

mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
```diff
@@ -383,41 +350,35 @@ bool Conv2dGenerator::needExtraPadBwdWeight(OpBuilder &builder) const {
...
-      // the following code will return if the resulting config needs padding
-      needExtraPad =
-          calculatePaddingKernelSize(gemmSize, dir, dataType, populateParams)
-              .has_value();
-    }
+    uint32_t gridSize;
+    auto res = populateParams.obtainTuningParameters(info, 0, config.perfConfig,
+                                                     validParams, gridSize);
+    needExtraPad =
+        succeeded(res) &&
```
manupak: So this translate to not having a perf_config str --> no need to pad. Is that always true ?

manupak (commented):
~~Are we saying the ultimate consumers of this API (i.e. getKernelCount and hasWorkspace) will always have a perf config~~
I think we should exit compilation if obtainTuningParams fails because we need to concretely know whether padding is needed or not.

mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
manupak: ok my bad that is not correct either -- it will go through the list.

manupak: however, I dont like the fact obtainTuningParameters failing always meaning there is not going to be pad. 
If we dont expect obtainTuningParameters to fail, maybe we exit the compilation instead ? WDYT ?

giuseros: About the perf_config, yes, it will go through the default list (ordered by padding). About exit the compilation when `obtainTuningParameters` fails, I don't have a strong opinion, but how do I do this? Do you mean adding an assert?

manupak: AFAIK, this is how these are ultimately used... 

https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/f432b86debf296cd32ce36384d2ee01bcc762dfa/mlir/tools/rocmlir-lib/rocmlir-lib.cpp#L132-L139

So we'd need to return nullptr ? 
@jerryyin thoughts ?

giuseros: You mean something like:
``` 
if (failed(conv2dGenerator.getWorkspaceSize(module))) { 
     return nullptr; 
} 
```
?

manupak: Yea, something like that..

manupak: (It is also used in kernel count as well...)

giuseros: Alright, I did it.

mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
```diff
@@ -40,7 +40,7 @@ bool isWrWAtomicKernel(GemmFeatures features, const Type &dataType,
 // reasonable reduction of GemmK after splitting, without incurring too much
 // overheads on atomic adds. One potential future work is to make this value be
 // tunable.
-LogicalResult calculateKBlockNum(const ConvolutionDims &convDims,
+LogicalResult calculateKBlockNum(const int64_t convGroupSize,
```
krzysz00: Shouldn't this be `convBatchSize`? Group is the `G` dimension

krzysz00 (commented):
Ok, so, just for my understanding: when does getting the kernel count fail?

My current theory of when those functions fail is "there exists no tuning config that can handle the given problem, no matter how much padding we add".

manupak:
I think so... and also it can be an layering (contrived) artifact of using higher order function such as obtainTuningParams that is possible to fail. 

I think mostly it fails when a given perf config is not applicable (because applicability is checked under obtainTuningParams)  ... rather than non-existence of a valid config in the list -- but not 100% sure about the latter either.

krzysz00:
Yeah, I think passing in an invalid perf config is the way to fail this ... and even then, in affix-tuning-parameters, you can ask for MLIR to fall back to heuristic mode - I think we added that option for MIGraphX's purposes

krzysz00:
(Or at least that used to be how things worked)

manupak:
If a perf-config is given and it doesn't work -- IMO we shouldn't fall back to heuristic mode. 
It will confuse a client who is a tuner....

manupak:
Instead, we could lift the populateDerive call for a given perf-config scenario out of obtainTuningParams.. 
In reality, there are no tuning params to be obtained... it is given

krzysz00:
There are still derived parameters: grid size and block size are the big ones.

Also, the fallback option is, IIRC, off by default, but it's there in case someone wants it. I vaguely member that the MIGraphX people did at one point.

manupak (approved):
LGTM!

PR #998, opened 2023-03-06
Refactor type lists in RockOps.td, declare Float8 support

krzysz00:
This commit does not actually add any support for FP8. It does, however, declare that all the relevant ops (Gemms, loads/stores, ...) support the float8 types.

I also went ahead and extracted out various repeated lists of types into common variables.

While I was here, I tightened up the validations for what the arguments to xdlops gemm can be, which required fixing some tests tthat wouldn't've lead to valid code.


jerryyin (approved):

sjw36:
Nice cleanup!

giuseros (approved):
LGTM

PR #999, opened 2023-03-06
Adding migraphx int8 quantization e2e support

jerryyin:
This PR targets at fixing issues related with migraphx int8 quantization and its e2e tests:
 - migraphx.conv2d bias type was wrong to set to filter type
 - Clone flow validation should use naive type comparisons in quantized int8
 - Added additional e2e test for int8 conv and int8 conv with quantization
 - Added convenient script `xmir-run`


mlir/utils/widgets/xmir-run
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+declare -g BASEDIR="none-such"
+for dir in "$PWD/.." "$PWD" "$PWD/build" \
+                  "$HOME/rocmlir/build"; do
+    if [[ -x "$dir/bin/rocmlir-driver" ]]; then
```
krzysz00: `xmir-runner`?

sjw36 (approved):
Looks good.

mlir/utils/widgets/xmir-run
jerryyin: Good point, will do.

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
```diff
@@ -629,3 +630,55 @@ extern "C" void mcpuVerifyInt32Int64(int32_t *gpuAllocated, int32_t *gpuAligned,
...
+
+  if (failure_count == 0) {
+    printf("[1 1 1]\n");
+  } else {
+    if (print_option == PrintOption::Always ||
+        print_option == PrintOption::Failure ||
+        print_option == PrintOption::Summary) {
+      printf("Number of elements: %ld\n", dataSize);
+      printf("Number of incorrect elements: %ld\n", failure_count);
+      std::cout << "maxAbsDiff: " << maxAbsDiff << std::endl;
```
manupak, line 668: super-nit : why mixed c/c++ style prints here ?

manupak (approved):
LGTM!

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
jerryyin: This allows me to print numbers in type T without enforcing a format. The other printf are consistent with other verification functions, in which the size/failure count are in similar format. I figured this is the short-cut to make :-p

PR #1001, opened 2023-03-06
Rocmlir package

yiqian1:
Use rocm-cmake to create a deb package for librockcompiler.

To create the package:
cmake -G Ninja .. -DBUILD_FAT_LIBROCKCOMPILER=ON -DCPACK_INSTALL_PREFIX=instdir [-DEXPORT_ALL_HEADERS=ON]
cmake --build . --target package

This will generate librockcompiler-rocmlir_1.1.0_amd64.deb and rockcompiler-headers-rocmlir_1.1.0_amd64.deb if EXPORT_ALL_HEADERS=ON.

To install
dpkg -i librockcompiler-rocmlir_1.1.0_amd64.deb

If CPACK_INSTALL_PREFIX is not defined, the package will be installed at /usr/local.

### DIFF
--- a/mlir/tools/rocmlir-lib/CMakeLists.txt
+++ b/mlir/tools/rocmlir-lib/CMakeLists.txt
@@ -247,6 +247,6 @@ if(BUILD_FAT_LIBROCKCOMPILER)
   rocm_create_package(
     NAME ${PACKAGE_NAME}
     DESCRIPTION "MLIR packages"
-    MAINTAINER "rocMLIR"
+    MAINTAINER "rocMLIR Dev Team dl.dl-mlir@amd.com"
   )
 endif()
--- a/mlir/tools/rocmlir-lib/CMakeLists.txt
+++ b/mlir/tools/rocmlir-lib/CMakeLists.txt
@@ -144,11 +144,11 @@ if(BUILD_FAT_LIBROCKCOMPILER)
     if(ARG_EXTRA_INCLUDES)
       set(EXTRA_INCLUDES ${ARG_EXTRA_INCLUDES})
     endif()
+    set(ROCM_INSTALL_LIBDIR lib/${package_name})
     rocm_install(TARGETS ${lib_name}
       EXPORT ${export_set}
-      ARCHIVE DESTINATION lib/${package_name}
-      INCLUDES DESTINATION ${INCLUDE_DIR} ${EXTRA_INCLUDES}
       COMPONENT ${component_name}
+      INCLUDES DESTINATION ${INCLUDE_DIR} ${EXTRA_INCLUDES}
     )
 
     # Generate package config and version file
--- a/mlir/tools/rocmlir-lib/CMakeLists.txt
+++ b/mlir/tools/rocmlir-lib/CMakeLists.txt
@@ -40,6 +40,13 @@ mlir_check_link_libraries(rocmlir-lib-test)
 llvm_canonicalize_cmake_booleans(BUILD_FAT_LIBROCKCOMPILER)
 # Static library target, enabled only when building static libs
 if(BUILD_FAT_LIBROCKCOMPILER)
+  find_package(ROCM 0.8 REQUIRED PATHS /opt/rocm)
+  include(ROCMInstallTargets)
+  include(ROCMCreatePackage)
+
+  set(ROCMCHECKS_WARN_TOOLCHAIN_VAR OFF)
+  set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_SOURCE_DIR}/LICENSE")
+
   set(PACKAGE_NAME ${CMAKE_PROJECT_NAME})
   set(LIBRARY_NAME rockCompiler)
   function(combine_archives output_archive)
@@ -76,9 +83,18 @@ if(BUILD_FAT_LIBROCKCOMPILER)
     ${LIBRARY_NAME}
   )
 
+  rocm_package_setup_component(lib${LIBRARY_NAME}
+        LIBRARY_NAME lib${LIBRARY_NAME}
+        PACKAGE_NAME ${PACKAGE_NAME}
+  )
+
+  rocm_install(FILES LICENSE
+    DESTINATION share/doc/${PACKAGE_NAME}
+    COMPONENT lib${LIBRARY_NAME})
+
   # Install Miir.h to ${CMAKE_INSTALL_PREFIX}/include/${PACKAGE_NAME}/
   # as part of component libroCompiler
-  install(FILES Miir.h
+  rocm_install(FILES Miir.h
     DESTINATION include/${PACKAGE_NAME}
     COMPONENT lib${LIBRARY_NAME})
 
@@ -128,11 +144,11 @@ if(BUILD_FAT_LIBROCKCOMPILER)
     if(ARG_EXTRA_INCLUDES)
       set(EXTRA_INCLUDES ${ARG_EXTRA_INCLUDES})
     endif()
-    install(TARGETS ${lib_name}
+    rocm_install(TARGETS ${lib_name}
       EXPORT ${export_set}
       ARCHIVE DESTINATION lib/${package_name}
-      COMPONENT ${component_name}
       INCLUDES DESTINATION ${INCLUDE_DIR} ${EXTRA_INCLUDES}
+      COMPONENT ${component_name}
     )
 
     # Generate package config and version file
@@ -152,14 +168,14 @@ if(BUILD_FAT_LIBROCKCOMPILER)
     )
 
     # Generate Targets.cmake file for the export-set to contain imported targets
-    install(EXPORT ${export_set}
+    rocm_install(EXPORT ${export_set}
       NAMESPACE ${package_name}::
       DESTINATION ${LIB_CMAKE_DIR}
       COMPONENT ${component_name}
     )
 
     # Install package configuration and version files
-    install(
+    rocm_install(
       FILES
       "${CMAKE_CURRENT_BINARY_DIR}/${pkg_config_file}"
       "${CMAKE_CURRENT_BINARY_DIR}/${pkg_version_file}"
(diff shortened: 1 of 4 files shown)


jerryyin:
Looks reasonable. I think this PR needs at least a sanity test to actually use the make package code path. I'd say pick a MIOpen integration test and make it to install using the debian package instead of `cmake --install`

I'd abstain from the review the details but am willing to approve if:
 - @lawruble13 and @amd-aakash  think this is good
 - All CI pass, including the newly added ones that use the debian package and doesn't impact any existing functional test

@amd-aakash Once Yi shipped this feature, could you take MLIR repo for a test?

mlir/tools/rocmlir-lib/CMakeLists.txt
```diff
@@ -76,9 +83,18 @@ if(BUILD_FAT_LIBROCKCOMPILER)
...
+  rocm_install(FILES LICENSE
+    DESTINATION share/doc/${PACKAGE_NAME}
+    COMPONENT lib${LIBRARY_NAME})
+
   # Install Miir.h to ${CMAKE_INSTALL_PREFIX}/include/${PACKAGE_NAME}/
   # as part of component libroCompiler
-  install(FILES Miir.h
+  rocm_install(FILES Miir.h
     DESTINATION include/${PACKAGE_NAME}
     COMPONENT lib${LIBRARY_NAME})
```
lawruble13, line 99: Why is this being installed to this component rather than the MLSE standard packaging?

mlir/tools/rocmlir-lib/CMakeLists.txt
```diff
@@ -128,6 +144,8 @@ if(BUILD_FAT_LIBROCKCOMPILER)
     if(ARG_EXTRA_INCLUDES)
       set(EXTRA_INCLUDES ${ARG_EXTRA_INCLUDES})
     endif()
+    # Call install() instead of rocm_install() here in order to install
+    # the package in a specific direcotry. 
     install(TARGETS ${lib_name}
       EXPORT ${export_set}
       ARCHIVE DESTINATION lib/${package_name}
```
lawruble13, line 151: Why is this being installed to a subdirectory of lib?

mlir/tools/rocmlir-lib/CMakeLists.txt
```diff
@@ -171,7 +189,11 @@ if(BUILD_FAT_LIBROCKCOMPILER)
   set(extra_include_dir "")
 
   if ((NOT LLVM_INSTALL_TOOLCHAIN_ONLY) AND EXPORT_ALL_HEADERS)
-    install(DIRECTORY
+  rocm_package_setup_component(${LIBRARY_NAME}-headers
+        LIBRARY_NAME ${LIBRARY_NAME}-headers
+        PACKAGE_NAME ${PACKAGE_NAME}
+  )
```
lawruble13, line 195: Why is this `${LIBRARY_NAME}-headers` component being used? I would expect these and the header file above to simply go into the `devel` component.

mlir/tools/rocmlir-lib/CMakeLists.txt
```diff
@@ -184,7 +206,7 @@ if(BUILD_FAT_LIBROCKCOMPILER)
       PATTERN "LICENSE.TXT"
     )
 
-    install(DIRECTORY ${MLIR_INCLUDE_DIRS}
+    rocm_install(DIRECTORY ${MLIR_INCLUDE_DIRS}
       DESTINATION include/${PACKAGE_NAME}/external
       COMPONENT ${LIBRARY_NAME}-headers
```
lawruble13, line 211: As above, why can this not use the standard `devel` component?

mlir/tools/rocmlir-lib/CMakeLists.txt
```diff
@@ -221,4 +243,10 @@ if(BUILD_FAT_LIBROCKCOMPILER)
     EXTRA_INCLUDES              # extra include directories
     ${extra_include_dir}
   )
+
+  rocm_create_package(
+    NAME ${PACKAGE_NAME}
+    DESCRIPTION "MLIR packages"
+    MAINTAINER "rocMLIR"
```
lawruble13: Maintainer should include an email in angle brackets after the name.

lawruble13 (changes requested):
Thanks for prompting me to review this, I have a few questions about design decisions related to this packaging. Can somebody weigh in on why the packaging for rocMLIR differs from the standard packaging scheme?

mlir/tools/rocmlir-lib/CMakeLists.txt
yiqian1: I'm not the right person to answer this question. I'd guess probably we wanted to deliver this package separately? Otherwise it would be nice if we follow the standard MLSE packaging scheme. @jerryyin  might common on this when he's back.

mlir/tools/rocmlir-lib/CMakeLists.txt
yiqian1: @zhanglx13 Any specific reason why we used a separate component here?

zhanglx13: This is because we provide static libraries for both MIOpen and MIGraphX. These headers are required only by MIGraphX.
I'm not aware of the motivation of using rocm-cmake to create a cmake package of rocMLIR. If this is to make things more standard, maybe we should think about putting everything into the standard `devel` component.

krzysz00:
I'll note that - to my understanding - the point right now is not to make us part of the ROCm release as a standalone thing, but to create packages that MIOpen and MIGraphX can install into their QA/CI/... to avoid having to recompile us every time even though they haven't changed the commit of ours they're using

manupak:
Actually, this is one of things why we cant ship MIGraphX with MLIR by default because of a missing packaging flow.

I don't see a strong reason why we need to have two different packages.

cc: @sjw36

sjw36:
Yes, this has become a strong requirement from MIGraphX to enable the rocMLIR flow by default. This should be a rocm package for 5.6.

krzysz00:
Does this mean we should also go ahead and do thing where we rename our directory to `rocmlir/` to avoid collisions between our headers and upstream's?

sjw36:
> Does this mean we should also go ahead and do thing where we rename our directory to `rocmlir/` to avoid collisions between our headers and upstream's?

Yes, good call.

yiqian1:
I'm including @pfultz2, @JehandadKhan and @aserio for more ideas on how MIOpen and MIGraphiX want to use the rocMLIR package. 

Do we want a standard devel package or two separate packages?

Does the static library have to be installed in a subdirectory of lib?

mlir/tools/rocmlir-lib/CMakeLists.txt
jerryyin: @lawruble13 Could you clarify what the question means? What is the typical practice for `MLSE standard packaging` to install a header file that client will consume??

jerryyin:
> Does this mean we should also go ahead and do thing where we rename our directory to rocmlir/ to avoid collisions between our headers and upstream's?

I will point out that it is very close to the release time frame and changing directory name shipped from our end "seems" to break backward compatibility. If we indeed want to do that, we'll likely have to make a sym-link to the original directory as a follow-up. Thinking of this, it is probably best not to break backward compatibility now but rather do all the renaming/moving changes in a separate release, at a much earlier stage. (For example, `Miir.h` is also not a desirable name, and I'd rather not to open that rabbit hole in this work.)

> Do we want a standard devel package or two separate packages?

I don't see why not, but there's also a pro in the original (separate packages) practice. It was a defensive approach to separate between the MLIR/MIGraphX integrations as they require very different header files. My opinion probably lean towards to keep it that way. 

> Does the static library have to be installed in a subdirectory of lib?

What's the alternative?

> more ideas on how MIOpen and MIGraphiX want to use the rocMLIR package.

I think this is a good clarifying question on the scope of this work. At least to my understanding this is strictly a QA effort to make them use our package. Our client, in order to keep their component build-able as a opensource component will likely continue to use the existing practice to build and install us. So long as we are backward compatible with the old way of doing it, we should be good.

yiqian1:
>

yiqian1:
> Does the static library have to be installed in a subdirectory of lib?

> What's the alternative?

The default installation directory is lib, not its subdirectory.

yiqian1:
Thanks to all for the feedback. Per our discussion at today's team meeting, I suggest to merge this PR as is and address the standard packing issue in a separate PR.

krzysz00:
@lawruble13 
1. This isn't planned to go into ROCm packaging itself at the moment
2. However, having our library be something that goes into a `.deb` is a hard requirement for an MIGraphX integration (which improves model performance and is scheduled for 5.6)
3. Feature complete for 5.6 is next Wednesday, at least for us
4. Therefore, this PR is blocking our release

In order to not break clients so close to the branch date, we are not willing to change the installation paths of our libraries or headers at this time.

Please, with some urgency
- Break your comments into those that change installation paths or otherwise affect the build output and those that don't
- Remove your merge blocker

(Also, wrt headers, MIOpen only uses `Miir.h`, while MIGraphX uses a different set of headers, hence the two different sets of header exports - it's more a historical thing, but we can't change it this close to release)

amd-aakash:
I have no issues with the PR. We are wanting to make use of MLIR packages as build dependency for MIOpen.

krzysz00:
We may want to open a copy of this PR to dodge the hold ... although the bit about the email address probably should happen.

jerryyin:
@yiqian1 Please file a ticket to document all @lawruble13's suggestions and assign to yourself for next release. Then I can help dismiss to allow you to move forward with merge.

mlir/tools/rocmlir-lib/CMakeLists.txt
lawruble13: I am going to resolve this so that you can include this fix in the future if you are going to publish these packages. Specifically, MLSE standard packaging installs exactly and only *shared* libraries into the runtime component, named for the library (in this case it would be named `rocMLIR`), and any header files or *static* libraries should be installed into the devel component, named either `<library>-dev` for debians or `<library>-devel` for RPMs. This behaviour can be accomplished by not specifying any component, as rocm-cmake automatically installs these items to the correct components.

lawruble13:
Per @krzysz00 above, I am resolving my comments and removing the merge blocker, with those changes to be included in the next release, as the packaging of rocMLIR is not going to be included in ROCm for this release.

lawruble13 (approved):
Noting that the changes I mentioned above should be included in an upcoming release, this looks fine to me as it's not being publicly released.

yiqian1:
The new ticket is https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/847.

PR #1003, opened 2023-03-06
Update Dockerfile to install rocm-cmake

yiqian1:
Will use rocm-cmake to generate MLIR packages.

### DIFF
--- a/mlir/utils/jenkins/Dockerfile
+++ b/mlir/utils/jenkins/Dockerfile
@@ -119,11 +119,6 @@ RUN apt-get update && rm /usr/share/keyrings/kitware-archive-keyring.gpg && \
   apt-get clean && \
   rm -rf /var/lib/apt/lists/*
 
-# Install the new rocm-cmake version
-RUN git clone -b master https://github.com/RadeonOpenCompute/rocm-cmake.git  && \
-  cd rocm-cmake && mkdir build && cd build && \
-  cmake  .. && cmake --build . && cmake --build . --target install
-
 # --------------------- Section 3: MIOpen dependencies ---------------
 WORKDIR /MIOpenDepsWork
 RUN wget https://raw.githubusercontent.com/ROCmSoftwarePlatform/MIOpen/develop/requirements.txt \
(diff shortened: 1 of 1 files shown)


mlir/utils/jenkins/Dockerfile
```diff
@@ -118,6 +119,11 @@ RUN apt-get update && rm /usr/share/keyrings/kitware-archive-keyring.gpg && \
   apt-get clean && \
   rm -rf /var/lib/apt/lists/*
 
+# Install the new rocm-cmake version
```
jerryyin: Why is this necessary? I think just get the rocm-cmake from default ROCm distro is fine?

yiqian1: This is what CK does, probably because the default ROCM doesn't include some newest updates.

jerryyin: I just took a look at the master branch, it is 5 months old now. `develop` branch is actually newer than it. I'd say get rid of those line and give it a shot to see this is really necessary on our end.

krzysz00: I think what's going on here is that `rocm-cmake` has the CMake files you need to do a ROCm build, which aren't installed by default because most people don't need them?

jerryyin: I may be wrong but I thought its exported CMake files are the key to make it useful to compile other components.

The worry I have is that we'd always install a stale rocm-cmake with the manual steps, and hardly notice it until it implicitly break us in the ROCm upgrade.

lawruble13: `rocm-cmake` doesn't get frequently updated, and these days we're making a strong effort to ensure that it's backwards compatible. The version of `rocm-cmake` installed with ROCm 5.4 (which is `rocm-cmake` 0.8.0 IIRC) should be sufficient for you folks.

krzysz00: Ok, having looked at the context:

1. Why are we manually installing from Git here? What's wrong with the apt-get-able package from the latest release?
2. If you look at `sudo dpkg -L rocm-cmake`, you'll see it's all stuff you need for ROCm packaging

yiqian1: OK, I'll remove the manual build and installation. Thanks!

jerryyin (approved):
LGTM

PR #1004, opened 2023-03-09
Fixes tuning key vs perfRunner arg mismatches

manupak:
This commit fixes the issue the tuning hooks in rocMLIR generates args that are not consistent with perfRunner.py that is used to tune the said problem config.

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/825

### DIFF
--- a/mlir/utils/performance/tuningRunner.py
+++ b/mlir/utils/performance/tuningRunner.py
@@ -226,7 +226,7 @@ def main(args=None):
     if parsed_args.config:
         configs = parsed_args.config
     elif opType == Operation.CONV:
-        configs = perfRunner.getConvConfigurations(paths.configuration_file_path, do_sweep=False)
+        configs = perfRunner.getConvConfigurations(paths.configuration_file_path)
     elif opType == Operation.GEMM:
         configs = perfRunner.getGemmConfigurations(paths.configuration_file_path)
     else:
--- a/mlir/utils/performance/tuningRunner.py
+++ b/mlir/utils/performance/tuningRunner.py
@@ -226,7 +226,7 @@ def main(args=None):
     if parsed_args.config:
         configs = parsed_args.config
     elif opType == Operation.CONV:
-        configs = perfRunner.getConvConfigurations(paths.configuration_file_path)
+        configs = perfRunner.getConvConfigurations(paths.configuration_file_path, do_sweep=False)
     elif opType == Operation.GEMM:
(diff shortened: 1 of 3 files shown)


mlir/utils/performance/tuningRunner.py
```diff
@@ -226,7 +226,7 @@ def main(args=None):
     if parsed_args.config:
         configs = parsed_args.config
     elif opType == Operation.CONV:
-        configs = perfRunner.getConvConfigurations(paths.configuration_file_path)
+        configs = perfRunner.getConvConfigurations(paths.configuration_file_path, do_sweep=False)
```
krzysz00: I don't think is the right behavior? The resnet50 file is set up to expect a sweep?

Or at least add a flag?

manupak: flag you mean in the config file ? 
(whether to do a sweep)

I ll describe the problem:
* In the MIGraphX tuning prototype it relies on tuning key to extract exact configs that needs to be tuned.
* Since the tuning key already has layout and friends already populated, they wont work as they work with gemm s out of the box.
* In order to not to overtune (i.e. not tune more than what client asks for) -- we might need a switch.

Having said that, did you mean a flag to indicate a sweep in the config line ?
OR I could do a check to see what is already present and do the sweep for rest ?

krzysz00: What I'm saying is that `mlir/utils/jenkins/miopen-tests/resnet50-miopen-configs` (the default config file) doesn't indicate fwd/bwd/..., and the convolution tuning flow takes that config and expects that you run a sweep over it.

manupak: Yes I gathered that. 

Maybe the solution is to do a sweep on parameters that are missing in the config line in a such file ? WDYT?

manupak: (rather than assuming its not there)

krzysz00: Yeah, I think sweeping on unspecified parameters is the right call - I _think_ we do that for GEMM performance tuning, for example.

manupak: yea -- it is correctly handled in gemm tuning.. I've did the same for conv now.

manupak:
@krzysz00 thanks for the reviews! I've addressed the comment.

krzysz00 (approved):
This all looks reasonable to me

manupak:
Thanks @krzysz00 !

PR #1005, opened 2023-03-09
Adding QuantConvolution and consolidate with MIGraphX Conv

jerryyin:
In MIGraphX, a default op doesn't allow input and output type to be different. Therefore below is invalid MIGraphX IR:

```
%1 = migraphx.convolution(%arg1, %arg2) : (tensor<1x128x56x56xi8>, tensor<128x128x3x3xi8>) -> tensor<1x128x28x28xi32>
```

I created the QuantConvolution Op and make it consolidate with the original convolution Op. This allow our MIGraphX `quant_convolution` to be captured and lowered correctly in our end.

### DIFF
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -66,10 +66,9 @@ static tosa::CastOp createCastOp(PatternRewriter &rewriter, Location loc,
   return op;
 }
 
-class ConvConverter final
-    : public OpConversionPattern<migraphx::ConvolutionOp> {
+template <typename T> class ConvConverter : public OpConversionPattern<T> {
 public:
-  using OpConversionPattern<migraphx::ConvolutionOp>::OpConversionPattern;
+  using OpConversionPattern<T>::OpConversionPattern;
 
   Value getZeroBias(Location loc, Type elemType, int64_t filterOutputChannels,
                     ConversionPatternRewriter &rewriter) const {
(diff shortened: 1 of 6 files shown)


mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -99,16 +98,15 @@ class ConvConverter final
   }
 
   LogicalResult
-  matchAndRewrite(migraphx::ConvolutionOp op, OpAdaptor adaptor,
-                  ConversionPatternRewriter &rewriter) const final {
-    auto operands = adaptor.getOperands();
+  matchAndRewrite(T op, typename OpConversionPattern<T>::OpAdaptor adaptor,
```
manupak: first, I agree with your attempt to avoiding code duplication.

Please put a comment this will be used only with a convolution and possibly with a todo to use a base class (i.e. an interface) as it becomes hard to read what this conversion pattern is converting now.

(Im fine with this for now but however it would be better to introduce a MIGraphXConvInterface rather than a generic template -- if you agree, lets put a todo with a ticket do the interface later.)

manupak (commented):
Overall looks good.

I just feel conversion pattern is not that generic to be using a generic template but fine for now.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jerryyin: Good point, I think I can change T to be ConvType and add a comment now to indicate what this conversion pattern targets at.

As far as the quant_convolution, I don't think we need to generalize it more: The tablegen type field captured the difference between the two ops. It is straightforward to tell for the existing implementation.

Giving a comparison with what we did we rock conv2d ops, we have both practice available:
 - In Conv2DRewritePattern, it is using a templatized manner to do it, very similar with here. This is because we need to materialize each instance of class for that particular lowering type. 
 - In MatchLayoutsToInput pattern, it is using `RockConvInterface`. In this creating class instance copies would be create duplicated compilation unit therefore we should be using a common interface to address it.

I'm hoping my generalization of the two cases are right. Now with that assumption, in the quant_convolution case we indeed need to materialize the two independent lowering types. Therefore it is reasonable to pick the first practice. I'd agree with us creating a ticket only when the second scenario exists: That there comes a use case where a lowering pipeline to handle this group of conv ops.

manupak (approved):
LGTM!

> As far as the quant_convolution, I don't think we need to generalize it more:

This is right. I was actually commenting to narrow it down by not using a template as this will act as a fallback for any op migraphx dialect -- should there not be an implementation.

> In MatchLayoutsToInput pattern, it is using RockConvInterface. In this creating class instance copies would be create duplicated compilation unit therefore we should be using a common interface to address it.

I feel this is the right way to go in the long run.
to be precise:

```mlir
class ConvConverter : public OpInterfaceConversionPattern<MIGraphXConvInterface> {
```

jerryyin:
> I feel this is the right way to go in the long run... OpInterfaceConversionPattern<MIGraphXConvInterface>

I keep thinking that interface is only worth the effort only if the two ops has a subset of similarities that we want to generalize about. As of right now, since two ops are effectively duplication of each other, creating interface will not give us much benefit, but cause more duplications because we need to put every op invocation into the interface. (Implementation wise, creating a common base class is probably cleaner than a `OpInterface`, but compared with current templatized implementation it'd be only style preferences over on another.) I think we are on the same page that we'll leave this as-is and reevaluate when two ops diverges, that would justify an interface.

PR #1006, opened 2023-03-13
Adds tosa.reduce_max based fusion support

manupak:
This commit implements tosa-level hookups
for reduce max operations.

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/776


mlir/test/Conversion/TosaToRock/reductions/tosa-to-rock-reduce-e2e.mlir
```diff
@@ -1,9 +0,0 @@
...
-// cat %s | rocmlir-driver -host-pipeline xmodel -kernel-pipeline full -targets %arch | xmir-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_c_runner_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CLONE
-// CLONE: [1 1 1]
-// CLONE-NEXT: Unranked Memref base
-
-func.func @test_basic(%arg0: tensor<200x100xf32>) -> tensor<200x1xf32> {
-  %1 = "tosa.reduce_sum"(%arg0) {axis = 1 : i64} : (tensor<200x100xf32>) -> tensor<200x1xf32>
```
manupak, line 7: Im removing this for now. It has become tricky do hookup to tosa to get standalone reductions to work in the absence of copy opt as a separate pass. We can revisit them, if go on to optimize the standalone kernel -- i'd think we would want to have the copy opt seperate pass run after align tiling.

For now, we are good for fusions with gemm-like ops.

krzysz00 (commented):
While we're here, could you
1. Add support for the prefill attr to the testing flow in `rocmlir-gen`
2. Update the zero_init_kernel to a constant_init_kernel

manupak:
> While we're here, could you
> 
> 1. Add support for the prefill attr to the testing flow in `rocmlir-gen`

We dont use rocmlir-gen in testing fusion kernels, hence I dont see the relevance in this particular PR.

> 2. Update the zero_init_kernel to a constant_init_kernel

Likewise, we dont generate init_kernels (yet) for fusion flows.

Generally they seem good suggestions -- Maybe we should create tickets and discuss the priority ?
I dont see the need or relevance for them to be done here.

krzysz00:
I thought we were using rocmlir-gen's clone to do the testing?

manupak:
> I thought we were using rocmlir-gen's clone to do the testing

If you meant, rocmlir-gen reading the prefill attr and inserting linalg fills before async launch, that is there. See below : 

https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/af89ecb51db6fba8150c1125ddfd2d87ca158392/mlir/tools/rocmlir-gen/rocmlir-gen.cpp#L2395-L2396

I thought you were asking for plain CLI args to generate a harness from scratch.

krzysz00:
... Ah, I somehow missed that we already had that.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -426,18 +426,70 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
...
+  Optional<uint32_t> num_cu;
+  rock::GemmFeatures features;
+  std::tie(arch, num_cu, features) = getArchAttributes(op);
+
+  int32_t blockSize = 256;
+  auto elementCount =
+      op.getInput().getType().template cast<ShapedType>().getNumElements();
+  int32_t gridSize = (elementCount + blockSize - 1) / blockSize;
+  if (num_cu.has_value()) {
+    gridSize = std::max((int32_t)(20 * num_cu.value()), gridSize);
```
krzysz00: Shouldn't this, perhaps, be a min()?
Otherwise we're launching redundant work?

krzysz00 (commented):
Overall, this seems reasonable.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
manupak: you are right! fixed now.

krzysz00 (approved):

manupak:
Thanks @krzysz00 !

PR #1008, opened 2023-03-14
Phase 1 of making the tosa-partition pass into a generic utility.

pcf000:
Simplify isConstantZero() and make it more general.

Restructure partition walk to collect anchors then process them back-to-front, and drop transpose special case.

Update rocmlir tests that are perturbed by the partitioning rearrangement.

Rearranged the code to make a stub TosaPartitionPass with all the action in a tosa-free Partition class.  Adjusted one test for different behavior with tosa::transpose.

This is a state that incorporates most of Simon's pull request, but not the traceInputs() change to the partitioner.  It does pass all the Dialect/Tosa tests. In sum:  generic partitioning, Tosa implementation, some of Simon's changes.

Tweak traceInputs() and use it on trailing-op operands like Simon does, and some cleanup.

First cut at separating the partitioner into a separate utility, used by tosa-partition.

IS NOT:  partition passes for dialects other than Tosa;  terminating ops as a special class (they can still be addressed in the trailing-op predicate).

### DIFF
--- a/external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
+++ b/external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
@@ -1,0 +1,501 @@
+//===- OutlinerUtils.cpp ------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// Replace conv2d followed by elementwise op with call to function containing
+// them.  Generalised, outline any anchor op, all its trailing elementwise ops,
+// and all its leading elementwise ops.
+//
+//===----------------------------------------------------------------------===//
+
+#include "mlir/Transforms/OutlinerUtils.h"
+#include "mlir/Analysis/SliceAnalysis.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "mlir/IR/Dominance.h"
+#include "mlir/IR/IRMapping.h"
+#include "mlir/IR/Matchers.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Pass/PassManager.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "mlir/Transforms/Passes.h"
+#include "mlir/Transforms/RegionUtils.h"
+#include "mlir/Transforms/TopologicalSortUtils.h"
+#include "llvm/ADT/DenseSet.h"
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SetVector.h"
+#include "llvm/ADT/SmallVector.h"
+#include <algorithm>
+#include <deque>
+#include <iostream>
+
+using llvm::SmallVector;
+using namespace mlir;
+
+static bool isZeroAttribute(Attribute value) {
+  if (auto intValue = value.dyn_cast<IntegerAttr>())
+    return intValue.getValue().isNullValue();
+  if (auto fpValue = value.dyn_cast<FloatAttr>())
+    return fpValue.getValue().isZero();
+  if (auto splatValue = value.dyn_cast<SplatElementsAttr>())
+    return isZeroAttribute(splatValue.getSplatValue<Attribute>());
+  if (auto elementsValue = value.dyn_cast<ElementsAttr>())
+    return llvm::all_of(elementsValue.getValues<Attribute>(), isZeroAttribute);
+  if (auto arrayValue = value.dyn_cast<ArrayAttr>())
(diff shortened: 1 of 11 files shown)


external/llvm-project/mlir/test/Dialect/Tosa/tosa-partition.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: mlir-opt --split-input-file --tosa-partition %s -verify-each=0 -o - | FileCheck %s
+// RUN: mlir-opt --split-input-file --tosa-partition='trailing-only=false' %s -verify-each=0 -o - | FileCheck %s
```
krzysz00, line 1: Why verify-each=0?

external/llvm-project/mlir/test/Dialect/Tosa/tosa-partition.mlir
```diff
@@ -88,10 +88,10 @@ func.func @test_fusion6(%arg0: tensor<128x8x32x32xf32>, %arg1: tensor<128x8x3x3x
 }
 
 // CHECK-LABEL: func private @test_fusion7__part_0
+// CHECK-NEXT: tosa.abs
```
krzysz00, line 91: Why'd the abs() move before the conv2d here? Also, we don't currently have load-side fusion in RocMLIR ... but maybe we won't hit this case by setting trailing-only in our pipelines.

krzysz00 (commented):
I've got some initial questions

external/llvm-project/mlir/test/Dialect/Tosa/tosa-partition.mlir
pcf000: Because I blindly copied the options from the test I started with.  Seriously, though, I was just following what was there.

external/llvm-project/mlir/test/Dialect/Tosa/tosa-partition.mlir
pcf000: I'm running it with trailing-only=false, which was the default until recently.  That gathers the abs into the leadingOps for the conv2d and thus includes it in the __part function.

Simon's recent patch changed the default because it fit his larger cases better;  I still wanted to test the leading-ops functionality in these little tests.

pcf000:
Renamed the partitioner utility to "outliner."

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
```diff
@@ -0,0 +1,510 @@
...
+          OperationEquivalence::Flags::IgnoreLocations))
+    return false;
+
+  // Compare the operands of the two operations. If the operand is within
+  // the block, it must refer to the same operation.
+  auto lhsOperands = lhs->getOperands(), rhsOperands = rhs->getOperands();
+  if (lhs->getNumOperands() != rhs->getNumOperands()) {
+    return false;
+  }
+  for (int operand : llvm::seq<int>(0, lhs->getNumOperands())) {
```
sjw36: use llvm::zip.

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
```diff
@@ -0,0 +1,510 @@
...
+
+OutliningCandidate::OutliningCandidate(Operation *anchorOp_,
+                                       ArrayRef<Operation *> &trailingOps_,
+                                       ArrayRef<Operation *> &leadingOps_,
+                                       ArrayRef<Value> &params_,
+                                       ArrayRef<Value> &returnVals_,
+                                       StringRef partFnName_)
+    : anchorOp(anchorOp_), partFnName(partFnName_), hash(0), function(nullptr) {
+  // We'll need to grab the cloned ops to avoid use-after-free.
+  for (auto *op : trailingOps_) {
```
sjw36: Should be a copy from ArrayRef to SmallVector.

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
```diff
@@ -0,0 +1,510 @@
...
+static bool outliningCandidatesEquivalent(OutliningCandidate &one,
+                                          OutliningCandidate &two) {
+  if (one.hash != two.hash) {
+    return false;
+  }
+
+  if (one.params.size() != two.params.size()) {
+    return false;
+  }
+  for (unsigned i = 0; i < one.params.size(); i++) {
```
sjw36: could also use llvm::zip

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
```diff
@@ -0,0 +1,510 @@
...
+    // Don't outline a kernel;  it may already have been outlined.
+    if (func->hasAttr(outlineTag))
+      continue;
+
+    std::vector<Operation *> anchors;
+    auto callback = [&](Operation *op) {
+      if (isAnchorOp(op))
+        anchors.push_back(op);
+    };
+    // Gather the anchor ops so we can process them back-to-front.
```
sjw36, line 405: Great approach.

sjw36 (approved):
Looks good. Just some minor suggestions.

sjw36:
For trailingOpsOnly == false, we may need 2 passes to first collect all successors, plus predecessors that are transposes. Then another pass to fuse any remaining predecessors. This would require building all the kernel funcs, then going back and modifying them before looking for equivalence.

PR #1009, opened 2023-03-14
Add a feature for the presence of the gfx940 MFMAs

krzysz00:
This feature currently isn't hooked up to anything, but at least it's there so we can use it in the future.

### DIFF
--- a/mlir/lib/Dialect/Rock/Generator/AmdArchDb.cpp
+++ b/mlir/lib/Dialect/Rock/Generator/AmdArchDb.cpp
@@ -18,6 +18,8 @@ using namespace mlir::rock;
 static constexpr AmdArchInfo gcnInfo(GemmFeatures::none, /*waveSize=*/64),
     cdnaInfo(GemmFeatures::mfma | GemmFeatures::dot | GemmFeatures::atomic_add,
              /*waveSize=*/64),
+    cdna3Info(cdnaInfo.defaultFeatures | GemmFeatures::gfx940_mfma,
+              /*waveSize=*/64),
     rdnaNoDotInfo(GemmFeatures::atomic_fmax_f32, /*waveSize=*/32),
     rdnaInfo(GemmFeatures::dot | GemmFeatures::atomic_fmax_f32,
              /*waveSize=*/32),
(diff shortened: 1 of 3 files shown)


manupak:
Overall changes are fine!.

Just naming question : why not use mfma_i8 and mfma_xf32 ? 

if not, Is it more of a random assortment of intrinsics?

krzysz00:
It's very much a random assortment. We have FP8, changes to I8, XF32, a bunch of old BF16's being deprecated ...

mlir/lib/Dialect/Rock/Generator/AmdArchDb.cpp
```diff
@@ -39,7 +41,8 @@ AmdArchInfo mlir::rock::lookupArchInfo(StringRef arch) {
   StringRef major = chip.slice(0, chip.size() - 2);
   if (major == "gfx9") {
     return llvm::StringSwitch<AmdArchInfo>(minor)
-        .Cases("08", "0a", "40", cdnaInfo)
+        .Cases("08", "0a", cdnaInfo)
+        .Case("40", cdna3Info)
```
manupak, line 45: Please mirror this change in mlir/test/lit.site.cfg.py.in:set_arch_features()

manupak (commented):
Please adjust lit cfg accordingly. 

Re: naming,
oh dear -- I dont have a better solution either given the bit enum format of features we use.
(An alternative might've been to use a hierarchical target structure -- but thats a larger change)

mlir/lib/Dialect/Rock/Generator/AmdArchDb.cpp
krzysz00: Right, good point.

Part of me is thinking we want to bring the arch string back ...

Ok, yeah, actually, given the _rest_ of the MFMA changes, I'm bringing the arch string back.

jerryyin:
I'm confused about this PR. It sounds like in #1013 you have already decided to add the architecture field in the mfma selection. So why not go one step further and encode it in the mfma map itself? Then at selection time it will be straightforward to check if the mfma is supported by the architecture encoded in the map entry.

The approach you are taking in this PR seems to suggest we'll represent all those in a separate way using the high level gemm features. In my opinion, it goes in conflict with #1013.

krzysz00:
Yeah, this is supersceded by #1013 so closing

PR #1011, opened 2023-03-15
Allow rock.buffer_{load,store} to support scalar memrefs

giuseros:
Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/827

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/AMDGPU.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/AMDGPU.td
@@ -87,7 +87,7 @@ def AMDGPU_RawBufferLoadOp :
   let assemblyFormat = [{
     attr-dict $memref `[` $indices `]`
       (`sgprOffset` $sgprOffset^)? `:`
-      type($memref) `,` type($indices) `->` type($value)
+      type($memref) (`,` type($indices)^)? `->` type($value)
   }];
   let hasCanonicalizer = 1;
   let hasVerifier = 1;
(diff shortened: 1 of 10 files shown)


giuseros:
@sjw36 the `rocmlir-driver` command works now. Is there a way I can test it runs correctly (all tests - PR+nightly- pass)?

@krzysz00 Is there something more to be done here? I guess those guards in the `verify` function were there for a reason? But I wasn't able to understand if there is anything else to be changed.

giuseros:
Ok, if I run `rocmlir-driver -c test.mlir` I get a segfault . I think this is related with this change :)

sjw36:
You should be able to use one of the fusion tests and change the bias for instance to be passed as tensor<f32>. The e2e tests will validate the results with the cpu result.

krzysz00 (approved):
If we've confirmed that this does actually lower to something sensible, then yeah, approved.

krzysz00:
What's the segfault?

giuseros:
Ok, there were few things to change to make it work. I added few tests and I would like to verify it really works with a final integration test (in chat with @sjw36  to come up with one).

krzysz00 (approved):
Still approved, but if you could do some `git rebase -i` surgery to get this into two commits, one that's the external changes and one that's the internal ones, that'd be good - it'll make this easier to send this upstream

giuseros:
Ok, I split the commit in three (llvm, rocm, integration). With the help of @sjw36  added also an integration test to check that everything works fine when we use scalar tensors.

krzysz00:
LGTM

sjw36 (approved):
Looks good.

PR #1013, opened 2023-03-16
Make MFMA selection chipset-dependent

krzysz00:
Whereas
- The upcoming gfx940 deprecates certain int8-related instructions, namely the 32x32x8 and 16x16x16 ones, in favor of the 32x32x16 and 16x16x32 ones, respectively, and;
- New, more efficient bfloat16 operations were introduced in gfx90a, and the older versions have been removed in gfx940, and;
- FP8-based MFMAs are only introduced in gfx940, and the future PR to add support for FP8 should be able to reject requests for these kernels on older architectures, and;
- Trying to add features describing all this variation really won't buy us much, as per discussion on #1009 

This PR
- Threads the `arch` attribute back through `gridwise_gemm_v2`, `blockwise_gemm_v2`, and `xdlops_gemm_v2`
- Extends the MFMA instruction group selection system to accept an `arch` parameter
- Uses the passed-in `arch` to choose the sets of tables that will be queried for MFMA groups
- Extends the tuning info structure with an `arch` field
- Cleans up some of the MFMA database implementation to avoid pointless copies
- Adds tests

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
+++ b/mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
@@ -124,8 +124,8 @@ private:
   MfmaInsnGroupAttr groupAttr;
 
 public:
-  static FailureOr<MfmaInsnGroup> select(Type elementType, int64_t mPerWave,
-                                         int64_t nPerWave);
+  static FailureOr<MfmaInsnGroup> select(Type elementType, StringRef arch,
+                                         int64_t mPerWave, int64_t nPerWave);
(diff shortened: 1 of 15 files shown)


manupak (approved):
I dont have major concerns here given past conversations.

I am trusting you that you got mfma assortments correct per arch.

krzysz00:
@manupak If you'd like to double-check my work, the definitive table for what MFMAs can be used where is at https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/AMDGPU/VOP3PInstructions.td#L626 and down (`is_gfx940_xdl` there is just related to scheduling and you can ignore it)

PR #1014, opened 2023-03-17
[TRIVIAL] fixes overtuning of duplicates

manupak:
Currently, we tend to retune the same config
in conv case. this commits fixes that.

### DIFF
--- a/mlir/utils/performance/perfRunner.py
+++ b/mlir/utils/performance/perfRunner.py
@@ -263,7 +263,8 @@ def getConvConfigurations(fileName):
                     output_layout = ""
 
                 oneConfig = f"{datatype}{direction}{filter_layout}{input_layout}{output_layout}{line}"
-                configs.append(oneConfig)
+                if oneConfig not in configs:
+                    configs.append(oneConfig)
     return configs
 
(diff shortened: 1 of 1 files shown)


krzysz00 (approved):
Was this actually happening?

Also, efficiency note:
- Make a `configsSet` that's the same as `configs()` but a `set()` so that we can get `O(1)` membership checks instead of `O(N)`
- But keep the list so that we still have a deterministic order.

That aside, approved

manupak:
Yes, I was scratching my head why a single conv takes so much time to tune, especially when all things specified (dir, datatype, etc).

I acknowledge the efficiency note (also yours as a solution for that) but I have an alternative approach for that which I'd prefer to do in a seperate PR. Also gemms have same efficiency concern as well.

I think a slightly cleaner approach would be to take lines out of the itertools product space and dynamically construct the product space based on the presence of args in the line. So that we dont have do these checks because sweep itself will be done on a smaller product space that is only required. WDYT ?
(if you agree, I ll create a different ticket for that)

krzysz00:
Yeah, dynamically constructing the product space based on what the line doesn't specify makes a lot of sense, I think that'd make a good separate PR.

PR #1017, opened 2023-03-21
[CI][MIGraphX] add explicit driver target and gpu backend

manupak:
Make the tests build the driver and explicitly
use gpu backend (not relying on the default).

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -972,10 +972,10 @@ pipeline {
                                 timeout(time: 60, activity: true, unit: 'MINUTES') {
                                     withEnv(['MIGRAPHX_ENABLE_MLIR=1']) {
                                         // Verify MLIR unit tests
-                                        sh 'make -j$(nproc) test_gpu_mlir'
+                                        sh 'make -j$(nproc) driver test_gpu_mlir'
                                         sh 'ctest -R test_gpu_mlir'
                                         // Verify ResNet50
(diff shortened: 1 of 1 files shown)


umangyadav (dismissed):

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -972,10 +972,10 @@ pipeline {
                                 timeout(time: 60, activity: true, unit: 'MINUTES') {
                                     withEnv(['MIGRAPHX_ENABLE_MLIR=1']) {
                                         // Verify MLIR unit tests
-                                        sh 'make -j$(nproc) test_gpu_mlir'
+                                        sh 'make -j$(nproc) driver test_gpu_mlir'
```
umangyadav, line 975: `make driver` wouldn't be necessary.

`driver` would get built as part of the `buildMigraphx()` call.

manupak: Just to confirm, we can rely on test_gpu_mlir target for that ?

pfultz2: `test_gpu_mlir` wont build the driver. You need to build it explicitly.

pfultz2 (approved):

mlir/utils/jenkins/Jenkinsfile
umangyadav: @pfultz2 https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/3116b350d684f3aefbce98b154f737f8610095c8/mlir/utils/jenkins/Jenkinsfile#L52 

This would build `driver`. I don't see need to add it  explictly.

PR #1018, opened 2023-03-21
[DO NOT MERGE] Use COMgr to compile our GPU binaries

krzysz00:
Since we're considering shipping as a ROCm package, I've investigated if it's possible to use COMgr instead of shipping our own copy of LLVM.

This branch shows that it's possible, but I'm not convinced we want to merge it.

The good news is that the setup works.

The bad news is that, because our LLVM copy is *newer* than the one shipped in ROCm 5.4 (which is what we're using for our CI) we'll sometimes need to **downgrade** the LLVM IR we produce to be acceptable input to ROCm's LLVM. If theirs were newer, it'd be fine, because LLVM has strong backwards compatibility guarantees, but this is the other direction.

On top of that, I had to do a few ugly hacks to make COMgr do the compiler invocations we want.

So ... yeah.

(I also should add some tests to make sure printf() works, but that's a tomorrow me problem - I wanted to get this PR out)

### DIFF
--- a/mlir/lib/RocmlirTransforms/CMakeLists.txt
+++ b/mlir/lib/RocmlirTransforms/CMakeLists.txt
@@ -13,6 +13,9 @@ endif()
 add_rocmlir_dialect_library(RocMLIRTransforms
   CompileGpuWithComgr.cpp
 
+  ADDITIONAL_HEADER_DIRS
+  ${MLIR_MAIN_INCLUDE_DIR}/mlir/RocmlirTransforms
+
   LINK_COMPONENTS
   ${COMGR_PASS_COMPONENTS}
 
--- a/mlir/lib/RocmlirTransforms/CMakeLists.txt
+++ b/mlir/lib/RocmlirTransforms/CMakeLists.txt
@@ -1,0 +1,40 @@
+# Instead of using the LLVM and LLD living next to MLIR, use the COMGR interface
+# to invoke the versions of these tools shipped with ROCm. This is used to
+# avoid packaging a separate copy of the compiler when creating MLIR-based tools
+# targetting ROCm, espcecially those living inside the ROCm distribution itself.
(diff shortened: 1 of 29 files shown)


krzysz00:
@jerryyin Can you ping in some MIOpen and PyTorch folks to clarify what their requirments wrt COMgr usage are?

mlir/lib/RocmlirTransforms/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
+# Instead of using the LLVM and LLD living next to MLIR, use the COMGR interface
+# to invoke the versions of these tools shipped with ROCm. This is used to
+# avoid packaging a separate copy of the compiler when creating MLIR-based tools
+# targetting ROCm, espcecially those living inside the ROCm distribution itself.
+if (ROCMLIR_ENABLE_COMGR)
+  find_package(amd_comgr REQUIRED PATHS /opt/rocm)
```
jerryyin, line 6: I suspect this won't work with pytorch's build environment. It'd be good if you can grab their docker and give it a test.

krzysz00:
If it helps, MIOpen's COMgr constaint is
```
set_var_to_condition(MIOPEN_USE_COMGR_DEFAULT (NOT DEFINED MIOPEN_BACKEND_OPENCL) AND (NOT (MIOPEN_BACKEND STREQUAL "HIPNOGPU")))
option(MIOPEN_USE_COMGR "Use comgr to build kernels instead of offline tools" ${MIOPEN_USE_COMGR_DEFAULT})
```
and so it defaults to on for us

krzysz00:
@jithunnair-amd and @JehandadKhan for comment on what, exactly, PyTorch's requirements about what parts of ROCm we can use are

krzysz00:
skip-ci for clang-tidy weirdness, but also for uncertainty about requirements.

PR #1020, opened 2023-03-22
Improve handling of contexts in the MIOpen integration

krzysz00:
Currently, our MIOpen integration uses one MLIR context for all compilations (!!!), which is why we have to use a mutex to ensure we don't have parallel compilations. This isn't how contexts are supposed to work - you want one context per independent unit of compilation. In fact, upstream has explicit examples for how to have a compilation server that re-uses a thread pool between contexts (to prevent each context from allocating a thread pool) but using a different context for each compilation.

This PR moves us to one-context-per-handle, which will, in the future (probably after 5.6 branch) allow us to drop the mutex from the MIOpen integration.

### DIFF
--- a/mlir/tools/rocmlir-lib/rocmlir-lib.cpp
+++ b/mlir/tools/rocmlir-lib/rocmlir-lib.cpp
@@ -3,6 +3,8 @@
 #include "mlir/Dialect/Rock/Pipelines/Pipelines.h"
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/BuiltinOps.h"
+#include "mlir/IR/Location.h"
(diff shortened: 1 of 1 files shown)


krzysz00:
On that note, @pfultz2 for whether (probably in 6.0) there are similar changes to context creation we may want to make over in MIGraphX

jerryyin (approved):
LGTM

pfultz2:
> On that note, @pfultz2 for whether (probably in 6.0) there are similar changes to context creation we may want to make over in MIGraphX

We already use a context for each compilation unit. I believe there was still issues with parallel compilation.

PR #1022, opened 2023-03-27
Fix globalLoadOp lowering to not use vector magic if it isn't needed

giuseros:
@krzysz00 we discussed this fix last week. While overall it does not bring performance uplift, it makes some configurations work better :
For instance, consider this sequence of commands:
```
$ echo "./bin/rocmlir-gen -ph -operation gemm --arch gfx90a -t i8 --transA --transB -g 64 -m 1024 -k 1024 -n 384 --perf_config=128,64,4,64,32,16,1,1 | ./bin/rocmlir-driver -c  |  ../mlir/utils/widgets/rocm-run" > toprofile.sh
$ rocprof --stats toprofile.sh && cat results.stats.csv
```
With current `rocMLIR`

```
"rock_gemm.kd",5,3917131,783426,100.0
```

With this PR:
```
"rock_gemm.kd",5,2332965,466593,100.0
```
So, it is about 84% improvement. Again, when tuned there is very little improvement, because the tuner is smart and selects a configuration that gives about the same performance.


krzysz00 (approved):
Looks good to me, approved!

PR #1023, opened 2023-03-27
 Use an intermediate buffer to transpose data from global memory to LDS

giuseros:
This is implementing the change proposed in https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/837

Unfortunately the performance improvements are not as good as expected. 

However:
- There is still a minimal speed-up in some cases
- There are minimal regressions
- We now have the high level code very similar to CK. The tuner indeed selects (in many cases) the same perf configs, so future performance improvements should be simpler after this change.

I am attaching the usual perf comparison (before/after) to this PR

[transpose.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/11081154/transpose.zip)

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
@@ -345,15 +345,23 @@ createGlobalLoadLoop(PatternRewriter &b, Location loc, GpuAllocOp loadBuffer,
   return outerLoop;
 }
 
-void copyAndTransposeFromLoadToStoreBuffer(
-    PatternRewriter &b, Location loc, Type elementType, GemmDimension vectorDim,
-    int64_t kpack, Value loadBuffer, Value storeBuffer, int64_t copyDPerThread,
-    int64_t copyKPerThread) {
+/// This function pack the load buffer into a store buffer ready to be copied
+/// into LDS:
+///  - The load buffer is a KPerThread x DPerThread or DPerThread x KPerThread
+///  depending
+///    on the direction of the (global vectorization)
+///  - The store buffer needs to be packed as a [KOuterPerThread, dPerThread,
+///  kpackPerThread]
+///    buffer
+TransformingForOp
+packLoadBufferToStoreBuffer(PatternRewriter &b, Location loc, Type elementType,
+                            GemmDimension vectorDim, int64_t kpack,
+                            Value loadBuffer, Value storeBuffer,
+                            int64_t copyDPerThread, int64_t copyKPerThread) {
 
-  // Copy loadBuffer[i,j] into storeBuffer[j,i]. This can
-  // be implemented in multiple ways, but for now we stick
-  // with a scalar loop that goes through the buffer and
-  // copies elements one by one
+  bool transpose =
+      ((vectorDim == GemmDimension::K && kpack == 1 && copyDPerThread > 1) ||
+       (vectorDim == GemmDimension::MorN && kpack > 1 && copyKPerThread > 1));
 
   SmallVector<StringRef, 2> loadBufferNames;
   SmallVector<int64_t, 2> loadBufferShape;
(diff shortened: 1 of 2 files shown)


mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -334,43 +330,84 @@ static TransformingForOp createGlobalLoadLoop(PatternRewriter &b, Location loc,
       ArrayRef<Attribute>{matrixToTensor, b.getArrayAttr({})},
       /*bounds=*/ArrayRef<int64_t>{1, 1, 1, dataPerThread},
       /*strides=*/ArrayRef<int64_t>{1, 1, 1, vectorLen}, forceUnroll,
-      /*useIndexDiffs=*/true, resultInit);
+      /*useIndexDiffs=*/true);
```
krzysz00, line 333: In silly questions: could this be replaced by just creating a `threadwise_read_into` op? 

That would give us a nice marker operation that'd let us fuse operations onto loads.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -334,43 +330,84 @@ static TransformingForOp createGlobalLoadLoop(PatternRewriter &b, Location loc,
...
-    }
-    b.create<rock::YieldOp>(loc, innerLoop.getResult(0));
+
+    b.create<InBoundsStoreOp>(loc, loaded, loadBuffer,
+                              outerLoop.getLowerCoords(/*domain*/ 1)[3]);
   }
   return outerLoop;
 }
 
+void copyAndTransposeFromLoadToStoreBuffer(
```
krzysz00: I'm not sure I like this approach. Could we, instead, have this `packLoadBufferToStoreBuffer()`, where `loadBuffer` is either `D x K` or `K x D` and `storeBuffer` is `kOuter x D x kpack`?

That way, we'd be calling this code unconditionally and the writes to LDS would appear to be more vectorizable.

krzysz00 (commented):
I think there might be a better way to do this that might also improve perf

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
giuseros: So the way it's working now is:
- global -> loadBuffer 
- loadBuffer -> storeBuffer
- storeBuffer -> lds

Are you suggesting to have something like:
- global -> loadBuffer
- loadBuffer -> lds
?

Because `kOuter` is only an LDS thing. Each thread loads elements into a `copyDPerThread*copyKPerThread` buffer (in local memory) This needs to be transposed (or not) depending on how we read from global and how we are storing in LDS. I need to use two buffers (I think) because we are pipelining

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
giuseros: I gave it a quick try, but in our reduction loop we have:
``` 
      Value iv = loopOp.getInductionVar();
      IRMapping loadAUpdates, loadBUpdates;
      auto blockwiseLoadAClone = cast<TransformingForOp>(
          b.clone(*blockwiseLoadA.getOperation(), loadAUpdates));
      blockwiseLoadAClone.setOperand(
          blockwiseLoadAClone.getUpperInits(/*domain=*/0)
              .getBeginOperandIndex(),
          iv);
```
To update the reduction index (`blockwiseLoadA` is a `transformingFor` operation). How can I do something similar for `threadwise_read_into` ?

krzysz00: ... right, good point, that won't work.

We'd need to define a new op, which would probably be a good idea anyway, and that's a future PR.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: I'm suggesting this:

```
loadBuffer : KPerThread x DPerThread or DPerThread x KPerThread
storeBuffer : KOuterPerThread x DPerThread x KPackPerThread
lds : KOuter x D x KPack

global memory -> loadBuffer
loadBuffer -[pack]> storeBuffer
storeBuffer -> lds
```

giuseros: I am not sure. I think this is already working like you are proposing. Because we are vectorizing by kpack in the LDS storing routine.

Let's say we have `kperblock=64` , `mperblock=128`, `kpack=16` . `loadBuffer` is already packed into a `16x2` buffer (we load 16 times vectors of length 2) . The final thing we need is to have it transposed in a `2x16` buffer, because we need to store it into the LDS buffer which is `4x128x16`. The fact that `kouter=4` is a workgroup thing, the thread doesn't know it. 

Basically having a private `storeBuffer` sized `4x2x16` wouldn't be useful for the single thread (I think).

krzysz00: Ok, let me rephrase, the pack buffer is KPackPerThread, not Kpack - my bad

krzysz00: The point is to arrange the data in the order that it'll be sent to LDS by each thread

giuseros: Ok, I applied the changes. I thought of not wrapping the buffers with transforms, but use the transforms directly in the for loop. It's mostly because it seemed too much for a simple transform like this, to have a separate wrapping step.

krzysz00 (approved):
Looks like it should work fine to me

giuseros:
Ok, I re-ran the benchmarks and they are fine (even slightly better than before). Attached the second perf comparison (upstream vs new) for future records. I also ran all the nightlies and I didn't get any issues.  Merging this in. 

[final2.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/11104392/final2.zip)

PR #1024, opened 2023-03-27
Matching quant and dequant ops from MIGraphX

jerryyin:
MIGraphX provided us with correct kernel prototype with both ops as following:

Dequantizelinear op:
```cpp
output[i] = static_cast<fp32>(input[i] - zero_pts[i]) * scales[i];
```

Quantizelinear op:
```cpp
int64_t quantized = static_cast<int32>(std::round(input[i] / scales[i])) + zero_pts[i];
output[i] = std::max(-128, std::min(127, quantized));
```

MLIR's implementation will not be identical to MIGraphx's because of the abstraction mismatch mentioned in https://github.com/ROCmSoftwarePlatform/rocMLIR/discussions/981. The math between using TOSA corresponding ops will be more or less similar.

We will explore to expose MIGraphX specific ops in TOSA extensions, but for the short term of this release, we continue choose to converge on the high level op.

### DIFF
--- a/mlir/include/mlir/Dialect/MIGraphX/MIGraphXOps.td
+++ b/mlir/include/mlir/Dialect/MIGraphX/MIGraphXOps.td
@@ -166,7 +166,8 @@ def MIGraphX_QuantizeLinearOp :
 	  Results<(outs TensorOf<[I8]>:$output)> {
   let summary = "Channelwise quantization";
   let description = [{
-    Quantization tensor elementwise
+    Quantization tensor channelwise. It computes the following:
+    tensor[n,c,h,w] = tensor[n,c,h,w] / quantScale[c] + quantBias[c]
   }];
   let assemblyFormat = "`(` operands `)` attr-dict `:` `(`type(operands)`)` `->` type(results)";
 }
@@ -179,7 +180,8 @@ def MIGraphX_DeQuantizeLinearOp :
 	  Results<(outs TensorOf<[F32]>:$output)> {
   let summary = "Channelwise dequantization";
   let description = [{
-    De-Quantization tensor elementwise
+    De-Quantization tensor channelwise. It computes the following:
+    tensor[n,c,h,w] = (tensor[n,c,h,w] - quantBias[c]) * quantScale[c]
   }];
   let assemblyFormat = "`(` operands `)` attr-dict `:` `(`type(operands)`)` `->` type(results)";
 }
--- a/mlir/include/mlir/Dialect/MIGraphX/MIGraphXOps.td
+++ b/mlir/include/mlir/Dialect/MIGraphX/MIGraphXOps.td
@@ -160,7 +160,7 @@ def MIGraphX_FloorOp :
 
 def MIGraphX_QuantizeLinearOp :
     MIGraphX_Op<"quantizelinear">,
-    Arguments<(ins TensorOf<[I32]>:$input,
+    Arguments<(ins TensorOf<[F32]>:$input,
                    TensorOf<[F32]>:$scale,
                    Optional<TensorOf<[I32]>>:$bias)>,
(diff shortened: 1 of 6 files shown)


mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -75,6 +75,10 @@ static tosa::CastOp createCastOp(PatternRewriter &rewriter, Location loc,
   return op;
 }
 
+static Type getShapedElementTy(const Value &v) {
```
krzysz00: `Value` is a smart pointer, there's no need to `const Value &` it.

krzysz00 (commented):
I've got a nit.

Also, for a higher-level perspective ... which definitions of quantize and de-quantize do frameworks actually want?

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jerryyin: Good point!

> which definitions of quantize and de-quantize do frameworks actually want?

Turns out none of framework want this, it driven exclusively by migraphx integration. The prototype used to sent to us is wrong.

krzysz00: ... ok, so what _do_ the frameworks want? And why does MIGraphX want this weird thing?

(I'm willing to approve on the basis that this is important integration stuff, but I'd like to understand)

mlir/include/mlir/Dialect/MIGraphX/MIGraphXOps.td
```diff
@@ -160,15 +160,28 @@ def MIGraphX_FloorOp :
...
+
+def MIGraphX_DeQuantizeLinearOp :
+    MIGraphX_Op<"dequantizelinear">,
+    Arguments<(ins TensorOf<[I32]>:$input,
+                   TensorOf<[F32]>:$scale,
+                   Optional<TensorOf<[I32]>>:$bias)>,
+	  Results<(outs TensorOf<[F32]>:$output)> {
+  let summary = "Channelwise dequantization";
+  let description = [{
+    De-Quantization tensor elementwise
```
sjw36: Please document the conversion here (as specified in the PR description).

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -484,27 +488,59 @@ class QuantizeLinearConverter final
...
-    auto input = op.getInput();
-    auto scale = op.getScale();
-    ShapedType inputType = input.getType().cast<ShapedType>();
-    auto elementType = inputType.getElementType();
+    Value input = op.getInput();
+    Value scale = op.getScale();
+    Value output = op.getOutput();
+    Location loc = op->getLoc();
+
+    Type elementType = getShapedElementTy(input);
```
sjw36, line 500: Would be good to also comment with the conversion pseudo code here as well.

sjw36 (approved):
Looks good. Just want a bit more documentation.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jerryyin: > so what do the frameworks want?

int8 optimizations shoot for inference, framework doesn't want it because they care about training instead.

> And why does MIGraphX want this weird thing?

Not sure how is it "weird"? I'd agree that they can implement this in a more clever way, like TOSA rescale, but the math they have looks reasonable to me. It is one way to do it, but why they do what they do is out of scope for this review.

PR #1025, opened 2023-03-27
Add support for fp8 gemm/conv inside our dialect (no hookups)

krzysz00:
This commit allows our operations (rock.conv2d, rock.gemm, rock.xdlops_gemm_v2, and so on) to work with the two 8-bit float types that will be available on gfx940.

This commit does not enable generating or testing these kernels, nor does it allow for a non-xdlops fp8 path (which would require some sort of software float8 pass).

Unlike the other MFMA instructions, the fp8 instructions support mixed-format inputs: we can freely mix-and-match fp8 (aka Float8E4M3FNUZ) and bf8 (Float8E5M2FNUZ) within GEMMs.

Therefore:
- Various structures (such as the tuning parameter information, MFMA selection, and the GemmWrapperInterface) have been extended to distinguish between the types of matrix A and matrix B.
- gridwise_gemm[_v2] has been updated to use two LDS buffers, one for matrix A and one for matrix B, instead of one large buffer. The compiler folks assured me this was fine - I'll give it a perf check tomorrow.
- As a consequence, the static LDS buffer offsets have been removed from blockwise_gemm_v2.
- In the few places it was needed (creating constants), fp8 and bf8 have been added as cases
- This _ought_ to hook up fine to the AMDGPU dialect and below, but I haven't been in a position to run an e2e test.

Future work includes some some flavor of non-xdlops path, hooking up rocmlir-gen, and CPU verification.

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
+++ b/mlir/include/mlir/Dialect/Rock/IR/MfmaInsnGroup.h
@@ -22,7 +22,16 @@
 namespace mlir {
 namespace rock {
 
-enum class MfmaTypeId : uint32_t { Fp32TyId = 0, Fp16TyId, Bf16TyId, I8TyId };
+enum class MfmaTypeId : uint32_t {
+  Fp32TyId = 0,
+  Fp16TyId,
+  Bf16TyId,
+  I8TyId,
+  Fp8Fp8TyId,
+  Fp8Bf8TyId,
+  Bf8Fp8TyId,
+  Bf8Bf8TyId
+};
 
 struct MfmaInsnInfo {
   MfmaTypeId type;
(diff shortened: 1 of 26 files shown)


krzysz00:
Preliminary observations: due to weird phase ordering stuff inside the compiler causing this change to break LDS vectorization, I think this may be blocked on the version of #1023 that I've talked over with @giuseros (which would have an unconditional "pack" step) after which we'd make the LDS buffers be `memref<kOuter x D x vector<kPack x T>, #workgroup>`, so as to ensure we get the wide reads we want. Still waiting on the perf runs to finish, but still.

krzysz00:
Update: even though the ISA looked a lot uglier, the perf impact is basically noise ... and so the "LDS is vectors" change can be done in the future. (untuned GEMM is down ... a tenth or a half of a %, convolutions are +- on this)
[split-lds.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/11093029/split-lds.zip)

krzysz00:
skip-ci because release merging

giuseros (approved):
LGTM!

PR #1026, opened 2023-03-28
Fixed rocMLIR submodule'ing

sjw36:
Make relative paths for sub-projecting
* when importing rocMLIR as a submodule paths should derive from CMAKE_CURRENT_...
Also moved all LLVM relevant cmake settings to cmake/llvm-project.cmake


jerryyin (approved):
LGTM

I'm not clear about the motivation of this PR though? This looks to me a NFC, or does it address any problem you run into?

krzysz00:
I think this fixes stuff like embedding us inside Triton

PR #1027, opened 2023-03-28
Replacing miirLazyInit with dialect specific registration

jerryyin:
This is a follow-up to @giuseros 's #847. @sjw36 's question on why we need those x86 dependencies brought this to my attention. Then I realize `miirLazyInit()` is a subset of `registerRocMLIRPasses()`. So we can safely get rid of the lazy initialization.

Please let me know if I'm doing this properly.


krzysz00:
Please also take a look at #1018 - specifically, the limited change I'd like to see is that you should set `LLVM_TARGETS_TO_BUILD` to  just "AMDGPU" in the static library case, since we only need "X86;AMDGPU" for JIT when we're not building for clients.

And, having looked, the AMDGPU target initilizations are called be registering the SerializeToHsaco pass, which is why we can safely remove them.

krzysz00:
While you're here, can you get the various LLVM init calls out of the C API tests?

sjw36 (approved):
Looks good.

mlir/test/fusion/e2e/resnet50/lit.local.cfg
```diff
@@ -1,2 +1,3 @@
-if not config.enable_rock_driver_e2e_test or config.no_AMD_GPU:
-  config.unsupported = True
+#if not config.enable_rock_driver_e2e_test or config.no_AMD_GPU:
+#  config.unsupported = True
+config.unsupported = False
```
krzysz00: ... why this change?

krzysz00 (changes requested):
I think you committed the wrong change, and you've enabled e2e tests in cases where they won't work.

jerryyin:
> you should set LLVM_TARGETS_TO_BUILD to just "AMDGPU" in the static library case

Nice suggestion.

>  the AMDGPU target initilizations are called be registering the SerializeToHsaco pass, which is why we can safely remove them.

Yes, I figured that following the existing init/register function routines is the more robust way to do it (compared with calling the register function from SerializeToHsaco)

> While you're here, can you get the various LLVM init calls out of the C API tests?

Which one are you talking about? I find no test do LLVM init under `/tools/rocmlir-lib` and its related test.

krzysz00:
The C API tests are in `mlir/test/CAPI/` and cal the LLVM init functions by hand too

krzysz00:
(and also probably have an unnecessary x86 dependency)

mlir/test/fusion/e2e/resnet50/lit.local.cfg
jerryyin: Yes, I somehow committed the wrong file. That was me shortcut our lit file to make it test a larger subset when doing quantization work...

Corrected now.

krzysz00 (approved):
LGTM!

jerryyin:
> The C API tests are in mlir/test/CAPI/ and cal the LLVM init functions by hand too

I did what I can do there. Nice suggestion.

> (and also probably have an unnecessary x86 dependency)

That turns out to be necessary, I think mixr runtime is dependent on x86 for host side. (Would otherwise run into linker error for dependency on LLVMInitializeX86AsmPrinter)

krzysz00:
Having looked at your CI failure, what needs to happen (because I hit the same failure on the COMgr PR) is that any targets that depend on a runtime library (ex. `mlir_async_runtime`) need to only be added when the execution engine is enabled (or, equivalently, with ROCM_RUNNER enabled, or so on and so forth).

The point is that, when we don't build X86, the execution engine doesn't get built because the JIT isn't available, and we haven't been as disciplined as we need to be about not depending on the execution engine libraries in cases where the e2e tests are disabled.

jerryyin:
@krzysz00 Thanks for helping me to look at the CI. I will probably ping you offline on this. But I actually think we should do the other way around: Always include x86 as our backend because testing is part of our repo's requirement. Then make the inclusion of x86 backend to be a library dependent decision. The reason why I prefer this way is such that we provide as much test coverage to static library as possible. The unit test, because they are there, used to help me rule out possibilities of failure and I don't want to get rid of them to enforce no-x86 backend policy.

krzysz00:
@jerryyin To quickly comment, we can't run any tests that require the MLIR execution engine using the static library (so we can't run out e2e tests on it) because the execution engine utilitty libraries are inherently shared objects. Therefore, we must fix our test suite to not depend on those runtime wrappers when they're not available

jerryyin:
> we can't run any tests that require the MLIR execution engine using the static library

You are right about this. I get myself confused and didn't realize that we can only run lowering test case but not runtime ones. Fixing test suite is a reasonable approach here.

mlir/test/CAPI/CMakeLists.txt
```diff
@@ -5,6 +5,10 @@ set(LLVM_OPTIONAL_SOURCES
   mixr_full.c
 )
 
+set(LLVM_LINK_COMPONENTS
+  X86
+)
+
```
jerryyin: Note that this is the blocking factor for removing x86 backend from static lib. Those tests (as executables) should be able to continue to run in static lib scenarios.

This make me to decide to revert the change that get rid of x86 backend. And if we indeed want to push for that, we'll need to make a decision how to deal with future components/tests that would dependent on x86 backend (and in a different PR).

krzysz00: ... Those tests should be amemned to not need the x86 backend

krzysz00: In that, as far as I'm aware, they don't actually use the JIT, so they don't need the x86 backend.

jerryyin: I'll cherry-pick dd2b727 and 8309f20 in the next PR to figure out what's causing the linking issue from CI compilation.

PR #1028, opened 2023-03-28
Add migraphx.mul to illegal ops

manupak:
Currently the dialect transformation does not
define mul as an illegal op. Hence the tablegen
define lowering is not taking place. This commit
fixes that and tests to assure that.

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/849

### DIFF
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosaPass.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosaPass.cpp
@@ -44,17 +44,8 @@ public:
     auto &ctx = getContext();
     RewritePatternSet patterns(&ctx);
     ConversionTarget target(ctx);
-    target.addLegalDialect<tosa::TosaDialect, migraphx::MIGraphXDialect,
-                           func::FuncDialect>();
-    target.addIllegalOp<
-        migraphx::AddOp, migraphx::ConstantOp, migraphx::ConvolutionOp,
-        migraphx::QuantConvolutionOp, migraphx::RsqrtOp, migraphx::ReluOp,
-        migraphx::TransposeOp, migraphx::BroadcastOp,
-        migraphx::MultiBroadcastOp, migraphx::ReshapeOp, migraphx::DotOp,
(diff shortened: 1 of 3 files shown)


krzysz00 (approved):

sjw36 (approved):
Good. But this list still seems very limited.. what about Sub,Div,Tanh,Sqrt,Floor,Ceil, etc.

sjw36:
Probably need to just make the whole dialect illegal..

krzysz00:
I think the point is that you only add illegal ops if you have a conversion for them?

krzysz00:
... yeah, that's a good point. Might as well make the whole dialect illegal, and then any ops we don't support would get kicked back as errors

manupak:
Agreed and added the dialect as illegal.

> Sub,Div,Tanh,Sqrt,Floor,Ceil

For some of these conversions exist (e.g. sub), but for some they dont (e.g. sqrt). I think its better to create a separate ticket(s) to implement those.  Some are super trivial (1:1) and some are not (e.g. sqrt as there is no tosa.sqrt, instead we need to do tosa.rsqrt + tosa.reciprocal)

Here in this PR, I'd like to scope it to be just 'mul' because it seems the next most beneficial one (as long as transformer models are concerned) and enabling that on MIGraphX has exposed several other issues (well exclusive to MIGraphX side of things). Therefore, after we get mul enabled e2e (migraphx) it should be easy to get others in (easy but not automatic).

sjw36:
> > Sub,Div,Tanh,Sqrt,Floor,Ceil
> 
> For some of these conversions exist (e.g. sub), but for some they dont (e.g. sqrt). I think its better to create a separate ticket(s) to implement those. Some are super trivial (1:1) and some are not (e.g. sqrt as there is no tosa.sqrt, instead we need to do tosa.rsqrt + tosa.reciprocal)
> 
> Here in this PR, I'd like to scope it to be just 'mul' because it seems the next most beneficial one (as long as transformer models are concerned) and enabling that on MIGraphX has exposed several other issues (well exclusive to MIGraphX side of things). Therefore, after we get mul enabled e2e (migraphx) it should be easy to get others in (easy but not automatic).

Yes, sounds good.

PR #1030, opened 2023-03-29
Fixing benchmark/tuner issues

giuseros:
This is fixing discrepancies between the tuner, the driver and `rocmlir-gen`. The main issues were:
- The benchmark driver was not filling the input
- The tuner was filling the input uniformly
This meant that tuner and benchmark driver were faster than `rocmlir-gen` (since the latter was using a fixed pattern, filling the tensors with actual data).

### DIFF
--- a/mlir/tools/rocmlir-tuning-driver/CMakeLists.txt
+++ b/mlir/tools/rocmlir-tuning-driver/CMakeLists.txt
@@ -67,7 +67,4 @@ set_property(TARGET rocmlir-tuning-driver
   target_link_libraries(rocmlir-tuning-driver PRIVATE ${LIBS})
   target_link_libraries(rocmlir-tuning-driver PUBLIC benchmark-driver-utils)
   target_link_libraries(rocmlir-tuning-driver PUBLIC hip::host hip::amdhip64)
-  target_include_directories(rocmlir-tuning-driver PUBLIC
-   "$<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}>"
-  )
 endif()
--- a/mlir/tools/rocmlir-tuning-driver/CMakeLists.txt
+++ b/mlir/tools/rocmlir-tuning-driver/CMakeLists.txt
@@ -67,4 +67,7 @@ set_property(TARGET rocmlir-tuning-driver
   target_link_libraries(rocmlir-tuning-driver PRIVATE ${LIBS})
   target_link_libraries(rocmlir-tuning-driver PUBLIC benchmark-driver-utils)
   target_link_libraries(rocmlir-tuning-driver PUBLIC hip::host hip::amdhip64)
+  target_include_directories(rocmlir-tuning-driver PUBLIC
+   "$<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}>"
+  )
 endif()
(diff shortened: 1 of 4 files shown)


giuseros:
Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/846

sjw36 (approved):
Looks good.

mlir/tools/rocmlir-tuning-driver/rocmlir-tuning-driver.cpp
```diff
@@ -41,27 +41,8 @@
...
-  static_assert(!std::is_reference<T>{},
-                "A __global__ function cannot have a reference as one of its "
-                "arguments.");
-  _vargs[n] =
-      const_cast<void *>(reinterpret_cast<const void *>(&std::get<n>(formals)));
-  return pArgs<n + 1>(formals, _vargs);
-}
-#endif
+// Utilities to allocate buffers
+#include "../utils/performance/common/benchmarkUtils.h"
```
sjw36, line 45: Can we export these somewhere? (so we don't have a relative path)

krzysz00: Nah, these are internal headers

krzysz00 (approved):
LGTM - can we do a followup on making the GPU memcopy an async operation

mlir/tools/rocmlir-tuning-driver/rocmlir-tuning-driver.cpp
krzysz00: ... on the other hand, this could go to `mlir/include/utils/`, perhaps.

giuseros: I added a `target_include_directories` so that we don't have relative paths. We could move things in a later (after 5.6) PR. I am always afraid of fiddling with CMake so late in the release cycle :)

mlir/tools/rocmlir-tuning-driver/CMakeLists.txt
```diff
@@ -65,5 +65,9 @@ set_property(TARGET rocmlir-tuning-driver
 
   llvm_update_compile_flags(rocmlir-tuning-driver)
   target_link_libraries(rocmlir-tuning-driver PRIVATE ${LIBS})
+  target_link_libraries(rocmlir-tuning-driver PUBLIC benchmark-driver-utils)
   target_link_libraries(rocmlir-tuning-driver PUBLIC hip::host hip::amdhip64)
+  target_include_directories(rocmlir-tuning-driver PUBLIC
+   "$<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}>"
```
krzysz00: @sjw36 Will this break the submodule stuff?

I think we might want to stick with a relative path for now

giuseros: I am happy both ways, but what is the "submodule stuff"?

sjw36: If someone makes rocMLIR a submodule, then all our CMake files need to use CMAKE_CURRENT_<SOURCE/BINARY>_DIR. Please update this one.

giuseros: As per agreed, I reverted to using relative paths. We should refactor this in the next release.

PR #1031, opened 2023-03-29
Amending quantized ops type requirement according to migraphx

jerryyin:
This is to help conform to migraphx/onnx standard. Quotes from @pfultz2 : 

- for migraphx quantizelinear, the bias and output type must be the same, so if its int8 output then it needs to be int8 bias. this also matches the onnx spec.
- For dequantizelinear the inputs and bias must be the same type, so if the input is int32 then the bias needs to be int32. That also matches the onnx spec.

### DIFF
--- a/mlir/include/mlir/Dialect/MIGraphX/MIGraphXOps.td
+++ b/mlir/include/mlir/Dialect/MIGraphX/MIGraphXOps.td
@@ -160,30 +160,32 @@ def MIGraphX_FloorOp :
 
 def MIGraphX_QuantizeLinearOp :
     MIGraphX_Op<"quantizelinear">,
-    Arguments<(ins TensorOf<[F32]>:$input,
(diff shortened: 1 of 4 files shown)


krzysz00:
Is this release-critical?

jerryyin:
@krzysz00 Yes, I'm also hoping this is the last one to be merge on our side for quantization.

krzysz00 (approved):
Well, I don't see any reason not to merge this.

krzysz00:
I'll note that, in general, the `get*()` methods should be returning a `TypedValue<ShapedType>`, so the cast isn't needed, but we can fix that later.

jerryyin:
@krzysz00 Not sure I catch it, I can address it in next PR once I understand. Are you talking about `getShapedElementTy()` function? Why/how can I return a `TypedValue<ShapedType>` and how does that help to give back the element type?

PR #1032, opened 2023-03-30
Fix a typo in PR #1030

giuseros:


manupak (dismissed):
LGTM

giuseros:
Hi @manupak , @krzysz00 , @sjw36 , 
This is a bit more than I was expecting. In the original version of the tuner we were mem-setting the input and output buffers each time we run a tuning step:
```
for 1:num_configs:
   allocate(buffers)
   memset(buffers)
   runKernel(buffers)
   free(buffers)
```
This was fine, since memset was quite fast. But now we are actually setting each value on the CPU and `memCopy` over to the GPU. So the additional change that I am doing in this PR is restructuring the previous loop to:

```
allocate(buffers)
memset(buffers)
for 1:num_configs:
   runKernel(buffers)
free(buffers)
```
In this way the tuner runs at a reasonable speed. I am not sure we want to include this last commit in the release, but I would  honestly vouch for it.

krzysz00:
@giuseros I don't think that's good. Specifically, the problem here is that after you've run enough kernels, you might get a buffer like [inf, inf, inf, ... inf], which'd have the same implications as the all 1s thing.

manupak:
> @giuseros I don't think that's good. Specifically, the problem here is that after you've run enough kernels, you might get a buffer like [inf, inf, inf, ... inf], which'd have the same implications as the all 1s thing.

What is the all 1s thing ?

I'd naively think this fine for verify-mode=none run ? Is there a perf implication based on the contents of the buffer to affect tuning ?

manupak:
Im trying to get the context here (I admit i've been not following our tuning driver work as i would've liked)

> But now we are actually setting each value on the CPU and memCopy over to the GPU

When did this happen and why ?
(Im seeing your before & after both to be using memset, does one of it means to be a cpu memset ?)

krzysz00:
... on further contemplation, we're only writing to matrix C, except in the case of an atomic add kernel, which only adds in the atomic add case ... yeah, only sending the data once should be fine

giuseros:
@krzysz00 , you are right, I reverted my last change. I guess tuner will be slower for now and then we can fix it later. 

@manupak , the change (from memset to allocate->populate->gpuCopy) is here: https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1030 and this was done to address https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/846 . 

The main issue is that if you feed uniform input to the kernels, the hardware goes faster

krzysz00:
That is
1. Moving the host and GPU mallocs, and the host buffer fill, out of the tuning loop is 100% fine and I'd be happy to land that in the release
2. Only doing one round of host to device memcpy is 99% fine (I'd have minor concerns about caches) and so we should ... maybe not push that as the last commit for 5.6?

krzysz00 (approved):

giuseros:
Alright, let's think about this after the release

sjw36 (approved):
The new process is good since it will let the GPU settle a bit before running the kernels. It may get unrealistic numbers for smaller problems since caches may be already filled.. but this also simulates a model execution when the result may already be in cache for the next kernel in the model sequence.

For bigger problems this should be a non-issue.

PR #1034, opened 2023-03-31
Amending compute sequence of quantize op according to migraphx

jerryyin:
Note, this PR is release critical.

I the previous computation, the cast happens before the addition, so it is:
> static_cast\<int8\>(scale) + bias

(For migraphx) Now, the cast happen after the addition, so it is:
> static_cast\<int8\>(scale + static_cast\<fp32\>(bias))

The above computation will make a difference when scale overflows, say scale = 2000, and bias = -5:
 - For previous computation, result = 127 - 5 = 122
 - For adjusted computation, result = static_cast<int8>(2000 - 5) = 127

### DIFF
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -506,17 +506,16 @@ public:
     Value scaled = createOpAndInfer<tosa::MulOp>(
         rewriter, loc, elementType, input, inverseScale, /*shift=*/0);
 
+    Value shifted = scaled;
     if (auto bias = op.getBias()) {
-      auto biasElementType = getShapedElementTy(bias);
-      Value scaleCast = createCastOp(rewriter, loc, biasElementType, scaled);
-      Value shifted = createOpAndInfer<tosa::AddOp>(
-          rewriter, loc, biasElementType, scaleCast, bias);
-      rewriter.replaceOp(op, {shifted});
-    } else {
-      Type outputType = getShapedElementTy(output);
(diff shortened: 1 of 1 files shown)


krzysz00 (approved):
Approved on the assumption that this is actually the right change. @pfultz2 and anyone else who actually needs to look at this to confirm that this is actually, for real, what we need to implement.

sjw36 (approved):
That looks right. And matches the logic in pseudo code above.

jerryyin:
Note that the migraphx corresponding code is [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/rewrite_quantization.cpp#L50-L52)

I made the mistake because it "seems" redundant that we do the up-cast of bias, which is already int8 type, to fp32 type. I could instead start a fight with migraphx on which one is more correct. But then I figured that migraphx's implementation is based on onnx, who did the convenient addition in python too (refer to [here](https://github.com/onnx/onnx/blob/9962a0cfa38c69aefb5f5852f8934886dfc167cd/onnx/reference/ops/op_quantize_linear.py#L27)).

It honestly still don't quite make sense that onnx spec would require a int8 type for bias (because output type is int8) but then immediately up-cast it to fp32. Guess I'm just playing by their rules now.

manupak:
Im not exactly sure this is correct.

So the biases are in quantized int8 domain, if I am not mistaken. 
In theory, If so the conversion qint8 -> fp32 should come with by multiplying it with the scale and not upcast.

Am I missing something here ?

jerryyin:
@manupak Exactly my doubt. But apparently the reference implementation on migraphx and onnx are not using this way to do it. 

I can confirm my new implementation passing after this PR, it would yield same results with migraphx reference implementaiton.

PR #1035, opened 2023-04-05
Add support for wmma in the rocMLIR backend

giuseros:


external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/AMDGPU.td
```diff
@@ -337,6 +337,10 @@ def MFMAOutTypes : AnyTypeOf<[F64,
                               VectorOfLengthAndType<[4, 16, 32], [F32]>,
                               VectorOfLengthAndType<[4, 16, 32], [I32]>,
                               VectorOfLengthAndType<[4], [F64]>]>;
+// wmma
+def WMMAInTypes : AnyTypeOf<[VectorOfLengthAndType<[16], [F16,BF16,I8]>]>;
```
krzysz00: Nit: spaces after the commas in that list.

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/AMDGPU.td
```diff
@@ -395,4 +399,31 @@ def AMDGPU_MFMAOp :
   let hasVerifier = 1;
 }
 
+def AMDGPU_WMMAOp :
+    AMDGPU_Op<"wmma", [AllTypesMatch<["destC", "destD"]>,
+                       AllTypesMatch<["sourceA", "sourceB"]>,
+                        Pure]>,
+    Arguments<(ins
+                   WMMAInTypes:$sourceA,
```
krzysz00: We could go ahead and name these `$a`, `$b`, and `$c` like with MFMA?

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/AMDGPU.td
```diff
@@ -395,4 +399,31 @@ def AMDGPU_MFMAOp :
...
+                   DefaultValuedAttr<BoolAttr, "true">:$zeroIndexing,
+                   DefaultValuedAttr<BoolAttr, "true">:$signedA,
+                   DefaultValuedAttr<BoolAttr, "true">:$signedB,
+                   DefaultValuedAttr<BoolAttr, "false">:$clamp)>,
+    Results<(outs WMMAOutTypes: $destD)> {
+  let summary = "MLIR wrapper for RDNA3 wmma instructions";
+  let description = [{
+    The `amdgpu.wmma` op is an MLIR wrapper around intrinsics
+    for various `wmma` instructions in the RDNA3 architecture, which perform
+    a 16x16 matrix multiplication for different data types.
```
krzysz00: I don't think `zeroIndexing` is the right name. I'd go for something like `subwordOffset`, defaulting to `0`.

Also, this documentation block should explain what all the attributes do, and when they're applicable.

Finally, how about `signedA` the boolean attribute becoming `unsignedA` the unit attribute (absence => signed), and `clamp` becoming a unit attribute as well?

external/llvm-project/mlir/lib/Dialect/AMDGPU/IR/AMDGPUDialect.cpp
```diff
@@ -185,6 +185,34 @@ void RawBufferAtomicUminOp::getCanonicalizationPatterns(
   results.add<RemoveStaticallyOobBufferWrites<RawBufferAtomicUminOp>>(context);
 }
 
+//===----------------------------------------------------------------------===//
+// WMMAOp
+//===----------------------------------------------------------------------===//
+LogicalResult WMMAOp::verify() {
```
krzysz00, line 200: While you're here, verify attribute sanity (especially with the unit attribute changes above). Make sure, for example, that `unsignedA` is only present on integer operations, or that you're not trying to set a `subwordOffset` that's too large for the given type.

krzysz00 (commented):
I've got some minor design quibbles (mainly around how the attributes are spelled). The main thing I'm going to complain about, however, is documentation. Since we'll be pulling this patch to send externally, we should describe the semantics of WMMA's arguments and return values and what all the attributes do.

We should also get a good commit message that starts something like `[mlir][AMDGPU] Define wrappers for WMMA matrix ops` and says a few words about RDNA 3 in the body.

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/AMDGPU.td
giuseros: Do you mean the `AMDGPU_MFMAOp`? I "think" they are named `sourceA` `sourceB` and `destC`. Or are you suggesting to rename them all?

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -334,6 +340,57 @@ static Value mfmaConcatIfNeeded(ConversionPatternRewriter &rewriter,
...
+
+  if (!elemType.isInteger(8)) {
+    operands.push_back(llvmInput);
+    return;
+  }
+
+  int64_t numBytes = vectorType.getNumElements();
+  Type i32 = rewriter.getI32Type();
+  VectorType vectorType32bits = VectorType::get(numBytes * 8 / 32, i32);
+  auto llvmVectorType32bits = typeConverter->convertType(vectorType32bits);
```
krzysz00, line 363: ... right, none of these are 4xi8, so vectors are fine. Never mind this comment I'm leaving for my own sanity

external/llvm-project/mlir/lib/Dialect/AMDGPU/IR/AMDGPUDialect.cpp
```diff
@@ -194,6 +194,50 @@ void RawBufferAtomicCmpswapOp::getCanonicalizationPatterns(
...
+
+  Type sourceAElemType = sourceVectorAType.getElementType();
+  Type sourceBElemType = sourceVectorBType.getElementType();
+  Type destElemType = destVectorType.getElementType();
+
+  if ((sourceAElemType.isSignedInteger() ||
+       sourceAElemType.isSignlessInteger()) &&
+      getUnsignedA()) {
+    return emitOpError(
+        "A is signed/signless but `unsignedA` passed to the Operation");
```
krzysz00: This isn't right. We use signless integers everywhere in our MLIR setup, and so does the care at this level. Just keep signed/unsigned as an operation property.

("signless" means "the data type doesn't tell you the signedness, the operation does", which is what we have)

krzysz00: ... ok, maybe it's just the error message being broken.

Basically, I don't think we need to validate the integer type here

krzysz00: ... unless you explicitly added handling for signed/signless integers, which isn't a bad idea

external/llvm-project/mlir/test/Target/LLVMIR/rocdl.mlir
```diff
@@ -216,6 +216,66 @@ llvm.func @rocdl.xdlops(%arg0 : f32, %arg1 : f32,
...
+  // CHECK: call <4 x float> @llvm.amdgcn.wmma.f32.16x16x16.bf16.v4f32(<16 x i16> %{{.*}}, <16 x i16> %{{.*}}, <4 x float> %{{.*}})
+  %r8 = rocdl.wmma.f32.16x16x16.bf16 %arg2, %arg2, %arg6 : (vector<16xi16>, vector<16xi16>, vector<4xf32>) -> vector<4xf32>
+
+  // f16 -> f16 (OPSEL = {0,1})
+  // CHECK: call <8 x half> @llvm.amdgcn.wmma.f16.16x16x16.f16.v8f16(<16 x half> %{{.*}}, <16 x half> %{{.*}}, <8 x half> %{{.*}}, i1 {{.*}})
+  %r9 = rocdl.wmma.f16.16x16x16.f16 %arg1, %arg1, %arg7, %zero: (vector<16xf16>, vector<16xf16>, vector<8xf16>, i1) -> vector<8xf16>
+
+  // bf16 -> bf16 (OPSEL = {0,1})
+  // CHECK: call <8 x i16> @llvm.amdgcn.wmma.bf16.16x16x16.bf16.v8i16(<16 x i16> %{{.*}}, <16 x i16> %{{.*}}, <8 x i16> %{{.*}}, i1 {{.*}})
+  %r11 = rocdl.wmma.bf16.16x16x16.bf16 %arg2, %arg2, %arg8, %zero: (vector<16xi16>, vector<16xi16>, vector<8xi16>, i1) -> vector<8xi16>
```
krzysz00: Nit: there's usually a space before the type colon

krzysz00 (commented):
A few minor questions about the validation setup, but overall, looks reasonable to me. My bad for sleeping on this

external/llvm-project/mlir/lib/Dialect/AMDGPU/IR/AMDGPUDialect.cpp
giuseros: Yeah, I misinterpreted the meaning of signless. So signless should always  work. I can get rid of the unsigned/signed types, but yes I added handling for them here:
``` 
def WMMAInTypes : AnyTypeOf<[VectorOfLengthAndType<[16], [F16, BF16, I8, SI8, UI8]>]>;
``` 
So I can keep them and keep the check just for `si8` and `ui8` or remove the check and handling only `i8`. I am happy with both options. What do you suggest?

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
```diff
@@ -379,6 +379,10 @@ def MFMAOutTypes : AnyTypeOf<[F64,
                               VectorOfLengthAndType<[4, 16, 32], [F32]>,
                               VectorOfLengthAndType<[4, 16, 32], [I32]>,
                               VectorOfLengthAndType<[4], [F64]>]>;
+// wmma
+def WMMAInTypes : AnyTypeOf<[VectorOfLengthAndType<[16], [F16, BF16, I8, SI8, UI8]>]>;
+def WMMAOutTypes : AnyTypeOf<[VectorOfLengthAndType<[4, 8], [F32, I32]>,
+                              VectorOfLengthAndType<[8, 16], [F16, BF16]>]>;
```
jerryyin, line 385: [Question] If the result is always 16x16 elements, why there are two alternative vector length each? Does it have anything to do with wave32 and wave64 mode? Are we using/enabling both modes?

jerryyin: Also, get rid of the `f16` as Jing mentioned that they are same bandwidth but lower precision?

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -334,6 +340,57 @@ static Value mfmaConcatIfNeeded(ConversionPatternRewriter &rewriter,
...
+  operands.push_back(sign);
+  operands.push_back(result);
+}
+
+/// Push the output operand. For many cases this is only pushing the output
+/// in the operand list. But when we have f16 -> f16 or bf16 -> bf16 intrinsics,
+/// since the same numbers of VGPRs is used, we need to decide if to store the
+/// result in the upper 16 bits of the VGPRs or in the lower part. To store the
+/// result in the lower 16 bits, set isZeroIndexing to true, otherwise result
+/// will be stored it in the upper part
```
jerryyin: What about fp32? I don't see it in comments or handled in the code below?

BTW, when we populate the wmma output operand, we need to populate in pairs so half of them are stored in lower 16 and another half stored in higher 16?

jerryyin (commented):
Overall looks good, I have mostly questions.

external/llvm-project/mlir/lib/Dialect/AMDGPU/IR/AMDGPUDialect.cpp
krzysz00: If we still want SI8 and UI8, we'd want to make sure the `unsigned` attribute matches ... or, more interestingly, we could ignore the `unsigned` attribute when an explicitly signed/unsigned type gets passed in.

The latter approach (if you call for a sign in the types, we ignore the attribute) feels like it'd make writing a pipeline that handles both signed and unsigned integers more natural.

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
krzysz00: This is a backend wrapper - if the hardware can do it, we add it.

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
giuseros: So for `fp32` and `int8` there is nothing to do, we simply push the output operand (because the output fits entirely in a VGPR). If you mean doing something like a `vector<2x8>` output for the fp16->fp16 version, yes we initially discussed that with @krzysz00 but the problem is that [multidimensional MLIR vectors translate in llvm arrays](https://mlir.llvm.org/docs/Dialects/Vector/#alternatives-for-lowering-an-n-d-vector-type-to-llvm), which cannot be simply flattened. So for now if a user of wmma wants to use the `fp16->fp16` version it will provide a `16xf16` output vector. The operation will fill it at position 0,2,..14 (`subwordOffset==0`) or 1,3,..15 (`subwordOffset==1`)

external/llvm-project/mlir/lib/Dialect/AMDGPU/IR/AMDGPUDialect.cpp
giuseros: Yep, I agree on the latter approach

jerryyin (approved):
LGTM

PR #1037, opened 2023-04-06
[DO NOT SQUASH] Make LDS buffers vector-typed

krzysz00:
In order to properly communicate size and alignment information to the
backend when using kpack > 1, change the LDS buffer types
from (kpacksPerBlock * dPerBlock * kpack) x T to
(kpacksPerBlock * dPerBlock) x vector<kpack x T>.

This required the creation of a memref.reinterpret_elements operation.

The LDS matrix wrapping and LDS store loop wrapping functionality has
been merged into one larger function, and a special-purpose function
for handling the requirements of the non-xdlops pipeline has been
added.

There's also a minor extension to the vectorizer to handle the fact
that we have a scalar iteration map but vector-valued buffers.

### DIFF
--- a/external/llvm-project/mlir/lib/Dialect/MemRef/IR/MemRefOps.cpp
+++ b/external/llvm-project/mlir/lib/Dialect/MemRef/IR/MemRefOps.cpp
@@ -22,6 +22,7 @@
 #include "mlir/Interfaces/ViewLikeInterface.h"
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/SmallBitVector.h"
+#include "llvm/Support/MathExtras.h"
 
 using namespace mlir;
 using namespace mlir::memref;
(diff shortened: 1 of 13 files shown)


krzysz00:
(by-hand nightlies on a lockhart pass)

krzysz00:
Perf results were substantial improvements for some perf configs (mostly on f16 and i8) that faded away with tuning

krzysz00:
Or at least that's what I think happened, those swings showed up when I only ran with perf configs as well, so ... benchmarking weirdness, I guess

krzysz00:
@giuseros In interesting notes, this change brings the number of `ds_read` instructions in the `rocmlir-gen` assembly down from 32 to 8 - so this might help with bank conflict stuff. So this might have more of a perf impact on convolutions. Or might work better  with other tunings or so on.

krzysz00:
Ok, so, having done some profiling on one of the lockhart nodes (and making up for the fact that the LDS bank conflicts metric seems to be broken in ROCm 5.4)

```python
>>> def bank_conflicts(count, gui_active):
...     return 100 * (count / (gui_active * 110))
...
>>> bank_conflicts(7077888,765116) # upstream
8.409762227379424
>>> bank_conflicts(7077888,770759) # new changes
8.348191375467087
```

So it looks like this change doesn't do much for bank conflicts in and of itself. We may want to look into, for example, if we can un-align the LDS a bit to prevent conflicts. (ex. some sort of stride > size scheme for the M/N dimension)

external/llvm-project/mlir/lib/Dialect/MemRef/IR/MemRefOps.cpp
```diff
@@ -2044,6 +2045,146 @@ void ReinterpretCastOp::getCanonicalizationPatterns(RewritePatternSet &results,
...
+  // long, or don't divide evenly, the verifier will complain.
+  int64_t scaleFactor = std::max(1L, sourceWidth / destWidth);
+  if (scaleFactor == 1)
+    return state.addTypes(sourceType.cloneWith(std::nullopt, newElementType));
+  SmallVector<int64_t> newShape(sourceType.getShape());
+  newShape.push_back(scaleFactor);
+  auto resultType =
+      MemRefType::get(newShape, newElementType, MemRefLayoutAttrInterface{},
+                      sourceType.getMemorySpace());
+  return state.addTypes(resultType);
```
giuseros, line 2100: Is this fine? I mean you are returning something here but the function signature is `void(..)`

giuseros (commented):
From a first look, this is fine, but I would like to understand it more before approving it. Also, I have a broader question: from what I understand, you are forcing the elements in LDS to be vectors, so that the backend emits the correct `ds_read_b128` instructions. I am wondering why the backend was not emitting these. Is there a simpler alignment trick that we can do to make the backend do what we want? Also, thinking more about this, to load from LDS we use `InBoundsLoadOp` that should issue `vector::transferRead`. How come that those are not translated into `ds_read_128b`?

external/llvm-project/mlir/lib/Dialect/MemRef/IR/MemRefOps.cpp
krzysz00: Yeah, I'm returning the result of the function, which is also of type `void`

krzysz00:
Having gone and stared at the passes, it looks like vector.transfer_read (or vector.load, which it lowers to) are doing bitcasts and so everything broadly works. However, looking at `./bin/rocmlir-gen --arch gfx90a -t f16 -p | ./bin/rocmlir-driver -c --debug-only=serialize-to-blob 2>&1` the main difference between the old and new code is that the old code loads has LDS reads like this
```
        ds_read2st64_b64 v[20:23], v0 offset0:4 offset1:5
        v_lshlrev_b32_e32 v29, 6, v2
        v_or_b32_e32 v24, v1, v29
        v_lshlrev_b32_e32 v36, 3, v24
        ds_read_b64 v[24:25], v36
        v_or_b32_e32 v38, 0x80, v29
        v_or_b32_e32 v26, v1, v38
        v_lshlrev_b32_e32 v41, 3, v26
        ds_read_b64 v[26:27], v41
        s_waitcnt lgkmcnt(1)
        v_mfma_f32_32x32x8f16 a[16:31], v[20:21], v[24:25], 0
        v_lshl_or_b32 v2, v2, 2, s2
        s_mov_b32 s2, 0x1c20000
        s_waitcnt lgkmcnt(0)
        v_mfma_f32_32x32x8f16 a[16:31], v[22:23], v[26:27], a[16:31]
        v_or_b32_e32 v26, 32, v1
        v_or_b32_e32 v24, v26, v29
        v_lshlrev_b32_e32 v29, 3, v24
        ds_read_b64 v[24:25], v29
        v_mfma_f32_32x32x8f16 a[16:31], v[20:21], v[24:25], 0
```
while the new code has
```
        ds_read2st64_b64 v[20:23], v0 offset0:4 offset1:5
        v_lshlrev_b32_e32 v24, 3, v1
        v_lshl_or_b32 v42, v2, 9, v24
        ds_read2_b64 v[24:27], v42 offset1:32
        ds_read2_b64 v[28:31], v42 offset0:128 offset1:160
        s_waitcnt lgkmcnt(1)
        v_mfma_f32_32x32x8f16 a[16:31], v[20:21], v[24:25], 0
```
I suspect this is partly related to alignment assumptions or a lack thereof, so it may not be the best idea to merge this right away.

giuseros:
I have some additional questions now that you showed me the assembly (disclaimer: these might be me not having a clue :) ):
- How the LLVM generated (between the two versions) compare? Is `ds_read2st64_b64` an LLVM intrinsic?
- If I understand correctly, the new code is loading 6*8 bytes while the old code was loading 5*8 bytes. Why is there such a difference? Might this be related to the alignment, like you are suggesting?
- Indeed, looking at the ISA reference it seems that `ds_read2st64_b64` is reading 2 `qwords` (64 bits) with a larger 64 bits stride. How come we are generating these and not the simpler `ds_read2_b64`?

krzysz00:
I think the fact that we're missing one qword's worth of reads is down to me doing a questionable job cutting up the ISA - I tried to pull identical segments but must have failed. I'll give it another shot tomorrow.

`ds_read2st64_b64` isn't an intrinsic.

The main difference in the generated IR is that
1. `vector.transfer_read` explicitly sets the alignment on the load to be `2`, the alignment of the scalar type. This might impede any more general alignment inference since the alignment is already there.
2. Similarly (since we're still on a MLIR version that uses typed pointers - aka we haven't updated to upstream yet), there's a `llvm.bitcast` in there to change the pointer type from `f16` to `vector<2xf16>`
3. All the combining of scalar (or smaller-vector) loads is a matter for the backend, mainly the instruction scheduler and certain optimization passes.

It's possible that what we actually want here is the ability to provide stronger alignment hints to the MLIR->LLVM lowering, which is probably work everyone else won't mind too much.

(that is, for instance, if we're doing one of our vectorized loops, we know the alignment of the load, which isn't something that's true in general)

krzysz00:
So, here's the post-`opt` (by way of `--print-after-all` only working on the old pass manager and thus the backend) IR for the first set of LDS loads
```
  tail call void asm sideeffect "s_waitcnt lgkmcnt(0)\0As_barrier", ""() #5
  %290 = shl i32 %.fr96, 2
  %291 = and i32 %290, 252
  %292 = getelementptr half, ptr addrspace(3) @__wg_rock_conv2d_gkcyx_ngchw_ngkhw_0_1, i32 %291
  %293 = load <4 x half>, ptr addrspace(3) %292, align 8
  %294 = shl i32 %.fr96, 2
  %295 = and i32 %294, 252
  %296 = or i32 %295, 256
  %297 = getelementptr half, ptr addrspace(3) @__wg_rock_conv2d_gkcyx_ngchw_ngkhw_0_1, i32 %296
  %298 = load <4 x half>, ptr addrspace(3) %297, align 8
  %299 = or i32 %178, %179
  %300 = shl i32 %299, 2
  %301 = getelementptr half, ptr addrspace(3) @__wg_rock_conv2d_gkcyx_ngchw_ngkhw_0_0, i32 %300
  %302 = load <4 x half>, ptr addrspace(3) %301, align 8
  %303 = or i32 %179, 128
  %304 = or i32 %178, %303
  %305 = shl i32 %304, 2
  %306 = getelementptr half, ptr addrspace(3) @__wg_rock_conv2d_gkcyx_ngchw_ngkhw_0_0, i32 %305
  %307 = load <4 x half>, ptr addrspace(3) %306, align 8
  %308 = tail call <16 x float> @llvm.amdgcn.mfma.f32.32x32x8f16(<4 x half> %293, <4 x half> %302, <16 x float> zeroinitializer, i32 0, i32 0, i32 0)
```

and with this change
```
  tail call void asm sideeffect "s_waitcnt lgkmcnt(0)\0As_barrier", ""() #5
  %288 = and i32 %.fr96, 63
  %289 = getelementptr <4 x half>, ptr addrspace(3) @__wg_rock_conv2d_gkcyx_ngchw_ngkhw_0_1, i32 %288
  %290 = load <4 x half>, ptr addrspace(3) %289, align 8
  %291 = and i32 %.fr96, 63
  %292 = or i32 %291, 64
  %293 = getelementptr <4 x half>, ptr addrspace(3) @__wg_rock_conv2d_gkcyx_ngchw_ngkhw_0_1, i32 %292
  %294 = load <4 x half>, ptr addrspace(3) %293, align 8
  %295 = or i32 %176, %177
  %296 = getelementptr <4 x half>, ptr addrspace(3) @__wg_rock_conv2d_gkcyx_ngchw_ngkhw_0_0, i32 %295
  %297 = load <4 x half>, ptr addrspace(3) %296, align 8
  %298 = or i32 %177, 128
  %299 = or i32 %176, %298
  %300 = getelementptr <4 x half>, ptr addrspace(3) @__wg_rock_conv2d_gkcyx_ngchw_ngkhw_0_0, i32 %299
  %301 = load <4 x half>, ptr addrspace(3) %300, align 8
  %302 = tail call <16 x float> @llvm.amdgcn.mfma.f32.32x32x8f16(<4 x half> %290, <4 x half> %297, <16 x float> zeroinitializer, i32 0, i32 0, i32 0)
```

krzysz00:
Another option that presents itself is to allocate a byte array and then use the existing `memref.view` (or a wrapper around it) in order to get IR that'll probably behave similarly in the grand scheme of things.

I'll make a branch of this branch that uses that approach and test it out.

krzysz00:
Closing as implemented in a different way

PR #1038, opened 2023-04-07
[cmake] Fixed submoduling for mlir

sjw36:
* upstream still has at least one CMAKE_BINARY_DIR
  * see https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/861
* also some general cleanup

### DIFF
--- a/cmake/llvm-project.cmake
+++ b/cmake/llvm-project.cmake
@@ -1,16 +1,14 @@
 message(STATUS "Adding LLVM git-submodule src dependency")
 
-# LLVM settings that have an effect on the MLIR dialect
-set(LLVM_ENABLE_ZLIB "OFF" CACHE STRING "")
-set(LLVM_ENABLE_TERMINFO OFF CACHE BOOL "")
-set(LLVM_TARGETS_TO_BUILD "X86;AMDGPU" CACHE STRING "")
+set(LLVM_PROJECT_DIR "${CMAKE_CURRENT_SOURCE_DIR}/external/llvm-project")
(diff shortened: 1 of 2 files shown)


cmake/llvm-project.cmake
```diff
@@ -43,26 +42,26 @@ list(APPEND CMAKE_MODULE_PATH
   "${ROCM_PATH}/hip/cmake"
 )
 list(APPEND CMAKE_MODULE_PATH
-  "${CMAKE_CURRENT_BINARY_DIR}/lib/cmake/mlir"
+  "${CMAKE_BINARY_DIR}/lib/cmake/mlir"
```
krzysz00, line 45: `LLVM_BINARY_DIR`, perhaps? Or is there something else here?

krzysz00 (commented):
One probable typo, looks good otherwise

cmake/llvm-project.cmake
sjw36: Unfortunately that is the upstream problem (see https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/861).

krzysz00: Ok, hold on, what's the problem upstream? `llvm-project.cmake` here is ours, not theirs.

krzysz00:
The CI failures on this one seem to have to do with the RocMLIR tests not knowing where to find `lit`

krzysz00 (approved):
Looks like a solid cleanup, approved

PR #1039, opened 2023-04-11
[MIGraphX] adding support for inlined constants

manupak:
This commit adds support for inlined constants
for migraphx offloads.

### DIFF
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -262,7 +262,7 @@ public:
     if (outRank > inRank) {
       SmallVector<int64_t, 5> newShape = llvm::to_vector<5>(inShape);
       for (uint32_t i = inRank; i < outRank; i++) {
-        newShape.push_back(i);
+        newShape.push_back(1);
       }
       tosa::ReshapeOp sameRankReshapedOp = createOpAndInfer<tosa::ReshapeOp>(
           rewriter, loc, elemType, op.getInput(),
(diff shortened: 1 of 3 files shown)


mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -262,7 +262,7 @@ class MultiBroadcastConverter final
     if (outRank > inRank) {
       SmallVector<int64_t, 5> newShape = llvm::to_vector<5>(inShape);
       for (uint32_t i = inRank; i < outRank; i++) {
-        newShape.push_back(i);
+        newShape.push_back(1);
```
krzysz00, line 265: This need backported?

krzysz00 (approved):
So ... what was the problem, here?

(but yeah, these both feel like obvious bugfixes)

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
manupak: Not necessarily... (could do as well) because we wont hit the case unless we offload scalar literals from migraphx (which we currently dont) all others offloads has iso-rank for multibroadcasts.

Im fine with either.

manupak:
> So ... what was the problem, here?

migraphx dialect passes doesn't allow tosa.consts to be present and a bugfix in differing rank multibroadcasts which weirdly occurs only in scalar constants in how migraphx partitions kernels -- otherwise they create 1x1x1x... shapes to make rank 
match.

krzysz00:
Ok, cool, so, go ahead and land this and either file backports or don't

PR #1040, opened 2023-04-11
[Backport][5.6][MIGraphX] adding support for inlined constants

manupak:
This commit adds support for inlined constants
for migraphx offloads.


manupak:
I've created the backport just in case.
They could be viewed as bug fixes or they could be viewed as we never supported inlined/bound constants from migraphx.
Thus, Im confused a bit. 

If you guys think, we should take these in for 5.6. Im fine too.
(we anyway will use mid release features for migraphx integration anyway)

krzysz00 (approved):
Approved on the basis that shipping obvious bugs is bad

giuseros (approved):
LGTM!

PR #1042, opened 2023-04-12
Adds support for transpose+reshape folding

manupak:
Adds support for transpose+reshape folding

Currently the transposes on inputs are only folded
if they the transpose output immediatedly reaches
a matmul or a conv2d. If they encounter, a reshape
in the middle the folding stops. This will confuse
the gemm-like op.

This commit adds further folding to fold reshapes
i.e. expand/collapse shape ops in between.

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/862

### DIFF
--- a/mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
+++ b/mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
@@ -433,9 +433,9 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
   // with tInput. If there are collapse shapes encountered, the collapse
   // is applied on the tInput.
   LogicalResult mergeTransposeWithGemmLikeOp(PatternRewriter &rewriter,
-                                             const Value &tOutput,
+                                             Value tOutput,
                                              ArrayRef<int32_t> dims,
-                                             const Value &tInput) const {
+                                             Value tInput) const {
     for (auto &use : tOutput.getUses()) {
       if (auto op = dyn_cast<tensor::CollapseShapeOp>(use.getOwner())) {
         SmallVector<ReassociationIndices, 4> reassocIndices =
@@ -464,9 +464,8 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
                                   "other than data or weight");
         }
       } else if (auto matMulOp = dyn_cast<tosa::MatMulOp>(use.getOwner())) {
-        LogicalResult res = checkMatMulTransposeValid(matMulOp, dims);
-        if (res.failed()) {
-          return res;
+        if (checkMatMulTransposeValid(matMulOp, dims).failed()) {
+          return failure();
         }
         bool mmNonTrivial = isMatMulNonTrivial(dims);
         if (matMulOp.getA() == tOutput) {
@@ -507,18 +506,15 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
       top->replaceAllUsesWith(convOp);
     } else if (tosa::MatMulOp matMulOp =
                    tInput.getDefiningOp<tosa::MatMulOp>()) {
-      LogicalResult res = checkMatMulTransposeValid(matMulOp, dims);
-      if (res.failed()) {
-        return res;
+      if (checkMatMulTransposeValid(matMulOp, dims).failed()) {
+        return failure();
       }
       setTranspose(matMulOp, "transpose_c", isMatMulNonTrivial(dims));
       matMulOp->getResult(0).setType(tOutput.getType());
       top->replaceAllUsesWith(matMulOp);
     } else {
-      LogicalResult res =
-          mergeTransposeWithGemmLikeOp(b, tOutput, dims, tInput);
-      if (res.failed()) {
-        return res;
+      if (mergeTransposeWithGemmLikeOp(b, tOutput, dims, tInput).failed()) {
+        return failure();
       }
     }
 
--- a/mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
+++ b/mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
@@ -385,15 +385,21 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
                      const char *layoutDefault, ArrayRef<int32_t> permDims,
                      bool isInput = false) const {
     StringRef currentLayout(layoutDefault);
+    size_t expectedRank = currentLayout.size();
+    size_t offset = 0;
+    if (permDims.size() > expectedRank) {
+      offset = permDims.size() - expectedRank;
+    }
+    SmallVector<int32_t, 4> dims{permDims.begin() + offset, permDims.end()};
     if (auto attr = op->getAttrOfType<StringAttr>(attrKey))
       currentLayout = attr.getValue();
     SmallString<4> layout(currentLayout);
     if (isInput) {
-      for (int i = 0, e = permDims.size(); i < e; ++i)
-        layout[permDims[i]] = currentLayout[i];
+      for (int i = 0, e = dims.size(); i < e; ++i)
+        layout[dims[i] - offset] = currentLayout[i];
     } else {
-      for (int i = 0, e = permDims.size(); i < e; ++i)
-        layout[i] = currentLayout[permDims[i]];
+      for (int i = 0, e = dims.size(); i < e; ++i)
+        layout[i] = currentLayout[dims[i] - offset];
     }
     op->setAttr(attrKey, StringAttr::get(op->getContext(), layout));
   }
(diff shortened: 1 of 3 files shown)


mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -398,39 +398,112 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
...
+          "Can't transpose the batch dimension out of place");
+    }
+    return success();
+  }
+
+  bool isMatMulTrivial(ArrayRef<int32_t> dims) const {
+    return (dims[dims.size() - 2] == 2 && dims[dims.size() - 1] == 1);
+  }
+
+  LogicalResult mergeTransposeWithGemmLikeOp(PatternRewriter &rewriter,
```
krzysz00, line 429: Could this be `FailureOr<Value>`?

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -398,39 +398,112 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
...
+                                             Value tOutput,
+                                             ArrayRef<int32_t> dims,
+                                             Value &tInput) const {
+    for (auto &use : tOutput.getUses()) {
+      if (auto op = dyn_cast<tensor::CollapseShapeOp>(use.getOwner())) {
+        SmallVector<ReassociationIndices, 4> reassocIndices =
+            op.getReassociationIndices();
+        tensor::CollapseShapeOp newCollapseShapeOp =
+            rewriter.create<tensor::CollapseShapeOp>(op.getLoc(), tInput,
```
krzysz00, line 438: Isn't `tInput` an output parameter?

krzysz00: ... Oh. It's both. Could use a comment.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -398,39 +398,112 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
...
+        rewriter.eraseOp(op);
+      } else if (auto op = dyn_cast<tensor::ExpandShapeOp>(use.getOwner())) {
+        SmallVector<ReassociationIndices, 4> reassocIndices =
+            op.getReassociationIndices();
+        ArrayRef<int64_t> outShape = op.getResultType().getShape();
+        ArrayRef<int64_t> inShape = op.getSrcType().getShape();
+        SmallVector<int64_t, 4> newOutShape = llvm::to_vector<4>(outShape);
+        // Assumption is shape expansions happen only in higher dimensions.
+        for (size_t i = 0; i < dims.size(); i++) {
+          newOutShape[newOutShape.size() - 1 - i] =
```
krzysz00: What's this supposed to be doing?

krzysz00: If you're trying to apply the transpose to the output shape, maybe go through the reassociation expressions and map those over?

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -443,46 +516,22 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
       top->replaceAllUsesWith(convOp);
     } else if (tosa::MatMulOp matMulOp =
                    tInput.getDefiningOp<tosa::MatMulOp>()) {
-      setTranspose(matMulOp, "transpose_c", matmulNonTrivial);
+      LogicalResult res = checkMatMulTransposeValid(matMulOp, dims);
+      if (res.failed()) {
```
krzysz00: You can do `if(failed(checkMatMulTranspose(...)) { return failure(); }` which is somewhat more the style.

krzysz00: Or `.failed() { return failure(); }`

krzysz00 (commented):
Minor nits and slight confusion, but the idea seems fine.

Also, can we get a test for convolutions?

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
manupak: Basically, I want to lift the expand shape on top of transpose.

I thought about faithfully following the reassociation expressions but then it would associate expanded dim strictly as before -- which is oddly what I dont want.

E.g. 

Before:
tp:                    m x n ---> n x m      
expand_shape: n x m ---> 1 x n x m
                         reassoc : [0, 1] , [2]

After (what I want it to be with no tp)
new_expand_shape: m x n --> 1 x m x n
                                 reassoc : [0, 1] , [2]  is still the same.
                                 
So reassociation seems identical -- just some dims are swapped. 
Here I just wanted to calculate the outShape of new_expand_shape while using the old reassociation.

I maybe wrong -- but strictly associating with reassociation felt like it would be give something like m x 1 x n which is not what we want here ?

manupak: or I could drop expand shape support -- we dont strictly need it (yet). What I m seeing is expanded shapes coming from clients (that has more dims than the ops require) and then our lowering add reshapes to collapse.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
manupak: no, there is nothing returned here except LogicalResult -- Ive changed others const refs to make it clear.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
manupak: I've added a comment all are inputs to the function.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
manupak: removed it for now -- it seems confusing and I dont see us exercising the case either (unless in a contrived test) -- we can revisit if needed.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
manupak: I want to propagate the error.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
krzysz00: ... ah, that makes more sense. 

I think the cases where we have expand_shape after this all could be stuff like adding a group dimension ... but that's usually done with coordinate transforms.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
krzysz00: There isn't an error. `LogicalResult` is a wrapper around `bool`

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -406,31 +412,89 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
...
+    int32_t prevLastDim = dims.size() - 2;
+    return (dims[prevLastDim] == lastDim && dims[lastDim] == prevLastDim);
+  }
+
+  // This function traverses the uses of tOutput and then modifies
+  // the uses to indicate the input are transposed and replaces them
+  // with tInput. If there are collapse shapes encountered, the collapse
+  // is applied on the tInput.
+  LogicalResult mergeTransposeWithGemmLikeOp(PatternRewriter &rewriter,
+                                             const Value &tOutput,
```
krzysz00: `Value` is a smart pointer, just pass it by value. That is, any and all uses of `const Value&` can be `Value`

manupak: done.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
manupak: right.. need to learn more about diagnostic handling. done.

krzysz00 (commented):
Wait, hold on, I'm not sure that a uniform `offset` is right? Don't we want to subtract off the number of dimensions that have been removed at that point in the transpose?

manupak:
Ok, yea I was taking the luxury of assuming that collapses affect only higher order dims.

Now, I've generalized that a bit to track associations and removed the offset. Now the permuted dims are modified as and when a collapse is encountered and added a test case where lower order dims are collapsed as well (though they are unlikely for us to encounter)

krzysz00 (approved):
This all seems pretty reasonable to me. I'm trusting that the `sort` call is needed for some reason I don't fully understand right now.

PR #1044, opened 2023-04-14
Allocates LDS buffers as bytes and view them as appropriate vectors

krzysz00:
In order to generate more vectorized reads from LDS (using things like ds_read2_b64), we need LLVM to be using simpler address computation logic, namely, we need the getelementpointer operations we're doing to be vector-typed. This doesn't necessarily mean we need to allocate our LDS buffers as vector-typed values, but it does mean that we need to do the view conversion instead of calling vector.transfer_read / vector.load , as this leads to computing a scalar index and then loading a vector based off of it.

Since we don't need to actually make the buffer vector-typed, we can instead use a buffer of bytes and the existing `memref.view` op, which is intended for exactly this sort of situation (and may have been motivated by similar LDS-using code in someone else's kernel).

Since our kernels have a bunch of nice powers of 2 running around, we don't necessarily need to explicitly pass the alignment of the global to LLVM, but I'm fixing to do that in a followup commit anyway.

The LDS matrix wrapping and LDS store loop wrapping functionality has been merged into one larger function to make this work - partly because that's how the previous PR did it.

There's also a minor extension to the vectorizer to handle the fact that we have a scalar iteration map but vector-valued buffers.

Superscedes PR #1037, for which see more info on what this changes about our output (there doesn't seem to be much difference in the ultimate effects)

(@zhanglx13 , you're on this because it might be relevant to Triton integration stuff, even if the code changes from this change shouldn't be much )

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
@@ -84,6 +84,24 @@ static Type obtainAccumulatorType(OpBuilder &b, Type elementTypeA,
   return accumulatorType;
 }
 
+/// Construct a `memref.view` operation that interprets the buffer `buffer`,
+/// whose elements are bytes, as a buffer of `type`.
+static TypedValue<MemRefType> viewBufferAs(OpBuilder &b, Value buffer,
+                                           Type type) {
+  Location loc = buffer.getLoc();
+  Value zeroByteOffset = b.createOrFold<arith::ConstantIndexOp>(loc, 0);
+  auto bufferType = buffer.getType().cast<MemRefType>();
+  int64_t byteWidth = getByteWidth(type);
+  int64_t numBytes = bufferType.getShape()[0];
+  assert(numBytes % byteWidth == 0 && "Can't evenly fit type into buffer");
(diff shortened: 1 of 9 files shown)


mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -485,18 +549,20 @@ createLdsStoreLoop(PatternRewriter &b, Location loc, Value storeBuffer,
     b.setInsertionPointToStart(loop.getBody());
     Value toStore = b.create<InBoundsLoadOp>(loc, storeType, storeBuffer,
                                              loop.getLowerCoords(0));
-    b.create<InBoundsStoreOp>(loc, toStore, rawBuffer,
-                              loop.getLowerCoords(/*domain=*/1));
+    if (ldsWriteLen == 1) // kpack = 1, vectorized in D
```
giuseros, line 552: Shouldn't that happen already when you lower `InBoundsStoreOp`?

giuseros:
It's all clear to me, except why you need to introduce a buffer of bytes. I mean why you can't have a normal typed buffer and then view it as a vector-typed buffer?

giuseros (approved):
After a chat and another look, LGTM!

PR #1045, opened 2023-04-14
Remove locks in MIOpen API

krzysz00:
Since the MIOpen integration has been updated to use context properly in #1020, we now will have one context per MIOpen thread (if they have multiple threads) so we can safely run compilation in parallel. This commit makes this happen.


jerryyin (changes requested):
You change LGTM. But testing is a huge part of this change. I remember there's this issue that only exists under circumstance of running with framework but nothing else can reproduce. If you can run us with either TF or Pytorch would be fine.

To give a concrete suggestion, please follow up with @rsanthanam-amd on the previous procedures of Tensorflow-xla + MIOpen + mlir integration model run. I'll merge once a resnet50 model can run with a couple of steps.

krzysz00:
Closed because this doesn't matter anymore

PR #1047, opened 2023-04-17
Collect locations of all outlined ops into a FusedLoc placed on the CallOp

pcf000:
@sjw36 , is this what you had in mind?  The current version caches the FusedLoc in the OutliningCandidate.  The first version is the commented-out block at the bottom, which calculates it anew every time but is compact.  The CallOp is down there, too.

### DIFF
--- a/external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
+++ b/external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
@@ -79,6 +79,7 @@ struct OutliningCandidate {
   std::string partFnName;
   llvm::hash_code hash;
   func::FuncOp function;
+  Location fusedLoc;
 
   /// Return the order index for the given value that is within the block of
   /// this data.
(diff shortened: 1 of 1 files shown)


external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
```diff
@@ -344,7 +353,14 @@ static void outlineOps(Operation *anchorOp, ArrayRef<Operation *> trailingOps,
   if (!trailingOps.empty())
     lastOp = trailingOps[trailingOps.size() - 1];
   b.setInsertionPointAfter(lastOp);
-  func::CallOp callOp = b.create<func::CallOp>(loc, outlinedFunc, values);
+  //   // +++pf: Move this bit up if we want fusedLoc on the FuncOp.
+  //   SmallVector<Location> fusedLocs{anchorOp->getLoc()};
+  //   for (auto *op : leadingOps)           // llvm::concat<Operation*> barfs
+  //     fusedLocs.push_back(op->getLoc());
+  //   for (auto *op : trailingOps)
+  //     fusedLocs.push_back(op->getLoc());
```
krzysz00: It might make sense to say "this function came from pulling together all of these things", yeah.

sjw36 (approved):
I like this approach. Thanks!

PR #1048, opened 2023-04-17
Add migraphx.slice op

manupak:
This commit adds migraphx.slice op and its lowering to tosa.


krzysz00 (approved):
The code itself looks fine, but I do have a question - do we have a rewrite from `tosa.slice` to `ro,ck.transform ... by Slice`?

manupak:
yep but not tosa.slice : we have to rewrite tensor.extract_slice to rock.transform by slice.. working on it.

PR #1049, opened 2023-04-17
[DO NOT SQUASH] Atomic emulation pass (for gfx10, etc.)

krzysz00:
[mlir][AMDGPU] Add emulation pass for atomics on AMDGPU targets

Not all AMDGPU targets support all atomic operations. For example,
there are not atomic floating-point adds on the gfx10 series. Add a
pass to emulate these operations using a compare-and-swap loop, by
analogy to the generic atomicrmw rewrite in MemrefToLLVM.

This pass is named generally, as in the future we may have a
memref-to-amdgpu that translates constructs like atomicrmw fmax (which
doesn't generally exist in LLVM) to the relevant intrinsics, which may
themselves require emulation.

Since the AMDGPU dialect now has a pass that operates on it, the
dialect's directory structure is reorganized to match other similarly
complex dialects.

The pass should be run before amdgpu-to-rocdl if desired.

This commit also adds f64 support to atomic_fmax.

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.h
+++ b/external/llvm-project/mlir/include/mlir/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.h
@@ -8,7 +8,7 @@
 #ifndef MLIR_CONVERSION_AMDGPUTOROCDL_AMDGPUTOROCDL_H_
 #define MLIR_CONVERSION_AMDGPUTOROCDL_AMDGPUTOROCDL_H_
 
-#include "mlir/Conversion/AMDGPUToROCDL/Chipset.h"
+#include "mlir/Dialect/AMDGPU/Utils/Chipset.h"
 #include <memory>
 #include <string>
 
--- a/external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/AMDGPU.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/AMDGPU.td
@@ -136,6 +136,48 @@ def AMDGPU_RawBufferStoreOp :
   let hasVerifier = 1;
 }
 
+// Raw buffer atomic compare-and-swap
+def AMDGPU_RawBufferAtomicCmpswapOp :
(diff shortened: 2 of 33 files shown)


manupak:
(FYI, from skimming : I like the fact that we add generic (more or less) support for atomics still doing a detailed look into the code)

krzysz00:
Yeah, we might be able to get rid of the `atomic_add` feature as a hard gate in the fusion code ... we'd probably want to keep it around for split-K gemms or something though. Maybe rename it.

manupak:
So atomic fmax ops could be emulated using smax and umin (if those are there), it might be perfomant than a generic CAS loop.
However, Im not exactly sure of the scope of pass : as in whether it stands a very generic fall back to CAS loops as final resort or whether its doing fixup using whatever is available at its disposal.

see here : 
https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/7f87ab006e3d9c275d31e8189f0790c88c72429b/mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp#L909-L952

krzysz00:
1) I'm not sure that smax/unim construction is NaN-correct and negative-zero correct (sure, in our setup, we don't care much, _but_ this goes upstream, so)
2) We may, in principle, end up with min and friends
2b) When address space 8 happens, there's a decent chance we'll end up with a generic operation that takes an `AtomicRMWKind` (I'd clean it up now, but backwards compatibility and since there'll be a new interface Real Soon Now (tm) ...)
3) Yeah, this pass is meant to be a generic fallback for any atomic an architecture might not support

manupak (approved):
fair enough -- in our cases we can assume NaNs will not be in the buffer (thus do checks on the src) or negative zeros might be swapped with pos zeros (and vice versa) -- which is fine.

I take your point on being the general fallback and upstream where those assumption may not hold or good enough.

Having a done detailed look I have no concerns but I would like to request a test for preservation of original attrs to cover that part of the code.

krzysz00:
I think I've got a test that preserves original attributes, there's some random `foo`s in the test designed to make sure they go through?

krzysz00:
https://reviews.llvm.org/D148722
and
https://reviews.llvm.org/D148724

PR #1050, opened 2023-04-18
[DO NOT SQUASH] Adds slice op input fusion support

manupak:
This commit adds lowering of slice op fused to
gemm-like ops by converting the eventually lowered
tensor.extract_slice op to rock.transform by slice.

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/863

### DIFF
--- a/external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
+++ b/external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
@@ -117,6 +117,17 @@ bool isTransposeOp(Operation *op) {
   return isa<tosa::TransposeOp, tosa::ReshapeOp>(op);
 }
 
+bool isSliceOp(Operation *op) { return isa<tosa::SliceOp>(op); }
+
+bool isConstSplatOp(Operation *op) {
+  if (tosa::ConstOp constOp = dyn_cast<tosa::ConstOp>(op)) {
+    if (constOp.getValue().isSplat()) {
+      return true;
+    }
+  }
+  return false;
+}
+
 bool isTransposeConfigConstant(Operation *op) {
   return op->hasTrait<OpTrait::ConstantLike>() &&
          llvm::any_of(op->getUsers(), [&](Operation *u) {
@@ -125,8 +136,8 @@ bool isTransposeConfigConstant(Operation *op) {
 }
 
 bool isAlwaysLeadingOp(Operation *op) {
-  return isConstantZero(op) || isTransposeOp(op) ||
-         isTransposeConfigConstant(op);
+  return isConstantZero(op) || isTransposeOp(op) || isSliceOp(op) ||
+         isConstSplatOp(op) || isTransposeConfigConstant(op);
 }
 
 bool isLeadingOp(Operation *op, bool trailingOnly) {
--- a/mlir/lib/Dialect/Rock/Transforms/ViewToTransform.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/ViewToTransform.cpp
@@ -86,6 +86,50 @@ struct ExpandShapeRewritePattern
   }
 };
 
+struct ExtractSliceRewritePattern
+    : public OpConversionPattern<tensor::ExtractSliceOp> {
+  using OpConversionPattern<tensor::ExtractSliceOp>::OpConversionPattern;
+
+  LogicalResult matchAndRewrite(tensor::ExtractSliceOp sliceOp,
+                                OpAdaptor adaptor,
+                                ConversionPatternRewriter &b) const final {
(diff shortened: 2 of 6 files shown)


external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
```diff
@@ -117,6 +117,17 @@ bool isTransposeOp(Operation *op) {
   return isa<tosa::TransposeOp, tosa::ReshapeOp>(op);
 }
 
+bool isSliceOp(Operation *op) { return isa<tosa::SliceOp>(op); }
+
+bool isConstantOp(Operation *op) {
```
krzysz00: Maybe rename this to `isSplatOp()`?

mlir/lib/Dialect/Rock/Transforms/ViewToTransform.cpp
```diff
@@ -86,6 +86,50 @@ struct ExpandShapeRewritePattern
...
+        return sliceOp.emitError("dynamic sizes are not supported.");
+      }
+    }
+
+    ArrayRef<int64_t> offsets = sliceOp.getStaticOffsets();
+    ArrayRef<int64_t> sizes = sliceOp.getStaticSizes();
+    ArrayRef<int64_t> strides = sliceOp.getStaticStrides();
+
+    for (int64_t stride : strides) {
+      if (stride != 1) {
```
krzysz00, line 117: Let's go ahead and support it? "pick every Nth element" is the function `(..., i, ...) -> (..., N * i, ...)`, which is just an Embed{}. This'd also let us support dilations in the Tosa interchange layer, I think.

The way I'd think to do this is to collect up all the non-unit slice dimensions, and then add a second `rock.transform` that applies the strides.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -1244,3 +1244,26 @@ TransformMapAttr mlir::rock::transformExpandShape(
...
+                                                   ArrayRef<int64_t> sizes) {
+  rock::BottomUpTMBuilder transform(b, inpShape, loc);
+  SmallVector<StringRef, 4> lowerNameRefs;
+  transform.getStartNames(lowerNameRefs);
+  SmallVector<SmallString<8>> upperNameStores;
+  SmallVector<StringRef, 4> upperNameRefs;
+  for (StringRef lowerName : lowerNameRefs) {
+    upperNameStores.emplace_back();
+    upperNameRefs.push_back(
+        (lowerName + Twine("_sliced")).toStringRef(upperNameStores.back()));
```
krzysz00, line 1261: Huh, that's a neat method.

krzysz00 (commented):
Do we want to also add this to `TransformToMemref`?

Or, better yet, now that we have `threadwise_read_into` etc ... do we even need `TransformToMemref`? We might want to switch up the utility kernels to allow for coordinate transforms... But that's a separate PR anyway.

Overall, none of this is blocking (though I'm feeling somewhat strongly about the `isConstant` rename) and I'm willing to approve this as is.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
sjw36: Would be nice to have a utility class that encapsulated/collected both the stores and the refs..

sjw36 (approved):
Looks great. Thanks for adding to partitioner as well!

mlir/lib/Dialect/Rock/Transforms/ViewToTransform.cpp
manupak: agreed with the suggestion though I dont want to do that given the on going things.
I've created a ticket for next sprint so that we dont forget  : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/870

external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp
manupak: IsConstSplatOp() ?

krzysz00: Yeah, even better name, good call.

manupak: done

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
manupak: Agreed. Im going to assume thats a seperate piece of work here for now :)

PR #1052, opened 2023-04-19
Avoid assuming the wave size during transformations

giuseros:
This is trying to avoid the issue of "assuming" `waveSize==64` causing issue when dealing with `wmma`. 

The only place where we still assume the wave size is in `MfmaInsnsGroup.h`. 

This is because `MfmaInsnsGroup` lives in `MLIRRockOps` and `MLIRUtility` uses `MLIRRockOps`, so I would create a cyclic dependency. 

For the time being, I decided to not care, because for now the `mfma` instruction is specific to architectures for which `waveSize` is 64.


mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -479,25 +480,25 @@ struct BlockwiseGemmV2RewritePattern
       }
     };
 
-    auto ldsToRegisterCopyKdim =
```
manupak: What is the change here ?

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
```diff
@@ -272,7 +270,8 @@ class PopulateParamsXDL : public BasePopulateParams<InitParamsXDL> {
 
   LogicalResult isValidBlockwiseGemmXDLOPS(const InitParamsXDL &param,
                                            Type dataTypeA, Type dataTypeB,
-                                           StringRef arch, uint32_t blockSize);
+                                           StringRef arch, int64_t waveSize,
```
manupak: nit : given we look up waveSize using arch, the expansion of func signature is to save that lookup ?

manupak (approved):
Looks good with a question and a nit.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: I think it's clang-format

krzysz00 (approved):
Re moving around the MfmaInstrGroup, we can probably factor the `Rock` utilities - maybe move the AmdArchDb and other stuff like that into something like `mlir/lib/Util/`.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
giuseros: Yep, exactly

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
giuseros: Yeah, I agree, it's probably better to look up again

giuseros:
> Re moving around the MfmaInstrGroup, we can probably factor the `Rock` utilities - maybe move the AmdArchDb and other stuff like that into something like `mlir/lib/Util/`.

Yeah like a place where no rock ops are used so we don't have any cycles

PR #1053, opened 2023-04-20
Gracefully not fail when transpose cannot be folded.

manupak:
There can be scenarios where tp cant be folded
such as a reshape collasing non-contigous dims
following a transpose.

This commit fixes/reverts op creations if the folding is not successful.


manupak:
somewhat related discussion : https://github.com/ROCmSoftwarePlatform/rocMLIR/discussions/1054

krzysz00 (commented):
Wait, so what's this actually fixing?

manupak:
This is fixing that bug of prematurely creating new collapse shape ops while the pass can still fail to fold.

krzysz00 (approved):
Yeah, makes sense, approved

manupak:
Thanks @krzysz00!

PR #1055, opened 2023-04-24
Add support for wmma in the rocMLIR frontend

giuseros:
This is adding WMMA support into our core passes:
- `GridwiseGemmToBlockwise`
- `BlockwiseGemmToThreadwise`
- `ThreadwiseGemmLowering`

I also add basic support to `rocmlir-gen` so that we can run the end-to-end flow. However, testing for now is only at the IR level. Runtime test should be added after this PR.

Please note that the changes to the backend are not part of this PR but they are part of (#1035)

The high level idea has been to implement a `AccelEmitter` class to abstract away details about the calls to the specific accelerator. Wmma will implement `WmmaEmitter` and mfma will immplement `MfmaEmitter`. 

Before approving this PR we should:
- Agree if this is the correct design
- If needed, select a name to replace XDLOPS (`Accelerator`? @sjw36 was suggesting `xma`) . The idea is to create a separate PR to do the name replacement
- Approve PR https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1035

After approving this PR we should:
- Add the correct default and tuning parameters
- Add runtime tests
- Improve `wmma` performance (e.g., select different `waveSize` during tuning time and using the `f16->f16` wmma instruction)

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -1055,8 +1055,6 @@ def Rock_BlockwiseGemmOp:
     $matrixC `+` `` `=` $matrixA `*` $matrixB attr-dict
     `:` type($matrixC) `+` `` `=` type($matrixA) `*` type($matrixB)
   }];
-
-  let hasVerifier = 1;
 }
 
 // This is less general than the one in the AMDGPU dialect, since it only
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -1063,9 +1063,9 @@ def Rock_BlockwiseGemmOp:
 // accounts for how we call these operations.
 defvar AccelArgTypes = [F32,
                     VectorOfLengthAndType<[2, 4], [F32]>,
-                    VectorOfLengthAndType<[2, 4, 8], [F16, BF16]>,
-                    VectorOfLengthAndType<[4, 8], [I8, F8E5M2FNUZ, F8E4M3FNUZ]>];
-defvar AccelResTypes = [VectorOfLengthAndType<[4, 16, 32], [F32, I32]>];
+                    VectorOfLengthAndType<[2, 4, 8, 16], [F16, BF16]>,
+                    VectorOfLengthAndType<[4, 8, 16], [I8, F8E5M2FNUZ, F8E4M3FNUZ]>];
+defvar AccelResTypes = [VectorOfLengthAndType<[4, 8, 16, 32], [F32, I32]>];
 
 // blockwise_gemm_accel
 def Rock_BlockwiseGemmAccelOp:
@@ -1078,6 +1078,7 @@ def Rock_BlockwiseGemmAccelOp:
                    MemRefOf<AccelArgTypes>:$bufferB,
                    MemRefOf<AccelResTypes>:$matrixC,
                    StrAttr:$arch,
+                   Rock_GemmFeaturesAttr:$features,
                    I32Attr:$blockSize,
                    RockAccelTuningParamAttrInterface:$params)>{
   let summary = "Blockwise GEMM accelerated version";
@@ -1091,7 +1092,7 @@ def Rock_BlockwiseGemmAccelOp:
   }];
   let assemblyFormat = [{
     $matrixC `+` `` `=` $bufferA `from` $matrixA `[` $waveOffsetA `]` `*`
-                        $bufferB `from` $matrixB `[` $waveOffsetB `]` attr-dict
+                        $bufferB `from` $matrixB `[` $waveOffsetB `]` `features` `=` $features attr-dict
     `:` type($matrixC) `+` `` `=` type($bufferA) `from` type($matrixA) `*`
                                   type($bufferB) `from` type($matrixB)
   }];
@@ -1128,6 +1129,7 @@ def Rock_AccelGemmOp:
                    MemRefRankOf<AccelArgTypes, [1]>:$matrixB,
                    MemRefRankOf<AccelResTypes , [1]>:$matrixC,
                    StrAttr:$arch,
+                   Rock_GemmFeaturesAttr:$features,
                    RockAccelTuningParamAttrInterface:$params)> {
   let summary = "Accelerated GEMM";
   let description = [{
@@ -1137,7 +1139,7 @@ def Rock_AccelGemmOp:
     Matrices A and B reside in LDS, the buffers live in registers, C is a vector
   }];
   let assemblyFormat = [{
-    $matrixC `+` `` `=` $matrixA`[`$mRepeat`]` `*` $matrixB`[`$nRepeat`]` attr-dict
+    $matrixC `+` `` `=` $matrixA`[`$mRepeat`]` `*` $matrixB`[`$nRepeat`]` `features` `=` $features attr-dict
     `:` type($matrixC) `+` `` `=` type($matrixA) `*` type($matrixB)
   }];
   let hasVerifier = 1;
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -1055,6 +1055,8 @@ def Rock_BlockwiseGemmOp:
     $matrixC `+` `` `=` $matrixA `*` $matrixB attr-dict
     `:` type($matrixC) `+` `` `=` type($matrixA) `*` type($matrixB)
   }];
+
+  let hasVerifier = 1;
 }
 
 // This is less general than the one in the AMDGPU dialect, since it only
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -1065,7 +1065,7 @@ defvar AccelArgTypes = [F32,
                     VectorOfLengthAndType<[2, 4], [F32]>,
                     VectorOfLengthAndType<[2, 4, 8, 16], [F16, BF16]>,
                     VectorOfLengthAndType<[4, 8, 16], [I8, F8E5M2FNUZ, F8E4M3FNUZ]>];
-defvar AccelResTypes = [VectorOfLengthAndType<[4, 8, 16, 32], [F32, I32]>];
+defvar AccelResTypes = [VectorOfLengthAndType<[4, 8, 16, 32], [F32, I32, F16, BF16]>];
 
 // blockwise_gemm_accel
 def Rock_BlockwiseGemmAccelOp:
--- a/mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
+++ b/mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
@@ -26,8 +26,8 @@ static Type getRetType(Type inputType) {
 }
 
 FailureOr<WmmaInsn> WmmaInsn::select(mlir::Type elementTypeA,
-                                     mlir::Type elementTypeB, int64_t mPerWave,
-                                     int64_t nPerWave) {
+                                     mlir::Type elementTypeB, int64_t waveSize,
+                                     int64_t mPerWave, int64_t nPerWave) {
   LLVM_DEBUG(llvm::dbgs() << "Invoke Wmma group selection:\n"
                           << "elementTypeA: " << elementTypeA << "\n"
                           << "elementTypeB: " << elementTypeB << "\n"
@@ -37,9 +37,10 @@ FailureOr<WmmaInsn> WmmaInsn::select(mlir::Type elementTypeA,
   if (elementTypeA != elementTypeB)
     return failure();
 
-  int64_t inputLen = 16;
+  if (waveSize != 32)
+    return failure();
 
-  // We are assuming waveSize==32
+  int64_t inputLen = 16;
   int64_t outLen = 8;
   int64_t outStride = 2;
 
--- a/mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
+++ b/mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
@@ -17,7 +17,7 @@
 using namespace mlir;
 using namespace mlir::rock;
 
-Type getRetType(Type inputType) {
+static Type getRetType(Type inputType) {
   Builder b(inputType.getContext());
   if (inputType.isInteger(8))
     return b.getI32Type();
--- a/mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
+++ b/mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
@@ -1,0 +1,67 @@
+
+#include "mlir/Dialect/Rock/IR/WmmaInsnGroup.h"
+
+#include "mlir/Dialect/AMDGPU/IR/AMDGPUDialect.h"
+#include "mlir/Dialect/LLVMIR/ROCDLDialect.h"
+#include "mlir/Dialect/Rock/utility/AmdArchDb.h"
+#include "mlir/Dialect/Rock/utility/math.h"
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/BuiltinTypes.h"
+
+#include "llvm/Support/Debug.h"
+#include "llvm/Support/ErrorHandling.h"
+#include <cstdint>
+
+#define DEBUG_TYPE "rock-wmma-insn-group"
+
+using namespace mlir;
+using namespace mlir::rock;
+
+Type getRetType(Type inputType) {
+  Builder b(inputType.getContext());
+  if (inputType.isInteger(8))
+    return b.getI32Type();
+
+  return b.getF32Type();
+}
+
+FailureOr<WmmaInsn> WmmaInsn::select(mlir::Type elementTypeA,
+                                     mlir::Type elementTypeB, StringRef arch,
+                                     int64_t mPerWave, int64_t nPerWave) {
+  LLVM_DEBUG(llvm::dbgs() << "Invoke Wmma group selection:\n"
+                          << "elementTypeA: " << elementTypeA << "\n"
+                          << "elementTypeB: " << elementTypeB << "\n"
+                          << "arch: " << arch << "\n"
+                          << "mPerWave: " << mPerWave << "\n"
+                          << "nPerWave: " << nPerWave << "\n");
+
+  if (!arch.contains("gfx11"))
+    return failure();
+
+  if (elementTypeA != elementTypeB)
+    return failure();
+
+  int64_t inputLen = 16;
+  int64_t outLen = 8;
+
+  int64_t mRepeats = mPerWave / inputLen;
+  int64_t nRepeats = nPerWave / inputLen;
+
+  VectorType argTypeA = VectorType::get({inputLen}, elementTypeA);
+  VectorType argTypeB = VectorType::get({inputLen}, elementTypeB);
+  VectorType retType = VectorType::get({outLen}, getRetType(elementTypeA));
+
+  StringRef insn;
+  if (elementTypeA.isF16()) {
+    insn = ROCDL::wmma_f32_16x16x16_f16::getOperationName();
+  } else if (elementTypeA.isBF16()) {
+    insn = ROCDL::wmma_f32_16x16x16_bf16::getOperationName();
+  } else if (elementTypeA.isInteger(8)) {
+    insn = ROCDL::wmma_i32_16x16x16_iu8::getOperationName();
+  } else {
+    return failure();
+  }
+
+  return WmmaInsn{insn,     inputLen, outLen,   mRepeats,
+                  nRepeats, argTypeA, argTypeB, retType};
+}
--- a/mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
+++ b/mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
@@ -26,23 +26,22 @@ static Type getRetType(Type inputType) {
 }
 
 FailureOr<WmmaInsn> WmmaInsn::select(mlir::Type elementTypeA,
-                                     mlir::Type elementTypeB, StringRef arch,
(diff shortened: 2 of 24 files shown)


mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1063,9 +1063,9 @@ def Rock_BlockwiseGemmOp:
 // accounts for how we call these operations.
 defvar MfmaArgTypes = [F32,
                     VectorOfLengthAndType<[2, 4], [F32]>,
-                    VectorOfLengthAndType<[2, 4, 8], [F16, BF16]>,
-                    VectorOfLengthAndType<[4, 8], [I8, F8E5M2FNUZ, F8E4M3FNUZ]>];
-defvar MfmaResTypes = [VectorOfLengthAndType<[4, 16, 32], [F32, I32]>];
+                    VectorOfLengthAndType<[2, 4, 8, 16], [F16, BF16]>,
+                    VectorOfLengthAndType<[4, 8, 16], [I8, F8E5M2FNUZ, F8E4M3FNUZ]>];
+defvar MfmaResTypes = [VectorOfLengthAndType<[4, 8, 16, 32], [F32, I32, F16, BF16]>];
```
krzysz00: Maybe `AccelResTypes` and so on while we're here?

mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "llvm/Support/Debug.h"
+#include "llvm/Support/ErrorHandling.h"
+#include <cstdint>
+
+#define DEBUG_TYPE "rock-wmma-insn-group"
+
+using namespace mlir;
+using namespace mlir::rock;
+
+Type getRetType(Type inputType) {
```
krzysz00: Nit: `static`

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -0,0 +1,431 @@
...
+  toMatrixC.embed("gemmN", 2, N, {"n", "wave_n", "n_i", "blk_col", "n_tid"},
+                  {nPerBlock, nPerWave, nPerRepeat, n, 1});
+  TransformMapAttr toMatrixCAttr = toMatrixC.get();
+
+  ArrayAttr idToMatrixCMaps =
+      b.getArrayAttr({splitMemoryCoordsAttr, toRowsAndColsAttr, toMatrixCAttr});
+  return idToMatrixCMaps;
+}
+
+Value MfmaEmitter::computeOutputConversion(PatternRewriter &b, Location loc,
```
krzysz00: Note to future us: if we ever do the "output type conversions as fusions" bit, we may not need this.

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -0,0 +1,431 @@
...
+      VectorType destVectorType = accVectorType.clone(destType);
+      cast = createTypeConversionOp(b, loc, loaded, destVectorType);
+    }
+    b.create<InBoundsStoreOp>(loc, cast, regDest,
+                              convertLoop.getLowerCoords(/*domain*/ 1));
+  }
+  return regDest;
+}
+
+Value MfmaEmitter::computeLdsSourceOffset(OpBuilder &kb, Value k_i,
```
krzysz00: Can we clean up these variable names? For one, `MN` -> `d` per gridwise conventions, but also, some of these could use lenghtening now that they're a meaningfully separate function

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -0,0 +1,431 @@
...
+                          {mRepeats, nRepeats, retNumElements});
+  TransformMapAttr splitMemoryCoordsAttr = splitMemoryCoords.get();
+
+  auto toMatrixC =
+      TopDownTMBuilder::below(splitMemoryCoords, splitMemoryCoordsAttr);
+
+  toMatrixC.passThrough({"gemmG"}, {0}, {"g"});
+
+  toMatrixC.embed("gemmM", 1, M, {"m", "wave_m", "m_tid", "rep_i", "item_i"},
+                  {mPerBlock, mPerWave, 1, wmmaInsn.inputLen, 2});
```
krzysz00: Something about these coefficients feels off, mainly that `m_tid = 2` and `item_i = 0` is the same is `m_tid = 1, item_i = 2`

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -0,0 +1,431 @@
...
+                  {nPerBlock, nPerWave, 1, wmmaInsn.inputLen});
+
+  TransformMapAttr toMatrixCAttr = toMatrixC.get();
+
+  ArrayAttr idToMatrixCMaps =
+      b.getArrayAttr({splitMemoryCoordsAttr, toMatrixCAttr});
+  return idToMatrixCMaps;
+}
+
+Value WmmaEmitter::computeOutputConversion(PatternRewriter &b, Location loc,
```
krzysz00: Since this code is basically the same in both paths, is there a way to factor it out some?

krzysz00 (commented):
Minor nitpicking here, but, overall, I think this is a good design I'd be happy to adopt

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
giuseros: But `m_tid` is liimited to 0 or 1 (in the case of wave32. It would be 0,1,2,3 for wave64). Basically every workitem is computing 8 values, which represents a sub-column of a 16x16 tile. For workitems 0 to 15 `m_tid` is 0 and they write to 0, 2, 4, 6, ..., 14. For workitems 16 to 31 `m_tid` is 1 and they write at positions 1,3,5,..,15 . This is outlined here: 

image

I will add a better comment to that transformation

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
giuseros: I can, but my idea (feel free to disagree) was to park this while doing a separate PR about renaming, where I can take care of renaming types, operations, etc...

sjw36: Let's have the renaming discussion on the next PR.

krzysz00: Agreed.

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
krzysz00: Ok, having looked at the diagram, this makes sense. However, given that, can we, instead of having that magic `2`, have a value called something like `itemsPerRegister` that's a function of the wave size? That'll make it easier to understand where we got that constant.

krzysz00 (commented):
The above was a minor nitpick, this is, effectively, approved, I just want to land the upstreamable part of this first

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
jerryyin: Are we too relaxed about the type and size constraint? I don't think all combinations can work?

mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
```diff
@@ -0,0 +1,67 @@
...
+                                     int64_t mPerWave, int64_t nPerWave) {
+  LLVM_DEBUG(llvm::dbgs() << "Invoke Wmma group selection:\n"
+                          << "elementTypeA: " << elementTypeA << "\n"
+                          << "elementTypeB: " << elementTypeB << "\n"
+                          << "arch: " << arch << "\n"
+                          << "mPerWave: " << mPerWave << "\n"
+                          << "nPerWave: " << nPerWave << "\n");
+
+  if (!arch.contains("gfx11"))
+    return failure();
```
jerryyin: This check probably shouldn't be here. It'd be too late if we try to select the wmma on a wrong architecture and fail. It should be caller's responsibility to make sure right select is invoked... Therefore, get rid of this argument?

mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
```diff
@@ -0,0 +1,67 @@
...
+  StringRef insn;
+  if (elementTypeA.isF16()) {
+    insn = ROCDL::wmma_f32_16x16x16_f16::getOperationName();
+  } else if (elementTypeA.isBF16()) {
+    insn = ROCDL::wmma_f32_16x16x16_bf16::getOperationName();
+  } else if (elementTypeA.isInteger(8)) {
+    insn = ROCDL::wmma_i32_16x16x16_iu8::getOperationName();
+  } else {
+    return failure();
+  }
```
jerryyin, line 63: [Nit] I'd put those in a static map just like mfma insn header does. But since this is only 3 entries, I'll leave this as a nit-pick...

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
```diff
@@ -179,7 +179,8 @@ void rock::buildBackendPipeline(OpPassManager &pm,
    */
   pm.addPass(createStripDebugInfoPass());
   pm.addPass(createLowerGpuOpsToROCDLOpsPass(
-      options.chip, options.indexBitwidth, /*useBarePtrCallConv=*/true));
+      options.chip, options.indexBitwidth, /*useBarePtrCallConv=*/true,
+      gpu::amd::Runtime::HIP));
```
jerryyin: No, this will break MIOpen opencl backend. Please revert.

jerryyin (commented):
Posting the first half of the review... This is a massive PR.

mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
```diff
@@ -105,7 +105,8 @@ void AffixTuningParameters::affixTuningParametersImpl(
     perfConfig = perfConfigAttr.getValue().str();
   }
   GemmFeatures features = op.getGemmFeatures();
-  if (bitEnumContainsAll(features, GemmFeatures::mfma)) {
+  if (bitEnumContainsAll(features, GemmFeatures::mfma) ||
+      bitEnumContainsAll(features, GemmFeatures::wmma)) {
```
jerryyin, line 109: You will continue to review the tuning details after this PR, right? I don't quite expect the tuning parameter derive logic to be all compatible.

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.h
```diff
@@ -0,0 +1,197 @@
...
+  //
+  // Tuning parameters
+  //
+  int64_t mPerBlock;
+  int64_t nPerBlock;
+  int64_t kPerBlock;
+  int64_t mPerWave;
+  int64_t nPerWave;
+  int64_t kPack;
+  int64_t waveSize;
```
jerryyin: Can we reuse the tuning parameters structure instead of making a copy of every parameter here.

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.h
```diff
@@ -0,0 +1,197 @@
...
+  // numOutputVectorElements =
+  // nResultVectors*mRepeats*nRepeats*accVectorType.getNumElements()
+  int64_t numOutputVectorElements;
+
+  //
+  // Accelerator data types
+  //
+  Type argTypeA;            // Type of the arguments (might be scalar or vector)
+  Type argTypeB;            // Type of the arguments (might be scalar or vector)
+  VectorType accVectorType; // Accumulator vector type (always vector type)
```
jerryyin: I have strong opinions on not keeping all states as public members as it will become impossible to track, and becomes a maintenance nightmare very soon. Also, I believe it is wrong that you are creating data members from parent class only to have child class to fill them in.

jerryyin (changes requested):
I like the idea of AccelEmitter but I am strongly against the current class design of AccelEmitter. It's tracking too much state. I'm putting a blocker now in case this get merged prematurely.

mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
giuseros: Yes, after this PR I will open another one about tuning

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
giuseros: Done, but I used `outputStride` and I set it in the `wmmaInsn`

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
giuseros: So these are used in blockwisegemmV2 and threadwisegemmV2 which don't even have a verifier. My assumption was that it is ok to be not too strict here, because it's an internal operator and there is no risk in doing anything wrong because any length/type mismatch would be caught later.

mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
giuseros: I moved into the `select` fuction

mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
giuseros: Yeah I  think that a map is a bit too much for just three entries. I will probably do it when we experiment with waveSize != 32 (which will add more entries to the table).

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.h
giuseros: Done. I guess it's a matter of preference, because now I have to extract those parameters locally in different functions (it's no big deal, just style)

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.h
giuseros: Good point. So I created a separate struct `AccelEmitterParameters` that now contains those parameters and each subclass has its own `initAccelEmitterParameters` which is called at construction. I tend to like less structure, but I hope this middle ground solution works for you

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -0,0 +1,468 @@
...
+  return sourceOffset;
+}
+
+void WmmaEmitter::emitThreadwiseLoop(OpBuilder &b, Location loc, Value argA,
+                                     Value argB, Value bufferC,
+                                     Value regCOffset) {
+  VectorType vectorType = wmmaInsn.retType;
+  auto vectorC = b.create<memref::LoadOp>(loc, vectorType, bufferC, regCOffset);
+
+  auto mfma = b.create<amdgpu::WMMAOp>(loc, vectorType, argA, argB, vectorC);
```
jerryyin: Ok, I think `clamp` should always be set to true because we want it to saturate instead of overflow?

Also, what about `subwordOffset` field? So we are by hard code only reading from `0, 2, 4, ..., 14` indices with this field on? I guess I don't think I have a good understanding about why this field matters (so it is okay that you are hard coding this behavior?).

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -0,0 +1,468 @@
...
+      b.getArrayAttr({splitMemoryCoordsAttr, toMatrixCAttr});
+  return idToMatrixCMaps;
+}
+
+std::unique_ptr<AccelEmitter>
+AccelEmitter::select(GemmFeatures features, Type dataTypeA, Type dataTypeB,
+                     StringRef arch, XdlopsGemmParamsAttr tuningParams) {
+  bool isMfma = rock::bitEnumContainsAll(features, GemmFeatures::mfma);
+  bool isWmma = rock::bitEnumContainsAll(features, GemmFeatures::wmma);
+  if (isMfma && arch.contains("gfx90")) {
```
jerryyin: `arch.contains()` shouldn't be necessary because you are not trusting features enum here. Suggest remove it from here and below.

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -0,0 +1,468 @@
...
+    : AccelEmitter{arch, tuningParams,
+                   initAccelEmitterParams(mfmaGroup, tuningParams)},
+      mfmaGroup{mfmaGroup} {
+  MfmaInsnAttr mfmaAttr = mfmaGroup.getInsnAttr();
+
+  // Specific mfma parameters
+  int64_t blocksInOutRegs = mfmaAttr.blocksInOutRegs;
+  inputSpansPerMfmaIn = mfmaAttr.inputSpansPerMfmaIn;
+  inputSpanLen = mfmaAttr.inputSpanLen;
+  isKReduction = (blocksInOutRegs == 1) && (inputSpansPerMfmaIn > 1);
```
jerryyin: Add `isKReduction` into the mfmaAttr? I think the fact that you are computing this again justified it.

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -0,0 +1,468 @@
...
+  sourceOffset = dBuilder.create<AddIOp>(
+      loc, sourceOffset, dBuilder.create<MulIOp>(loc, inputLen, d_i));
+  sourceOffset = kBuilder.create<AddIOp>(
+      loc, sourceOffset, kBuilder.create<MulIOp>(loc, dPerBlock, k_i));
+  return sourceOffset;
+}
+
+void WmmaEmitter::emitThreadwiseLoop(OpBuilder &b, Location loc, Value argA,
+                                     Value argB, Value bufferC,
+                                     Value regCOffset) {
```
jerryyin, line 354: [Comment] Ok, so there's no such thing as nResultVectors in wmma side because we always populate the 16x16x16 threadwise tile sizes. And the reason why you can make such simplification is due to all wmma has same shape, therefore arbitrary tile size can solely be computed from repeat. We can make same simplification if we only use reduction xdlops on mfma too.

[Nit] Speaking of which, `nResultVecotrs` parameter kept in `AccelEmitterParams` can be confusing, because wmma compute doesn't need it. I'd move it to be a mfma thing only.

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -0,0 +1,468 @@
...
+  return params;
+}
+
+Value WmmaEmitter::computeLdsSourceOffset(OpBuilder &kBuilder, Value k_i,
+                                          OpBuilder &dBuilder, Value d_i,
+                                          OpBuilder &builder, Value dPerBlock,
+                                          Location loc, Value sourceBase,
+                                          Value laneId) {
+
+  // srcOffset = k_i * MN + (laneId % wmmaInputLen) + wmmaInputLen * mn_i;
```
jerryyin, line 341: [Question] Looks very suspicious that this can give us the right result. Could you remind me how this works? It looks to me that if we have 64 lanes and wmmaInputLen = 16. Then:
 - lane0's offset = k_i * MN + 0 + 16 * mn_i
 - lane16's offset = k_i * MN + 0 + 16 * mn_i

Since there are 4 x 16 lanes within each wave, are they just duplicating in loading from the same place 4 times?

mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
jerryyin: If that's the case could you add an assert somewhere just to make sure wave size is indeed 32? The switch between 32 and 64 is only governed by size of the output?

jerryyin (commented):
Sorry for splitting into multiple reviews, just had time to look at the `AccelEmitter` implementation. Your updates are making the data structure much better now. Thank you!

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
giuseros: Yes, this is exactly how `wmma` works (https://gpuopen.com/learn/wmma_on_rdna3/)!

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
giuseros: The problem is that `nResultVectors` is queried in our passes. This is why I need to have it in both

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
giuseros: Yes, so the intrinsic works exactly in this way: for `fp16` it only fills half of the VGPRs they use. It can fill the top half or the bottom half. So the output is 8 VGPRs which means vector of 16 fp16 elements. If `subwordOffset` is 0, then the top half is used (hence 0, 2, .. 14) , while if it is 1, the bottom half is used.

mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
giuseros: Yes, you are right the switch is only governed by the size of the output, so probably there will still be only three entries in  (while `outLen` and `outStride` will change).

jerryyin (approved):
Talked with @giuseros offline and who cleared my last doubts about LDS->register read.

LGTM and thanks for your work!

krzysz00 (approved):
All this looks good to me too

giuseros:
Thanks for the review guys. Question: should we merge this given that there is PR #1058  that is taking care of naming etc...?

krzysz00:
Wait, hold on, is the renaming PR landing first or is this landing first?

(I think either order makes sense)

giuseros:
Yeah, at the end I landed the renaming first and rebased this one on top

PR #1056, opened 2023-04-24
Add unet conv and gemm configs to perfRunner.py invocation, along with bert.

pcf000:
I believe this is what Manupa asked for -- running perfRunner.py with the collection of configs for unet, separate collections for conv and gemm.  Please let me know if it needs to be adjusted.


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -851,7 +851,13 @@ pipeline {
                                 // Run MLIR perf benchmarks
                                 sh 'python3 ./bin/perfRunner.py'
                                 sh """python3 ./bin/perfRunner.py --op=gemm --batch_all \
-					--configs_file=${WORKSPACE}/mlir/utils/performance/bert-configs \
+                                       --configs_file=${WORKSPACE}/mlir/utils/performance/bert-configs \
+                                       --tuning_db=${WORKSPACE}/mlir_tuning_${CHIP}.tsv""" \
```
manupak: This tuning db does not incoporate unet instances.
It needs to be added here as well : 

https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/a997d5f51314b45d7a4c04f1599966dcf53f9b4d/mlir/utils/jenkins/Jenkinsfile#L603-L611

Need same for convs.

manupak (dismissed):
Thanks @pcf000 !

The tuning has to be updated as well..

@giuseros how do we visualize these bert gemm perf ? (I am only familiar with MIOpen comparisons)

mlir/utils/jenkins/Jenkinsfile
manupak: The tsv's are appended

pcf000: I think I've done what you asked in the next commit.  Running a private Jenkins test now.

krzysz00 (commented):
Why not, instead of this, have one file `gemm-configs` that comments where they came from?

pcf000:
> Why not, instead of this, have one file `gemm-configs` that comments where they came from?

Because Manupa asked me to separate them, specifically "gemms and convs should be provided for separate runs."

krzysz00:
Ok, so, in that case, we should have a `gemm-configs` and a `conv-comfigs`, which comments showing the source, and then separate `.tsv` files for the gemms and the convs, probably.

Heck, what we could do is take `resnet50-configs`, rename it to `conv-configs`, and run the unet convolutions through the MIOpen tuning infrastructure.

Or start running `resnet50` through our own setup.

manupak:
I agree to unifying the configs as : conv-configs and gemm-configs (that we care about) with comments indicating the source.

I only meant the configs are meant to be seperate just because the scripts expect gemms and convs to be tuned/measured separately.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -606,9 +606,15 @@ pipeline {
...
                                 sh """python3 ./bin/tuningRunner.py --operation gemm \
---configs_file=../mlir/utils/performance/bert-configs \
---output=mlir_tuning_${ARCH}.tsv
-[ -f mlir_tuning_${ARCH}.tsv ]"""
+                                        --configs_file=../mlir/utils/performance/bert-configs \
+                                        --output=mlir_tuning_${ARCH}.tsv
+                                      python3 ./bin/tuningRunner.py --operation gemm \
+                                        --configs_file=../mlir/utils/performance/unet-gemm-configs \
+                                        --output=mlir_tuning_${ARCH}.tsv
+                                      python3 ./bin/tuningRunner.py --operation conv \
```
manupak, line 611: This sits under "Tune Gemm" which is not right.
Maybe we should either broaden the stage name or add a seperate stage.

manupak (commented):
Just a nit about stage naming.

I think it might be just easier to rename them to be gemm-configs and conv-configs (and append the configs) as @krzysz00 suggested. The requirement was just to do continuous comparisons.

Im fine with either approach.

mlir/utils/jenkins/Jenkinsfile
krzysz00: Yeah, "rocMLIR tuning" could be a good name here.

mlir/utils/performance/conv-configs
```diff
@@ -0,0 +1,17 @@
+conv -F 1 -f NCHWG -I NCHWG -O NCHWG -n 2 -c 1280 -H 32 -W 32 -k 640 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -g 1
```
manupak, line 2: IIUC, this file stands as a generic conv-configs. Therefore, it would be best that we put a comment indicating the source of these configs (like in gemm configs).

Something as future work for us :
Also noting we have 4 other similiar files which are defined per model (resnet50 and resnext101). Though I dont personally see that it is necessary bring them under here in the PR, it would be good to set that as the direction we want to go (prolly in a future PR). The 4 files seems to be coming from : model x {nightly,pr} combinations. If we are agreeing on unifying the filenames be to model agnostic, we should probably be renaming these to be two files along the lines (or similiar) : 

* conv_configs
* conv_configs_pr_subset

manupak (approved):
This all looks good to me bar adding a comment to state source where conv-configs came from.

mlir/utils/performance/conv-configs
manupak: (I dont see resnext101 configs even used, if I m not missing anything)

PR #1057, opened 2023-04-24
Add infrastructure for tracking fusion performance

yiqian1:
Add tuning fusion test cases in weekly CI.
Testing and reporting of fusion performance are currently disabled until a tuning DB is availabe. Currently the performance report only shows 'TFLops'. May include other performance metrics later.

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -600,7 +600,7 @@ pipeline {
                             showEnv()
                         }
                     }
-                    stage("Tune Gemm") {
+                    stage("Tune Gemm and Fusion") {
                         steps {
                             buildProject('check-rocmlir-build-only ci-performance-scripts', '')
                             dir('build') {
@@ -609,10 +609,19 @@ pipeline {
 --configs_file=../mlir/utils/performance/bert-configs \
 --output=mlir_tuning_${ARCH}.tsv
 [ -f mlir_tuning_${ARCH}.tsv ]"""
+
+                                // Tune resnet50
+                                sh """python3 ./bin/tuningRunner.py --op fusion \
+--test_dir ../mlir/test/fusion/e2e/resnet50/ -o tuning_fusion_resnet50_${ARCH}.tsv"""
+
+                                // Tune bert
+                                sh """python3 ./bin/tuningRunner.py --op fusion \
+--test_dir ../mlir/test/xmir/e2e/bert-torch-tosa/ -o tuning_fusion_bert_${ARCH}.tsv"""
                             }
                             sh 'rm build/CMakeCache.txt'
                         }
                     }
+
                     stage("Build MIOpen with librockCompiler") {
                         steps {
                             buildMIOpenWithMLIR()
@@ -626,9 +635,9 @@ pipeline {
                         steps {
                             tuneMLIRKernels()
                             // Save user database for nightly jobs
-                            // Hack: move the GEMM tuning DB to MIOpenUserDB
+                            // Hack: move the GEMM/Fusion tuning DB to MIOpenUserDB
                             // to make all the stashes work
-                            sh "cp ${WORKSPACE}/build/mlir_tuning_${ARCH}.tsv ${WORKSPACE}/MIOpen/build/MIOpenUserDB/"
+                            sh "cp ${WORKSPACE}/build/*_${ARCH}.tsv ${WORKSPACE}/MIOpen/build/MIOpenUserDB/"
                             dir ('MIOpen/build/MIOpenUserDB') {
                                 stash name: "MLIR-PerfDB-${ARCH}", includes: "**"
                             }
@@ -790,7 +799,7 @@ pipeline {
                                 selector: lastSuccessful(),\
                                 target: 'MIOpen/build/MIOpenUserDB'
                                 sh 'ls MIOpen/build/MIOpenUserDB'
-                                // Move GEMM tuning DB out of the way if it exists
+                                // Move GEMM and fusion tuning DB out of the way if it exists
                                 // This is needed because the tuned vs. untuned MIOpen
                                 // code will delete the MIOpenUserBD directory
                                 catchError (buildResult: null) {
@@ -856,6 +865,18 @@ pipeline {
                             }
                         }
                     }
+                    //stage("Test Fusion") {
+                    //    steps {
+                    //        dir('build') {
+                                // Run fusion resnet50 perf benchmarks
+                                //sh """python3 ./bin/perfRunner.py --op=fusion \
+          //--test_dir=${WORKSPACE}/mlir/test/fusion/e2e/resnet50/ --tuning_db=${WORKSPACE}/tuning_fusion_resnet50_${CHIP}.tsv"""
+                                // Run bert perf benchmarks
+                                //sh """python3 ./bin/perfRunner.py --op fusion \
+          //--test_dir=${WORKSPACE}/mlir/test/xmir/e2e/bert-torch-tosa/ --tuning_db=${WORKSPACE}/tuning_fusion_bert_${CHIP}.tsv"""
+                            //}
+                        //}
+                    //}
                     stage("Test MLIR vs CK") {
                         when {
                             beforeAgent true;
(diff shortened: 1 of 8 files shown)


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -600,7 +600,7 @@ pipeline {
                             showEnv()
                         }
                     }
-                    stage("Tune Gemm") {
+                    stage("Tune Gemm and Fusion") {
```
sjw36: Please make a separate stage.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -856,6 +865,18 @@ pipeline {
                             }
                         }
                     }
+                    //stage("Test Fusion") {
```
sjw36, line 872: Remove commented code.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -896,6 +917,7 @@ pipeline {
                                 sh 'ls -l'
                                 sh 'python3 ./bin/createPerformanceReports.py ${CHIP}'
                                 sh 'python3 ./bin/createGemmPerformanceReports.py ${CHIP}'
+                                //sh 'python3 ./bin/createFusionPerformanceReports.py ${CHIP}'
```
sjw36, line 924: why is this disabled?

sjw36 (commented):
What does this compare against? Would be good to compare vs just the anchor op perf.

Also a few minor changes.

mlir/utils/jenkins/Jenkinsfile
krzysz00: It's the other approach to not running this script until we have a tuning database, which is that we'll uncomment this after the database is populated in weekly

mlir/utils/performance/perfRunner.py
```diff
@@ -793,6 +791,144 @@ def benchmarkMIOpenWithMLIRKernels(configs, arch, filename, paths: Paths):
...
+        # find the best perf_config
+        bestPerf =""
+        if tuningDb:
+            if (arch, testVector) in tuningDb:
+               bestPerf = tuningDb[arch, testVector]
+               config.setPerfConfig(bestPerf)
+            else: # Tuning DB present but doesn't contain config, return N/A
+               oneEntry = config.tableEntry(np.nan)
+               oneEntry['FileName'] = filename
+               allData.append(oneEntry)
```
krzysz00, line 922: Can you comment this logic. It feels like it's adding an extra row to the table

mlir/utils/performance/perfRunner.py
```diff
@@ -793,6 +791,144 @@ def benchmarkMIOpenWithMLIRKernels(configs, arch, filename, paths: Paths):
...
+               continue;
+
+        # find the name of test function
+        futName = None
+        with open(filename, 'r') as f:
+            for line in f:
+              if re.search(r"rocmlir-gen", line):
+                  words = line.split()
+                  futName = words[words.index("-fut") + 1]
```
krzysz00: Break out of the loop after this at the very least

mlir/utils/performance/perfRunner.py
```diff
@@ -793,6 +791,144 @@ def benchmarkMIOpenWithMLIRKernels(configs, arch, filename, paths: Paths):
...
+               config.setPerfConfig(bestPerf)
+            else: # Tuning DB present but doesn't contain config, return N/A
+               oneEntry = config.tableEntry(np.nan)
+               oneEntry['FileName'] = filename
+               allData.append(oneEntry)
+               continue;
+
+        # find the name of test function
+        futName = None
+        with open(filename, 'r') as f:
```
krzysz00: Is there a way to extract this data from the file earlier so we don't have to loop through the whole thing?

mlir/utils/performance/reportUtils.py
```diff
@@ -21,6 +24,7 @@
                         'X', 'DilationH', 'DilationW', 'StrideH', 'StrideW',
                         'PaddingH', 'PaddingW', 'PerfConfig']
 GEMM_TEST_PARAMETERS = ['DataType', 'Chip', 'TransA', 'TransB', 'G', 'M', 'K', 'N', 'PerfConfig']
+FUSION_TEST_COL = ['TestFile', 'PerfConfig', 'Time']
```
krzysz00: This name doesn't seem to be used anywhere else

krzysz00 (commented):
I think I've found a few bugs in here

mlir/utils/performance/perfRunner.py
yiqian1: If the tuning DB doesn't contain the config, add a row with Nan for that test file into the table, such that we know something wrong with the test case.

krzysz00: Yeah but it still then runs the test

krzysz00: I think the right thing to do is to set the perf config to "" or NaN and then continue, letting the row be added naturally

yiqian1: > Yeah but it still then runs the test

No, it skips the test because there is a 'continue' in next line.

yiqian1: > I think the right thing to do is to set the perf config to "" or NaN and then continue, letting the row be added naturally

 If we run the test with an empty perf config, the test will use the heuristic config. That's not what we want.

krzysz00: No, it is what we want - that's the fallback path for missing gemm tuning records, for example

yiqian1: That's what we want in real world. But when we're running validations, don't we want to catch any unexpected issues? And benchmarkMLIR() does the same, i.e. mark the test case with NaN (or FAILED) when the best config doesn't exist.

mlir/utils/jenkins/Jenkinsfile
yiqian1: Have made a separate stage.

mlir/utils/performance/perfRunner.py
yiqian1: Yes.

mlir/utils/performance/reportUtils.py
yiqian1: Removed.

yiqian1:
> What does this compare against? Would be good to compare vs just the anchor op perf.
> 
Currently it just shows the performance without comparisons. I will add comparisons in next PR.

mlir/utils/performance/perfRunner.py
krzysz00: ... yeah, you're right. But you'll still want to return a full table row

mlir/utils/performance/perfRunner.py
```diff
@@ -793,6 +791,161 @@ def benchmarkMIOpenWithMLIRKernels(configs, arch, filename, paths: Paths):
     chip = GFX_CHIP_RE.search(arch).group(0)
     df.to_csv(chip + '_' + filename, index=False)
 
+RUNNABLE_TEST_RE = re.compile(r"//\s*RUN\s*:")
+ROCMLIRGEN_RE = re.compile(r"rocmlir-gen.*-fut")
```
krzysz00: You could probably use a match group (something like `rocmlir-gen.*-fut\s*(\w+)`) to pull the name out during the regex match

mlir/utils/performance/perfRunner.py
```diff
@@ -793,6 +791,161 @@ def benchmarkMIOpenWithMLIRKernels(configs, arch, filename, paths: Paths):
     chip = GFX_CHIP_RE.search(arch).group(0)
     df.to_csv(chip + '_' + filename, index=False)
 
+RUNNABLE_TEST_RE = re.compile(r"//\s*RUN\s*:")
```
krzysz00: Similarly, here could go with `RUN\s*:\s*([^|]+)\s*|`

mlir/utils/performance/perfRunner.py
krzysz00: .. which you do, I can't read.

krzysz00 (commented):
I've got minor suggestions re using regex more to save on random string manipulations, but overall this seems fine to me. Ping me for an approve if you want one

krzysz00 (approved):

PR #1058, opened 2023-04-26
Modify ops/passes/parameters to use accel instead of xdlops

giuseros:
This is a PR required for https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1055 . I am changing mostly:
- `XdlopsGemmParams` -> `AccelGemmParams`
- `Rock_XdlopsGemmV2Op` -> `Rock_AccelGemmV2Op`

I am not touching tuning yet, so for now `InitParamsXDL` and `PopulateParamsXDL` will stay that way. Please, feel free to point to places that I forgot and/or to places where the rename was not correct.


krzysz00:
I'm not sure I agree with this. I think it might make more sense to have a separate `xdlops_gemm_params` and a distinct `wmma_gemm_params`

giuseros:
But then don't you also have to have two separate operations? I mean in:
```
def Rock_AccelGemmV2Op:
    Rock_Op<"accel_gemm_v2">,
    Arguments<(ins Index:$mRepeat,
                   Index:$nRepeat,
                   MemRefRankOf<AccelArgTypes, [1]>:$matrixA,
                   MemRefRankOf<AccelArgTypes, [1]>:$matrixB,
                   MemRefRankOf<AccelResTypes , [1]>:$matrixC,
                   StrAttr:$arch,
                   Rock_AccelGemmParamsAttr:$params)> 
```
We need to have one with `Rock_XdlopsGemmParamsAttr` and one with `Rock_WmmaGemmParamsAttr`. 

Or are you suggesting to have the two types  inherit from a common base class "accel_param"? Consider also that the parameters for wmma and xdlops are the same.

krzysz00:
I'm suggesting that there's a future where they might not be the same. For example, wmma might tune wavesize, where mfma can't.

And yeah, inheritance is probably a good call here - we already have a tuning parameter interface, after all.

krzysz00:
Or, more interestingly, instead of inheritance, composition: we have accel_params, and then xdlops_params and wmma_params that include an accel_params which carries the general stuff.

giuseros:
Re future proofness, fair enough, let's try to make the design future proof :) 

Re composition. how would the signature of  `Rock_AccelGemmV2Op` (and gridwise/blockwise/threadwise)  be?

giuseros:
So one possible idea, the only one that comes to my mind is to have:
```
class Rock_AccelGemmParamsAttr 
def  Rock_XdlopsGemmParamsAttr : Rock_AccelGemmParamsAttr
def Rock_WmmaGemmParamsAttr : Rock_AccelGemmParamsAttr
```
The operations/passes will accept `Rock_AccelGemmParamsAttr` and the passes will pass the `Rock_AccelGemmParamsAttr` to the `AccelEmitter` whose specialisations (`wmmaEmitter` and `mfmaEmitter`) will know to `dyn_cast` to the right type.

krzysz00:
Right, I forgot TableGen can do that. I'm not 100% sure it'll show up as a `dyn_cast`able type.

Another thought I have is either
- AccelTuringParamsInterface or;
- (since it just popped into my head, I think it might work): Adding an `std::optional<AccelTuningParamsAttr> getCommonAccelParams()` to the tuning parameter interface.

giuseros:
I used the interface approach, with the same `getKPerBlock`, `getMPerBlock`, etc... defined in the interface. 

In this way if the specialised paramsAttr has any of the common parameters it doesn't have to do anything (because table gen already defines it from the record in the `td` file). 

In other words `Rock_XdlopsGemmParamsAttr` already adheres to the interface (no additional methods need to be defined).

However, if we want to remove `kpack` from `Rock_WmmaGemmParamsAttr` we will have to implement only `getKpack()` to be compliant to the `RockAccelTuningParamAttrInterface`.

sjw36:
Another naming suggestion, GemmCore or MMACore (instead of AccelGemmV2..). Good with whatever yall want to go with.

jerryyin:
I like it that the `AccelGemmV2` has `Gemm` in it so it is specific enough to tell what it does. (Core is a bit too generic in that sense). But I'd like to drop the `V2` in it. I think that used to differentiate the V1 version of lots of implementations co-exist in our repo (like xdlopsgemm and xdlopsgemmV2). Now that doesn't apply anymore in majority of cases.

krzysz00:
Yeah, I'm with Giuseppe on Core being a bit generic, and I think `accel_gemm` is clear enough - and it pairs nicely with the alternative op at that level being `threadwise_gemm`

mlir/include/mlir/Dialect/Rock/IR/RockAccelTuningParamAttrInterface.td
```diff
@@ -0,0 +1,110 @@
...
+        /*methodBody=*/"",
+        /*defaultImplementation=*/""
+      >,
+    InterfaceMethod<
+        /*desc=*/[{
+          Given that the GEMM input blocks layout is `[kOuter, dPerBlock, kPack]` (where `d` can be `m` or `n`)
+          this method returns `kOuter`.
+        }],
+        /*retType=*/"int64_t",
+        /*methodName=*/"getKPerBlock",
```
krzysz00: Since we're already renaming stuff, can we do ourselves a favor and call this `getKpacksPerBlock` or something? We might maybe clean up the main tuning parameters interface as well while we're here.

mlir/include/mlir/Dialect/Rock/IR/RockAccelTuningParamAttrInterface.td
```diff
@@ -0,0 +1,110 @@
...
+          Given that the GEMM input blocks layout is `[kOuter, dPerBlock, kPack]` (where `d` can be `m` or `n`)
+          this method returns `kPack`.
+        }],
+        /*retType=*/"int64_t",
+        /*methodName=*/"getKpack",
+        /*args=*/(ins),
+        /*methodBody=*/"",
+        /*defaultImplementation=*/""
+      >,
+    InterfaceMethod<
```
krzysz00: Could we move this onto the main tuning parameters interface, and have it as statically `true` on the general gemm parameters?

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -363,10 +363,10 @@ def Rock_GridwiseGemmV2Op :
                    StoreMethodAttr:$storeMethod,
                    I32Attr:$blockSize,
                    I32Attr:$gridSize,
-                   Rock_XdlopsGemmParamsAttr:$params)> {
+                   RockAccelTuningParamAttrInterface:$params)> {
   let summary = "Gridwise GEMM V2";
```
krzysz00: Heck, if we're doing the big renaming kick, maybe this could be named `gridwise_gemm_accel` and we'd have `blockwise_gemm_accel` as well?

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1119,20 +1119,20 @@ def Rock_ThreadwiseGemmOp:
   let hasVerifier = 1;
 }
 
-// xdlops_gemm_V2
-def Rock_XdlopsGemmV2Op:
-    Rock_Op<"xdlops_gemm_v2">,
+// accel_gemm_V2
+def Rock_AccelGemmV2Op:
```
krzysz00: Yeah, this doesn't need a v2

krzysz00 (commented):
One organizational thought, a few naming thoughts

krzysz00:
Actually, come to think of it, `getKpack()` could also be a method on the general tuning parameter interface.

giuseros:
From what I see, also `getMPerBlock`, `getNPerBlock` and `getKpackPerBlock` is used in general and accel params. SHould I move them too? Just to be sure, the accel ops, still need the accelparaminterface. This means that I will have to dyn_cast to get to the general parameters (right?)

giuseros:
Ok, I found out that we can have interface inheritance in td files :)

 I applied most of your changes, except I didn't rename `kPerBlock` -> `kpackPerBlock` in the general parameter. Point is that I don't know if there are differences. To sum up I have:

General parameters: `{blockSize, kPerBlock, mPerBlock, nPerBlock, kPerThread, mPerThread, nPerThread, kpack}`
Accel parameters: `{mPerBlock, nPerBlock, kpackPerBlock, mPerWave, nPerWave, kpack, forceUnroll}`
Generic interface: `{kpack, forceUnroll}`
Accel interface: `{mPerBlock, nPerBlock, kpackPerBlock, mPerWave, nPerWave}`

What I don't know is if identical name means identical concept. If so, we can move many things in the generic interface.

krzysz00:
So the current situation with general GEMM is that they have kpack = 1 always, so kPerBlock == kpacksPerBlock

(The name kPerBlock came from before kpack was a thing, was what happened, and the design decision was made to repurpose it as kpacksPerBlock)

krzysz00:
(Yeah, I'd've thought you'd need to dyn_cast, but I guess not)

krzysz00 (approved):
I'm happy with these changes!

PR #1059, opened 2023-05-01
Do an ugly workaround issues with multiple LDS variables.

krzysz00:
Looks like, in some cases, the backend gets something wrong with multiple LDS buffers allocated. So, use one buffer, and the offset argument to memref.view to generate appropriate types for A and B buffer views as needed.

Hopefully we can resolve the underlying issue and this won't be neeeded anymore.


giuseros (approved):
LGTM!

PR #1060, opened 2023-05-02
Create the devel package for header files and the static library

yiqian1:
- Remove librockCompiler and librockCompiler-headers components
- Include header files and static libraries into the default devel component
- Install static library at lib directory instead of its subdirectory
- Rename our installation directory as "rocmlir"
- Rename the cmake filenames: rocmlir-targets.cmake, rocmlir-targets-release.cmake, rocmlir-config.cmake and rocmlir-config-version.cmake. These are default names defined in rocm-cmake

MIGraphX tests passed
http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-1060/4/pipeline


CMakeLists.txt
```diff
@@ -35,11 +35,6 @@ set(ROCMLIR_BIN_DIR "${CMAKE_CURRENT_BINARY_DIR}/bin" CACHE PATH "")
 set(ROCMLIR_LIB_DIR "${CMAKE_CURRENT_BINARY_DIR}/lib" CACHE PATH "")
 message(STATUS "ROCMLIR_BIN_DIR: ${ROCMLIR_BIN_DIR}")
 
-if(BUILD_MIXR_TARGET)
```
krzysz00, line 38: We might still want to support BUILD_MIXR_TARGET as a flag for backwards-compatibility reasons

CMakeLists.txt
```diff
@@ -35,6 +35,11 @@ set(ROCMLIR_BIN_DIR "${CMAKE_CURRENT_BINARY_DIR}/bin" CACHE PATH "")
 set(ROCMLIR_LIB_DIR "${CMAKE_CURRENT_BINARY_DIR}/lib" CACHE PATH "")
 message(STATUS "ROCMLIR_BIN_DIR: ${ROCMLIR_BIN_DIR}")
 
+# BUILD_MIXR_TARGET is deprecated but still available
+if(BUILD_MIXR_TARGET)
+  set(BUILD_FAT_LIBROCKCOMPILER ON CACHE BOOL "Build static librockCompiler")
+-endif()
```
krzysz00: Extra - here, probably causing your build failure.

Also, can we verify that this packages all the header files MIGraphX needs now?

yiqian1: I can confirm that it packages everything that old packages included. But we might have packaged more files than what MIGraphX actually needs, but I'm not sure about that.

krzysz00:
I'm willing to approve pending that QA et al are happy with this

yiqian1:
@lawruble13 Could you please review this PR?

lawruble13 (approved):
LGTM, apologies for the delays in approving this.

PR #1061, opened 2023-05-02
Fix ck-benchmark-driver for int8 data-type.

giuseros:
Fixes https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/855

This is a dirty hack to solve the int8 perf degradation (<50% perfs) when comparing CK vs MLIR. 

### The problem
What was happening is that MLIR quantized gemms are `int8->int32`, while in CK they are `int8->int8`. This meant that when G was big and K was small, the majority of time was spent writing data into global memory. We were writing `int32`, they were writing `int8` hence they were significantly faster. 

### The dirty solution
Unfortunately CK does not support `int8->int32` gemms by default. The dirty solution I came up with was to add the `int8->int32` instances manually, and reroute the `ck-benchmark-driver` to those instances (instead of the canonical `int8->int8` instances) when `int8` datatype is passed. The only drawback I see with this, is that the compile time incrased significantly for the `ck-benchmark-driver` because now we are building part of CK (which is very slow to build)

### The proper solution
I think the proper solution should be to support `int8->int8` in MLIR. If I am not mistaken @jerryyin is working on this. So @jerryyin as soon as you enable the quantization in MLIR this hack should be removed

### Performance comparison
You can see the new performance comparison attached to this PR. The last two reds are gone, and also some yellows turned into blues. 

[gfx90a_MLIR_vs_CK.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/11371765/gfx90a_MLIR_vs_CK.zip)

### DIFF
--- a/mlir/utils/jenkins/static-checks/clang-tidy.ignore
+++ b/mlir/utils/jenkins/static-checks/clang-tidy.ignore
@@ -1,5 +1,6 @@
 # Reference: https://github.com/google/llvm-premerge-checks/blob/master/scripts/clang-tidy.ignore
 # Files to be ignored by clang-tidy. Will not appear in short report and inline comments.
 mlir/examples/standalone
+mlir/utils/performance/ck-benchmark-driver
 mlir/test
 
--- a/mlir/utils/performance/ck-benchmark-driver/CMakeLists.txt
+++ b/mlir/utils/performance/ck-benchmark-driver/CMakeLists.txt
@@ -5,10 +5,11 @@ endif()
 
 if (composable_kernel_FOUND)
   find_package(hip REQUIRED PATHS /opt/rocm)
+  FILE(GLOB Int8Instances int8_int8_int32_instances/*.cpp)
   set(LIBS composable_kernel::device_operations hip::device benchmark-driver-utils)
   set(EXCLUDE_FROM_ALL TRUE)
 
-  add_executable(ck-benchmark-driver ck-benchmark-driver.cpp)
+  add_executable(ck-benchmark-driver ck-benchmark-driver.cpp ${Int8Instances})
   target_link_libraries(ck-benchmark-driver PRIVATE ${LIBS})
   set_target_properties(ck-benchmark-driver
     PROPERTIES
--- a/mlir/utils/performance/ck-benchmark-driver/ck-benchmark-driver.cpp
+++ b/mlir/utils/performance/ck-benchmark-driver/ck-benchmark-driver.cpp
@@ -27,6 +27,9 @@
 #include "ck/tensor_operation/gpu/device/tensor_layout.hpp"
 #include "ck/tensor_operation/gpu/element/element_wise_operation.hpp"
 
+#include "int8_int8_int32_instances/batched_gemm.h"
+#include "int8_int8_int32_instances/gemm.h"
+
 // System includes
 #include <iostream>
 #include <string>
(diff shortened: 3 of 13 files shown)


jerryyin:
> What was happening is that MLIR quantized gemms are int8->int32, while in CK they are int8->int8.

That sounds to be a very reasonable conclusion for performance comparison result. Congrats on finding this out!

> I think the proper solution should be to support int8->int8 in MLIR.

I'm curious about this. With the output int8 in such limited range, how do they guarantee the precision is under control? Unless they do fusion to treat the output int8 as another quantized output. Then they will need to support `conv` + `dequantize` + `activation` + `quantize` fusion as well as a number of quantization arguments. That's going to make it look like a sub-graph, not a convolution. If that's what you are talking about, yes we already support (limited combination of) this, but in the scope of migraphx integration.

> Unfortunately CK does not support int8->int32 gemms by default.

I'm surprised about this. If we go with above reasoning, I think the int8 -> int8 result is the byproduct of fusion therefore they really should have a way to do the convolution in non-fusion way. Ultimately, the xdlops output is only int32 and there should be an option for them to do "nothing" after xdlops compute completes and just write back.

> Unfortunately CK does not support int8->int32 gemms by default. The dirty solution I came up with was to add the int8->int32 instances manually

Continue the reasoning here, I think the more correct approach is to not "add" additional computes to push int8 back to int32, because that's going to be unfair to CK. Rather, the approach should be to "drop" those additional fusion related ops from their convolution to make the int32 native computed result to be written out directly.

giuseros:
Hi @jerryyin , if I understood your comment correctly, I expressed myself in the wrong way :) 

CK produces the gemm/conv output in int32. They have a check (inside the gemm operation) that then takes care of converting int32 to int8. So they do something like 
```
if (outType != accType) { convertToOutType(accType)}
```
The problem is that [in their configurations](https://github.com/ROCmSoftwarePlatform/composable_kernel/tree/develop/library/src/tensor_operation_instance/gpu/batched_gemm) they only add `int8*int8=int8`, always enabling the conversion. What I am doing is to add a set of `int8*int8=int32` where they don't need to convert. 

Basically, I think I am implementing exactly what you are proposing. I am not converting int8 to int32, but I am stopping the conversion from the int32 accumulator to the int8 result (which I believe is what you are referring to).

jerryyin:
Thanks for your explanation, that make much more sense now! So I would not call this PR a hack but a reasonable patch to CK.

I'm just giving a reminder that at the end of our gridwisegemm, we also have this

> if (outType != accType) { convertToOutType(accType)}

I think this was only used for fp16 now but you can easily extend to int8 if you'd like. We can potentially also support this new flavor of int8 just for CK comparison if we want to do it. You approach is also a fair way to do it too. Now my question is, how stable is their configuration files? As long as they don't periodically break us then we can stick to what you have.

giuseros:
Extending MLIR to support `int8*int8=int8` seems something that require more thought, because we need to understand what conversion schema they are using and replicate into MLIR. 

The problem is also that  different frameworks might use different schemas, so hardcoding what CK has might also be not ideal. 
 I thought that your work on quantization was exactly to come up with a common quantization schema. 

As per extending CK, I don't think they can break us, unless they make a change for which one of those configurations is not valid anymore (which for now never happened). But if they add new (faster) configurations for `int8*int8=int8` case we would need to add them in our extension to keep the comparison fair.

jerryyin:
> Extending MLIR to support int8*int8=int8 seems something that require more thought

I was not saying that we make MLIR to support int8*int8 = int8, but rather to make it available so as to do only perf comparison with CK. I don't think this is a reasonable configuration to expose to the external world. I don't think it is reasonable for CK to expose such either. We should find out why they'd expose such a low precision version of int8 conv. Could you ask CK how they address this concern?

> I thought that your work on quantization was exactly to come up with a common quantization schema.

But that's going to make it to be quantization, not convolution. It'd not be fair comparison if we compare the quantization with their native convolution. I'd much more prefer a simple single op to go into our performance report.

> which one of those configurations is not valid anymore (which for now never happened).

Okay that sounds good enough. The only concern here is about whether we can catch up with their updates. Like if they updated their configurations, how do we know it, and update ours accordingly?

giuseros:
> Okay that sounds good enough. The only concern here is about whether we can catch up with their updates. Like if they updated their configurations, how do we know it, and update ours accordingly?

Well the easiest way would be to create a PR on their repo to add the configurations I added. Once this happen we can remove this hack

krzysz00 (commented):
I'd be good to approve this for now

jerryyin:
I will not block this PR from merging but I strongly suggest not to based on lack of understanding from the key questions below:
 - Why do they want to support `int8xint8->int8` configurations knowing that precision will suck?
 - How do we keep up to date with CK on those duplications that should belong to CK's repository?
 - Is it even a fair comparison to twist their configurations this way? (We don't want to introduce a new bias to fix an old bias)

giuseros:
Closing in favour of a different approach sketched in this ticket: https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/855

PR #1063, opened 2023-05-02
[EXTERNAL] add tosa erf and its lowering

manupak:
This commit adds tosa erf operator and its lowering
to math lib functions.

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOps.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOps.td
@@ -390,6 +390,35 @@ def Tosa_TanhOp : Tosa_Op<"tanh", [
   );
 }
 
+
+//===----------------------------------------------------------------------===//
+// Operator: erf
+//===----------------------------------------------------------------------===//
+def Tosa_ErfOp : Tosa_Op<"erf", [
+    DeclareOpInterfaceMethods<InferShapedTypeOpInterface,
+                              ["inferReturnTypeComponents"]>,
+    Pure]> {
+  let summary = "Computes gauss error function of input";
+
+  let description = [{
+    Gauss error function:
+                           - x
+    erf(x) = 2 / sqrt(pi) /  e^(-t^2) dt
+                         - 0
+    For quantized integer data types, the TABLE operator should be used instead
+    with the following definition.  The erf_table has 513 entries each of
+    16-bit/8-bit precision and covering the input range -4.0 to +4.0 in steps of 1/64.
+  }];
+
+  let arguments = (ins
+    Tosa_Tensor:$input
+  );
+
+  let results = (outs
+    Tosa_Tensor:$output
+  );
+}
+
 //===----------------------------------------------------------------------===//
 // TOSA Spec Section 2.4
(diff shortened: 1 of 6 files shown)


manupak:
depends on https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1062

manupak:
already started tosa discussion : https://discuss.mlplatform.org/t/adding-erf-operator-to-tosa/189/3.
lets see.

external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOps.td
```diff
@@ -390,6 +390,32 @@ def Tosa_TanhOp : Tosa_Op<"tanh", [
...
+// Operator: erf
+//===----------------------------------------------------------------------===//
+def Tosa_ErfOp : Tosa_Op<"erf", [
+    DeclareOpInterfaceMethods<InferShapedTypeOpInterface,
+                              ["inferReturnTypeComponents"]>,
+    Pure]> {
+  let summary = "Computes gauss error function of input";
+
+  let description = [{
+    Gauss error function.
```
krzysz00: Do we perhaps want to stick the definition of the gauss error function here to save people from having to run to a separate document?

external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOps.td
```diff
@@ -390,6 +390,32 @@ def Tosa_TanhOp : Tosa_Op<"tanh", [
...
+    For quantized integer data types, the TABLE operator should be used instead
```
krzysz00, line 408: Does `TABLE` already exist? And should it maybe be spelled `tosa.table`?

krzysz00 (commented):
Add a test for the lowering, please, but looks good otherwise

external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOps.td
manupak: sure, I ll add that.

external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOps.td
manupak: yes it does. Im being consistent with other descriptions.

manupak:
Its is kind of tested in the next PR (migraphx one) that suits our needs.
I thought to keep this as minimal as possible for now and the tests in upstream in the way they'd want it.

WDYT ? -- Im not against adding it here though, just wanted to know your thoughts.

krzysz00:
I figured upstream will want the tests anyway, so might as well add them here. They probably have tests for all the other math functions, so you'll probably want to find those and add `erf` as a case.

manupak:
OK I've added tests.

Also started upstreaming...
https://reviews.llvm.org/D150357
https://discuss.mlplatform.org/t/adding-erf-operator-to-tosa/189/9

krzysz00 (approved):

manupak:
Thanks @krzysz00!

PR #1064, opened 2023-05-02
Add MIGraphX erf op and its lowering

manupak:
This commit adds migraphx.erf op and its
lowering to tosa.erf with tests.

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/867

### DIFF
--- a/mlir/include/mlir/Dialect/MIGraphX/MIGraphXOps.td
+++ b/mlir/include/mlir/Dialect/MIGraphX/MIGraphXOps.td
@@ -132,6 +132,17 @@ def MIGraphX_NegOp :
   let summary = "Negation";
   let description = [{
     get neg
+      }];
+  let assemblyFormat = "`(`$inA`)` attr-dict `:` `(`type($inA)`)` `->` type($output)";
+}
+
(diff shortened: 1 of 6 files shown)


krzysz00 (commented):
The MIGraphX part of this looks perfectly fine to me.

manupak:
Rebased now. -- now only has MIGraphX things here.

krzysz00 (approved):
Approved, but do we perhaps want to restrict this to f32 tensors because we don't have lower-precision erf?

krzysz00:
(Or, more generally, do we want to fix up gpu-to-rocdl to handle input types not supported by the functions by way of some extf's)

manupak:
We do have it tested here for f16 too, but do u mean like fp8?

manupak:
> Approved, but do we perhaps want to restrict this to f32 tensors because we don't have lower-precision erf?

We have it for f16 too.

krzysz00:
Ah. In which case, assuming we have a lowering pattern for f16, never mind all that

manupak:
No worries

manupak:
Thanks!

PR #1065, opened 2023-05-02
Add migraphx bert configs to perf tracking

manupak:
This commit adds the gemm configs that we
see from MIGraphX when we run bert models.

### DIFF
--- a/mlir/utils/performance/gemm-configs
+++ b/mlir/utils/performance/gemm-configs
@@ -1,4 +1,4 @@
-# This section of the file has been generated with the following command:
+# The immediately following configs are generated with the following command:
 # convertRocBlasToPerfRunner.py -c bert-configs-raw --output-file bert-configs
 
 -transA true -transB true -g 64 -m 1024 -n 384 -k 1024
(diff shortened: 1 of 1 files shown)


giuseros:
Hi @manupak , I was wondering, to make perf comparisons easier, how about adding a `--set` options to `perfRunner.py` ? `--set=0` would be the original dataset, `--set=1` the one you added and `--set=all` all together (which would be the default in the nightly)? Another option would be to add a `--config-set=[0..X]` . What do you think?

manupak:
Hmm....

I dont personally like adding another level of hierarchy to configs. I think we can get the same effect by adding this as a seperate file if thats what we want -- but that undo s the conversation here : https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1056. cc : @krzysz00 .

Also why is the original set has to be seperately measured ? as in what is the significance of that.
If we want a seperation, maybe we can add these as batch size 1 configs.

giuseros:
Well it was more for ease of use when debugging performance. But yeah I can simply comment out the lines in the config that I don't want to evaluate.

giuseros (approved):
LGTM!

manupak:
@giuseros can you take another look because I added all the gemms from all bert models ?

PR #1067, opened 2023-05-04
Improve tuner benchmark function

giuseros:
This is a changed we discussed in the previous release. I am hoisting out from the tuning loop:
- The allocation of host memory
- The initialisation of host memory
- The allocation of device memory

The tuning loop calls an async copy from Host to Memory to initialise the device buffers.

### DIFF
--- a/mlir/tools/rocmlir-tuning-driver/rocmlir-tuning-driver.cpp
+++ b/mlir/tools/rocmlir-tuning-driver/rocmlir-tuning-driver.cpp
@@ -109,11 +109,13 @@ static benchmark::DataType getDataType(Type inputType) {
   }
 
 // In order to match rocprof, returns time in nanoseconds
-static FailureOr<double>
-benchmarkKernel(const char *binary, const char *funcName, uint32_t blockSize,
-                uint32_t gridSize, benchmark::DataType dataType,
-                std::vector<void *> hostBuffers, std::vector<void *> gpuBuffers,
-                std::vector<size_t> bufferSizes) {
+static FailureOr<double> benchmarkKernel(const char *binary,
+                                         const char *funcName,
+                                         uint32_t blockSize, uint32_t gridSize,
+                                         benchmark::DataType dataType,
+                                         ArrayRef<void *> hostBuffers,
+                                         MutableArrayRef<void *> gpuBuffers,
+                                         ArrayRef<size_t> bufferSizes) {
   constexpr double msToNs = 1e6;
   hipModule_t mod;
   HIPCHECK(hipModuleLoadData(&mod, binary))
--- a/mlir/tools/rocmlir-tuning-driver/rocmlir-tuning-driver.cpp
+++ b/mlir/tools/rocmlir-tuning-driver/rocmlir-tuning-driver.cpp
@@ -102,18 +102,19 @@ static benchmark::DataType getDataType(Type inputType) {
   }
 }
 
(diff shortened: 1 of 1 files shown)


giuseros:
@manupak This is the change to the tuner I was mentioning yesterday

mlir/tools/rocmlir-tuning-driver/rocmlir-tuning-driver.cpp
```diff
@@ -102,18 +102,19 @@ static benchmark::DataType getDataType(Type inputType) {
...
 #define HIPCHECK(expr)                                                         \
   if (hipSuccess != (expr)) {                                                  \
     return failure();                                                          \
   }
+
+// In order to match rocprof, returns time in nanoseconds
+static FailureOr<double>
+benchmarkKernel(const char *binary, const char *funcName, uint32_t blockSize,
+                uint32_t gridSize, benchmark::DataType dataType,
+                std::vector<void *> hostBuffers, std::vector<void *> gpuBuffers,
```
manupak:
```suggestion
                ArrayRef<void *> hostBuffers, ArrayRef<void *> gpuBuffers,
```

manupak (commented):
Thanks @giuseros! just a single nit.

mlir/tools/rocmlir-tuning-driver/rocmlir-tuning-driver.cpp
giuseros: Yeah, I had it originally like this, but I "think" that `ArrayRef` is a const reference, and I have:
```
// HIP wants an array of pointers to each argument
  std::vector<void *> argPointers;
  for (void *&item : gpuBuffers) {
    argPointers.push_back(reinterpret_cast<void *>(&item));
  }
``` 
This was making the compiler unhappy (because they should be `void const*` but I am casting to `void *`). `HostBuffer` can be an `ArrayRef`, but I thought it nicer to have the same type for both (at the end it's a vector with three pointers and we can afford copying it around). What do you think?

krzysz00: Other types worth knowing about include `SmallVectorImpl<T> &` for passing SmallVector by reference and also `MutableArrayRef`.

krzysz00: Or (since someone in a different review reminded me of it), we could go with `std::array<void*, 3>&`

krzysz00 (approved):
Seems quite reasonable to me, and type nitpicking isn't blocking, approved.

mlir/tools/rocmlir-tuning-driver/rocmlir-tuning-driver.cpp
giuseros: I opted for `MutableArrayRef`. So that I have `ArrayRef`s for host buffers and sizes, and `MutableArrayRefs` for buffers than need to be allocated.

manupak: I think the problem is : (void *&item : gpuBuffers). Here you iterate by reference rather than value -- so it can be problematic when the container is a const reference. I think if you iterate by value it shouldn't be an issue but then the saving is copying of the container vs copying of each item -- thus maybe not worth reworking in this case.

manupak (approved):
LGTM

PR #1068, opened 2023-05-04
Improve regression reports

giuseros:
I added a small usability improvement to the regression reports. My use case is that I tend to generate a lot of regression reports, using different `csv` , but then I tend to don't remember which `csv` file I used. 

This changes adds the old/new files to the regression reports in its own subsection

### DIFF
--- a/mlir/utils/performance/perfRegressionReport.py
+++ b/mlir/utils/performance/perfRegressionReport.py
@@ -5,6 +5,7 @@ import reportUtils
 from pathlib import PurePath
 import sys
 from typing import Tuple
+import os
 
 import pandas as pd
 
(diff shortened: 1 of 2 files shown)


mlir/utils/performance/perfRegressionReport.py
```diff
@@ -120,6 +121,8 @@ def getPerfDate(statsPath: PurePath, default="???"):
         oldLabel = "copy"
 
     data, summary = computePerfStats(oldDf, newDf, oldLabel, newLabel)
+    oldNewPaths = {"old" : os.path.abspath(oldDataPath),
```
krzysz00, line 124: On the one hand, this is why we have `perf-run-date`, so you can mark which file is which (so long as they're in separate directories).
This also seems like it'll add useless noise to the CI-generated regression reports.

I'm not opposed, but I figured I'd point this out.

giuseros: Yeah, I didn't think at the noise in the CI. Let's do this, I will close this for now and see if I can create multiple folders where I store my local data (so that csv, tsv and reports are connected, which is what I really need)

PR #1069, opened 2023-05-04
add migraphx neg,exp and div ops.

manupak:
This commit adds neg,exp and div ops for
migrapx dialect and its lowering to tosa.

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/876


jerryyin (approved):
LGTM

PR #1071, opened 2023-05-05
Initial changes to move conversions to f32 into the verify wrapper function

pcf000:
... and let tests run in their native types.  Also update three tests that were checking for f32, and allow kernel-generation cases to use clone-verification.

### DIFF
--- a/mlir/test/rocmlir-driver/populate_prc.mlir
+++ b/mlir/test/rocmlir-driver/populate_prc.mlir
@@ -5,17 +5,18 @@
 // F32-NEXT:    call @printMemrefF32(%[[RES]]) : (memref<*xf32>) -> ()
 
 // RUN: rocmlir-gen --arch %arch -p -prc -t f16 | FileCheck %s --check-prefixes=F16,CHECK
-// RUN: rocmlir-gen --arch %arch -p -prc -t bf16 | FileCheck %s --check-prefix=BF16
+// RUN: rocmlir-gen --arch %arch -p -prc -t bf16 | FileCheck %s --check-prefixes=BF16,CHECK
 
 // F16:    %{{.*}} = memref.alloc() : memref<1x128x8x3x3x[[type:f16]]>
-// BF16:   %{{.*}}= memref.alloc() : memref<1x128x8x3x3x[[type:bf16]]>
-// CHECK:  call @_memcpy_[[type]]_f32_9216(%{{.*}}, %{{.*}}) : (memref<9216x[[type]]>, memref<9216xf32>) -> ()
+// BF16:   %{{.*}} = memref.alloc() : memref<1x128x8x3x3x[[type:bf16]]>
+// CHECK:  memref.copy %{{.*}} : memref<9216x[[type]]> to memref<9216x[[type]]>
 // CHECK:  %{{.*}} = memref.alloc() : memref<128x1x8x32x32x[[type]]>
-// CHECK:  call @_memcpy_[[type]]_f32_1048576(%{{.*}}, %{{.*}}) : (memref<1048576x[[type]]>, memref<1048576xf32>) -> ()
-// CHECK:  %{{.*}} = memref.alloc() : memref<{{.*}}>
+// CHECK:  memref.copy %{{.*}} : memref<1048576x[[type]]> to memref<1048576x[[type]]>
+// CHECK:  %{{.*}} = memref.alloc() : memref<{{.*}}x[[type]]>
+// CHECK:  memref.copy %{{.*}} : memref<{{.*}}x[[type]]> to memref<{{.*}}x[[type]]>
+// CHECK-NEXT: call @conv2d_cpu(%{{.*}}, %{{.*}}, %[[RES1:.*]]) : (memref<1x128x8x3x3x[[type]]>, memref<{{.*}}>, memref<{{.*}}>) -> ()
 // CHECK:  call @_memcpy_[[type]]_f32_{{[0-9]+}}(%{{.*}}, %{{.*}}) : (memref<{{.*}}x[[type]]>, memref<{{.*}}xf32>) -> ()
-// CHECK-NEXT: call @conv2d_cpu(%{{.*}}, %{{.*}}, %[[RES1:.*]]) : (memref<1x128x8x3x3xf32>, memref<{{.*}}>, memref<{{.*}}>) -> ()
-// CHECK:  %[[RES2:.*]] = memref.cast %[[RES1]] : memref<{{.*}}> to memref<*xf32>
+// CHECK:  %[[RES2:.*]] = memref.cast %{{.*}} : memref<{{.*}}> to memref<*xf32>
 // CHECK:  call @printMemrefF32(%[[RES2]]) : (memref<*xf32>) -> ()
 
 // RUN: rocmlir-gen --arch %arch -p -prc -t i8 | FileCheck %s --check-prefix=INT8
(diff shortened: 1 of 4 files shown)


mlir/test/rocmlir-driver/populate_prc.mlir
```diff
@@ -5,17 +5,18 @@
...
 
 // RUN: rocmlir-gen --arch %arch -p -prc -t f16 | FileCheck %s --check-prefixes=F16,CHECK
-// RUN: rocmlir-gen --arch %arch -p -prc -t bf16 | FileCheck %s --check-prefix=BF16
+// RUN: rocmlir-gen --arch %arch -p -prc -t bf16 | FileCheck %s --check-prefixes=BF16,CHECK
 
 // F16:    %{{.*}} = memref.alloc() : memref<1x128x8x3x3x[[type:f16]]>
-// BF16:   %{{.*}}= memref.alloc() : memref<1x128x8x3x3x[[type:bf16]]>
-// CHECK:  call @_memcpy_[[type]]_f32_9216(%{{.*}}, %{{.*}}) : (memref<9216x[[type]]>, memref<9216xf32>) -> ()
+// BF16:   %{{.*}} = memref.alloc() : memref<1x128x8x3x3x[[type:bf16]]>
+// CHECK:  memref.copy %{{.*}} : memref<9216x[[type]]> to memref<9216x[[type]]>
```
pcf000, line 12: Because we now have the test function and the validation function producing the same type, we have memref.copy to directly copy the values rather than memcpy_f16_f32 (for instance) to copy and convert.

mlir/test/rocmlir-driver/populate_pv_with_gpu.mlir
```diff
@@ -31,15 +31,15 @@
 // F16-CHECK: func.func @rock_conv2d_gkcyx_ngchw_ngkhw_0({{.*}}) attributes {kernel = 0 : i32, xmodel.arch = "{{.*}}"} {
 // F16-CHECK: rock.conv2d({{.*}}) features = dot {[[PARMS:.*]]} : memref<[[FILTERDIMS:[x0-9]+]]xf16>, memref<[[INPUTDIMS:[x0-9]+]]xf16>, memref<[[OUTPUTDIMS:[x0-9]+]]xf16>
 // F16-CHECK: call @rock_conv2d_gkcyx_ngchw_ngkhw_0_gpu({{.*}}) : (memref<[[FILTERDIMS]]xf16>, memref<[[INPUTDIMS]]xf16>, memref<[[OUTPUTDIMS]]xf16>) -> ()
-// F16-CHECK: call @rock_conv2d_gkcyx_ngchw_ngkhw_0_ver_gpu({{.*}}) : (memref<[[FILTERDIMS]]xf32>, memref<[[INPUTDIMS]]xf32>, memref<[[OUTPUTDIMS]]xf32>) -> ()
+// F16-CHECK: call @rock_conv2d_gkcyx_ngchw_ngkhw_0_ver_gpu({{.*}}) : (memref<[[FILTERDIMS]]xf16>, memref<[[INPUTDIMS]]xf16>, memref<[[OUTPUTDIMS]]xf16>) -> ()
```
pcf000, line 34: The validation function is in f16, just like the test kernel.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -1700,8 +1700,10 @@ createCPUConvFunc(ModuleOp module,
   OpBuilder b(module.getContext());
   auto loc = b.getUnknownLoc();
 
-  Type elemType = b.getF32Type();
-  Type outputElemType = b.getF32Type();
+  Type elemType = typeFromString(genConfig.dataTypeStr, module.getContext());
```
pcf000, line 1703: Not forcing things to f32, but using the actual types.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -1882,6 +1880,7 @@ static func::FuncOp createCpuGemmKernelWithMlir(ModuleOp module,
   return func;
 }
 
+// Make an element-by-element copy that converts from srcType to destType.
```
pcf000, line 1883: "memcpy" is kind of a misnomer.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2006,9 +2005,13 @@ static void emitMemcpy(OpBuilder &b, Value src, Value dst) {
   Value dstFlat = makeNDMemRef(b, dst, 1);
   auto srcFlatType = srcFlat.getType().cast<MemRefType>();
   auto dstFlatType = dstFlat.getType().cast<MemRefType>();
-  auto memcpyFunc = getMemcpyFuncDecl(module, srcFlatType, dstFlatType);
 
-  b.create<func::CallOp>(loc, memcpyFunc, ValueRange{srcFlat, dstFlat});
+  if (srcFlatType == dstFlatType) {
```
pcf000, line 2009: The old code assumed/knew that there would be a type conversion here, and called getMemcpyFuncDecl() in all cases.  The new version accounts for cases without a conversion.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2133,87 +2142,65 @@ static func::FuncOp createVerifierFunc(ModuleOp module, const KernelIF &kernel,
...
-  } else if (valElemType.isInteger(64)) {
-    verifyFuncName += "Int32Int64";
-  } else if (valElemType.isInteger(8)) {
-    verifyFuncName += "Int8";
-  } else {
-    llvm::errs() << "Unsupported type of validation function output: ";
-    llvm::errs() << " (Only f32, int32 and int64 are supported)\n";
-    exit(1);
-  }
+  Value testResult, valResult;       // Values passed to the verify function
```
pcf000, line 2145: Choose the function name further down.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2133,87 +2142,65 @@ static func::FuncOp createVerifierFunc(ModuleOp module, const KernelIF &kernel,
...
+  auto mr1DUnkF32Type =
+      MemRefType::get({mlir::ShapedType::kDynamic}, floatType);
 
-  bool isTestAndValSameType =
-      (testOutType.isInteger(32) || testOutType.isF32());
-
-  Value testResult, valResult; // Values passed to the verify function
-  Value testResultNew;         // Values used for type conversion
-  if (!isTestAndValSameType) {
+  if (testOutType.isa<FloatType>() && !testOutType.isF32()) {
```
pcf000, line 2155: The old side, using the isTestAndValSameType variable (not quite a descriptive name), would convert the f16 kernel results to f32, and take the validation results (which were always f32) and convert to f16 and back to mimic the same precision.

The new side converts both kernel and validation results identically, since they're going to be the same type -- if float and not f32, convert to f32.

Both sides then pass the results to the mcpuVerify function.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2231,10 +2218,14 @@ static func::FuncOp createVerifierFunc(ModuleOp module, const KernelIF &kernel,
                            ValueRange{testResult, valResult, printDebugVal});
   }
 
-  if (!isTestAndValSameType) {
+  if (testResultNew && !testResultNew.use_empty()) {
```
pcf000, line 2221: If we allocated a new buffer, and it's still in use, deallocate it.  Same effect as checking the original flag.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2363,7 +2349,6 @@ static void insertValidationCalls(const GenParams &genParams, OpBuilder &b,
           exit(1);
         }
         KernelIF kernel(conv2dGenerator.getKernelFunc());
-        rock::Conv2dGenerator::Config newConfig = conv2dGenerator.getConfig();
```
pcf000, line 2366: newConfig is never used.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2775,12 +2750,6 @@ int main(int argc, char **argv) {
     });
 
   } else {
```
pcf000, line 2752: Remove this check, to allow clone-verification for generated-kernel cases too.  The old code didn't try to make the types work out, but now we make both kernel and validation function have the same types.

krzysz00 (approved):
Probably not a bad idea, and I don't see any issues with this, so approved

giuseros (approved):
Yes, looks fine to me too!

PR #1073, opened 2023-05-08
[DO NOT SQUASH] Brute-force compiler patch to fix all the miscompiles we've tripped over

krzysz00:
1. Revert the old workaround, since it only shifted around what caused the breakage and is something we'd rather not have running around if we don't need it.
2. Disable (by making it do nothing) a rewrite in the compiler that seems to be in the chain of code leading to the bug - that is, I don't think the function I patched is the buggy one, but that if it runs, it triggers the bug somewhere down the line. This is definitely an ugly workaround, but shouldn't have much of an impact perf-wise.

I've tested all the configs that've failed so far and all of them pass on this branch.

### DIFF
--- a/external/llvm-project/llvm/lib/Target/AMDGPU/SIISelLowering.cpp
+++ b/external/llvm-project/llvm/lib/Target/AMDGPU/SIISelLowering.cpp
@@ -9468,6 +9468,12 @@ SDValue SITargetLowering::performSHLPtrCombine(SDNode *N,
   if (!CAdd)
     return SDValue();
 
+  // EMERGENCY WORKAROUND! When this rewrite triggers, it causes some other
+  // currently unknown pattern causes a miscompile for certain cases.
+  // Therefore, disable this unconditionally. We'll want to remove this as
+  // soon as these's an actual fix to whatever the underlying problem is.
+  return SDValue();
+
   // If the resulting offset is too large, we can't fold it into the addressing
   // mode offset.
   APInt Offset = CAdd->getAPIntValue() << CN1->getAPIntValue();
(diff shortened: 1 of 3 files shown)


jerryyin (commented):
I'm impressed that you can find out such a work around in AMDGPU lowering. However, I don't understand what you have done and whether that will impact something else in kernel compilation. I'd prefer you pull up at least one LLVM developer (or someone in our team that understand more of this) to guarantee us that it doesn't have any other side effect.

----

Taking one step back. Could you evaluate how wide will the original bug affect? I think we roughly have 3 options:
1. Revert your original PR that triggers it, discards your attempt until backend is fixed. Then we reapply your original PR.
2. Disable the (subset of the 4) test if it only hit on very limited configs, reenable them when backend is fixed.
3. Go with your current PR (If we are absolutely sure this will not have any side effects)

krzysz00:
The problem with 1. is that it breaks fp8 support

jerryyin:
@krzysz00 Fair point. I forget about the motivation of your PR is to enable fp8 support. That's making it a fairly high priority item and we should just revert and reapply.

For 2, I started a tuning with the existing code base [here](http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir-weekly/detail/mlir-weekly/160/pipeline/106/). If the tuning is passing, then I'd be good to just disable that small subset of parameter sweep test.
 - The follow up item for this is a backup plan: how can the rest of the team (who are actively working on a much larger range of config know that when kernel run fails, it isn't a "real failure"? but from the backend bug. Any more apparent attributes other than what we already know about the failing configs are mostly backward weights?
 - It seems to me that answer to that question is to re-run and apply -O0 to the rocm runner?

For 3, do you know someone from compiler team that can review it? I'd not hesitate to merge if you can get an approval from there.

krzysz00:
Closed as done differently in #1080

PR #1077, opened 2023-05-10
Add synchronization primitives

giuseros:
This PR adds two synchronization primitives to the `ROCDL` operations. Those primitives are needed by https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/856

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
@@ -131,6 +131,18 @@ def ROCDL_BarrierOp : ROCDL_Op<"barrier"> {
   let assemblyFormat = "attr-dict";
 }
 
+def ROCDL_SetPrioOp : ROCDL_IntrOp<"s.setprio", [], [], [], 0>,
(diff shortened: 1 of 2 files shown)


external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
```diff
@@ -131,6 +131,18 @@ def ROCDL_BarrierOp : ROCDL_Op<"barrier"> {
   let assemblyFormat = "attr-dict";
 }
 
+def ROCDL_SetPrioOp : ROCDL_IntrOp<"s.setprio", [], [], [], 0>,
+  Arguments<(ins I16:$priority)> {
+  let results = (outs);
+  let assemblyFormat = "$priority attr-dict `:` type($priority)";
```
krzysz00: You don't actually have to give the type here - since it's always I16/I32, it'll be inferred

krzysz00 (approved):
Nits re assembly format (since there's a weak preference for not listing out fixed types if possible upstream), but approved

We'll probably want to edit the commit message slightly when upstreaming, but that's not too big a deal.

PR #1078, opened 2023-05-10
Enabling quantized op broadcast for migraphx integration

jerryyin:
This PR:
 - Enabled a less restrictive broadcast, so broadcasting on dequantized linear become possible
 - Added `Commutative` attribute to migraphx ops
 - Introduced an actual use case from resnet50 quantized model

This patch allows UIF quantized resnet50 model to successfully compile with MLIR enabled.


krzysz00 (approved):
LGTM

PR #1079, opened 2023-05-10
[mlir][Arith] Add pass for emulating unsupported float ops

krzysz00:
To complement the bf16 expansion and truncation patterns added to ExpandOps, define a pass that replaces, for any arithmetic operation op,
%y = arith.op %v0, %v1, ... : T
with
%e0 = arith.expf %v0 : T to U
%e1 = arith.expf %v1 : T to U
...
%y.exp = arith.op %e0, %e1, ... : U
%y = arith.truncf %y.exp : U to T

This allows for "emulating" floating-point operations not supported on a given target (such as bfloat operations or most arithmetic on 8-bit floats) by extending those types to supported ones, performing the arithmetic operation, and then truncating back to the original type (which ensures appropriate rounding behavior).

The lowering of the extf and truncf ops introduced by this transformation should be handled by subsequent passes.

(Our note: this is part 1 of "get fp8 at least kinda validatable", part 2 will be adding a `ArithToROCDL` that exposes the fp8 conversion intrinsics)

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/Arith/Transforms/Passes.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/Arith/Transforms/Passes.td
@@ -58,6 +58,28 @@ def ArithIntRangeOpts : Pass<"int-range-optimizations"> {
   }];
 }
 
+def ArithEmulateUnsupportedFloats : Pass<"arith-emulate-unsupported-floats"> {
+  let summary = "Emulate operations on unsupported floats with extf/truncf";
+  let description = [{
+    Emulate arith and vector floating point operations that use float types
+    which are unspported on a target by inserting extf/truncf pairs around all
+    such operations in order to produce arithmetic that can be performed while
+    preserving the original rounding behavior.
+
+    This pass does not attempt to reason about the operations being performed
+    to determine when type conversions can be elided.
+  }];
+
+  let options = [
+    ListOption<"sourceTypeStrs", "source-types", "std::string",
+      "MLIR types without arithmetic support on a given target">,
+    Option<"targetTypeStr", "target-type", "std::string", "\"f32\"",
+      "MLIR type to convert the unsupported source types to">,
+  ];
+
+  let dependentDialects = ["vector::VectorDialect"];
+}
+
 def ArithEmulateWideInt : Pass<"arith-emulate-wide-int"> {
   let summary = "Emulate 2*N-bit integer operations using N-bit operations";
   let description = [{
--- a/external/llvm-project/mlir/lib/Dialect/Arith/Transforms/EmulateUnsupportedFloats.cpp
+++ b/external/llvm-project/mlir/lib/Dialect/Arith/Transforms/EmulateUnsupportedFloats.cpp
@@ -1,0 +1,185 @@
+//===- EmulateUnsupportedFloats.cpp - Promote small floats --*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+// This pass promotes small floats (of some unsupported types T) to a supported
+// type U by wrapping all float operations on Ts with expansion to and
+// truncation from U, then operating on U.
+//===----------------------------------------------------------------------===//
+
+#include "mlir/Dialect/Arith/Transforms/Passes.h"
+
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Vector/IR/VectorOps.h"
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/Location.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/IR/SubElementInterfaces.h"
+#include "mlir/Transforms/DialectConversion.h"
+#include "llvm/ADT/STLExtras.h"
+#include <optional>
+
+namespace mlir::arith {
+#define GEN_PASS_DEF_ARITHEMULATEUNSUPPORTEDFLOATS
+#include "mlir/Dialect/Arith/Transforms/Passes.h.inc"
+} // namespace mlir::arith
+
+using namespace mlir;
+
+namespace {
+struct EmulateUnsupportedFloatsPass
+    : arith::impl::ArithEmulateUnsupportedFloatsBase<
+          EmulateUnsupportedFloatsPass> {
+  using arith::impl::ArithEmulateUnsupportedFloatsBase<
+      EmulateUnsupportedFloatsPass>::ArithEmulateUnsupportedFloatsBase;
+
+  void runOnOperation() override;
+};
+
+struct EmulateFloatPattern final : ConversionPattern {
+  EmulateFloatPattern(TypeConverter &converter, MLIRContext *ctx)
+      : ConversionPattern(converter, Pattern::MatchAnyOpTypeTag(), 1, ctx) {}
+
+  LogicalResult match(Operation *op) const override;
+  void rewrite(Operation *op, ArrayRef<Value> operands,
+               ConversionPatternRewriter &rewriter) const override;
+};
+} // end namespace
+
+/// Map strings to float types. This function is here because no one else needs
+/// it yet, feel free to abstract it out.
+static std::optional<FloatType> parseFloatType(MLIRContext *ctx,
+                                               StringRef name) {
+  Builder b(ctx);
+  return llvm::StringSwitch<std::optional<FloatType>>(name)
+      .Case("f8E5M2", b.getFloat8E5M2Type())
+      .Case("f8E4M3FN", b.getFloat8E4M3FNType())
+      .Case("f8E5M2FNUZ", b.getFloat8E5M2FNUZType())
+      .Case("f8E4M3FNUZ", b.getFloat8E4M3FNUZType())
+      .Case("bf16", b.getBF16Type())
+      .Case("f16", b.getF16Type())
+      .Case("f32", b.getF32Type())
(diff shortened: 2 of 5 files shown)


external/llvm-project/mlir/lib/Dialect/Arith/Transforms/EmulateUnsupportedFloats.cpp
```diff
@@ -0,0 +1,180 @@
...
+}
+
+void mlir::arith::populateEmulateUnsupportedFloatsLegality(
+    ConversionTarget &target, TypeConverter &converter) {
+  // Don't try to legalize functions and other ops that don't need expansion.
+  target.markUnknownOpDynamicallyLegal([](Operation *op) { return true; });
+  target.addDynamicallyLegalDialect<arith::ArithDialect>(
+      [&](Operation *op) -> std::optional<bool> {
+        return converter.isLegal(op);
+      });
```
jerryyin, line 130: I can only understand this to be useful, that you want to make sure the operation has legal operand and result types. What does the rest of the function do, in particular, why is it necessary to enumerate and mark a subset of vector op and arith op to also be legal? Can they go in conflict with your `addDynamicallyLegalDialect()` call here?

external/llvm-project/mlir/include/mlir/Dialect/Arith/Transforms/Passes.td
```diff
@@ -58,6 +58,28 @@ def ArithIntRangeOpts : Pass<"int-range-optimizations"> {
...
+
+    This pass does not attempt to reason about the operations being performed
+    to determine when type conversions can be elided.
+  }];
+
+  let options = [
+    ListOption<"sourceTypeStrs", "source-types", "std::string",
+      "MLIR types without arithmetic support on a given target">,
+    Option<"targetTypeStr", "target-type", "std::string", "\"f32\"",
+      "MLIR type to convert the unsupported source types to">,
```
jerryyin, line 77: For understanding, so you would require a client to explicitly specify a conversion source and target, and only stick to the conversions that are explicitly specified? (with the exception that target type is assumed to be f32)

external/llvm-project/mlir/lib/Dialect/Arith/Transforms/EmulateUnsupportedFloats.cpp
```diff
@@ -0,0 +1,180 @@
...
+    TypeConverter &converter, ArrayRef<Type> sourceTypes, Type targetType) {
+  converter.addConversion([sourceTypes = SmallVector<Type>(sourceTypes),
+                           targetType](Type type) -> std::optional<Type> {
+    if (llvm::is_contained(sourceTypes, type))
+      return targetType;
+    if (auto shaped = type.dyn_cast<ShapedType>())
+      if (llvm::is_contained(sourceTypes, shaped.getElementType()))
+        return shaped.clone(targetType);
+    // All other types legal
+    return type;
```
jerryyin, line 110: Based on this lambda, it looks like the type naturally comes from `sourceTypes` therefore it is unecessary to check whether it `is_contained()`? So it should be instead:

```cpp
    if (auto shaped = type.dyn_cast<ShapedType>())
      if (llvm::is_contained(sourceTypes, shaped.getElementType()))
        return shaped.clone(targetType);
    return targetType;
```

external/llvm-project/mlir/lib/Dialect/Arith/Transforms/EmulateUnsupportedFloats.cpp
```diff
@@ -0,0 +1,180 @@
...
+                                          "' to a known floating-point type");
+      return signalPassFailure();
+    }
+    sourceTypes.push_back(*maybeSourceType);
+  }
+  if (sourceTypes.empty())
+    (void)emitOptionalWarning(
+        std::nullopt,
+        "no source types specified, float emulation will do nothing");
+
```
jerryyin, line 169: Could you add a bit width check that, if the target type bit width is shorter, mark it as failure because we don't intend to lower the precision of computation here?

external/llvm-project/mlir/lib/Dialect/Arith/Transforms/EmulateUnsupportedFloats.cpp
```diff
@@ -0,0 +1,180 @@
...
+void EmulateFloatPattern::rewrite(Operation *op, ArrayRef<Value> operands,
+                                  ConversionPatternRewriter &rewriter) const {
+  Location loc = op->getLoc();
+  TypeConverter *converter = getTypeConverter();
+  SmallVector<Type> resultTypes;
+  assert(
+      succeeded(converter->convertTypes(op->getResultTypes(), resultTypes)) &&
+      "type conversions shouldn't fail in this pass");
+  Operation *expandedOp =
```
jerryyin, line 88: Add a condition check before this point that if `oldType == newType`, then skip everything below, and also add a unit test to make sure same type conversion doesn't add redundant extf and truncf.

jerryyin (commented):
Looks good overall, mostly questions and nitpicks.

external/llvm-project/mlir/lib/Dialect/Arith/Transforms/EmulateUnsupportedFloats.cpp
krzysz00: The chain is set up like this
- By default, all dialects have a legality callback that just returns `true`
- Except the `arith` dialect, where we check for type legality
- We also list out some `vector` ops that do arithmetic and send them down that same legality check
- And all of this doesn't apply for certain `arith` ops, like `extf` and `truncf`, which are allowed to mention the otherwise banned types
-

external/llvm-project/mlir/include/mlir/Dialect/Arith/Transforms/Passes.td
krzysz00: Yeah. Heck, that f32 default might be removable.

external/llvm-project/mlir/lib/Dialect/Arith/Transforms/EmulateUnsupportedFloats.cpp
krzysz00: Nope, this lambda gets passed all `Type`s.

external/llvm-project/mlir/lib/Dialect/Arith/Transforms/EmulateUnsupportedFloats.cpp
krzysz00: Yeah, fair, though I'd expect that to explode more reasonably later when `extf : f64 to f32` shows up. ... or maybe you do want to support that on a weird future target.

external/llvm-project/mlir/lib/Dialect/Arith/Transforms/EmulateUnsupportedFloats.cpp
krzysz00: Emulating `T` with `T` should be a pass-level error, come to think of it. I'll fix that.

external/llvm-project/mlir/include/mlir/Dialect/Arith/Transforms/Passes.td
jerryyin: It's probably reasonable to assume it for f32 from our perspective. It is indeed the golden standard of support.

external/llvm-project/mlir/lib/Dialect/Arith/Transforms/EmulateUnsupportedFloats.cpp
jerryyin: Okay, so for any other types, this falls into the `Emulating T with T` situation because we just return with whatever types that come in? If that's the case then we'd need an assert before your original `return type`? Because that really shouldn't happen

external/llvm-project/mlir/lib/Dialect/Arith/Transforms/EmulateUnsupportedFloats.cpp
jerryyin: Maybe you do, who knows... Probably still a good assert or at least a debug log to make sure someone wants to do such kind of stuff.

external/llvm-project/mlir/lib/Dialect/Arith/Transforms/EmulateUnsupportedFloats.cpp
jerryyin: Ok thanks for the background.

> And all of this doesn't apply for certain arith ops, like extf and truncf, which are allowed to mention the otherwise banned types

If that's the case we can mark only those two `illegal` and everything else to be legal? Which seems to be easier then enumerating the legal ops?

external/llvm-project/mlir/lib/Dialect/Arith/Transforms/EmulateUnsupportedFloats.cpp
jerryyin: Agreed

external/llvm-project/mlir/lib/Dialect/Arith/Transforms/EmulateUnsupportedFloats.cpp
krzysz00: The point is that all `arith` operations, except the named ones, are only legal when they don't mention one of the source types. Same goes for certain named vector operations.

external/llvm-project/mlir/lib/Dialect/Arith/Transforms/EmulateUnsupportedFloats.cpp
krzysz00: No, in a type converter, saying `T` converts to `T` means "`T` is legal". For example, if this lambda were called with `i32`, or `memref<4xbf16>`, those should pass through unmodified.

krzysz00:
@jerryyin Can I get another look?

sjw36 (approved):
Looks good.

PR #1080, opened 2023-05-11
[DO NOT SQUASH] Disable the miscompiling perf config(s)

krzysz00:
Since the problem we're seeing only happens on a specific perf config, go ahead and disable that perf config since it looks like that'll be fine.

Also revert the previous workaround because it didn't help.


jerryyin (approved):
LGTM.

So parameter sweep will skip this because this is marked as inapplicable? That sounds about right.

krzysz00:
Yep, that's the plan.

PR #1082, opened 2023-05-12
[DO NOT SQUASH] Apply the compiler fix for the LDS miscomplies

krzysz00:
Revert the earlier temporary disabling of certain perf configs as the underlying issue has been resolved by the patch included here.

The underlying upstream patch is https://reviews.llvm.org/D150246


jerryyin (approved):
LGTM!

Could you update their PR address in your pull request description? Just for future/upstream merge reference.

PR #1083, opened 2023-05-12
Change conv-configs layouts to NCHW, raise exception on errors

pcf000:
MIOpen doesn't accept NCHWG as in the original commit, only NCHW, NHWC, CHWN, NCDHW, and NDHWC.  The parser in perfRunner.py is more generous.

Raise an exception in perfRunner.py for this kind of error, to keep it from happening silently.  Suppress warnings from deleting a file that isn't there, because they annoy me.

Emit the elapsed time for the conv configs, necessary for perf reporting.

For issue [#880](https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/880).


mlir/utils/performance/perfRunner.py
```diff
@@ -33,7 +33,7 @@
 DATA_TYPES_GEMM = ['f32', 'f16', 'i8']
 
 # Compiled regexp object used for extracting elapsed time from MIOpenDriver's output
-ELAPSED_TIME_RE = re.compile(r"Elapsed: (.*)ms")
+ELAPSED_TIME_RE = re.compile(r"Elapsed: ([0-9\.]*) ms")
```
pcf000, line 36: I briefly used "--wall 2" which printed a line like "Elapsed: 9 ms  aux time 3 ms" and the original regexp matched "9 ms  aux time 3".  But it also printed two lines with elapsed time so I stopped using it.

mlir/utils/performance/perfRunner.py
```diff
@@ -462,7 +462,7 @@ def __init__(self, dtype: str, direction: str, filterLayout: str, inputLayout:st
     @classmethod
     def benchmarkExternal(cls, commandLine, paths: Paths, arch, envs=dict()):
         config = cls.fromCommandLine(commandLine, arch)
-        MIOpenDriverCommand = [paths.miopen_driver_path, *commandLine, '-V', '0']
+        MIOpenDriverCommand = [paths.miopen_driver_path, *commandLine, '-V', '0', '-t', '1']
```
pcf000, line 468: Needed to print the elapsed time.

mlir/utils/performance/perfRunner.py
```diff
@@ -472,6 +472,7 @@ def benchmarkExternal(cls, commandLine, paths: Paths, arch, envs=dict()):
             outs, errs = p1.communicate(timeout=300)
             if len(errs) > 0:
                 print("MIOpen benchmark produced errors: ", errs.decode('utf-8'))
+                raise OSError(errs.decode('utf-8'))
```
pcf000: Timeouts are caught and ignored.  I wanted parse errors like these to cause a test failure and draw attention.  If we have other conditions that should be ignored, we can add them, but maybe should abstract the whole idiom.  Could also accumulate the "errs" strings and signal at the end of the loop if it's not empty.

mlir/utils/performance/perfRunner.py
```diff
@@ -627,7 +628,7 @@ def benchmarkExternal(cls, commandLine, paths: Paths, arch, envs=dict()):
             raise ValueError("rocblas-benchmark-driver not built")
         benchmarkArgs = config.generateMlirDriverCommandLine("")
         # remove the result file generated by rocprof in previous benchmarking
-        os.system("rm "+BENCHMARKING_RESULT_FILE_NAME)
+        os.system("rm -f "+BENCHMARKING_RESULT_FILE_NAME)
```
pcf000, line 635: Got tired of "file not found: results.stats.csv" over and over.

manupak:
lets run a nightly on this PR? (I can do that for you if you want)

I'd like to see those stages verified in the PR itself (rather than skipped)

pcf000:
Hang on, Yi has pointed out some other flaws (over on the issue) and I should address those first.  Argh.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -865,13 +866,12 @@ pipeline {
                             dir('build') {
                                 sh 'date --utc +%Y-%m-%d > perf-run-date'
                                 // Run MLIR perf benchmarks
-                                sh 'python3 ./bin/perfRunner.py'
```
pcf000, line 863: This run used the default conv configs.  Added those into the conv-configs file, and now we don't need this run, which avoids overwriting the implicit output file.

mlir/utils/performance/perfRunner.py
```diff
@@ -472,7 +472,8 @@ def benchmarkExternal(cls, commandLine, paths: Paths, arch, envs=dict()):
             outs, errs = p1.communicate(timeout=300)
             if len(errs) > 0:
                 print("MIOpen benchmark produced errors: ", errs.decode('utf-8'))
-                raise OSError(errs.decode('utf-8'))
+                if p1.returncode != 0:
```
pcf000, line 478: This is happening:
```
Running MIOpen Benchmark:  conv -F 1 -f NCHW -I NCHW -O NCHW -n 256 -c 512 -H 7 -W 7 -k 512 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1

MIOpen benchmark produced errors:  MIOpen(HIP): Warning [FindSolutionImpl] Invalid config loaded from Perf Db: ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC: fwd,nhwc,fp32,0,1,128,128,16,32,32,2,1,1,2,2,0,0,1,0,0,1,4,2,1,1,4,1,64,1,4,2,1,1,4,1,64,0. Performance may degrade.
```
Before I started raising signals, it would just message and continue, but with a signal, it halts.  I don't know where the invalid configs come from, but to preserve existing behavior I check the return code to see if the process itself failed.

pcf000:
Fixed a couple of mistakes that Yi noticed, and changed the signal-raising slightly for cases that warn but don't fail.

yiqian1:
I just realized that tuningRunner appends new configs to existing ones, instead overwriting them. Let's keep one single database for both conv and gemm. Apologies for the confusion.

mlir/utils/jenkins/Jenkinsfile
yiqian1: The resnet50 configs are used to compare MLIR vs. MIOpen with tuned and untuned MLIR kernels, while conv-configs are run with MLIR, tuned MLIR and MIOpen with no MLIR kernels. Currently createPerformanceReports.py combines the results of this perfRunner.py and two previous perfRunner executions (with --miopen* options) to create html reports. It won't work for the results of  running perfRunner with conv-configs. It's better to keep resnet50 configs and conv-configs separate and have two versions of createPerformancerReport to create different html reports. Alternatively add conv-configs into the resnet50 config file to get tuned/untuned MIOpen results, if we're interested in those performance and can tolerate the extended testing time.

mlir/utils/performance/createPerformanceReports.py
```diff
@@ -9,39 +9,33 @@
...
+    if df.empty:
+        return
+    tuned_df = readReport(reportUtils.MIOPEN_TUNED_REPORT_FILE,
+                          'MIOpen with tuned MLIR')
+    untuned_df = readReport(reportUtils.MIOPEN_UNTUNED_REPORT_FILE,
+                            'MIOpen with untuned MLIR')
 
     # Add tuned and untuned performance to the existing performance table
-    if tunedReportFound == True and untunedReportFound == True :
+    if tuned_df.empty and untuned_df.empty:
```
yiqian1: If they are NOT empty.

yiqian1 (approved):

PR #1084, opened 2023-05-15
Add user interface support for WMMA

giuseros:
This PR is adding user interface code to rocMLIR to execute (and test) WMMA. Main changes:
- Abstract accelerator specifics behind an `PopulateParamsAccel` which has `wmma` and `xdlops` as subclasses 
- Check for `isAccel()` instead of `isXdlops` in many places of `AffixTuningParam`, `GridwiseGemmParams` `GemmToGridwise` and `rocmlir-gen`. 

The main issue I found was that `isAcces != (isXdlops || isWmma)`, because wmma can only run on fp16 and int8 (and not on fp32). So `isAccel` is a function of the feature set and of the input datatype. 

**Note**: I tested it on my local system and it works. PR testing should confirm if this works also with xdlops, so that we can mark this as "tested"


mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
```diff
@@ -266,6 +266,11 @@ LogicalResult Conv2dGenerator::hasValidChip() const {
   if (bitEnumContainsAll(config.features, GemmFeatures::mfma) &&
       (chipHexNumber != 0x908 && chipHexNumber != 0x90a))
     return failure();
+
+  // WMMA is only supported on gfx1100
```
krzysz00: Hold on a minute, WMMA is on all gfx11xx, let's support all of those preemptively.

mlir/test/mlir-rock-lib/applicability/chip.mlir
```diff
@@ -4,7 +4,7 @@
...
-// RUN: ! rocmlir-lib-test --args " --operation conv2d --arch amdgcn-amd-amdhsa:gfx1100 --num_cu 64 --in_type fp32 --fil_type fp32 --out_type fp32 --fil_layout GNCHW --in_layout NGCHW --out_layout NGCHW --batchsize 64 --in_channels 1024 --out_channels 1024 --in_h 14 --in_w 14 --out_h 14 --out_w 14 --fil_h 1 --fil_w 1 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h 0 --padding_w 0 --kernel_name conv2d_nchw_kcyx_nkhw --groupsize 1" --option tuningparams
+// RUN: ! rocmlir-lib-test --args " --operation conv2d --arch amdgcn-amd-amdhsa:gfx1130 --num_cu 64 --in_type fp32 --fil_type fp32 --out_type fp32 --fil_layout GNCHW --in_layout NGCHW --out_layout NGCHW --batchsize 64 --in_channels 1024 --out_channels 1024 --in_h 14 --in_w 14 --out_h 14 --out_w 14 --fil_h 1 --fil_w 1 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h 0 --padding_w 0 --kernel_name conv2d_nchw_kcyx_nkhw --groupsize 1" --option tuningparams
```
krzysz00, line 9: 1100?

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2784,6 +2784,17 @@ static ModuleOp generateKernel(MLIRContext *context, GenParams &genParams,
       enabledFeatures =
           bitEnumSet(enabledFeatures, rock::GemmFeatures::atomic_fmax_f32,
                      atomicFMaxF32Feature == FeatureToggle::on);
+
+    if (wmmaFeature != FeatureToggle::infer) {
+      enabledFeatures = bitEnumSet(enabledFeatures, rock::GemmFeatures::wmma,
+                                   wmmaFeature == FeatureToggle::on);
+    } else {
+      Type elemType = typeFromString(tensorDataType, context);
+      if (!elemType.isF16() && !elemType.isInteger(8)) {
```
krzysz00: WMMA also supports bf16?

krzysz00 (commented):
Minor nitpicks, reasonable change otherwise

mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
```diff
@@ -25,6 +25,25 @@ static Type getRetType(Type inputType) {
...
+      return false;
+    }
+    return true;
+  } else {
+    if (kPerBlock < inputLen) {
+      LLVM_DEBUG(llvm::dbgs()
+                 << "When KPack is 1, KPerBlock must be at least k_base\n");
+      return false;
+    }
+    return true;
```
jerryyin: (I might be wrong, but) I don't quite remember we have such limitation for existing xdlops path? It looks like you explicitly rejected configs with smaller K

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
```diff
@@ -206,121 +216,154 @@ template <typename InitParamType> class BasePopulateParams {
...
+  // Tuning parameters for fp16/bf16 convolutions.
+  static const InitParamsAccel initParametersFp16[nInitParametersFp16];
 
-  std::vector<InitParamsXDL> getTuningParameters(KernelType opType,
-                                                 Type dataTypeA, Type dataTypeB,
-                                                 StringRef arch) const;
+  static constexpr size_t nInitParametersForward8Bit = 5;
+  // Tuning parameters for i8 convolutions.
+  static const InitParamsAccel
+      initParametersForward8Bit[nInitParametersForward8Bit];
```
jerryyin, line 324: Why hard code the array size? I think it is perfect valid to declare it without size but initialize it in cpp file.

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
```diff
@@ -206,121 +216,154 @@ template <typename InitParamType> class BasePopulateParams {
...
 
-class PopulateParams : public BasePopulateParams<InitParamsNonXDL> {
+//
+// Non acceleration parameter initialization interface
+//
+class PopulateParams : public BasePopulateParams<InitParamsNonAccel> {
 private:
   static constexpr size_t nInitParameters = 21;
-  static const InitParamsNonXDL initParameters[nInitParameters];
+  static const InitParamsNonAccel initParameters[nInitParameters];
```
jerryyin, line 225: Same here, why hard-code the array size?

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
```diff
@@ -19,11 +19,29 @@ using namespace mlir::rock;
...
+  if (!isWmma)
+    return false;
+
+  Type elementType = dataType;
+  if (auto vectorType = dataType.dyn_cast<VectorType>())
+    elementType = vectorType.getElementType();
+  else if (auto memrefType = dataType.dyn_cast<MemRefType>())
+    elementType = memrefType.getElementType();
+
+  return (elementType.isF16() || elementType.isInteger(8));
```
jerryyin: Add a debug statement that we reverted the bit enum when it comes to an unsupported type?

Coming to think of it, I'm not sure I like this approach: I'm leaning towards we don't call unsupported WMMAs, and give an explicit error such that user knows that scenario is unsupported. Comparatively speaking, you want to continue the execution but override `GemmFeatures::wmma` decision that has already been required.

mlir/test/rocmlir-driver/populate.mlir
```diff
@@ -1,7 +1,7 @@
-// RUN: rocmlir-gen --arch %arch -p -mfma=off -atomic_add=off -atomic_fmax_f32=off -dot=on | FileCheck %s --check-prefix=F32
```
jerryyin, line 1: `-mfma=off -wmma=off -atomic_add=off` seems to be redundant as they should by default be off?

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -132,6 +132,12 @@ makeRockConv2D(ConversionPatternRewriter &rw, Operation *op, Value input,
   std::optional<uint32_t> num_cu;
   rock::GemmFeatures features;
   std::tie(arch, num_cu, features) = getArchAttributes(op);
+  // Even if acceleration is supported by the arch, datatype might not be
+  // valid accelerated data types
+  if (!rock::isAccel(features, input.getType())) {
+    features = bitEnumClear(features, rock::GemmFeatures::mfma |
+                                          rock::GemmFeatures::wmma);
+  }
```
jerryyin: Following the last comment, I don't quite like this approach that in middle of lowering we'd override the `features` in unsupported scenarios. If we want to have such kind of conditions it should happen very early in a pipeline (or before the lowering started, like in conv2d generator).

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
```diff
@@ -149,29 +150,33 @@ GemmRewritePattern::matchAndRewrite(GemmOp op, GemmOpAdaptor adaptor,
...
+          op.getStoreMethodAttr(), blockSize, gridSize,
+          params.cast<XdlopsGemmParamsAttr>());
+      rw.eraseOp(op);
+    } else if (isWmma) {
+      // Onne the attribute copies are gone, make this a replaceOp
+      rw.create<GridwiseGemmAccelOp>(
+          loc, a, b, c, op.getArchAttr(), op.getFeaturesAttr(),
+          op.getStoreMethodAttr(), blockSize, gridSize,
+          params.cast<WmmaGemmParamsAttr>());
+      rw.eraseOp(op);
```
jerryyin: Can we pass in the parent tuning parameters type but dynamically cast them when in gridwise gemm op? It looks inconvenient that we have to do such kind of branching considering that you already have both ops unified under the same type.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -507,46 +544,133 @@ PopulateParamsXDL::getTuningParameters(KernelType opType, Type dataTypeA,
...
+    std::make_tuple(64, 32, 2),
+    std::make_tuple(32, 64, 2),
+    std::make_tuple(32, 32, 2),
+    std::make_tuple(32, 16, 2),
+    std::make_tuple(16, 32, 2),
+    std::make_tuple(32, 32, 2),
+    std::make_tuple(64, 16, 2),
+    std::make_tuple(16, 64, 2),
+    std::make_tuple(16, 16, 2),
+  };
```
jerryyin, line 601: This seems to be a copy-paste from xdlops gemm. Now will some configuration such as 16x128 ever become performant in those combinations? How does CK handle those?

jerryyin (commented):
It looks like you have tuning part of the code ready. Have you done any testing related with tuning yet?

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
krzysz00: Because that's how I did it back in the day. I think there's some sort of `std::array` stuff you can use to infer the size. I also think there's an issue where you can't have a variable-sized static constant in a class like that because the compiler doesn't know how much `.rodata` it needs for it.

mlir/test/rocmlir-driver/populate.mlir
krzysz00: Nah, if `%arch` supports them, they'll be on

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
krzysz00: Yeah, it shouldn't be needed to have this if statement here. You could go with `isXdlops || isWmma` and cast to the interface

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
giuseros: The problem here is that I made so that wmma is by default on for gfx1100. Should I change this? I guess it makes sense, because the architecture won't be sufficient to determine what features are enabled. So to enable wmma the user should always need to ask for it. What do you think?

krzysz00: I'd suggest that the function for default arch features grows a data type argument, or that we turn off WMMA for f32, or so on

giuseros: The data type argument was why I added  `isAccel(features, dataTYpe)`, so I can get the features + datatype and check if I can run the acceleator gemms. I guess @jerryyin is worried that those things might fail silently if we assume that the feature is on, but somehow the compiler decides to not use it. This is why I was thinking to turn the wmma feature off on gfx11 and then erroring if the user wants `wmma=on` for the wrong data type.

krzysz00: So there's two separate issues here
1. Being able to control whether WMMA is on/off despite the defaults
2. That when the architecture supports wmma, that only applies for f16 and i8
3. We need to know what kind of gemm we'll be doing when we create the gemm so we can, for example, handle perf configs correctly

I claim that what we want to do here is that, instead of just looking up the architecture features in the arch db directly, we want a function that wraps around that that takes the arch name and the data type and masks off features that aren't relevant - specifically, f32 wmma.

Then, having wmma be `infer` works correctly, constructing GEMMs etc. from tosa works correctly, and so on ... _and_ if I decide to do `-t f32 -wmma=on` this'll generate that kernel (which will promptly trip on an error message but anyway)

mlir/test/mlir-rock-lib/applicability/chip.mlir
giuseros: This is a negative test. So I need an architecture that is not supported

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
giuseros: I don't think they do any sort of tuning yet (from what I see). I mean it's not too different from Xdlops. I added some configuration, but I plan on pruning it once I get down to tuning

mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
giuseros: The message is wrong, but (I think) [we have a similar limitation for in the xdlops path](https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/ef74abec7457f1446c8707161d6e48505a4c446e/mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp#L438). For wmma  `k_base` is 16 (the `inputLen`) and we don't have to care about reduction vs broadcast.

giuseros:
I restructured the code in the way @krzysz00 was proposing. Now the ArchDB is providing a function that checks the dataType (for wmma) before returning the default features. The user can override the default at her own risk, but then an error will be generated during the lowering (if the dataytpe is not compatible with the feature). 

Re Tuning, @jerryyin  I plan to give a go at tuning once this is in (so that I can focus solely on performance). 

There is a last doubt about the param list being a static array (with the array size hardcoded) - we can follow up in the comments.

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
giuseros: Yes, the code was already like this originally. I thought the limitation was also that this would be serialized, used by static libraries, and this is why we wanted a plain array instead of something fancier. Should I try something else?

giuseros: @jerryyin @krzysz00 any further thoughts? I guess this is the last standing point to get this merged?

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -60,3 +60,32 @@ AmdArchInfo mlir::rock::lookupArchInfo(StringRef arch) {
...
+mlir::rock::AmdArchInfo::getDefaultFeatures(StringRef strDataType) {
+  GemmFeatures theseFeatures = defaultFeatures;
+  bool isWmma = bitEnumContainsAll(theseFeatures, GemmFeatures::wmma);
+  if (isWmma) {
+    if (strDataType != "f16" && strDataType != "bf16" && strDataType != "i8") {
+      theseFeatures = bitEnumClear(theseFeatures, GemmFeatures::wmma);
+    }
+  }
+  return theseFeatures;
+}
```
jerryyin: This seems unnecessary? We have a `Conv2dGenerator::getDataType()` function that does type conversion. You should extract that to a header so this function doesn't have to operate on a string. Besides, having a duplicate compared to `getDefaultFeatures(Type dataType)` hard to maintain.

mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
```diff
@@ -105,7 +105,9 @@ void AffixTuningParameters::affixTuningParametersImpl(
     perfConfig = perfConfigAttr.getValue().str();
   }
   GemmFeatures features = op.getGemmFeatures();
-  if (rock::isAccel(features, op.getAType())) {
+  bool isAccel =
+      bitEnumContainsAny(features, GemmFeatures::wmma | GemmFeatures::mfma);
```
jerryyin: This applies to lots of places. I think it has appeared enough times to justify a dedicated function

```cpp
bool isAccel(GemmFeatures features){
   return bitEnumContainsAny(features, GemmFeatures::wmma | GemmFeatures::mfma);
}
```

So you'd just need to call this function instead of having the expand out the specific features everywhere.

jerryyin (commented):
> The user can override the default at her own risk, but then an error will be generated during the lowering (if the dataytpe is not compatible with the feature).

May I ask where this error string is? I was looking through the changes you made but can't find it.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
giuseros: You are 100% right! The reason I had two versions was because in the library we check the config "before" we build the module. But I didn't realize I could do the other way round. I just did, and got rid of the string version.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -502,7 +502,7 @@ static LogicalResult verifyGemmTypes(Operation *op, GemmFeatures features,
   if (bitEnumContainsAll(features, GemmFeatures::wmma)) {
     if (!(elemTypeA.isF16() || elemTypeA.isBF16() || elemTypeA.isInteger(8))) {
       return op->emitOpError(
-          "Wmma gridwise supports only F16 and Int8 data types");
+          "Wmma gridwise supports only F16/BF16/int8 data types");
```
giuseros, line 505: > May I ask where this error string is? I was looking through the changes you made but can't find it.

@jerryyin  the error is thrown here (it was actually wrong/incomplete so I improved it)

jerryyin: Thanks! This is indeed a reasonable place for checking it!

jerryyin (approved):
LGTM

PR #1088, opened 2023-05-18
[mlir][AMDGPU] Add packed 8-bit float conversion ops and lowering

krzysz00:
Define operations that wrap the gfx940's new operations for converting between f32 and registers containing packed sets of four 8-bit floats.

Define rocdl operations for the intrinsics and an AMDGPU dialect wrapper around them (to account for the fact that MLIR distinguishes the two float formats at the type level but that the LLVM IR does not).

Define an ArithToAMDGPU pass, meant to run before conversion to LLVM, that replaces relevant calls to arith.extf and arith.truncf with the packed operations in the AMDGPU dialect. Note that the conversion currently only handles scalars and vectors of rank <= 1, as we do not have a usecase for multi-dimensional vector support right now.

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
@@ -390,4 +390,83 @@ def ROCDL_DPPMovOp : ROCDL_IntrOp<"mov.dpp", [], [0],
   }];
 }
 
+//===---------------------------------------------------------------------===//
+// 8-bit float intrinsics
+//===---------------------------------------------------------------------===//
+def ROCDL_CvtF32Bf8Op :
+    ROCDL_IntrOp<"cvt.f32.bf8", [], [], [Pure], 1>,
+    Arguments<(ins I32:$srcA, I32:$byteSel)> {
+  let summary = "Convert bf8 to f32";
+  let description = [{
+    Convert 8-bit bf8 value from the `byteSel`th bit of `srcA` to fp32.
+  }];
+  let assemblyFormat = [{
+    attr-dict $srcA `[` $byteSel `]` `:` type($res)
+  }];
+}
+
+def ROCDL_CvtF32Fp8Op :
+    ROCDL_IntrOp<"cvt.f32.fp8", [], [], [Pure], 1>,
+    Arguments<(ins I32:$srcA, I32:$byteSel)> {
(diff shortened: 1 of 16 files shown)


external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
```diff
@@ -390,4 +390,83 @@ def ROCDL_DPPMovOp : ROCDL_IntrOp<"mov.dpp", [], [0],
...
+  let description = [{
+    Convert `srcA` and `srcB` to bf8 and store into the low/high word of
+    `old`, preserving the other word.
+  }];
+  let assemblyFormat = [{
+    attr-dict $srcA `,` $srcB `->` $old `[` $wordSel `]` `:` type($res)
+  }];
+}
+
+def ROCDL_CvtPkFp88F32Op :
```
sjw36: too many 8's?

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -638,6 +640,161 @@ struct WMMAOpLowering : public ConvertOpToLLVMPattern<WMMAOp> {
...
+                                                    wordSel);
+  }
+  return success();
+}
+
+LogicalResult PackedTruncFp8x2OpLowering::matchAndRewrite(
+    PackedTruncFp8x2Op op, PackedTruncFp8x2OpAdaptor adaptor,
+    ConversionPatternRewriter &rewriter) const {
+  Location loc = op.getLoc();
+  if (chipset.majorVersion != 9 || chipset.minorVersion < 0x40)
```
sjw36, line 730: Move to an arch db lookup.

sjw36 (approved):
Looks good overall. 

Broadly speaking I would advocate for a more generalized archdb expanded for these features so we don't have the explicit chipset checks sprinkled around.

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
krzysz00: This is external, there is no arch db, chipset checks are how this pass has been for a while.

Though coordinating with the compiler team to expose the feature list is a long-term option.

PR #1089, opened 2023-05-18
Adding migraphx quant-dot op support

jerryyin:
This PR adds support for `migraphx.quant_dot` and its lowering. With it, migraphx kernel is able to do `rock.gemm()` computation on int8 type.

This PR does the following:
 - Abstracted the dot op into a common base: MIGraphX_DotOpBase
 - A number of bug fixes in the dot op lowering to enabling i8 support
 - Conversion tests and end to end tests

### DIFF
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -124,8 +124,7 @@ public:
     auto input_t = op.getInput();
     auto filter_t = op.getFilter();
     auto results = op->getResults();
-    auto elementTy =
-        input_t.getType().template cast<ShapedType>().getElementType();
+    auto elementTy = input_t.getType().getElementType();
     auto outputTy = results[0].getType().template cast<ShapedType>();
     SmallVector<int64_t> NCHW2NHWC{0, 2, 3, 1};
     SmallVector<int64_t> NHWC2NCHW{0, 3, 1, 2};
@@ -177,9 +176,9 @@ public:
                             {padTop, padBottom, padLeft, padRight}));
 
     // Convert optional attributes
-    if (auto attr = op->getAttr("xdlopsV2"))
+    if (auto attr = (*op).template getAttrOfType<BoolAttr>("xdlopsV2"))
       cop->setAttr("xdlopsV2", attr.template cast<BoolAttr>());
-    if (auto attr = op->getAttr("perf_config"))
+    if (auto attr = (*op).template getAttrOfType<BoolAttr>("perf_config"))
       cop->setAttr("perf_config", attr.template cast<StringAttr>());
 
     // Note: For TOSA convolution, a non-float type is considered as a
@@ -298,9 +297,8 @@ public:
     TypedValue<TensorType> in_A = op.getInA();
     TypedValue<TensorType> in_B = op.getInB();
     auto results = op->getResults();
-    auto elementTy =
-        in_A.getType().template cast<ShapedType>().getElementType();
-    ShapedType outputTy = results[0].getType().template cast<ShapedType>();
+    auto elementTy = in_A.getType().getElementType();
+    ShapedType outputTy = results[0].getType();
 
     // check batch dimension. Tosa matmul only allow a single dimension for it,
     // add reshape ops to flatten and restore the original dimension.
@@ -364,10 +362,10 @@ public:
     auto mop = rewriter.create<tosa::MatMulOp>(loc, newOutType, in_A, in_B);
 
     // Convert optional attributes
-    if (auto attr = op->getAttr("xdlopsV2"))
-      mop->setAttr("xdlopsV2", attr.template cast<BoolAttr>());
-    if (auto attr = op->getAttr("perf_config"))
-      mop->setAttr("perf_config", attr.template cast<StringAttr>());
+    if (auto attr = (*op).template getAttrOfType<BoolAttr>("xdlopsV2"))
+      mop->setAttr("xdlopsV2", attr);
+    if (auto attr = (*op).template getAttrOfType<BoolAttr>("perf_config"))
+      mop->setAttr("perf_config", attr);
 
     // Note: For TOSA matmul, a non-float type is considered as a
     // quantized convolution. For quantized convolution, it is required
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -177,9 +177,9 @@ public:
 
     // Convert optional attributes
     if (auto attr = (*op).template getAttrOfType<BoolAttr>("xdlopsV2"))
-      cop->setAttr("xdlopsV2", attr.template cast<BoolAttr>());
-    if (auto attr = (*op).template getAttrOfType<BoolAttr>("perf_config"))
-      cop->setAttr("perf_config", attr.template cast<StringAttr>());
+      cop->setAttr("xdlopsV2", attr);
+    if (auto attr = (*op).template getAttrOfType<StringAttr>("perf_config"))
+      cop->setAttr("perf_config", attr);
 
     // Note: For TOSA convolution, a non-float type is considered as a
     // quantized convolution. For quantized convolution, it is required
@@ -364,7 +364,7 @@ public:
     // Convert optional attributes
     if (auto attr = (*op).template getAttrOfType<BoolAttr>("xdlopsV2"))
       mop->setAttr("xdlopsV2", attr);
-    if (auto attr = (*op).template getAttrOfType<BoolAttr>("perf_config"))
+    if (auto attr = (*op).template getAttrOfType<StringAttr>("perf_config"))
       mop->setAttr("perf_config", attr);
 
     // Note: For TOSA matmul, a non-float type is considered as a
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -285,25 +285,28 @@ public:
   }
 };
 
-class DotConverter final : public OpConversionPattern<migraphx::DotOp> {
+template <typename DotType>
+class DotConverter final : public OpConversionPattern<DotType> {
 public:
-  using OpConversionPattern<migraphx::DotOp>::OpConversionPattern;
+  using OpConversionPattern<DotType>::OpConversionPattern;
 
   LogicalResult
-  matchAndRewrite(migraphx::DotOp op, OpAdaptor adaptor,
+  matchAndRewrite(DotType op,
+                  typename OpConversionPattern<DotType>::OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const final {
     Location loc = op->getLoc();
     TypedValue<TensorType> in_A = op.getInA();
     TypedValue<TensorType> in_B = op.getInB();
     auto results = op->getResults();
     auto elementTy =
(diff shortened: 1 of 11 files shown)


jerryyin:
Gentle ping for review @sjw36 @manupak

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -361,13 +364,27 @@ class DotConverter final : public OpConversionPattern<migraphx::DotOp> {
     auto mop = rewriter.create<tosa::MatMulOp>(loc, newOutType, in_A, in_B);
 
     // Convert optional attributes
-    if (auto attr = op->getAttrOfType<BoolAttr>("xdlopsV2"))
-      mop->setAttr("xdlopsV2", attr);
-    if (auto attr = op->getAttrOfType<StringAttr>("perf_config"))
-      mop->setAttr("perf_config", attr);
+    if (auto attr = op->getAttr("xdlopsV2"))
+      mop->setAttr("xdlopsV2", attr.template cast<BoolAttr>());
```
sjw36: This will assert if given the wrong attr type (i.e. UnitAttr). I would keep the former since you are expecting these types anyway.

sjw36 (changes requested):
Looks good overall. Just need to fix one below.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jerryyin: > This will assert if given the wrong attr type (i.e. UnitAttr)

Not sure I catch your point. It was previously "assumed" to be `BoolAttr` as well. (So keeping the former way will still not allow a different attr type.) The only change I made is to move it from `getAttr` side to `setAttr` side. The reason why I need that is due to it is easier to this whole class is not templatized and it is easier for compiler to comply with C++ dot operator on the second line. 

Refer to: https://stackoverflow.com/questions/3505713/c-template-compilation-error-expected-primary-expression-before-token

sjw36: In the former, the `getAttrOfType` will return a null `attr` type if the op 1) does not contain the attr named or 2) contains the attr named but of a different type. Also, the `attr` will always have the specified type (ie. StringAttr).

In the latter, the 'getAttr' will return a generic Attribute which will fail the cast if it is a different type (UnitAttr vs. BoolAttr).

If you are having compile problems, you can explicitly assign type to `attr` accordingly (BoolAttr or StringAttr) in the conditional statement.

sjw36: You can try a test with `xdlopsV2` as a UnitAttr:
> migraphx.dot ... attributes {xdlopsV2, perf_config = "..."} ...

Should be ignored in the first case and assert in the second case.

sjw36: If you prefer it not to silently ignore, you can use your method but check the type and error out accordingly before the explicit cast.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -285,25 +285,28 @@ class MultiBroadcastConverter final
...
+                  typename OpConversionPattern<DotType>::OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const final {
     Location loc = op->getLoc();
     TypedValue<TensorType> in_A = op.getInA();
     TypedValue<TensorType> in_B = op.getInB();
     auto results = op->getResults();
     auto elementTy =
-        op->getOperand(0).getType().cast<ShapedType>().getElementType();
-    ShapedType outputTy = results[0].getType().cast<ShapedType>();
+        in_A.getType().template cast<ShapedType>().getElementType();
```
manupak: This shouldn't need the template cast<ShapedType> now as it is a TypedValue< TensorType >. i.e. getElementType(); should work out of the box.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -285,25 +285,28 @@ class MultiBroadcastConverter final
...
+    ShapedType outputTy = results[0].getType().template cast<ShapedType>();
```
manupak: same here

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -313,7 +316,7 @@ class DotConverter final : public OpConversionPattern<migraphx::DotOp> {
     // A, B, Out have the same rank. rank=2 assumes batch=1.
     // Here handling special cases.
     if (outRank != 3 || rankA != rankB ||
-        (outRank == 3 && orgDimsA != orgDimsB)) {
+        (outRank == 3 && orgDimsA[0] != orgDimsB[0])) {
```
manupak, line 317: Why this change ?

manupak (commented):
few comments and questions

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jerryyin: Good point! I didn't realize that this is already implicitly casted to a `TensorType` (instead of just a generic `Type`) and who has a method `getElementType()`

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jerryyin: This was a bug before. `orgDimsA` will always be different than `orgDimsB`, because the layout is [m, k] * [k, n]. So the effect of the bug is that we'd almost always generate a pair of redundant reshape before and after the dot op.

I think the original intent was to allow the batch dimension to broadcast when they don't align. So I did exactly that check in here that make the redundant reshape to go away.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
jerryyin: Good point, thanks for elaborating it. Indeed spell out the type of the attribute in the conditional is a more defensive way to do it. Addressed in d7b8ad8

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
manupak: I see... 
Can you add a test for this ? maybe in mixr_to_tosa ?

jerryyin: This branch will realistically never happen because migraphx won't implicitly broadcast, but rather always use a multibroadcast op to achieve this. So when it comes to this op, the first dimension must be same between A and B. The original intention to support such kind of broadcast is kind of mis-leading.

But sure, I can add a test to make sure that unrealistic use case do not yield a bug.

jerryyin: Done in 17dc34ef9b0913693dd54c4144a84e4e3dd462a3

mlir/test/Conversion/MIGraphXToTosa/mixr-to-tosa-ops.mlir
```diff
@@ -215,11 +222,21 @@ module  {
...
     %2 = "migraphx.mul"(%0, %arg2) {} : (tensor<1x5x3xf32>, tensor<1x5x3xf32>)-> tensor<1x5x3xf32>
     return %2 : tensor<1x5x3xf32>
   }
 
+  // CHECK-LABEL: func.func @func_mul_bcast
+  // CHECK: tosa.reshape{{.*}}1x10x4
+  // CHECK: tosa.reshape{{.*}}1x4x3
+  // CHECK: tosa.matmul
+  // CHECK: tosa.reshape{{.*}}2x5x3
+  func.func @func_mul_bcast(%arg0: tensor<2x5x4xf32>, %arg1: tensor<1x4x3xf32>, %arg2: tensor<2x5x3xf32>) -> tensor<2x5x3xf32> attributes{kernel, arch = ""} {
```
manupak: Thanks! this should've been here :) .
However, I was more asking for CHECK-NOT: tosa.reshape as you were saying w/o your change a redundant reshape is added but I dont want to hold this on that account.

manupak (approved):
LGTM bar other's comments.

mlir/test/Conversion/MIGraphXToTosa/mixr-to-tosa-ops.mlir
jerryyin: You are right. On a second look the reshape has already been checked over here:

https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/db5e885579222a7cf28b5158bd70d78dbd5dec83/mlir/test/Conversion/MIGraphXToTosa/mixr-to-tosa-ops.mlir#L73-L95

I was wrong to say that this is an unrealistic use case. Since this test is better than the one I have written, I deleted the case in f3ab7eb.

I also added the `CHECK-NOT` according to your suggestion.

jerryyin:
@sjw36 Let me know if you have any additional concerns? Requested change has been made and your review is blocking this PR from merging.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -361,13 +362,27 @@ class DotConverter final : public OpConversionPattern<migraphx::DotOp> {
     auto mop = rewriter.create<tosa::MatMulOp>(loc, newOutType, in_A, in_B);
 
     // Convert optional attributes
-    if (auto attr = op->getAttrOfType<BoolAttr>("xdlopsV2"))
+    if (auto attr = (*op).template getAttrOfType<BoolAttr>("xdlopsV2"))
       mop->setAttr("xdlopsV2", attr);
-    if (auto attr = op->getAttrOfType<StringAttr>("perf_config"))
+    if (auto attr = (*op).template getAttrOfType<BoolAttr>("perf_config"))
```
sjw36: Not sure why these lines need to change. And this should still be a StringAttr right?

sjw36 (changes requested):
Still not quite right.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -177,9 +176,9 @@ class ConvConverter : public OpConversionPattern<ConvType> {
                             {padTop, padBottom, padLeft, padRight}));
 
     // Convert optional attributes
-    if (auto attr = op->getAttr("xdlopsV2"))
+    if (auto attr = (*op).template getAttrOfType<BoolAttr>("xdlopsV2"))
       cop->setAttr("xdlopsV2", attr.template cast<BoolAttr>());
-    if (auto attr = op->getAttr("perf_config"))
+    if (auto attr = (*op).template getAttrOfType<BoolAttr>("perf_config"))
       cop->setAttr("perf_config", attr.template cast<StringAttr>());
```
sjw36: This line doesn't need the cast now.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
krzysz00: Yeah, perf_config is a string, not a boolean

jerryyin: I must be drunk when I write this.

Addressed in c31d296

sjw36 (approved):
Looks good!

PR #1090, opened 2023-05-18
[DO NOT SQUASH] Handle gfx941 (including the compiler patches)

krzysz00:
- Pull in the atomics workaround and add gfx941 to it
- Update our internal support lists
- And so on

Note that the compiler patches (aka, _not_ the non-imported patches in this PR) are needed to handle bringup on gfx940.


krzysz00:
Update: I have successfully pulled in the sc0/sc1 patches and their dependenencies. I have tested (by way of `--debug-only`) confirmed that buffer_store operations on gfx940 get their sc0 and sc1 bits set, but that this doesn't happen on any other architectures.

jerryyin (approved):
Looks reasonable. Thanks for the work

PR #1093, opened 2023-05-23
[DO NOT SQUASH] Import MHAL - MLIR Hardware Abstraction Layer

sjw36:
XModel has been extracted out as a separate repo: mlir-hal [https://github.com/ROCmSoftwarePlatform/mlir-hal]

This PR removes XModel and re-aligns with upstream LLVM/MLIR.

Still some upstream changes will be needed to support mhal.launch through bufferization.


sjw36:
Implements https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/890

sjw36:
I think this is going to pass now. Please commence reviews. @krzysz00 @jerryyin @giuseros

cmake/llvm-project.cmake
```diff
@@ -61,6 +61,7 @@ list(APPEND LLVM_INCLUDE_DIRS
 
 # Linker flags
 set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-rpath -Wl,${LLVM_BINARY_DIR}/llvm/lib")
+list(APPEND CMAKE_BUILD_RPATH "${LLVM_BINARY_DIR}/llvm/lib")
```
krzysz00, line 63: Can we double check the LLVM files to see what impact this has on our build?

external/llvm-project/mlir.diff
krzysz00, line 1: Delete pointless file

external/mlir-hal/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+set(MHAL_MAIN_INCLUDE_DIR "${MHAL_PROJECT_DIR}/include")
+option(MHAL_ENABLE_TRANSFORMS "Enable MHAL transforms" ON)
+
+set(LLVM_ENABLE_RTTI OFF CACHE BOOL "")
+
+include_directories("${MHAL_MAIN_INCLUDE_DIR}")
+include_directories("${MHAL_BINARY_DIR}/include")
+
+add_subdirectory(include/mlir)
+add_subdirectory(lib)
```
krzysz00, line 17: Note to future us about making this buildable standalone. Can you set up a ticket for that?

external/mlir-hal/include/mlir/Conversion/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+
+set(LLVM_TARGET_DEFINITIONS MHALPasses.td)
+mlir_tablegen(MHALPasses.h.inc -gen-pass-decls -name MHALConversion)
+#mlir_tablegen(MHALPasses.capi.h.inc -gen-pass-capi-header --prefix MHALConversion)
```
krzysz00, line 4: Let's go ahead and enable a C API?

external/mlir-hal/include/mlir/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+add_subdirectory(Conversion)
```
krzysz00, line 1: I'm going to go ahead and hang "Why not `include/mhal/`?" off of this.

krzysz00 (approved):
I'm not seeing any obvious issues with the refactor, though I still don't quite like the design of `mhal`'s packaging format and would really like to move us to a "one module per target, which includes the relevant host code" system, but that's outside the scope of the PR.

cmake/llvm-project.cmake
sjw36: This setting is the proper way to encode the RPATHs into the binaries for BUILD. I have removed the previous line as it is no longer necessary.

We may need to set CMAKE_INSTALL_RPATH for installation runtime paths.

sjw36:
See ticket for mlir-hal project improvements: https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/983

PR #1095, opened 2023-05-25
[CI] fix build dockerfile script version extraction

manupak:
Current build-dockerfile.sh assumes the rocm version to be 3 digit figure which is not the case always
(e.g. rocm-5.5).

This commit fixes this to extract major and minor
versions looking from the left rather and dropping the last digit.


manupak:
I just found that dockerhub contained a image named "rocm5" because of this issue : https://hub.docker.com/r/rocm/mlir/tags.

jerryyin (approved):
Nice catch, thank you!

PR #1096, opened 2023-05-25
Disable ZSTD which may affect the rocmlir packages

yiqian1:
Disable ZSTD which may affect the rocmlir packages.

### DIFF
--- a/cmake/llvm-project.cmake
+++ b/cmake/llvm-project.cmake
@@ -19,6 +19,7 @@ set(MLIR_CMAKE_CONFIG_DIR
 set(MLIR_TABLEGEN_EXE mlir-tblgen)
 
 # LLVM settings that have an effect on the MLIR dialect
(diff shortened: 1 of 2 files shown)


jerryyin (approved):

cmake/llvm-project.cmake
```diff
@@ -19,6 +19,7 @@ set(MLIR_CMAKE_CONFIG_DIR
 set(MLIR_TABLEGEN_EXE mlir-tblgen)
 
 # LLVM settings that have an effect on the MLIR dialect
+set(LLVM_ENABLE_ZSTD "OFF" CACHE STRING "")
 set(LLVM_ENABLE_ZLIB "OFF" CACHE STRING "")
```
jerryyin, line 23: `OFF CACHE BOOL` instead of STRING?

PR #1097, opened 2023-05-25
[DO NOT MERGE] Rename and enhance threadwise_read_into to threadwise_memcpy

manupak:
This is a draft commit (more like to stand as a proposal).

Problem Description :

On my way to implement blockwise_reduction, I encountered that I m needing to write to LDS from regs. I end up creating transforming_for loops for that. However, I think we can benefit from having a dedicated op for that.

We do have threadwise_read_into but it is specific to reading from global memory into registers.

Proposal :

* Rename threadwise_read_into to threadwise_memcpy
* Add support for reading from regs to LDS
* Also add support for views on both source and dest while not requiring them if they are regs.

Alternative :

* We could add threadwise_(somename that I cant think of) that is specifically for regs to LDS, but I felt it shares more things with read_into (subjective).

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -904,57 +904,102 @@ def Rock_GlobalStoreOp :
   }];
 }
 
-// threadwise_read_into
-def Rock_ThreadwiseReadIntoOp :
-    Rock_Op<"threadwise_read_into">,
+def Rock_ThreadwiseMemCpyOp :
+    Rock_Op<"threadwise_memcpy">,
     AllElementTypesMatch<["source", "dest"]>,
     Arguments<(ins Arg<MemRefOf<SupportedMemoryElems>, "source view", [MemRead]>:$source,
       Arg<MemRefRankOf<SupportedMemoryElems, [1]>,
         "destination registers", [MemWrite]>:$dest,
-      TransformMapArrayAttr:$extraViews,
+      OptionalAttr<TransformMapArrayAttr>:$extraSrcViews,
+      OptionalAttr<TransformMapArrayAttr>:$extraDstViews,
       UnitAttr:$forceUnroll,
       UnitAttr:$useIndexDiffs
     )> {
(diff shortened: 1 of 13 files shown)


manupak:
If you think this is a good direction (otherwise we can create seperate ops): 

On having another look, we could just have one op (threadwise_memcpy) that supports : 
* global to regs
* regs to lds
* lds to regs
* regs to global

krzysz00:
I think we can go either way, though if we're going for the generalized version, we might want to get rid of `threadwise_write_all` as well.

Though I like having separate ops because it makes, for example, finding fusion locations easier instead of trying to filter matches.

manupak:
I think for fusion, what we really need is to find a regs to some memory copy, right? Im not sure whether we'd want to 'filter' it. i.e. if linalg generic is taking input that is a memref that is being copied with data, one could fuse it to the copy.

Even if we want seperate ops, I d think we only want 2 flavors: 1) regs to lds/global And 2) global/lds to regs. Then we ll look for 1). Having said that, I still think we d generally don't want such a separation because, fusing into 2) represents fusing into input reads -- though not supported could be a valid scenario in future.

WDYT?

krzysz00:
The one thing that makes me hesitant about merging moves to/from global and to/from LDS is that going to LDS doesn't involve bounds checks but going to global might.

But we could probably make it work as one general operation for movement loops.

On the third hand, I like the nice look of fusion looping through write_all's and generating the corresponding read_into's. But that's not too big a deal.

sjw36:
I was initially on-board with unifying write_all/read_into but having second thoughts. The fusion flow looks for the write_all after the conv/gemm pattern, this can be non-trivial and probably best to keep separate.

Also having separate ops more aligns with ld/st ops and as @krzysz00 points out more readable but they could certainly be derived from a base class to share common functionality.

manupak:
Cool! 

ok, I ll enhance write_all/read_into to support LDS as well then and keep the seperation.

manupak:
superseded by : https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1104

PR #1098, opened 2023-06-01
[DO NOT SQUASH] Add the amdgpu-waves-per-eu attribute to the kernel

giuseros:
Adding this attribute seem to improve things (~10% on a TT configuration and ~20% on a NN configuration) for the workload I am focusing. 

It is basically saying to the backend compiler to optimize the code so that at least `x` waves need to be executing at the same time on the Execution Units. 

Since I didn't test it yet on all the workloads, I will put this as a draft (also because I hardcoded a "2" , but probably we want to have something more elegant) so that @manupak can test this on MIGraphX. 

Thanks @sjw36 for the help!

### DIFF
--- a/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
+++ b/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
@@ -200,6 +200,10 @@ struct LowerGpuOpsToROCDLOpsPass
         op->setAttr(ROCDL::ROCDLDialect::getFlatWorkGroupSizeAttrName(),
                     flatSizeAttr);
       }
+      if (auto minWavesPerEu = op->removeAttr("min_wave_per_eu")
+                                   .dyn_cast_or_null<IntegerAttr>()) {
+        op->setAttr(ROCDL::ROCDLDialect::getMinWavesPerEu(), minWavesPerEu);
+      }
     });
 
     OpPassManager pm("gpu.module");
--- a/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
+++ b/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
@@ -200,10 +200,6 @@ struct LowerGpuOpsToROCDLOpsPass
         op->setAttr(ROCDL::ROCDLDialect::getFlatWorkGroupSizeAttrName(),
                     flatSizeAttr);
       }
-      if (auto minWavesPerEu = op->removeAttr("min_wave_per_eu")
-                                   .dyn_cast_or_null<IntegerAttr>()) {
-        op->setAttr(ROCDL::ROCDLDialect::getMinWavesPerEu(), minWavesPerEu);
-      }
     });
 
     OpPassManager pm("gpu.module");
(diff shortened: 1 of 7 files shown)


giuseros:
I added a better heuristic to decide when to ask the compiler to generate code that fits at least two waves per EU and also gave a go at performance evaluation. All in all, I see between 5% and 12% improvements on specific single batch size cases (fp16 and int8) so I think this is worth doing it. 

There is a spike of -23% which is a glitch. I retuned and actually saw slightly better performance

[regression.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/11651359/regression.zip)

external/llvm-project/mlir/lib/Target/LLVMIR/Dialect/ROCDL/ROCDLToLLVMIRTranslation.cpp
```diff
@@ -135,6 +135,21 @@ class ROCDLDialectLLVMIRTranslationInterface
       llvmAttrValue.append(value.getValue());
       llvmFunc->addFnAttr("amdgpu-flat-work-group-size", llvmAttrValue);
     }
+    if ("rocdl.amdgpu_waves_per_eu" == attribute.getName()) {
```
krzysz00: Can we name this `rocdl.waves_per_eu`?

external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
```diff
@@ -200,6 +200,10 @@ struct LowerGpuOpsToROCDLOpsPass
         op->setAttr(ROCDL::ROCDLDialect::getFlatWorkGroupSizeAttrName(),
                     flatSizeAttr);
       }
+      if (auto minWavesPerEu = op->removeAttr("min_wave_per_eu")
```
krzysz00: I don't think, in this case, that we need this, and we should instead set `rocdl.waves_per_eu` directly

krzysz00 (commented):
I've got minor nitpicks design-wise, but no fundamental objections to the idea.

krzysz00 (approved):
Approved, though I think you'll want to `git rebase -i` this into two comits

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
```diff
@@ -169,6 +169,13 @@ void LowerRockOpsToGPUPass::runOnOperation() {
                        b.getDenseI32ArrayAttr({gridSize, 1, 1}));
     }
 
+    if (auto attr = theFunc->getAttr("wave_size")) {
+      int32_t waveSize = attr.template cast<IntegerAttr>().getInt();
+      if (blockSize / waveSize >= 2) {
+        gpuFunc->setAttr("rocdl.waves_per_eu", b.getI32IntegerAttr(2));
+      }
```
jerryyin, line 176: No objection but just a curiosity question. Have we unleashed all its potential?

Any idea adding a `if (blockSize / waveSize >= 4)` branch and set it as 4 before this condition would impact the performance?

giuseros: Yes, I tried with multiple combinations of: 
- 2,4
- 2,2
- 4,4
But anything different than 2 (when blockSize/waveSize >=2 ) impacts performance negatively

jerryyin: Thanks for clarifying!

PR #1087, opened 2023-05-16
Add a rocmlir-gen option --out_datatype

yiqian1:
The rocmlir-gen option --out_datatype specifies the datatype of output tensors. Currently only int8 is allowed for int8 conv and gemm. 

This feature is added to get compatible performances with CK as stated in [issue#855](https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/855).

PR #1099, opened 2023-06-01
Generalize generation infrastructure to account for mixed types (FP8)

krzysz00:
1. In the ConvGenerator, expand on the work from #1087 to have proper filter, input, and output data type strings.
2. Amend rocmlir-gen to take --fil_dtype/--a_dtype/-tf/-ta, --in_dtype/-b_dtype/-ti/-tb, and --out_dtype/--c_dtype/-to/-tc to allow for specifying any combination of types for a convolution or gemm.
2b. Make the `-t` option an alias for quickly setting `-t{a,b,c}`, including the 8-bit {int,float} input  -> 32-bit {int,float} output default.
3. While doing the above, extend the various strToType() methods to accept 8-bit float types.
4. Make the GenParams struct in rocmilr-gen take a vector of types to account for our new generality.
5. Instead of hard-coding f16 and bf16 as the small floats that use f32 for validation, send all floats of size < 32 to f32. That way we can't miss a case.

### DIFF
--- a/mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
+++ b/mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
@@ -37,12 +37,12 @@ Conv2dGenerator::Conv2dGenerator(
     const std::string &arch, const std::string &chip, const std::string &triple,
     const std::string &chipFeatures, const std::string &perfConfig, int num_cu,
     GemmFeatures features, const std::optional<ConvOpType> operation,
-    const std::string &dataTypeStr, int dilationHeight, int dilationWidth,
+    const std::string &filterDataTypeStr, const std::string &inputDataTypeStr,
+    const std::string &outputDataTypeStr, int dilationHeight, int dilationWidth,
     int strideHeight, int strideWidth, int paddingHeightLeft,
     int paddingHeightRight, int paddingWidthLeft, int paddingWidthRight,
     const std::string &filterLayout, const std::string &inputLayout,
-    const std::string &outputLayout, const std::string &outputDataTypeStr,
-    const std::string &kernelBaseName)
+    const std::string &outputLayout, const std::string &kernelBaseName)
     : config{arch,
              chip,
              triple,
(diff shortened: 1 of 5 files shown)


krzysz00:
Note that, after this PR lands, we'll be able to run an fp8 kernel on MI-300, I think.
With #1079 and #1088, we'll be close to being able to do a GPU validation (probably).
We may end up needing to write CPU utility functions analogous to the ones used for f16 in order to get everything to codegen.

yiqian1:
LGTM

krzysz00:
@yiqian1 I think you forgot to hit approve?

yiqian1 (approved):

mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
```diff
@@ -686,8 +703,9 @@ void Conv2dGenerator::setKernelName(const std::string &newName) {
   config.kernelBaseName = newName;
 }
 
-void Conv2dGenerator::setDataType(std::string newType) {
-  config.dataTypeStr = newType;
+void Conv2dGenerator::setDataTypes(const std::string &dataTypeStr) {
+  config.filterDataTypeStr = config.inputDataTypeStr =
+      config.outputDataTypeStr = dataTypeStr;
```
jerryyin, line 708: This comment doesn't hold off merging.

I think you are just extending the original logic here. But this will tentatively set the data type to (potentially) wrong types, and we leave those field as-is. Then later, when the "real" data type is required, we do a second computation regardless of what the field value is using `getFilterDataType()`, `getInputDataType()` and `getOutputDataType()`.

I would much more prefer instead
 - setDataTypes() does everything, and populate the correct types for filter/intput/output member fields at one go
 - get*Type() only return the field and doesn't do any redundant derive process.

krzysz00: This API is literally for `setDataTypes("f32");` for verification code.

PR #1100, opened 2023-06-02
Add Wmma tuning infrastructure

giuseros:
This PR is adding the tuning infrastructure for WMMA:
- I added a `--data-type` argument to `perfRunner` and `tuningRunner`. This is because for wmma I want to focus on `i8` and `f16` performance, and not on `f32` (which takes a very long time to tune, since it is not using wmma)
- For now the tuning space is similar to Xdlops, but I think it can be pruned (Kpack should always be 16 and it looks that also Kouter always comes out as 16).

This PR depends on  https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1084

### DIFF
--- a/mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
+++ b/mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
@@ -45,6 +45,16 @@ void createGemmTuningRangeBF(struct TunableParams *newSpace,
                                                  {1, 4, 8, 16},
                                                  {0, 1}};
 
+  // M/block N/block K/block M/wave N/wave kPack aCopyMore/forceUnroll
+  const std::vector<std::vector<uint32_t>> ValidRangeWmmaGemmParams = {
+      {4, 8, 16, 32, 64, 128, 256},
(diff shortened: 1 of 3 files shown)


mlir/include/mlir/Dialect/Rock/utility/AmdArchDb.h
```diff
@@ -21,6 +21,11 @@ struct AmdArchInfo {
 
   constexpr AmdArchInfo(GemmFeatures defaultFeatures, int64_t waveSize)
       : defaultFeatures(defaultFeatures), waveSize(waveSize) {}
+
+  /// Get the default features for the pari <arch, datatype>. The data type
+  /// can be a string or an mlir::Type
+  GemmFeatures getDefaultFeatures(Type dataType);
+  GemmFeatures getDefaultFeatures(StringRef strDataType);
```
krzysz00: Do we need this? We could convert the string to a `Type` out in rocmlir-gen?

krzysz00 (approved):
As far as the code goes, approved!

I'll leave experiments wrt tuning space to you

PR #1102, opened 2023-06-03
Tune int8xint8->int8 gemm configurations

yiqian1:
The tuningRunner.py script generates gemm configurations with the int8 outputs for tuning.
The perfRunner.py script adds an "out_datatype" attribute to GemmConfiguration. 

ConvConfiguration does not change as we only tune convs with default output datatypes. 

Also clean up Jenkinsfile to use one fusion db for all fusion test cases.

### DIFF
--- a/mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
+++ b/mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
@@ -338,6 +338,42 @@ std::string getTuningProblemStr(ModuleOp &mod) {
   } else if (opType == KernelType::Gemm) { // gemm case
     rock::GemmOp rGemmOp = dyn_cast<rock::GemmOp>(gemmOp);
     // Please keep these in sync with mlir/utils/performance/perfRunner.py
+    // Data type
+    problemOS << "-t ";
+    Type elemTypeA = gemmIF.getAType(), elemTypeB = gemmIF.getBType();
+    if (elemTypeA.isF32() && elemTypeB.isF32()) {
+      problemOS << "f32";
+    } else if (elemTypeA.isF16() && elemTypeB.isF16()) {
+      problemOS << "f16";
+    } else if (elemTypeA.isBF16() && elemTypeB.isBF16()) {
+      problemOS << "bf16";
+    } else if (elemTypeA.isInteger(8) && elemTypeB.isInteger(8)) {
+      problemOS << "i8";
+    } else if (elemTypeA.isFloat8E4M3FNUZ() && elemTypeB.isFloat8E4M3FNUZ()) {
+      problemOS << "fp8_fp8";
+    } else if (elemTypeA.isFloat8E4M3FNUZ() && elemTypeB.isFloat8E5M2FNUZ()) {
+      problemOS << "fp8_bf8";
+    } else if (elemTypeA.isFloat8E5M2FNUZ() && elemTypeB.isFloat8E4M3FNUZ()) {
+      problemOS << "bf8_fp8";
+    } else if (elemTypeA.isFloat8E5M2FNUZ() && elemTypeB.isFloat8E5M2FNUZ()) {
+      problemOS << "bf8_bf8";
+    } else {
+      // Unknown data type
+      return std::string();
+    }
+
+    // OUtput datatype
+    auto outType = gemmIF.getOutArgument()->get().getType();
+    auto elemTypeC = outType.dyn_cast<mlir::MemRefType>().getElementType();
+    problemOS << " -out_datatype ";
+    if (elemTypeC.isFloat8E4M3FNUZ()) {
+      problemOS << "fp8" << sep;
+    } else if (elemTypeC.isFloat8E5M2FNUZ()) {
+      problemOS << "bf8" << sep;
+    } else {
+      problemOS << elemTypeC << sep;
+    }
+
     // TransA
     problemOS << "-transA ";
     if (rGemmOp.getATransposed())
@@ -359,30 +395,6 @@ std::string getTuningProblemStr(ModuleOp &mod) {
     problemOS << "-k " << gemmIF.getGemmSize().k << sep;
   } else {
     // Unknown op type, unreachable.
-    return std::string();
-  }
-
-  // Data type
-  problemOS << "-t ";
-  Type elemTypeA = gemmIF.getAType(), elemTypeB = gemmIF.getBType();
-  if (elemTypeA.isF32() && elemTypeB.isF32()) {
-    problemOS << "f32";
-  } else if (elemTypeA.isF16() && elemTypeB.isF16()) {
-    problemOS << "f16";
-  } else if (elemTypeA.isBF16() && elemTypeB.isBF16()) {
-    problemOS << "bf16";
-  } else if (elemTypeA.isInteger(8) && elemTypeB.isInteger(8)) {
-    problemOS << "i8";
-  } else if (elemTypeA.isFloat8E4M3FNUZ() && elemTypeB.isFloat8E4M3FNUZ()) {
-    problemOS << "fp8_fp8";
-  } else if (elemTypeA.isFloat8E4M3FNUZ() && elemTypeB.isFloat8E5M2FNUZ()) {
-    problemOS << "fp8_bf8";
-  } else if (elemTypeA.isFloat8E5M2FNUZ() && elemTypeB.isFloat8E4M3FNUZ()) {
-    problemOS << "bf8_fp8";
-  } else if (elemTypeA.isFloat8E5M2FNUZ() && elemTypeB.isFloat8E5M2FNUZ()) {
-    problemOS << "bf8_bf8";
-  } else {
-    // Unknown data type
     return std::string();
   }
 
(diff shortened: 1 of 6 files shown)


mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -306,6 +306,35 @@ std::string getTuningProblemStr(ModuleOp &mod) {
...
+      problemOS << "bf8_bf8";
+    } else {
+      // Unknown data type
+      return std::string();
+    }
+
+    // OUtput datatype
+    auto outType = gemmIF.getOutArgument()->get().getType();
+    auto memRefType = outType.dyn_cast<mlir::MemRefType>();
+    problemOS << " -out_datatype " << memRefType.getElementType() << sep;
```
krzysz00: This'll give us `f8E5M2FNUZ` and the like if you call for an fp8 output (not currently supported but still)

mlir/utils/performance/perfRunner.py
```diff
@@ -489,8 +489,17 @@ def benchmarkExternal(cls, commandLine, paths: Paths, arch, envs=dict()):
             outs, errs = p1.communicate()
         return config.tableEntry(nanoSeconds)
 
-def getGemmConfigurations(fileName):
+OUT_DATATYPE = re.compile(r"-out_datatype[=\s*](.*)")
+def getGemmConfigurations(fileName, rocmlir_gen_flags):
```
krzysz00: I'm not sure I like parsing out of rocmlir-gen flags here. If someone's trying to override the output data type, than they should use that.

mlir/utils/performance/perfRunner.py
```diff
@@ -1214,6 +1224,9 @@ def main(args=None):
         configs = getConvConfigurations(paths.configuration_file_path)
     elif opType == Operation.GEMM:
         configs = getGemmConfigurations(paths.configuration_file_path)
+        # Make MLIR to be compatible with CK by forcing int8xint8->int8
```
krzysz00: Will this do the right thing? Is `CKGemmConfig` the class that's being used to run the rocMLIR comparisons?

mlir/utils/performance/tuningRunner.py
```diff
@@ -181,6 +181,18 @@ def extractFusionConfigs(test_dir, paths: Paths, options: Options):
 
     return opType
 
+def getExtendedGemmConfigurations(filename):
```
krzysz00: I don't think I like the idea of implicitly adding more configs to tune

Would it make more sense to explicitly add the `i8 * i8 -> i8` configs to our files of things we want to tune?

krzysz00 (commented):
I've got some questions, but I don't see anything massively wrong with the PR.

Sorry about the long delay in reviews.

mlir/utils/performance/perfRunner.py
yiqian1: CKGemmConfig is used if perfRunner.py is invoked with '--external-gemm-library CK'

krzysz00: ... right, these ones have `benchmarkExternal` and are a `GemmConfig` otherwise. Makes sense, concern withdrawn.

mlir/utils/performance/tuningRunner.py
yiqian1: I've thought about that method. The overhead is that we'd have to maintain another config files that doesn't include i8*i8->i8 for running non-CK comparison, instead of having one config files for both tuningRunner and perfRunner for all externals.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -306,6 +306,42 @@ std::string getTuningProblemStr(ModuleOp &mod) {
   } else if (opType == KernelType::Gemm) { // gemm case
     rock::GemmOp rGemmOp = dyn_cast<rock::GemmOp>(gemmOp);
     // Please keep these in sync with mlir/utils/performance/perfRunner.py
+    // Data type
+    problemOS << "-t ";
+    Type elemTypeA = gemmIF.getAType(), elemTypeB = gemmIF.getBType();
```
giuseros, line 343: If I understand correctly the data type won't be added in the case of convolution?

mlir/utils/performance/tuningRunner.py
giuseros: THe problem here is that we are extending our tuning time by 30%. 

The way I see it is that you are adding another implicit datatype. In other words we have: f32, f16, i8i8, i8i32. Can you parametrize which datatype we want to tune? In my recently merged PR (https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1100) I have introduced a `--datatype` parameter to the tuning runner. Maybe you can remove `i8` and add `i8i8` and `i8i32` . The deafult might be all of them (or maybe just `i8i8`)

mlir/utils/performance/perfRunner.py
```diff
@@ -1214,6 +1224,9 @@ def main(args=None):
...
+        if confClass == CKGemmConfig:
```
giuseros: I still think we are adding a further datatype in `perfRunner` (see my previous comment). Previously we had f32, f16, i8 and now we have f32, f16,i8i8, i8i32. Can we by default generate all the configuration, but give the user the possibility to pick which datatype they want to use?

giuseros:
Hi @yiqian1 , sorry for the late review. I think the code looks fine, I only have comments/questions. Also, a general question (out of curiosity): did you run the tuner for i8i8 and i8i32? Is there any change in the configurations picked?

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
yiqian1: The convolution config is using MIOpen format, e.g. conv for f32, convfp16 for f16 etc. Now we decided to remove MIOpen integration from our CI, I think I will change it to use rocMLIR format (i.e. what rocmlir-gen can accept).

mlir/utils/performance/tuningRunner.py
yiqian1: Yes, because that's what we need.

yiqian1:
> Hi @yiqian1 , sorry for the late review. I think the code looks fine, I only have comments/questions. Also, a general question (out of curiosity): did you run the tuner for i8i8 and i8i32? Is there any change in the configurations picked?

Yes. I run the tuner for both i8i8 and i8i32. There are a few changes in perf-configs. If you look at the MLIR_vs_CK performance report, there are 3 test cases where Tuned/CK is worse than MLIR/CK. That's because Tuned MLIR picked wrong perf-config. This PR should fix that.

yiqian1:
I don't want to have extra config files. I adopted Giuseppe's suggestion and used --data-type to explicitly specify the input and output datatypes for i8xi8->i8 cases. Thanks to @giuseros  and @krzysz00

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
krzysz00: ... this is a breaking change for anyone holding a tuning DB, I think. Not a huge deal, but it's worth pointing out.

mlir/utils/performance/tuningRunner.py
```diff
@@ -263,7 +263,7 @@ def main(args=None):
     parser.add_argument(
         '--data-type',
          nargs='+',
-         choices=["f32", "f16", "i8"],
+         choices=["f32", "f16", "i8", "f32_f32", "f16_f16", "i8_i32", "i8_i8"],
```
krzysz00: I don't think we need the 'f32_f32` ones.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -599,11 +599,15 @@ pipeline {
                         steps {
                             buildProject('check-rocmlir-build-only ci-performance-scripts', '')
                             dir('build') {
-                                // Tune gemms, fail if the DB is not created
+                                // Tune gemms with default datatypes, fail if the DB is not created
                                 sh """python3 ./bin/tuningRunner.py --operation gemm \
                                         --configs_file=../mlir/utils/performance/gemm-configs \
                                         --output=mlir_tuning_${ARCH}.tsv
                                       [ -f mlir_tuning_${ARCH}.tsv ]"""
+                                // Tune gemms int8xint8->int8
```
krzysz00: Let's go ahead and update this with "for performance comparisons against CK. Ideally, this'll be removed in the future."

krzysz00 (commented):
Overall, (but also @giuseros for comments) I think the approach here is fine.

mlir/utils/performance/tuningRunner.py
giuseros: Yes, I would say that we only need "i8_i8" and "i8_i32". We can assume that the others mean same input/output type.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -338,6 +338,42 @@ std::string getTuningProblemStr(ModuleOp &mod) {
...
+    if (elemTypeA.isF32() && elemTypeB.isF32()) {
```
giuseros, line 344: same suggestion as above: can we only add i8i8 and i8i32?

yiqian1: We don't handle i8i8 or i8i32 here. This is the rocmlir-gen argument for filter and input data types.

giuseros (approved):
LGTM

PR #1104, opened 2023-06-05
Add reading and writing to LDS via threadwise_*

manupak:
This commit adds writing to LDS using the
threadwise_read_into and threadwise_write_all ops.

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/894

### DIFF
--- a/mlir/lib/Dialect/Rock/IR/RockDialect.cpp
+++ b/mlir/lib/Dialect/Rock/IR/RockDialect.cpp
@@ -1444,11 +1444,12 @@ LogicalResult ThreadwiseReadIntoOp::verify() {
     inputShape = extraViews[0].cast<TransformMapAttr>().getUpperBounds();
 
   MemRefType srcType = getSource().getType().cast<MemRefType>();
-  Attribute srcMemSpaceAttr = srcType.getMemorySpace();
+  gpu::AddressSpaceAttr srcMemSpaceAttr =
+      srcType.getMemorySpace().dyn_cast_or_null<gpu::AddressSpaceAttr>();
   // Unless specified it is assumed to be global
   gpu::AddressSpace srcGpuMemSpace = gpu::AddressSpace::Global;
   if (srcMemSpaceAttr) {
-    srcGpuMemSpace = srcMemSpaceAttr.cast<gpu::AddressSpaceAttr>().getValue();
+    srcGpuMemSpace = srcMemSpaceAttr.getValue();
   }
 
   if (srcGpuMemSpace == gpu::AddressSpace::Global) {
--- a/mlir/lib/Dialect/Rock/IR/RockDialect.cpp
+++ b/mlir/lib/Dialect/Rock/IR/RockDialect.cpp
@@ -1435,7 +1435,7 @@ LogicalResult ThreadwiseReadIntoOp::verify() {
   auto gpuMemSpaceAttr = memSpaceAttr.dyn_cast_or_null<gpu::AddressSpaceAttr>();
   if (memSpaceAttr && (!gpuMemSpaceAttr || gpuMemSpaceAttr.getValue() !=
                                                gpu::AddressSpace::Private))
-    return emitOpError("source must be private registers");
+    return emitOpError("dest must be private registers");
   ArrayAttr extraViews = getExtraViews();
   ArrayRef<int64_t> inputShape;
   if (extraViews.empty())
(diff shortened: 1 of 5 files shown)


mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -1435,16 +1435,35 @@ LogicalResult ThreadwiseReadIntoOp::verify() {
...
     inputShape = extraViews[0].cast<TransformMapAttr>().getUpperBounds();
 
-  if (inputShape.size() != 3)
-    return emitOpError("source view must accept (bid, tid, iter) coordinates");
+  MemRefType srcType = getSource().getType().cast<MemRefType>();
+  Attribute srcMemSpaceAttr = srcType.getMemorySpace();
+  // Unless specified it is assumed to be global
+  gpu::AddressSpace srcGpuMemSpace = gpu::AddressSpace::Global;
+  if (srcMemSpaceAttr) {
+    srcGpuMemSpace = srcMemSpaceAttr.cast<gpu::AddressSpaceAttr>().getValue();
```
krzysz00: `auto srcMemSpaceAttr = srcType.getMemorySpace.dyn_cast_or_null<gpu::AddressSpaceAttr>();`
and then the if statement. Or possibly erroring out on memory spaces that aren't either absent or the GPU ones.
That cast<> will cause crashes on user input as opposed to programmer error.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -738,21 +793,46 @@ LogicalResult ThreadwiseWriteAllRewritePattern::matchAndRewrite(
...
+      b.create<GlobalStoreOp>(
+          loc, source, buffer, b.getIndexAttr(vectorLen), op.getFeaturesAttr(),
+          op.getStoreMethodAttr(),
+          outLoop.getLowerCoords(/*domain=*/0)[iterDim[dstAddrSpace]],
+          outLoop.getValidity(/*domain=*/1),
+          outLoop.getLowerCoords(/*domain=*/1));
+    } else {
+      Type loadType = vectorTypeOrSelf(elementType, vectorLen);
+      TypedValue<IntegerType> valid = outLoop.getValidity(/*domain=*/0);
+      Value zeroInt = createZeroConstantOp(b, loc, valid.getType());
```
krzysz00: This `cmp` is reduntant, - `validity` is already bool-typed

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -686,21 +699,52 @@ LogicalResult ThreadwiseReadIntoRewritePattern::matchAndRewrite(
...
+      Value isValid = b.createOrFold<arith::CmpIOp>(
+          loc, arith::CmpIPredicate::ne, valid, zeroInt);
+      scf::IfOp ifb =
+          b.create<scf::IfOp>(loc, isValid, /*withElseRegion=*/true);
+      {
+        OpBuilder thenb = ifb.getThenBodyBuilder(b.getListener());
+        Value loaded = thenb.create<InBoundsLoadOp>(
+            loc, loadType, buffer, loadLoop.getLowerCoords(/*domain=*/0));
+        thenb.create<InBoundsStoreOp>(
```
krzysz00: Why not instead have the load be in the if, creating an if/else that yield's the value, and then unconditionally create the InBoundsStore?

krzysz00 (commented):
Style nits more than fundamental objections

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
manupak: good call! done.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
manupak: right! fixed.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
manupak: you are right. fixed now!

manupak:
Thanks for the review @krzysz00! I believe they are addressed now.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -664,13 +665,25 @@ LogicalResult ThreadwiseReadIntoRewritePattern::matchAndRewrite(
...
+  MemRefType srcBufferType = buffer.getType().cast<MemRefType>();
+  ArrayRef<int64_t> bufferShape = srcBufferType.getShape();
+
+  // Unless specified it is assumed to be global
+  gpu::AddressSpace srcAddrSpace = gpu::AddressSpace::Global;
+  if (srcBufferType.getMemorySpace()) {
+    srcAddrSpace =
+        srcBufferType.getMemorySpace().cast<gpu::AddressSpaceAttr>().getValue();
+  }
+  DenseMap<gpu::AddressSpace, size_t> iterDim{
```
krzysz00, line 677: In silly minor notes, I think there's a `SmallDenseMap` we could use here? And I wonder if we can `constexpr` this

krzysz00 (approved):
Looks good to me!

PR #1108, opened 2023-06-08
Add blockwise_broadcast_reduce op and lowering

manupak:
This commit adds blockwise_broadcast_reduce op that
uses a LDS workspace to reduce. It accepts
the inputs in regs and produces a broadcasted
reduced output to regs. The workspace does not
need to be initialized.

* adds nr space larger than blocksize cases
* made lds workspace buffer flat to allow reuse
* added more tests to check max

closes : https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/882


manupak:
~~depends on https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1104 (and contains content of both PRs)~~

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1145,6 +1145,45 @@ def Rock_AccelGemmOp:
...
+def Rock_BlockwiseReduceOp:
+    Rock_Op<"blockwise_reduce">,
+    Arguments<(ins MemRefOf<GemmOutputTypes>:$input,
+                   MemRefOf<LdsBufferTypes>:$workspace_buffer,
+                   MemRefOf<GemmOutputTypes>:$output,
+                   IndexAttr:$axis,
+                   ReduceMethodAttr:$reduceMethod,
+                   TransformMapArrayAttr:$inputRegView,
+                   I32Attr:$blockSize,
+                   I32Attr:$gridSize
```
krzysz00: Why do we need the grid size here?

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -1673,6 +1673,91 @@ LogicalResult ReduceOp::verify() {
...
+  if (!wsMemSpaceAttr) {
+    return emitError("No gpu memspace attr found in workspace memref; the "
+                     "workspace memref should be in LDS");
+  } else {
+    if (wsMemSpaceAttr.getValue() != gpu::AddressSpace::Workgroup) {
+      return emitError("workspace should be in LDS.");
+    }
+  }
+
+  if (inputTensorShape[0] != gridSize) {
```
krzysz00: Ah, now I get it - coordinate transforms don't have symbols so we can't slice off the workgroup ID component

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -830,17 +830,617 @@ LogicalResult ThreadwiseWriteAllRewritePattern::matchAndRewrite(
...
+  }
+
+  // This function will append views to target a flat LDS buffer
+  // where non-reduction dims are laid contigously as they are expected
+  // function on parallel.
+  ArrayAttr createLDSWorkspaceView(Location loc, PatternRewriter &rewriter,
+                                   ArrayAttr regTensorView, int64_t reduceAxis,
+                                   bool makeRDimZero = false) const {
+
+    TransformMapAttr lowestTr =
```
krzysz00, line 928: Could you instead `getShape()` the underlying object instead of taking the output shape of the last transform? This feels unnecessary.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -830,17 +830,617 @@ LogicalResult ThreadwiseWriteAllRewritePattern::matchAndRewrite(
...
+        "flatDim", 0, {"nrDim", "rDim"},
+        {nonReduceMergeDimSize, lowestShape[reduceAxis]});
+    TransformMapAttr flatLDSView = flatLDSViewBuilder.get();
+    SmallVector<Attribute> threadsToLDSViewAttrs;
+    for (Attribute trMap : regTensorView) {
+      threadsToLDSViewAttrs.push_back(trMap);
+    }
+    threadsToLDSViewAttrs.push_back(twoDimLDSView);
+    threadsToLDSViewAttrs.push_back(flatLDSView);
+    return rewriter.getArrayAttr(threadsToLDSViewAttrs);
```
krzysz00, line 969: Instead of doing this, could you add some `rock.transform` operations and then `untransform()` later? Or would that make things uglier?

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -830,17 +830,617 @@ LogicalResult ThreadwiseWriteAllRewritePattern::matchAndRewrite(
...
+  ArrayAttr createThreadViewForNRLargerThanThreads(
+      Location loc, ArrayRef<int64_t> toReduceShape, int64_t blockSize,
+      int64_t reduceAxis, PatternRewriter &rewriter) const {
+    BottomUpTMBuilder threadsToTensor(rewriter, toReduceShape, loc);
+    SmallVector<StringRef, 4> lowerNameRefs;
+    threadsToTensor.getStartNames(lowerNameRefs);
+
+    int64_t nonReduceMergeDimSize = 1;
+    SmallVector<StringRef, 4> nonReduceNameRefs;
+    for (auto dimAndSize : llvm::enumerate(toReduceShape)) {
```
krzysz00, line 990: It feels like this could be abstracted out into a utility function

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -830,17 +830,617 @@ LogicalResult ThreadwiseWriteAllRewritePattern::matchAndRewrite(
...
+    threadsToTensor = BottomUpTMBuilder::above(threadsToTensor, padTrMap);
+    threadsToTensor.unmerge({"tid", "nrIter"}, {0, 1}, "nrDim",
+                            {blockSize, nrThreads});
+    threadsToTensor.passThrough({"rIter"}, {2}, {"rIter"});
+    TransformMapAttr unmergeTrMap = threadsToTensor.get();
+
+    return rewriter.getArrayAttr({unmergeTrMap, padTrMap, mergeTrMap});
+  }
+
+  // This should only be used if product non-reduction dims is
```
krzysz00, line 1018: Can this and the function above be combined in some reasonable way?

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -830,17 +830,617 @@ LogicalResult ThreadwiseWriteAllRewritePattern::matchAndRewrite(
...
+    threadsToTensor = BottomUpTMBuilder::above(threadsToTensor, padTrMap);
+    threadsToTensor.unmerge({"rtid", "rIter"}, {1, 2}, "rDim",
+                            {rthreads, rDimPerRThread});
+    threadsToTensor.passThrough({"nrtid"}, {0}, {"nrDim"});
+    TransformMapAttr unmergeTrMap = threadsToTensor.get();
+
+    return rewriter.getArrayAttr({unmergeTrMap, padTrMap, mergeTrMap});
+  }
+
+  ArrayAttr dropDim(Location loc, ArrayAttr view, size_t dropDimIdx,
```
krzysz00: What's this for?

krzysz00 (commented):
I haven't read this thing in its full detail, but it seems like a lot of this could be cleaned up by using `rock::TransformOp` to build up the view chains you need and then `untransform()`-ing after. I might be wrong though.

Another note: it might be not be a bad idea to create `*TmBuilder::below/above(TransformMapAttr transform)` that uses the notional `TransformMapAttr::collectNames()` (which we could add) to make names consistent and perhaps make debugging easier, but that's a separate PR.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: mostly to verify as you may have found later... 

So this is a bit of conundrum I had as well; we dont have a way to easily remove a dimension from the view.
Between the two options : 
D1) Make the view upperdims just be (tid, iter)
D2) Make the view upperdims just be (bid, tid, iter)

So D1) doesn't need gridsize nor the slicing of bid dimension that happens later, however, most higher level ops at this time will have a view that have the upperdims be (bid, tid, iter). Therefore, that slicing has to be done by the client op that uses blockwise reduction.

So D2) will need gridsize mostly as a verification to make sure the correct views are being passed on, however,  the implementation (current) will strip it off after verifying it. This way it is bit convenient for client ops but its a bit icky that this op needs a view that has (bid, ... ).

I went for D2 due to convenience of client ops.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
manupak: I dont have such an object/op -- just a bunch of manipulated transforms (basically bid stripped)

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
manupak: I dont have a preference either way in terms of clarity -- I chose this because I felt it was unnecessary to re-traverse the IR (i.e. do untransform) within the same lowering pass. I really dont have an opinion between passing the ArrayAttr of transforms vs a stack of transform ops apart from the latter has to spend marginally more cycles compiling.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
manupak: It does two different things even from the first transform?

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
manupak: This to drop the bid dim from the view. Please see the reply for having gridsize here..

manupak:
Ok, Ive done the changes to be: 
D1) Make the view upperdims just be (tid, iter) and renamed to "blockwise_broadcast_reduce" as per our offline discussion.

@krzysz00 please take a look -- I can still create transform ops for ArrayAttr and do untransform later, but let me know.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1145,6 +1145,46 @@ def Rock_AccelGemmOp:
...
+                   MemRefOf<GemmOutputTypes>:$output,
+                   IndexAttr:$axis,
+                   ReduceMethodAttr:$reduceMethod,
+                   TransformMapArrayAttr:$inputRegView,
+                   I32Attr:$blockSize
+                   )> {
+  let summary = "Blockwise reduce operation with output broadcasting";
+  let description = [{
+    The `rock.blockwise_broadcast_reduce` op does reduce at a blockwise level and
+    broadcast the reduce values to a size of input tensor. Note that both input/output
```
krzysz00: "to fill all reduced elements in the input tensor"

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1145,6 +1145,46 @@ def Rock_AccelGemmOp:
...
+    tensors live in registers.
```
krzysz00, line 1215: Can we confirm (by staring at ISA etc.) that redundant loads from LDS get eliminated?

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1145,6 +1145,46 @@ def Rock_AccelGemmOp:
...
+    axis (Dr) is not sliced because it is not supported.
+
+    $output tensor resides in registers in threads. It would have
+    the same dimensionality of the input tensor but broadcasted.
+    So the users of the op, has to add a extraView if they are writing back
+    to some memory (LDS/Global) to bar threads (or iteration within threads)
+    to not to create a broadcasted tensor.
+
+    $workspace_buffer resides in LDS. This should be a flat tensor with at least
+    with number of items per block.
```
krzysz00: with at least `blockSize` elements.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -1673,6 +1673,84 @@ LogicalResult ReduceOp::verify() {
   return success();
 }
 
+//===-----------------------------------------------------===//
+// Blockwise_ReduceOp
+//===-----------------------------------------------------===//
+
+LogicalResult BlockwiseBroadcastReduceOp::verify() {
+  ArrayAttr inputViewArrayAttr = getInputRegViewAttr();
+  // This view should be {bid, tid, iter} to {bid, d0, ... , Dr , ... , dn};
```
krzysz00: Update comments

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -830,17 +830,558 @@ LogicalResult ThreadwiseWriteAllRewritePattern::matchAndRewrite(
...
+        nonReduceNameRefs.push_back(lowerNameRefs[dim]);
+      }
+    }
+    threadsToTensor.merge("nrDim", 0, nonReduceNameRefs);
+    threadsToTensor.passThrough({"rDim"}, {1}, {lowerNameRefs[reduceAxis]});
+    TransformMapAttr mergeTrMap = threadsToTensor.get();
+
+    threadsToTensor = BottomUpTMBuilder::above(threadsToTensor, mergeTrMap);
+    // Distribute threads that is extra than nonReduceMergeDimSize and use them
+    // for reductions. This is a floor because max threads per non-reduction
```
krzysz00: I have no idea what this comment is trying to say, please reword.

mlir/test/Dialect/Rock/integration/reduce/blockwise_reduce/blockwise_reducemax_nr_threads_gt_blocksize.mlir
```diff
@@ -0,0 +1,22 @@
+// RUN: cat %s | rocmlir-gen -ph -print-results -rand fixed - | rocmlir-driver -arch %arch -c  | mlir-cpu-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void | FileCheck %s
```
krzysz00: Useless use of cat nit.

krzysz00 (commented):
Documentation must be improved

I think the transform stack editing approach works broadly OK and we can always find cleaner ways to do this later.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: Done

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: What redundant loads ?

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
manupak: done

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
manupak: reworded, see if it makes sense.

mlir/test/Dialect/Rock/integration/reduce/blockwise_reduce/blockwise_reducemax_nr_threads_gt_blocksize.mlir
manupak: changed.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: hmm no, a single thread could have multiple items to reduce with other items in other threads. I reworded with "with at least number of items to reduce per block."

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: See if what you mean is related to https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/982

krzysz00: I'm saying that if you load the result values for (i, 0) == (i, 1) ... on a lane (after reducing in j) that you'll only access LDS once

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -1673,6 +1673,82 @@ LogicalResult ReduceOp::verify() {
...
+    return emitError("No gpu memspace attr found in workspace memref; the "
+                     "workspace memref should be in LDS");
+  } else {
+    if (wsMemSpaceAttr.getValue() != gpu::AddressSpace::Workgroup) {
+      return emitError("workspace should be in LDS.");
+    }
+  }
+
+  if (inputThreadView[0] != blockSize) {
+    return emitError("second dim of the input thread should tid");
```
krzysz00: "first dimension of the input view should be equal to the block size"

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1145,6 +1145,46 @@ def Rock_AccelGemmOp:
...
+                   )> {
+  let summary = "Blockwise reduce operation with output broadcasting";
+  let description = [{
+    The `rock.blockwise_broadcast_reduce` op does reduce at a blockwise level and
+    to fill all elements of the input tensor with reduced values. Note that both input/output
+    tensors live in registers.
+
+    $input tensor resides in registers in threads.
+
+    $inputRegView should be register {tid, iter}
```
krzysz00: "should be a coordinate transform taking `{tid, iter}` (where `iter` is the register number in `$input` and `tid` is the ID of the lane within the workgroup and producing indices into a virtual tensor `{d0, ..., dR, ... dN} which is spread across the lanes of the workgroup."

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1145,6 +1145,46 @@ def Rock_AccelGemmOp:
...
+                   TransformMapArrayAttr:$inputRegView,
+                   I32Attr:$blockSize
+                   )> {
+  let summary = "Blockwise reduce operation with output broadcasting";
+  let description = [{
+    The `rock.blockwise_broadcast_reduce` op does reduce at a blockwise level and
+    to fill all elements of the input tensor with reduced values. Note that both input/output
+    tensors live in registers.
+
+    $input tensor resides in registers in threads.
```
krzysz00: "in registers, where each thread holds a subset of the elements of the tensor as defined by $inputView"

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1145,6 +1145,46 @@ def Rock_AccelGemmOp:
...
+
+    $inputRegView should be register {tid, iter}
+    to virtual tensor {d0, ... , Dr , ... , dn} which is a slice of
+    a larger tensor {D0, ... , Dr , ... , Dn}. Note that the reduction
+    axis (Dr) is not sliced because it is not supported.
+
+    $output tensor resides in registers in threads. It would have
```
krzysz00: I don't understand this paragraph at all, please reword.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1145,6 +1145,46 @@ def Rock_AccelGemmOp:
...
+    the same dimensionality of the input tensor but broadcasted.
+    So the users of the op, has to add a extraView if they are writing back
+    to some memory (LDS/Global) to bar threads (or iteration within threads)
+    to not to create a broadcasted tensor.
+
+    $workspace_buffer resides in LDS. This should be a flat tensor with at least
+    number of items to reduce per block.
```
krzysz00: Document how that number is computed here?

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -830,17 +830,558 @@ LogicalResult ThreadwiseWriteAllRewritePattern::matchAndRewrite(
...
+
+  // This should only be used if product non-reduction dims is
+  // less than number threads in a block.
+  //
+  // Given a input tensor : D0, ... , Dr , ... , DN to reduce,
+  // This function creates a view that maps the space of
+  // [D0, ... , Dr , ... , DN] --> [nrtid, rtid, rIter] where
+  // nrtid = tid / product(non-reduction dims) is a reduction subgroup leader.
+  // rtid = tid % product(non-reduction dims) is thread idx within a reduction
+  // subgroup. | rtid | is the number of threads that'd participate in a
```
krzysz00: Nit: Why the pipes?

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -830,17 +830,558 @@ LogicalResult ThreadwiseWriteAllRewritePattern::matchAndRewrite(
...
+        // each individual thread could do its own reduction.
+        ArrayAttr threadsToTensorTrs = createThreadViewForNRLargerThanThreads(
+            loc, regTensorShape, blockSize, axis, rewriter);
+        ArrayAttr threadToLDSViewTrs =
+            createLDSWorkspaceView(loc, rewriter, threadsToTensorTrs, axis);
+        ArrayAttr threadsToLDSViewReducedTrs = createLDSWorkspaceView(
+            loc, rewriter, threadsToTensorTrs, axis, /*makeRDimZero-*/ true);
+        ArrayRef<int64_t> threadViewShape =
+            threadToLDSViewTrs[0].cast<TransformMapAttr>().getUpperBounds();
+        ArrayRef<int64_t> ldsBufferShape =
```
krzysz00, line 1192: This should be `ldsBuffer.getType().getShape()`, no ?

krzysz00 (commented):
Mostly documentation notes.

In addition ... weren't there unit tests before? Could we add some unit tests?

manupak:
> weren't there unit tests before? Could we add some unit tests?

There weren't; I ll add those now -- was waiting on alignment on the implementation as unit tests can be a pain to keep changing.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
manupak: done

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: done

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: done

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: ok changed to num elements in the input tensor as we went with D1 now.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
manupak: It was there to signify the magnitude of the dimension but worded as is instead

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
manupak: no, it is fine for the ldsBuffer to be larger to enable reuse between ops

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: reworded and removed additional/unncessary info on "how to retrieve the non-broadcasted" tensor which has nothing to do with this op.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
krzysz00: Poke re what the codegen looks like

krzysz00 (commented):
Code is approved, will approve given some unit tests

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: I like to propose to defer this exercise to perf improvement stage of this work : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/988

manupak:
@krzysz00 , ok Ive added unit tests

mlir/test/Dialect/Rock/lowering_blockwise_broadcast_reduce.mlir
```diff
@@ -0,0 +1,122 @@
+// RUN: rocmlir-opt -rock-blockwise-gemm-to-threadwise -canonicalize -split-input-file %s | FileCheck %s
+
+// CHECK: #[[MAP:.*]] = affine_map<(d0, d1) -> (d1, d0)>
```
krzysz00: Nit: CHECK-DAG here so these can be reordered safely

mlir/test/Dialect/Rock/lowering_blockwise_broadcast_reduce.mlir
```diff
@@ -0,0 +1,122 @@
...
+// CHECK: #[[MAP3:.*]] = affine_map<(d0, d1) -> (d0, d1)>
+// CHECK: #[[MAP4:.*]] = affine_map<(d0, d1) -> (0, d0)>
+// CHECK: #[[TMAP:.*]] = #rock.transform_map<#[[MAP]] {{.*}}
+// CHECK: #[[TMAP1:.*]] = #rock.transform_map<#[[MAP]] by {{.*}}
+// CHECK: #[[TMAP2:.*]] = #rock.transform_map<#[[MAP1]] by {{.*}}
+// CHECK: #[[TMAP3:.*]] = #rock.transform_map<#[[MAP2]] by {{.*}}
+// CHECK: #[[TMAP4:.*]] = #rock.transform_map<#[[MAP3]] by {{.*}}
+// CHECK: #[[TMAP5:.*]] = #rock.transform_map<#[[MAP3]] by {{.*}}
+// CHECK: #[[TMAP6:.*]] = #rock.transform_map<#[[MAP4]] by {{.*}}
+// CHECK: func @rock_blockwise_reducesum_nr_threads_gt_blocksize
```
krzysz00, line 15: Nit: CHECK-LABEL

krzysz00 (approved):
Approved, nitpicks on the unit test

mlir/test/Dialect/Rock/lowering_blockwise_broadcast_reduce.mlir
manupak: We cant use that because CHECK-LABEL splits the scoping and we need stuff from above func for stuff below right ?

krzysz00: If you go poking around the FileCheck docs there's ways to have global variables. I don't recall if there's a flag or if it's just leading `$`. I know I've used it in other tests.

PR #1109, opened 2023-06-09
Add "rocm" prefix build-dockerfile.sh

manupak:
I have mistakenly dropped the "rocm" prefix
in the image tag when fixing the version extraction previously.

This commit fixes that to re-instate "rocm" prefix.


jerryyin (approved):

manupak:
Can we cheat the CI on this one ? because only one matters is the one where jenkinsfile is updated to use the images.

jerryyin:
@manupak Fine by me. In case it is a machine related issue, could you the following:
 - Take a look at whether rebooting scripts has been triggered. Persistent failure usually are triggered by those.
 - Ask Omkar to take a look if the reboot script doesn't work any more.

PR #1111, opened 2023-06-12
[DO NOT SQUASH] arith.extf on 8-bit floats by tables

krzysz00:
1. Set up the lowering of fp8 constants to i8 everywhere so translation to LLVM IR doesn't explode
2. Add a pass that implements a translation from `arith.extf` on 8-bit floats to table lookup and inserts the relevant tables.

(End-to-end hookups are a followup PR)


jerryyin (approved):
LGTM

PR #1112, opened 2023-06-12
Add E2E test support for fp8

krzysz00:
1. Fix up the "rocdl" pipeline option to rocmilr-driver which has gotten out of sync with the backend pipeline
2. Fix nesting of atomics emulation so that it actually runs and only runs on the GPU modules.
3. Add a boolean flag for whether a target has fp8 conversion instructions to the arch DB (since we can't toggle their use on a per-kernel basis, this isn't a gemm feature)
4. Add --convert-arith-to-amdgpu (if supported based on the flag above) and --fp8-ext-to-tables to the backend pipelines. (**NOTE**: These fp8 passes are not running in the xmodel pipeline. This is a consequence of the hacky fp8 tables pass being downstream and not a patch onto upstream, so it can't be added to Xmodel pipelines. Since xmodel is being moved anyway this is fine for now?)
5. Add tests to make sure fp8 ops are lowered as expected.
6. Add fp8 to select end-to-end tests to ensure minimal coverage of the datatype (which is all we really need because it acts like int8 for most purposes).

### DIFF
--- a/mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
+++ b/mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
@@ -309,14 +309,13 @@ LogicalResult Conv2dGenerator::getBwdWeightKernelCount(OpBuilder &builder,
 
   kernelCount = 1;
   if (isAccel(config.features)) {
-    Type dataType = getInputDataType(builder);
     bool needExtraPad = false;
     if (failed(needExtraPadBwdWeight(builder, needExtraPad))) {
       return failure();
     }
     if (!needExtraPad) {
       Type dataType = getInputDataType(builder);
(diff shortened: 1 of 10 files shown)


mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp
```diff
@@ -309,14 +309,13 @@ LogicalResult Conv2dGenerator::getBwdWeightKernelCount(OpBuilder &builder,
 
   kernelCount = 1;
   if (isAccel(config.features)) {
-    Type dataType = getInputDataType(builder);
```
krzysz00, line 312: Misc warning cleanup while I was there.

sjw36 (approved):
Looks good.

PR #1116, opened 2023-06-13
Reorganize and add MIGraphX elementwise ops

krzysz00:
1. Move the `erf` conversion to the tablegen-based pattern rewrite system
2. Add other elementwise ops that both MIGraphX and Tosa support (abs, ceil, floor, sigmoid, log) to those patterns
3. Add tests
4. Organize everything into neat alphabetical order while I'm here.
5. Define i32 `div` just in case.

See https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/928 for where I got this list.


manupak (approved):
LGTM!

jerryyin (approved):
Good job!

PR #1117, opened 2023-06-15
Enable E2E test cases fusion/resnet50 and xmir/bert-torch-tosa

yiqian1:
Reorganize the e2e test cases to run resnet50 and bert-torch-tosa in nightly. 

Previously the parent directories (fusion/e2w and xmir/e2e) were only enabled in PR so it wouldn't go into the subdirectories (resnet50 and bert-torch-tosa).


krzysz00 (approved):
LGTM, approved

PR #1118, opened 2023-06-15
Use a more interleaved block layout

giuseros:
Use a more interleaved layout when mPerWave/nPerWave > 1. 

Instead of `wave[0]` computing `block[0:mPerWave]`, let `wave[0]` compute blocks `block[i*numWaves:(i+1)*numWaves+mPerAccel]`  where `0 <= i < mPerWave`

Fixes: https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/925

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
@@ -1,3 +1,27 @@
+//===- AccelEmitter.cpp - MLIR helper to emit acceleration intrinsics
+//---------------===//
+//
+// Copyright 2020 The MLIR Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// =============================================================================
+//
+// This class tries to abstract away the code-generation details needed to
+// generated calls to matrix multiply accelerator intrinsics (wmma, mfma).
+//
+//
+//===----------------------------------------------------------------------===//
+
 #include "AccelEmitter.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/Rock/utility/AmdArchDb.h"
@@ -108,8 +132,8 @@ AccelEmitterParams MfmaEmitter::initAccelEmitterParams(
   params.mRepeats = mfmaGroup.getMRepeats(mPerWave);
   params.nRepeats = mfmaGroup.getNRepeats(nPerWave);
   params.nResultVectors = mfmaGroup.getImms().size();
-  params.mPerAccel = mfmaGroup.getLenPerMfmaGroup(mPerWave);
-  params.nPerAccel = mfmaGroup.getLenPerMfmaGroup(nPerWave);
+  params.mPerAccel = mfmaAttr.mfmaNonKDim * mfmaGroup.getImms().size();
+  params.nPerAccel = mfmaAttr.mfmaNonKDim * mfmaAttr.blocksMfma;
   params.kpackPerThread =
       (mfmaAttr.isKReduction ? kpackPerBlock / mfmaAttr.inputSpansPerMfmaIn
                              : kpackPerBlock);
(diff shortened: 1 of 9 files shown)


mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -108,8 +132,8 @@ AccelEmitterParams MfmaEmitter::initAccelEmitterParams(
   params.mRepeats = mfmaGroup.getMRepeats(mPerWave);
   params.nRepeats = mfmaGroup.getNRepeats(nPerWave);
   params.nResultVectors = mfmaGroup.getImms().size();
-  params.mPerAccel = mfmaGroup.getLenPerMfmaGroup(mPerWave);
-  params.nPerAccel = mfmaGroup.getLenPerMfmaGroup(nPerWave);
+  params.mPerAccel = mfmaAttr.mfmaNonKDim * mfmaGroup.getImms().size();
+  params.nPerAccel = mfmaAttr.mfmaNonKDim * mfmaAttr.blocksMfma;
```
krzysz00: Is this right?

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -236,11 +265,12 @@ ArrayAttr MfmaEmitter::computeOutputTransforms(PatternRewriter &b, Location loc,
   toMatrixC.embed(
       "gemmM", 1, matrixM,
       {"m", "wave_m", "m_tid", "m_i", "blk_row", "vec_group", "vec_item"},
-      {mPerBlock, mPerWave, rowGroupSize, mPerRepeat, m,
+      {mPerBlock, mPerAccel, rowGroupSize, mPerAccel * mWaves, m,
```
krzysz00, line 271: These names are meaningfully different, and if you're going to change them, can you update `"wave_m"` to `"accel_m"` above as well?
And I think `mPerAccel * mWaves` should be `== mPerRepeat`?

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -424,12 +465,14 @@ ArrayAttr WmmaEmitter::computeOutputTransforms(PatternRewriter &b, Location loc,
   // tile. For workitems 0 to 15 m_tid is 0 and they write to 0,2,4,...,14.  For
   // workitems 16 to 31 m_tid is 1 and they write at positions 1,3,5,..,15 .
   // This is outlined in https://gpuopen.com/learn/wmma_on_rdna3/
-  toMatrixC.embed(
```
krzysz00, line 427: Update  comment?

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -424,12 +465,14 @@ ArrayAttr WmmaEmitter::computeOutputTransforms(PatternRewriter &b, Location loc,
...
-      "gemmM", 1, matrixM, {"m", "wave_m", "m_tid", "rep_i", "item_i"},
-      {mPerBlock, mPerWave, 1, wmmaInsn.inputLen, wmmaInsn.outputStride});
+  toMatrixC.embed("gemmM", 1, matrixM,
+                  {"m", "wave_m", "m_tid", "rep_i", "item_i"},
+                  {mPerBlock, wmmaInsn.inputLen, 1, mWaves * wmmaInsn.inputLen,
```
krzysz00, line 477: Similarly, let's fix or standardize on the repeat names?

krzysz00 (commented):
Minor maintenance concerns

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
giuseros: Why? We still need  `wave_m`. It's just the coefficient of the `embed` that is different. I mean, I am still looking at `wave_0`, `wave_1`, and so on, but their coefficient is different. 
- Before `wave_m` was mutiplied by `mPerWave`, because we were doing all the repeats together
- Now `wave_m` is multiplied by `mPerAccel`, because the i-th wave only computes `mPerAccel` and then "jump" of `mWaves`

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
giuseros: I think we miscomputed those, but we didn't use those values anywhere in the code (setting them both to (0,0) produces no failures). 

But I  over-complicated this calculation: what I needed was `mPerWave/mRepeats` and `nPerWave/nRepeats`. I.e., how many rows/colums a given wave computes in a single iteration.

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
giuseros: Same here, I think that `rep_i` is fine, it's just that the stride is larger (before it was `inputLen`, now it is `inputLen*waves`

giuseros:
Performance regression report: [regression.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/11772661/regression.zip)

### Summary
Few improvements in `int8`, flat for the rest (which is what I already expected)

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
krzysz00: Ah. Makes sense.

krzysz00 (approved):

PR #1119, opened 2023-06-16
Add rock.blockwise_fill op

manupak:
This commit add rock.blockwise_fill op
that could be used to fill up a LDS
memref with data.

closes : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/881

### DIFF
--- a/mlir/lib/Dialect/Rock/IR/RockDialect.cpp
+++ b/mlir/lib/Dialect/Rock/IR/RockDialect.cpp
@@ -1673,6 +1673,33 @@ LogicalResult ReduceOp::verify() {
   return success();
 }
 
+//===-----------------------------------------------------===//
+// BlockwiseFillOp
+//===-----------------------------------------------------===//
+
+LogicalResult BlockwiseFillOp::verify() {
+  MemRefType memrefType = getMemref().getType();
+  if (memrefType.getRank() != 1) {
+    return emitError("Blockwise fill expects a flat memref");
+  }
+  if (gpu::AddressSpaceAttr memSpace =
+          memrefType.getMemorySpace()
+              .dyn_cast_or_null<gpu::AddressSpaceAttr>()) {
+    if (memSpace.getValue() != gpu::AddressSpace::Workgroup) {
+      return emitError("Memory space is expected to be workgroup");
+    }
+  } else {
+    return emitError("Memory space is expected to be workgroup");
+  }
+  int64_t numElements = getMemref().getType().getNumElements();
+  if (VectorType vecType = dyn_cast<VectorType>(getValue().getType())) {
+    if (numElements % vecType.getNumElements() != 0) {
+      return emitError("The vector length is not a factor in memref size.");
+    }
+  }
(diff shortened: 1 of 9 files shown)


krzysz00 (commented):
Code-wise approved.
Test-wise, to save on testing load, could you please make those blockwise_fill tests unit tests (over in `mlir/Dialect/Rock`) and not integration tests? There's no real need to integration test this op.

manupak:
I'd like to keep the integration tests as they helped me fix somethings while in the implementation. If the CI load is the concern, I ll disable them but want to have them to run them if needed (to rule out that they are working)

manupak:
> could you please make those blockwise_fill tests unit tests (over in mlir/Dialect/Rock) 

this is already done

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -1673,6 +1673,33 @@ LogicalResult ReduceOp::verify() {
   return success();
 }
 
+//===-----------------------------------------------------===//
+// BlockwiseFillOp
+//===-----------------------------------------------------===//
+
+LogicalResult BlockwiseFillOp::verify() {
+  MemRefType memrefType = getMemref().getType();
+  if (memrefType.getRank() != 1) {
```
krzysz00, line 1682: Nit: I think there's a way to write this constraint in the tablegen

mlir/test/Dialect/Rock/integration/blockwise_fill/lit.local.cfg
```diff
@@ -0,0 +1,3 @@
+# These tests are disabled in the interest of CI load
+# However, they should be passing.
+config.unsupported = True
```
krzysz00, line 3: Won't it throw people that these tests are always unsupported? Wouldn't it be better to remove them?

krzysz00 (approved):
Approved with nits

mlir/test/Dialect/Rock/integration/blockwise_fill/lit.local.cfg
krzysz00: Ah, I saw the above comment.

And the concern is less CI resource and more "me running `check-rocmlir` resources", and that an integration test isn't that needed when you have unit tests.

manupak: Yes -- the thing is I'd like have one or two integration test per feature, especially say when refactors happen we'll need to change unit tests but it would be unlikely change the integration test. The unit tests more or less stands as documentation, IMO. So having the integration test kinds of gives confidence to change the unit tests in a refactor and also having the confidence that nothing is broken by the said refactor.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
manupak: Noted. I ll slide this in with another PR.

PR #1120, opened 2023-06-16
Add clip, convert (on floats), and where to MIGraphX lowering

krzysz00:
These operations have been identified as potential further fusion candidates and so this commit adds support for them to our lowering flow.

### DIFF
--- a/mlir/test/Conversion/MIGraphXToTosa/mixr-to-tosa-ops.mlir
+++ b/mlir/test/Conversion/MIGraphXToTosa/mixr-to-tosa-ops.mlir
@@ -114,6 +114,42 @@ module  {
     return %3 : tensor<1x64x112x112xf32>
   }
 
+  // CHECK-LABEL: func.func @clip_i32
+  func.func @clip_i32(%arg0: tensor<64x64xi32>, %arg1: tensor<64x64xi32>, %arg2: tensor<64x64xi32>) -> tensor<64x64xi32> attributes {arch = "gfx90a:sramecc+:xnack-", kernel = "mixr"} {
+    // CHECK: %[[MAX:.*]] = "tosa.maximum"(%arg0, %arg1)
+    // CHECK: %[[MIN:.*]] = "tosa.minimum"(%[[MAX]], %arg2)
+    // CHECK: return %[[MIN]]
+    %0 = migraphx.clip(%arg0, %arg1, %arg2) : (tensor<64x64xi32>, tensor<64x64xi32>, tensor<64x64xi32>) -> tensor<64x64xi32>
+    return %0 : tensor<64x64xi32>
+  }
+
+  // CHECK-LABEL: func.func @clip_broadcast
+  func.func @clip_broadcast(%arg0: tensor<64x64xf16>, %arg1: tensor<1x64xf16>, %arg2: tensor<1xf16>) -> tensor<64x64xf16> attributes {arch = "gfx90a:sramecc+:xnack-", kernel = "mixr"} {
+    // CHECK: %[[BCAST2:.*]] = "tosa.reshape"(%arg2) {new_shape = array<i64: 1, 1>}
+    // CHECK: %[[MAX:.*]] = "tosa.maximum"(%arg0, %arg1)
+    // CHECK: %[[MIN:.*]] = "tosa.minimum"(%[[MAX]], %[[BCAST2]])
+    // CHECK: return %[[MIN]]
+    %0 = migraphx.multibroadcast(%arg1) {out_dyn_dims = [], out_lens = [64, 64]} : (tensor<1x64xf16>) -> tensor<64x64xf16>
+    %1 = migraphx.multibroadcast(%arg2) {out_dyn_dims = [], out_lens = [64, 64]} : (tensor<1xf16>) -> tensor<64x64xf16>
+    %2 = migraphx.clip(%arg0, %0, %1) : (tensor<64x64xf16>, tensor<64x64xf16>, tensor<64x64xf16>) -> tensor<64x64xf16>
+    return %2 : tensor<64x64xf16>
+  }
+
+  // CHECK-LABEL: func.func @where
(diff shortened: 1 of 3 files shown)


mlir/test/Conversion/MIGraphXToTosa/mixr-to-tosa-ops.mlir
```diff
@@ -217,6 +253,13 @@ module {
      return %0 : tensor<16xf32>
   }
 
+  // CHECK-LABEL: func.func @func_convert
+  // CHECK: tosa.cast
+  func.func @func_convert(%arg0: tensor<16xf16>) -> tensor<16xf32> {
+    %0 = "migraphx.convert"(%arg0) : (tensor<16xf16>) -> tensor<16xf32>
```
jerryyin, line 261: Just to give a heads up that they are not actually equivalent. 
 - migraphx.convert doesn't saturate
 - tosa.cast saturate

I'd say this is fine because maybe 99% after a migraphx.convert it has additional op to clamp. In that case it is okay to make clamp a no-op.

mlir/test/Conversion/MIGraphXToTosa/mixr-to-tosa-ops.mlir
```diff
@@ -114,6 +114,42 @@ module  {
     return %3 : tensor<1x64x112x112xf32>
   }
 
+  // CHECK-LABEL: func.func @clip_i32
+  func.func @clip_i32(%arg0: tensor<64x64xi32>, %arg1: tensor<64x64xi32>, %arg2: tensor<64x64xi32>) -> tensor<64x64xi32> attributes {arch = "gfx90a:sramecc+:xnack-", kernel = "mixr"} {
+    // CHECK: %[[MAX:.*]] = "tosa.maximum"(%arg0, %arg1)
+    // CHECK: %[[MIN:.*]] = "tosa.minimum"(%[[MAX]], %arg2)
+    // CHECK: return %[[MIN]]
+    %0 = migraphx.clip(%arg0, %arg1, %arg2) : (tensor<64x64xi32>, tensor<64x64xi32>, tensor<64x64xi32>) -> tensor<64x64xi32>
```
jerryyin, line 122: Use tosa.clamp instead?

jerryyin (approved):
Comments are non-blocking, LGTM

mlir/test/Conversion/MIGraphXToTosa/mixr-to-tosa-ops.mlir
krzysz00: Ok, so, firstly, I think the bullet points are backwards.
Second, we're only implementing `convert` for float-to-float conversions. Can we get some confirmation that `float -> half` is supposed to saturate instead of rounding to infinity if needed? Because those are some weird semantics.

(since `half -> float` doesn't have this problem)

mlir/test/Conversion/MIGraphXToTosa/mixr-to-tosa-ops.mlir
krzysz00: Nope. `tosa.clamp` is for uniform, constant clamps only, which isn't the semantics of `migraphx.clip`. We could, hypothetically, try to optimize the case where the min and max are literal unit tensors, but that'll generate the same code when you lower the tosa.clamp anyway.

jerryyin: Fair enough, yah, you reminded me about it.

mlir/test/Conversion/MIGraphXToTosa/mixr-to-tosa-ops.mlir
krzysz00: Chat with MIGraphX confirms that their reference implementation isn't quite right for floats.

jerryyin: For float->half, 
 - could you double check how TOSA to linalg lowering works? Their unit tests are quite helpful there. 
   - If there's a saturate vs. infinity behavior than it's going to be a problem. Either way, Tosa.cast has a deterministic lowering, and by swapping that to a Tosa.cast may force it to a behavior it didn't expect.
 - BTW are there really float-> half conversions in middle of the model? I'd more or less expect that to happen as part of the quant/dequant lowering

krzysz00: This isn't related to quant - this is in fp16 models where you want to go for fp32 on some of the elementwise ops, I think.

`tosa.cast` just lowers to `arith.truncf` 

I'm going to @umangyadav here for our half of the discussion, because, as far as I can tell, MIGraphX's convert on floats is supposed to be `truncf()`

jerryyin: >  this is in fp16 models where you want to go for fp32 on some of the elementwise ops

Fair enough! Very reasonable up cast behavior.

umangyadav: > MIGraphX's convert on floats is supposed to be truncf() 

By `truncf()` if you mean this  https://en.cppreference.com/w/c/numeric/math/trunc  ? 

MIGraphX' Ref implementation just calculates : `std::min(std::numeric_limits<T>::max(), std::max(std::numeric_limits<T>::min(), T(x)));` where T is any type and x is the input.  Notice that it casts inputs to target_type before clamping. 
This should work for any Floating points as long as it is not INFs or NaNs because comparisons don't work on them.

krzysz00: Ah. Yeah, I'm seeing the cast to the target type.

... why are those mins and maxes there, then?

krzysz00: By truncf I mean, effectively, LLVM's floating-point truncation instruction, which is probably equivalent to C's `truncf()`

umangyadav: > .... why are those mins and maxes there, then? 

Not sure, looks like just casting would do it.

PR #1122, opened 2023-06-20
Add tosa -> rock.attention conversion

manupak:
This commit adds frontend op rock.attention
and creates conversions from tosa.

closes : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/959

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -202,6 +202,43 @@ def Rock_ReduceOp :
   }];
 }
 
+def Rock_AttentionOp :
+  Rock_Op<"attention", [AllElementTypesMatch<["queries", "keys", "values", "scale"]>]>,
+  Arguments<(ins
+    Arg<TensorOrMemRefOf<[F32, F16]>, "queries", [MemRead]>:$queries,
+    Arg<TensorOrMemRefOf<[F32, F16]>, "keys", [MemRead]>:$keys,
+    Arg<TensorOrMemRefOf<[F32, F16]>, "values", [MemRead]>:$values,
+    Arg<TensorOrMemRefOf<[F32, F16]>, "scale", [MemRead]>:$scale,
+    Arg<TensorOrMemRefOf<[F32, F16]>, "output", [MemRead, MemWrite]>:$out,
+    Rock_GemmFeaturesAttr:$features,
+    OptionalAttr<I32Attr>:$blockSize,
+    OptionalAttr<I32Attr>:$gridSize
+  )>,
+  Results<(outs Optional<TensorOf<[F32, F16]>>:$result)> {
+  let summary = "Attention operation of transformer models";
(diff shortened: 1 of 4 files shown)


mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -202,6 +202,43 @@ def Rock_ReduceOp :
   }];
 }
 
+def Rock_AttentionOp :
```
krzysz00, line 205: Can we implement `RockGemmWrapperInterface` for this, even if in a somewhat artificial way?

krzysz00 (commented):
I don't have any issues with the PR itself, but I would like to understand, for example, how we'll be tuning a `rock.attention`

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: RockGemmWrapperInterface seems specific to gemm-based kernels (and their interface methods).
Hence, I dont want to bring that here now. For the bring-up, we can set a universal param whilst rock.attention op could accept a "perf-config". I'd like to sort the tuning after we have something functional.

Going forward, one idea might be to introduce a RockTunableOpInterface that could prescribe the following:
1) op->getTuningRange(struct TunableParams* tuningSpace)
2) op->tuningSetParam(ParamEntry* paramEntry)

krzysz00: Ok, the only methods specific to gemms on `RockGemmWrapperInterface` are `getAType` and `getBType`. If it weren't for those, we could actually straight-up go ahead and rename the darn thing to `RockTunableKernelInterface` or something.

Given the constraints on flash attention, I'd be happy with an implementation that sets `getAType() == getBType() == [get the type of something in there]`.

On the third hand, I think there's an argument that we could have, for example `queries : fp8, keys : bf8, values : f32` or `queries : int8, keys : int8, values : int32` or something like that, so perhaps `getAType() = getElementType(queries)` and `getBType = getElementType(keys)`, but I don't fully know the usage context.

So I think there's an argument, for now, for using the inner GEMM (queries * keys) as the gemm for RockGemmWrapperInterface, which would let us plug `attention` into the high-level flows along with `conv` and `gemm`.

manupak: Lets move this discussion to : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/954.
I dont think this is relevant to this PR.

krzysz00: Agreed, and agreed that there's a longer design discussion to be had here that is out of scope for this PR.

krzysz00 (commented):
Approved per earlier comments, deferral of tuning discussion to a later date

krzysz00 (approved):
... this time with actually hitting the approve setting lol

PR #1123, opened 2023-06-21
Fix fp8 CI with -rand N by disabling random fp8 tests

krzysz00:
arith.truncf isn't implemented for fp8 so don't generate code that uses it, which means don't allow -rand for those types.


sjw36 (approved):
Looks good.

PR #1124, opened 2023-06-21
[mlir][CAPI] Expose the rest of MLIRContext's constructors

krzysz00:
It's recommended practice that people calling MLIR in a loop pre-create a LLVM ThreadPool and a dialect registry and then explicitly pass those into a MLIRContext for each compilation. However, the C API does not expose the functions needed to follow this recommendation from a project that isn't calling MLIR's C++ dilectly.

Add the necessary APIs to mlir-c, including a wrapper around LLVM's ThreadPool struct (so as to avoid having to amend or re-export parts of the LLVM API).

### DIFF
--- a/external/llvm-project/mlir/include/mlir-c/IR.h
+++ b/external/llvm-project/mlir/include/mlir-c/IR.h
@@ -83,7 +83,18 @@ typedef struct MlirNamedAttribute MlirNamedAttribute;
 //===----------------------------------------------------------------------===//
 
 /// Creates an MLIR context and transfers its ownership to the caller.
+/// This sets the default multithreading option (enabled).
 MLIR_CAPI_EXPORTED MlirContext mlirContextCreate(void);
+
+/// Creates an MLIR context with an explicit setting of the multithreading
+/// setting and transfers its ownership to the caller.
+MLIR_CAPI_EXPORTED MlirContext
+mlirContextCreateWithThreading(bool threadingEnabled);
+
+/// Creates an MLIR context, setting the multithreading setting explicitly and
+/// pre-loading the dialects from the provided DialectRegistry.
+MLIR_CAPI_EXPORTED MlirContext mlirContextCreateWithRegistry(
+    MlirDialectRegistry registry, bool threadingEnabled);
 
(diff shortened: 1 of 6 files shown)


manupak:
Interesting.. 

> It's recommended practice that people calling MLIR in a loop pre-create a LLVM ThreadPool and a dialect registry and then explicitly pass those into a MLIRContext for each compilation.

Is this something upstream is aware already ? (your comment sounds like it)

krzysz00:
@manupak It's straight-up from a comment upstream, see https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/IR/MLIRContext.h#L53

external/llvm-project/mlir/include/mlir-c/Support.h
```diff
@@ -56,6 +56,8 @@ extern "C" {
   };                                                                           \
   typedef struct name name
 
+/// Re-export llvm::ThreadPool so as to avoid
```
pcf000: Incomplete sentence?

external/llvm-project/mlir/include/mlir-c/IR.h
```diff
@@ -83,8 +83,19 @@ typedef struct MlirNamedAttribute MlirNamedAttribute;
...
+/// This sets the default multithreading option (enabled).
 MLIR_CAPI_EXPORTED MlirContext mlirContextCreate(void);
 
+/// Creates an MLIR context with an explicit setting of the multithreading
+/// setting and transfers its ownership to the caller.
+MLIR_CAPI_EXPORTED MlirContext
+mlirContextCreateWithThreading(bool threadingEnabled);
+
+/// Creates an MLIR context, setting the multithreading setting explicitly and
+/// pre-loading the dilects from the provided DialectRegistry.
```
pcf000: Spell-check.

pcf000:
Looks good to me.  I assume this is headed upstream.

krzysz00:
@pcf000  Yep, this'll head upstream once we land it, thanks for catching the typos!

sjw36 (approved):
Looks good.

PR #1125, opened 2023-06-22
Call rocblas in the same way migraphx does

giuseros:
This partially covers https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/932 

We were assuming MIGraphx was using a different layout and hence invoking rocBLAS in a different way. Let's fix this and run the tuner again

### DIFF
--- a/mlir/utils/performance/rocblas-benchmark-driver/rocblas-benchmark-driver.cpp
+++ b/mlir/utils/performance/rocblas-benchmark-driver/rocblas-benchmark-driver.cpp
@@ -108,17 +108,22 @@ int main(int argc, char **argv) {
 
   for (int i = 0, e = args.kernelRepeats; i < e; ++i)
     if (args.gemmG == 1) {
-      ROCBLAS_ABORT_IF_FAIL(rocblas_gemm_ex(
-          handle, rocblas_operation_none, rocblas_operation_none, args.gemmN,
-          args.gemmM, args.gemmK, alpha, bDevice, inType, ldb, aDevice, inType,
-          lda, beta, cDevice, outType, ldc, cDevice, outType, ldc,
-          /*computeType=*/outType, rocblas_gemm_algo_standard, 0,
-          rocblas_gemm_flags_none));
+      ROCBLAS_ABORT_IF_FAIL(
+          rocblas_gemm_ex(handle, /*trans_a=*/blasTransB,
+                          /*trans_b=*/blasTransA, /*m=*/args.gemmN,
+                          /*n=*/args.gemmM, /*k=*/args.gemmK, alpha,
+                          /*a=*/bDevice, /*a_type=*/inType, /*lda=*/ldb,
+                          /*b=*/aDevice, /*b_type=*/inType, /*ldb=*/lda, beta,
+                          cDevice, outType, ldc, cDevice, outType, ldc,
+                          /*computeType=*/outType, rocblas_gemm_algo_standard,
+                          0, rocblas_gemm_flags_none));
 
     } else {
       ROCBLAS_ABORT_IF_FAIL(rocblas_gemm_strided_batched_ex(
-          handle, blasTransB, blasTransA, args.gemmN, args.gemmM, args.gemmK,
-          alpha, bDevice, inType, ldb, strideB, aDevice, inType, lda, strideA,
+          handle, /*trans_a=*/blasTransB, /*trans_b=*/blasTransA,
+          /*m=*/args.gemmN, /*n=*/args.gemmM, /*k=*/args.gemmK, alpha,
+          /*a=*/bDevice, /*a_type=*/inType, /*lda=*/ldb, /*stride_a=*/strideB,
+          /*b=*/aDevice, /*b_type=*/inType, /*ldb=*/lda, /*stride_b=*/strideA,
           beta, cDevice, outType, ldc, strideC, cDevice, outType, ldc, strideC,
           args.gemmG,
           /*computeType=*/outType, rocblas_gemm_algo_standard, 0,
--- a/mlir/utils/performance/rocblas-benchmark-driver/rocblas-benchmark-driver.cpp
+++ b/mlir/utils/performance/rocblas-benchmark-driver/rocblas-benchmark-driver.cpp
@@ -87,6 +87,11 @@ int main(int argc, char **argv) {
 
   rocblas_datatype inType = getRocblasType(false, args.dataType);
   rocblas_datatype outType = getRocblasType(true, args.dataType);
+  rocblas_datatype computeType = outType;
+
+  if (inType == rocblas_datatype_f16_r) {
+    computeType = rocblas_datatype_f32_r;
+  }
 
   rocblas_initialize();
   rocblas_handle handle;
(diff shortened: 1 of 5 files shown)


mlir/utils/performance/rocblas-benchmark-driver/rocblas-benchmark-driver.cpp
```diff
@@ -54,25 +54,28 @@ int main(int argc, char **argv) {
...
-  // inverted
-  if (!args.transposeA) {
+  // Please note: MIGraphx and MLIR are using row-major format
+  // to store matrices, while rocBLAS is using a colum-major format.
+  // To be compliant to rocBLAS format, MIGraphx swaps the inputs
+  // and tells rocBLAS that B is nxk and A is kxm. So the result
+  // will be a nxm matrix stored in colum-major order. We can simply
+  // recover the original matrix C by reading the matrix in a row-major
+  // way.
+  if (args.transposeA) {
```
manupak, line 83: Something feels not right here...

shouldn't lda associated with M <-> N has to change here.
I mean as I see lda is associated with args.gemmM before/after this change.

giuseros: Yes, but it is the other way round. When it is not transposed, A is a `kxn` column-major  matrix, which means that lda is `k`. If it is transposed, this is stored as a `nxk` column-major matrix, which means lda is `n`

krzysz00: I think the point here is about matrix C.

That matrix, when untransposed, is stored `[M, N]`. However, rocBLAS expects it to be stored `[N, M]`. This means that, our  `C = A * B` is their `tr(C) = tr(B) * tr(A)`

krzysz00: That is, when we have our problem `tr?(C) = tr?(A) * tr?(B)`, that is equivalent to the rocBlas problem `tr(tr?(C)) = tr(tr?(B)) * tr(tr?(A))`, with `tr?` being either a transpose or a noop depending on the flags and `tr(tr(X))` cancelling out.

giuseros: t's really a matter of "interpretation". I will repeat what I understood to also convince myself :)

A row-major `2x3` `A=[1 2 3;4 5 6]` in memory is just a sequence of `[1 2 3 4 5 6]`. We tell rocBlas that this is a `3x2` matrix and he believes us because he will read ` 1 2 3` as a column and then ` 4 5 6` as the next column. So to rocBLAS that is a `3x2` matrix. You do the same exact reasoning with say a `3x10`  Bmatrix. For a row-major library is `3x10`, but we tell rocBLAS that this is a `10x3` matrix. 

After execution, now rocBLAS gives us a `10x2` matrix `C` , and it is fine for us, because we will read the first 10 consecutive elements which is a row for us. So a `10x2` matrix `C` is (in row-major) a `2x10` matrix, which is exactly what we wanted.

About `ld{a,b,c}` , this is the number of elements between element `i,j` and element `i+1,j` for row major and it is the number of elements between `i,j` and `i,j+1` for column major. Going back to our example, our matrix `2x3 `has `lda=3` (size of a row). When we pass this to rocBLAS, it'll become a `3x2` colum-major matrix, and the number of elements between` i,j` and `i,j+1` will still be 3 (size of a column).

Finally, transposition. For what I said above, the transposition flags are untouched. If it's not transposed for us, it won't be transposed for rocBLAS and viceversa (because we re-interpret the matrix)

giuseros: @manupak , @krzysz00  any update on this? I would like to have this in to test performance against rocblas

mlir/utils/performance/rocblas-benchmark-driver/rocblas-benchmark-driver.cpp
```diff
@@ -106,16 +109,16 @@ int main(int argc, char **argv) {
   for (int i = 0, e = args.kernelRepeats; i < e; ++i)
     if (args.gemmG == 1) {
       ROCBLAS_ABORT_IF_FAIL(rocblas_gemm_ex(
-          handle, blasTransA, blasTransB, args.gemmM, args.gemmN, args.gemmK,
-          alpha, aDevice, inType, lda, bDevice, inType, ldb, beta, cDevice,
-          outType, ldc, cDevice, outType, ldc,
+          handle, rocblas_operation_none, rocblas_operation_none, args.gemmN,
+          args.gemmM, args.gemmK, alpha, bDevice, inType, ldb, aDevice, inType,
```
manupak: Lets add the arg name ; i.e. /*arg=*/argValue type here.

manupak (approved):
This all looks good! 

I was not getting the operand swap for rocblas, then we discussed offline and I learnt that is being swapped in the rocblas api call -- thus, I think its better to add arg_name as a comment to highlight this is intentional.

mlir/utils/performance/rocblas-benchmark-driver/rocblas-benchmark-driver.cpp
krzysz00: @giuseros I'm still not convinced your code is correct - see my working out above

krzysz00: That is, `blasTransA` in this branch should be `rocblas_operation_none`

krzysz00: A = [1 2
 3 4
 5 6] : 3x2

B =
[a b c d
 e f g h] : 2x5

C = [1a + 2e, 1b + ...
 3a + 4e ...
 ...] : 3x4

mem(A) = [1 2 3 4 5 6]
idx_A(i, j) (2*i + j) // rocBLAS must do this 
memref<3x2xT> means that the index memref[i][j] is at (2 * i + j)

rocblas (trans off, lda = 2, m = 2, k = 3)
(i, j) -> (i + 2 * j)
(0, 0) -> 0 + (0, 1) -> 2 + (0, 2) -> 4
1? 

rocblas trans off map is (i, j) -> (i + lda * j)
rocblas trans on map is (i, j) -> (j + lda * i)

krzysz00: Ok so taking A : 3x2, B : 2x4, C : 3x4 as above, the MIGraphX method is to compute

C: 4x3 = B: 4x2 * A: 2x3 , passing in the underlying buffers.

This leads to map(B) = (i + 2*j) and map(A) = (j + 2*i), which are arg-swapped versions of what MLIR uses. Similarly, C is (i, j) -> (i + 3 * j), which is also a argument-swapped version of our (i, j) -> (3 * i + j).

So, this method is equivalent to computing C = AB by tr(C) = tr(B)tr(A) implicitly.

krzysz00 (approved):
Approved, now that I'm convinced of this

PR #1126, opened 2023-06-22
Find the gpu module of a tunable kernel

yiqian1:
Make the rocmlir-tuning-driver to tune backward-weight.

### DIFF
--- a/mlir/tools/rocmlir-tuning-driver/rocmlir-tuning-driver.cpp
+++ b/mlir/tools/rocmlir-tuning-driver/rocmlir-tuning-driver.cpp
@@ -288,13 +288,17 @@ static LogicalResult runTuningLoop(ModuleOp source) {
       return failure();
     }
 
-    // Extract binary anb benchmark
+    // Extract binary and benchmark
     std::string hipModule;
-    tuneCopy.walk([&hipModule](gpu::GPUModuleOp op) {
-      hipModule = op->getAttrOfType<StringAttr>("gpu.binary").getValue().str();
-      return WalkResult::interrupt();
+    tuneCopy.walk([&hipModule, &kernelFuncName](gpu::GPUModuleOp op) {
+      std::string moduleName = op.getName().str();
+      if (moduleName == kernelFuncName + "_module") {
+        hipModule =
+            op->getAttrOfType<StringAttr>("gpu.binary").getValue().str();
+        return WalkResult::interrupt();
+      }
+      return WalkResult::advance();
     });
-
     FailureOr<double> timing = benchmarkKernel(
         hipModule.c_str(), kernelFuncName.c_str(), blockSize, gridSize,
         dataType, hostBuffers, gpuBuffers, bufferLengths);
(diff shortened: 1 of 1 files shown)


mlir/tools/rocmlir-tuning-driver/rocmlir-tuning-driver.cpp
```diff
@@ -288,13 +288,17 @@ static LogicalResult runTuningLoop(ModuleOp source) {
...
     }
 
-    // Extract binary anb benchmark
+    // Extract binary and benchmark
     std::string hipModule;
-    tuneCopy.walk([&hipModule](gpu::GPUModuleOp op) {
-      hipModule = op->getAttrOfType<StringAttr>("gpu.binary").getValue().str();
-      return WalkResult::interrupt();
+    tuneCopy.walk([&hipModule, &kernelFuncName](gpu::GPUModuleOp op) {
+      std::string moduleName = op.getName().str();
```
krzysz00, line 294: I there there might be a way to avoid `.str()`-ing here with `op.getName()` and then, in the `if` path, using a `Twine`, but that's not entirely clear.
It's not a massive thing, but it'll save us some memory allocations.

mlir/tools/rocmlir-tuning-driver/rocmlir-tuning-driver.cpp
```diff
@@ -288,13 +288,17 @@ static LogicalResult runTuningLoop(ModuleOp source) {
...
+      if (moduleName == kernelFuncName + "_module") {
+        hipModule =
+            op->getAttrOfType<StringAttr>("gpu.binary").getValue().str();
+        return WalkResult::interrupt();
+      }
+      return WalkResult::advance();
```
krzysz00, line 302: Per discussion, if at all possible, can we get a `llvm::errs() << "Ignoring utility kernels, benchmark times will not match performance tests\n";` in the advance case?

krzysz00:
Approval pending creation of a ticket for doing this the more general way in addition to comments above

yiqian1:
A new ticket [#985](https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/985) for a more general solution has been created.

mlir/tools/rocmlir-tuning-driver/rocmlir-tuning-driver.cpp
yiqian1: probably I can do 
if (moduleName == Twine(kernelFunc.getSymName(), "_module").str())

krzysz00: Do you need the `.str()` there? I forget if there's a `StringRef == Twine`

krzysz00: (if you do then never mind)

krzysz00: ... actually, I'd think to do this
(and this is with `moduleName` staying a `StringRef`)
```
if (!moduleName.endsWith("_module")) {
  llvm::errs() << "GPU module arising from rocmlir pipeline has unexpected name format: << moduleName << "\n";
  [abort the tuning]
}
if (moduleName.drop_back(8) == funcName) {
  ...
}
```

krzysz00 (approved):
Approved to unblock kernel tuning and because this'll work fine, noting nitpicks about string processing that can be fixed later if need be.

PR #1128, opened 2023-06-23
[hotfix] Re-add SoftwareBF16 pass for bf16 emulation

sjw36:
* removed unnecessary libmlir_float16_utils runtime lib


sjw36:
I will run this with nightly enabled.

jerryyin (approved):
LGTM but why can changes here impact accuracy?

krzysz00 (approved):
Can you confirm that the pass is also run on the GPU kernels - that is, that it's still in place as part of `-gpu-to-rocdl`

sjw36:
> LGTM but why can changes here impact accuracy?

I had removed the SoftwareBF16 pass from the runner pipeline since it is not upstream. So I was getting unknown symbols so I added the mlir_float16_utils to cover, but that gave inaccurate conversion.

PR #1133, opened 2023-06-28
enable rocMLIR on Windows

apwojcik:
The UIF2.0 team is working on bringing parts of ROCm to Windows. This PR is to enable rocMLIR compilation on Windows. The team uses the Microsoft Visual Studio C++ 2022 toolchain to build the library.


mlir/tools/rocmlir-lib/CMakeLists.txt
```diff
@@ -45,57 +47,344 @@ if(BUILD_FAT_LIBROCKCOMPILER)
   set(PACKAGE_NAME ${CMAKE_PROJECT_NAME_LOWER})
 
   set(LIBRARY_NAME rockCompiler)
-  function(combine_archives output_archive)
-    set(mri_file ${CMAKE_CURRENT_BINARY_DIR}/${output_archive}.mri)
-    set(full_output_path ${ROCMLIR_LIB_DIR}/lib${output_archive}.a)
-    set(output_archive_dummy_file ./${output_archive}.dummy.cpp)
+  set(COMPONENT_NAME ${LIBRARY_NAME})
+
+  set(__rocmlir_targets
```
krzysz00: I'm still extremely dubious about this - what, exactly, are we gaining from hard-coding the list of transitive dependencies like this? 

If we're building static libraries that aren't needed, could you point out which ones, so we can get them out of the dependency tree?

Otherwise ... what's this solving?

CMakeLists.txt
```diff
@@ -9,6 +9,8 @@ if(POLICY CMP0116)
   cmake_policy(SET CMP0116 OLD)
 endif()
 
+set(CMAKE_POSITION_INDEPENDENT_CODE ON CACHE BOOL "")
```
krzysz00, line 12: We might want to explicitly set `LLVM_ENABLE_PIC` on just to be safe?

CMakeLists.txt
```diff
@@ -37,28 +39,30 @@ set(ROCMLIR_LIB_DIR "${CMAKE_CURRENT_BINARY_DIR}/lib" CACHE PATH "")
 message(STATUS "ROCMLIR_BIN_DIR: ${ROCMLIR_BIN_DIR}")
 
 if( ROCMLIR_ENABLE_BENCHMARKS STREQUAL "all" )
-  set(ROCMLIR_ENABLE_BENCHMARKS "ck;rocblas")
+  set(ROCMLIR_ENABLE_BENCHMARKS CACHE STRING "ck;rocblas" FORCE)
```
krzysz00, line 42: Nice catch!

cmake/llvm-project.cmake
```diff
@@ -27,6 +27,9 @@ set(LLVM_BUILD_EXAMPLES ON CACHE BOOL "")
 set(LLVM_INSTALL_UTILS ON CACHE BOOL "")
 set(LLVM_ENABLE_TERMINFO OFF CACHE BOOL "")
 set(LLVM_ENABLE_ASSERTIONS ON CACHE BOOL "")
+if(MSVC)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_SILENCE_NONFLOATING_COMPLEX_DEPRECATION_WARNING" CACHE INTERNAL "")
```
krzysz00, line 31: What's requiring this? Is there some upstream bug here? Is there a better way to do this, or a more targeted way?

krzysz00 (commented):
Overall, very reasonable changes, I'd be willing to merge just about all of these ... except the changes to how we link libRockCompiler. I'd really like more information on that.

cmake/llvm-project.cmake
apwojcik: According to the C++ ISO spec, §26.2/2, the effect of instantiation of the template `complex` for any type other than `float`, `double`, or `long double` is unspecified. Clang, GNU, and MSVC provide template specializations for those types and provide the template itself. However, the MSVC warns about using the template, not specializations, and mentions that it is deprecated. To silence the warning, one may use `_SILENCE_NONFLOATING_COMPLEX_DEPRECATION_WARNING`.
Here we may safely turn off the warnings because the data type used is a class for representing a floating point number, and the class provides all required operations for the `complex` template to work correctly.

mlir/tools/rocmlir-lib/CMakeLists.txt
apwojcik: I changed the hardcoded list of files to `file(GLOB ...)`. However, please consider the warning given by the authors of CMake.

> We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. The CONFIGURE_DEPENDS flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if CONFIGURE_DEPENDS works reliably, there is still a cost to perform the check on every rebuild.

Source: https://cmake.org/cmake/help/v3.26/command/file.html#glob
The same applies to globbing using a shell from CMake. That is even worse because CMake does not know we are doing such a thing.

cmake/llvm-project.cmake
krzysz00: Per discussion, we can safely put this here and then try to make it an upstream patch eventually.

CMakeLists.txt
apwojcik: I added this because MIOpen on Windows complained some symbols could not be relocated. The symbols were from rocMLIR, not from LLVM. If you insist, I will add the flag `LLVM_ENABLE_PIC` to `cmake\llvm-project.cmake`.

krzysz00:
I'm planning to do https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/995 next week (once I'm back from the long weekend, so Wednesday, but you're welcome to do this earlier if you want) since, as far as I can tell, this'll clean up our builds and prevent the generation of static libraries that we don't depend on. I don't know if that'll be enough to slide under the 4 GB limit, but I thought I'd point this out.

( https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/996 will also happen Eventually(tm) )

I suspect this isn't too relevant in the short term, since it looks like the INTERFACE approach works pretty well too, but I thought I'd mention it.

CMakeLists.txt
krzysz00: It's default on, so yeah, this shouldn't be a big deal.

krzysz00 (approved):
Looks good to me, though it'd be nice if you could wait for #1139 to land first so as to make this easier to revert.

PR #1134, opened 2023-06-28
Add outer indices to threadwise write/read ops

manupak:
This commit adds support for threadwise write/read ops to accept a additional coordinates that is already determined by the client op that creates them.

E.g. in a tiled gemm implementation the loaded values will be viewed as (kOuter, bid, tid, iter). However, kOuter will be iterated by the outerloop in a thread. We can use threadwise_write_all within the loop where kOuter, bid, and iter as an extra indices passed onto threadwise_write_all.

closes : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1009


mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -858,14 +872,23 @@ LogicalResult ThreadwiseWriteAllRewritePattern::matchAndRewrite(
...
 
-  SmallVector<Value, 3> writeStartCoords{tid, zero};
-  SmallVector<int64_t, 3> bounds{1, iterLen};
-  SmallVector<int64_t, 3> strides{1, vectorLen};
+  SmallVector<Value, 3> writeStartCoords;
+  SmallVector<int64_t, 3> bounds;
+  SmallVector<int64_t, 3> strides;
+  ValueRange extraIndices = op.getExtraIndices();
+  for (Value extraIdx : extraIndices) {
+    writeStartCoords.push_back(extraIdx);
```
krzysz00: Nit: There are constructors for both copying in a range and producing a vector of N copies of something.

krzysz00 (commented):
As a general thought, can we actually make the block ID one of these extra indices instead of making it implicit? This'll also simplify @giuseros's layout refactor, and make threadwise_write_all less special-case-y.

manupak:
Hmmm.... why not 'tid' too ? The lowering doesn't do anything special as compared to 'bid'.
I mean if the client is creating/passing the views, they might give the 'bid', 'tid' too and let the op handle 'iter'

krzysz00:
.. yeah, passing in `tid` isn't a bad call.

manupak:
ok, I've generalized it now. PTAL.

manupak:
A gentle ping for reviews!

krzysz00 (approved):
Terribly sorry for forgetting to review, great cleanup, approved!

manupak:
Thanks @krzysz00!

PR #1136, opened 2023-06-30
Add API for quick tuning, otherwise improve the MIGraphX API

krzysz00:
Note that this PR contains breaking changes (though, generally, minor ones) to MIGraphX, and the version define in the headers has been updated accordingly.

To summarize:
1. The tuning API has been split so that the full and quick tuning spaces are now separate objects with their own indexing, instead of having quick be a subset of full (which assumed that we'd have a sorting metric for the tuning space, which we don't). Therefore, mlirRockTuningParamGet has been renamed to mlirRockTuningParamGetFull and there is a mlirRockTuningParamGetQuick
2. The quick tuning space exists and is populated using our heuristic-mode logic, which uses a hard-coded list of configs per type and sorts them to prioritize kernels that don't require padding.
3. Both tuning spaces now filter out performance configurations that cannot result in valid kernels, but do **not** filter out every such config (they only remove those whose invalidity is clear from the parameters)
4. A new function has been added to construct the applicability pipeline, which is used to detect invalid performance configurations. When tuning, this should be run after applying a performance configuration to determine if that configuration is valid.
5. Various `int` parameters have been updated to more meaningful types.

Internal details for us:
1. The various PopulateParams classes have grown an isParamArtrPlausible() method that runs the AffixTuningParameters checks on a given perf config. This allows us to filter the tuning space.
2. The "prioritize paddilng-less kernels" sorter has now been made a public API so the tuning space constructrors can use it.
3. The InitParams* -> tuning parameters attribute constructor is now exposed for the non-xdlops case as well.
4. The tuning loop now iterates over these InitParams* structs, which are direct reflections of the perf config string.
5. Some of the C API tests have been updated, expaded, and, in one case, fixed.


krzysz00:
MIGraphX folks, y'all're reviewers on this in order to detect reasons this API sucks.

manupak (approved):
Given the fact it is a sunk cost that we have the applicability pipeline (in a different universe, we could've done it via a partial lowering and op verifiers when constructing the tuning space but agreeing on that to be larger task altogether), I dont have any issues of using it now.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -151,20 +188,29 @@ TunableParams *createTunableParamSpace(ModuleOp &mod) {
...
+    return false;
+  paramEntry->param = tuningSpace->tuningRangeFull[pos];
+  return true;
+}
+
+bool tuningGetParamQuick(TunableParams *tuningSpace, unsigned pos,
+                         ParamEntry *paramEntry) {
   // out of bound check.
-  if (pos < 0 || (unsigned int)pos > tuningSpace->tuningRange.size() - 1)
+  if (pos > tuningSpace->tuningRangeFull.size() - 1)
```
manupak: This seems not right...

manupak (commented):
There may be a typo...

I think what you've done additionally is passing a buffer to be filled with problemStr? 
which is good than piggybacking on the tuning table.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
krzysz00: The "buffer to be filled" thing was @pfultz2's request - I was going to go for `char **` arguments so we `malloc()` and return the buffer, but he was wanting to avoid the extra copy

krzysz00: ... right, comment on the wrong thing.

This was a typo, fixed

manupak (approved):
I dont see issues here....

mlir/include/mlir-c/Dialect/MIGraphX.h
```diff
@@ -38,12 +38,27 @@ MLIR_CAPI_EXPORTED void mlirGetKernelAttrs(MlirModule module, uint32_t *attrs);
...
 MLIR_CAPI_EXPORTED void mlirMIGraphXAddHighLevelPipeline(MlirPassManager pm);
 
+/// Adds the pipeline that checks if the kernel with a given tuning
+/// configuration will actually compile to the pass manager. If this pipeline
+/// fails, it's not a meaningful error. The input to this should have been run
+/// through the high-level pipeline. This pipeline is only needed when tuning,
+/// and should, ideally, be called on a clone of the results of teh highlevel
+/// pipeline.
+MLIR_CAPI_EXPORTED void
+mlirMIGraphXAddApplicabilityPipeline(MlirPassManager pm);
```
pfultz2, line 58: Exactly how this supposed to be called? Wouldnt we add this always?

krzysz00: The applicability pipeline is a subset of the backend pipeline (that is, `backend(applicability(x)) == backend(x)`) such that failures during the applicability portion of the pipeline are failures of the request (ex. the tuning parameter ultimately made us use too much LDS) and not errors that need to be reported.

PR #1137, opened 2023-07-05
Make packing of LDS store reg buffer accept a view

manupak:
Currently, the function packLoadBufferToStoreBuffer takes an assumption of the layout of the load buffer.
However, for fused attention, the load buffer for second gemm needs to have a flexible layout.
Therefore this commit changes the code, removing the assumption and make the packing accept a view of the load buffer.

closes : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1010


krzysz00 (commented):
Why do we need this change? Is it just a cleanup or?

manupak:
While it could be view as a cleanup, it is required for me to pass in a view of the output of the first gemm to subsequent ops of attention -- namely blockwise_broadcast_reduce and the second gemm.

krzysz00 (approved):
Approved because I don't nee any actual issue here

PR #1138, opened 2023-07-05
Enhance outputViews of accelEmitter

manupak:
Currently accelEmitter's computeOutputTransforms API always return the grid-level view of the output.

This commit changes the API to be able produce grid-level, block-level and thread-level view.

closes : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1011


krzysz00 (commented):
I'd argue that the better approach here would be to pass `0` in for the block size instead of doing all this special casing

manupak:
I need views that top-level takes (tid, iter) and (iter) to be passed into blockwise_broadcast_reduce and rock.transform ops to create views for subsequent ops following the the first gemm in attention.

Though, it looks special casey (I've made sure to narrow it down where the changes are...) this would look much pleasing to the eyes when looking at the IR (i.e. transform chains).

krzysz00:
Re having views that take `(tid, iter)` or `(iter)` as the top level ... why not `constDim()` it?

krzysz00:
I'm also seeing the issue that the views don't always start at `d0` and that could be a problem.

manupak:
> Re having views that take `(tid, iter)` or `(iter)` as the top level ... why not `constDim()` it?

It would technically work, in fact this is what I was doing in blockwise_broadcast_reductions, but in that discussion I thought we agreed to change the outputView emitter to conform to generate modular views -- gridwise, blockwise and threadwise.

I still would like to see leaner IR after this stage rather than carrying on a derivation from constDim dimension from the top.

> I'm also seeing the issue that the views don't always start at d0 and that could be a problem.

I dont follow what this means. The objective here will have different start dims -- namely
1) gridwise - bid, tid, iter
2) blockwise - tid, iter
3) threadwise - iter.

Then in the view creation logic, it will strip out any coordinate transform that is dependent on dimension that is not present.

In this version, I opted to reduce code duplication -- as a result needed special cases in the unified view creation function to exlcude/include certain transformations.

An alternate method would be to do code duplication and do seperate functions : 
1) computeGridwiseOutputTransforms 
2) computeBlockwiseOutputTransforms 
3) computeThreadwiseOutputTransforms

krzysz00:
You're right re previous discussion, and so given that we can stick with this approach.

The minor nitpick I'm having is that, when you give lower positions (say when you split up `iter`), you end up creating coordinate transforms that don't define the 0th, 1st, ... Kth output index, because you start at K (I think an example I saw yesterday starts at 6).

I don't remember off the top of my head if this trips an assert or if 0s get filled in when you do something like that, but I thought I'd point it out.

manupak:
The way I understood it is as follows : 
1) it starts with upper coords (bid?, tid?, iter)
2) then does bunch of split/merge (depending on how you look at it)
3) finally does an embed to construct a coordinate in [M, N] space.

So if some upper coords are absent, it may have holes -- that I agree.

> toMatrixC.embed("gemmN", 2, matrixN,
                  {"n", "wave_n", "n_i", "blk_col", "n_tid"},
                  {nPerBlock, nPerAccel, nPerAccel * nWaves, n, 1});

I think this is the problematic bit you are pointing out where tid crosses a line.

In retrospect, do we need a embed here ? can we do a merge here which is simpler and will work nicely with dimension removals ?
The dim sizes are there -- its just shifted to left to make it an embed -- we need mblocks and nblocks additionally

krzysz00:
The problematic bit is the `{6, 7, 8, 9}` in `{6, 7, 8, 9}, "item",`, which, in an `item`-only map, would leave the first 6 output dimensions unfilled.

And I think it was an `Embed` for weird historical reasons/not fully understanding what all the coefficients were. I *think* we should be able to refactor it into an `Unmerge` if we wanted to, but maybe that's a separate PR.

manupak:
So the top-level args maps to low level dims without jumping the queue in the most cases.

i.e. 

[bid, tid, item] -->    [ {some dims from transforming bid} , {some dims from transforming tid} , {some dims from transforming item} ]

So if one top-level arg is stripped out --> it would mean a top-level dims of low-level view to be stripped out.
Also in this embed, the right side co-effs are unaffected by what happens to the left-side coeffs because right-most is the fastest changing dimension (in the view).

So above work for most cases.
Except where : 

> toMatrixC.embed("gemmN", 2, matrixN,
{"n", "wave_n", "n_i", "blk_col", "n_tid"},
{nPerBlock, nPerAccel, nPerAccel * nWaves, n, 1});

where n_tid is inserted as the fastest changing dim in low-level view.
Therefore to handle that, now I've added a division by "n" for all the item coeffs.

I've also added low-level dim inference rather than straight away plugging the M, N dimension size of the original matrix.

manupak:
(I need to correct wmma side of things; which I haven't paid enough attention if thats what you are referring to)

manupak:
alright -- Im figuring out embed ordering is not strictly decreasing strides as I was assuming before... sigh.

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -209,19 +226,22 @@ ArrayAttr MfmaEmitter::computeOutputTransforms(PatternRewriter &b, Location loc,
...
+      createTopSplitTMBuilder(b, loc, numElements, gridSize, blockSize);
+  if (gridSize.has_value()) {
+    splitMemoryCoords.merge(
+        {"g", "m", "n"}, {0, 1, 2}, {"bid"},
+        {gridSize.value() / gStride, gStride / nBlocks, nBlocks});
+  }
+  if (blockSize.has_value()) {
+    int64_t wavesInKernelBlock = blockSize.value() / waveSize;
+    splitMemoryCoords.merge(
+        {"wave", "m_tid", "n_tid"}, {3, 4, 5}, "tid",
```
krzysz00: Let me be clear: the problem I'm seeing is that the indices `{3, 4, 5}` should not be `{3, 4, 5}`, they should be `{numGridDims, numGridDims + 1, numGridDims + 2}`, where `numGridDims == gridSize.has_value() ? 3 : 0`

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -472,17 +544,38 @@ ArrayAttr WmmaEmitter::computeOutputTransforms(PatternRewriter &b, Location loc,
...
+    SmallVector<int64_t, 5> embedDimCoeffsM;
+    if (gridSize.has_value()) {
+      embedDimNamesM.push_back("m");
+      embedDimCoeffsM.push_back(mPerBlock);
+    }
+    if (blockSize.has_value()) {
+      embedDimNamesM.insert(embedDimNamesM.end(), {"wave_m", "m_tid"});
+      embedDimCoeffsM.insert(embedDimCoeffsM.end(), {wmmaInsn.inputLen, 1});
+    }
+    embedDimNamesM.insert(embedDimNamesM.end(), {"rep_i", "item_i"});
```
krzysz00: This use of `insert()` is `append()`

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
manupak: ah right got it..

manupak:
Ive fixed all that now;

krzysz00 (commented):
Overall, I'm liking the approach.

However, this whole dimension removal thing could be massively simplified to (since you're always removing from the beginning), to `ArrayRef::drop_front()` (or there might just be an `ArrayRef::slice()`)

So we can save on the copies and such there.

manupak:
> (since you're always removing from the beginning)

Unfortunately no...
See n_tid of mfma emitter otherwise I would've thought so.

Also, I learned yesterday embed is order insensitive (well unless you sort it); 
There can be larger strides towards right.

krzysz00:
If the order of the Embed parameters is the issue, feel free to permute them to make the drop_front work

manupak:
Its both: 
1) we need to drop right/back dims (e.g. ntid) -- in essence producing a threadwise stride slice view of the output
2) coeffs do not have ordering restrictions.

I think due to 1) 'just' dropping front would not work.

krzysz00:
The point is, that the Embed coefficients can be in *any* order, so all you need to do is to put the ones you'll want to drop at the front.

Just rearrange the coefficients and their names.

krzysz00:
1) That is, move "ntid" and its coefficient up earlier

manupak:
> that the Embed coefficients can be in any order

This I agree

> so all you need to do is to put the ones you'll want to drop at the front.

This I dont; we can only do the drop in this fashion if they are larger coefficients than what is remaining after the drop.

> That is, move "ntid" and its coefficient up earlier

ntid's coefficient is 1. when I drop this I still want a 'dense' memref which is a strided slice of the gemm output.

manupak:
OK, Ive reordered the dims in the decreasing manner and removed the embeds altogether.

The embeds are making this hard to follow...

manupak:
PTAL; 

I think unmerges are making this much simpler to follow...

Though, I can split (and semi-duplicate) the code for blockwise and threadwise case, if thats needed. However, I would appreciate to know if thats the last thing to be done here, if possible.

krzysz00 (approved):
Nice utility, this looks pretty good and makes sense, approved

And my offhand comments earlier re factoring out the to-registers view into a second function aren't strictly needed but might be worth thinking about. That is, I'm not sure it's possible to define an accelerator map that both means something and doesn't take the thread ID as an argument. I'd like to understand what the usecase for that map is, but that doesn't hold up the PR.

manupak:
Thanks @krzysz00! 

It is mostly for (softmax or otherwise) partial normalizations that are rank-sensitive but not ordering within dimensions. 
i.e. I would still like to see a matrix with same dimensions but is a (strided) subset of the larger one.

PR #1145, opened 2023-07-11
Add a super-grouped grid layout transformation

giuseros:
This PR is aiming at imposing an layout on the distribution of the workgroups, aiming at minimize cache trashing. The idea is the same of: https://triton-lang.org/main/getting-started/tutorials/03-matrix-multiplication.html#sphx-glr-getting-started-tutorials-03-matrix-multiplication-py

The main changes are:
- I added a transformation to map `bid` -> `g_block, m_block, n_block` to mimic what Triton does. Since we cannot have a piecewise affine transformation, I had to manually insert the transformation directly in MLIR. This is not super elegant and might be corrected in followup PR. 
- The group size is calculated via a simple heuristic on the number of CUs (`groupSize=sqrt(numCU)`). For now we don't add any additional tuning parameter
- `numCU` is now mandatory in gridwise gemm and it is used as part of the problem description (i.e., `numCU` will be part of the key in the tuning db). This means I had to change `perfRunner.py` and `tuningRunner.py`.


mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -213,11 +215,10 @@ ArrayAttr MfmaEmitter::computeOutputTransforms(PatternRewriter &b, Location loc,
...
   int64_t retNumElements = accVectorType.getNumElements();
   int64_t numElements = retNumElements * mRepeats * nRepeats * nResultVectors;
-  TopDownTMBuilder splitMemoryCoords(b, {"bid", "tid", "item"},
-                                     {gridSize, blockSize, numElements}, loc);
-
-  splitMemoryCoords.merge({"g", "m", "n"}, {0, 1, 2}, {"bid"},
-                          {gridSize / gStride, gStride / nBlocks, nBlocks});
+  TopDownTMBuilder splitMemoryCoords(
+      b, {"g_block", "m_block", "n_block", "tid", "item"},
+      {gridSize / gStride, mBlocks, nBlocks, blockSize, numElements}, loc);
```
manupak: Can we not re-calculate grid coordinates here from gridSize ? 
The caller of computeOutputTransforms should already have bidGridLengths.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -766,9 +776,20 @@ LogicalResult ThreadwiseReadIntoRewritePattern::matchAndRewrite(
...
   SmallVector<int64_t, 3> strides{1, vectorLen};
   if (srcAddrSpace == gpu::AddressSpace::Global) {
-    readStartCoords.insert(readStartCoords.begin(), bid);
-    bounds.insert(bounds.begin(), 1);
-    strides.insert(strides.begin(), 1);
+    SmallVector<Value> blockCoords;
+    if (inputShape.size() == 3) {
+      blockCoords = SmallVector<Value>{bid};
+    } else {
+      int64_t numCU = 110;
```
manupak: We dont need to this after https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1134.
i.e. we dont need this logic recalculated at this level in the lowering.

Please have a look at the PR and I think could simply pass in Values : g_block, m_block and n_block when threadwise_read / threadwise_write ops are created.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -863,9 +884,20 @@ LogicalResult ThreadwiseWriteAllRewritePattern::matchAndRewrite(
...
   SmallVector<int64_t, 3> strides{1, vectorLen};
   if (dstAddrSpace == gpu::AddressSpace::Global) {
-    writeStartCoords.insert(writeStartCoords.begin(), bid);
-    bounds.insert(bounds.begin(), 1);
-    strides.insert(strides.begin(), 1);
+    SmallVector<Value> blockCoords;
+    if (outputShape.size() == 3) {
+      blockCoords = SmallVector<Value>{bid};
+    } else {
+      int64_t numCU = 110;
```
manupak: Same here... (see the above comment)

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -376,13 +376,17 @@ createGlobalLoadLoop(PatternRewriter &b, Location loc, GpuAllocOp loadBuffer,
...
   Value zero = b.createOrFold<arith::ConstantIndexOp>(loc, 0);
 
-  SmallVector<Value, 4> globalStart = {zero, bid, tid, zero};
-  SmallVector<Value, 4> vectorStartOuter(4, zero);
+  Value g_block, m_block, n_block;
+  std::tie(g_block, m_block, n_block) =
+      layout::gridLayout(b, loc, bid, upperBounds[2], upperBounds[3], numCU);
+  SmallVector<Value, 4> globalStart = {zero,    g_block, m_block,
+                                       n_block, tid,     zero};
+  SmallVector<Value, 4> vectorStartOuter(6, zero);
```
manupak, line 380: This could all just be a threadwise_read_into... 
i.e. no need to repeat the logic

mlir/lib/Dialect/Rock/Transforms/LayoutEmitter.cpp
```diff
@@ -0,0 +1,80 @@
+//===- AccelEmitter.cpp - MLIR helper to emit acceleration intrinsics
```
manupak: This is not right...

mlir/lib/Dialect/Rock/Transforms/LayoutEmitter.cpp
```diff
@@ -0,0 +1,80 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// =============================================================================
+//
+// This class tries to abstract away the code-generation details needed to
+// generated calls to matrix multiply accelerator intrinsics (wmma, mfma).
```
manupak: As this is not a transform pass, IMO this file should live in utility? 
(same could be said for AccelEmitter)

manupak (dismissed):
The intent and outcome is sound! I agree with the direction.

I just have concerns with layering.
I am doing some cleanup/improvement on parallel to increase modularity/extensibility of the codebase that could be useful here. 

Please have a look at the PR : 
1) https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1134

With the above, you shouldn't need to redo the logic restore block id in the lowering of threadwise_read_into / write_all ops. I believe you already have them when those ops are created .. therefore, you just pass them as "extraIndices" field.

2) AccelEmitter should not be re-calculating grid layout back from gridSize. Im also doing something here : https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1138. One of the changes here is to pass the bidGridLengths to accelEmitter. I believe with that you dont need to re-do g_block, m_block and n_block extractions again in AccelEmitter.

So generally, my comments are to push to a singular place where this change should happen and others should 'just' work(TM) :) .

mlir/lib/Dialect/Rock/Transforms/LayoutEmitter.cpp
krzysz00: I think Transforms/ isn't required to just be passes, it can also have pass-specific support code

manupak: which I always thought why Transforms/Utils/ or Utils/ exist... (but yea it is not strictly followed upstream either)
Anyways I dont have a strong opinion; just felt nice to have an criterion to decide which goes where.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -362,10 +360,12 @@ static TransformingForOp
 createGlobalLoadLoop(PatternRewriter &b, Location loc, GpuAllocOp loadBuffer,
                      Value wrappedMatrix, ArrayAttr vectorMap,
                      int64_t dataPerThread, int64_t vectorLen, Value bid,
```
manupak, line 362: Instead of bid here,  lets pass in the g_block val, m_block val and n_block val if this function is being kept.
I still think we can replace the whole function with a threadwise_read_jnto.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1388,9 +1391,9 @@ struct GridwiseGemmAccelRewritePattern
         b, loc, M, N, blockSize, gridSize, regCAllocOp, convertedC,
         forceUnroll);
 
-    b.create<ThreadwiseWriteAllOp>(loc, registerC, op.getC(), idToMatrixCMaps,
-                                   op.getFeatures(), op.getStoreMethod(),
-                                   forceUnroll, useIndexDiffs);
+    b.create<ThreadwiseWriteAllOp>(
+        loc, registerC, op.getC(), idToMatrixCMaps, op.getFeatures(),
```
manupak: you'll need to pass in extraIndices here as  = {g_block, m_block, n_block, tid} to match the top dims of idToMatrixCMaps

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -116,8 +115,7 @@ def Rock_Conv2DBwdDataOp : Rock_Conv2DOpBase<"conv2d_bwd_data">
 def Rock_Conv2DBwdWeightOp : Rock_Conv2DOpBase<"conv2d_bwd_weight">
 {
   dag additionalArgs = (ins Optional<TensorOrMemRefRankOf<[F32], [5]>>:$workspace,
-                            OptionalAttr<IndexAttr>:$kBlocks,
-                            I32Attr:$numCu);
+                            OptionalAttr<IndexAttr>:$kBlocks, I32Attr:$numCU);
```
manupak, line 118: why not OptionalAttr on numCU here ?

manupak (commented):
few further pointers/comments...

mlir/lib/Dialect/Rock/Transforms/LayoutEmitter.cpp
giuseros: Yeah I don't know, I mean my (also not strong) preference is more toward a flat structure. I guess if we reach a point where we have a lot of utility code, then we can move everything in a separate folder

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
giuseros: This was like this because the `numCU` is mandatory in backward weight convolution (because it is needed to compute the `KBlock` and the workspace)

giuseros:
Hi @manupak , I rebased and restructured the code like you proposed, and I agree it's way cleaner. Thanks!

mlir/include/mlir/Dialect/Rock/IR/RockGemmWrapperInterface.td
```diff
@@ -102,6 +102,16 @@ def RockGemmWrapperInterface : OpInterface<"RockGemmWrapperInterface"> {
           return $_op.getFeatures();
         }]
       >,
+    InterfaceMethod<
+        /*desc=*/[{
+          Return the features attribute of this op.
```
krzysz00: Failed to update comment

mlir/lib/Dialect/Rock/Transforms/LayoutEmitter.cpp
```diff
@@ -0,0 +1,79 @@
...
+//
+//
+//===----------------------------------------------------------------------===//
+#include "mlir/Dialect/Rock/IR/Rock.h"
+#include "mlir/Dialect/Rock/IR/TransformMapBuilder.h"
+#include "mlir/Dialect/Rock/Passes.h"
```
krzysz00: What's this include for?

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -16,19 +16,21 @@
...
+              /*hasFp8ConversionInstrs=*/true),
+    rdnaNoDotInfo(GemmFeatures::atomic_fmax_f32, /*waveSize=*/32, 80,
                   /*hasFp8ConversionInstrs=*/false),
     rdnaInfo(GemmFeatures::dot | GemmFeatures::atomic_fmax_f32,
-             /*waveSize=*/32, /*hasFp8ConversionInstrs=*/false),
+             /*waveSize=*/32, 80, /*hasFp8ConversionInstrs=*/false),
     gfx11Info(GemmFeatures::dot | GemmFeatures::atomic_add |
                   GemmFeatures::atomic_fmax_f32 | GemmFeatures::wmma,
-              /*waveSize=*/32, /*hasFp8ConversionInstrs=*/false);
+              /*waveSize=*/32, 96, /*hasFp8ConversionInstrs=*/false);
```
krzysz00: Can we get the little informative comments uniformly?

mlir/utils/performance/perfRunner.py
```diff
@@ -707,6 +713,7 @@ def runConfigWithMLIR(config: PerfConfiguration, paths: Paths, rocmlir_gen_flags
     if debug:
         print("Running MLIR Benchmark: ", repr(config))
     rocmlirGenCommand = paths.mlir_paths.rocmlir_gen_path + ' -ph ' + commandLineOptions
+    print(rocmlirGenCommand)
```
krzysz00: Stray debug print

krzysz00 (commented):
Minor nitpicks

mlir/lib/Dialect/Rock/Transforms/LayoutEmitter.cpp
```diff
@@ -0,0 +1,64 @@
+//===- LayoutEmitter.cpp - MLIR helper that contains the layout logic -===//
```
manupak: nit (naming) : this could be GridLayoutEmitter as it not a generic layout emitter

mlir/lib/Dialect/Rock/Transforms/LayoutEmitter.cpp
```diff
@@ -0,0 +1,64 @@
...
+//===----------------------------------------------------------------------===//
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Rock/utility/AmdArchDb.h"
+#include "mlir/Dialect/Rock/utility/loweringUtils.h"
+
+#include "llvm/Support/Debug.h"
+
+#include "LayoutEmitter.h"
+
+#define DEBUG_TYPE "layout-transform"
```
manupak: nit (naming) : "rock-grid-layout-emitter" using this in CLI with -debug-only="layout-transform" might be a little too vague.

mlir/lib/Dialect/Rock/Transforms/LayoutEmitter.h
```diff
@@ -0,0 +1,54 @@
...
+  Value g_block;
+  Value m_block;
+  Value n_block;
+};
+
+/// This function emits the right triplet of <group,block_m,block_n> identifers,
+/// given a flat blockId. This has been adapted from:
+/// https://triton-lang.org/main/getting-started/tutorials/03-matrix-multiplication.html#sphx-glr-getting-started-tutorials-03-matrix-multiplication-py
+///
+GridCoordinates gridLayout(PatternRewriter &b, Location loc, Value bid,
```
manupak: nit (naming) : this could be "makeGroupedGridLayout" just because we might end up different ways here

manupak (commented):
Looking good! few naming nit picks.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
krzysz00: It's only needed for atomic_add kernels, though, so it's kinda optional still. On the other hand, MIOpen integration is off so _shrug_.

krzysz00 (commented):
The code here seems fine to me.

Do we have perf numbers?

giuseros:
> The code here seems fine to me.
> 
> Do we have perf numbers?

Yes, they are coming.  Need to collect base results and then we can compare

manupak (commented):
One doc fix that you prolly missed otherwise looks good on the condition perf agree to numCU based (full occupancy) heuristic. 

I ll wait for perf before hitting approve.

mlir/lib/Dialect/Rock/Transforms/GridLayoutEmitter.h
```diff
@@ -0,0 +1,55 @@
+//===- LayoutEmitter.h - MLIR helper that contains the layout logic -===//
```
manupak: I think you missed this.

giuseros:
As I mentioned in the meeting, I have had an issue with `i8xi8->i32` performance. 

Basically, the fact that threads were loading i8 and storing i32 were messing up with the cache (I think). I added a sort of normalization factor based on the ratio `outBitwidth/inBitwidth` that sorted out the issue. I am attaching here the comparison:

[perf.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/12074095/perf.zip)

giuseros:
A friendly reminder @manupak  @krzysz00 :)

mlir/lib/Dialect/Rock/Transforms/GridLayoutEmitter.cpp
```diff
@@ -0,0 +1,71 @@
...
+  // This also covers the cases where the output width is larger
+  // than the input width
+  int64_t bitWidthIn = info.inputType.getIntOrFloatBitWidth();
+  int64_t bitWidthOut = info.outputType.getIntOrFloatBitWidth();
+  int64_t groupSize =
+      std::ceil(std::sqrt(info.numCU)) * (bitWidthOut / bitWidthIn);
+
+  Value mBlocksPerGroup = b.createOrFold<ConstantIndexOp>(loc, groupSize);
+  Value blocksPerGroup =
+      b.createOrFold<ConstantIndexOp>(loc, groupSize * info.nBlocks);
```
manupak, line 52: Im just realizing we do 1-d tiling here... why ?

manupak: and this is not equivalent to 

```py
# Number of programs ids along the N axis
num_pid_n = tl.cdiv(N, BLOCK_SIZE_N)
# Number of programs in group
num_pid_in_group = GROUP_SIZE_M * num_pid_n
```

where num_pid_n is value of current n_block rather than n_blocks right ?

manupak (commented):
Sorry (didnt catch this early), I have last min question about the logic...

mlir/lib/Dialect/Rock/Transforms/GridLayoutEmitter.cpp
manupak: oops, yes it is

manupak:
(it was my bad)

manupak (approved):
Codewise approved!
perf-wise I can see -6.94 to +10 swing.. 

do you have any idea for the negative ones ?

giuseros:
My guess would be that the heuristic we are using is too simple. I think this PR will enable us to do this sort of investigation, hoping not only to sort out the regressions you mention, but also to improve performance further.

giuseros:
I created an investigation ticket here: https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1022

PR #1148, opened 2023-07-11
[mlir][SerializeToHsaco] Initialize LLVM exactly once.

krzysz00:
No matter how one constructs their SerializeToHsaco pass, we want to ensure that the LLVM initialization code runs once and only once. This commit adds a static once_flag to ensure that.

I have run into issues when doing parallel GPU compiles where LLVM options were added more than once, and this may be part of the issue.


sjw36 (approved):
Looks good.

PR #1151, opened 2023-07-14
Add the codepath navi3x to Jenkinsfile. Currently navi3x is enabled only in PR CI

yiqian1:


krzysz00 (commented):
(There's a weird method too large error we might want to look in to, not sure if it's related)

As a high-level question, how many gfx1100 nodes do we have?

The change itself looks fine.

I claim that now we've got too many systems we're testing, and we might want to leave, for example, navi21 CI for nightly/weekly only.

yiqian1:
> (There's a weird method too large error we might want to look in to, not sure if it's related)
> 
> As a high-level question, how many gfx1100 nodes do we have?

We don't have gfx1100 nodes. We have two gfx1101 nodes.

krzysz00:
... hold on, can you make sure we actually have arch db entries for gfx1101? I think we do, but

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -297,7 +311,7 @@ pipeline {
 
         // choose the codepath for testing
         choice(name: 'codepath',
-               choices: ['default', 'mfma', 'navi21', 'vanilla'],
+               choices: ['default', 'mfma', 'navi21', 'navi3x', 'vanilla'],
                description: 'Choose the codepath to test')
         // option to disable navi21 cells in case nodes are offline
         booleanParam(name: 'disableNavi21', defaultValue: false,
```
jerryyin, line 317: Add an option to `disableNavi31`, and this PR should be ready to merge.

You may want to let the team know:
 - What does Navi3x node related issues look like
 - How do they manually kick a run skipping Navi3x nodes when that happens

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -190,15 +192,20 @@ def rebootNode() {
 void build_fixedE2ETests(String codepath) {
     // Limit the number of lit workers to 8 for navi21 codepath on nightly CI as a workaround for issue#702
     limit_lit_workers = false
+    // Filter out dense_output_bf16.mlir for navi32 codepath as a workaround for issue#1023
+    filter_out_tests = false
     if ( (codepath == 'navi21') && (params.nightly == true) ) {
```
krzysz00, line 197: Do you need `navi31` here as well?

krzysz00: (We might also want to drop this just being on nightly?)

krzysz00 (approved):
Seems reasonable, approved

mlir/utils/jenkins/Jenkinsfile
yiqian1: I'm dropping this test for both navi31 and navi32, although I could have done this only for navi32. The comment is indeed confusing. We will drop this for nightly as well.

PR #1152, opened 2023-07-14
[DO NOT SQUASH] Add support for noalias atributes.

stefankoncarevic:


giuseros:
Nice one! It looks fine to me, but I think you should give it a round of `clang-format` because the formatting looks a bit odd :)
Also, could you write a description to help the reviewers understand the aim of the PR (maybe also pointing out the ticket you are working on?)

external/llvm-project/mlir/lib/Conversion/GPUCommon/GPUOpsLowering.cpp
```diff
@@ -181,6 +186,30 @@ GPUFuncOpLowering::matchAndRewrite(gpu::GPUFuncOp gpuFuncOp, OpAdaptor adaptor,
...
+    auto memrefTy = en.value().dyn_cast<MemRefType>();
+    NamedAttrList argAttr = argAttrs
+                                ? argAttrs[en.index()].cast<DictionaryAttr>()
+                                : NamedAttrList();
+    if (argAttr.empty())
+      continue;
+
+    if (memrefTy && argAttr.erase(LLVM::LLVMDialect::getNoAliasAttrName())) {
+      auto remapping = signatureConversion.getInputMapping(en.index());
+      for (size_t i = 0; i < remapping->size; i++)
```
krzysz00: Nit: LLVM house style calls for `++i` and, even more paranoidly, `(size_t i = 0, e = remapping->size; i < e; ++i)`

external/llvm-project/mlir/lib/Conversion/GPUCommon/GPUOpsLowering.cpp
```diff
@@ -181,6 +186,30 @@ GPUFuncOpLowering::matchAndRewrite(gpu::GPUFuncOp gpuFuncOp, OpAdaptor adaptor,
...
+        if (llvmFuncOp.getArgument(remapping->inputNo + i)
+                .getType()
+                .template isa<LLVM::LLVMPointerType>()) {
+          llvmFuncOp.setArgAttr(remapping->inputNo + i,
+                                LLVM::LLVMDialect::getNoAliasAttrName(),
+                                rewriter.getUnitAttr());
+        }
+    }
+    assert(argAttr.empty() && "Only noalias attributes support");
```
krzysz00: I'd rather not assert here and just let argument attributes fall through if unhandled. It's what the behavior was before and this could crash someone's working code.

external/llvm-project/mlir/lib/Conversion/GPUCommon/GPUOpsLowering.cpp
```diff
@@ -181,6 +186,30 @@ GPUFuncOpLowering::matchAndRewrite(gpu::GPUFuncOp gpuFuncOp, OpAdaptor adaptor,
...
+  // pointer arguments.
+  for (const auto &en : llvm::enumerate(gpuFuncOp.getArgumentTypes())) {
+    auto memrefTy = en.value().dyn_cast<MemRefType>();
+    NamedAttrList argAttr = argAttrs
+                                ? argAttrs[en.index()].cast<DictionaryAttr>()
+                                : NamedAttrList();
+    if (argAttr.empty())
+      continue;
+
+    if (memrefTy && argAttr.erase(LLVM::LLVMDialect::getNoAliasAttrName())) {
```
krzysz00: Let's generalize this a bit:
1. Stick a lambda together that handles copying an arbitrary pointer attribute (`noalias`, `readonly`, `nonnull`, `dereferencable_or_null(N)` and so on)
2. Instead of creating a new attribute, save the one you `erase()`d so you can add it to the pointer arguments

external/llvm-project/mlir/lib/Conversion/GPUCommon/GPUOpsLowering.cpp
```diff
@@ -181,6 +186,30 @@ GPUFuncOpLowering::matchAndRewrite(gpu::GPUFuncOp gpuFuncOp, OpAdaptor adaptor,
...
+      auto remapping = signatureConversion.getInputMapping(en.index());
+      for (size_t i = 0; i < remapping->size; i++)
+        if (llvmFuncOp.getArgument(remapping->inputNo + i)
+                .getType()
+                .template isa<LLVM::LLVMPointerType>()) {
+          llvmFuncOp.setArgAttr(remapping->inputNo + i,
```
krzysz00: Correctness issue:
If `noalias` is present and the calling convention is not bare pointers (or, equivalently, if the remapping is of size > 1, currently), do **not** copy `noalias` and instead emit a warning. This is because memrefs, as you've probably noticed, have the allocation pointer and the aligned pointer, and those do, in fact, alias.

(In the future, we might want to add add an "aligned pointers only" scheme, but that's far future work we don't need)

mlir/unittests/Dialect/Rock/TransformMapBuilderTests.cpp
```diff
@@ -245,3 +250,28 @@ TEST_F(TMBuilderTest, GemmOut) {
                      &context));
   EXPECT_EQ(resDown, resUp);
 }
+
+TEST_F(TMBuilderTest, TestNoAliasAttributeForMemrefArgs) {
```
krzysz00: This is completely unneeded and not what you should've been changing.

krzysz00 (changes requested):
1. Correctness issue.
2. We might as well solve some of the general case while we're here
3. Add tests for the changes you just made to GPU function lowering:
You need to add a new file to `external/llvm-project/mlir/test/Conversion/GPUCommon/` . Use `lower-memory-space-attrs.mlir` in that directory as an example for how to set these tests up. Other tests in that directory or the surrounding ones may also be useful.

To be clear, I'm not upset, this is nonobvious stuff when you're new to the codebase.

external/llvm-project/mlir/lib/Conversion/GPUCommon/GPUOpsLowering.cpp
krzysz00: We don't actually have to do the general case, but it'll save everyone headache, I think.

krzysz00:
Final note: this commit touches both our code and upstream. We'll want to send the upstream part to LLVM. To make that easy you'll need to:
1. Stick a `[DO NOT SQUASH]` in the title of this PR to warn everyone.
2. When merging it, use a merge commit
3. Most importantly, have the final work (you don't need to do this immediately, and you can just push commits here and then `rebase -i` your way to the correct form once you're approved) split into two commits: one containing only changes to `external/llvm-project/` and one containing all other changes.

The external commit should have a longer commit message (explaining, at a high level, what this is doing and why), and its first line will be something like `[mlir][GPU] Handle LLVM pointer attributes on memref arguments` or `[mlir][GPU] Handle llvm.noalias on memref arguments` depending on which change we actually make.

external/llvm-project/mlir/lib/Conversion/GPUCommon/GPUOpsLowering.cpp
```diff
@@ -181,8 +186,46 @@ GPUFuncOpLowering::matchAndRewrite(gpu::GPUFuncOp gpuFuncOp, OpAdaptor adaptor,
...
+        argAttrs ? argAttrs[en.index()].cast<DictionaryAttr>() : NamedAttrList();
+
+  auto copyPointerAttribute = [&](StringRef attrName) {
+    Attribute attr = argAttr.erase(attrName);
+    if (!attr)
+      return;
+    auto remapping = signatureConversion.getInputMapping(en.index());
+    if (remapping->size > 1 &&
+        attrName == LLVM::LLVMDialect::getNoAliasAttrName()) {
+      mlir::emitWarning(
```
krzysz00: No need for a leading `mlir::`
Also, per upstream style, instead of `if {} else {}`ing, just `return;` after emitting the warning

external/llvm-project/mlir/lib/Conversion/GPUCommon/GPUOpsLowering.cpp
```diff
@@ -181,8 +186,46 @@ GPUFuncOpLowering::matchAndRewrite(gpu::GPUFuncOp gpuFuncOp, OpAdaptor adaptor,
...
+          llvmFuncOp.getLoc(),
+          "Warning: Cannot copy noalias with non-bare pointers.\n");
+    } else {
+      for (size_t i = 0, e = remapping->size; i < e; ++i) {
+        if (llvmFuncOp.getArgument(remapping->inputNo + i)
+                .getType()
+                .template isa<LLVM::LLVMPointerType>()) {
```
krzysz00: I don't think this needs a `template`

external/llvm-project/mlir/lib/Conversion/GPUCommon/GPUOpsLowering.cpp
```diff
@@ -181,8 +186,46 @@ GPUFuncOpLowering::matchAndRewrite(gpu::GPUFuncOp gpuFuncOp, OpAdaptor adaptor,
...
+    auto memrefTy = en.value().dyn_cast<MemRefType>();
+    NamedAttrList argAttr =
+        argAttrs ? argAttrs[en.index()].cast<DictionaryAttr>() : NamedAttrList();
+
+  auto copyPointerAttribute = [&](StringRef attrName) {
+    Attribute attr = argAttr.erase(attrName);
+    if (!attr)
+      return;
+    auto remapping = signatureConversion.getInputMapping(en.index());
+    if (remapping->size > 1 &&
```
krzysz00: Since that's not an obvious condition, comment that this is a proxy for the bare pointer calling convention.

external/llvm-project/mlir/lib/Conversion/GPUCommon/GPUOpsLowering.cpp
```diff
@@ -181,8 +186,46 @@ GPUFuncOpLowering::matchAndRewrite(gpu::GPUFuncOp gpuFuncOp, OpAdaptor adaptor,
...
+  };
+
+  if (argAttr.empty())
+    continue;
+
+  if (memrefTy) {
+    copyPointerAttribute(LLVM::LLVMDialect::getNoAliasAttrName());
+    copyPointerAttribute(LLVM::LLVMDialect::getReadonlyAttrName());
+    copyPointerAttribute(LLVM::LLVMDialect::getNonNullAttrName());
+    copyPointerAttribute(LLVM::LLVMDialect::getDereferenceableOrNullAttrName());
```
krzysz00: I think straight-up `Dereferencable` exists as well, could go here?

krzysz00 (commented):
The main code looks file (aside from some style nitpicks), this just needs a test.

external/llvm-project/mlir/test/Conversion/GPUCommon/memref-arg-attrs.mlir
```diff
@@ -0,0 +1,50 @@
...
+
+gpu.module @kernel {
+  gpu.func @test_func_dereferenceable_or_null(%arg0 : memref<f32> {llvm.dereferenceable_or_null = 4 : i64} ) {
+    gpu.return
+  }
+}
+
+// CHECK-LABEL:  llvm.func @test_func_dereferenceable_or_null
+// ROCDL-SAME:  !llvm.ptr {llvm.dereferenceable_or_null = 4 : i64}
+//  NVVM-SAME:  !llvm.ptr {llvm.dereferenceable_or_null = 4 : i64}
```
krzysz00: Nit: missing newline

external/llvm-project/mlir/test/Conversion/GPUCommon/memref-arg-noalias-attrs.mlir
```diff
@@ -0,0 +1,35 @@
+// RUN: mlir-opt %s -split-input-file -convert-gpu-to-rocdl='use-opaque-pointers=1 use-bare-ptr-memref-call-conv=1' | FileCheck %s --check-prefixes=CHECK,ROCDL
+// RUN: mlir-opt %s -split-input-file -convert-gpu-to-nvvm='use-opaque-pointers=1 use-bare-ptr-memref-call-conv=1' | FileCheck %s --check-prefixes=CHECK,NVVM
+// RUN: mlir-opt %s -split-input-file -convert-gpu-to-rocdl='use-opaque-pointers=1 use-bare-ptr-memref-call-conv=0' -o - 2>&1 | FileCheck %s --check-prefixes=CHECK-WARNING
```
krzysz00: This needs to be in a separate file `memref-arg-noalias-warning.mlir` that uses `-verify-diagnostics`

external/llvm-project/mlir/test/Conversion/GPUCommon/memref-arg-noalias-attrs.mlir
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// CHECK-LABEL:  llvm.func @func_without_any_attr
+// ROCDL-SAME:  !llvm.ptr
+//  NVVM-SAME:  !llvm.ptr
+
+// -----
+
+// CHECK-WARNING: Cannot copy noalias with non-bare pointers.
+gpu.module @kernel {
```
krzysz00: That is, this test in particular should be in its own file.

krzysz00 (commented):
Overall looks good, just need to shuffle around the diagnostic test

krzysz00 (approved):
Looks good to me, but could you add newlines to those test files.

PR #1154, opened 2023-07-14
fix constructing the FAT rockCompiler library on Windows

apwojcik:
Unfortunately, the `file(GLOB ..)` does not work on Windows as desired. __On the clean build, the command creates an empty list__. That is obvious - the command is executed during build configuration, and the searched files do not exist during that time. When a build run again, it works. However, the list contains files from the previous build. In case of new files are added/old ones are removed, those will not be included in the current build to form the FAT rockCompiler library.

Therefore, we provide a list of required files on Windows instead of searching for them. The Linux version keeps searching for the files as it was before. We could follow the Linux approach, but the size of a single static library on Windows can be, at most, 4GB. The size of Debug library is currently 5GB. Therefore we cannot provide FAT rockCompiler on Windows the same way it is on Linux.

The command on Linux is executed post-build - all required files are there to be found. However, the CONFIGURE_DEPENDS option does not work with MSVC multi-config generator in this scenario (the documentation mentions that might happen).

### DIFF
--- a/mlir/tools/rocmlir-lib/CMakeLists.txt
+++ b/mlir/tools/rocmlir-lib/CMakeLists.txt
@@ -1,9 +1,11 @@
 get_property(rocmlir_dialect_libs GLOBAL PROPERTY ROCMLIR_DIALECT_LIBS)
 get_property(rocmlir_capi_libs GLOBAL PROPERTY ROCMLIR_PUBLIC_C_API_LIBS)
+get_property(rocmlir_conversion_libs GLOBAL PROPERTY ROCMLIR_CONVERSION_LIBS)
 
 set(LIBS
   ${rocmlir_dialect_libs}
   ${rocmlir_capi_libs}
+  ${rocmlir_conversion_libs}
 )
 
 set(CMAKE_BUILD_RPATH ${CMAKE_BUILD_DIR}/external/llvm-project/llvm/lib)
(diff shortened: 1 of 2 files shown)


krzysz00 (changes requested):
I'm blocking this unless there's flat-out no other way to fix this.

Can we look into whether there's a way to get the required list of libraries from walking some sort of CMake state?

Alternatively, would it be possible to use shared libraries on Windows?

apwojcik:
The number of changes required to unblock building DLLs on Windows is enormous! By default, on Windows, building DLLs is disabled when `BUILD_FAT_LIBROCKCOMPILER` is `ON`. The blockage is coming from the llvm-project. However, more is needed to solve the main issue - the missing information on components to form the rockCompiler interface library required during build configuration. I tried getting it from various global properties, such as `ROCMLIR_DIALECT_LIBS`, `ROCMLIR_PUBLIC_C_API_LIBS`, `LLVM_LIBS`, `MLIR_STATIC_LIBS`, `MLIR_CONVERSION_LIBS`, etc. The conclusion is `ROCMLIR_*` properties contain incomplete information - missing targets. The `LLVM/MLIR_*` properties have more libraries on the list than were constructed during a FAT rockCompiler library build. I ended up in add/removing libraries manually to be able to fit in. That resulted in an even more complicated CMake code and far less maintainable than the one presented in this PR.
Furthermore, we must install LLVM libraries into a different directory than the default. Otherwise, we hit the problem of using incorrect LLVM libraries or overriding them by other `cmake install` commands. Several other components use different versions of the llvm-project, and we need to install each into a separate directory discoverable only by that particular component.

krzysz00:
Noted on DLLs not being a valid option.

The reason I'm still blocking this merge is that having this hardcoded list of unfolded dependencies deep in the build tree feels like a recipe for things breaking down the line.

I'd much rather find a way to generate this list at configure-time.
Or at least to make it shorter - if there's a few LLVM libraries we need to manually list out, it's not the end of the world, but this is ... not great.

On the third hand, per https://randomascii.wordpress.com/2023/03/08/when-debug-symbols-get-large/, it looks like that 4 GB limit for debug symbols has been removed in recent Windows.

apwojcik:
The limit is not about PDB file size or the page size used. It is related to the size of a single static library which cannot exceed 4GB. Here's the error message I am getting from the LIB.EXE:
`C:\rocMLIR\build.debug\lib\rockCompiler.lib : fatal error LNK1248: image size (1004B340E) exceeds maximum allowable size (FFFFFFFF)`
I checked again with the latest MSVC toolchain version 17.6.5. The limit is still there, unfortunately.
LNK1248: https://learn.microsoft.com/en-us/cpp/error-messages/tool-errors/linker-tools-error-lnk1248?view=msvc-170
The toolchain and host are 64bit, and the output binaries are 64bit. I even replaced the default linking options for DEBUG (i.e., I replaced `/Z7` with `/Zi`). Unfortunately, LLVM requires `/Z7` and returns to the default value.

apwojcik:
I changed the PR so the information on targets is read from the global properties. However, as I mentioned, the list has more targets than built, and some still need to be included. Other global properties store only partial lists, and `MLIR_ALL_LIBS` is the most complete. Others require fewer libraries to remove, but there are more libraries missed and to be added.
I prefer the previous version, which listed the targets explicitly because the list's purpose is straightforward. However, with the current version of PR, we still need to provide specific libraries to remove and add, and the list's goal might be unclear, plus it is large, too.

BTW: I removed the `workaround` tag because it is not the workaround anymore.

krzysz00:
Per discussion elsewhere, I'd be happy with the big list if we can somehow (maybe a CI step using Ninja's ability to output a graph or something) guarantee that the list is up to date.

apwojcik:
I have added the script to check the correctness of both lists that may be run using the target `check-rockCompiler`. The target is available only when `BUILD_FAT_LIBROCKCOMPILER` is `ON`. The target will automatically build the FAT library if not built or needs to be rebuilt. However, when the library is valid, it only runs the checks. The target `check-rocmlir` depends on the target `check-rockCompiler`, so it is triggered in CI, too. Example output of misalignment between build libraries and libraries from the lists:

```
$ cmake --build . --target check-rockCompiler
[2/2] Running utility command for __check_llvm_libs
FAILED: mlir/tools/rocmlir-lib/CMakeFiles/__check_llvm_libs.util
cmd.exe /C "cd /D C:\rocMLIR\build.release\mlir\tools\rocmlir-lib && C:\rocMLIR\mlir\tools\rocmlir-lib\checkTargets.py rockCompiler __llvm_targets llvm_targets.txt llvm_libs.txt"
** checkTargets.py: error: `__llvm_targets` of rockCompiler is missing elements - ADD them or UPDATE ignored targets in this script:
        tests
ninja: build stopped: subcommand failed.
```
Unfortunately, the target `check-rocmlir` creates more libraries than the FAT library requires. Therefore, I added ignored library list in the script. In case of an error like the above, a developer should decide whether to add the library to the required or ignored libraries list.
Below is an example of libraries missing from one list and removed from the other.
```
$ ninja check-rockCompiler
[1/2] Running utility command for __check_llvm_libs
FAILED: mlir/tools/rocmlir-lib/CMakeFiles/__check_llvm_libs.util
cmd.exe /C "cd /D C:\rocMLIR\build.release\mlir\tools\rocmlir-lib && C:\rocMLIR\mlir\tools\rocmlir-lib\checkTargets.py rockCompiler __llvm_targets llvm_targets.txt llvm_libs.txt"
** checkTargets.py: error: `__llvm_targets` of rockCompiler is missing elements - ADD them or UPDATE ignored targets in this script:
        tests
[2/2] Running utility command for __check_rocmlir_libs
FAILED: mlir/tools/rocmlir-lib/CMakeFiles/__check_rocmlir_libs.util
cmd.exe /C "cd /D C:\rocMLIR\build.release\mlir\tools\rocmlir-lib && C:\rocMLIR\mlir\tools\rocmlir-lib\checkTargets.py rockCompiler __rocmlir_targets rocmlir_targets.txt rocmlir_libs.txt"
** checkTargets.py: error: `__rocmlir_targets` of rockCompiler has invalid elements - REMOVE them:
        test2
ninja: build stopped: subcommand failed.
```

apwojcik:
Replaced with #1160

PR #1156, opened 2023-07-14
Remove MIOpen Integration from nightly and weekly CI

yiqian1:
- Remove MIOpen from Jenkinsfile.
- Tune convolution configs in MLIR.
- perfRunner uses installed MIOpenDriver to run benchmarks.


mlir/utils/performance/createGemmPerformanceReports.py
yiqian1, line 1: This file was renamed to createPerformanceReports.py and can create MLIR vs MIOpen reports.

krzysz00 (approved):
I'm approving to unbreak CI, but I'd really like to still see convolutions broken down by fwd/bwd/wrw like they were in the old MIOpen report.

mlir/utils/performance/createPerformanceReports.py
```diff
@@ -8,53 +8,47 @@
...
     plotMean.name = "Geo. mean"
     plotMean = pd.DataFrame(plotMean).T
 
-    plotMean[['MLIR TFlops', 'MIOpen TFlops (no MLIR Kernels)']]\
-        .to_csv(chip + '_' + reportUtils.PERF_PLOT_REPORT_FILE, index=False)
+    plotMean[['MLIR TFlops', f'{lib} TFlops (no MLIR Kernels)']]\
+        .to_csv(chip + '_' + reportUtils.PERF_PLOT_REPORT_FILE[lib], index=False)
 
-    means = df.groupby(["Direction", "DataType", "InputLayout"])[COLUMNS_TO_AVERAGE]\
+    means = df.groupby(["DataType"])[COLUMNS_TO_AVERAGE]\
```
krzysz00: More specifically, even though this is now a unified file, could we please include "Direction" and "InputLayout" when they exist? That's an important breakdown.

PR #1159, opened 2023-07-17
[mlir][AMDGPU] Move to new buffer resource intrinsics

krzysz00:
The AMDGPU backend now has buffer resource intrinsics that take a ptr addrspase (8) instead of a vector<4xi32>, improving LLVM's ability to reason about their memory behavior. This commit moves MLIR to these new functions.

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
@@ -58,10 +58,12 @@ class ROCDL_Op<string mnemonic, list<Trait> traits = []> :
 }
 
 class ROCDL_IntrOp<string mnemonic, list<int> overloadedResults,
-  list<int> overloadedOperands, list<Trait> traits, int numResults> :
+  list<int> overloadedOperands, list<Trait> traits, int numResults,
+  int requiresAccessGroup = 0, int requiresAliasAnalysis = 0> :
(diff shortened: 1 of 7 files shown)


external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
```diff
@@ -223,7 +225,166 @@ def ROCDL_wmma_bf16_16x16x16_bf16 : ROCDL_Wmma_IntrOp<"wmma.bf16.16x16x16.bf16">
 def ROCDL_wmma_i32_16x16x16_iu8 : ROCDL_Wmma_IntrOp<"wmma.i32.16x16x16.iu8">;
 def ROCDL_wmma_i32_16x16x16_iu4 : ROCDL_Wmma_IntrOp<"wmma.i32.16x16x16.iu4">;
 
+//===---------------------------------------------------------------------===//
```
sjw36, line 228: Code reordering makes it harder to correlate changes. Can you keep them in the same order.

sjw36 (approved):
Looks good. Would be better to keep the same order of ops in the td file.

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
krzysz00: I could, but I wanted to stick the old ops down where they'd be harder to find

PR #1161, opened 2023-07-18
[DO NOT SQUASH] Decouple VectorToLLVM, make host runner includes conditional

krzysz00:
This depends on #1160 .

1. Split VectorToLLVM into separate libraries for the pass and the conversion patterns so that you can include VectorToLLVM and not bring in, for example, the amx dialect (which has its own conversion patterns)
2. MHAL's host runner used vector-to-llvm as a pass, and is pulled in to the static library builds. We should, ideally, not be doing that. So this commit makes the host runner part (and its dependencies) optional - and turns them off for fat library builds.
3. Misc. changes made in the mlir directory, mostly around the new MHAL CMake options and removing a stray include.


sjw36 (approved):
Looks good. Please upstream to MHAL after.

krzysz00:
We are skip-ci because there's some issues with getting some of the sanity tests running with this patch and it's not super high-priority.

PR #1163, opened 2023-07-18
[Outliner] Refactor for divergence/re-convergence case in bert

sjw36:
* Add op when all operands are:
   * already collected or
   * external to the anchor op chain
      * depth first collection of these as leading ops 
* Use DenseSet for candidate lookup 
* Increased static size for leading/trailing ops vectors 
* Added debug traversal messaging 
* Also added Terminal op support  
  * don't traverse users 
* Disable for func containg MemRefTypes

https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1018

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Transforms/OutlinerUtils.h
+++ b/external/llvm-project/mlir/include/mlir/Transforms/OutlinerUtils.h
@@ -45,22 +45,22 @@ bool isConstantZero(Operation *op);
 
 class Outliner {
 public:
-  Outliner(function_ref<bool(Operation *)> isAnchorOp,
-           function_ref<bool(Operation *)> isLeadingOp,
-           function_ref<bool(Operation *)> isTrailingOp, StringRef outlineTag)
-      : isAnchorOp(isAnchorOp), isLeadingOp(isLeadingOp),
-        isTrailingOp(isTrailingOp), outlineTag(outlineTag) {}
+  Outliner(function_ref<bool(Operation *)> _isAnchorOp,
+           function_ref<bool(Operation *)> _isLeadingOp,
+           function_ref<bool(Operation *)> _isTrailingOp,
+           function_ref<bool(Operation *)> _isTerminatingOp,
+           StringRef _outlineTag)
+      : isAnchorOp(_isAnchorOp), isLeadingOp(_isLeadingOp),
+        isTrailingOp(_isTrailingOp), isTerminatingOp(_isTerminatingOp),
+        outlineTag(_outlineTag) {}
 
   void outline(ModuleOp module);
 
-private:
   function_ref<bool(Operation *)> isAnchorOp;
   function_ref<bool(Operation *)> isLeadingOp;
   function_ref<bool(Operation *)> isTrailingOp;
+  function_ref<bool(Operation *)> isTerminatingOp;
   StringRef outlineTag;
-  void traceInputs(Operation *op, Operation *ignoreOp,
-                   SetVector<Operation *> &predecessors,
-                   SetVector<Value> &inputNodes);
 };
 
 } // namespace mlir
--- a/external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
+++ b/external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
@@ -17,7 +17,6 @@
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
-#include "mlir/IR/Dominance.h"
 #include "mlir/IR/IRMapping.h"
 #include "mlir/IR/Matchers.h"
 #include "mlir/IR/PatternMatch.h"
@@ -33,14 +32,11 @@
 #include "llvm/ADT/SmallVector.h"
 #include <algorithm>
 #include <deque>
-#include <iostream>
 
 using llvm::SmallVector;
 using namespace mlir;
 
 #define DEBUG_TYPE "outliner-utility"
-
-#define FUSION_CAPACITY_DEFAULT 32
 
 static bool isZeroAttribute(Attribute value) {
   if (auto intValue = value.dyn_cast<IntegerAttr>())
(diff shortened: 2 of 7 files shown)


external/llvm-project/mlir/include/mlir/Transforms/OutlinerUtils.h
```diff
@@ -45,22 +45,22 @@ bool isConstantZero(Operation *op);
 
 class Outliner {
 public:
-  Outliner(function_ref<bool(Operation *)> isAnchorOp,
-           function_ref<bool(Operation *)> isLeadingOp,
-           function_ref<bool(Operation *)> isTrailingOp, StringRef outlineTag)
-      : isAnchorOp(isAnchorOp), isLeadingOp(isLeadingOp),
-        isTrailingOp(isTrailingOp), outlineTag(outlineTag) {}
+  Outliner(function_ref<bool(Operation *)> _isAnchorOp,
```
krzysz00, line 48: I don't think we need the leading underscores here?

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
```diff
@@ -66,23 +72,15 @@ bool mlir::isConstantZero(Operation *op) {
 // Inspired by / adapted from outlineIfOp() in SCF/Transforms/Utils.cpp
 // and mergeIdenticalBlocks() in Utils/RegionUtils.cpp.
 
-struct OutliningCandidate {
-  OutliningCandidate(Operation *anchorOp, ArrayRef<Operation *> &trailingOps,
-                     ArrayRef<Operation *> &leadingOps, ArrayRef<Value> &params,
-                     ArrayRef<Value> &returnVals, StringRef partFnName);
+class OutliningCandidate {
+  SmallVector<Type, FUSION_CAPACITY_DEFAULT> params;
```
krzysz00: Those'll be some pretty big stack allocations. Might as well leave these at their default sizes? Or drop the constant down to, say, 4 or 8?

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
```diff
@@ -173,337 +209,380 @@ static bool opsMatch(Operation *lhs, Operation *rhs, OutliningCandidate &one,
...
+    if (inOp == nullptr)
+      return false;
+    else if (inOp == _anchorOp)
+      return true;
+    else if (encountered.contains(inOp))
       return false;
+    encountered.insert(inOp);
+    for (auto opr : inOp->getOperands()) {
+      if (Operation *oprOp = opr.getDefiningOp()) {
+        debug(">>>", oprOp);
```
krzysz00, line 131: I think there's some magic incantations for showing recursion on a tree in the debug stream - it might be in the pattern rewriters somewhere. Or maybe I'm way off.

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
```diff
@@ -173,337 +209,380 @@ static bool opsMatch(Operation *lhs, Operation *rhs, OutliningCandidate &one,
...
+      }
+    }
+    // get Result Types
+
+    for (auto opPair : llvm::zip(ops, that.ops)) {
+      if (!opsMatch(std::get<0>(opPair), std::get<1>(opPair), that)) {
+        return false;
+      }
+    }
+    return true;
```
krzysz00: Could this be the built-in `OperationEquivalence` with appropriate flags set?

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
```diff
@@ -173,337 +209,380 @@ static bool opsMatch(Operation *lhs, Operation *rhs, OutliningCandidate &one,
...
-      else if (vtype.isa<MemRefType>())
-        outlinedFunc.setArgAttrs(
-            pair.index(),
-            b.getDictionaryAttr({readAccessAttr, writeAccessAttr}));
+    auto getAccessAttrs = [&](Type t,
+                              bool inputs) -> std::optional<DictionaryAttr> {
+      if (t.isa<VectorType, RankedTensorType, UnrankedTensorType>())
+        return b.getDictionaryAttr({inputs ? readAttr : writeAttr});
+      if (t.isa<MemRefType>())
+        return b.getDictionaryAttr({readAttr, writeAttr});
```
krzysz00, line 279: Is this overly conservative, or am I misremembering when this code runs?

krzysz00 (commented):
I'm not seeing anything obviously wrong here, but I'm also not very familiar with this code.

external/llvm-project/mlir/include/mlir/Transforms/OutlinerUtils.h
pcf000: Old hangover from me, and compilers that objected to identical names there.

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
```diff
@@ -173,337 +209,380 @@ static bool opsMatch(Operation *lhs, Operation *rhs, OutliningCandidate &one,
...
     std::vector<Operation *> anchors;
     auto callback = [&](Operation *op) {
+      for (auto operand : op->getOperands()) {
+        if (isa<MemRefType>(operand.getType()))
+          hasMemrefs = true;
+      }
+      for (auto result : op->getResults()) {
+        if (isa<MemRefType>(result.getType()))
+          hasMemrefs = true;
+      }
```
pcf000: Will be some redundant touches here, since (a) most results become somebody's operand and (b) we keep looking after hasMemrefs is true.

pcf000: Since we skip the rest of the function if hasMemrefs is true, perhaps just return instead of setting a flag?

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
```diff
@@ -173,337 +209,380 @@ static bool opsMatch(Operation *lhs, Operation *rhs, OutliningCandidate &one,
...
+    // Erase the ops we outlined, which should be safe now.
+    for (auto &op : llvm::make_early_inc_range(llvm::reverse(_leadingOps))) {
+      if (op->use_empty())
+        op->erase();
+    }
+  }
+
+  bool collect() {
+    // Given a Conv2DOp (or other anchor op), gather all Leading
+    // and Trailing ops in the PBV chain up to and including Terminal
```
pcf000, line 196: What does "PBV" mean?

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
pcf000: I'm having trouble understanding what this is for.  Wouldn't any op between the current and the anchor be included in the trailing-ops set?  When traversing forward from anchor, an op that doesn't qualify as trailing should stop the traversal.

Depending on circumstances, this could also be asking a post-dominance question.

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
```diff
@@ -173,337 +209,380 @@ static bool opsMatch(Operation *lhs, Operation *rhs, OutliningCandidate &one,
...
-}
 
-static OutliningCandidate *
-findOutliningCandidate(OutliningCandidate &newCandidate,
-                       std::vector<OutliningCandidate> &candidates) {
-  for (auto &candidate : candidates) {
-    if (outliningCandidatesEquivalent(candidate, newCandidate)) {
-      return &candidate;
+  // Make sure all operands are external or immediate in the chain from anchorOp
+  bool isImmediateOp(Operation *inOp) const {
```
pcf000, line 141: What does "immediate" mean in this context?  Its operands are in the outlined function, or aren't reachable from the anchor, but I can't relate that to the word "immediate."

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
pcf000: It started that way, but I didn't note why I made my own version.  Could be it didn't quite do what I wanted, or could be that I didn't want to mess with mainstream code (back in 2021).

external/llvm-project/mlir/include/mlir/Transforms/OutlinerUtils.h
sjw36: I like the leading _ for some classes to distinguish between local vars and members. See OutlineBuilder.

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
sjw36: This fixes the divergence problem. For example:

```
                         anchor
                               |
                           mul0
                          /      \
                          |     add0
                          |        |
                          |      mul1
                          |         |
                          \       /
                           add1
                             | 
                            ...

```
Previous impl would stop at the mul0 and the subsequent add0 and have 2 outputs because mul1 did not dominate all uses of output of mul0. 

This code will allow an op to be collected if all it's inputs are external (outside anchor chain) or immediately available in the collection. So mul1 will only consider the output of add0 and that add0 is collected.

sjw36: Hmm, got formatted.. does it make sense?

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
sjw36: Actually I am now disallowing MemRefTypes so it could be removed.

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
sjw36: pass-by-value

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
sjw36: Basically it means the op has already been collected as part of outline. 

In the example above, add1 is checked right after mul0 but before add0,mul1 so it has an input operand from mul0 that is in chain and collected but it has another input from mul1 that is in chain but not collected yet. So in order to keep from having holes the outline (i.e. one of the intermediates fails isTrailingOp) we postpone collecting it until we encounter it again on the second path.

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
pcf000: The benefit of a large initial allocation is that we delay allocating when adding to the vector.  Is the time taken for the realloc significant enough to worry about?

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
```diff
@@ -173,337 +209,380 @@ static bool opsMatch(Operation *lhs, Operation *rhs, OutliningCandidate &one,
...
+  void collectInputs(Operation *op, Operation *ignoreOp) {
+    for (const auto &operand : op->getOperands()) {
+      Operation *usedOp = operand.getDefiningOp();
+      if (usedOp) {
+        // Ignore input from anchor chain or already collected
+        if (usedOp != ignoreOp && !contains(usedOp)) {
+          if (_outliner.isLeadingOp(usedOp)) {
+            // Depth first collection of Leading ops
+            collectInputs(usedOp, op);
+            addPredecessor(usedOp);
```
pcf000, line 163: The reason for the little remove-and-insert dance that my version did is this situation:

```
A
|\
| - op2
|    |
 \   |
  - op1
     |
     |
   anchor
```

When collecting the inputs of op1, we see A first and add it to the leading ops.  Then we visit op1 and add it, and it also has A as an operand, but they're in the wrong order and a later stage will complain.  I had this happen but don't recall the example.  I can't convince myself that doing recur-then-add as you have here will avoid the problem.

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
```diff
@@ -173,337 +209,380 @@ static bool opsMatch(Operation *lhs, Operation *rhs, OutliningCandidate &one,
...
+    // _inputs gathers what will become the parameters of the
+    // outlined function;  initially it's the anchor's arguments,
+    // and it accumulates arguments to other ops that don't come
+    // from inside the outlined function.
+    //
+    // _results will become the results of the outlined function.
+    // These are gathered after all ops have been collected, if any
+    // op has external uses.
+
+    // BFS
```
pcf000, line 208: What does "BFS" mean?

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
```diff
@@ -173,337 +209,380 @@ static bool opsMatch(Operation *lhs, Operation *rhs, OutliningCandidate &one,
...
-  for (auto &candidate : candidates) {
-    if (outliningCandidatesEquivalent(candidate, newCandidate)) {
-      return &candidate;
+  // Make sure all operands are external or immediate in the chain from anchorOp
+  bool isImmediateOp(Operation *inOp) const {
+    for (auto opr : inOp->getOperands()) {
+      if (Operation *oprOp = opr.getDefiningOp()) {
+        debug("IN>", oprOp);
+        DenseSet<Operation *> encountered;
+        if (!contains(oprOp) && inChainToAnchor(oprOp, encountered))
```
pcf000, line 146: I believe, but haven't rigorously proven, that this inChainToAnchor() call is equivalent to domInfo.properlyDominates(anchor, oprOp);  thus more conservative than the old version and somewhat uglier, but otherwise okay.

pcf000:
Overall, a couple little concerns but otherwise it looks good.

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
sjw36: breadth-first-search

sjw36:
I encountered a crash yesterday when running ResNet50 in the candidate comparison code. As @krzysz00 and I were looking for problems in the DenseMapInfo configuration, we stumbled upon a pass, namely 'duplicate-function-elimination', that takes care of the duplicates. So I have removed all the candidate comparison code.

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
sjw36: I walked thru that code, but it seemed to only check that they were in the same block and one op came after another.. It didn't check for any relation via operands.

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
sjw36: Yes, I made leadingOps add in order of encounter, so I reversed the order here of recursion to go recurse and add all inputs and ops before adding the current op. So leadingOps no longer needs reversed.

external/llvm-project/mlir/lib/Transforms/Utils/OutlinerUtils.cpp
```diff
@@ -554,34 +361,32 @@ void Outliner::outline(ModuleOp module) {
...
         anchors.push_back(op);
+      return WalkResult::advance();
     };
+
     // Gather the anchor ops so we can process them back-to-front.
     func.walk(callback);
 
     if (!hasMemrefs) {
+      uint32_t cnt = 0;
       // (Problems with node mismatches and unexpected uses if we have the
```
pcf000, line 381: Can get rid of this comment, since "candidates" are gone.  (line 381;  silly github)

pcf000 (approved):
Aside from nitpicks, it looks okay to me.  I don't know why it's crashing.

PR #1164, opened 2023-07-18
[NFC] Refactor input tile loading and storing to LDS to be based on just views

manupak:
Currently there are a number of significant coupling between functions in gridwise to blockwise lowering that assumes a certain shapes and transforms. This kind of restricts usage of these for other gridwise kernels.

In this first refactor PR,

Im introducing a singular view creation function -- createGemmInputViewsFromGlobal --  that gives gridwise, blockwise and threadwise views of the global input tensors based on derivative tuning params.

Then refactor the rest code where input tile loading to just to rely on the those views generate further code.
In the process Im removing functions : 
* wrapMatrixForGlobalLoad (most of the view creation logic is lifted from here to createGemmInputViewsFromGlobal)
* globalVectorLayout - this is trying to re-generate threadwise view independently and not even used by createGlobalLoadLoop
* createGlobalLoadLoop - this seems identical to threadwise_read_into op.
* viewLoadBufferDK - I introduced this earlier to view loadedBuffer as DxK threadwise slice of the input tensor. However, it is just the inverted view of createGemmInputViewsFromGlobal.threadwiseView)

~~If these seems reasonable, I ll further refactor wrapLDSforStore (which is what I really want to be based on views) in the next PR~~

UPDATE : 

* I have done the further refactor to decouple wrapLDSStoreBuffer function from global tile loading assumption on this PR.
* Basically problem was wrapLDSStoreBuffer integrates knowledge from how the tile was loaded into regs and creates the views that it does not have control over. I've decoupled this now.

closes : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1029


mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -356,89 +263,9 @@ static ArrayAttr ldsVectorLayout(OpBuilder &b, Location loc,
...
-  TransformMapAttr kdView = kdViewBuilder.get();
-  ret = b.create<TransformOp>(loc, ret, kdView);
-
-  return ret;
-}
-
 /// This function pack the load buffer into a store buffer ready to be copied
 /// into LDS:
-///  - The load buffer is (viewed as) a DPerThread x KPerThread
+///  - The load buffer is (viewed as) a KPerThread x DPerThread
```
manupak, line 243: This is to be consistent with all types of views to end with ... x K x D

giuseros: Does it mean we are transposing while loading?

manupak: no; it is a just a view of the load buffer.

manupak:
```c++
static void makeGemmInputViewsIter(TopDownTMBuilder &viewBuilder,
                                   StringRef dIterName, int64_t dPerThread,
                                   int64_t kPerThread,
                                   ArrayRef<unsigned> outDims,
                                   bool isKContigousDim) {
  if (isKContigousDim) {
    viewBuilder.merge({dIterName, "k_iter"}, outDims, "iter",
                      {dPerThread, kPerThread});
  } else {
    viewBuilder.merge({"k_iter", dIterName}, outDims, "iter",
                      {kPerThread, dPerThread});
  }
}
```
i.e. how iter is flattened will differ based on what is the contiguous dim

mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
```diff
@@ -22,6 +22,18 @@ namespace rock {
 struct ConvolutionDims;
 struct GemmSize;
 
+struct GPUViews {
+  ArrayAttr gridwiseView;
+  ArrayAttr blockwiseView;
```
krzysz00: Can we get some comments here on what these views actually are?

krzysz00 (commented):
I'm broadly happy here aside from some documentation.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -252,9 +185,7 @@ static FailureOr<Value> wrapMatrixForGlobalLoad(
 static FailureOr<Value> wrapLDSBufferForStore(OpBuilder &b, Location loc,
                                               Value buffer, Type ldsReadType,
                                               int64_t kOuter, StringRef dName,
-                                              int64_t d, int64_t kPerThread,
-                                              int64_t dPerThread,
-                                              GemmDimension vectorDim) {
```
manupak, line 257: Ive removed the vectorDim now

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -810,42 +617,89 @@ struct GridwiseGemmRewritePattern : public OpRewritePattern<GridwiseGemmOp> {
...
+    // so that we can view loadBuffer as a K x D tensor
+    ArrayAttr loadBufferAViews =
+        invertTransforms(b, loc, maybeABufferViews->threadwiseView);
+    Value viewLoadBufferA = transform(b, loadBufferA, loadBufferAViews);
+    // Prior to LDS store, we need re-arrange register buffer to maxmize LDS
+    // vectorization Hence, creating the view w.r.t global that correspond to
+    // such re-arranged register buffer
+    auto [aKLdsStoreIterSplit, mLdsStoreiterSplit] =
+        createLDSStoreIterSplits(aCopyKPerThread, copyMPerThread, kpack);
+    FailureOr<GPUViews> maybeALdsStoreViews = createGemmInputTileViews(
```
manupak: We create the re-shuffled regs views first throughout (not just threadwise).
Then we transform the storeBuffer with inverted threadwise view that will produce K x D view of whats inside the regs.
This makes this composable with the K x D view of the loadBuffer.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -810,42 +617,89 @@ struct GridwiseGemmRewritePattern : public OpRewritePattern<GridwiseGemmOp> {
...
+    auto packBLoop = packLoadBufferToStoreBuffer(
+        b, loc, elementTypeB, kpack, viewLoadBufferB, viewStoreBufferB);
+
     Type ldsReadTypeA = vectorTypeOrSelf(elementTypeA, kpack);
-    Type ldsReadTypeB = vectorTypeOrSelf(elementTypeB, kpack);
-    FailureOr<Value> maybeWrappedLdsA = wrapLDSBufferForStore(
-        b, loc, ldsBufferA, ldsReadTypeA, kpacksPerBlock, "m", mPerBlock,
-        aCopyKPerThread, copyMPerThread, aVectorDim);
+    FailureOr<Value> maybeWrappedLdsA =
+        wrapLDSBufferForStore(b, loc, ldsByteBufferA, ldsReadTypeA,
```
manupak, line 670: Now wrapping here will produce K x D to flatLDS byte buffer because (tid, iter) to K x D is defined by the tile loading process , not here.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -810,42 +617,89 @@ struct GridwiseGemmRewritePattern : public OpRewritePattern<GridwiseGemmOp> {
...
+                              kpacksPerBlock, "m", mPerBlock, aCopyKPerThread);
     if (failed(maybeWrappedLdsA))
       return maybeWrappedLdsA;
-    FailureOr<Value> maybeWrappedLdsB = wrapLDSBufferForStore(
-        b, loc, ldsBufferB, ldsReadTypeB, kpacksPerBlock, "n", nPerBlock,
-        bCopyKPerThread, copyNPerThread, bVectorDim);
+    // This is KxD view of the flat LDS buffer
+    Value wrappedLdsA = std::move(*maybeWrappedLdsA);
+    // This will produce a (tid, iter) --> flat LDS view
+    wrappedLdsA = transform(b, wrappedLdsA, maybeALdsStoreViews->blockwiseView);
```
manupak: Here we compose : (tid, iter) --> K x D (defined by tile loading process) and K x D --> flatLDSbuffer

manupak (commented):
I've done somewhat significant refactor here.. and put down some comments to follow the PR. 
Since it was all in my head, I m happy to anwser questions / improve docs.

The main gist is I've decoupled view creations to always be based on the operator space which is K x D (basically logical tile dims). Then construct & compose views to reg/lds buffers as needed. This basically allows to reuse much of the components with the operations that can produce K x D views.

mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
manupak: I've added for these and others here.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
manupak: or it could be defined by a prior gemm for e.g.

mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
```diff
@@ -22,6 +22,76 @@ namespace rock {
 struct ConvolutionDims;
 struct GemmSize;
 
+// This is a helper data structure
```
krzysz00: This needs more clarity.

Tile of what? What'll might use these outputs? What are K and D? Are they the same for all of these?

mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
```diff
@@ -22,6 +22,76 @@ namespace rock {
...
+// to server as a return type for
+// view creating utilities.
+struct GPUViews {
```
krzysz00: This struct name is _way_ too generic. `LoadedTileViews`? `GridToMatrixMaps`?

mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
```diff
@@ -22,6 +22,76 @@ namespace rock {
...
+  // This is a [gridIdx0, ... , gridIdxN, tid, iter] to K x D view
+  // This is useful to define a global tile once loaded into regs.
+  ArrayAttr gridwiseView;
+  // This is a [tid, iter] to flat K x D view
+  // Useful to define a blockwise tile once loaded into regs.
+  ArrayAttr blockwiseView;
+  // This is a [iter] to flat K x D view
+  // Useful to define a threadwise tile once loaded into regs.
```
krzysz00: Somewhere, either in this file or elsewhere, define what this function is and how you're meant to use it.

mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
```diff
@@ -22,6 +22,76 @@ namespace rock {
...
+  ArrayAttr threadwiseView;
+};
+
+// This is helper data structure to allow
+// lowering define lower dimensions (K or D) re-shuffling
+// into thread_id or iter into reg buffers in threads.
```
krzysz00: I do not understand this comment

mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
```diff
@@ -22,6 +22,76 @@ namespace rock {
...
+// It is particularly useful to arrange data
+// in registers before they are loaded/written
+// to some memory (global or lds).
+struct LowerDimPartInfo {
+  // This defines sub-dimension names of either tid or iter.
+  // These sub-dimensions will be part of the merger when creating
+  // the tid or iter dimensions.
+  SmallVector<StringRef> lowerDimPartNames;
```
krzysz00: Returning a `StringRef` is extremely suspicious on memory safety grounds.

mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
```diff
@@ -22,6 +22,76 @@ namespace rock {
...
+//    This provides how d dimension should be split into sub-dimensions of tid
+// 3) kDimIterPartInfo :
+//    This provides how k dimension should be split into sub-dimensions of iter
+// 4) dDimIterPartInfo :
+//    This provides how d dimension shuold be split into sub-dimensions of iter
+FailureOr<GPUViews> createGemmInputTileViews(
+    OpBuilder &b, Location loc, Value gBuffer, StringRef dName,
+    ArrayRef<StringRef> bidGridOrder, ArrayRef<int64_t> bidGridLengths,
+    int64_t gridSize, int64_t blockSize, int64_t kPerBlock, int64_t dPerBlock,
+    LowerDimPartInfo kDimTidPartInfo, LowerDimPartInfo dDimTidPartInfo,
```
krzysz00: These dim part info things should probably be taken by reference, they're big

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
```diff
@@ -125,3 +128,225 @@ Type mlir::rock::vectorTypeOrSelf(Type elementType, int64_t len) {
...
+static SmallVector<unsigned> getOutDimIndices(int64_t outDimStartIdx,
+                                              int64_t newDimSize) {
+  SmallVector<unsigned> outDims;
+  for (int i = 0; i < newDimSize; i++) {
+    outDims.push_back(outDimStartIdx + i);
+  }
+  return outDims;
+}
+
+static std::tuple<SmallVector<StringRef>, SmallVector<int64_t>>
```
krzysz00: Take these as out parameters of `SmallVectorImpl`

mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
krzysz00: (It'll probably work, but I'd be careful here if anyone is constructing derived strings from this value. If nothing else, warn about it)

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
```diff
@@ -125,3 +128,225 @@ Type mlir::rock::vectorTypeOrSelf(Type elementType, int64_t len) {
...
+    }
+
+    {
+      SmallVector<StringRef, 3> dSubDimNames{thisBlockDim};
+      dSubDimNames.append(dDimTidPartInfo.lowerDimPartNames.begin(),
+                          dDimTidPartInfo.lowerDimPartNames.end());
+      dSubDimNames.append(dDimIterPartInfo.lowerDimPartNames.begin(),
+                          dDimIterPartInfo.lowerDimPartNames.end());
+      SmallVector<int64_t, 3> dSubDimSizes{dGlobal / dPerBlock};
+      dSubDimSizes.append(dDimTidPartInfo.lowerDimPartSizes.begin(),
```
krzysz00: I'm a bit suspicious of this

This looks like uniformly assuming that the iter dimension moves faster than the tid dimension, which doesn't feel right.

mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
```diff
@@ -22,6 +22,76 @@ namespace rock {
...
+  ArrayAttr threadwiseView;
+};
+
+// This is helper data structure to allow
+// lowering define lower dimensions (K or D) re-shuffling
+// into thread_id or iter into reg buffers in threads.
+// It is particularly useful to arrange data
+// in registers before they are loaded/written
+// to some memory (global or lds).
+struct LowerDimPartInfo {
```
krzysz00: I'm not liking these.

It doesn't seem like you can quite accomplish what you're going for by just returning a `rock::TransformAttr` and some result sizes, but ... maybe? 

Especially if the transform map builders grew an `insertTransform(TransformAttr, ArrayRef<int64_t>)`.

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
```diff
@@ -125,3 +128,225 @@ Type mlir::rock::vectorTypeOrSelf(Type elementType, int64_t len) {
...
+                                              int64_t newDimSize) {
+  SmallVector<unsigned> outDims;
+  for (int i = 0; i < newDimSize; i++) {
+    outDims.push_back(outDimStartIdx + i);
+  }
+  return outDims;
+}
+
+static std::tuple<SmallVector<StringRef>, SmallVector<int64_t>>
+getCombinedOrder(const LowerDimPartInfo &lhs, const LowerDimPartInfo &rhs) {
```
krzysz00: I'm very suspicious of the fact that you need this.

krzysz00 (changes requested):
I'm sticking a block on this because I don't understand it and it's just ... really suspicious. 

It feels both overly general and like it's making a lot of strong assumptions about clients at the same time.

I'm also calling for improved documentation, but I'm not sure it'd be good to write it until we're happy with the design.

mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
manupak: These are an artifact that we cant create a graph of transforms rather than a chain of transform. 

I can flat out create two independent view creators here : 
1) loaded tile from global
2) reshuffled tile from global for lds store

Then remove these utils. 
I did this to factor the common thing between the above two.

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
manupak: it works between :

1) loaded tile from global
2) reshuffled tile from global

and yea it wont work for others -- not as general that bit.
I could just simply make two independent view creators.

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
manupak: The fact this was needed to factor out create similiar views for 1) loaded tile from global and 2) reshuffled tile from global where the difference is only how iter maps to indices in regs.

manupak:
I think broadly the complaints are about the attempt in trying to factor out two view generators which might as well be independent of each other though they share partial coord transform of bid and tid but not iter.  I might remove all that utils and create two seperate functions : 
1) createGemmInputTileViewsFromGlobal
2) createGemmInputTileViewsToLds

please confirm if that would clear things up for you.

krzysz00:
I'd like to understand what it is you meant by "input tile" in those function names before I say anything concrete.

manupak:
Input tile is the tile of the input passed to the gemm. It is two dimensional but 3 different sizes when you look at it from gridwise, blockwise, and threadwise level.

Gridwise: this is the largest 2d tile all blocks collectively hold in the GPU
Blockwise: this is the tile all threads in a block collectively hold in the GPU
Threadwise: this is the tile each thread hold

After transformations, we keep track of those views.

krzysz00:
Ok, so, because now that I've slept on things I think it's important to get some more names in here because I'm having trouble what's going on here.

For simplicity, I'm leaving the `G` dimension out of all of this.

First, some variable names and their ranges (they may not match the code or the transform dims, but)
- `kg : [0, K)` a coordinate in the K dimension of the global tensor
- `dg : [0, M/N)` a coordinate in the D (M/N) dimension of the global tensor
- `kbl : [0, KPerBlock)` a coordinate in the K dimension of the tile loaded by a workgroup
- `dbl : [0, DPerBlock)` same for the D dimension
- `kObs : [0, KPacksPerBlock)` a coordinate in the kOuter dimension of the LDS matrix
- `dbs : [0, DPerBlock)` same for the D dimension
- `kPbs : [0, KPack)` same for kPack
- `krl : [0, KPerThread)` a coordinate in the registers loaded from global memory, in the K dimension.
- `drl: [0, DPerThread)` similarly
- `kOrs : `[0, KOuterPerThread)`: a coordinate in the registers as they're being stored to LDS, KOuter dimension
- `drs: [0, DPerThread)`: similarly
- `kPrs : [0, kPackPerThread)`: similarly
- `...bid`: The coordinates representing the block ID
- `tid : [0, blockSize)`: ID of a thread within a workgroup
- `iter : [0, kPerThread * dPerThread = R)` the number of values each thread posseses

We have, or might want  have, the following functions
- `toGlobalMatrixCoords(... bid, tid, iter) -> (kg, dg)` (... to actual global memory)
- `loadRegsAsThreadTile(iter) -> (krl, drl)` We don't currently have this. In your case you'll probably want `loadRegsAsThreadTile(iter) -> (krl, drl)` and `accelRegsAsThreadTile(iter) -> (krl, drl)` because they could be distinct
  - Or, if I've correctly understood the generalization you're going for, there's `loadRegsAsThreadTile(tid, iter) -> (krl, drl)` and `accelRegsAsThreadTile(tid, iter) -> (krl, drl)`
- We have `threadTileToPackedThreadTile(krl, drl) -> (kOrs, drs, kPrs)`, which handles the packing and transposes
- We have `storeRegAsPackedThreadTile(kOrs, drs, kPrs) -> (iter)`
- We have `toLDSMatrixCoords(tid, iter) -> (kObs, dbs, kPbs)`

I think some of your changes come down to making those `RegsAsTile` functions go the other way: that is, we'll define `threadTileToLoadRegs(krl, drl) -> (iter)` and that sort of thing.

Have I missed some functions you need?

manupak:
Apart from naming, we are mostly talking about the same thing except : 

> We have threadTileToPackedThreadTile(krl, drl) -> (kOrs, drs, kPrs), which handles the packing and transposes
> We have storeRegAsPackedThreadTile(kOrs, drs, kPrs) -> (iter)
> We have toLDSMatrixCoords(tid, iter) -> (kObs, dbs, kPbs)

One of the qualities I want to preserve at each stage of lowering, I'd like to see that as a tile that is going into the gemm 'logically'. 
e.g. instead of threadTileToPackedThreadTile (krl, drl) -> (kOrs, drs, kPrs): 

we do : 
iter -->  (kOrs, drs, kPrs) --> (krl, drl)  
Lets call it packedRegsAsThreadTile

This view tells me what what the registers holds after the packing as well as 'how' it is being held.
Moreover, it allows me to do the packing as follows :
i.e. 
loadRegsAsThreadTile[loadRawBuffer] --> packedRegsAsThreadTile[storeRawBuffer]

Now because I have that view -- packedRegsAsThreadTile -- I can compose this further.

Extending that quality if every macro op signoff using that logical shape views -- we can keep composing ops.

manupak:
So packedRegAsThreadTile is no different in structure to loadRegsAsThreadTile or accelRegsAsThreadTile

manupak:
On top of that I want 

XAsThreadTile
XAsBlockTile
XAsGridTile

to be generated simultaneously so we have them at every point of lowering without one going further than the others...

manupak:
End of the day, following your terminology, I'd like to have : 
1) loadRegsAsThreadTile, loadRegsAsBlockTile, loadRegsAsGridTile
2) packedRegsAsThreadTile, packedRegsAsBlockTile, packedRegsAsGridTile
3) accelRegsAsThreadTile, accelRegsAsBlockTile, accelRegsAsGridTile

Since I wanted each of them to be simultaneously generated : 

1) loadRegsAsTileViews(...) -> GPUViews
2) packedRegsAsTileViews(...) -> GPUViews
3) accelRegsAsTileViews(...) --> GPUViews

manupak:
Now Ive done : 
* GPUViews --> RegsAsMatrixSubTiles
* created getLoadRegsAsTileViews
* created getPackedRegsAsTileViews

and removed the rest of the utils as discussed.
PTAL @krzysz00;

krzysz00 (approved):
This all looks pretty comprehensible ... approving, but I'd like an assembly diff check (and a quick perf check) before landing this just to be safe.

manupak:
Yes -- agreed.
The gemms look flat but convs have regressions.
The regressions are for conv nchw x [i8, f16] cases which shouldn't be there.
 Im looking into it what causes that.

[PR1164_perf.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/12239718/PR1164_perf.zip)

manupak:
[PR1164_conv_perf16.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/12247895/PR1164_conv_perf16.zip)

Ok conv perfs are more or less flat now.. It turned out to be that vector length inference is not working for all the cases for global loads when its looked at after the input is viewed as gridSubTile. I just added a forceVecLen option for threadwise_read_into workaround the issue and filed a bug report : 
https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1049

This made the perf flat as expected.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -176,73 +176,6 @@ computeCopyPerThread(Type elementType, int64_t copyPerThread, int64_t kPerBlock,
...
-  SmallString<8> dThreadName = llvm::formatv("{0}_thread", dName);
-
-  // Note: (kThreads * dThreads) = (kPerBlock * dPerBlock) / dataPerThread) =
-  // blockSize
-  int64_t kThreads = kPerBlock / kPerThread;
-  int64_t dThreads = dPerBlock / dPerThread;
-
-  TopDownTMBuilder splitId(
-      b, {"k_loop", "g_block", "m_block", "n_block", "tid", "iter"},
-      {kIters, bidGridLengths[0], bidGridLengths[1], bidGridLengths[1],
```
manupak, line 213: In my analysis (mostly IR diffs), I have found and fixed this bug currently; we use m_blocks as n_blocks.

krzysz00:
Ok, wait, so does the vectorization analysis still need to be tweaked or?

manupak:
So the current vectorization analysis "works" on the view/memref present to the op (though our perf tests does only checks tranposes realistically, aren't they?). 

However, if I do further views on it (i.e. gridSubTile) it kind of breaks.

So there are only two explanations : 
1) either my views are wrong
2) or vectorization analysis doesn't work in some cases

If you are suspecting 1) I fail to see how that would produce the correct result with forced vectorization.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -356,89 +238,9 @@ static ArrayAttr ldsVectorLayout(OpBuilder &b, Location loc,
...
-  Value zero = b.createOrFold<arith::ConstantIndexOp>(loc, 0);
-
-  SmallVector<Value, 4> globalStart = {
-      zero, coords.g_block, coords.m_block, coords.n_block, tid, zero};
-  SmallVector<Value, 4> vectorStartOuter(6, zero);
-  auto outerLoop = b.create<TransformingForOp>(
-      loc, ArrayRef<ValueRange>{globalStart, vectorStartOuter},
-      ArrayRef<Attribute>{matrixToTensor, b.getArrayAttr({})},
-      /*bounds=*/ArrayRef<int64_t>{1, 1, 1, 1, 1, dataPerThread},
-      /*strides=*/ArrayRef<int64_t>{1, 1, 1, 1, 1, vectorLen}, forceUnroll,
```
manupak, line 385: Also I'd like to point out that vectorLen is already forced in the current code where there isn't vec len inference done after untransform. So this PR doesn't change this aspect.

manupak: I can reproduce the same issue with the following changes to develop as well...

```c++
--- a/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
@@ -359,20 +359,26 @@ static ArrayAttr ldsVectorLayout(OpBuilder &b, Location loc,
 static TransformingForOp
 createGlobalLoadLoop(PatternRewriter &b, Location loc, GpuAllocOp loadBuffer,
                      Value wrappedMatrix, ArrayAttr vectorMap,
-                     int64_t dataPerThread, int64_t vectorLen, Value bid,
+                     int64_t dataPerThread, int64_t vectorLenExt, Value bid,
                      Value tid, layout::GridCoordinates coords,
                      bool forceUnroll) {
   Value tensor;
   ArrayAttr matrixToTensor;
   std::tie(tensor, matrixToTensor) = untransform(b, wrappedMatrix);
+  ArrayRef<int64_t> tensorSize = tensor.getType().cast<ShapedType>().getShape();
+
+  ArrayRef<int64_t> wrappedMatrixShape =
+      wrappedMatrix.getType().cast<ShapedType>().getShape();
+  Type elemType = wrappedMatrix.getType().cast<MemRefType>().getElementType();
+  int64_t vecLen = getMaxVectorizationForDatatype(matrixToTensor, /*dim=*/5,
+                                          wrappedMatrixShape[5], tensorSize, elemType);

   // Optimize the transform chain.
-  ArrayRef<int64_t> tensorSize = tensor.getType().cast<ShapedType>().getShape();
   matrixToTensor = collapseContiguousMerges(matrixToTensor, tensorSize);

   Type elementType =
       wrappedMatrix.getType().cast<MemRefType>().getElementType();
-  Type loadType = vectorTypeOrSelf(elementType, vectorLen);
+  Type loadType = vectorTypeOrSelf(elementType, vecLen);
   Value zero = b.createOrFold<arith::ConstantIndexOp>(loc, 0);
```

manupak: rocmlir-gen --operation conv2d -t i8 --arch gfx908:sramecc+:xnack- --num_cu 120 --fil_layout kcyx --in_layout nchw --out_layout nkhw --batchsize 256 --in_channels 64 --in_h 56 --in_w 56 --out_channels 256 --fil_h 1 --fil_w 1 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h 0 --padding_w 0 --kernel-repeats 5 --perf_config=32,128,32,32,32,1,1,1 | rocmlir-driver -c --mlir-print-ir-before=rock-blockwise-gemm-to-threadwise &> dump_br.log

manupak:
Ok, I've reverted back forceVec changes and included the commit from https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1179 to check the CI.

Perfs are flat now as well...
[PR1164_conv_perf17.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/12261123/PR1164_conv_perf17.zip)

manupak:
I've taken this in on the basis it was approved code (at the point of approval) and perfs are flat.

PR #1166, opened 2023-07-19
Add atomic add support everywhere, remove spurious errors

krzysz00:
1. Replace some "errors" that don't actually stop the compilation with warnings.
2. Remove the limitation on atomic_add since we have atomic_add emulation
3. Add a test for said atomic emulation
4. Disable align-tiling on non-kernel funcitons

This'll improve MIGraphX's usability on Navi and get a bunch of spam out of people's logs.

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
@@ -606,6 +606,10 @@ ReduceRewritePattern::matchAndRewrite(rock::ReduceOp reduceOp,
 
 void RockLinalgAlignPass::runOnOperation() {
   MLIRContext *ctx = &getContext();
+  func::FuncOp func = getOperation();
+  // Only run this pass on GPU kernel functions.
+  if (!func->hasAttr("kernel"))
+    return;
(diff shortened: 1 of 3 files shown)


mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -606,21 +606,24 @@ ReduceRewritePattern::matchAndRewrite(rock::ReduceOp reduceOp,
 
 void RockLinalgAlignPass::runOnOperation() {
   MLIRContext *ctx = &getContext();
+  func::FuncOp func = getOperation();
+  // Only run this pass on GPU kernel functions.
+  if (!func->hasAttr("kernel"))
+    return;
```
sjw36, line 612: Thought this check was already here.. strange

sjw36 (approved):
Looks good.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
krzysz00: It's there in regularize but not in align tiling

PR #1167, opened 2023-07-19
Extract test vector from the problem description string

yiqian1:
Also only do forward convolution on unet configurations.

### DIFF
--- a/mlir/utils/performance/conv-configs
+++ b/mlir/utils/performance/conv-configs
@@ -24,20 +24,20 @@
 -n 256 -c 3 -H 230 -W 230 -k 64 -y 7 -x 7 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -t 1
 
 # This section of the file is selected conv configs for unet (stable diffusion).
--n 2 -c 1280 -H 32 -W 32 -k 640 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
--n 2 -c 1280 -H 32 -W 32 -k 640 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
--n 2 -c 1920 -H 16 -W 16 -k 1280 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
--n 2 -c 1920 -H 16 -W 16 -k 1280 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
--n 2 -c 1920 -H 32 -W 32 -k 640 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
--n 2 -c 1920 -H 32 -W 32 -k 640 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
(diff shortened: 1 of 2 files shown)


mlir/utils/performance/perfRunner.py
```diff
@@ -837,7 +837,7 @@ def getFusionTestInfo(filename, paths: Paths):
     p2.stdout.close()
     output, _ = tuningKey.communicate()
     result = output.decode('utf-8').strip().split('\t')
-    testEntry = {'filename' : filename, 'testVector' : result[1], 'futName' : futName}
+    testEntry = {'filename' : filename, 'testVector' : result[2], 'futName' : futName}
```
yiqian1, line 840: [PR#1145](https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1145) added 'numCU' to the problem description string, as a result the tuning script should skip it and extract the test vector from the third item.

mlir/utils/performance/conv-configs
```diff
@@ -24,20 +24,20 @@
...
+-F 1 -n 2 -c 1920 -H 16 -W 16 -k 1280 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
+-F 1 -n 2 -c 1920 -H 16 -W 16 -k 1280 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
+-F 1 -n 2 -c 1920 -H 32 -W 32 -k 640 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
+-F 1 -n 2 -c 1920 -H 32 -W 32 -k 640 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
+-F 1 -n 2 -c 320 -H 32 -W 32 -k 640 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
+-F 1 -n 2 -c 320 -H 32 -W 32 -k 640 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
+-F 1 -n 2 -c 320 -H 64 -W 64 -k 320 -y 3 -x 3 -p 1 -q 1 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -t 1
+-F 1 -n 2 -c 320 -H 64 -W 64 -k 4 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
+-F 1 -n 2 -c 640 -H 16 -W 16 -k 1280 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
+-F 1 -n 2 -c 640 -H 16 -W 16 -k 1280 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
```
krzysz00, line 38: Why is `-F 1` here? Can you leave a comment?

I assume this is a temporary measure because bwd weight and bwd data tuning are broken.

Could you perhaps, instead, go into the tuner and temporarily remove `-F 2` and `-F 4` as options?

krzysz00 (commented):
I'm not sure about how bwd weight/data were disabled.

mlir/utils/performance/conv-configs
yiqian1: Because for MIGraphX we only want forward. The original config file had "-F 1", but we removed it in [this PR](https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1083), because miopen-validate.sh couldn't handle 'forward only' configs. I just restored the changes.

krzysz00 (approved):

PR #1168, opened 2023-07-20
[DO NOT SQUASH] Remove our software BF16 pass and make related upstream changes

krzysz00:
Commit messages are listed below

The first is a backport from upstream around calling math functions on vectors that's needed to make other diffs behave.

---

[Internal changes and cleanups for SoftwareBF16 removal](https://github.com/ROCmSoftwarePlatform/rocMLIR/commit/4043a2edd37fdf7a3557fce2cf516b06aefeaa64)

1. Add the mlir_float16_utils.so file to all our test runs to pick up
__truncsbf16 and such.
2. Make the "rocdl" pipeline option less error-prone by just making it
the "backend" compile option minus the actual serialize-to-hsaco
3. Add bf16 emulation to the GPU module pipeline (but not to host
code, LLVM can handle things there)
4. Clean up Rock's pipelines.cpp by summing a single function pass
manager.

---

[Remove the SoftwareBF16 pass](https://github.com/ROCmSoftwarePlatform/rocMLIR/commit/ff518444b5785287b3ed5e69eadd132e480d6d39)

This is separate from the other external/ changes to make said changes
upstreamable.

---

[[mlir][AMDGPU] Improve BF16 handling through AMDGPU compilation](https://github.com/ROCmSoftwarePlatform/rocMLIR/commit/824ea97193bf22f85772c470045f54445792f5ac)

Many previous sets of AMDGPU dialect code have been incorrect in the
presence of the bf16 type (when lowered to LLVM's bfloat) as they were
developed in a setting that run a custom bf16-to-i16 pass before LLVM
lowering.

An overall effect of this patch is that you should run
--arith-emulate-unsupported-floats="source-types=bf16 target-type=f32"
on your GPU module before calling --convert-gpu-to-rocdl if your code
performs bf16 arithmetic.

While LLVM now supports software bfloat, initial experiments showed
that using this support on AMDGPU inserted a large number of
conversions around loads and stores which had substantial performance
imparts. Furthermore, all of the native AMDGPU operations on bf16
types (like the WMMA operations) operate on 16-bit integers instead of
the bfloat type.

First, we make the following changes to preserve compatibility once
the LLVM bfloat type is reenabled.
1. The matrix multiplication operations (MFMA and WMMA) will bitcast
bfloat vectors to i16 vectors.
2. Buffer loads and stores will operate on the relevant integer
datatype and then cast to bfloat if needed.

Second, we add type conversions to convert bf16 and vectors of it to
equivalent i16 types.

Third, we add the bfloat <-> f32 expansion patterns to the set of
operations run before the main LLVM conversion so that MLIR's
implementation of these conversion routines is used.

Finally, we extend the "floats treated as integers" support in the
LLVM exporter to handle types other than fp8.

We also fix a bug in the unsupported floats emulation where it tried
to operate on `arith.bitcast` due to an oversight.

---

[[mlir][Arith] Change F32 to BF16 truncation to match __truncsfbf2](https://github.com/ROCmSoftwarePlatform/rocMLIR/commit/c9d82b3accd031c5ceb67521907673cdffef496b)

The current implementation of truncf %* : f32 to bf16 in the expand
ops pass is slow and does not match the __truncsfbf2 / float2bf16
function in the execution utilities. This has caused divergences in
how floats are truncated between code that uses this pass and code
which relies on __truncsfbf2 implementations, creating spurious
failures in conformance testing.


external/llvm-project/mlir/include/mlir/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.h
```diff
@@ -21,6 +21,9 @@ class Pass;
 #define GEN_PASS_DECL_CONVERTAMDGPUTOROCDL
 #include "mlir/Conversion/Passes.h.inc"
 
+/// Note: The ROCDL target does not support the LLVM bfleat type at this time
+/// and so this function will add conversions to change all `bfloat` uses
```
sjw36, line 25: typo 'bfleat'

external/llvm-project/mlir/lib/Conversion/GPUCommon/GPUOpsLowering.cpp
```diff
@@ -485,8 +485,8 @@ LogicalResult impl::scalarizeVectorOp(Operation *op, ValueRange operands,
     auto scalarOperands = llvm::map_to_vector(operands, extractElement);
     Operation *scalarOp =
         rewriter.create(loc, name, scalarOperands, elementType, op->getAttrs());
-    rewriter.create<LLVM::InsertElementOp>(loc, result, scalarOp->getResult(0),
-                                           index);
+    result = rewriter.create<LLVM::InsertElementOp>(
```
sjw36, line 488: good find. Looks like it was fixed upstream.

external/llvm-project/mlir/lib/Dialect/Arith/Transforms/ExpandOps.cpp
```diff
@@ -246,55 +246,29 @@ struct BFloat16TruncFOpConverter : public OpRewritePattern<arith::TruncFOp> {
...
-    Value shr = b.create<arith::ShRUIOp>(rounded, c16);
-    Value trunc = b.create<arith::TruncIOp>(i16Ty, shr);
+    Value bitcast = b.create<arith::BitcastOp>(i32Ty, operand);
+    Value isNaN =
+        b.create<arith::CmpFOp>(arith::CmpFPredicate::UNO, operand, operand);
+    Value sign =
+        b.create<arith::TruncIOp>(i1Ty, b.create<arith::ShRUIOp>(bitcast, c31));
+    Value nanVal = b.create<arith::SelectOp>(sign, sNaN, qNaN);
+
+    Value lsb =
```
sjw36, line 264: This looks pretty straight forward but probably going to want some comments for upstream.

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
```diff
@@ -122,28 +123,29 @@ void rock::buildKernelPipeline(OpPassManager &pm,
    *   --rock-gemm-to-gridwise --rock-regularize
    *   --rock-gridwise-gemm-to-blockwise
    */
-  pm.addNestedPass<func::FuncOp>(rock::createRockAffixTuningParametersPass(
+  auto &funcPm = pm.nest<func::FuncOp>();
```
sjw36, line 126: Does this consolidate the passes into 1 list?

sjw36 (approved):
Looks good.

external/llvm-project/mlir/lib/Conversion/GPUCommon/GPUOpsLowering.cpp
krzysz00: It's a backport of the upstream patch

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
krzysz00: I think it does from the point of view of printing the pass pipeline, so it shouldn't have a practical effect, but yeah.

PR #1173, opened 2023-07-25
MLIR#842: fixed mixed layout in rock.conv2d()

sjw36:
* when 1x1 filters get tosa.reshape then converted to collapse(NCHW) -> expand(NHWC)
* added unit test

https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/842

### DIFF
--- a/mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
+++ b/mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
@@ -564,7 +564,8 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
 //     %1 = expand(%0[KC]) -> [KHWC]
 // If this feeds into a conv2d as filter, we will drop the collapse/expand and
 // update the filter_layout attribute.
-struct CollapseExpandRewritePattern : public OpRewritePattern<tensor::ExpandShapeOp> {
+struct CollapseExpandRewritePattern
+    : public OpRewritePattern<tensor::ExpandShapeOp> {
   using OpRewritePattern<tensor::ExpandShapeOp>::OpRewritePattern;
 
   bool checkExpand(tensor::ExpandShapeOp expOp) const {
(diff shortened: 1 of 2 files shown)


mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -557,6 +557,75 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
...
+    LogicalResult lres = failure();
+    
+    Value expInp = expOp.getOperand();
+    Value expOut = expOp.getResult();
+
+    if (!checkExpand(expOp))
+      return failure();
+
+    auto colOp = expInp.getDefiningOp<tensor::CollapseShapeOp>();
+    if (colOp && checkCollapse(colOp)) {
```
krzysz00, line 606: Nit: we could get rid of `lres` and make this `if (!colOp || !checkCollapse(colOp)) { return failure(); }`

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -557,6 +557,75 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
...
+      auto colInp = colOp.getOperand();
+
+      for (Operation *usr : expOut.getUsers()) {
+        if (auto convOp = dyn_cast<tosa::Conv2DOp>(usr)) {
+          if (convOp.getOperand(1) == expOut) {
```
krzysz00, line 611: Do we want this pattern for input and output tensors too?

krzysz00 (commented):
Overall seems reasonable, nothing I commented is a blocker.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
sjw36: Still need it for the user loop, there could theoretically be multiple users.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
sjw36: Limiting to filters since they are probably the only 1x1 tensors we will see.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
krzysz00: Ah, missed that

We could still do the early return to save on a level of nesting

krzysz00 (approved):
Approved since there's one real nitpick left (inverting the collapse check if statement) but there's no reason to make you wait for me to hit approve after that's fixed.

PR #1179, opened 2023-08-03
Fix sorting of Embed{} inputs in vectorization

krzysz00:
Fixes https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1049

The issue found in that bug related to how the dimensions that were the arguments to Embed{} were sorted for processing. The old logic used the position in the embed arguments as the tiebreaker for arguments with equal coefficients, which caused missed vectorization opportunities as shown in the above bug.

This commit adds improved sorting logic to prevent this sort of thing.

### DIFF
--- a/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
+++ b/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
@@ -511,10 +511,48 @@ propagateVectorizationInfo(TransformMapAttr map, const VectorizationData &input,
     // isn't equal to the accumulated value times the smallest coefficient seen
     // on a dimension being vectorized.
     case TransformType::Embed: {
-      // Since embed coefficients can go in any order, we need them sorted
+      // Since embed coefficients can go in any order, we need them sorted.
+      // We first sort by the Embed coefficient
+      // We tiebreak dimensions with equal coefficients as follows:
+      //  - Dimensions that have vectorization info are processed before those
+      //    that don't.
+      //   - Two vectorization dimensions have ties broken by whether their
+      //      needed coefficient is equal to their embed coefficient.
+      //   - If both inputs have a matching coefficient, the one with the longer
+      //     vectorization length is used (sending equality to input order).
+      //   - Failing that, since neither of the inputs being compared are
+      //     going to succesfully advance the vectorizaion, we fall back to
+      //     order in the input map.
+      // - In all otther cases, ties are broken by position in the input map.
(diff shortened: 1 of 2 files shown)


mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -511,10 +511,48 @@ propagateVectorizationInfo(TransformMapAttr map, const VectorizationData &input,
...
+      //   - Failing that, since neither of the inputs being compared are
+      //     going to succesfully advance the vectorizaion, we fall back to
+      //     order in the input map.
+      // - In all otther cases, ties are broken by position in the input map.
       auto &&zip = llvm::zip(params, upperDims);
       SmallVector<std::tuple<int64_t, uint32_t>> data(zip.begin(), zip.end());
-      std::sort(data.begin(), data.end());
+      std::sort(data.begin(), data.end(), [&](const auto &a, const auto &b) {
+        int64_t paramA = std::get<0>(a), paramB = std::get<0>(b);
+        uint32_t dimA = std::get<1>(a), dimB = std::get<1>(b);
```
manupak: nit : you could do structured bindings or std::tie here.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -511,10 +511,48 @@ propagateVectorizationInfo(TransformMapAttr map, const VectorizationData &input,
...
+        bool aHasVec = input[dimA].has_value(),
+             bHasVec = input[dimB].has_value();
+        if (!aHasVec && !bHasVec)
+          return dimA < dimB;
+        if (aHasVec && !bHasVec)
+          return true;
+        if (!aHasVec && bHasVec)
+          return false;
+
+        bool aIsGood = input[dimA]->needsCoefficient == paramA,
```
manupak: nit : could use a better name here... as in I dont follow the relationship between being 'good' vs needed coefficient equating to embed coefficient.
Also it would be better use coeffA vs paramA to make things clearer...

manupak (commented):
This all looks good to me! also I checked with this commit I dont see missed vectorization opportunities anymore. Thanks!

I have naming nits here as a first timer looking at this part of the code.
Just curious -- so do we canonicalize merges to be embeds at some point ?

manupak (approved):
Im going to approve this (as they are naming nits)! In case you want to merge it in my night time.

manupak:
Thanks for the quick fix @krzysz00 !

PR #1180, opened 2023-08-03
Give validation kernel same type signature as test kernel (except -pv_with_gpu).

pcf000:
* Move the setting of opd1/opd2/result outside the loop-body function to allow for type conversion outside the loop nest.
    * Insert memcpy-f16-f32 ahead of CPU conv loop.
    * Insert deallocs and memcpy-f32-f16 ahead of return.
    * Similar changes for all CPU conv and gemm cases.
    * Exempt gpu validation from the same-signature principle, so it can make an f32 kernel for reference.

### DIFF
--- a/mlir/lib/Dialect/Rock/IR/RockDialect.cpp
+++ b/mlir/lib/Dialect/Rock/IR/RockDialect.cpp
@@ -457,9 +457,8 @@ ConvOpType mlir::rock::convOpTypeFromKernelType(KernelType kernelType) {
   case KernelType::Gemm:
     llvm_unreachable(
         "Gemm ops shouldn't be in convolution-specific lowering passes");
-  default:
-    llvm_unreachable("Unsuppported KernelType");
   }
+  llvm_unreachable("Unsuppported KernelType");
 }
 
 KernelType mlir::rock::kernelTypeFromConvOpType(ConvOpType convOpType) {
@@ -470,9 +469,8 @@ KernelType mlir::rock::kernelTypeFromConvOpType(ConvOpType convOpType) {
     return KernelType::Conv2DBwdData;
   case ConvOpType::BwdWeight:
     return KernelType::Conv2DBwdWeight;
-  default:
-    llvm_unreachable("Unsupported ConvOpType");
   }
+  llvm_unreachable("Unsupported ConvOpType");
 }
 
 GemmSize GemmSize::fromConvolution(ConvOpType type,
@@ -1462,9 +1460,6 @@ LogicalResult ThreadwiseReadIntoOp::verify() {
   else
     inputShape = extraViews[0].cast<TransformMapAttr>().getUpperBounds();
 
-  MemRefType srcType = getSource().getType().cast<MemRefType>();
-  gpu::AddressSpaceAttr srcMemSpaceAttr =
-      srcType.getMemorySpace().dyn_cast_or_null<gpu::AddressSpaceAttr>();
   size_t extraIdxCount = getExtraIndices().size();
   if (inputShape.size() != extraIdxCount + 1) {
     return emitOpError("source view must be extraIndices + 1");
@@ -1489,8 +1484,6 @@ LogicalResult ThreadwiseWriteAllOp::verify() {
   else
     outputShape = extraViews[0].cast<TransformMapAttr>().getUpperBounds();
 
-  MemRefType dstType = getDest().getType().cast<MemRefType>();
-  Attribute dstMemSpaceAttr = dstType.getMemorySpace();
   size_t extraIdxCount = getExtraIndices().size();
   if (outputShape.size() != extraIdxCount + 1) {
     return emitOpError("dest view must be extraIndices + 1");
--- a/mlir/tools/rocmlir-gen/rocmlir-gen.cpp
+++ b/mlir/tools/rocmlir-gen/rocmlir-gen.cpp
@@ -955,7 +955,6 @@ static func::FuncOp createGPUWrapper(ModuleOp module, const KernelIF &kernel) {
   auto kfunc = kernel.func;
   std::string funcName = kfunc.getName().str() + "_gpu";
   auto gpuWrapperFuncType = b.getFunctionType(kernel.params, {});
-
   auto gpuWrapperFunc =
       func::FuncOp::create(loc, StringRef(funcName), gpuWrapperFuncType);
   module.push_back(gpuWrapperFunc);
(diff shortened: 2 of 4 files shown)


mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -457,9 +457,8 @@ ConvOpType mlir::rock::convOpTypeFromKernelType(KernelType kernelType) {
   case KernelType::Gemm:
     llvm_unreachable(
         "Gemm ops shouldn't be in convolution-specific lowering passes");
-  default:
-    llvm_unreachable("Unsuppported KernelType");
   }
+  llvm_unreachable("Unsuppported KernelType");
```
pcf000, line 461: Warnings about default case when all possibilities explicit.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -1462,9 +1460,6 @@ LogicalResult ThreadwiseReadIntoOp::verify() {
   else
     inputShape = extraViews[0].cast<TransformMapAttr>().getUpperBounds();
 
-  MemRefType srcType = getSource().getType().cast<MemRefType>();
```
pcf000, line 1465: Unused variables.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -1268,6 +1267,152 @@ getConv2dBounds(rock::ConvOpType dir,
   return std::make_tuple(dim, dimH, dimW);
 }
 
+static func::FuncOp getMemcpyFuncDecl(ModuleOp module, const MemRefType srcType,
```
pcf000, line 1270: This function and emitMemcpy() moved up, unchanged, to be available for ensureFloatIsF32().

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -1268,6 +1267,152 @@ getConv2dBounds(rock::ConvOpType dir,
...
+
+  if (srcFlatType == dstFlatType) {
+    b.create<memref::CopyOp>(loc, srcFlat, dstFlat);
+  } else {
+    auto memcpyFunc = getMemcpyFuncDecl(module, srcFlatType, dstFlatType);
+    b.create<func::CallOp>(loc, memcpyFunc, ValueRange{srcFlat, dstFlat});
+  }
+}
+
+Value ensureFloatIsF32(OpBuilder &b, Location loc, Value ref, Type floatType) {
```
pcf000, line 1405: If the ref is float and not F32, make an F32 buffer and copy into it.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -1402,6 +1547,32 @@ createCPUConvWithMLIR(ModuleOp module, func::FuncOp &func,
     break;
   }
 
+  Value opd1, opd2, result;
+
+  switch (genConfig.operation.value()) {
+  case rock::ConvOpType::Fwd:
```
pcf000, line 1555: Split the block->getArgument parts of this switch out and ahead of the nested function, so we can ensure they're F32 if necessary.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -1553,8 +1714,29 @@ createCPUConvWithMLIR(ModuleOp module, func::FuncOp &func,
   affine::buildAffineLoopNest(b, loc, lowerBounds, upperBounds, steps,
                               createConv2dLoopNest);
 
+  if (!opd1.dyn_cast<BlockArgument>())
```
pcf000: Undo the F32-buffer copies at the end -- dealloc for arguments, copy back to blockArgument for the result.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -1572,10 +1754,20 @@ createCPUConvWithCPP(ModuleOp module, func::FuncOp &func,
     elemType = b.getI8Type();
   }
 
+  Value filter = block->getArgument(0);
```
pcf000, line 1760: Similar to above, potentially make F32 buffers for parameters ahead of some action.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -1746,6 +1938,16 @@ createCPUConvWithCPP(ModuleOp module, func::FuncOp &func,
                  paddingWidthRightConstantOp, dilationHeightConstantOp,
                  dilationWidthConstantOp, accelConstantOp});
 
+  if (!filter.dyn_cast<BlockArgument>())
```
pcf000: And undo at the end.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -1766,8 +1968,10 @@ createCPUConvFunc(ModuleOp module,
   OpBuilder b(module.getContext());
   auto loc = b.getUnknownLoc();
 
-  Type elemType = b.getF32Type();
-  Type outputElemType = b.getF32Type();
+  Type elemType = typeFromString(genConfig.inputDataTypeStr, module.getContext());
```
pcf000: Work with the kernel data types instead of forcing F32 (in two places).

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -1891,25 +2095,27 @@ static func::FuncOp createCpuGemmKernelWithMlir(ModuleOp module,
...
-      type = b.getF32Type();
-  }
-
   SmallVector<Type, 3> argTypes;
   getGemmTypes(cpuTypes, argTypes, /*isCpuVerifier=*/true);
 
   constexpr llvm::StringLiteral cpuKernName("host_naive_gemm");
   auto func =
       b.create<func::FuncOp>(loc, cpuKernName, b.getFunctionType(argTypes, {}));
+  module.push_back(func);
```
pcf000, line 2107: Move this up from the end, because emitMemcpy() traces back to the ModuleOp from inside the function.  Tried to make it consistent around the file.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2232,10 +2317,19 @@ static func::FuncOp createVerifierFunc(ModuleOp module, const KernelIF &kernel,
...
 
-    testResultNew = b.create<memref::AllocOp>(loc, valFlatType);
+    auto f32FlatType = MemRefType::get(valFlatType.getShape(), floatType);
+    testResultNew = b.create<memref::AllocOp>(loc, f32FlatType);
     emitMemcpy(b, testFlat, testResultNew);
-    testResult = b.create<memref::CastOp>(loc, mr1DUnkValType, testResultNew);
-    mr1DUnkTestType = mr1DUnkValType;
+    testResult = b.create<memref::CastOp>(loc, mr1DUnkF32Type, testResultNew);
+    mr1DUnkTestType = mr1DUnkF32Type;
+    if (!valElemType.isF32()) {
```
pcf000, line 2329: Treat the validation data exactly like the test-result data and promote to F32 before calling mcpuVerify.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2452,6 +2536,16 @@ static void insertValidationCalls(const GenParams &genParams, OpBuilder &b,
           valVars.resize(valVars.size() - 1);
         }
 
+        // Make the kernel and its wrapper function, and call it.
```
pcf000: I had moved this when messing with createGPUWrapper(), to have the parameter list trimmed when I touched it.  Now I don't mess with createGPUWrapper() and could undo this specific change.  It's only a block swap;  the only code change is removing the unused variable newConfig.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2644,7 +2738,11 @@ static LogicalResult populateHostHarnessLogic(
       } else if ((genValidation == "clone") || elemType.isInteger(8) ||
                  elemType.isInteger(32)) {
         valElemType = elemType;
+      } else if (!gpuValidation && isSmallFloatIn &&
```
pcf000, line 2744: Not GPU validation, small floats, and we're generating a kernel:  use the kernel's types for the validator.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
krzysz00: Instead of commenting this here, could you make this a comment on the function and mention briefly what the function does?

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
krzysz00: These `dyn_cast`s can be `isa<>`s instead

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -1553,8 +1714,29 @@ createCPUConvWithMLIR(ModuleOp module, func::FuncOp &func,
...
+    case rock::ConvOpType::BwdWeight:
+      resultBlockArg = block->getArgument(0);
+      break;
+    case rock::ConvOpType::BwdData:
+      resultBlockArg = block->getArgument(1);
+      break;
+    }
+
+    Value resultFlat = makeNDMemRef(b, result, 1);
+    emitMemcpy(b, resultFlat, resultBlockArg);
```
krzysz00, line 1738: Doesn't the result need to be free'd?

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2197,29 +2284,27 @@ static func::FuncOp createVerifierFunc(ModuleOp module, const KernelIF &kernel,
 
   // obtain function name of the verifier wrapper
   std::string verifyFuncName = "mcpuVerify";
-  if (valElemType.isF32()) {
+  if (valElemType.isa<FloatType>()) {
```
krzysz00: Why is this test relaxed, especially since below you assume `valElemType` is `f32`?

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
krzysz00: I think `newConfig` was needed to capture feature/type changes arising from creating the validation kernel. I'm a bit suspicious that it's going away

pcf000: It's a getter without side effects.  It used to be passed as argument to createGPUWrapper(), but that parameter was removed in March 2022 and newConfig was forgotten.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
pcf000: Yup.  In several other cases, too.  Changes coming.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
pcf000: Not sure where I'm assuming, but I think it's a hangover from when I thought the CPU could handle any float type.  Anyway, undone.

krzysz00 (approved):
Looks solid to me.

krzysz00:
... though, long-term thought: perhaps we should make `rocmlir-gen` work on tensors and then bufferize the results. Heck, that might work for the "move bufferization after gemm-to-gridwise" thing I've been eyeing for a while

yiqian1 (approved):

PR #1181, opened 2023-08-03
[DO NOT SQUASH] Large tensor support

krzysz00:
Fixes https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1032

This commit adds support for cases where a tensor holds more than 4 GB
of data or when any part of an index computation would overflow
INT32_MAX to rocMLIR.

The support is added as follows:
1. To save on redundancy, and to get rid of some weirdness that'll go
away once address space 7 works properly, remove rock.buffer_load and
rock.buffer_store in favor of rock.global_load/rock.global_store
1b. Update the global_load/store assembly format to reflect what
buffer_load/store used to have because that format's better

2. Update untransform() to return a third argument indicating whether
the untransformed tensor is "big" (intermediate dimension over
INT32_MAX or total storage over 4 GB).

All untransform() call sites now handle that value, either by setting
it as an attribute on the relevant operation or (for cases like LDS)
asserting that it can't be true.

3. global_load/global_store grew the needs64BitIdx attribute, which,
on a per load/store basis, records whether this memory operations
requires "index" to be a 64-bit integer for correctness

4. During blockwise-gemm-to-threadwise, after all global_load or
global_store ops have been produced, each function is walked to see if
any of the memory operations are `needs64BitIdx`. If they are, the
function itself is annotated with rock.64bitindex (this is safe
because between the annotation and -rock-to-gpu, we're only running
our passes or passes that run at the function level and don't touch
the function itself).

5. SugarToLoops has been modified to lower global_load/global_store
instead of buffer_load/buffer_store. This means:
- For 64-bit indexing or loads that don't require bounds checks, we
use regular memref operations (lowering to LLVM loads) after
memory_space_cast'ing the memory to the global address space. This
ensures that the backend doesn't use worse-performing flat_*
instructions.
- If a 64-bit indexing operation needs a bounds check that would
ordinarily be done with buffer ops, we use if statements instead
- In the marginal case where we can use 32 bit indexing but the buffer
has over 2 GB of stuff in it, we use buffer_* unconditionally because
that takes an unsigned offset unlike getelementptr .
- The logic for splitting up operations into the correct series of
buffer load operations has been simplified and moved into
sugar-to-loops.

5. Atomics are similarly handled with a combination of
atomic_rmw (corresponding directly to the LLVM) and buffer ops.

6. Since upstream doesn't support type puns on memrefs, and since we
don't use reduction fusion anyway (and since the compiler backend
doesn't do the right thing with fmax), the integer atomics hack for
fmax has been ***REMOVED***.

7. After all those changes, when doing a rock-to-gpu, we set the data
layout attribute on each kernel's GPU module, specifying the index
bitwidth based on whether rock.64bitindex is present.

8. The backend pipeline (more fundamentally, the gpu-to-rocdl
lowering) looks at the data layout to get the index width. This menas
that we do not set the index bitwidth override anymore.
8b. Note that, because of this, whether a kernel uses 32-bit or 64-bit
index is a *per-kernel* decision.

Tests have been added to cover this new functionality, and the PR
tests have been extended with a large GEMM
((32768 x 1) * (1 x 32769)) and the nightly tests have grown more
such GEMMs.


mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -171,6 +172,11 @@ static int64_t getUtilityVectorizationLen(ShapedType shape,
                         kMaxVectorOpLen * elemsPerWord);
 }
 
+static bool isLargeTensor(ShapedType type) {
+  return (type.getNumElements() * (type.getElementTypeBitWidth() / 8)) >=
+         ((int64_t)1 << (int64_t)32);
```
manupak: Cant we use std::numeric_limits here ?

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -237,6 +243,33 @@ LogicalResult createElementwiseLoop(
   return success();
 }
 
+/// Create a private buffer that can hold type `type`.
+static Value makeGpuAlloc(OpBuilder &b, Location loc, Type type) {
```
manupak: nit naming : makePrivateGpuAlloc

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
```diff
@@ -853,223 +857,333 @@ static Value computeMemRefNumElements(OpBuilder &b, Location loc,
...
+        loc, data, dest, coords, useBufferOobChecks, nullptr, nullptr);
+  return b.create<memref::AtomicRMWOp>(loc, AtomicRMWKind::maxf, data, dest,
+                                       coords);
+// Disabled because we can't make this hack work in general.
+#if 0
   }
+  Value dataAsInt = b.create<arith::BitcastOp>(loc, b.getI32Type(), data);
+  // Note: this doesn't work, and you'd need to add an upstream op which does
+  // this.
+  Value destAsInt = b.createOrFold<memref::CastOp>(
```
manupak, line 1024: Why do you need to do this ? 
Whilst I understand memref type punning is not a thing, our intrinsics dont require it -- maybe only memref::AtomicRMWOp needs it ? 

Is there underlying thing that Im missing here?

memref::AtomicRMWOp seems to be happily doing it with floats.

So why cant we just keep integer atomics approach there (which is only needs the type pun on the loaded val not necessarily on the memref) and fallback to memref::AtomicRMWOp with floats ?

manupak: We might need a non-binary `useBuffers` to indicate which buffer intrinsics to to use.

All in all, Im not against removing it but I just want to know why we did it -- and it is not obvious to me from this PR.

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
```diff
@@ -853,223 +857,333 @@ static Value computeMemRefNumElements(OpBuilder &b, Location loc,
...
+// because upstream doesn't have the ability to do type puns on a memref.
+// We don't know if the smax/umin trick is faster than a CAS loop,
+// so we're keeping it around in the comments below for performance evaluation
+// once we have a reduction-utilizing client.
+//
+// Returns the operation that the underlying global_store could be replaced
+// with.
+static Operation *makeAtomicFmax(PatternRewriter &b, Location loc, Value data,
+                                 Value dest, ArrayRef<Value> coords,
+                                 bool useBuffers, bool useBufferOobChecks) {
```
manupak: nit naming : it'd better to use useBufferIntrinsics as just buffers are too generic

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -46,15 +47,41 @@ mlir::rock::untransform(OpBuilder &b, Value transformed, ArrayAttr existing) {
...
+  auto bufferType = cast<ShapedType>(ret.getType());
+  int64_t elemBytes;
+  if (auto shapedTy = dyn_cast<ShapedType>(bufferType.getElementType()))
+    elemBytes =
+        (shapedTy.getNumElements() * shapedTy.getElementTypeBitWidth()) / 8;
+  else
+    elemBytes = bufferType.getElementTypeBitWidth() / 8;
+
+  if (bufferType.hasStaticShape())
+    isBig = (bufferType.getNumElements() * elemBytes) >=
```
manupak: this code was elsewhere as well.. might worth factoring to a general util

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -46,15 +47,41 @@ mlir::rock::untransform(OpBuilder &b, Value transformed, ArrayAttr existing) {
...
+std::tuple<Value, ArrayAttr, bool>
 mlir::rock::untransform(OpBuilder &b, Value transformed,
                         ArrayRef<Attribute> existing) {
   return untransform(b, transformed, b.getArrayAttr(existing));
 }
 
+bool mlir::rock::needs64BitIndices(TransformMapAttr map) {
+  // Negative lengths are some form of dynamic index and therefore big.
+  auto isBig = [](int64_t l) -> bool {
+    return l < 0 || l > (int64_t)(std::numeric_limits<int32_t>::max());
```
manupak, line 82: same here... we might factor this out

manupak (commented):
I did a initial look. Mostly nits/factoring suggestions.

However, I'd like to understand why you needed a memref cast; to me it feels like it is an artifact of using a binary switch `useBuffer` where it might as well indicate which intrinsics are available; so we can do the lowering.

I'd expect a hardware intrinsic to be faster than a SW cmp/swap loop but I might be wrong given that your discussions with the backend compiler team.

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
krzysz00: `memref::CastOp` doesn't allow for float-to-int type change.

And you **cannot** use buffers with tensors > 4 GB, so whatever we're doing has to work without buffers.

And our intrinsics _should_ be enforcing the type correctness there.

`atomicrmw` lowers to a CAS loop on architectures where there's no `fmax`.

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
krzysz00: I don't think there's a numeric_limits for 0x1_00_00_00_00

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
krzysz00: Where is the duplication?

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
manupak: ah ha... my bad then... why that value though ? just curious

manupak: hold on isn;t this equivalent to :  > std::numeric_limits<int32_t>::max()
Note : >= vs >

manupak: hence the duplication from below

giuseros:
From a first look, it looks ok to me. I need some more time to better understand what is going on, but won't object if Manupa approves this

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
manupak: > And you cannot use buffers with tensors > 4 GB, so whatever we're doing has to work without buffers.

Now I know.. and that makes this reasonable.

> And our intrinsics should be enforcing the type correctness there.

Sure.. hope it;d not be at the cost of perf though and it is not incorrect.
hope between us and compiler folks, someone could figure out a way generate code for that.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
manupak: {l > (int64_t)(std::numeric_limits<int32_t>::max() } == 
 {l >= 1 << 32} , right ?

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
krzysz00: ... no, it would be `> std::numeric_limits<uint32_t>::max()` ... note the u, and even then we'd need to cast to a longer type, so I might as well ignore that and write out 4 GB as a constant.

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
krzysz00: I mean, I think there's a solid correctness debate about that thing you came up with and NaN.

Also, no one's using this path.

And thirdly, the compiler often generates cas loops even when the instructions should exist and we don't quite know why and that's a problem for future us.

krzysz00: I'm also not convinced that your hack was actually faster? Branching on +/- is probably very non-uniform, while I'd think a CAS loop might be a uniform branch in a particular wave in a lot of cases

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
krzysz00: No, int32_t::max() is `(1 << 31) - 1`

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
manupak: > And you cannot use buffers with tensors > 4 GB, so whatever we're doing has to work without buffers.
> Now I know.. and that makes this reasonable.

I'd rest my case coz of this and the following : 

> Also, no one's using this path.

TLDR;

> I mean, I think there's a solid correctness debate about that thing you came up with and NaN.

Sure NaNs are an issue.. its just this being rocMLIR pass we didn't care as much. 
Even then I dont expect us to initialize a buffer to be filled with NaNs for reductions and we can 'potentially' handle a stVal being a NaN.

> I'm also not convinced that your hack was actually faster? Branching on +/- is probably very non-uniform, while I'd think a CAS loop might be a uniform branch in a particular wave in a lot of cases

https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/7180de65cfc407eec9ea6c13cdcc47d3b1f13917/external/llvm-project/mlir/lib/Dialect/AMDGPU/Transforms/EmulateAtomics.cpp#L103-L142

Im not convinced how the alternative (above ^) is faster either. I can appreciate with that we dont know perfectly about it. I am thinking above is what will be codegen'd if not please do correct me. 

In the current hack, I expected the branches to be balanced and divergence to only to be couple (closer to a single instruction in it).

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
manupak: yea you right, uint32_t::max() ?
(I just dont like these magic numbers; was looking a way to get rid of them)

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
manupak: 4GB constant also work. 

or 

As you correctly pointed out : 
`> (int64_t)std::numeric_limits<uint32_t>::max()`

krzysz00: I'm not seeing the problem with `2 ^ 32` directly?

krzysz00: ... but yeah, is a minor point, I can fix

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
krzysz00: Updating to `useBufferOps` to match the rest of the code

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
krzysz00: Yeah, that did need to be a utility, fixed.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -270,12 +270,15 @@ struct BlockwiseGemmRewritePattern
...
-    std::tie(matrixA, transformsA) =
+    bool ldsANeedsi64, ldsBNeedsi64;
+    std::tie(matrixA, transformsA, ldsANeedsi64) =
         untransform(b, adaptor.getMatrixA(),
                     b.getArrayAttr({splitTidAAttr, toLdsIndexAAttr}));
-    std::tie(matrixB, transformsB) =
+    std::tie(matrixB, transformsB, ldsBNeedsi64) =
         untransform(b, adaptor.getMatrixB(),
                     b.getArrayAttr({splitTidBAttr, toLdsIndexBAttr}));
+    assert(!ldsANeedsi64 && !ldsBNeedsi64 &&
```
manupak: Can't a `user`/`tuning` trigger this ? 
If so me thinks, should this be an EmitError ?

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -908,20 +831,31 @@ def Rock_GlobalLoadOp :
     Rock_Op<"global_load", [AllElementTypesMatch<["source", "result"]>]>,
     Arguments<(ins Arg<MemRefOf<SupportedMemoryElems>, "source memory", [MemRead]>:$source,
                    I1:$valid,
-                   Variadic<Index>:$sourceCoord)>,
+                   Variadic<Index>:$sourceCoord,
+                   UnitAttr:$needs64BitIdx,
+                   UnitAttr:$canReadOffEnd)>,
     Results<(outs AnyType:$result)> {
-  let summary = "Marker for one or more vectorized global loads";
+  let summary = "Load a vector from global memory, applying bounds checks";
```
manupak: nit : it is not necessarily a vector right?

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
```diff
@@ -200,7 +200,7 @@ void rock::buildBackendPipeline(OpPassManager &pm,
   floatEmuOpts.targetTypeStr = "f32";
   gpuPm.addPass(arith::createArithEmulateUnsupportedFloats(floatEmuOpts));
   gpuPm.addPass(createLowerGpuOpsToROCDLOpsPass(
-      options.chip, options.indexBitwidth, /*useBarePtrCallConv=*/true));
+      options.chip, /*indexBitwidth=*/0, /*useBarePtrCallConv=*/true));
```
manupak: Why is this set to zero now? might worth a comment here for future us.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -777,11 +634,13 @@ LogicalResult ThreadwiseReadIntoRewritePattern::matchAndRewrite(
     if (srcAddrSpace == gpu::AddressSpace::Global) {
       Value loaded = b.create<GlobalLoadOp>(
           loc, loadType, buffer, loadLoop.getValidity(/*domain=*/0),
-          loadLoop.getLowerCoords(/*domain=*/0));
+          loadLoop.getLowerCoords(/*domain=*/0), needs64BitIdx);
       b.create<InBoundsStoreOp>(loc, loaded, dest,
                                 loadLoop.getLowerCoords(
                                     /*domain=*/1)[extraIdxCount]);
     } else {
+      assert(!needs64BitIdx &&
```
manupak: Im not sure asserts are the right choice here... 
Cant a user trip over this ? or are they covered by verifiers ?

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -881,8 +741,12 @@ LogicalResult ThreadwiseWriteAllRewritePattern::matchAndRewrite(
                               outLoop.getLowerCoords(
                                   /*domain=*/0)[extraIdxCount],
                               outLoop.getValidity(/*domain=*/1),
-                              outLoop.getLowerCoords(/*domain=*/1));
+                              outLoop.getLowerCoords(/*domain=*/1),
+                              needs64BitIdx ? b.getUnitAttr() : nullptr,
+                              /*canStoreOffEnd=*/nullptr);
     } else {
+      assert(!needs64BitIdx &&
```
manupak: same here as well.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -115,7 +115,9 @@ bestGlobalVectorization(OpBuilder &b, Value matrix, int64_t copyDPerThread,
                         Type elementType) {
   Value tensor;
   ArrayAttr transforms;
-  std::tie(tensor, transforms) = untransform(b, matrix);
+  bool needs64Bit;
+  std::tie(tensor, transforms, needs64Bit) = untransform(b, matrix);
+  (void)needs64Bit;
```
manupak: use std::ignore here for needs64Bit

manupak (commented):
Looking good! I dont have major concerns. 
Just a question about asserts and a minor nit pick on docs.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: I don't think any combination of tuning parameters we have could set this off.

I also think anyone trying to manually craft an example that sets this off would hit checks in gridwise-gemm-to-blockwise.

krzysz00: Moved to a match failure just to be safe

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
krzysz00: Updated to use the actual name of the constant

manupak (approved):
LGTM!

PR #1182, opened 2023-08-04
Add `inbounds` to gep instructions

ravil-mobile:
Closes ROCmSoftwarePlatform/rocMLIR-internal#1038

Mark all `GEP` operations inbounds


ravil-mobile:
Seems like the pass finds `GEPs` in a kernel

```
> rocmlir-driver -c ~/project/digest/ex1/rocmlir-gen.txt -o  /dev/null
llvm.func: attrs: , "CConv", "block_size", "function_type", "gpu.kernel", "gpu.known_grid_size", "grid_size", "linkage", "rocdl.flat_work_group_size", "rocdl.kernel", "rocdl.reqd_work_group_size", "rocdl.waves_per_eu", "sym_name", "visibility_"
%52 = llvm.getelementptr %51[0, 0] : (!llvm.ptr<3>) -> !llvm.ptr<3>, !llvm.array<2048 x i8>
%63 = llvm.getelementptr %62[0, 0] : (!llvm.ptr<3>) -> !llvm.ptr<3>, !llvm.array<1024 x i8>
%231 = llvm.getelementptr %230[%171] : (!llvm.ptr<5>, i32) -> !llvm.ptr<5>, f32
%271 = llvm.getelementptr %270[%171] : (!llvm.ptr<5>, i32) -> !llvm.ptr<5>, f32
%307 = llvm.getelementptr %306[%164] : (!llvm.ptr<5>, i32) -> !llvm.ptr<5>, f32
%313 = llvm.getelementptr %312[%171] : (!llvm.ptr<5>, i32) -> !llvm.ptr<5>, f32
%316 = llvm.getelementptr %315[%171] : (!llvm.ptr<5>, i32) -> !llvm.ptr<5>, f32
%324 = llvm.getelementptr %323[%321] : (!llvm.ptr<5>, i32) -> !llvm.ptr<5>, f32
%327 = llvm.getelementptr %326[%321] : (!llvm.ptr<5>, i32) -> !llvm.ptr<5>, f32
%334 = llvm.getelementptr %333[%171] : (!llvm.ptr<3>, i32) -> !llvm.ptr<3>, i8
%346 = llvm.getelementptr %345[%171] : (!llvm.ptr<3>, i32) -> !llvm.ptr<3>, i8
%361 = llvm.getelementptr %360[%171] : (!llvm.ptr<5>, i32) -> !llvm.ptr<5>, f32
%364 = llvm.getelementptr %363[%359] : (!llvm.ptr<3>, i32) -> !llvm.ptr<3>, f32
```

where `~/project/digest/ex1/rocmlir-gen.txt` is a dump of 
```
rocmlir-gen -g 3 -m 1024 -k 769 -n 1024 --transA=false -t f32  --operation gemm --arch gfx908:sramecc+:xnack- -pv 
```

ravil-mobile:
Seems like the pass finds `GEPs` in a kernel. For example,

```
> rocmlir-driver -c ~/project/digest/ex1/rocmlir-gen.txt -o  /dev/null
llvm.func: attrs: , "CConv", "block_size", "function_type", "gpu.kernel", "gpu.known_grid_size", "grid_size", "linkage", "rocdl.flat_work_group_size", "rocdl.kernel", "rocdl.reqd_work_group_size", "rocdl.waves_per_eu", "sym_name", "visibility_"
%52 = llvm.getelementptr %51[0, 0] : (!llvm.ptr<3>) -> !llvm.ptr<3>, !llvm.array<2048 x i8>
%63 = llvm.getelementptr %62[0, 0] : (!llvm.ptr<3>) -> !llvm.ptr<3>, !llvm.array<1024 x i8>
%231 = llvm.getelementptr %230[%171] : (!llvm.ptr<5>, i32) -> !llvm.ptr<5>, f32
%271 = llvm.getelementptr %270[%171] : (!llvm.ptr<5>, i32) -> !llvm.ptr<5>, f32
%307 = llvm.getelementptr %306[%164] : (!llvm.ptr<5>, i32) -> !llvm.ptr<5>, f32
%313 = llvm.getelementptr %312[%171] : (!llvm.ptr<5>, i32) -> !llvm.ptr<5>, f32
%316 = llvm.getelementptr %315[%171] : (!llvm.ptr<5>, i32) -> !llvm.ptr<5>, f32
%324 = llvm.getelementptr %323[%321] : (!llvm.ptr<5>, i32) -> !llvm.ptr<5>, f32
%327 = llvm.getelementptr %326[%321] : (!llvm.ptr<5>, i32) -> !llvm.ptr<5>, f32
%334 = llvm.getelementptr %333[%171] : (!llvm.ptr<3>, i32) -> !llvm.ptr<3>, i8
%346 = llvm.getelementptr %345[%171] : (!llvm.ptr<3>, i32) -> !llvm.ptr<3>, i8
%361 = llvm.getelementptr %360[%171] : (!llvm.ptr<5>, i32) -> !llvm.ptr<5>, f32
%364 = llvm.getelementptr %363[%359] : (!llvm.ptr<3>, i32) -> !llvm.ptr<3>, f32
%375 = llvm.getelementptr %374[%171] : (!llvm.ptr<5>, i32) -> !llvm.ptr<5>, f32
%378 = llvm.getelementptr %377[%373] : (!llvm.ptr<3>, i32) -> !llvm.ptr<3>, vector<2xf32>
```

where `~/project/digest/ex1/rocmlir-gen.txt` is a dump of

```
rocmlir-gen -g 3 -m 1024 -k 769 -n 1024 --transA=false -t f32  --operation gemm --arch gfx908:sramecc+:xnack- -pv 
```

mlir/include/mlir/Dialect/Rock/Passes.td
```diff
@@ -98,4 +98,8 @@ def RockLowerReducePass : Pass<"rock-lower-reduce", "::mlir::func::FuncOp"> {
   let dependentDialects = ["rock::RockDialect", "func::FuncDialect", "gpu::GPUDialect"];
 }
 
+def RockMakeGepsInboundsPass : Pass<"rock-make-geps-inbounds", "::mlir::LLVM::LLVMFuncOp"> {
```
krzysz00: It might be reasonable to make set this value to `""` so this can be run on any root.

mlir/lib/Dialect/Rock/Transforms/MakeGepsInbounds.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+// Remove these includes after we're not working with a test pass upstream
+#include "mlir/Analysis/DataFlow/DeadCodeAnalysis.h"
+#include "mlir/Analysis/DataFlow/IntegerRangeAnalysis.h"
+#include "mlir/Transforms/FoldUtils.h"
+
+
+namespace mlir {
+namespace rock {
+// TODO: ask why?
```
krzysz00: Magic incantations that let us autogenerate a _lot_ of boilerplate.

mlir/lib/Dialect/Rock/Transforms/MakeGepsInbounds.cpp
```diff
@@ -0,0 +1,64 @@
...
+void RockMakeGepsInboundsPass::runOnOperation() {
+  auto func = getOperation();
+  llvm::outs() << func->getName() << ": attrs: ";
+  for (auto attr: func->getAttrs()) {
+    llvm::outs() << ", " << attr.getName();
+  }
+  llvm::outs() << '\n';
+
+  func.walk([](LLVM::GEPOp gepOp) {
+    llvm::outs() << gepOp << '\n';
```
krzysz00: ... ah, right, this is a draft.

Here, you'll probably just want to add the `inbounds` attribute (I forget if it's a boolean or a `UnitAttr`)

mlir/lib/Dialect/Rock/Transforms/MakeGepsInbounds.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include "mlir/Dialect/Arith/Transforms/Passes.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/Rock/Passes.h"
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "mlir/Pass/PassManager.h"
+#include "mlir/Transforms/Passes.h"
+
+// Remove these includes after we're not working with a test pass upstream
+#include "mlir/Analysis/DataFlow/DeadCodeAnalysis.h"
```
krzysz00: These  you'll want to remove right now

krzysz00 (commented):
Figured I'd toss in some comments on what I'm seeing so far

Glad you managed to find all the relevant infrastructure so quickly!

mlir/lib/Dialect/Rock/Transforms/MakeGepsInbounds.cpp
ravil-mobile: Forgot to delete) Thanks for answering!

mlir/lib/Dialect/Rock/Transforms/MakeGepsInbounds.cpp
ravil-mobile: yes, it is a draft. I just decided to share with you and @manupak. I mean I just wanted to be sure that I am moving to a correct direction.

Yes, I will just add `inbounds` attribute then.

mlir/lib/Dialect/Rock/Transforms/MakeGepsInbounds.cpp
ravil-mobile: Copied it from `CleanMath`. Took it as a template)

mlir/include/mlir/Dialect/Rock/Passes.td
ravil-mobile: let  me try

ravil-mobile: done

mlir/lib/Dialect/Rock/Transforms/MakeGepsInbounds.cpp
ravil-mobile: done

mlir/include/mlir/Dialect/Rock/Passes.td
```diff
@@ -98,4 +98,8 @@ def RockLowerReducePass : Pass<"rock-lower-reduce", "::mlir::func::FuncOp"> {
   let dependentDialects = ["rock::RockDialect", "func::FuncDialect", "gpu::GPUDialect"];
 }
 
+def RockMakeGepsInboundsPass : Pass<"rock-make-geps-inbounds", ""> {
```
manupak: you ll probably need to define this as a func pass.
Have a look at other such passes

mlir/lib/Dialect/Rock/Transforms/MakeGepsInbounds.cpp
```diff
@@ -0,0 +1,56 @@
...
+namespace {
+struct RockMakeGepsInboundsPass
+    : public rock::impl::RockMakeGepsInboundsPassBase<RockMakeGepsInboundsPass> {
+  void runOnOperation() override;
+};
+} // end namespace
+
+
+void RockMakeGepsInboundsPass::runOnOperation() {
+  auto func = getOperation();
```
manupak: Lets use the FuncOp type here.. I think it will only work if the pass was designated to run on func ops. See above

mlir/lib/Dialect/Rock/Transforms/MakeGepsInbounds.cpp
manupak: We might get away doing that as it is a simple pass rather doing OpRewrite here...

manupak (commented):
Few comments about declaring the pass to work on func ops.

mlir/lib/Dialect/Rock/Transforms/MakeGepsInbounds.cpp
krzysz00: I disagree

I'd like the type here to be `Operation *`

mlir/include/mlir/Dialect/Rock/Passes.td
krzysz00: Nah, per my comments, `""` (any root operation) is sufficient.

krzysz00: And enables scheduling onto `gpu.module` without weird implicit nesting stuff.

While calling for `LLVM::FuncOp` might enable parallelism in general, we've got one function per module so

krzysz00:
Also, having looked at the autogenerated bits, `setInbounds(true)` is a valid method on GEPs

mlir/lib/Dialect/Rock/Transforms/MakeGepsInbounds.cpp
manupak: What happens if we have nested modules? i remember seeing them in xmir sometime back.. @sjw36 might know better 

It feels too broad on checking the existence of kernel attr on any root op where we perfectly know we need to do this only on funcs

krzysz00: I'm arguing we shouldn't be checking for the kernel attribute either.

This could be run on host code just as well in the future.

krzysz00: Though we could go for `LLVM::FuncOp` and then do an `addNestedPass`.

It's not too important

manupak: Hmmmm interesting.. are we sure about all the host code especially in tosa partitioned scenario?

krzysz00: This is pretty hypothetical, and I'm not planning to enable it currently, but yeah, I'm confident that high-level Tosa code won't lower to array indexing violations.

ravil-mobile:
> Also, having looked at the autogenerated bits, `setInbounds(true)` is a valid method on GEPs

Thanks. I was looking for something like that in `llvm-project/mlir/include/mlir/Dialect/LLVMIR' but I failed to find. I also checked my build directory (I thought there will something generated with tablegen) I and still couldn't find it.

ravil-mobile:
Thanks @manupak,

> We might get away doing that as it is a simple pass rather doing OpRewrite here...

As @krzysz00 mentioned, we can just simply change the attribute of each `gep`. The following worked

https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/d70096a37770c900898e33be07ba4c6bab4556ce/mlir/lib/Dialect/Rock/Transforms/MakeGepsInbounds.cpp#L44-L47

krzysz00:
@ravil-mobile  Re where `setInbounds()` was (I found it via go-to-definition), it was in `build/external/llvm-project/llvm/tools/mlir/include/mlir/Dialect/LLVMIR/LLVMOps.h.inc` , which is, you'll note' a rather nonobvious path (`external/llvm-project/llvm/tools/mlir` as opposed to `external/llvm-project/mlir` in the source is a weird build system quirk)

krzysz00:
As a broad comment, could you rename this pass to something like `rock-add-assumptions-for-llvm`, since there's a few more things we'll want to stick in here later, like fixing up the `align` value on loads and stores of vectors.

manupak:
Its not a pass acting on rock dialect... so how about something similiar to : https://mlir.llvm.org/docs/Passes/#-llvm-optimize-for-nvvm-target. Can we call it to be optimize for amdgpu-target ?

manupak:
or rocdl maybe?

ravil-mobile:
> Its not a pass acting on rock dialect... so how about something similiar to : https://mlir.llvm.org/docs/Passes/#-llvm-optimize-for-nvvm-target. Can we call it to be optimize for amdgpu-target ?

@manupak, makes sense. @krzysz00, what do you think?

krzysz00:
@manupak I see your point, but this isn't a general optimization pass that anyone should be running. This is specifically something we won't be upstreaming because it's something of a footgun if you're not very confident that you're generating correct code.

I do see the point on `rock` though.

... come to think of it, there's a pass or two in the `Dialects/Rock/Transforms/` directory that doesn't really belong there - the math cleaner, for one thing.

So, with all that in mind, how about `rocmlir-optimize-kernel-llvm`

(and we can have a future todo item to actually make a `Transforms/` directory at the top level that holds the passes we define against upstream)

(@ravil-mobile )

ravil-mobile:
thanks @krzysz00. @manupak, what do you think about `rocmlir-optimize-kernel-llvm` ? I personally like it

manupak:
> , but this isn't a general optimization pass that anyone should be running. This is specifically something we won't be upstreaming because it's something of a footgun if you're not very confident that you're generating correct code.

I agree with this sentiment but I think it could live as a Transform pass in llvm dialect of MLIR that we dont intend to upstream.
I dont think we'll have collisions with that ?

Therefore, it might as well be llvm-optimize-for-rocmlir-kernels

krzysz00:
I'm in agreement with `rocmlir-optimize-kernel-llvm` still.

@manupak I want these passes to live in our directory structure (which I keep meaning to rename to `rocmlir/` instead of `mlir/`) instead of `external/llvm-project/` because keeping things in `external/llvm-project/` that aren't meant to be upstreamed is bad and will keep messing up future upstream merges.

ravil-mobile:
> because keeping things in external/llvm-project/ that aren't meant to be upstreamed is bad and will keep messing up future upstream merges.

Sounds reasonable.

manupak:
ok, I dont feel as strongly to object further.

The issue I still see is a its a transform pass on LLVM dialect and it felt to be belong to that; irrespective of whether we upstream it or not. In a summary, its a downstream pass acting on a upstream dialect.

I think directory structure is orthogonal because my argument is based on the dialect that transform pass belong not based on the project it belongs.

> instead of external/llvm-project/ because keeping things in external/llvm-project/ that aren't meant to be upstreamed is bad and will keep messing up future upstream merges.

I'd agree with this statement in the general case.

But if we think about 'how' it messes up with upstream merges, it would be still be a downstream pass acting on a upstream dialect irrespective of the directory we put that in. Therefore any changes to the upstream dialect will equate same amount of 'messing', IMHO.

mlir/lib/Dialect/Rock/Transforms/MakeGepsInbounds.cpp
```diff
@@ -0,0 +1,47 @@
+//===- CleanMath.cpp - Clean up math after lowering/unrolling loops  ---===//
```
ravil-mobile: fix header

ravil-mobile:
Hi @krzysz00 , @manupak.

I renamed `MakeGepsInbounds` to `RockPrepareLLVM`. What do you think?

manupak (commented):
Broadly looks good! I think we'd need a simple unit test (LIT based)

krzysz00:
Yeah, agreed on needing a test.

ravil-mobile:
> Broadly looks good! I think we'd need a simple unit test (LIT based)

> Yeah, agreed on needing a test.

Agree with @krzysz00  and @manupak . Do we have a similar test in our code base, or the one I can look as a reference?

manupak:
https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/7180de65cfc407eec9ea6c13cdcc47d3b1f13917/external/llvm-project/mlir/test/Dialect/LLVMIR/optimize-for-nvvm.mlir#L1-L24

ravil-mobile:
@krzysz00 , @manupak , Is this test ok?

https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/e47e28215f02e361c86707c839a87ceaf7b1042d/mlir/test/Dialect/Rock/prepare_llvm.mlir#L1-L11

krzysz00:
You'll want to use variables instead of hard-coding the names like you did.
I'd also argue for more descriptive variable names.

Also, LLVM's moved to opaque pointers, so you can lose the `<f32>` on the pointer types.
Something like
```mlir
// CHECK-LABEL: @access_ptr
// CHECK-SAME: (%[[:BASE.+]]: !llvm.ptr, %[[IDX:.+]]: i64)
llvm.func @access_ptr(%base: !llvm.ptr<f32>, %idx i64) -> () { 
  // CHECK: = llvm.getelementptr inbounds %[[BASE][%[[IDX]]]
  %p0 = llvm.getelementptr %base[%idx] : (!llvm.ptr, i64) -> !llvm.ptr, f32
  %c1 = llvm.mlir.constant(1) : i64 
  // CHECK: %[[NEXT:.+]] = llvm.add %[[IDX]]
  %next = llvm.add %arg1, %1 : i64
  // CHECK: = llvm.getelementptr inbounds %[[BASE]][%[[NEXT]]]
  %p2 = llvm.getelementptr %arg0[%next] : (!llvm.ptr, i64) -> !llvm.ptr, f32
  llvm.return 
} 
```

ravil-mobile:
Hi @krzysz00 ,

> You'll want to use variables instead of hard-coding the names like you did.
I'd also argue for more descriptive variable names.

I thought it was not important for the test. But, I can see your point; the named-variables look better.

manupak (approved):
LGTM!

PR #1185, opened 2023-08-08
Improve rocmlir-driver debugging, sanity testing

krzysz00:
- Add a `-dump-pipelines` option to rocmlir-driver to show the pass pipelines we're running in various driver invocations in text form. This'll help us make sure we're accurately running portions of the pipeline when we want to stop early. (This representation should be passed to the `-pass-pipeline` command-line argument in either rocmlir-driver or rocmlir-opt)

Similarly, add tests that encode the current state of the various rocmlir pipelines. These ensure that we don't accidentally change the pipelines.

While I was there, fix various pipelines to use fewer nested pass managers and have one longer funcOp pipeline.

Also, change rock-loops-to-cf to be a func pass like it logically should be.

I have also removed some passes from the "kernel" pipeline that were meant to run on host code and have been subsumed into the MHAL runner pipeline. This means that our kernel pipeline will, like we all expect, end on -convert-rock-to-gpu .

---

I have also added the --verify-passes option to rocmlir-driver and used to enable verification in rocmlir pipelines, but not by default. When --verify-passes is set, this will causes rocmlir-driver to act like rocmlir-opt and check for IR validity in between each psss.

I have used this option to massively simplify our sanity tests, which we have not been keeping up to date with new pipeline changes. This also means the sanity tests will stop being O(N^2).

Fixes
https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1040


djramic (approved):
LGTM!

sjw36:
Very cool @krzysz00 ! Thanks

PR #1186, opened 2023-08-08
Erase dead transforms during rock-regularize

krzysz00:
Fixes
https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/940 and is highly likely to resolve
https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1072 (but that'll require testing).

The crashes were tracable to branching read chains (a transform op with multiple users), which are an issue that we punted on. However, we did not need to resolve this issue in this commit to fix the bug.

The underlying issues was that there were dead transform ops (the originals of the operations we inverted) cluttering up use lists and causing the "a buffer is read multiple times with different transform chains, we never wrote that code" detector to fire spuriously. This commit explicitly erases transform ops post-inversion in order prevent this crash.

The actual assertion failure came from the fact that the multiple-readers detector would simply cut off the chain collection early, and then the assert would observe that we never found an actual reader in at least one of the chains. It wasn't the most meaningful assert, but it did do its job at spotting an issue.

We also witch up the rock-regularize pattern to collect the inverses before creating any new IR, in the interests of debugability and the fact that upstream MLIR would really prefer we not modify IR in failing patterns (perhaps we should enable those checks in debug builds...).

We also add the original bug to our nightly testing.


mlir/test/fusion/nightly-misc-e2e/issue-940.mlir
```diff
@@ -0,0 +1,24 @@
...
+    %7 = migraphx.slice(%6) {axes = [1], ends = [24], starts = [12]} : (tensor<1x36x384x64xf32>) -> tensor<1x12x384x64xf32>
+    %8 = migraphx.transpose(%7) {permutation = [0, 1, 3, 2]} : (tensor<1x12x384x64xf32>) -> tensor<1x12x64x384xf32>
+    %9 = migraphx.transpose(%4) {permutation = [0, 1, 3, 2]} : (tensor<1x12x384x64xf32>) -> tensor<1x12x64x384xf32>
+    %10 = migraphx.transpose(%8) {permutation = [0, 1, 3, 2]} : (tensor<1x12x64x384xf32>) -> tensor<1x12x384x64xf32>
+    %11 = migraphx.dot(%10, %9) : (tensor<1x12x384x64xf32>, tensor<1x12x64x384xf32>) -> tensor<1x12x384x384xf32>
+    %12 = migraphx.transpose(%11) {permutation = [0, 1, 3, 2]} : (tensor<1x12x384x384xf32>) -> tensor<1x12x384x384xf32>
+    %13 = migraphx.mul(%12, %1) : (tensor<1x12x384x384xf32>, tensor<1x12x384x384xf32>) -> tensor<1x12x384x384xf32>
+    return %13 : tensor<1x12x384x384xf32>
+  }
+}
```
manupak: super nit : newline

manupak (approved):
Good job in figuring this out so quickly!

PR #1187, opened 2023-08-09
Replace createLDSStore loop with threadwise_write_all

manupak:
Currently, the gridwise lowering creates a transforming
for loop to store LDS for gemms. This could be avoided
by using threadwise_write_all op that can target dest
with vector-typed elements.

Therefore, this commit adds support to threadwise_write_all
ops to target dest with vector-typed elements as well.

closes : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1073


manupak:
[PR1187_perf.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/12299715/PR1187_perf.zip)

perfs are flat as well (just checked in case..)

krzysz00 (approved):
Looks like a solid cleanup to me!

manupak:
Thanks @krzysz00!

PR #1189, opened 2023-08-10
Refactor AccelEmitter to create RegsAsMatrixSubTiles

manupak:
We have introduced composite structue to hold views to view register memref as a sub-tile of a matrix in a prior refactor.
This commit further uses this structure to generate such views for output registers of accelerators.

closes : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1060


krzysz00 (approved):
This looks pretty reasonable to me, and I'm not seeing any big reason not to land this.

(And this isn't quite the time to nitpick about code duplication, as that's a cleanup for Later (tm))

manupak:
Thanks @krzysz00 !

PR #1193, opened 2023-08-11
Add configurations for dense gemms

giuseros:
This is to cover and evaluate the configs in https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1024


manupak (approved):
LGTM!

PR #1194, opened 2023-08-11
Add error check for unsupported WMMA configuration (e.g., kouter*kpack < 16)

giuseros:
This is a check that is present in the `mfma` branch but I probably forgot to add it in the `wmma` branch.

### DIFF
--- a/mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
+++ b/mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
@@ -627,6 +627,19 @@ PopulateParamsWmma::isValidBlockwiseGemm(const InitParamsAccel &param,
   }
 
   // Reject invalid KPACK values.
+  auto maybeWmmaInsn = WmmaInsn::select(dataTypeA, dataTypeB, waveSize,
+                                        param.gemmMPerWave, param.gemmNPerWave);
+  if (failed(maybeWmmaInsn)) {
+    LLVM_DEBUG(llvm::dbgs() << "Failed to select wmma instruction.\n");
+    return failure();
+  }
+  WmmaInsn wmmaInsn = *maybeWmmaInsn;
+  if (!wmmaInsn.isCoherentWithK(param.gemmKPack, param.gemmKPerBlock)) {
+    LLVM_DEBUG(llvm::dbgs()
+               << "Wmma instruction selection is not compatible with k.\n");
+    return failure();
+  }
+
   return success();
 }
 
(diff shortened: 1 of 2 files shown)


giuseros:
Fixes https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1078

yiqian1:
Have you tried this change with the develop branch? I ran the test case in the issue#1078 and got the error "Getting kernel count failed". With your branch, rocmlir-driver had lowering failure. However, both failures happened at the check wmmaInsn.isCoherentWithK().

giuseros:
> Have you tried this change with the develop branch? I ran the test case in the issue#1078 and got the error "Getting kernel count failed". With your branch, rocmlir-driver had lowering failure. However, both failures happened at the check wmmaInsn.isCoherentWithK().

Hi Yi, yes, this is the goal of the PR. Wmma perf configs where `kPerBlock  <16` and `kPack==1` should fail, in the same way they fail for mfma when `kPerBlock < attr.k_base` . The perf config in https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1078 is indeed: `256,128,8,64,128,1,1,1` and `kPerBlock=8<16`

yiqian1:
> Hi Yi, yes, this is the goal of the PR. Wmma perf configs where kPerBlock  <16 and kPack==1 should fail, in the same way they fail for mfma when kPerBlock < attr.k_base . The perf config in https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1078 is indeed: 256,128,8,64,128,1,1,1 and kPerBlock=8<16

I think we should not use this perf_config for wmma. This is an E2E test and should not fail. I don't remember why we had this test in the first place. We probably can remove this test. But for this PR, I'd suggest use an applicable perf_config, or mark it as XFAIL for wmma, or simply disable this test for wmma.

giuseros:
> But for this PR, I'd suggest use an applicable perf_config, or mark it as XFAIL for wmma, or simply disable this test for wmma.

Do you want me to do in this PR? Or are you going to do that while enabling tests for wmma?

krzysz00:
This seems like a reasonable check to add and will enable better tuning for MIGraphX.

yiqian1:
> > But for this PR, I'd suggest use an applicable perf_config, or mark it as XFAIL for wmma, or simply disable this test for wmma.
> 
> Do you want me to do in this PR? Or are you going to do that while enabling tests for wmma?

No need to address it in this PR. Apologies for the confusion, as I had assumed wmma was already enabled (given my recent work on it), and this PR would cause failure of the nightly job.

yiqian1 (approved):

PR #1197, opened 2023-08-14
[DO NOT SQUASH] Fix SetDefaultDeviceOp rewrite, suppress rm warning in tuningRunner.py

pcf000:
The SetDefaultDeviceOp rewriter looks like a simple oversight -- it doesn't pass the replacement op to the replaceOp() call.

tuningRunner.py removes results.stats.csv, which may not exist;  add "-f" so it doesn't warn about the missing file.


external/llvm-project/mlir/lib/Conversion/GPUCommon/GPUToLLVMConversion.cpp
```diff
@@ -1333,8 +1333,9 @@ LogicalResult ConvertSetDefaultDeviceOpToGpuRuntimeCallPattern::matchAndRewrite(
     gpu::SetDefaultDeviceOp op, OpAdaptor adaptor,
     ConversionPatternRewriter &rewriter) const {
   Location loc = op.getLoc();
-  setDefaultDeviceCallBuilder.create(loc, rewriter, {adaptor.getDevIndex()});
-  rewriter.replaceOp(op, {});
```
pcf000, line 1337: Nobody else in MLIR passes {} to replaceOp().

giuseros: I think the original intention was to delete `op`, but I guess that since the `setDefaultDeviceCall` is returning nothing, this should be equivalent.

giuseros (approved):
LGTM!

krzysz00:
Has this been upstreamed?

pcf000:
Not yet.  I wanted to double-check with you, since it was your code originally (from February 2022).
I bumped into it by using --device with rocmlir-gen, trying to get Tuna and tuningRunner.py to distribute work among GPUs.

krzysz00:
Yeah, go ahead and send this upstream

PR #1198, opened 2023-08-14
Fixed rock.blockwise_broadcast_reduce wrong views

manupak:
The lower views of the virtual distributed
tensor was incorrect. Therefore it was reducing in wrong axis.

This commit fixes that and adds a bit more barriers to be safe between LDS stores and reads.

closes : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1082
(though I added the code to handle it, it turned out due to not taking lower views correctly)


krzysz00:
Looks like you've got a test that depends on the wave size in some interesting fashion and so it's failing on Navi.

manupak:
yep! that is my assumption so far as well.. looking into it

manupak:
fixed now.

krzysz00 (approved):

manupak:
Thanks @krzysz00!

PR #1199, opened 2023-08-15
Fix embed vector length alignment inference

manupak:
There was a bug in when propagating alignment
information where the coefficients are the same.
It was exposed after landing https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1164 where we are using vector length inference upon viewing it for global->regs loading. 
The bug was exposed in the weekly CI (https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1083)

This commit fixes that to use the minimum alignment when same embed coefficients are used.

### DIFF
--- a/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
+++ b/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
@@ -578,6 +578,22 @@ propagateVectorizationInfo(TransformMapAttr map, const VectorizationData &input,
       // of 1
       bool hasNegativeCoefficients = false;
       std::optional<VectorizationInfo> ourResult;
+
+      // When there are equal embed coefficients
+      // we should always refer to the one with lowest
+      // alignment as the alignment
+      llvm::SmallDenseMap<int64_t, int64_t> coeffToAlignment;
+      for (auto [coefficient, upperDim] : data) {
+        if (input[upperDim].has_value()) {
+          if (coeffToAlignment.count(coefficient)) {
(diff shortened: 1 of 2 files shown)


manupak:
As a first timer in this part of the code, I'd appreciate a double check here. 

*I think* -- as I've understood the concept of "alignment" (thanks to @giuseros) -- is that we should be always using the minimum one where indices just get added (i.e. iso embed coefficients) because when one of such dimensions change in 'aligned' manner it adds offset if we are vectorizing on another -- which might screw up the alignment on the the dimension that is being vectorized, especially if there is padding involved.

This seems to fix the reproducer  in https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1083 and a couple of failing tests in the  log : http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir-weekly/detail/mlir-weekly/182/pipeline/229

manupak:
Running a weekly with parameterSweeps enabled : http://rocmhead:8080/job/MLIR/job/mlir/job/PR-1199/2

manupak:
http://rocmhead:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-1199/2/pipeline/

param sweeps are succesful @jerryyin!
Im rerunning it with normal PR CI

giuseros (approved):
LGTM!

jerryyin:
@manupak Good job, thank you!

krzysz00:
I confirm this is a correct bug fix

PR #1200, opened 2023-08-17
Make the "exhaustive" tuning space maximally exhaustive.

krzysz00:
This commits adds known-inapplicable configs back to the exhaustive tuning space in the interests of potentially finding a performance regression.

### DIFF
--- a/mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
+++ b/mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
@@ -84,9 +84,9 @@ void createGemmTuningRangeBF(TuningParamSet *newSpace,
                   InitParamsAccel gemmParams(
                       gemmMPerBlock, gemmNPerBlock, gemmKPerBlock, gemmMPerWave,
                       gemmNPerWave, gemmKPack, forceUnroll, true);
-                  if (succeeded(
-                          tuningInfo.paramsProbablyValid(info, gemmParams)) &&
-                      (kind == TuningParamSetKind::Exhaustive ||
+                  if (kind == TuningParamSetKind::Exhaustive ||
+                      (succeeded(
+                           tuningInfo.paramsProbablyValid(info, gemmParams)) &&
                        succeeded(
                            tuningInfo.couldBePerformant(info, gemmParams))))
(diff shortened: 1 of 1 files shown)


pfultz2 (approved):

manupak:
Im not fully aware of the context of the PR, if I had to guess, I'd think its coming from a perf regression on migraphx side ? 

One other explanation could be newly added elementwise fusions which could potentially make 'slow' gemms/convs to be offloaded to rocMLIR, especially if perf wasn;t measured before/after adding them. 

This was one of the reasons, I didn't enable neg, exp, div, and erf fusions. All of them reduces the perf because they offload 'slow' gemms in transformer models and later these cases were confirmed to be slow by @giuseros when we change how we invoke rocBLAS to measure perf (a.k.a. migraphx way). Im not sure whether those have improved yet... maybe they have ? 

Now I can see all that is enabled : https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/7ed4954a7114b2273a2759914adaec81a564056d/src/targets/gpu/fuse_mlir.cpp#L213-L243.

Did we do a model-wide perf after enabling these ?

As I said, I am not aware of the context, so above might not be relevant at all.

krzysz00:
@manupak That's actually pretty useful context. I enabled them on the theory that fusions are good.

What we may instead want to do is to get a better "should this go to MLIR" heuristic, analogous to how MIGraphX doesn't send split-K gemms to CK but to rocBLAS instead.

manupak:
that'd be ideal if we can define a heuristic... 
I honestly dont know a "rocMLIR is going to be bad on this gemm/conv" heuristic unless we come up with a empirical one looking at where we stand against rocBLAS/CK.

@giuseros any high-level logic to detect currently bad performing cases ?

pfultz2:
So end-to-end there is very little perf difference between Full and Exhaustive. However, there are 2 configs that are about 10% faster, but it doesn't seem like it makes a difference end-to-end. Although, there are many more configs that do run with Exhaustive as the slowest time as 10x more than the slowest with Full. Perhaps we are removing too many configs as there was some that was faster, but I dont know. 

Overall the perf seems to be something else than just pruning too many configs though. Either way, this PR would be good to have so we can diagnose perf issues in the future.

PR #1201, opened 2023-08-17
Fix HIP warnings in static library builds.

krzysz00:
The warnings we usually see apparently sometimes turned into errors in some environments.


sjw36 (approved):
LGTM

yiqian1 (approved):

PR #1204, opened 2023-08-21
Add test for SetDefaultDeviceOp rewrite

pcf000:
The upstream reviewer asked for a test.  I verified that the test fails without the already-merged patch.


krzysz00 (approved):

krzysz00:
On the other hand (I know I just approved this) ... can you look for a file of existing tests for `-gpu-to-llvm` and add this there?

pcf000:
The GPUCommon directory is the closest thing to such a file.  There are 13 uses of --gpu-to-llvm in external/llvm-project/mlir/test;  2 are in SparseTensor tests and the rest are in GPUCommon.

PR #1205, opened 2023-08-21
[DO NOT SQUASH] Fix MIGraphX quantization / dequantization 

krzysz00:
1. Remove the fp32 constraint on scales
2. Remove the constraint that the bias and output have the same type,
as this is not necessarily the case (for example, from what I can
tell, some frameworks allow i32 bias in a quantization to i8).
2b. When the bias type is not the output type in a quantization,
manually do a clamping truncation.
Note that manual clamps aren't needed in the usual case because
tosa.cast performs a clamping float-to-int.
3. Remove handling for input/output type != scale type, since that
can't happen in MIGraphX IR
4. Add constraints to encode the MIGraphX IR's assumptions about
quantize ops.

### DIFF
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -552,28 +552,52 @@ public:
                   ConversionPatternRewriter &rewriter) const final {
     Value input = op.getInput();
     Value scale = op.getScale();
+    Value bias = op.getBias();
     Value output = op.getOutput();
     Location loc = op->getLoc();
 
     Type elementType = getShapedElementTy(input);
     Value inverseScale =
         createOpAndInfer<tosa::ReciprocalOp>(rewriter, loc, elementType, scale);
-    if (getShapedElementTy(inverseScale) != getShapedElementTy(input)) {
-      inverseScale = createCastOp(rewriter, loc, elementType, inverseScale);
-    }
     Value scaled = createOpAndInfer<tosa::MulOp>(
         rewriter, loc, elementType, input, inverseScale, /*shift=*/0);
 
-    Value shifted = scaled;
-    if (auto bias = op.getBias()) {
-      Value biasCast = createCastOp(rewriter, loc, elementType, bias);
-      shifted = createOpAndInfer<tosa::AddOp>(rewriter, loc, elementType,
-                                              scaled, biasCast);
-    }
-
     Type outputType = getShapedElementTy(output);
-    Value downCast = createCastOp(rewriter, loc, outputType, shifted);
-    rewriter.replaceOp(op, downCast);
+    // If there is a bias, we upcast to the larger of the bias type and int32_t
+    // (which is what the bias type is in dequantize, the MLIR quantization
+    // implementation, and other ML frameworks) and then do a clamping
+    // truncation to the output type so that adding a bias saturates
+    // instead of overflowing.
+    Type biasType = outputType;
+    if (bias) {
+      biasType = getShapedElementTy(bias);
+      if (biasType.getIntOrFloatBitWidth() < 32) {
+        biasType = rewriter.getI32Type();
+        bias = createCastOp(rewriter, loc, biasType, bias);
+      }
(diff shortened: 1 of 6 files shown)


mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -558,22 +558,36 @@ class QuantizeLinearConverter final
...
-      Value biasCast = createCastOp(rewriter, loc, elementType, bias);
-      shifted = createOpAndInfer<tosa::AddOp>(rewriter, loc, elementType,
-                                              scaled, biasCast);
-    }
-
     Type outputType = getShapedElementTy(output);
-    Value downCast = createCastOp(rewriter, loc, outputType, shifted);
-    rewriter.replaceOp(op, downCast);
+    // We cast the scaled value to the type of the bias, which might be larger
+    // than the final output type.
```
jerryyin: What if this assumption doesn't satisfy and bias type is smaller than the final output type? If that's the case, we want to cast to scaled type and not clamp it.

BTW, is the motivation of this has to do with any existing migraphx IR/workload that has mixed type in bias?

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -558,22 +558,36 @@ class QuantizeLinearConverter final
     Type elementType = getShapedElementTy(input);
     Value inverseScale =
         createOpAndInfer<tosa::ReciprocalOp>(rewriter, loc, elementType, scale);
-    if (getShapedElementTy(inverseScale) != getShapedElementTy(input)) {
-      inverseScale = createCastOp(rewriter, loc, elementType, inverseScale);
-    }
```
jerryyin, line 563: > Remove handling for input/output type != scale type, since that can't happen in MIGraphX IR

I could be wrong, but I remember (either by looking at migraphx code base, or referenced by @pfultz2) that it is entirely possible to have fp32 to be the type of scale, which becomes different than input in quantizelinear or output in dequantizelinear.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
krzysz00: I don't know if there's an existing MIGraphX workload, but the MLIR quantization documentation specifically mentions int32 bias and int{8,16} outputs.

So we might as well support that.

But a bias smaller than the quantization type seems extremely unlikely

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
krzysz00: @pfultz2 specifically told me that the input and the scale have to have the same type in MIGraphX IR yesterday

pfultz2: Yes the input and scale should be the same type.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
pfultz2: In MIGraphX, the bias and quantized type must be the same.

krzysz00: However, does this mean that adding the bias in is meant to wrap around?

pfultz2: It should clamp.

jerryyin (approved):

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
krzysz00: @pfultz2 But that means that I have to compute `scaled + bias` in a larger type, or else, with `scaled = 126` and `bias = 2`, we'd get `scaled + bias = -128`

PR #1206, opened 2023-08-22
Fix installation of librockCompiler on Windows

apwojcik:
The default ROCm create package ruleset does not install all the required rocMLIR libraries on Windows. To fix that, we use `rocm_install(FILES ${__rocmlir_libraries} ...)` to install the complete list of libraries required.

### DIFF
--- a/mlir/tools/rocmlir-lib/CMakeLists.txt
+++ b/mlir/tools/rocmlir-lib/CMakeLists.txt
@@ -69,6 +69,7 @@ if(BUILD_FAT_LIBROCKCOMPILER)
     endforeach()
 
     foreach(__library ${__rocmlir_libs})
+      list(APPEND __rocmlir_libraries ${ROCMLIR_LIB_DIR}/${__library}.lib)
       target_link_libraries(${LIBRARY_NAME}
           INTERFACE
               $<BUILD_INTERFACE:${__library}>
@@ -78,9 +79,10 @@ if(BUILD_FAT_LIBROCKCOMPILER)
     rocm_install(FILES ${__mlir_libraries}
             DESTINATION lib/llvm)
 
-    rocm_install(FILES ${ROCMLIR_LIB_DIR}/MLIRRockThin.lib
+    rocm_install(FILES ${__rocmlir_libraries}
             DESTINATION lib)
 
+    unset(__rocmlir_libraries)
     unset(__mlir_libraries)
 
     # Linux compatibility custom target


krzysz00 (commented):
Do you need to install `__mlir_libraries` as well?

apwojcik:
Yes, both `__mlir_libraries` and `__rocmlir_libraries` are required.

mlir/tools/rocmlir-lib/CMakeLists.txt
```diff
@@ -78,9 +79,10 @@ if(BUILD_FAT_LIBROCKCOMPILER)
     rocm_install(FILES ${__mlir_libraries}
             DESTINATION lib/llvm)
 
-    rocm_install(FILES ${ROCMLIR_LIB_DIR}/MLIRRockThin.lib
+    rocm_install(FILES ${__rocmlir_libraries}
```
krzysz00, line 82: Then shouldn't this also include `${__mlir_libraries}`

apwojcik: The line above does `${__mlir_libraries}` separately and install the libraries into the `lib\llvm` directory. The rocMLIR libraries are installed into the `lib` directory. That is the safety precaution for rocMLIR to interface with the correct libraries (other components we depend on are using llvm-project in different versions and scopes).
Of course, there is no issue when someone installs into `<base>/rocmlir/`. However, we cannot guarantee following that rule. Instead, one may install it in a shared location overriding/skipping existing files and creating hard-to-find bugs.

krzysz00: Makes sense

krzysz00:
I have one question: does that mean that, on Linux, we'll install both the fat library and a copy of each static library?

I think that's bad and could lead to double-linking or blow up everyone's disk for no reason.

(unless this is an `if (WIN32)` that I missed)

apwojcik:
Only the fat library and `__rocmlir_libs` are installed on Linux. No llvm libraries are installed. That is the original behavior of the installation command for the `librockCompiler`/`all` target before we added the Windows part.

krzysz00:
Then that might've been a bug in the installation, unless I've massively missed how the build works.

I'm pretty sure we only need the fat library, since it pulls in all those rocmlir libraries (by concatenation). It's possible we're including those other libraries to get their header files, but that doesn't seem right.

apwojcik:
Let's fix it in a separate PR. Currently, this bug is blocking us from building MIOpen on Windows.

krzysz00 (approved):

PR #1207, opened 2023-08-22
Add rock.attention's gridwise lowering

manupak:
This commit mainly adds support for
lowering rock.attention to rock.gridwise_attention op

Additionaly it completes rock.attention op
spec by adding transpose flags and other required
things from rock.gridwise_attention op.

Also, this commit adds support to generate
attention kernels without scale in the interest
of verification of initial attention kernel implementations.

closes : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/955


mlir/include/mlir/Dialect/Rock/Passes.td
```diff
@@ -43,6 +43,11 @@ def RockGemmToGridwisePass : Pass<"rock-gemm-to-gridwise", "::mlir::func::FuncOp
   let dependentDialects = ["rock::RockDialect", "func::FuncDialect"];
 }
 
+def RockAttentionToGridwisePass : Pass<"rock-attention-to-gridwise", "::mlir::func::FuncOp"> {
```
krzysz00, line 46: Would it be more reasonable to stick this in `gemm-to-gridwise` since it operates at the same "level" of lowering? The same way that utility kernels get handled in `conv-to-gemm`?

mlir/lib/Dialect/Rock/Transforms/AttentionToGridwise.cpp
```diff
@@ -0,0 +1,179 @@
...
+};
+
+struct AttentionRewritePattern : public OpConversionPattern<AttentionOp> {
+  using OpConversionPattern<AttentionOp>::OpConversionPattern;
+  LogicalResult matchAndRewrite(AttentionOp op, AttentionOpAdaptor adaptor,
+                                ConversionPatternRewriter &b) const override;
+};
+} // end namespace
+
+static Attribute getTuningParams(ConversionPatternRewriter &rw,
```
krzysz00, line 64: This goes in `affix-tuning-params`, I claim

mlir/lib/Dialect/Rock/Transforms/AttentionToGridwise.cpp
```diff
@@ -0,0 +1,179 @@
...
+                                         AttentionOpAdaptor adaptor,
+                                         ConversionPatternRewriter &rw) const {
+  Location loc = op->getLoc();
+
+  if (!adaptor.getQueries().getType().isa<MemRefType>())
+    return op.emitOpError("Cannot lower unbufferized gemm to gridwise");
+
+  bool isAccel = rock::isAccel(op.getFeatures());
+  if (!isAccel) {
+    return op.emitError("Currently, attention op is only supported on GPUs "
```
krzysz00, line 98: We'll need to reflect this constraint in MIGraphX, I think

mlir/lib/Dialect/Rock/Transforms/AttentionToGridwise.cpp
```diff
@@ -0,0 +1,179 @@
...
+  auto accelParams = params.cast<RockAccelTuningParamAttrInterface>();
+  int64_t waveSize = rock::lookupArchInfo(op.getArchAttr()).waveSize;
+  int64_t blockSize = waveSize * accelParams.getNPerBlock() *
+                      accelParams.getMPerBlock() /
+                      (accelParams.getMPerWave() * accelParams.getNPerWave());
+  IntegerAttr blockSizeAttr = rw.getI32IntegerAttr(blockSize);
+  int64_t gridSize = (gemm0Size.m / accelParams.getMPerBlock()) * gemm0Size.g;
+  IntegerAttr gridSizeAttr = rw.getI32IntegerAttr(gridSize);
+
+  func::FuncOp func = op->getParentOfType<func::FuncOp>();
```
krzysz00, line 152: Yeah, this bit definitely goes in with the other tuning parameter affixation logic

krzysz00 (commented):
As a general note, given that generalized problem keys are on the 6.1 todo list, it might make sense to not bother with the transpose flags on `attention`

mlir/include/mlir/Dialect/Rock/Passes.td
manupak: If we want to do that, I'd like to rename the pass to something like : 'lower-to-gridwise' ?

manupak: For future (6.1 maybe), 

I think it;d be cleaner to have 4 dialects : 
1) RockHighLevel / RockClient / Rock -- similiar to CHLO in XLA.
2) RockGridwise
3) RockBlockwise
4) RockThreadwise

and move these transforms to be conversions really.

mlir/lib/Dialect/Rock/Transforms/AttentionToGridwise.cpp
manupak: yea -- as much as I dont like that pass , you are right in terms of consistency -- hence I ll do the movement.

Something for the future,
It;d probably better to ops 'own' the tuning params rather than a pass doing it for them. IIRC, you had some thoughts about removing that pass altogether.. anyway I ll park this conversation for later. (now is not the time)

krzysz00: Yeah, I'm with you on having ops own the tuning parameters and having a `addHeuristicTuningParams(op)` and `computeDerivedTuningParemeters(op)` function ... probably with making the op require the tuning parameter attributes.

But that's a future thing.

mlir/include/mlir/Dialect/Rock/Passes.td
krzysz00: Yeah, lower-to-gridwise is a good pass.

I'm not sure about the dialect split: `vector` has a lot of progressive lowering within itself, too.

manupak:
Superseded by https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1225/

PR #1209, opened 2023-08-23
Reduce LDS bank conflicts

giuseros:
Given the `Kouter x D x Kpack` layout of LDS in `rocMLIR`, this PR is implementing an LDS-write bank conflicts reduction policy in  two cases: 
- When the global matrix is `DxK`, this means we would write `Kpack` vectors in the same column (because we are basically transposing the global matrix in LDS). In this case we rotate the colum dimension by the row dimension : i.e., when we want to write at position `(i,j)` in LDS, we get rerouted to position `(i, (i+j)%D)`. 
- When the global matrix is `KxD`, this means that we are writing along the `D` dimension, and that would be fine if every work-item is only writing one single vector. If a workitem is writing >2 vectors, then they are writing in a strided fashion which generates bank conflicts. The solution proposed is to do a coordinate-swap when storing in LDS, so that the contiguous dimension is the workitem id. The output will be wapped back before producing the final result (thanks @manupak for the suggestion!)

These two policies required a bit of code reshuffling:
- I have to propagate the `isKContiguousDim{A/B}` down to the `blockwiseGemm` operator
- I need `copyMPerThread` and `copyMPerThread` in `blockwiseGemm` to address cases where `kpack==1`, this means that `computeCopyPerThread` has been promoted into `loweringUtils.h` because it'll be used by two operations. 
- `wrapLDSBufferForStore` is building the transformation in a top down fashion (instead than bottom up) because I need a transformation like `(d0, d1)->(d0, d1+d0)` which cannot be build bottom up. 

There are at least three big refactors that stem from this PR:
- Moving the `BlockwiseGemmAccel` operation to use transforms (as opposed to plain `AffineLoop`s)
- Wrap the LDS as a view and pass it down, so that we don't need to propagate (or recompute) LDS layout information in `BlockwiseGemmAccel`/`BlockwiseGemm`
- Unify accel and non-accel code paths

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -1046,7 +1046,9 @@ defvar LdsBufferTypes = GemmInputTypes # [VectorOfRankAndType<[1], GemmInputType
 def Rock_BlockwiseGemmOp:
     Rock_Op<"blockwise_gemm">,
     Arguments<(ins MemRefRankOf<GemmInputTypes, [3]>:$matrixA,
+                   BoolAttr : $isKContiguousDimA,
                    MemRefRankOf<GemmInputTypes, [3]>:$matrixB,
+                   BoolAttr : $isKContiguousDimB,
                    MemRefRankOf<GemmOutputTypes, [2]>:$matrixC,
                    Rock_GeneralGemmParamsAttr:$params
                    )> {
@@ -1084,7 +1086,9 @@ defvar AccelResTypes = [VectorOfLengthAndType<[4, 8, 16, 32], [F32, I32, F16, BF
 def Rock_BlockwiseGemmAccelOp:
     Rock_Op<"blockwise_gemm_accel">,
     Arguments<(ins MemRefOf<LdsBufferTypes>:$matrixA,
+                   BoolAttr:$isKContiguousDimA,
                    MemRefOf<LdsBufferTypes>:$matrixB,
+                   BoolAttr:$isKContiguousDimB,
                    Index:$waveOffsetA,
                    Index:$waveOffsetB,
                    MemRefOf<AccelArgTypes>:$bufferA,
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -1067,6 +1067,11 @@ def Rock_BlockwiseGemmOp:
     Each {m,n}PerThread group of elements (which may themselves be kPacks)
     is read from LDS by a given thread, then the next mPerThread group is
     mRepeatStride elements later in the buffer.
+
+    `rotate{M,N}PerThread` is used to rotate the access to LDS to avoid bank-conflicts in cases
+    where subsequent threads try to write data in the same column of the output matrix. If `kpack>1`
+    the rotation is simply the row index. If `kpack==1` the rotation takes into account the number of
+    elements reads by a thread along the `M` and/or `N` dimension, namely `in{M,N}PerThread`
   }];
   let assemblyFormat = [{
     $matrixC `+` `` `=` $matrixA `*` $matrixB attr-dict
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -1046,12 +1046,12 @@ defvar LdsBufferTypes = GemmInputTypes # [VectorOfRankAndType<[1], GemmInputType
 def Rock_BlockwiseGemmOp:
     Rock_Op<"blockwise_gemm">,
     Arguments<(ins MemRefRankOf<GemmInputTypes, [3]>:$matrixA,
-                   BoolAttr : $isKContiguousDimA,
                    MemRefRankOf<GemmInputTypes, [3]>:$matrixB,
-                   BoolAttr : $isKContiguousDimB,
                    MemRefRankOf<GemmOutputTypes, [2]>:$matrixC,
-                   I32Attr:$copyMPerThread,
-                   I32Attr:$copyNPerThread,
+                   I32Attr:$inMPerThread,
+                   I32Attr:$inNPerThread,
+                   UnitAttr:$rotateMWithK,
+                   UnitAttr:$rotateNWithK,
                    Rock_GeneralGemmParamsAttr:$params
                    )> {
   let summary = "Blockwise GEMM non accelerated version";
@@ -1088,11 +1088,11 @@ defvar AccelResTypes = [VectorOfLengthAndType<[4, 8, 16, 32], [F32, I32, F16, BF
 def Rock_BlockwiseGemmAccelOp:
     Rock_Op<"blockwise_gemm_accel">,
     Arguments<(ins MemRefOf<LdsBufferTypes>:$matrixA,
-                   BoolAttr:$isKContiguousDimA,
                    MemRefOf<LdsBufferTypes>:$matrixB,
-                   BoolAttr:$isKContiguousDimB,
-                   I32Attr:$copyMPerThread,
-                   I32Attr:$copyNPerThread,
+                   I32Attr:$inMPerThread,
+                   I32Attr:$inNPerThread,
+                   UnitAttr:$rotateMWithK,
+                   UnitAttr:$rotateNWithK,
                    Index:$waveOffsetA,
                    Index:$waveOffsetB,
                    MemRefOf<AccelArgTypes>:$bufferA,
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -1050,6 +1050,8 @@ def Rock_BlockwiseGemmOp:
                    MemRefRankOf<GemmInputTypes, [3]>:$matrixB,
                    BoolAttr : $isKContiguousDimB,
                    MemRefRankOf<GemmOutputTypes, [2]>:$matrixC,
+                   I32Attr:$copyMPerThread,
+                   I32Attr:$copyNPerThread,
                    Rock_GeneralGemmParamsAttr:$params
                    )> {
   let summary = "Blockwise GEMM non accelerated version";
@@ -1089,6 +1091,8 @@ def Rock_BlockwiseGemmAccelOp:
                    BoolAttr:$isKContiguousDimA,
                    MemRefOf<LdsBufferTypes>:$matrixB,
                    BoolAttr:$isKContiguousDimB,
+                   I32Attr:$copyMPerThread,
+                   I32Attr:$copyNPerThread,
                    Index:$waveOffsetA,
                    Index:$waveOffsetB,
                    MemRefOf<AccelArgTypes>:$bufferA,
--- a/mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
+++ b/mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
@@ -10,6 +10,7 @@
 #define ROCK_UTILITY_LOWERINGUTILS_H
 
 #include "mlir/Dialect/Rock/IR/RockTypes.h"
+#include "mlir/Dialect/Rock/IR/TransformMapBuilder.h"
 #include "mlir/Support/LLVM.h"
 #include "llvm/ADT/SmallVector.h"
 
@@ -107,6 +108,22 @@ backwardDataKernelIds(int64_t strideHeight, int64_t strideWidth,
 /// Return a vector type of length `len` if `len` is more than 1, otherwise,
 /// return `type`.
 Type vectorTypeOrSelf(Type elementType, int64_t len);
+
+/// Compute a thread copy layout, i.e., how many elements a single thread (or
+/// workitem) reads along K and M (independently on how we vectorize the reads)
+FailureOr<std::pair<int64_t, int64_t>>
+computeCopyPerThread(Type elementType, int64_t copyPerThread, int64_t kPerBlock,
+                     int64_t dPerBlock, int64_t kpack, Location loc);
+
+// if K is not the contiguous dimension, we swapped (on each axis) the thread id
+// and the iter id dimensions, so that the threads write in a contiguous fashion
+// minimizing LDS bank conflicts.  This transformation swap those dimensions
+// back before producing the final output view
+TopDownTMBuilder swapThreadIdAndIteration(
+    TopDownTMBuilder &toMatrixC, ArrayRef<int64_t> bidGridLengths,
+    int64_t copyMPerThread, int64_t copyNPerThread, int64_t mPerBlock,
+    int64_t nPerBlock, bool isKContiguousDimA, bool isKContiguousDimB,
+    SmallVector<Attribute> &transformAttrs);
 
 } // end namespace rock
 } // end namespace mlir
--- a/mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
+++ b/mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
@@ -60,7 +60,8 @@ FailureOr<RegsAsMatrixSubTiles> getPackedRegsAsTileViews(
     OpBuilder &b, Location loc, Value globalBuffer, StringRef dName,
     ArrayRef<StringRef> bidGridOrder, ArrayRef<int64_t> bidGridLengths,
     int64_t blockSize, int64_t kPerBlock, int64_t dPerBlock, int64_t kPerThread,
-    int64_t dPerThread, int64_t kpack, bool isKContigousDim);
+    int64_t dPerThread, int64_t kpack, bool isKContigousDim,
+    bool doSwapThreadIterSubDimsForD);
 
 bool isWrWAtomicKernel(GemmFeatures features, Type dataType,
                        bool requiredPadding);
--- a/mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
+++ b/mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
@@ -113,12 +113,13 @@ Type vectorTypeOrSelf(Type elementType, int64_t len);
 // and the iter id dimensions, so that the threads write in a contiguous fashion
 // minimizing LDS bank conflicts.  This transformation swap those dimensions
 // back before producing the final output view
-TopDownTMBuilder swapThreadIdAndIteration(
-    TopDownTMBuilder &toMatrixC, ArrayRef<int64_t> bidGridLengths,
-    int64_t copyMPerThread, int64_t copyNPerThread, int64_t mPerBlock,
-    int64_t nPerBlock, bool doSwapThreadIterSubDimsForM,
-    bool doSwapThreadIterSubDimsForN, bool isBlockwise,
-    SmallVector<Attribute> &transformAttrs);
+TopDownTMBuilder
+swapThreadIdAndIteration(TopDownTMBuilder &toMatrixC, int64_t mBlocks,
+                         int64_t nBlocks, int64_t copyMPerThread,
+                         int64_t copyNPerThread, int64_t mPerBlock,
+                         int64_t nPerBlock, bool doSwapThreadIterSubDimsForM,
+                         bool doSwapThreadIterSubDimsForN, bool isBlockwise,
+                         SmallVector<Attribute> &transformAttrs);
 
 } // end namespace rock
 } // end namespace mlir
--- a/mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
+++ b/mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
@@ -109,12 +109,6 @@ backwardDataKernelIds(int64_t strideHeight, int64_t strideWidth,
 /// return `type`.
 Type vectorTypeOrSelf(Type elementType, int64_t len);
 
-/// Compute a thread copy layout, i.e., how many elements a single thread (or
-/// workitem) reads along K and M (independently on how we vectorize the reads)
-FailureOr<std::pair<int64_t, int64_t>>
-computeCopyPerThread(Type elementType, int64_t copyPerThread, int64_t kPerBlock,
-                     int64_t dPerBlock, int64_t kpack, Location loc);
-
 // if K is not the contiguous dimension, we swapped (on each axis) the thread id
 // and the iter id dimensions, so that the threads write in a contiguous fashion
 // minimizing LDS bank conflicts.  This transformation swap those dimensions
@@ -122,7 +116,8 @@ computeCopyPerThread(Type elementType, int64_t copyPerThread, int64_t kPerBlock,
 TopDownTMBuilder swapThreadIdAndIteration(
     TopDownTMBuilder &toMatrixC, ArrayRef<int64_t> bidGridLengths,
     int64_t copyMPerThread, int64_t copyNPerThread, int64_t mPerBlock,
-    int64_t nPerBlock, bool isKContiguousDimA, bool isKContiguousDimB,
+    int64_t nPerBlock, bool doSwapThreadIterSubDimsForM,
+    bool doSwapThreadIterSubDimsForN, bool isBlockwise,
     SmallVector<Attribute> &transformAttrs);
 
 } // end namespace rock
--- a/mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
+++ b/mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
@@ -9,6 +9,7 @@
 #define ROCK_UTILITY_TRANSFORMMAPUTILS_H
 
 #include "mlir/Dialect/Rock/IR/Rock.h"
+#include "mlir/Dialect/Rock/IR/TransformMapBuilder.h"
 #include "mlir/Dialect/Utils/ReshapeOpsUtils.h"
 
 namespace mlir {
@@ -135,6 +136,16 @@ TransformMapAttr transformExtractSlice(OpBuilder &b, Location loc,
                                        ArrayRef<int64_t> outShape,
                                        ArrayRef<int64_t> offsets,
                                        ArrayRef<int64_t> sizes);
+
+// If the condition is satified, rotate the dimension `d` by `k` using
+// `d = (d+k*stride) % len(d)`
+rock::TopDownTMBuilder rotateIf(bool condition, TopDownTMBuilder &builder,
+                                TransformMapAttr &attr, int64_t stride,
+                                StringRef dName, int64_t d, int64_t dPos,
+                                StringRef kName, int64_t k,
+                                ArrayRef<StringRef> beforeDims,
+                                ArrayRef<StringRef> afterDims,
+                                SmallVector<Attribute, 4> &transformAttrs);
 
 // This utility function will take an ordered decreasing dimension strides and
 // total number of elements to produce an array of dimension sizes. This
--- a/mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
+++ b/mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
@@ -139,13 +139,11 @@ TransformMapAttr transformExtractSlice(OpBuilder &b, Location loc,
 
 // If the condition is satified, rotate the dimension `d` by `k` using
 // `d = (d+k*stride) % len(d)`
-rock::TopDownTMBuilder rotateIf(bool condition, TopDownTMBuilder &builder,
-                                TransformMapAttr &attr, int64_t stride,
-                                StringRef dName, int64_t d, int64_t dPos,
-                                StringRef kName, int64_t k,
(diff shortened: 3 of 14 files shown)


giuseros:
Fixes https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/806

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1155,9 +1144,11 @@ struct GridwiseGemmAccelRewritePattern
 
       // Emit blockwise GEMM.
       blockwiseGemmAccelOp = b.create<BlockwiseGemmAccelOp>(
-          loc, ldsViewForGemmA, ldsViewForGemmB, mMyWaveOffsetA, mMyWaveOffsetB,
-          arrayA, arrayB, regCAllocOp, op.getArchAttr(), op.getFeaturesAttr(),
-          op.getBlockSizeAttr(), op.getParamsAttr());
+          loc, ldsViewForGemmA,
+          BoolAttr::get(op->getContext(), isKContiguousDimA), ldsViewForGemmB,
```
krzysz00: I think there's a builder method for this?

krzysz00 (commented):
Very minor comments currently.

I don't see any real issues with the approach, if it works

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -193,7 +194,9 @@ makeViewsForRowsAndCols(TopDownTMBuilder &viewBuilder, int64_t mPerRepeat,
 
 RegsAsMatrixSubTiles MfmaEmitter::computeOutputTransforms(
     PatternRewriter &b, Location loc, int64_t mLen, int64_t nLen,
-    int64_t blockSize, ArrayRef<int64_t> bidGridLengths) {
+    bool isKContiguousDimA, bool isKContiguousDimB, int64_t computeMPerThread,
```
manupak: nit-naming : While I agree we do the swap based on K being contigous, might for these utils I;d prefer a more direct command such as 'doSwapThreadIterSubDimsForM' / 'doSwapThreadIterSubDimsForN'.

Also lets not use 'A' and 'B' to make these utils gemm agonstic but more related to an individual matrix, because A & B is only relevant here because we take M from A and N from B. I feel using N & M is more consistent with rest of the code here w/o introducing the concepts of A & B.

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -305,9 +308,17 @@ RegsAsMatrixSubTiles MfmaEmitter::computeOutputTransforms(
...
     toMatrixC.unmerge("gemmN", 2, dimNamesN, dimSizesN);
-    TransformMapAttr toMatrixCAttr = toMatrixC.get();
-    ret.gridSubTile = b.getArrayAttr(
-        {splitMemoryCoordsAttr, toRowsAndColsAttr, toMatrixCAttr});
+
+    // Before returning the output view, if necessary, swap back the
+    // threadid/iter dimensions on both the M/N axis.
+    SmallVector<Attribute> transformAttrs{splitMemoryCoordsAttr,
+                                          toRowsAndColsAttr};
+    mlir::rock::swapThreadIdAndIteration(
```
manupak, line 316: This swap needs to happen for blockwise subTile view as well.

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -514,7 +525,9 @@ void WmmaEmitter::emitThreadwiseLoop(OpBuilder &b, Location loc, Value argA,
 
 RegsAsMatrixSubTiles WmmaEmitter::computeOutputTransforms(
     PatternRewriter &b, Location loc, int64_t mLen, int64_t nLen,
-    int64_t blockSize, ArrayRef<int64_t> bidGridLengths) {
+    bool isKContiguousDimA, bool isKContiguousDimB, int64_t computeMPerThread,
+    int64_t computeNPerThread, int64_t blockSize,
+    ArrayRef<int64_t> bidGridLengths) {
```
manupak, line 539: nothing meaninful have been done with these changes ?
dont we need to do the swap here as well ?

mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
```diff
@@ -108,6 +109,22 @@ backwardDataKernelIds(int64_t strideHeight, int64_t strideWidth,
...
+/// workitem) reads along K and M (independently on how we vectorize the reads)
+FailureOr<std::pair<int64_t, int64_t>>
+computeCopyPerThread(Type elementType, int64_t copyPerThread, int64_t kPerBlock,
+                     int64_t dPerBlock, int64_t kpack, Location loc);
+
+// if K is not the contiguous dimension, we swapped (on each axis) the thread id
+// and the iter id dimensions, so that the threads write in a contiguous fashion
+// minimizing LDS bank conflicts.  This transformation swap those dimensions
+// back before producing the final output view
+TopDownTMBuilder swapThreadIdAndIteration(
```
manupak: See the comment below on MfmaEmitter::computeOutputTransforms.
I'd like this utility tied to swapping sub dims of 'M' (i.e. rows) and 'N' (i.e. cols) where you should only need
mLen, mPerBlock, mPerThread (same for N). This could only use that information with a more direct naming for the bools doSwapThreadIterSubDimsofM (or similiar).

Lets not introduce concepts of A. B and K here.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -185,7 +150,9 @@ computeCopyPerThread(Type elementType, int64_t copyPerThread, int64_t kPerBlock,
 static FailureOr<Value> wrapLDSBufferForStore(OpBuilder &b, Location loc,
                                               Value buffer, Type ldsReadType,
                                               int64_t kOuter, StringRef dName,
-                                              int64_t d, int64_t kPerThread) {
+                                              int64_t d, int64_t kPerThread,
+                                              int64_t dPerThread,
+                                              bool isKContiguousDim) {
```
manupak: What if K & D is neither the contiguous dimension ? 
Now I think about it again, I feel we should only do this if global kVecLen > 1.

Nit-naming - id would be better to give a direct name such as 'rotateKwithD' or similar instead of the proxy 'isKContiguousDim' and let the caller decide when to do the rotation.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -185,7 +150,9 @@ computeCopyPerThread(Type elementType, int64_t copyPerThread, int64_t kPerBlock,
 static FailureOr<Value> wrapLDSBufferForStore(OpBuilder &b, Location loc,
                                               Value buffer, Type ldsReadType,
                                               int64_t kOuter, StringRef dName,
-                                              int64_t d, int64_t kPerThread) {
+                                              int64_t d, int64_t kPerThread,
```
manupak, line 191: Also please update the function description as it is misleading now to be said LDS layout is <kOuter * d * kpack * sizeof(T) x i8

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -254,6 +276,13 @@ struct BlockwiseGemmRewritePattern
     toLdsIndexA.ignore("n_cuwave");
     toLdsIndexA.passThrough({"kpack"}, {2}, {"kpack"});
     TransformMapAttr toLdsIndexAAttr = toLdsIndexA.get();
+    SmallVector<Attribute, 4> transformAttrsA{splitTidAAttr, toLdsIndexAAttr};
+
+    // Since the LDS layout is `kOuter x rotate(m) x kpack` we want to rotate
```
manupak: This needs to be said in the function docs for wrapLDSforStore as well.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -242,6 +242,28 @@ struct BlockwiseGemmRewritePattern
...
+
+    Type elementTypeA = blockAType.getElementType();
+    if (auto vectorDataType = elementTypeA.dyn_cast<VectorType>())
+      elementTypeA = vectorDataType.getElementType();
+
+    auto maybeCopyAPerThread = computeCopyPerThread(
+        elementTypeA, aCopyPerThread, k * kPack, m, kPack, loc);
+    if (failed(maybeCopyAPerThread))
+      return maybeCopyAPerThread;
+    int64_t copyMPerThread = (*maybeCopyAPerThread).second;
```
manupak: This feels fragile to me...
We shouldn;t be recomputing this as it was decided before the op is created and this coule be problematic if the op is created with a seperate logic.

I;d say we just put an attribute on the op to indicate mPerThread and nPerThread; I think that is all you want to know rather than recomputing it here. It makes the debugging easier in future also less problematic for the second gemm in the attention kernel.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -254,6 +276,13 @@ struct BlockwiseGemmRewritePattern
     toLdsIndexA.ignore("n_cuwave");
     toLdsIndexA.passThrough({"kpack"}, {2}, {"kpack"});
     TransformMapAttr toLdsIndexAAttr = toLdsIndexA.get();
+    SmallVector<Attribute, 4> transformAttrsA{splitTidAAttr, toLdsIndexAAttr};
+
+    // Since the LDS layout is `kOuter x rotate(m) x kpack` we want to rotate
+    // the dimension `m` before reading from LDS
+    int64_t strideA = (kPack == 1 ? copyMPerThread : 1);
+    rotateIf(op.getIsKContiguousDimA(), toLdsIndexA, toLdsIndexAAttr, strideA,
```
manupak: Please put a cross-reference link to wrapLDSforStore to say that these two needs to be kept in sync.

TLDR;

I spent some time thinking about can we pass the transformMaps of the LDS so these need to not be maintained here.. it feels complex so lets keep a cross-reference here.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -423,6 +457,21 @@ struct BlockwiseGemmAccelRewritePattern
     Value mWavesConstantOp = b.create<ConstantIndexOp>(loc, mWaves);
     Value nWavesConstantOp = b.create<ConstantIndexOp>(loc, nWaves);
 
+    int64_t aCopyPerThread = (K * M * KPack) / op.getBlockSize();
+    int64_t bCopyPerThread = (K * N * KPack) / op.getBlockSize();
+
+    auto maybeCopyAPerThread = computeCopyPerThread(dataTypeA, aCopyPerThread,
```
manupak: Here also, lets not recompute mPerThread and nPerThread here and pass in as attributes on the basis it was decided when the op is created and it should be upto the caller to decide that.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -455,11 +504,31 @@ struct BlockwiseGemmAccelRewritePattern
...
+                                 bool isKContiguousDim,
+                                 int64_t copyMNPerThread) {
       // Compute source offset
       Value sourceOffset = accelEmitterPtr->computeLdsSourceOffset(
           kb, k_i, mnb, mn_i, b, MN, loc, sourceBase, mnWaves, laneId);
 
+      // Since the LDS layou is `k0 x rotate(m) x kpack` we want to rotate
+      // the dimension `n` before reading from LDS. Since we don't use
+      // any transform here, this is done by hand with rem/div operators
+      if (isKContiguousDim) {
```
manupak: as before, I think we should be doing this only if kVecLen > 1.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
manupak: 'isKContiguousDim' is going to be rather confusing when I call wrapLDSBufferForStore for the seccond gemm. I think 'rotateKwithD' is better because it gives the sense it is optional/perf improvement flag that called can use as long as blockwiseGemm knows that.

 With that I feel its better to name for blockwiseGemm 's new attrbitues to indicate K + D rotation is done on the input rather than saying it was done because K was contigous dim -- which is true but that should only be present in GridwiseGemmtoBlockwise where we make the decision to rotateKwithD when the input tiles for A & B is loaded from global memory.

manupak (dismissed):
Thanks for the PR! Its great to see you have sorted things when K global vectorized scenario.
I m an agreement to binary generated from this PR (I think) and my requested changes are more structural in nature.

I did a detailed review on the PR, -- keeping an eye out specially where coupling is being introduced between gridwise and blockwise lowering is happenning. I wanted to point out places where assumptions being taken how gridwise lowering is happenning when we are doing blockwise lowering. So my main suggestions are to hoist them up to gridwise lowering so that we still allow blockwise gemm to be used in other kernels (Im obviously thinking of the attention kernel).

Also I note wmma computeOutputTransforms are not affected by non-K sub-dim swaps. Is that correct ? Or Is it because we dont have Navi3x in our CI that this is passing ?

Other than that my other comments are naming nits.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
giuseros: It is super fragile, but I feared that I was adding too many attributes on the op. But if it is fine with you, I will go ahead and do it :)

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
giuseros: The problem is also that on the accel path we don't have transform maps. I would pass the maps once we bring the accel path to use them.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
giuseros: > Now I think about it again, I feel we should only do this if global kVecLen > 1.

Can we talk more about that? If I understand what is the issue you see, I am not sure kVecLen>1 is either a good condition.

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
giuseros: Yes, I fear that this is because we don't have wmma in our CI. I will try to run the tests on my local card

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
giuseros: Yeah I "think" I see why you want that, but I am not sure this is the right solution. Resolving this discussion, but leaving the other one open.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1046,8 +1046,12 @@ defvar LdsBufferTypes = GemmInputTypes # [VectorOfRankAndType<[1], GemmInputType
 def Rock_BlockwiseGemmOp:
     Rock_Op<"blockwise_gemm">,
     Arguments<(ins MemRefRankOf<GemmInputTypes, [3]>:$matrixA,
+                   BoolAttr : $isKContiguousDimA,
                    MemRefRankOf<GemmInputTypes, [3]>:$matrixB,
+                   BoolAttr : $isKContiguousDimB,
```
manupak: as discussed, it might be better to call these rotateKwithD (to denote what exactly they are doing rather than the 'why') and also I think they could be UnitAttr (i.e. optional)

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1046,8 +1046,12 @@ defvar LdsBufferTypes = GemmInputTypes # [VectorOfRankAndType<[1], GemmInputType
...
                    MemRefRankOf<GemmOutputTypes, [2]>:$matrixC,
+                   I32Attr:$copyMPerThread,
```
manupak: nit : we could drop the copy* prefix as it doesn;t need to be copied -- it could also be computed.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1084,7 +1088,11 @@ defvar AccelResTypes = [VectorOfLengthAndType<[4, 8, 16, 32], [F32, I32, F16, BF
 def Rock_BlockwiseGemmAccelOp:
     Rock_Op<"blockwise_gemm_accel">,
     Arguments<(ins MemRefOf<LdsBufferTypes>:$matrixA,
+                   BoolAttr:$isKContiguousDimA,
                    MemRefOf<LdsBufferTypes>:$matrixB,
+                   BoolAttr:$isKContiguousDimB,
+                   I32Attr:$copyMPerThread,
+                   I32Attr:$copyNPerThread,
```
manupak: Same comment as above

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
giuseros: We clarified this point with a chat :) Marking this as resolved

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.h
```diff
@@ -114,7 +114,9 @@ struct AccelEmitter {
   /// matrix multiplication tile.
   virtual RegsAsMatrixSubTiles
   computeOutputTransforms(PatternRewriter &b, Location loc, int64_t mLen,
-                          int64_t nLen, int64_t blockSize,
+                          int64_t nLen, bool isKContiguousDimA,
+                          bool isKContiguousDimB, int64_t computeMPerThread,
```
manupak: naming : I think you missed the changes here.

manupak (commented):
I dont have major concerns here assuming Navi3x are working :).

As we discussed offline,
I'd like to gridwise common utils and blockwise gemm to have explicit flags for the rotation (where applicable) and subdim swaps (where applicable) . I can still see KContigousDimA in some places.

We can use the heuristic contiguous dim being actual driver to fix those flags when we do the gridwise to blockwise lowering.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
giuseros: The idea here is that the `dPerThread` is how many elements each thread copies from global to LDS, which can be different from how many elements each thread computes. Happy to proceed if you want to still change the name, I just wanted to give you the context of why I picked it

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
giuseros: In the operator? The point is that we will use that attribute to rotate and/or swap. If we name the operator property `rotateKwithD`, during the lowering we will have:
```
bool rotateDWithK = op.getRotateKWithD
bool doSwapThreadIterSubDimsForM = !rotateDWithK
```
If you wish, we might add a `doSwapThreadIterSubDimsForM` additional property to the blockwise gemm operator to keep things separated. What do you think?

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
giuseros: I ran the tests on my navi31 board and they all passed. Marking this as resolved.

giuseros:
@krzysz00 , @manupak I re-ran the tuner and got the performance results, which look quite good (better than I expected). I am attaching the results here:

[ldsPerfComparison.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/12439463/ldsPerfComparison.zip)

Please note that in the original tuning db, entries for 1k,2k4k,8k,  are missing (hence why the `FAILED` rows).

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: What I meant to say was the input to blockwiseGemm doesn;t need to come from global -- always.
I think you are talking assumption that input tensors live in global --> regs --> LDS.

```mermaid
flowchart TD
   In1 --> gemm0
   In2 --> gemm0
   gemm0 --> gemm1
   In3 --> gemm1
   gemm1 --> Out1
```

So for `gemm1`'s LHS input data in the thread (dPerThead) needs to be written to LDS. These doesn't match with definition of 'each thread copies from global to LDS'

manupak: That dPerThread is computed not copied.

manupak:
Perf is amazing!

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: as discussed offline, I think they are not mutually exclusive from blockwiseGemm PoV.
i.e. it will still result in a valid kernel that is computing accurate output.

They become mutually exclusive in gridwise to blockwise lowering; so lets hoist the decision to the lowering.

giuseros: That was my bad, because we don't use the swap flag in blockwise, but we only use the rotation flag. So I added the two rotateMWithK and rotateNWithK  as you suggested

giuseros:
I did an analysis of mlir vs rocblas perf, and we actually improved in many cases, so I can confirm there is a good perf improvement, although in same cases it might not be as high as the report I shared. I might still want to do another round of tuning, but perf-wise I think this feature is ready to go

giuseros:
Hi @manupak , is there anything else that needs to be addressed ?

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
```diff
@@ -344,3 +352,81 @@ FailureOr<RegsAsMatrixSubTiles> mlir::rock::getPackedRegsAsTileViews(
   }
   return gpuViews;
 }
+
+TopDownTMBuilder mlir::rock::swapThreadIdAndIteration(
+    TopDownTMBuilder &toMatrixC, ArrayRef<int64_t> bidGridLengths,
```
manupak: bidGridLengths should not be needed here. I mean how the grid is partitioned should not be relevant.
I think what you need is mblocks and nblocks.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1048,6 +1048,10 @@ def Rock_BlockwiseGemmOp:
     Arguments<(ins MemRefRankOf<GemmInputTypes, [3]>:$matrixA,
                    MemRefRankOf<GemmInputTypes, [3]>:$matrixB,
                    MemRefRankOf<GemmOutputTypes, [2]>:$matrixC,
+                   I32Attr:$inMPerThread,
+                   I32Attr:$inNPerThread,
+                   UnitAttr:$rotateMWithK,
+                   UnitAttr:$rotateNWithK,
```
manupak, line 1054: We might need to describe these flags in the description (below) for others who haven;t followed this PR.

manupak (commented):
almost there... (Im droping my hold as comments are minor)

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
manupak: to clarify further
bidGridLengths[1] == mblocks and  bidGridLengths[2] == nblocks will not always be the case..

manupak:
> I did an analysis of mlir vs rocblas perf, and we actually improved in many cases, so I can confirm there is a good perf improvement, although in same cases it might not be as high as the report I shared. I might still want to do another round of tuning, but perf-wise I think this feature is ready to go

Since this is a purely a perf related PR, we might want to attach some numbers here.
IIRC, the early data was not relevant because the tuning was not *fair* ? 

If getting tuning data is the problematic aspect,
I feel we could do a iso-tuning (using same tuning db) perf comparison -- taking the tuning db from develop (i.e. weeklies)

manupak:
Code-wise approved! Thanks for bearing with me...
(just waiting on to see the perf diff)

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -143,9 +143,9 @@ bestGlobalVectorization(OpBuilder &b, Value matrix, int64_t copyDPerThread,
 
 /// Compute a thread copy layout, i.e., how many elements a single thread (or
 /// workitem) reads along K and M (independently on how we vectorize the reads)
-static FailureOr<std::pair<int64_t, int64_t>>
-computeCopyPerThread(Type elementType, int64_t copyPerThread, int64_t kPerBlock,
-                     int64_t dPerBlock, int64_t kpack, Location loc) {
+FailureOr<std::pair<int64_t, int64_t>> static computeCopyPerThread(
```
krzysz00: Why'd you move the `static`?

mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
```diff
@@ -136,6 +137,16 @@ TransformMapAttr transformExtractSlice(OpBuilder &b, Location loc,
...
 
+// If the condition is satified, rotate the dimension `d` by `k` using
+// `d = (d+k*stride) % len(d)`
+rock::TopDownTMBuilder rotateIf(bool condition, TopDownTMBuilder &builder,
+                                TransformMapAttr &attr, int64_t stride,
+                                StringRef dName, int64_t d, int64_t dPos,
+                                StringRef kName, int64_t k,
+                                ArrayRef<StringRef> beforeDims,
+                                ArrayRef<StringRef> afterDims,
+                                SmallVector<Attribute, 4> &transformAttrs);
```
krzysz00: Nit: `SmallVectorImpl<Attribute>&` so you don't need to specify the size.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -176,16 +176,19 @@ computeCopyPerThread(Type elementType, int64_t copyPerThread, int64_t kPerBlock,
...
 }
 
-/// Wraps the LDS buffer "buffer", which is <kOuter * d * kpack * sizeof(T) x i8
-/// into a tid x iter view, where `iter` iterates over nominal scalar indices
-/// into a buffer of type T. `buffer` will be reinterpreted as a buffer with
-/// element type vector<kpackPerThread x T> (with kpackPerThread == 1 meaning
-/// just T). The resulting view must be iterated over with a stride of no less
-/// than min(kPerThread, kpack).
+/// Wraps the LDS buffer "buffer", which is <kOuter * rotate(d) * kpack *
+/// sizeof(T) x i8 into a tid x iter view, where `iter` iterates over nominal
```
krzysz00: Comment nit: I woludn't put the rotate in here, and add back the bracket after i8, to emphasize that the input is a 1D buffer.

Then I'd mention that the `d` dimension is rotated in a separate sentence

krzysz00 (approved):
Looks reasonable, I've mainly got nitpicks.

Do you know where some of those -7% drops in the perf report come from? Are they still there on a rerun (using the same perf configs)?

giuseros:
Thanks @krzysz00 ! About the perf degrations, I am not sure. I can reproduce them, and it seems only to happen in `int8` world. I can see the LDS bank conflicts going down, but maybe we were relying on the loads to take time to balance between different threadd

krzysz00:
Re the int8 degradations, if they don't show up consistently, then it's probably fine.

giuseros:
Ok, some news about this. Even with rocm-head data, the perf seemed skewed, so I bit the bullet and ran another tuning session over the week-end. The result was that we were having performance regressions in int8, ranging between -8%  up to -15%. 

I investigated a bit, and the problem is that the tricks that we are using to improve bank conflicts might cause troubles during vectorization. 

E.g., if if maxVecLen=16 and kpack=4, each thread is writing 4 contiguous vectors (each of 4 elements) in LDS. LLVM can vectorize this by a single ds_write_128. However, if we swap thread-id and iter-id dimensions we will loose this vectorization opportunity.

So I added some more conditions to decide when to rotate/swap dims. This removed the int8 regressions. 

Attaching here the new report (which can considered final)

[final.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/12512957/final.zip)

PR #1210, opened 2023-08-23
[DO NOT MERGE] Handle linalg.generic blocks that reduced to constant initilizers

krzysz00:
Fixes
https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1098

Some linalg.generics blocks can (due to, as far as I can tell, weirdness involved in how broadcasting, linalg.generic fusion. and such are handled) turn into constant initializers.

The existence of such blocks broke our assumption that all the niputs to a linalg.generic either traced to the threadwise_write_all from a gemm or to a function input. This failed to account for the cases where the input was a local allocation that was filled with constants by a linalg.generic. (It also failed to account for the possibility of the value coming from a `memref.get_global`, which might happen if we end up with large tensor constants in our models).

This commit resolves the issues as follows:
1. Adding special logic to the handling of linalg.generic operations in -rock-align tiling. When such an operation has no inputs, we replace the memref.alloc it writes to (which will be the immediate argument to that op, thanks to regularization) with a rock.alloc in registers of the same type but take no other action.
2. Improve moveTransformsBefore() to handle the said allocations. If a transform chain terminates at a private allocation, we move it before the corresponding threadwise_write_all if it isn't already there, and move the initializing linalg.generic along with it. (As noted in the comments, there's a hypothetical failure mode here involving non-constant initialization, but I'm taking that risk for now).

A minimized version of the crash reproducer has been added as a pass test.

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/Passes.td
+++ b/mlir/include/mlir/Dialect/Rock/Passes.td
@@ -55,7 +55,7 @@ def RockGridwiseGemmToBlockwisePass : Pass<"rock-gridwise-gemm-to-blockwise", ":
 
 def RockLinalgAlignPass : Pass<"rock-linalg-align", "::mlir::func::FuncOp"> {
   let summary = "expand linalg ops aligned with threadwise copy";
-  let dependentDialects = ["rock::RockDialect", "linalg::LinalgDialect", "vector::VectorDialect", "memref::MemRefDialect"];
+  let dependentDialects = ["rock::RockDialect", "linalg::LinalgDialect", "vector::VectorDialect", "memref::MemRefDialect", "gpu::GPUDialect"];
 }
 
 def RockBlockwiseGemmToThreadwisePass : Pass<"rock-blockwise-gemm-to-threadwise", "::mlir::func::FuncOp"> {
--- a/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
@@ -21,6 +21,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/GPU/IR/GPUDialect.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/Linalg/Utils/Utils.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
@@ -87,15 +88,55 @@ struct ReduceRewritePattern : public OpRewritePattern<rock::ReduceOp> {
 };
 } // end anonymous namespace
 
+// Returns, if it exists, the linalg.generic that initilizes this linalg.generic
+// with some value that does not depend on any inputs, or the null operation
+// if there is none such.
+static linalg::GenericOp getConstInitLaGeneric(rock::GpuAllocOp alloc) {
+  linalg::GenericOp ret;
+  for (Operation *user : alloc->getUsers()) {
+    if (auto generic = dyn_cast<linalg::GenericOp>(user)) {
+      if (!generic.getInputs().empty())
+        continue;
+      if (generic.getOutputs().size() != 1)
+        continue;
+      if (generic.getOutputs().front() == alloc) {
+        if (ret) {
+          LLVM_DEBUG(llvm::dbgs() << "Multiple writers to constant, wtf?\n");
+          return linalg::GenericOp();
+        }
+        ret = generic;
+      }
+    }
+  }
+  return ret;
+}
+
 static void moveTransformsBefore(PatternRewriter &b, Value val) {
-  if (auto defOp = val.getDefiningOp()) {
+  if (auto *defOp = val.getDefiningOp()) {
     if (auto rtop = dyn_cast<TransformOp>(defOp)) {
       moveTransformsBefore(b, rtop.getOperand());
+      defOp->moveBefore(b.getBlock(), b.getInsertionPoint());
+      b.setInsertionPointAfter(defOp);
+    } else if (isa<memref::GetGlobalOp, memref::AllocOp, rock::GpuAllocOp>(
+                   defOp)) {
+      if (b.getInsertionPoint()->isBeforeInBlock(defOp)) {
+        defOp->moveBefore(b.getBlock(), b.getInsertionPoint());
+        b.setInsertionPointAfter(defOp);
+      }
 
-      defOp->remove();
-      b.insert(defOp);
+      if (auto zeroArgConst = dyn_cast<rock::GpuAllocOp>(defOp)) {
+        linalg::GenericOp initLago = getConstInitLaGeneric(zeroArgConst);
+        if (!initLago)
+          llvm_unreachable(
+              "Sanity checks earlier in the pattern will have rejected this");
+        if (b.getInsertionPoint()->isBeforeInBlock(initLago)) {
+          initLago->moveBefore(b.getBlock(), b.getInsertionPoint());
+          b.setInsertionPointAfter(initLago);
+        }
+      }
     } else {
-      llvm_unreachable("must trace to func.arg");
+      llvm_unreachable("Cannot trace non-gemm input to known initilizer or "
+                       "function argument");
     }
   }
 }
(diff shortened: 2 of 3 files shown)


manupak:
Oh right if only thing this PR does handle orphaned linalg blocks that fills memref with constant data those could be folded early on. I just figured that this happens because that we dont fold unit dims before linalg elemwise fusion.

```c++
 git diff
diff --git a/mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp b/mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
index 18d3973402d6..8536fe9749f4 100644
--- a/mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
+++ b/mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
@@ -79,8 +79,8 @@ void rock::buildBufferizePipeline(OpPassManager &pm,
   // linalg tensor opts
   /* rocmlir-opt --linalg-fuse-elementwise-ops --linalg-fold-unit-extent-dims
    */
-  pm.addNestedPass<func::FuncOp>(createLinalgElementwiseOpFusionPass());
   pm.addNestedPass<func::FuncOp>(createLinalgFoldUnitExtentDimsPass());
+  pm.addNestedPass<func::FuncOp>(createLinalgElementwiseOpFusionPass());
   pm.addNestedPass<func::FuncOp>(rock::createRockViewToTransformPass());

   // bufferization
```

above would generate 

```mlir
// -----// IR Dump After LinalgElementwiseOpFusion (linalg-fuse-elementwise-ops) //----- //
func.func @mlir_quant_dot_dequantizelinear_add_mul_erf_add_mul_mul_quantizelinear(%arg0: tensor<1x1x3072xf32>, %arg1: tensor<32x384x768xi8>, %arg2: tensor<32x768x3072xi8>) -> tensor<32x384x3072xi8> attributes {arch = "gfx908", kernel = "mixr", num_cu = 48 : i64} {
  %cst = arith.constant 34.907238 : f32
  %cst_0 = arith.constant 5.000000e-01 : f32
  %cst_1 = arith.constant 1.000000e+00 : f32
  %cst_2 = arith.constant 0.707106769 : f32
  %cst_3 = arith.constant 4.23844496E-4 : f32
  %cst_4 = arith.constant 1.270000e+02 : f32
  %cst_5 = arith.constant -1.280000e+02 : f32
  %0 = bufferization.alloc_tensor() : tensor<32x384x3072xi32>
  %1 = rock.gemm %0 = %arg1 * %arg2 features =  mfma|dot|atomic_add storeMethod =  set {arch = "gfx908", numCU = 48 : i32} : tensor<32x384x3072xi32> = tensor<32x384x768xi8> * tensor<32x768x3072xi8> -> tensor<32x384x3072xi32>
  %expanded = tensor.expand_shape %1 [[0], [1], [2, 3, 4]] : tensor<32x384x3072xi32> into tensor<32x384x1x1x3072xi32>
  %2 = tensor.empty() : tensor<32x384x1x1x3072xi8>
  %3 = linalg.generic {indexing_maps = [affine_map<(d0, d1, d2, d3, d4) -> (d0, d1, d2, d3, d4)>, affine_map<(d0, d1, d2, d3, d4) -> (d2, d3, d4)>, affine_map<(d0, d1, d2, d3, d4) -> (d0, d1, d2, d3, d4)>], iterator_types = ["parallel", "parallel", "parallel", "parallel", "parallel"]} ins(%expanded, %arg0 : tensor<32x384x1x1x3072xi32>, tensor<1x1x3072xf32>) outs(%2 : tensor<32x384x1x1x3072xi8>) {
  ^bb0(%in: i32, %in_6: f32, %out: i8):
    %4 = arith.sitofp %in : i32 to f32
    %5 = arith.mulf %4, %cst_3 : f32
    %6 = arith.addf %5, %in_6 : f32
    %7 = arith.mulf %6, %cst_2 : f32
    %8 = math.erf %7 : f32
    %9 = arith.addf %8, %cst_1 : f32
    %10 = arith.mulf %6, %9 : f32
    %11 = arith.mulf %10, %cst_0 : f32
    %12 = arith.mulf %11, %cst : f32
    %13 = math.roundeven %12 : f32
    %14 = arith.minf %13, %cst_4 : f32
    %15 = arith.maxf %14, %cst_5 : f32
    %16 = arith.fptosi %15 : f32 to i8
    linalg.yield %16 : i8
  } -> tensor<32x384x1x1x3072xi8>
  %collapsed = tensor.collapse_shape %3 [[0], [1], [2, 3, 4]] : tensor<32x384x1x1x3072xi8> into tensor<32x384x3072xi8>
  return %collapsed : tensor<32x384x3072xi8>
}
```

Apologies for not catching this earlier, but I thought you were going to 'recursively' view through all the elementwise ops in when tracing to func arg source that are in the other branch (i.e. the branch in which rock.gemm is not found) when tracing from binary elementwise ops.

manupak:
I can confirm the crash does not happen when the const linalg block is folded away as you found out as well.

krzysz00:
Re the pipeline, wouldn't we want "fold unit dims; elementwise fusion; fold unit dims?"

Do you think that's a fix we should do instead of/in addition to this PR?

manupak:
Im fine with either : instead of or in addition. I ll let you decide what you want to do here.

>  wouldn't we want "fold unit dims; elementwise fusion; fold unit dims?"

I dont think elementwise fusion will introduce unit dims to be folded away but not folding unit dims might block some elemwise fusions as we are experiencing with this issue.

It really shouldn't have the need to do because linalg op space to i/o space affine mapping are capable of denoting any broadcasting whatsoever -- at least in theory. However, it wouldn't hurt either to do that if we are suspicious about that. 

TLDR: ( I ll dump all my thoughts here just for ya'll information and not as any objection to the PR)

# Long term solution 

As far as I see the broader problem is that we need to propogate up the tiling in a potential elemwise chain of the other input of immediate binary elementwise operator that is being fused.

Let me illustrate the 'a' general problem
### Fig1
```mermaid
flowchart TD
   In1 --> BinElemwiseOp1
   In2 --> BinElemwiseOp1
   Const1 --> BinElemwiseOp2
   BinElemwiseOp1 --> BinElemwiseOp2
   In3 --> rock.gemm
   In4 --> rock.gemm
   rock.gemm -->  BinElemwiseOp3
   BinElemwiseOp2 --> BinElemwiseOp3
   BinElemwiseOp3 --> Out1
```
## Direction 1
Ordinarily, if we 'decide' to rely on `linalg-elemwise-fusion` pass I d expect this to take the form of
### Fig2
```mermaid
flowchart TD
   In3 --> rock.gemm
   In4 --> rock.gemm
   rock.gemm -->  FusedLinalgBlock
   In1 --> FusedLinalgBlock
   In2 --> FusedLinalgBlock
   FusedLinalgBlock --> Out1
```

Note here that I took the assumption that we bound the Const1 because its 'small enough' to be put in registers.
If it involved a large constant tensor; it should not be bound and treated as any other input by the graph compiler -- this is generally what graph compilers does -- MIGraphX currently only binds splat constants.

## Direction 2 
We want to rely less on  `linalg-elemwise-fusion` pass going into the future...

Then we need to handle Fig1 without taking any assumption about high-level passes (at least being the goal).
If this is the direction we want take, I can see an initial step is taken in this PR by handling the specific scenario as follows : 

```mermaid
flowchart TD
   Const1 --> BinElemwiseOp1
   In1 --> rock.gemm
   In2 --> rock.gemm
   rock.gemm -->  BinElemwiseOp1
   BinElemwiseOp1 --> Out1
```

Here, this PR handle this specific case where Const1 is an orphaned linalg.generic block pushing constants to memref -- which resolves the immediate problem.

## Summary

So `Direction1` and `Direction2` are not strictly mutually exclusive.
If  `Direction2` is what we want to take, I agree with you this PR does take a step in that direction.

for  `Direction1`, I dont think this sort of handling is needed -- however, we need to think if there are cases where we wont be able to use  `linalg-elemwise-fusion` -- which I doubt. we could even call that pass just before `rock-align-tiling`.  However, since you have done the work, It doesn;t hurt to keep the pass resilient to this case.

manupak (approved):
While this is needed or not is a seperate concern, but since you've done this I have no issues of having this capability in rock-align-tiling.

krzysz00:
Having investigated the sigmoid failure, it turns out MIGraphX is entirely willing to send us the following case

```
module {
  func.func @mlir_transpose_slice_dot_add_add_tanh_add_sigmoid_sub_mul_mul_add(%arg0: tensor<1x5xf32>, %arg1: tensor<2x5xf32>, %arg2: tensor<2x5xf32>, %arg3: tensor<2x5xf32>, %arg4: tensor<2x5xf32>, %arg5: tensor<2x5xf32>, %arg6: tensor<2x5xf32>, %arg7: tensor<15x5xf32>) -> tensor<2x5xf32> attributes {arch = "gfx1100", kernel = "mixr", num_cu = 48 : i64} {
    %0 = migraphx.multibroadcast(%arg0) {out_dyn_dims = [], out_lens = [2, 5]} : (tensor<1x5xf32>) -> tensor<2x5xf32>
    %1 = migraphx.transpose(%arg7) {permutation = [1, 0]} : (tensor<15x5xf32>) -> tensor<5x15xf32>
    %2 = migraphx.slice(%1) {axes = [1], ends = [15], starts = [10]} : (tensor<5x15xf32>) -> tensor<5x5xf32>
    %3 = migraphx.dot(%arg6, %2) : (tensor<2x5xf32>, tensor<5x5xf32>) -> tensor<2x5xf32>    %4 = migraphx.add(%3, %0) : (tensor<2x5xf32>, tensor<2x5xf32>) -> tensor<2x5xf32>
    %5 = migraphx.add(%arg1, %4) : (tensor<2x5xf32>, tensor<2x5xf32>) -> tensor<2x5xf32>    %6 = migraphx.tanh(%5) : (tensor<2x5xf32>) -> tensor<2x5xf32>
    %7 = migraphx.add(%arg2, %arg3) : (tensor<2x5xf32>, tensor<2x5xf32>) -> tensor<2x5xf32>
    %8 = migraphx.sigmoid(%7) : (tensor<2x5xf32>) -> tensor<2x5xf32>
    %9 = migraphx.sub(%arg4, %8) : (tensor<2x5xf32>, tensor<2x5xf32>) -> tensor<2x5xf32>    %10 = migraphx.mul(%9, %6) : (tensor<2x5xf32>, tensor<2x5xf32>) -> tensor<2x5xf32>
    %11 = migraphx.mul(%8, %arg5) : (tensor<2x5xf32>, tensor<2x5xf32>) -> tensor<2x5xf32>
    %12 = migraphx.add(%10, %11) : (tensor<2x5xf32>, tensor<2x5xf32>) -> tensor<2x5xf32>    return %12 : tensor<2x5xf32>
  }
}
```

Please note `%7 = migraphx.add(%arg2, %arg3)`, which is, presumably, the result of non-splat constants being passed as arguments (which is, IMO, a decision that should be revisited, but anyway).

On account of this - and because I don't trust case 1 above to not cause resiliency issues down the road, I suspect I'll closing this PR in favor of a strategy I'd rejected earlier - adding a dummy threadwise_write_all during regularization in order to give the fusion flow something to latch on to.

krzysz00:
... actually, no, that won't be needed, I can generalize this change

krzysz00:
Or, on the third hand, the pass-shuffling solution is an option.

Though I'll also note
`/// Pass that fuses generic ops on tensors. Used only for testing.`
on the ELementwiseFusion pass we're using.

We may want to end up assembling our own version of that pass overall.

I'll be dealing with all this on Monday or something.

manupak:
https://github.com/llvm/llvm-project/blob/0c9c9dd9a24f9d715d950fef0ac7aae01437af96/mlir/lib/Dialect/Linalg/Transforms/ElementwiseOpFusion.cpp#L9

If your worry is about 'only for testing', it seems to have disappeared.
It is used by many projects I know and it should be good enough.

I am generally opposed to maintaining duplicate passes downstream. I d
rather fix things upstream using that effort. Im sure they d be more than
happy to accept fixes/improvements.

It kind of make sense to do the fusion on SSA - we dont have to do this
'tracing' on non-SSA IR.

Also with planned work -- moving type conversions out of gridwise_gemm --
we should probably be moving that out of rock.gemm on tensors. We should be
calling the linalg fusion pass there.

krzysz00:
Superseded by #1222, closing (@manupak for taking a look at that approach)

PR #1211, opened 2023-08-23
Improve loads and stores involving scalar memrefs.

krzysz00:
Expand the scalar memref optimization the global load lowering had to apply to threadwise_read_into operations from non-global memories.

Extend that optimization in global loads and stores to handle the case where the tensor has 1 element, not just when it's 0D.

However, remove "assume it's in bounds" optimization from atomic stores whose validity isn't trivially true. This is because, just because all threads write to the same place doesn't mean they all have valid data to atomically {add, max, ...} there.

We were probably going to trip over this eventually and waste a bunch of time, hopefully this fix wards that off.

Improve the documentation on moveTransformsBefore() to make it more obvious what that function does.

Adjust the fusion tiling flow to not create unrolled loops for what are essentially fill/memset operations.

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
@@ -530,14 +530,10 @@ static Value makeRegs(LinalgAlignRewriter &b, MemRefType::Builder &mrb,
                                        srcMemType.getElementType())));
 }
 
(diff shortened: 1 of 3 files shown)


krzysz00:
@manupak @giuseros Ping

manupak (approved):
(extremely sorry for missing this)
LGTM!

PR #1217, opened 2023-08-25
Convert the existing Resnet50 to be consistent how migraphx run their verification.#1053

stefankoncarevic:
Change the run line for verify model ResNet50 in way how MIGraphX team use that.

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -874,7 +874,7 @@ pipeline {
                                         sh 'make -j$(nproc) driver test_gpu_mlir'
                                         sh 'ctest -R test_gpu_mlir'
                                         // Verify ResNet50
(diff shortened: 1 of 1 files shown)


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -852,7 +852,7 @@ pipeline {
                                         sh 'make -j$(nproc) driver test_gpu_mlir'
```
jerryyin, line 874: You may want to `make -j$(nproc) migraphx-driver`?

stefankoncarevic: The target name is actually driver. I don't know original reason for that.

jerryyin (approved):

krzysz00:
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/1796 is related, I think

PR #1219, opened 2023-08-25
Add back a few large GEMM tests

krzysz00:
I suspect that we're running into more oom issues because of tharnsB and transC and the like. So, move large gemm tests into their own file where there'll only be one or two tests that run, thus ensuring we don't overload the machines.

(Note: historicallly, we tried to add large GEMM tests to the usual GEMM test set and we'd get OOMs from parallel runs on memeory-constrained machines)

### DIFF
--- a/mlir/test/e2e/CMakeLists.txt
+++ b/mlir/test/e2e/CMakeLists.txt
@@ -50,6 +50,7 @@ if (ROCK_E2E_TEST_ENABLED)
   if (ROCK_E2E_TEST_SUITES STREQUAL "" OR ROCK_E2E_TEST_SUITES STREQUAL "part3")
     list(APPEND CONFIGS
     Resnext101Config
+    LargeGemm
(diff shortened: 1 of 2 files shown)


krzysz00:
Ping

manupak:
> I suspect that we're running into more oom issues because of tharnsB and transC and the like

Is this true now ? I cant recall any issues of the like... (unless its a thing that happened last week whilst im out)

> So, move large gemm tests into their own file where there'll only be one or two tests that run, thus ensuring we don't overload the machines.

If anything this PR adds more tests and does not reduce any workload on the machine. So what is the actual motivation for the addition of tests? 

Im not opposed to merging this in; its just the motivation written in the description is confusing.

krzysz00:
@manupak The motivation was from back when we were trying to add even more of these tests and they were causing CI hangs because the test took so long

krzysz00:
Have updated description

PR #1221, opened 2023-08-25
Enable Navi3x in nightly CI

yiqian1:
Navi3x is enabled in nightly E2E tests, but not in performance tests yet.

- Disable conv2d_wrw_perf_config as the config is not applicable to Navi3x ([issue#1078](https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1078)).
- Increase difference thresholds for conv2d_regression_fwd tests on Navi3x as a workaround to [issue#1115](https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1115).
- Disable test cases using the bf16 data type as a workaround to [issue#1080](https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1080).
- Run ROCK_E2E tests in multiple smaller batches on Navi3 to increase the chance of successful completion. 
- Ensure that Jenkins jobs continue running in case of a timeout or failure during Navi3x E2E tests.

The nightly E2E tests passed on Navi3x in [this build](http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-1221/48/pipeline/172).
[This build](http://rocmhead.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-1221/42/pipeline/151) shows the jenkins job continues in case of a timeout on the Navi3x stage.


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -225,6 +225,9 @@ void check_randomE2ETests(String codepath) {
               -DLLVM_LIT_ARGS='-v --time-tests ${ limit_lit_workers ? '-j 8' : ' ' }'
               -DCMAKE_EXPORT_COMPILE_COMMANDS=1
              """)
+    timeout(time: 60, activity: true, unit: 'MINUTES') {
+        sh "cd build; ninja check-rocmlir || ${codepath} == 'navi3x'"
```
krzysz00: I don't think this is valid shell syntax. 

Also, it's wort checking (though it might not be possible) if there's a way to suppress failures from within a Jenkins pipeline.

krzysz00 (commented):
I'm not sure I like this arbitrary stages approach to the tests. I don't really understand what it buys us.

yiqian1:
> I'm not sure I like this arbitrary stages approach to the tests. I don't really understand what it buys us.

Through my experiments, I've discovered that running the E2E tests in multiple small batches would give me a better chance of passing them.

krzysz00:
I'd rather not do that and instead focus on fixing that underlying issue instead of complicating our test system

yiqian1:
> I'd rather not do that and instead focus on fixing that underlying issue instead of complicating our test system

We're collaborating with other groups to investigate the underlying causes to [the Navi3x stability issue](https://ontrack-internal.amd.com/browse/SWDEV-422198). We are uncertain about the timeline for resolving the issue. On the other hand, we want to cover Navi3 in our nightlies to identify any potential issues on our end.

jerryyin (approved):
Could you file a follow-up (tracking) ticket on reverting the test split-up once the stability issue has been figured out?

PR #1222, opened 2023-08-29
Fix handling of linalg.generic blocks not tied to gemm inputs

### LINKED ISSUE #1188: Crash on align tiling when running RNN GRU models enabling sigmoid
After fixing a missing comma, MLIR will now work with sigmoid operators, but this causes the `./bin/test_verify test_gru_bidirct` test to fail in migraphx on mi100:

```
UNREACHABLE executed at /usr/local/cget/build/tmp-832dcb3bff0f4556884f71aa299c95b7/rocMLIR-1ad9d6df32acc6d29d58e8ed6710e36746d0a4d6/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp:98!
```

krzysz00:
This commit fixes the handling of linalg.generic blocks where none of
the inputs participate in a chain of writers that lead to a gemm, but
that feed into such chains. For example, the linalg.generic blocks
that ultimately result from the `migraphx.add(%arg2, %arg3)` in the
example test.

The general solution requires us to add a second case to the
linalg.generic fusion matching. The fusion logic would previously scan
the inputs of a linalg.generic for a threadwise_write_all, which it
would use to get a tile size, set of global coordinates, and a set of
transformations that had to be applied to those arguments in order to
make them all indexable with said global coordinates.

Now, we support a second case: tracing from the output of a
linalg.generic to a threadwise_read_into that was generated by another
round of fusion. That threadwise_read_into then gives us the tiling
information we need to efficiently load a relevant tile of any
additional constant arguments.

Implementing this "visit a generic that we can't do
anything with but might handle later" pattern required control over
the pattern matching worklist that none of the upstream pattern
rewrites could provide. Therefore, this commit defines its own
subclass PatternRewriter - the LinalgAlignRewriter, that lets us
schedule revisits after an initial top-down pass. This rewriter also
provides failure behavior that is more in keeping with how we're using
the rewrite patterns - failure of any rewrite indicates malformed IR
that should fail the fusion pass, as opposed to the current approach
where it leads to stray memref.alloc() ops that throw errors near the
backend.

The fusion flow is relatively unchanged, aside from some variables
being remaned for clarity. The main changes are that the code which
creates threadwise_read_into operations now goes in and schedules the
linalg.generic that wrote the buffer (if any) for re-processing.

Other changes including traceToThreadwiseWriteAll() to
traceToWriter(), removing a function that was equivalant to
untransform().

The fusion flow is also changed to move the writes after the relevant
linalg.generic instead of trying to clone things before the write so
as to prevent use-before-def problems that may have arisen with the
old approach.

We also fix handling of 0-D inputs to threadwise_read_into and
threadwise_write_all in blockwise-gemm-to-threadwise since I needed
that in a previous iteration of the PR and there's no reason to let
that code go to waste.

There is also now a new version of untransform() that doesn't package
the attributes into an ArrayAttr.

Fixes #1188

Fixes
https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1098

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
@@ -49,6 +49,7 @@
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/ScopedPrinter.h"
 
+#include <deque>
 #include <numeric>
 
 namespace mlir {
@@ -91,14 +92,12 @@ protected:
   llvm::ScopedPrinter logger{llvm::dbgs()};
 #endif
 
-  // Ensure we completely process one generation of operations before moving
-  // to the next one by keeping two worklists that we copy into.
-  SmallVector<Operation *, 0> nextWorklist;
-  // Our patterns aren't stable under repetade application (ex. fuling an
+  std::deque<Operation *> worklist;
+  // Our patterns aren't stable under repeated application (ex. fuling an
   // already fused op) but we can have cases where an operation is marked for
   // multiple visits (ex. two arguments to a linalg.generic reading from
   // the same source linalg.generic), so we need to prevent duplicate
-  // scheduling within a generation.
+  // scheduling.
   llvm::SmallPtrSet<Operation *, 4> scheduledOps;
 
   SmallPtrSet<Operation *, 2> opsExpectingRevisit;
@@ -112,8 +111,7 @@ protected:
 public:
   LinalgAlignRewriter(MLIRContext *ctx);
 
-  /// Schedule the given operation for processing in a future iteration
-  /// through the patterns.
+  /// Schedule the given operation for processing.
   void scheduleVisit(Operation *op);
 
   /// Inform the pattern rewriter that this operation will need to be revisited
@@ -122,9 +120,9 @@ public:
   /// is never revisited, then fail.
   LogicalResult needsRevisit(Operation *op);
 
-  /// Match all worklist itemis against the patterns in `matcher`, repetaing
+  /// Match all worklist itemis against the patterns in `matcher`, repeating
   /// this procedure until all scheduled visits are complete, including those
-  /// added during a pattern match. Fails if any pattern applicatino fails.
+  /// added during a pattern match. Fails if any pattern application fails.
   /// or if an operation that is marked as needing revisiting is not visited
   /// after it is so marked.
   LogicalResult drainWorklist(PatternApplicator &matcher);
@@ -225,7 +223,7 @@ void LinalgAlignRewriter::scheduleVisit(Operation *op) {
   constexpr llvm::StringLiteral dupePrefix("** Duplicate:  ");
   if (scheduledOps.insert(op).second) {
     logOpActivity(prefix, op);
-    nextWorklist.push_back(op);
+    worklist.push_back(op);
   } else {
     logOpActivity(dupePrefix, op);
   }
@@ -239,54 +237,48 @@ LogicalResult LinalgAlignRewriter::needsRevisit(Operation *op) {
 }
 
 LogicalResult LinalgAlignRewriter::drainWorklist(PatternApplicator &matcher) {
-  SmallVector<Operation *, 0> currentWorklist;
-  while (!nextWorklist.empty()) {
-    // Copy the operations to be visited out of the way so we don't have
-    // iterator invalidation.
-    currentWorklist.append(nextWorklist);
-    nextWorklist.clear();
-    scheduledOps.clear();
-    for (Operation *op : currentWorklist) {
+  while (!worklist.empty()) {
+    Operation *op = worklist.front();
+    worklist.pop_front();
 #ifndef NDEBUG
-      auto canApply = [&](const Pattern &pattern) -> bool {
-        LLVM_DEBUG({
-          logger.startLine() << "Applying pattern " << pattern.getDebugName()
-                             << " (matches" << *pattern.getRootKind() << ")"
-                             << " on " << op->getName() << "\n";
-          logger.indent();
-        });
-        return true;
-      };
-      auto onFailure = [&](const Pattern &pattern) {
-        LLVM_DEBUG({
-          logger.unindent();
-          logger.startLine() << "Failed to match the " << pattern.getDebugName()
-                             << " pattern (matches " << *pattern.getRootKind()
-                             << ") on " << op->getName() << "\n";
-        });
-      };
-      auto onSuccess = [&](const Pattern &pattern) -> LogicalResult {
-        LLVM_DEBUG({
-          logger.unindent();
-          logger.startLine() << "Matched " << pattern.getDebugName()
-                             << " pattern on " << op->getName() << "\n";
-        });
-        return success();
-      };
+    auto canApply = [&](const Pattern &pattern) -> bool {
+      LLVM_DEBUG({
+        logger.startLine() << "Applying pattern " << pattern.getDebugName()
+                           << " (matches" << *pattern.getRootKind() << ")"
+                           << " on " << op->getName() << "\n";
+        logger.indent();
+      });
+      return true;
+    };
+    auto onFailure = [&](const Pattern &pattern) {
+      LLVM_DEBUG({
+        logger.unindent();
+        logger.startLine() << "Failed to match the " << pattern.getDebugName()
+                           << " pattern (matches " << *pattern.getRootKind()
+                           << ") on " << op->getName() << "\n";
+      });
+    };
+    auto onSuccess = [&](const Pattern &pattern) -> LogicalResult {
+      LLVM_DEBUG({
+        logger.unindent();
+        logger.startLine() << "Matched " << pattern.getDebugName()
+                           << " pattern on " << op->getName() << "\n";
+      });
+      return success();
+    };
 #else
-      function_ref<bool(const Pattern &)> canApply = {};
-      function_ref<void(const Pattern &)> onFailure = {};
-      function_ref<LogicalResult(const Pattern &)> onSuccess = {};
+    function_ref<bool(const Pattern &)> canApply = {};
+    function_ref<void(const Pattern &)> onFailure = {};
+    function_ref<LogicalResult(const Pattern &)> onSuccess = {};
 #endif
-      opsExpectingRevisit.erase(op);
-      LogicalResult matchResult =
-          matcher.matchAndRewrite(op, *this, canApply, onFailure, onSuccess);
-      if (failed(matchResult)) {
-        LLVM_DEBUG(llvm::dbgs() << "Pattern match failed\n");
-        return failure();
-      }
+    scheduledOps.erase(op);
+    opsExpectingRevisit.erase(op);
+    LogicalResult matchResult =
+        matcher.matchAndRewrite(op, *this, canApply, onFailure, onSuccess);
+    if (failed(matchResult)) {
+      LLVM_DEBUG(llvm::dbgs() << "Pattern match failed\n");
+      return failure();
     }
-    currentWorklist.clear();
   }
   if (!opsExpectingRevisit.empty()) {
 #ifndef NDEBUG
--- a/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
@@ -21,6 +21,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/GPU/IR/GPUDialect.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/Linalg/Utils/Utils.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
@@ -36,6 +37,8 @@
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Interfaces/SideEffectInterfaces.h"
 #include "mlir/Pass/PassManager.h"
+#include "mlir/Rewrite/FrozenRewritePatternSet.h"
+#include "mlir/Rewrite/PatternApplicator.h"
 #include "mlir/Transforms/DialectConversion.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 #include "mlir/Transforms/Passes.h"
@@ -44,6 +47,7 @@
 #include "llvm/ADT/SmallSet.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/Support/Debug.h"
+#include "llvm/Support/ScopedPrinter.h"
 
 #include <numeric>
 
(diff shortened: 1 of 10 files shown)


sjw36 (approved):
Great improvement @krzysz00 ! 

Where does it add the tiling/transform specifics to the pseudo tw-write for the real tiling of the kernel?

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -577,15 +577,39 @@ struct ThreadwiseWriteAllRewritePattern
 };
 } // end anonymous namespace
 
+/// Amend the operation chain (and copmuted shape) for a read/write to add a
```
sjw36: sp: copmuted

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -577,15 +577,39 @@ struct ThreadwiseWriteAllRewritePattern
...
+/// length-1 iteration index to 0-dimensinoal (scalar) buffers.
```
sjw36: sp: dimensinoal

krzysz00:
@sjw36 
> Where does it add the tiling/transform specifics to the pseudo tw-write for the real tiling of the kernel?
I'm not sure I understand the question

The basic structure of the transforms on the `rock.alloc()`'d buffer in the pseudo-write case is to take a 1D iteration index, expand it into N-D tensor indices, and then flatten the tensor indices back together when actually writing into the underlying buffer. That 1D -> ND expansion doesn't need to be propagated to the rest of the function, because they're expecting a N-D value to index into (the original `memref.alloc`). So, we replace that alloc with a N-D view of the 1-D array we created.

manupak:
Having another read through, I still have doubts in this approach - we might want to take that offline.
My concern is (allegedly) this might not work when other input is larger than a thread subtile being written by each thread in the gemm.

So let me check if I understand the added tests : 

1) [linalg-generic-sigmoid-elementwise-block.mlir](https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1222/files#diff-8092236c7bfbcb6b82fbeeaf247e2d3774502b709a61ccb0934f580564b55db7)
```mlir
...
%alloc = memref.alloc() : memref<2x5xf32>
%4 = rock.transform %alloc by #transform_map4 : memref<2x5xf32> to memref<1x2x5xf32>
%8 = rock.transform %4 by #transform_map8 : memref<1x2x5xf32> to memref<1x32x32xf32>
%c0 = arith.constant 0 : index
%11 = rock.alloc() : memref<16xf32, #gpu.address_space<private>>
...
rock.threadwise_write_all features = dot|atomic_add|atomic_fmax_f32 {forceUnroll, useIndexDiffs} %11 -> [#transform_map32, #transform_map33](%8) [%c0, %c0, %c0, %c0] by set : memref<16xf32, #gpu.address_space<private>> -> memref<1x32x32xf32>
```
In above test, a thread subtile has 16 elements and other input is 2x5=10 elements.

2) [linalg-generic-sigmoid-elementwise-scalar-block.mlir](https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1222/files#diff-3dc27ce005ac3ed5d6536f52275d61277de7db031ab0fd2c467c5eb97d1a4edf)

```mlir
%alloc = memref.alloc() : memref<f32>
    %4 = rock.transform %alloc by #transform_map4 : memref<f32> to memref<1x1x1xf32>
    %8 = rock.transform %4 by #transform_map8 : memref<1x1x1xf32> to memref<1x32x32xf32>
    %c0 = arith.constant 0 : index
    %11 = rock.alloc() : memref<16xf32, #gpu.address_space<private>>
    rock.threadwise_write_all features =  dot|atomic_add|atomic_fmax_f32 {forceUnroll, useIndexDiffs} %11 -> [#transform_map32, #transform_map33](%8) [%c0, %c0, %c0, %c0] by  set : memref<16xf32, #gpu.address_space<private>> -> memref<1x32x32xf32>
```
In above test, every I/O is a single element.

3) const initializer is not relevant to this concern as it does not involve an 'another' input.

manupak:
I would've approached this differently -- maybe this is because I dont understand how tiling info is propogated upwards towards other inputs in the current approach.

My approach would;ve been : 

1) We do the linalg rewrite on the one (there should be atleast one) that leads to threadwise_write_all of gemm-like op and add threadwise_read_into using correct tiling info (a.k.a. extraViews on threadwise_write_all ) the other inputs irrespective of them being func args.
2) Then we do threadwise_read_into rewrite to propagate upwards -- if threadwise_read_into traces to an linalg-generic output, we add the same threadwise_read_into s onto linalg-generic's input.
3) We keep doing this until allthreadwise_read_into s read from func args.

I just stated this because thats just 'one' way to do this. However, I would like to understand how the current approach does all that.

sjw36:
@manupak Current approach basically does #1 above, but fails to propagate up as in #2.

sjw36:
> The basic structure of the transforms on the `rock.alloc()`'d buffer in the pseudo-write case is to take a 1D iteration index, expand it into N-D tensor indices, and then flatten the tensor indices back together when actually writing into the underlying buffer. That 1D -> ND expansion doesn't need to be propagated to the rest of the function, because they're expecting a N-D value to index into (the original `memref.alloc`). So, we replace that alloc with a N-D view of the 1-D array we created.

But how do you get the tiling in a `generic` that isn't connected to a `threadwise_write_all`? 

If you process them independently and put `threadwise_read/writes`s around it, how do you reconcile with the tiling coming from the rock.gemm op?

krzysz00:
@sjw36 Here's what happens, going off the sigmoid example for concrete numbers
1. The original output of the non-aligned fusion was a `memref<2x5xf32>`
2. We replace that allocation with a `memref<2x5xf32, #private>`, which is a `rock.transform`'d version of the `memref<10xf32 #private>` we used to store the result.
3. The value we replaced is subject to any transforms needed do align it with the generics it participates in (in this case, a GEMM + pointwise fusion, but it could be another untiled one)
4. When the subsequent generics are processed, `rock.threadwise_read_into` will eventually reach that `memref<10xf32, #private>` as source. However, that source will be transformed so that it aligns with the coordinates in the fusion, which are (for the purposes of this example) the GEMM coordinates.

The `write_all` on the unalignable generic is a temporary construct that gets erased before processing is finished.

Also, re the "elementwise input > thread tile size" case, we'll still have that elementwise allocation transformed into GEMM coordinates and we'll `threadwise_read_into` the appropriate subset of the values.

Do we want to have a meeting?

krzysz00:
@manupak If you have time for a review before tomorrow's scrum, I'd really appreciate it

manupak:
Im not sure whether you need this "scheduling revisits" feature.

https://mlir.llvm.org/docs/PatternRewriter/#greedy-pattern-rewrite-driver

"This driver processes ops in a worklist-driven fashion and greedily applies the patterns that locally have the most benefit. The benefit of a pattern is decided solely by the benefit specified on the pattern, and the relative order of the pattern within the pattern list (when two patterns have the same local benefit). Patterns are iteratively applied to operations until a fixed point is reached or until the configurable maximum number of iterations exhausted, at which point the driver finishes."

"By default, ops that were modified in-place and newly created are added back to the worklist. Ops that are outside of the configurable “scope” of the driver are not added to the worklist. Furthermore, “strict mode” can exclude certain ops from being added to the worklist throughout the rewrite process:"

So please help me understand why you need this explicit scheduled rewrites because say you have two rewrite patterns:
A1) matching logic that traces input to threadwise_write_all
A2) matching logic that traces output to threadwise_read_into

So in the initial worklist, only A1) will be matched in a single op -- this is expected others would just be match failures.
Once that happens the modified op will be placed to worklist in the top as mentioned in the docs (unless you found an upstream bug). Then all the patterns will be tried iteratively in the worklist. Beyond the first round, there will be only matches for A2) and they;d only be happening closest to the threadwise_write_all. Thus for each match for A2) more items will be pushed to the worklist  until all there is no ouput leading to a threadwise_read_into.

Therefore my understanding contradicts with your statement : 
> Implementing this "visit a generic that we can't do
anything with but might handle later" pattern required control over
the pattern matching worklist that none of the upstream pattern
rewrites could provide.

manupak:
I think I get it now.. having looked at it again.
Because these are two different linalg blocks...

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -65,96 +69,329 @@ struct RockLinalgAlignPass
...
+struct LinalgAlignRewriter : public PatternRewriter,
+                             public RewriterBase::Listener {
+protected:
+#ifndef NDEBUG
+  llvm::ScopedPrinter logger{llvm::dbgs()};
+#endif
+
+  // Ensure we completely process one generation of operations before moving
+  // to the next one by keeping two worklists that we copy into.
+  SmallVector<Operation *, 0> nextWorklist;
```
manupak: usually this is done by maintaining a queue where pop from back and push from front.
No need to maintain two worklists. Why would that not work here ?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -65,96 +69,329 @@ struct RockLinalgAlignPass
...
+  // Our patterns aren't stable under repetade application (ex. fuling an
```
manupak: nit : typos

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -65,96 +69,329 @@ struct RockLinalgAlignPass
...
+  /// through the patterns.
+  void scheduleVisit(Operation *op);
+
+  /// Inform the pattern rewriter that this operation will need to be revisited
+  /// in the future to make everything correct. Returns success() (we matched,
+  /// but have to wait to actually run this) as a convenience. If the operation
+  /// is never revisited, then fail.
+  LogicalResult needsRevisit(Operation *op);
+
+  /// Match all worklist itemis against the patterns in `matcher`, repetaing
```
manupak: nit : typos

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -65,96 +69,329 @@ struct RockLinalgAlignPass
...
+  /// this procedure until all scheduled visits are complete, including those
+  /// added during a pattern match. Fails if any pattern applicatino fails.
```
manupak: nit : typos

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -65,96 +69,329 @@ struct RockLinalgAlignPass
...
+  opsExpectingRevisit.insert(op);
+  return success();
+}
+
+LogicalResult LinalgAlignRewriter::drainWorklist(PatternApplicator &matcher) {
+  SmallVector<Operation *, 0> currentWorklist;
+  while (!nextWorklist.empty()) {
+    // Copy the operations to be visited out of the way so we don't have
+    // iterator invalidation.
+    currentWorklist.append(nextWorklist);
```
manupak: yep! I think a queue should just work here.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -239,94 +459,255 @@ static LogicalResult checkUniqueReader(Operation *op, Operation *reader,
...
-    for (Operation *nonViewUse : nonViewUsers) {
-      if (isa<memref::DeallocOp>(nonViewUse)) {
-        // ignore
-        continue;
+  Operation *result = nullptr;
+  Operation *startValDef = startVal.getDefiningOp();
+  if (!startValDef)
+    return nullptr;
+  bool unique = true;
+  auto setResult = [&](Value val, Operation *theResult) {
```
manupak, line 476: This might read better as a seperate function and with a description. 
I dont think there aren;t many local variables you need to capture to warrant a lambda here. Anything Im mising ?

mlir/test/fusion/pr-e2e/mixr-complex-elementwise-functions-issue-1188.mlir
```diff
@@ -0,0 +1,18 @@
+// RUN: rocmlir-driver -kernel-pipeline=migraphx %s | rocmlir-driver -host-pipeline=partition,highlevel -targets %arch | rocmlir-gen -ph -rand 1 -rand_type float -fut mlir_transpose_slice_dot_add_add_tanh_add_sigmoid_sub_mul_mul_add --verifier clone - | rocmlir-driver -c -arch %arch | mlir-cpu-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_float16_utils%shlibext --entry-point-result=void | FileCheck %s
+// ALLOW_RETRIES: 2
+// CHECK: [1 1 1]
+module {
+  func.func @mlir_transpose_slice_dot_add_add_tanh_add_sigmoid_sub_mul_mul_add(%arg0: tensor<1x5xf32>, %arg1: tensor<2x5xf32>, %arg2: tensor<2x5xf32>, %arg3: tensor<2x5xf32>, %arg4: tensor<2x5xf32>, %arg5: tensor<2x5xf32>, %arg6: tensor<2x5xf32>, %arg7: tensor<15x5xf32>) -> tensor<2x5xf32> attributes {kernel = "mixr", num_cu = 48 : i64} {
```
manupak: This tests doesn;t help as partitioner doesn't do the right job here. 
Instead a useful test would be : 
1) take the output after partitioner
2) hand modify it to be how we want the kernel to be

kind of forcing the partition as we want it tested

manupak (commented):
Apologies for my early morning mis-read of the 'rewriter' issue you;ve addressed. I get it now.
It might be also be a good candidate to be upstreamed.

I broadly agree with the approach here. However, I;d like a test with a input memref that is larger than the thread sub-tile. (i.e. increasing the dimension sizes of the existing test as another test should work)

I'd also note partitioner will just partition the strictly trailing ops following the gemm-like kernel. So a manual partitioned test would be required here. 
here is an example : https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/develop/mlir/test/fusion/pr-e2e/reductions/tosa-gemm-add-reduce-max.e2e.mlir

mlir/test/fusion/pr-e2e/mixr-complex-elementwise-functions-issue-1188.mlir
krzysz00: I inspected the partitioner output and found it to include the problematic case?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
krzysz00: I'm capturing unique and result here.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
krzysz00: I'm concerned about iterator invalidation, but ... yeah, good point, this could be a queue

krzysz00:
Re your comment, I'm not sure I'd want to add another end-to-end test, as between the integration test and the unit tests, I think we can be confident that the tile propagates upward

manupak:
I dont think we have a test case in each of these where tile size != tensor size. Do we ?

manupak:
if you can make the 2x5 to something like 128x256, this would be more suitable to check tile size propogation.

krzysz00:
The tile size is 16 but the size of the tensors is 10 - though it gets padded out to 32x32

manupak:
> The tile size is 16 but the size of the tensors is 10 - though it gets padded out to 32x32

Exactly, its just that I dont doubt it will pass but its generally good have a 'proper' tiling done -- a.k.a. tile size < tensor size
Also I dont understand why it is hard to add one ? merely changing the migraphx.* test shape would do. Is there any underlying reason for the reluctance to add one ?

manupak (commented):
I have no further concerns here. 

Im just waiting on a tile size < other input tensor size test -- if your concern is about PR CI test time, Im fine with adding that to be in a nightly. I can understand with padding and all it might end being one but those are lowering details of the compiler -- but from a testing point of view, I cant see that we are *properly* testing tile size propagation case here when all operations deal with 2x5 tensors.

krzysz00:
@manupak I fundamentally don't understand your concern here. We know from the unit tests that we propagate the GEMM tile size up to all the other inputs.

manupak (approved):
First of all, I think the code here is doing the right thing.. my comments are not to be mistaken that I doubt the approach or the implementation.

My fundamental request is for a PR that is claiming to propagate tile size up to other inputs in the fusion process is to have a test that meaningfully exercise that in a e2e test (not just based on padding the input). 

I failed to see how a rock.alloc writing 16xf32 values per thread is being fused with a chain of operations in the middle that does elementwise ops for tensors/memrefs of 2x5xf32 is a good test to showcase what is being fixed here.

Anyhow, I dont feel strongly to drag this any further.

krzysz00:
On the other hand, I think I can add a test that exercizes another interesting case: a tile of size 1, but inputs that are meaningfully larger than that.

Yeah, I'll add a version of the test you're looking for, even if it'll be a bit synthetic.

krzysz00:
I'm going to land this to unblock MIGraphX and then add the nightly test in a followup PR.

manupak:
@krzysz00 I wrote a very similiar test to yours in a different PR. I still think e2e test here just verified CPU vs CPU.

rocmlir-driver -kernel-pipeline=migraphx /var/mkarunar/work/rocMLIR/mlir/test/fusion/pr-e2e/mixr-complex-elementwise-functions-issue-1188.mlir | rocmlir-driver -host-pipeline=partition,highlevel -targets gfx90a:sramecc+:xnack- | rocmlir-gen -ph -rand 1 -rand_type float -fut mlir_transpose_slice_dot_add_add_tanh_add_sigmoid_sub_mul_mul_add --verifier clone - | rocmlir-driver -c -arch gfx90a:sramecc+:xnack- > compile_1188.log

[compile_1188.log](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/12618681/compile_1188.log)

You ll find an empty xmodule as follows in the above log : 
```mlir
module @__xmodule_amdgcn_amd_amdhsa_gfx90a_srameccY_xnack_ attributes {gpu.container_module, mhal.arch = "amdgcn-amd-amdhsa:gfx90a:sramecc+:xnack-", mhal.module} {
  }
```

krzysz00:
Huh, yeah, you're right there. It really feels like this should work though.

krzysz00:
I found it! Use `-kernel-pipeline=gpu,binary -host-pipeline=mhal,runner` instead of `-c` in the last `rocmlir-driver`

manupak:
Let me try

manupak:
Thanks now it runs something in the GPU!

PR #1223, opened 2023-08-30
Add new heuristic initialization to perf_config

djramic:
Suggestion for a new heuristic list based on collected tuning data.


mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -452,29 +452,62 @@ LogicalResult PopulateParamsAccel::obtainTuningParameters(
...
+  {64, 64, 2, 32, 32, 8, true, true},   
+  {32, 64, 4, 32, 64, 4, true, true},   
+  {32, 64, 4, 8, 64, 4, true, true},   
+  {32, 64, 2, 32, 32, 8, true, true},   
+  {32, 32, 8, 16, 16, 4, true, true}, 
+  {32, 64, 2, 8, 64, 4, true, true},   
+  {16, 64, 8, 16, 16, 4, true, true}, 
+  {16, 32, 8, 16, 16, 8, true, true},  
+  {16, 16, 8, 16, 16, 8, true, true}, 
+  {8, 64, 8, 4, 64, 8, true, true}
```
krzysz00: Quick checks:
1. Were there no "4,*" configs near the top of tuning? I could've sworn I've seen them before.
2. Since @pfultz2 raised the importance of this point - are the best (or close to best) configs for the gemms in the tier 1 models all on this new list?
3. Does at least one of these configs apply to all problem sizes?

(for 3, you might want to chase through the applicability checks in the tuning code to make sure there's one that'll always work)

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -452,29 +452,62 @@ LogicalResult PopulateParamsAccel::obtainTuningParameters(
 const InitParamsAccel
 PopulateParamsXDL::initParameters[PopulateParamsXDL::nInitParameters] = {
```
jerryyin, line 453: You may need to update: `nInitParameters`, `nInitParametersFp16` and `nInitParametersForward8Bit` to match with the updated size.

Out of curiosity, the it might be able to compile getting rid of those hard-coded constants since this is all statically initialized and c++ compiler can figure out a size.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
jerryyin: I'd wait on @krzysz00's opinion but we'd likely to shoot to double the size of entries here (and fp16, int8) for better chance of hitting a good perf_config combination. (Especially those parameters in the range of 64, 128, that showed superior performance.)

giuseros: Hi @krzysz00 , about point 3) I was under the impression that every config is applicable to any size, given the right amount of padding.

krzysz00 (commented):
As a bit of feedback, do we need the lists split up by architecture? If the Navi3x list doesn't have a lot of overlap with the MI-x00 list, we should make them separately and make the function that returns the heuristic list take the architecture as a parameter.

krzysz00 (approved):

PR #1227, opened 2023-09-05
Introduce the rock.threadwise_transpose to ease the grouping of operations during a pipeline pass

giuseros:
This is to make the following IR generated in `gridwiseToBlockwisePass`: 
```
affine.for %arg3 = 0 to 12 {
      rock.threadwise_read_into ...
      rock.threadwise_read_into ...
      rock.transforming_for (%arg4) = {...} // -> packing A
      rock.transforming_for (%arg4) = {...} // -> packing B
      rock.lds_barrier
      rock.threadwise_write_all ...
      rock.threadwise_write_all ...
      rock.lds_barrier
      rock.blockwise_gemm_accel 
    }
```
To look like this:
```
affine.for %arg3 = 0 to 12 {
      rock.threadwise_read_into ...
      rock.threadwise_read_into ...
      rock.threadwise_transpose ... // -> packing A
      rock.threadwise_transpose ... // -> packing B
      rock.lds_barrier
      rock.threadwise_write_all ...
      rock.threadwise_write_all ...
      rock.lds_barrier
      rock.blockwise_gemm_accel 
    }
```

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -1260,7 +1260,7 @@ def Rock_ThreadwiseTransposeOp :
     Rock_Op<"threadwise_transpose">,
     AllElementTypesMatch<["source", "dest"]>,
     Arguments<(ins
-      Arg<MemRefOf<SupportedMemoryElems>, "source view", [MemWrite]>:$source,
+      Arg<MemRefOf<SupportedMemoryElems>, "source view", [MemRead]>:$source,
       Arg<MemRefOf<SupportedMemoryElems>, "detination view", [MemWrite]>:$dest,
       I64Attr : $kpack,
       UnitAttr:$forceUnroll,
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -1256,4 +1256,27 @@ def Rock_InWarpTransposeOp :
     # !cast<string>(swizzleGroupSize) # ";\n";
 }
 
+def Rock_ThreadwiseTransposeOp :
+    Rock_Op<"threadwise_transpose">,
+    AllElementTypesMatch<["source", "dest"]>,
+    Arguments<(ins
+      Arg<MemRefOf<SupportedMemoryElems>, "source view", [MemWrite]>:$source,
+      Arg<MemRefOf<SupportedMemoryElems>, "detination view", [MemWrite]>:$dest,
+      I64Attr : $kpack,
+      UnitAttr:$forceUnroll,
+      UnitAttr:$useIndexDiffs
+    )> {
+  let summary = "Threadwise register-to-register transpose";
+
+  let description = [{
+    `rock.threadwise_transpose` accepts a source view and a destination view and copy
+    data from source to destination.
+  }];
+
+  let assemblyFormat = [{
+    attr-dict $source  `->` $dest `:` type($source) `->` type($dest)
+  }];
+  let hasVerifier = 0;
+}
+
 #endif // ROCK_OPS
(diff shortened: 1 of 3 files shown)


mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1256,4 +1256,27 @@ def Rock_InWarpTransposeOp :
     # !cast<string>(swizzleGroupSize) # ";\n";
 }
 
+def Rock_ThreadwiseTransposeOp :
+    Rock_Op<"threadwise_transpose">,
+    AllElementTypesMatch<["source", "dest"]>,
+    Arguments<(ins
+      Arg<MemRefOf<SupportedMemoryElems>, "source view", [MemWrite]>:$source,
```
krzysz00: MemRead, perhaps?

krzysz00 (approved):
One nitpick, otherwise it's a nice little abstraction, approved pending that fix.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1256,4 +1256,27 @@ def Rock_InWarpTransposeOp :
...
+      Arg<MemRefOf<SupportedMemoryElems>, "detination view", [MemWrite]>:$dest,
+      I64Attr : $kpack,
+      UnitAttr:$forceUnroll,
+      UnitAttr:$useIndexDiffs
+    )> {
+  let summary = "Threadwise register-to-register transpose";
+
+  let description = [{
+    `rock.threadwise_transpose` accepts a source view and a destination view and copy
+    data from source to destination.
```
manupak, line 1273: This is a bit misleading description for what it does.
The current lowering only works for a very specific dimensionality in the upper views that needs to be K x D.
I think we need a op verifier too to ascertain this.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1256,4 +1256,27 @@ def Rock_InWarpTransposeOp :
...
+  }];
+
+  let assemblyFormat = [{
+    attr-dict $source  `->` $dest `:` type($source) `->` type($dest)
+  }];
+  let hasVerifier = 0;
```
manupak, line 1279: I think we need an op verifier because the lowering is very special cased.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -575,8 +576,98 @@ struct ThreadwiseWriteAllRewritePattern
...
+      untransform(b, destView);
+  assert(!needs64BitIdx && "Registers shouldn't need 64-bit indexing");
+  ArrayRef<int64_t> rawStoreBufferShape =
+      rawStoreBuffer.getType().cast<ShapedType>().getShape();
+
+  Value zero = b.createOrFold<arith::ConstantIndexOp>(loc, 0);
+  SmallVector<Value, 2> start(2, zero);
+  SmallVector<int64_t, 2> strides(2, 1);
+  int64_t vecLen = 1;
+  // The store buffer is a flattened < kouter x dPerThread x kpack >.
```
manupak, line 628: These assumption needs to be reflected in the operation description

manupak (commented):
@krzysz00 @giuseros I think the operation name is misleading because this operation cant be used to do *any* intra-thread transpose as it claims to be doing. 

This operation is very specific to transpose a register buffer in a thread in the way we want to load it to LDS for blockwise_gemm. Therefore, I suggest we change the name, description and op verifiers to be truthful to what it does. something along the lines : threadwise_pack_gemm_buffer (or simliar)

On the other hand,
a generic threadwise_transpose operation should not make assumption on the dimensionality and needs a better logic to figure out the vectorization that could work for both load buffer and store buffer -- That;d be something like getting the vector length inference from bottom dimensions of store buffer -- going through common view (e.g. k x d in this scenario) -- to the bottom dimensions of load buffer.

manupak:
Having thought about it a bit more....
You could canonicalize the generic version of the operation to be either : threadwise_read_into or threadwise_write_all that does reg -> reg read/write. It is just that view stacks need to be fused in the middle where they have common dimensionality.

PR #1228, opened 2023-09-05
Added `numCU` to `perfRunner.py` and `tuningRunner.py`

ravil-mobile:
* `perfTuner.py` accepts the legacy and new config formats - i.e., with and without `num_cu`
* `tuningRunner.py` generate tuning config according to the new format - i.e., with `num_cu`
* fixed trailing whitespace problem in `getTuningProblemStr`

Closes ROCmSoftwarePlatform/rocMLIR-internal#1104

### DIFF
--- a/mlir/utils/performance/tuningRunner.py
+++ b/mlir/utils/performance/tuningRunner.py
@@ -320,10 +320,10 @@ def main(args=None):
 
     # Note, appending results here to allow multiple config sets
     with open(parsed_args.output, 'a') as outFile:
-        print("# arch\ttestVector\tperfConfig", file=outFile)
+        print("# arch\tnumCUs\ttestVector\tperfConfig", file=outFile)
         for testVector, perfConfig in winners.items():
             print(f"Arch = {arch}({numCU} CUs), vector = '{testVector}', perfConfig = {perfConfig}")
-            print(f"{arch}\t{testVector}\t{perfConfig}", file=outFile)
+            print(f"{arch}\t{numCU}\t{testVector}\t{perfConfig}", file=outFile)
 
 if __name__ == '__main__':
     sys.exit(main())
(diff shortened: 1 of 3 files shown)


mlir/utils/performance/tuningRunner.py
```diff
@@ -320,10 +320,10 @@ def main(args=None):
 
     # Note, appending results here to allow multiple config sets
     with open(parsed_args.output, 'a') as outFile:
-        print("# arch\ttestVector\tperfConfig", file=outFile)
+        print("# arch\tnum CUs\ttestVector\tperfConfig", file=outFile)
```
manupak: super-nit : do we need the space between num and CU ?

manupak (approved):
LGTM!

@yiqian1 did we add something to run a single tuning cfg in PR CI ? 
Im just asking to see if changes tuningRunner.py gets verified in PR CI.

mlir/utils/performance/tuningRunner.py
ravil-mobile: I guess it doesn't matter because this line is just a comment. But, yep, let me remove the white space for consistency

yiqian1:
> LGTM!
> 
> @yiqian1 did we add something to run a single tuning cfg in PR CI ? Im just asking to see if changes tuningRunner.py gets verified in PR CI.

Yes. We tuned one gemm and one conv config in PR CI.

mlir/utils/performance/tuningRunner.py
yiqian1: We need the space to be consistent with the format of the following entries.

manupak: I suppose we are adding this for the first time in these scripts (I maybe wrong) .. which entries are you talking about ?

yiqian1 (approved):

mlir/utils/performance/tuningRunner.py
yiqian1: I was talking about line 326 where we added the entries to the database.

ravil-mobile: > super-nit : do we need the space between num and CU ?

I just removed the whitespace. As I mentioned before, line 323 generates just a comment which is getting ignored in migraphx. But, I still removed the whitespace for consistency

PR #1231, opened 2023-09-11
Fix wmma bug stemmed from LDS reduction algorithm

giuseros:
Second attempt to fix https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1119

There seems to be a bug in `unmerge`. 

Broadly speaking, if we consider three dimensions `d0`,`d1`,`d2` and  `d1` is always `0` (i.e., its length is `1`) there should be no difference between:
1.  `unmerge{2,1,16}` 
2. `unmerge{2,4,4}` 
3.  `unmerge{2,2,16}` 
4.  `unmerge{2,16,1}`

But 1 does not work, while 2,3,4 work. Unfortunately we are hitting a case where 1 is needed. 

If I move the `unmerge(2,1,16}` to an `embed{16,16,1}` this  seems to work fine.

I am adopting this solution for now, but I will spend some more time investigating this. If it is a simple fix, I will update this PR, otherwise I will create a ticket to track this bug.

### DIFF
--- a/mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
+++ b/mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
@@ -362,8 +362,6 @@ TopDownTMBuilder mlir::rock::swapThreadIdAndIteration(
     SmallVector<Attribute> &transformAttr) {
   TransformMapAttr toMatrixCAttr = toMatrixC.get();
   transformAttr.push_back(toMatrixCAttr);
-  int64_t mThreads = mPerBlock / copyMPerThread;
-  int64_t nThreads = nPerBlock / copyNPerThread;
 
   auto splitAgain = TopDownTMBuilder::below(toMatrixC, toMatrixCAttr);
   {
@@ -378,11 +376,12 @@ TopDownTMBuilder mlir::rock::swapThreadIdAndIteration(
       idx += 1;
     } else if (isBlockwise) {
       splitAgain.merge({"m_iter", "m_tid"}, {idx, idx + 1}, "gemmM",
-                       {copyMPerThread, mThreads});
+                       {copyMPerThread, mPerBlock / copyMPerThread});
       idx += 2;
     } else {
       splitAgain.merge({"m_block", "m_iter", "m_tid"}, {idx, idx + 1, idx + 2},
-                       "gemmM", {mBlocks, copyMPerThread, mThreads});
+                       "gemmM",
+                       {mBlocks, copyMPerThread, mPerBlock / copyMPerThread});
       idx += 3;
     }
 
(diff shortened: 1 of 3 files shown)


manupak:
Using an embed (unnecessarily) in this level, is going to make the view stacks non-invertible at this level.

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
```diff
@@ -409,20 +409,22 @@ TopDownTMBuilder mlir::rock::swapThreadIdAndIteration(
...
+      swapBack.embed("gemmM", idx, mBlocks * mPerBlock,
+                     {"m_block", "m_tid", "m_iter"},
+                     {mThreads * copyMPerThread, copyMPerThread, 1});
 
     if (!doSwapThreadIterSubDimsForN)
       swapBack.passThrough({"gemmN"}, {idx + 1}, {"gemmN"});
     else if (isBlockwise)
-      swapBack.unmerge("gemmN", idx + 1, {"n_tid", "n_iter"},
-                       {nPerBlock / copyNPerThread, copyNPerThread});
+      swapBack.embed("gemmN", idx + 1, mPerBlock, {"n_tid", "n_iter"},
```
manupak: why mPerBlock here ?

manupak (dismissed):
These embeds and unmerge should be identical.
However, the newly introduced embeds use mPerBlock where nPerBlock was used. Is that difference actually is the reason ? 

Im putting a hold because I strongly oppose introducing embeds unless the bug is confimed. IMO, which is unlikely.  If this PR is fixing the issue, I suppose it is because mPerBlock is used in the place of nPerBlock.

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
giuseros: This was a typo, but the bug is confirmed unfortunately

giuseros:
Unfortunately, the bug is confirmed. If you consider those maps:
```
#transform_map_unmerge_gemm =  #rock.transform_map<affine_map<(d0, d1, d2, d3, d4) -> (d0, d1, (d2 + d4) * 16 + d3)>
by [<PassThrough ["gemmG"] at [0] -> ["gemmG"] at [0]>,
     <PassThrough ["gemmM"] at [1] -> ["gemmM"] at [1]>,
     <Unmerge{48, 1, 16} ["n_block", "n_tid", "n_iter"] at [2, 4, 3] -> ["gemmN"] at [2]>] bounds = [1, 16, 48, 16, 1] -> [1, 16, 768]>

#transform_map_pad_gemm =  #rock.transform_map<affine_map<(d0, d1, d2) -> (d0, d1, d2)>
by [<PassThrough ["gemmG"] at [0] -> ["gemmG"] at [0]>,
    <Pad{0, 15} ["gemmMPad"] at [1] -> ["gemmM"] at [1]>,
    <PassThrough ["gemmN"] at [2] -> ["gemmN"] at [2]>] bounds = [1, 16, 768] -> [1, 1, 768]>
```
And you run the vectorizer on dim 2, the result will be one, but not for the right reason. It  thinks that the map chain is vectorizable, but the result of the vectorization is 1 because the last dimension (d2) is 1. However, this gives the framework to flatten the merge, which is wrong, because it is not vectorizable in this case.

giuseros:
Ok, this time I went to the core of the issue. Basically the engine was signalling the dimensions as contiguous, when they were not contiguous. This solution (hopefully) fixes that. Let's have a chat and see if this is acceptable

krzysz00 (approved):
Approved with the understanding that the comments will be improved per the call we had earlier today about how the dimensions exposed by the bug are actually contiguous but the structure of the `Unmerge` causes incorrect results with how we've implemented the contiguous dimension collapse logic.

@giuseros to file a followup ticket on a longer-term fix here

krzysz00 (approved):
Yep, that cleans up the documentation, you're now even more approved

PR #1233, opened 2023-09-12
Add code in Dockerfile.migraphx-ci to install the onnxruntime for accuracy_checker. 

stefankoncarevic:


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -878,6 +879,11 @@ pipeline {
                                     }
                                 }
                             }
+                            //Accuracy_checker will compare outputs from MIGraphX and onnx runtime
+                            dir('MIGraphX/tools/accuracy') {
+                                sh 'pip3 install -r requirements.txt'
```
krzysz00: I'm not happy with this CI-time install.

Could you move the requirements into the relevant Dockerfile instead?

krzysz00 (changes requested):

mlir/utils/jenkins/Jenkinsfile
stefankoncarevic: Thank you, I will remove that from Jenkinsfile and add to Dockerfile.

stefankoncarevic:
@krzysz00 
After your previous comment for CI-time install, I open the requirements and see the only two things to be install, one is the numpy and other is the onnxruntime. Because we have numpy already install, I add in our Dockerfile.migraphx-ci:
 RUN pip3 install onnxruntime==1.10.0.  After finish CI, I saw the message from MIGraphX: 
ModuleNotFoundError: No module named 'onnxruntime'.
I think that message is because onnxruntime is not install from Dockerfile.migraphx-ci? 
Is it a good way and place, where I wrote, or should it be different?

krzysz00:
@stefankoncarevic You'll need to make a separate commit for the Dockerfile changes

stefankoncarevic:
Thank you.

krzysz00 (approved):

PR #1234, opened 2023-09-12
Add support multiple n_blocks for second gemm in attention

manupak:
Currently the attention kernel always did all of N
of the second gemm in a single block according to
the original algorithm in : https://arxiv.org/pdf/2205.14135.pdf

However, this has two drawbacks:
1) Obvious parallelism loss
2) LDS usage becomes a function of problem cfg

2 is more concerning because this would require
the user to select tuning configs to compensate
larger N of the second gemm. I'd like to note
that this is also bit unlikely for self-attention
because N of the second gemm is head dimension and
it is considered to be relatively smaller.

Nonetheless, this makes kernel more robust in
general.

closes : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1124
builds on top of : https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1225

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
@@ -221,6 +221,7 @@ void AffixTuningParameters::affixTuningParametersImpl(AttentionOp op) {
   // Calculate (padded) grid size
   Value queries = op.getQueries();
   Value keys = op.getKeys();
+  Value values = op.getValues();
   SmallVector<int64_t, 3> queriesShape =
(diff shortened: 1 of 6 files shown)


manupak:
~~Put it to draft until https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1225 is merged...~~

krzysz00 (approved):
Looks reasonable to me, approved

manupak:
http://rocmhead:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-1234/3/pipeline/145 Navi3x passes test but hangs later in some tuning stage.. Im going to disable navi3x for the next one.

PR #1236, opened 2023-09-12
Add tuning space argument to tuning driver, tuning runner script

krzysz00:
This allows us to manually run quick tunings or other such exercises, which'll be useful for producing internal perf comparisons that better track what MIGraphX wants.

Fixes
https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1126


ravil-mobile (approved):
I am not a big expert in this part of the code by it looks good to me

PR #1237, opened 2023-09-13
Relax f16 threshold to cater for wmma

manupak:
Navi3x may need a relaxed threshold to cater for f16 discrepancies.


manupak:
FYI,
I ll just note that the attention kernel can amplify f16 inaccuracies due to exponents after the first gemm -- hence it is a bit of a open problem but nothing concerning because from a ML point of view what matters is the histogram shape rather than each individual value.

krzysz00 (approved):

PR #1240, opened 2023-09-13
Disable tuning fallback, as it interferes with tuning

krzysz00:
The tuning fallback option was added to the MIGraphX integration in

I can't find any rationale for why we added this, but I think it might have had to do with concerns that stale tuning data would cause model-wide failures.

Given that this is throwing off our ability to tune the model and to actually know what we're tuning, I'm disabling this fallback behavior in this commit. From now on, passing a value of `perf_config` that doesn't correspond to an applicable config will cause compilation failure, just like in MLIR's native flow.


manupak:
Hmmm so there can be three scenarios :
T1) no perf config is provided
T2) in-applicable perf config is provided
T3) applicable perf config is provided

I think it should still pass with T1 and T3. I think this change as it is will also make T1 fail as well.

krzysz00:
No, it won't make T1 fail

manupak:
maybe it works when perf_config std::string is empty

krzysz00:
(We exercise T1 all the time with rocmlir-gen)

manupak:
Cool! if thats the case I have no issues here.

manupak (approved):
LGTM!

PR #1242, opened 2023-09-14
Add support for explicit broadcast in MIGraphX dialect

manupak:
This commit adds tosa.add with splat zero for explicit broadcasts. Then this pattern is identifed in RockToTosa conversion and replaced with a rock.transform broadcast.

The test exposed a further bug in the collapseContigousMerges where there are broadcasted dimensions that are padded and merged.

Moreover, due to unavailability of infra to test user-defined partitions based clone, I had to change the test to look similiar after partition pipeline has been run.

closes : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1128

### DIFF
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosaPass.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosaPass.cpp
@@ -14,6 +14,7 @@
 #include "mlir/Conversion/MIGraphXToTosa/MIGraphXToTosa.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/MHAL/IR/MHAL.h"
 #include "mlir/Dialect/Quant/QuantOps.h"
 #include "mlir/Dialect/Tosa/Transforms/Passes.h"
 #include "mlir/Dialect/Tosa/Utils/QuantUtils.h"
@@ -44,7 +45,8 @@ public:
     auto &ctx = getContext();
     RewritePatternSet patterns(&ctx);
     ConversionTarget target(ctx);
-    target.addLegalDialect<tosa::TosaDialect, func::FuncDialect>();
+    target.addLegalDialect<tosa::TosaDialect, func::FuncDialect,
+                           mhal::MHALDialect>();
     target.addIllegalDialect<migraphx::MIGraphXDialect>();
     target.markUnknownOpDynamicallyLegal([](Operation *) { return true; });
 
--- a/mlir/lib/Dialect/MIGraphX/MiGraphXTransform.cpp
+++ b/mlir/lib/Dialect/MIGraphX/MiGraphXTransform.cpp
@@ -14,6 +14,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/MHAL/IR/MHAL.h"
 #include "mlir/Dialect/MIGraphX/MIGraphXOps.h"
 #include "mlir/Dialect/MIGraphX/Passes.h"
 #include "mlir/Dialect/Tosa/IR/TosaOps.h"
@@ -63,7 +64,7 @@ struct MIGraphXTransforms
     RewritePatternSet patterns(&ctx);
     ConversionTarget target(ctx);
     target.addLegalDialect<migraphx::MIGraphXDialect, func::FuncDialect,
-                           tosa::TosaDialect>();
+                           tosa::TosaDialect, mhal::MHALDialect>();
     target.addIllegalOp<migraphx::SqrtOp>();
     auto func = getOperation();
 
--- a/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
+++ b/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
@@ -249,6 +249,7 @@ struct DimInfo {
   TransformAttr transform;       // Transform to which the dimension belongs
   size_t positionInMerge; // Position of the dimension inside the transform
   bool isBroadcast;       // Is this a broadcast dimension?
+  bool isPadded;          // Is this a padded dimension?
 
   // Utility function to return a transform pair <map,transform>
   std::pair<TransformMapAttr, TransformAttr> transformPair() {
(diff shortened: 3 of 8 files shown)


mlir/test/fusion/pr-e2e/mixr-sd-explicit-broadcasting.mlir
```diff
@@ -0,0 +1,21 @@
+// RUN: rocmlir-driver -kernel-pipeline=migraphx %s | rocmlir-driver -host-pipeline=partition,highlevel -targets %arch | rocmlir-gen -ph -rand 1 -rand_type float -fut mlir_reshape_convolution --verifier clone - | rocmlir-driver -c -arch %arch | mlir-cpu-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_float16_utils%shlibext,%linalg_test_lib_dir/libmlir_c_runner_utils%shlibext --entry-point-result=void | FileCheck %s
+// ALLOW_RETRIES: 2
+// CHECK: [1 1 1]
+// Note the fake wrapper function to make it look like this has already been partitioned,
+// since we want "partition" for the xmodel packaging but don't actually want to use
+// the partition logic.
+// This is an awkward hack that should be fixed Later (tm)
+module {
+  func.func @mlir_reshape_convolution_real(%arg0: tensor<1x1x16x1x16x1xf32>, %arg1: tensor<1x1x3x3xf32>) -> tensor<1x1x32x32xf32> attributes {kernel = "mixr", num_cu = 48 : i64} {
```
manupak: This test isn;t doing anything meaningful as I have found out.
So are the e2e  tests in https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1222

krzysz00: Really? Can we sync up on this - I looked at the intermediates and saw an actual series of calls.

manupak: Ive put a reproducer on your past PR. 
Would you do me a favor and double check the output after rocmlir-driver -c just before cpu-runner.

I dont see any kernels there.

manupak: i.e. no long bitstream anywhere to be found.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosaPass.cpp
```diff
@@ -44,7 +45,8 @@ struct MIGraphXToTosa : public impl::MIGraphXToTosaPassBase<MIGraphXToTosa> {
     auto &ctx = getContext();
     RewritePatternSet patterns(&ctx);
     ConversionTarget target(ctx);
-    target.addLegalDialect<tosa::TosaDialect, func::FuncDialect>();
+    target.addLegalDialect<tosa::TosaDialect, func::FuncDialect,
```
krzysz00, line 48: Why this?

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -249,6 +249,7 @@ struct DimInfo {
   TransformAttr transform;       // Transform to which the dimension belongs
   size_t positionInMerge; // Position of the dimension inside the transform
   bool isBroadcast;       // Is this a broadcast dimension?
+  bool isPadded;          // Is this a padded dimension?
```
krzysz00, line 252: This is to do with broadcast and padding?

krzysz00 (commented):
I think the old implicit broadcast logic is still important when we can use it, can you put it back?

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosaPass.cpp
manupak: to run migraphx pipeline on the host for references for the clone test.

manupak:
> I think the old implicit broadcast logic is still important

Why do you think it is important ?
Its not important for the kernels -- because every broadcast (implicit or explicit) is going to be converted to a rock.transform in tosa to rock.

krzysz00:
Will it? I think the implicit broadcasts will convert to `linalg.generic` input maps which'll then get regularized away, and that'll help make the generic fusion work better?

Unless I'm way off

manupak:
it would convert to rock.transform before even linalg is introduced.

manupak:
all will be rock.transforms before even regularization happens... -- im going to claim regularization stays unaffected.

It is just that we just pre-emptively convert the explicit broadcast to a rock.transform prior to tosa-to-linalg happens -- so it wont need 'raising' from input maps of linalg generic.

krzysz00:
I could've sworn that `tosa-to-rock` doesn't go through, say, `tosa.add` and make the broadcasts explicit? But maybe I missed something?

manupak:
> I could've sworn that tosa-to-rock doesn't go through, say, tosa.add and make the broadcasts explicit? 

This might be true

But I thought we are discussing about `assignExpandedShapeVal` which is solely used by migraphx.broadcast and migraphx.multibroadcast lowering, which are explicit originally.

krzysz00:
Yes, I'm saying `migraphx.[multi]broadcast` should lower to Tosa implicit broadcasts if possible.

manupak:
> Yes, I'm saying migraphx.[multi]broadcast should lower to Tosa implicit broadcasts if possible.

migraphx.[multi]broadcast is an explicit broadcast operator and we have been using Tosa implicit broadcasts just because Tosa did not have an explicit broadcast.

Why would you want to maintain two flavors of lowering ? (and increase error surface unneccesarily)
Explicit broadcast operator should just lower to rock.transform passthrough + broadcasts -- we are just using an dummy add as a vehicle to do that. Why does that have to suffer through linalg.generic''s input affine map to be recovered later as a broadcast ? when the input IR (migraphx dialect) clearly specifies an explicit broadcast.

I remain unconvinced about 'old implicit broadcast logic is still important' that we need to lower an explicit broadcast operator to implicit broadcast.

krzysz00:
Yeah, I see your point, we might as well make it uniform, especially since we have our own fusion logic that can handle the interruption in the `linalg.generic`s caused by this approach

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosaPass.cpp
krzysz00: Yeah, but you only need to mark a dialect "legal" if you emit it during the pass

krzysz00 (approved):
Approved, minus that nit about `mhal`

mlir/lib/Dialect/MIGraphX/MiGraphXTransform.cpp
```diff
@@ -63,7 +64,7 @@ struct MIGraphXTransforms
     RewritePatternSet patterns(&ctx);
     ConversionTarget target(ctx);
     target.addLegalDialect<migraphx::MIGraphXDialect, func::FuncDialect,
-                           tosa::TosaDialect>();
+                           tosa::TosaDialect, mhal::MHALDialect>();
```
manupak, line 67: This does a full conversion as opposed to a partial conversion (see few lines below). 
https://mlir.llvm.org/docs/DialectConversion/

I think what you say is true only for a partial conversion.

krzysz00: Ah, then yeah, that's everything.

manupak:
Thanks @krzysz00 !

PR #1244, opened 2023-09-15
Add utility script for pulling modules from a MIGraphX trace.

krzysz00:


manupak (approved):
LGTM! 
Disclaimer: I have no clue about perl, so going to trust this is doing what is supposed to do.
(If you want a proper review, might want to ping someone who knows perl)

Issue #1245, opened 2023-09-18
Some observations from gfx1102 runs


jungpark-mlir:
Hi,

I've got a RX7600 card which has gfx1102 with wmma support and here's a couple of issues I found.

I tested a 1k*1k*1k gemm with this,
./bin/rocmlir-gen -ph -operation gemm -t f16 -out_datatype f16 --arch gfx1102 --num_cu 32 -g 1 -m 1024 -k 1024 -n 1024 -transA=True -transB=False

Initially picked tuning params is
```
// -----// IR Dump After RockAffixTuningParametersPass (rock-affix-params) //----- //
func.func @rock_gemm(%arg0: memref<1x1024x1024xf16>, %arg1: memref<1x1024x1024xf16>, %arg2: memref<1x1024x1024xf16>) attributes {block_size = 128 : i32, grid_size = 64 : i32, kernel, mhal.arch = "amdgcn-amd-amdhsa:gfx1102", wave_size = 32 : i32} {
  rock.gemm %arg2 = tr %arg0 * %arg1 features =  dot|atomic_add|atomic_fmax_f32|wmma storeMethod =  set {arch = "amdgcn-amd-amdhsa:gfx1102", derivedBlockSize = 128 : i32, gridSize = 64 : i32, numCU = 32 : i32, params = #rock.wmma_gemm_params<kpackPerBlock = 4, mPerBlock = 128, nPerBlock = 128, kpack = 16, mPerWave = 64, nPerWave = 64, forceUnroll = true>} : memref<1x1024x1024xf16> = memref<1x1024x1024xf16> * memref<1x1024x1024xf16>
  return
}
```
Which is  --perf_config='128, 128, `4`, 64, 64, 16, 1, `1`' 
This runs and verifies with cpu result (when enabled). 
But it looks mismatching with validRangeWmmaGemmParams in tuning, which has `{16, 32}` for kPerBlock and sets `false` to the last param.

The other issue is, it causes gpu hang for lots of tuning params, for example,
```
 --perf_config='16, 32, 16, 16, 32, 16, 1, 0'
 --perf_config='16, 64, 16, 16, 64, 16, 1, 1'
 --perf_config='32, 16, 32, 16, 16, 16, 1, 0'
 --perf_config='32, 32, 32, 16, 32, 16, 1, 0'
 --perf_config='64, 16, 16, 32, 16, 16, 1, 0'
```

[...]

PR #1246, opened 2023-09-19
Using less spilling values for kOuter

giuseros:
This is trying to address #1245 . Immediate changes
- I put the last config parameter back to `true` 
- I removed all the `kouter={16/32}` configurations from both the tuning and the default list. They spill in most of the cases, and if they don't spill they are outperformed by configuration with smaller `kOuter`. 
- I added the possibility to have `kPack < kBase(16)` .While for xdlops we had kbase < maxVectorLength, this is not the case for wmma. Indeed, kpack=16 would vectorize properly only for int8. 

Long term plan: find a new set of "default" configurations


giuseros:
@krzysz00 maybe you could try to see if this fixes the hangs you are experiencing

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.h
```diff
@@ -110,6 +110,9 @@ struct AccelEmitter {
                                        Location loc, Value baseOffset,
                                        Value dWaves, Value laneId) = 0;
 
+  /// Validate the accelerator structure
+  virtual void validateAcceleratorProperties(){};
```
krzysz00: `virtual LogicalResult` and then crashing on failure or something.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -710,50 +710,18 @@ const InitParamsAccel
...
-  {32, 64, 16, 32, 32, 16, true, true},
-  {32, 64, 16, 16, 64, 16, true, true},
-  {32, 64, 16, 16, 32, 16, true, true},
   {32, 64, 4, 32, 64, 16, true, true},
-  {32, 32, 32, 16, 32, 16, true, true},
-  {32, 32, 16, 32, 32, 16, true, true},
-  {32, 32, 16, 16, 16, 16, true, true},
-  {32, 32, 16, 16, 32, 16, true, true},
-  {16, 64, 16, 16, 64, 16, true, true},
-  {16, 64, 16, 16, 32, 16, true, true},
```
krzysz00, line 727: Let's put these back as `kPerBlock = 4` or, more generally, re-run.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -57,10 +57,10 @@ void createGemmTuningRangeBF(TuningParamSet *newSpace,
   const std::vector<std::vector<uint32_t>> validRangeWmmaGemmParams = {
       {4, 8, 16, 32, 64, 128, 256},
       {16, 32, 64, 128, 256},
-      {16, 32},
+      {1, 2, 4, 8},
       {4, 8, 16, 32, 64, 128},
       {4, 8, 16, 32, 64, 128},
-      {16},
+      {4, 8, 16},
```
krzysz00, line 63: Will we want kPack < 16?

(I think so but)

giuseros: Yes, for two reasons:
- For fp16 the maximum hardware vector is 8 (128bits)
- In general we saw that there are configurations where a vector smaller than the max allowed  is faster

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
giuseros: Coming up in a follow-up PR

krzysz00 (approved):

giuseros:
I stopped at analyzing ~240 gemm configs, but the results look good enough: 
- "Quick tuning" time reduced by a lot on navi32 (I didn't measure, but it was more than a 2x reduction)
- Performance improved by a lot
- There are 4 cases where performance decreased, but only 1 is real (and it is about 10% hit on a very small int8 gemm)

I am attaching the report here
[navi32.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/12699829/navi32.zip)

krzysz00 (approved):

giuseros:
Given the performance improvements and the simplicity of the change, I am merging this in. If worse comes to worst, we can always revert this change using the old configs.

PR #1249, opened 2023-09-21
Upload models on github repository and use wget to downloaded them.

stefankoncarevic:


krzysz00 (commented):
Per yesterday's MIGraphX meeting, we need to consult with other MLSE teams on how they're handling this

krzysz00 (approved):
Approved with the understanding that we'll find a way to clean this up next release

PR #1255, opened 2023-09-25
Bump rocMLIR supported major version up for ROCm 6.0

jerryyin:
In this release, we have switched our frontend from MIOpen to MIGraphX. I believe this justifies a major version bump:
 - 1.x.x releases we claim support for MIOpen
 - 2.x.x releases we deprecated MIOpen support and transition to MIGraphX


krzysz00 (approved):
Sure, why not!

PR #1256, opened 2023-09-25
Embed device libraries into build to minimize runtime requirements

krzysz00:
Fixes
https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1142

Patch SerializeToHsaco (we'll need to do a version of this again when we move to the new compilation infrastructure) and its cmake to find the device libraries at bulid time (if possible) and embed them (by way of a convenient little Python script) into the compilation of SerilazieToHsaco. This'll allow our stuff to work even when there's no /opt/rocm or we can't find it.

( @lamb-j since you're not an available reviewer but should see this )

### DIFF
--- a/external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp
+++ b/external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp
@@ -159,6 +159,10 @@ void SerializeToHsacoPass::getDependentDialects(
   gpu::SerializeToBlobPass::getDependentDialects(registry);
 }
 
+#ifdef ROCMLIR_DEVICE_LIBS_PACKAGED
+#include "Transforms/AmdDeviceLibs.cpp.inc"
+#endif
+
 std::optional<SmallVector<std::unique_ptr<llvm::Module>, 3>>
 SerializeToHsacoPass::loadLibraries(SmallVectorImpl<char> &path,
                                     SmallVectorImpl<StringRef> &libraries,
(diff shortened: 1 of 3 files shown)


external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp
```diff
@@ -157,26 +157,48 @@ void SerializeToHsacoPass::getDependentDialects(
...
     llvm::SMDiagnostic error;
-    llvm::sys::path::append(path, file);
-    llvm::StringRef pathRef(path.data(), path.size());
-    std::unique_ptr<llvm::Module> library =
-        llvm::getLazyIRFileModule(pathRef, error, context);
-    path.truncate(dirLength);
+    std::unique_ptr<llvm::Module> library;
+#ifdef ROCMLIR_DEVICE_LIBS_PACKAGED
+    if (packagedLibs.contains(file)) {
+      llvm::errs() << "Pulling in " << file << "\n";
```
giuseros: Would it be better something different than `llvm::errs`? Maybe `emitRemark`?

giuseros (commented):
I think this is very cool! Just a question: is there a way to test this?

external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp
krzysz00: ... Oh, goodness, stray debug message, thanks

giuseros:
Hi @krzysz00 I am happy to approve this,  I want just to understand how this will be tested. 

Is MIGraphX using this flow so that we know it works? Alternatively, can we add a build step in our tests to verify that it at least builds when `/opt/rocm` is not there?

krzysz00:
@giuseros Just like our `lld` embedding, there won't be any specific test to ensure that this is working - the verification is by manual `strace` more than anything, and I can't think how to easily make it not do that

giuseros (approved):

PR #1257, opened 2023-09-26
[mlir] Fix file leak in SeralizeToHsaco

krzysz00:
SerializetToHsaco, as currently implemented, leaks the file descriptor of the .hsaco temporary file, which causes issues in long-running parallel compilation setups.

In addition, move to an LLVM mutex for the linking stage rather than std::muterx to match upstream better.

Fixes
https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1151

### DIFF
--- a/external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp
+++ b/external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp
@@ -14,6 +14,8 @@
 #include "mlir/Dialect/GPU/Transforms/Passes.h"
 #include "mlir/IR/Location.h"
 #include "mlir/IR/MLIRContext.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/Mutex.h"
 
 #if MLIR_GPU_TO_HSACO_PASS_ENABLE
 #include "mlir/ExecutionEngine/OptUtils.h"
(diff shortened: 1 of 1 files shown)


jerryyin (approved):
I think it is useful to know:
1. How do you figure out it is hsaco file descriptor that's not closed? Through strace?
2. How do you think we can prevent this from happening again? Can we run a single lowering with strace and make a text match with keyword like "*leak*"

krzysz00:
@jerryyin 
As to 1., here's how the debug process went
```bash
```
strace -ff -e trace=%file,%desc -o calls env MIGRAPHX_ENABLE_MLIR=1 ./bin/test_verify test_lstm_reverse_last
cat calls.* > calls
rm calls.*
less calls
```
then skipping to the end (which was an MLIR process or had one close-by) to look at the patterns of `openat`s and `read`s and `close`s.
2. There isn't really an automated way to regression test this that I'm aware of, especially since we're running a bunch of stuff in the same process so it's hard to tell the difference between leaking something and just not having gotten around to closing it yet since you still need it.

File: external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp [hunk 1]
```diff
@@ -448,13 +452,14 @@ SerializeToHsacoPass::createHsaco(const SmallVectorImpl<char> &isaBinary) {
   }
 
   // Load the HSA code object.
-  auto hsacoFile = openInputFile(tempHsacoFilename);
+  auto hsacoFile =
+      llvm::MemoryBuffer::getFile(tempHsacoFilename, /*IsText=*/false);
```
sjw36, line 456: Good, should have a runtime improvement also!

krzysz00: I don't think there'll be a runtime improvement, we copy this binary out to a vector anyway

PR #1258, opened 2023-09-26
Fix attention padding kernels

manupak:
Padded attention kernels wont just
work out of the box as (implicit) gemm
kernels.

This is because attention kernels does a
softmax normalization after the first gemm
and having additional zeros -- zero not being
the least value in the range -- is going to affect the normalization.

Therefore, this commit introduces a mid-kernel
overwrite to first gemm output to make the out
of bound values to be negative infinity.

closes : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1146

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -427,6 +427,8 @@ def Rock_GridwiseAttentionAccelOp :
                    Rock_GemmFeaturesAttr:$features,
                    I32Attr:$blockSize,
                    I32Attr:$gridSize,
+                   OptionalAttr<I32Attr>:$prePadG0M,
+                   OptionalAttr<I32Attr>:$prePadG0N,
                    RockAccelTuningParamAttrInterface:$params)> {
   let summary = "Gridwise attention accelerated version";
   let description = [{
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -427,8 +427,8 @@ def Rock_GridwiseAttentionAccelOp :
                    Rock_GemmFeaturesAttr:$features,
                    I32Attr:$blockSize,
                    I32Attr:$gridSize,
-                   OptionalAttr<I32Attr>:$prePadG0M,
-                   OptionalAttr<I32Attr>:$prePadG0N,
+                   OptionalAttr<IndexAttr>:$prePadG0M,
+                   OptionalAttr<IndexAttr>:$prePadG0N,
                    RockAccelTuningParamAttrInterface:$params)> {
   let summary = "Gridwise attention accelerated version";
   let description = [{
--- a/mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
@@ -257,12 +257,19 @@ AttentionRewritePattern::matchAndRewrite(AttentionOp op,
   func::FuncOp func = op->getParentOfType<func::FuncOp>();
   IntegerAttr blockSizeAttr = func->getAttr("block_size").cast<IntegerAttr>();
   IntegerAttr gridSizeAttr = func->getAttr("grid_size").cast<IntegerAttr>();
-
+  IntegerAttr prePadG0MAttr;
+  if (gemm0ExtraPad.m) {
+    prePadG0MAttr = rw.getI32IntegerAttr(gemm0Size.m);
+  }
+  IntegerAttr prePadG0NAttr;
+  if (gemm0ExtraPad.n) {
+    prePadG0NAttr = rw.getI32IntegerAttr(gemm0Size.n);
+  }
   rw.replaceOpWithNewOp<GridwiseAttentionAccelOp>(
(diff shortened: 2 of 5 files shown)


mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
```diff
@@ -257,12 +257,19 @@ AttentionRewritePattern::matchAndRewrite(AttentionOp op,
   func::FuncOp func = op->getParentOfType<func::FuncOp>();
   IntegerAttr blockSizeAttr = func->getAttr("block_size").cast<IntegerAttr>();
   IntegerAttr gridSizeAttr = func->getAttr("grid_size").cast<IntegerAttr>();
-
+  IntegerAttr prePadG0MAttr;
```
krzysz00, line 260: `index` please, or else this'll break on large tensors

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1237,6 +1237,67 @@ struct GridwiseAttentionAccelRewritePattern
...
+        /*useIndexDiffs=*/true, /*forceUnroll=*/true);
+    {
+      OpBuilder::InsertionGuard guard(rewriter);
+      rewriter.setInsertionPointToStart(loop.getBody());
+
+      Block::BlockArgListType gemm0OutCoords = loop.getLowerCoords(0);
+      Block::BlockArgListType upperCoords = loop.getLowerCoords(1);
+      auto isLargerThanM = rewriter.create<arith::CmpIOp>(
+          loc, arith::CmpIPredicate::sge, gemm0OutCoords[1], prePadGemmMVal);
+      auto isLargerThanN = rewriter.create<arith::CmpIOp>(
```
krzysz00: You shouldn't need this? You could just query the `validity` argument of the transforming_for, which'll return `false` in the event of padding?

Heck, you could extend `global_load` with a `pad_value : T`, which you'd use in the case of invalidity (though then you'd want to not use buffer ops, since those only support 0 and sometimes 1 as padding values)

krzysz00 (commented):
Something's slightly bothering me about the approach here, but I might not understand the problem.

manupak:
Good point, I could use validity on output but not on global_load.

I was initially musing over specifying a pad value for loads but the
trouble is i need out of bounds output to be neg inf not out of bound
input. Also I cant think of pad value on input that would always result in
neg inf for out of bound output.

So I ll partially take ur suggestion to use validity

manupak:
Ok maybe not, because there is no pad tr maps in the first gemm output
which is not fully realized global buffer. So this has to be it.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
manupak: As I described below this has nothing to do with global loads. (as in I cant get the first gemm output contain neg infs in padded region by assigning a 'special' value to inputs)

Since its a intermediary tensor that is supposedly padded, there are no tr maps to indicate that as we dont write out that tensor to global.

krzysz00:
@manupak I'm having trouble understanding what you're trying to get across.

To clarify, the goal here is "if a register doesn't correspond to a global coordinate that would have been valid if this first gemm was real, replace it with -inf"?

How feasible is it to construct the maps for "writing out" (even if you never actually do it) the first gemm?

manupak:
> To clarify, the goal here is "if a register doesn't correspond to a global coordinate that would have been valid if this first gemm was real, replace it with -inf"?

Yep this is right

> How feasible is it to construct the maps for "writing out" (even if you never actually do it) the first gemm?

This is feasible. We can construct & append a padding rock.transform -- so that validity checks will stop composing affine maps until the last one and provide a validity output. I dont see that change having a meaningful change on the generated code -- a.k.a post rock-sugar-to-loops. 

If you want, I can create and append such transform map and remove the body of the current transforming for -- to be honest I dont see a difference as here as these get unrolled.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -427,6 +427,8 @@ def Rock_GridwiseAttentionAccelOp :
                    Rock_GemmFeaturesAttr:$features,
                    I32Attr:$blockSize,
                    I32Attr:$gridSize,
+                   OptionalAttr<I32Attr>:$prePadG0M,
```
krzysz00: Again, `index`

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: To rephrase, could you *create* the maps that would have been used if the intermediate tensor *were* written to global memory, and then use the validity bit those generate to do this `-inf` change?

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
manupak: done

manupak (commented):
@krzysz00 I think I ve done what was requested.

krzysz00 (approved):
Looks pretty nice

PR #1262, opened 2023-09-27
Enable trivial broadcasts of splat constants

manupak:
This commit just rewrites the splat constant
broadcasts with another splat constant.

It also workaround the issue of non-scalar
constants appearing in the kernel code which
does not seem to be properly handled.

closes : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1152


krzysz00 (approved):

krzysz00:
This is a reasonable optimization, and that it just so happens to fix a bug is useful

manupak:
Thanks @krzysz00!

PR #1263, opened 2023-09-28
Update Jenkins to rocm 5.7 and add the DGPU_TARGET for MIGraphX stage.

stefankoncarevic:


giuseros (approved):
LGTM

jerryyin:
@stefankoncarevic Did you check locally that the build works?

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -932,7 +932,13 @@ pipeline {
                         steps {
                             sh 'rm -rf MIGraphX'
                             dir('MIGraphX') {
-                                getAndBuildMIGraphX("-DCMAKE_PREFIX_PATH='${WORKSPACE}/MIGraphXDeps;/MIGraphXDeps;/opt/rocm' -DMIGRAPHX_ENABLE_MLIR=On -DCMAKE_CXX_COMPILER=/opt/rocm/llvm/bin/clang++ -DMIGRAPHX_USE_HIPRTC=Off")
+                                getAndBuildMIGraphX('''
+                                                    -DCMAKE_PREFIX_PATH='${WORKSPACE}/MIGraphXDeps;/MIGraphXDeps;/opt/rocm'
+                                                    -DMIGRAPHX_ENABLE_MLIR=On
+                                                    -DCMAKE_CXX_COMPILER=/opt/rocm/llvm/bin/clang++
+                                                    -DMIGRAPHX_USE_HIPRTC=Off
```
krzysz00, line 939: Would it be possible to make this `-DGPU_TARGETS=${whatever-the-current-arch-is}`?

stefankoncarevic: I think it can, I will try to add that the build MIGraphX only for the current GPU architecture.

yiqian1:
I believe the "method too large" error is introduced by the global variable. You can try to remove that variable, instead call get_gpu_architecture() from the build-MIGraphX stage and use a local variable to store the result.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -930,12 +946,21 @@ pipeline {
...
+                            script {
+                                def gpu_arch = get_gpu_architecture()
+                                sh 'rm -rf MIGraphX'
+                                dir('MIGraphX') {
+                                    getAndBuildMIGraphX('''
+                                                    -DCMAKE_PREFIX_PATH='${WORKSPACE}/MIGraphXDeps;/MIGraphXDeps;/opt/rocm'
+                                                    -DMIGRAPHX_ENABLE_MLIR=On
+                                                    -DCMAKE_CXX_COMPILER=/opt/rocm/llvm/bin/clang++
+                                                    -DMIGRAPHX_USE_HIPRTC=Off
+                                                    -DGPU_TARGETS=${gpu_arch}
```
yiqian1: missing "" here?

jerryyin:
@stefankoncarevic Please file successive ticket on the feedback from @krzysz00. If the problem is -DGPU_TARGETS not reading the string correctly, then it must be related with string handling in Jenkins file. Please talk with @yiqian1 to figure out the right way to do this.

I'm merging this considering:
 - You are on vacation till after the release
 - CI has full pass and we indeed want to land this for the release

PR #1264, opened 2023-09-28
Enalbe backward-data convolution in performance tests

yiqian1:


mlir/utils/performance/perfRunner.py
```diff
@@ -29,7 +29,7 @@
 BENCHMARKING_RESULT_FILE_NAME = 'results.stats.csv'
 #DIRECTIONS = ['-F 1', '-F 2', '-F 4']
 # temporarily disable backward-data until rocmlir-tuning-driver can handle multi-kernel code
```
krzysz00: Delete the comment, otherwise approved

krzysz00 (approved):

PR #1266, opened 2023-09-28
[ROCM 6.1] Refactor threadwise accel gemm

giuseros:
This PR is trying to make `threadwise_accel_gemm` more aligned to the other threadwise operator. I made it support a `m,n,k` view + external indices to "freeze" the iteration. 

The main reason I am doing this is that I want to have the possibility to play with the loop orders. 

If I want to move the `threadwise_accel_gemm` at a different nesting level, now I can simply pass a different view, while before it was not possible ([in the past](https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/976/files#diff-65ff09ada7be250263e6c3716d4c47e8989b32c13f8b755388b4d028aabfe2ef) we had to change the operator interface to implement a different loop schedule)


giuseros:
I am closing this PR, because I think that I found better ways of doing it and I will open a new PR

PR #1267, opened 2023-09-29
Fix windows regression caused by 3cff247

tperry-amd:
# Root Cause

This was caused by commit caused by 3cff247d0f8b35b8c8e356b3a02e9a74bea1674f "[mlir] Fix file leak in SeralizeToHsaco (#1257)"

This change caused the following error:
rocMLIR\external\llvm-project\mlir\lib\Dialect\GPU\Transforms\SerializeToHsaco.cpp(439): error C2664: 'std::error_code llvm::sys::fs::closeFile(llvm::sys::fs::file_t &)': cannot convert argument 1 from 'int' to 'llvm::sys::fs::file_t &'

# Fix

LLVM on Windows uses void* for file_t instead of int. CloseFile accepts this as a parameter and eventually uses it as an int anyway.

The least impactful fix is to initialize a file_t and set the lower int values to the int for the file descriptor. A direct reinterpret_cast from int to void* generates a warning about type sizes since int is 32-bit on Windows.

### DIFF
--- a/external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp
+++ b/external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp
@@ -436,7 +436,22 @@ SerializeToHsacoPass::createHsaco(const SmallVectorImpl<char> &isaBinary) {
   }
   llvm::FileRemover cleanupHsaco(tempHsacoFilename);
   // Close here to prevent weird behaviors when LLD opens this file.
+  // file_t is an int on linux and void* on Windows so an explicit cast is
+  // needed. This casting may go from a 32-bit type to a 64-bit type on
+  // Windows which generates a warning so the approach is to just set the low
+  // bytes to the int value.
+  // Internal to closeFile it just uses it as an int again so it should
+  // function.
+  // A proper fix would be to change createTemporaryFile to close the file
+  // after it opens it or to change closeFile to accept an int. Both changes
+  // require changing core LLVM code so this workaround was done.
+#if _WIN32
+  llvm::sys::fs::file_t tempHscacoFilePtr = 0;
+  *reinterpret_cast<int*>(&tempHscacoFilePtr) = tempHsacoFD;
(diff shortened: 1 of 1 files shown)


external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp
```diff
@@ -436,7 +436,22 @@ SerializeToHsacoPass::createHsaco(const SmallVectorImpl<char> &isaBinary) {
   }
   llvm::FileRemover cleanupHsaco(tempHsacoFilename);
   // Close here to prevent weird behaviors when LLD opens this file.
+  // file_t is an int on linux and void* on Windows so an explicit cast is
+  // needed. This casting may go from a 32-bit type to a 64-bit type on
+  // Windows which generates a warning so the approach is to just set the low
+  // bytes to the int value.
+  // Internal to closeFile it just uses it as an int again so it should
+  // function.
+  // A proper fix would be to change createTemporaryFile to close the file
```
giuseros, line 445: I guess another possible fix could be to have create temporary file to accept a `file_t&` type

giuseros (approved):

giuseros:
Hi @tperry-amd , 
this is failing because of formatting. You should run `clang-format external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp`

krzysz00 (commented):
The actual fix is being worked out at https://github.com/llvm/llvm-project/pull/67711 - I'll make those changes and backport the actual solution over instead.

krzysz00:
(It feels like the fix may be to use the version of `createTemporaryFile` that doesn't take an FD argument - could you try that? - I just pushed it to my upstream PR)

tperry-amd:
> (It feels like the fix may be to use the version of `createTemporaryFile` that doesn't take an FD argument - could you try that? - I just pushed it to my upstream PR)

I saw your recent comments on the upstream PR. I'll check your changes from the upstream PR shortly and close this PR since it sounds like you will backport it.

PR #1269, opened 2023-09-29
Mirror upstream fix to SerializeToHsaco Windows regression

krzysz00:
See
https://github.com/llvm/llvm-project/pull/67711


jerryyin (approved):

krzysz00:
Bad news - this workaround is really bad for (heuristic mode) fp16 performance, see attached.
It may be the case that we'll want a more targeted compiler patch here, and there're some theories of what that might look like floating around.
[s-clause-workaround-pr1270.html.txt](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/12784342/s-clause-workaround-pr1270.html.txt)

krzysz00:
... wrong PR

tperry-amd (approved):

PR #1270, opened 2023-10-02
Work around Navi3x compiler issue causing hangs

krzysz00:
This is a rather blunt instrument, but it does resolve the problem we were seeing with hangs caused by the compiler insterting s_clause instructions somehow.

However, I can say this fixes
https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1161

Performance numbers upcoming

### DIFF
--- a/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
+++ b/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
@@ -176,7 +176,7 @@ struct LowerGpuOpsToROCDLOpsPass
 
     // Manually rewrite known block size attributes so the LLVMIR translation
     // infrastructure can pick them up.
-    m.walk([ctx](LLVM::LLVMFuncOp op) {
+    m.walk([ctx, maybeChipset](LLVM::LLVMFuncOp op) {
       if (auto blockSizes = dyn_cast_or_null<DenseI32ArrayAttr>(
               op->removeAttr(gpu::GPUFuncOp::getKnownBlockSizeAttrName()))) {
         op->setAttr(ROCDL::ROCDLDialect::getReqdWorkGroupSizeAttrName(),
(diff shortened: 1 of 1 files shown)


giuseros (commented):
LGTM

giuseros (approved):

jerryyin (approved):
Mind to file a follow-up tracking ticket and reference this PR and its context so we can remember to retry removing when the proper fix is up in the backend.

krzysz00:
Bad news - this workaround is really bad for (heuristic mode) fp16 performance, see attached.
It may be the case that we'll want a more targeted compiler patch here, and there're some theories of what that might look like floating around.
[s-clause-workaround-pr1270.html.txt](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/12784342/s-clause-workaround-pr1270.html.txt)

krzysz00:
Closing this PR in favor of a more targeted change to our copy of the compiler that should have much lower performance implications

giuseros:
> Bad news - this workaround is really bad for (heuristic mode) fp16 performance, see attached. It may be the case that we'll want a more targeted compiler patch here, and there're some theories of what that might look like floating around. [s-clause-workaround-pr1270.html.txt](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/12784342/s-clause-workaround-pr1270.html.txt)

Did you try to re-tune? Because I don't think we can re-use the same configurations

krzysz00:
Given that #1271 exists, I didn't bother re-tuning, and also the perf drops were on a very wide variety of kernels, so I'd be surprised if re-tuning fixed anything

giuseros:
So about retuning, to quickly see if it made any difference you can take the worst case in your perf report, re-tune the single problem, and see if you get any benefits. 

I will comment on #1271 for the alternative solution

krzysz00:
Quick tuning gave the same perf config in both cases

It's possible that (going off the timings in said quick tuning) the fp16 drops were a particularly nasty round of benchmark failures, but still, given that the other PR seems to target the actual issue better, we probably don't want to use this one

PR #1273, opened 2023-10-03
Generalize and rename `threadwise_transpose`

giuseros:
Fixes https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1160

I generalized `threadwise_transpose` to accept any source/dest view and renamed it `threadwise_copy`

Thanks @manupak  for the guidance!

### DIFF
--- a/mlir/lib/Dialect/Rock/IR/RockDialect.cpp
+++ b/mlir/lib/Dialect/Rock/IR/RockDialect.cpp
@@ -1474,6 +1474,15 @@ LogicalResult ThreadwiseWriteAllOp::verify() {
   return success();
 }
 
+LogicalResult ThreadwiseCopyOp::verify() {
+  auto srcShape = getSource().getType().getShape();
+  auto dstShape = getDest().getType().getShape();
+  if (dstShape != srcShape)
+    return emitOpError("Source and dest shape need to have the same shape.");
+
+  return success();
+}
+
 //===----------------------------------------------------------------------===//
 // BlockwiseGemmOp
 //===----------------------------------------------------------------------===//
--- a/mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
@@ -624,92 +624,67 @@ struct ThreadwiseWriteAllRewritePattern
                                 ConversionPatternRewriter &b) const final;
 };
 
-struct ThreadwiseTransposeRewritePattern
-    : public OpConversionPattern<ThreadwiseTransposeOp> {
-  using OpConversionPattern<ThreadwiseTransposeOp>::OpConversionPattern;
+struct ThreadwiseCopyRewritePattern
+    : public OpConversionPattern<ThreadwiseCopyOp> {
+  using OpConversionPattern<ThreadwiseCopyOp>::OpConversionPattern;
 
-  LogicalResult matchAndRewrite(ThreadwiseTransposeOp op, OpAdaptor adaptor,
+  LogicalResult matchAndRewrite(ThreadwiseCopyOp op, OpAdaptor adaptor,
                                 ConversionPatternRewriter &b) const final;
 };
 } // end anonymous namespace
 
-LogicalResult ThreadwiseTransposeRewritePattern::matchAndRewrite(
-    ThreadwiseTransposeOp op, OpAdaptor adaptor,
+LogicalResult ThreadwiseCopyRewritePattern::matchAndRewrite(
+    ThreadwiseCopyOp op, OpAdaptor adaptor,
     ConversionPatternRewriter &b) const {
   Location loc = op.getLoc();
   auto sourceView = cast<TypedValue<MemRefType>>(adaptor.getSource());
   auto destView = cast<TypedValue<MemRefType>>(adaptor.getDest());
 
-  ArrayRef<int64_t> loadShape =
-      sourceView.getType().cast<ShapedType>().getShape();
-  Type elemType = sourceView.getType().cast<MemRefType>().getElementType();
-  int64_t copyDPerThread = loadShape[1];
-  int64_t copyKPerThread = loadShape[0];
-  int64_t kpack = op.getKpack();
-
-  // We use kpackPerThread instead of kpack to cover edge cases where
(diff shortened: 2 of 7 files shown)


mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -624,92 +624,67 @@ struct ThreadwiseWriteAllRewritePattern
...
-                                            copyKPerThread, rawLoadBufferShape,
-                                            elemType);
-    vecLen = math_util::gcd(vecLen, kpackPerThread);
-    strides[0] = vecLen;
-  }
-  loadBufferView = collapseContiguousMerges(loadBufferView, rawLoadBufferShape);
-  storeBufferView =
-      collapseContiguousMerges(storeBufferView, rawStoreBufferShape);
+
+  auto storeBufferViewInverted = invertTransforms(b, loc, storeBufferView);
```
krzysz00, line 673: Note: it looks like `invertTransforms` doesn't catch inversion failures properly, and this code should be testing for the return value of this being `nullptr` (you'll want to patch `invertTransforms` to propagate those up from `invertTransformMap`)

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -624,92 +624,67 @@ struct ThreadwiseWriteAllRewritePattern
...
 
   // Run the packing loop
-  auto packLoop = b.create<TransformingForOp>(
+  SmallVector<Value, 2> start{zero};
+  SmallVector<int64_t, 2> strides{vecLen};
+  auto copyLoop = b.create<TransformingForOp>(
       loc, ArrayRef<ValueRange>{start, start},
-      ArrayRef<Attribute>{loadBufferView, storeBufferView},
-      /*bounds=*/loadShape,
+      ArrayRef<Attribute>{srcToDstView, b.getArrayAttr({})},
```
krzysz00: I think we'll want to still want to use two domains in the uninvertable case? Or we'll want to ban the uninvertable case?

krzysz00 (commented):
Some notes re ways this could fall over

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
giuseros: What I can do is to not vectorize the uinvertible case and copy from/to the original views one element at the time

krzysz00: Makes sense, sounds like a plan

I think you'll just need to fix `invertTransforms` to report the failure cleanly

manupak: I think it just returns null ... does it not ?

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -1474,6 +1474,15 @@ LogicalResult ThreadwiseWriteAllOp::verify() {
   return success();
 }
 
+LogicalResult ThreadwiseCopyOp::verify() {
+  auto srcShape = getSource().getType().getShape();
+  auto dstShape = getDest().getType().getShape();
+  if (dstShape != srcShape)
+    return emitOpError("Source and dest shape need to have the same shape.");
+
```
manupak, line 1482: Should we check the dest is a flat buffer here as well ?

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -624,92 +624,67 @@ struct ThreadwiseWriteAllRewritePattern
...
-  std::tie(rawStoreBuffer, storeBufferView, needs64BitIdx) =
-      untransform(b, destView);
-  assert(!needs64BitIdx && "Registers shouldn't need 64-bit indexing");
   ArrayRef<int64_t> rawStoreBufferShape =
       rawStoreBuffer.getType().cast<ShapedType>().getShape();
-
+  if (rawLoadBufferShape.size() != 1)
+    return op.emitOpError("Raw load buffers have to be flat.");
+  if (rawStoreBufferShape.size() != 1)
+    return op.emitOpError("Raw store buffers have to be flat.");
```
manupak, line 657: It might be useful to lift these up to verifiers ?

manupak (commented):
Looks good! just nits modulo handling the uninvertible case.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
giuseros: In order to do that "I think" I  need to `untransform` during verification, and think this  means to have a circular linking dependency: `MLIRRockUtility` depends on `MLIRRockOps`, but now we would need also `MLIRRockOps` to depend on `MLIRRockUtility`.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: No, I don't think it does, or at least it wasn't clear that it did that when I looked yesterday

giuseros: Yeah, so it was still returning a non empty array attr. I made it so that if the inversion fails, it now return an empty ArrayAttr(`{}`).

krzysz00 (approved):
Approved! And, so, after two years, threadwise_copy is back ... but it looks much more reasonable this time

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
krzysz00: ... that's not actually an empty array attr, which is `b.getArrayAttr({})` or `ArrayAttr::get(ctx, {})` or the like. It's actually `ArrayAttr{}` aka `ArrayAttr(nullptr)`.

Which is good as a failure state.

manupak (approved):
LGTM!

PR #1276, opened 2023-10-05
Enable performance tests on Navi3x

yiqian1:
See [Navi3x performance reports](http://rocmhead.amd.com:8080/job/MLIR/job/mlir/view/change-requests/job/PR-1276/21/Performance_20report_20for_20gfx1100/).

It turned out that perfRunner.py already takes care of timeout cases. There is no need to try to catch timeouts in Jenkinsfile.

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -817,74 +817,29 @@ pipeline {
                     }
                     stage("Test MLIR vs MIOpen/rocBLAS") {
                         steps {
-                            script {
-                                dir('build') {
-                                    sh 'date --utc +%Y-%m-%d > perf-run-date'
-                                    // Run MLIR vs MIOpend perf benchmarks.
-                                    try {
-                                        timeout(time: 60, activity: true, unit: 'MINUTES') {
-                                            sh """python3 ./bin/perfRunner.py --op=conv --batch_all \
-                                               --configs_file=${WORKSPACE}/mlir/utils/performance/conv-configs \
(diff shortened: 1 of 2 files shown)


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -112,30 +112,32 @@ void preMergeCheckPackage(String codepath) {
 
 void postProcessPerfRes(String chip) {
     publishHTML (target: [
-        allowMissing: false,
+        allowMissing: ${chip=='gfx1100' ? true, false},
```
krzysz00: `== gfx1100` is the wrong way to check for Navi3x. You probably want `.startswith("gfx11")`

krzysz00 (commented):
I'm not sure I like this approach. Wouldn't it make more sense to exclude the convolution perf testing on Navi3x cards instead of having all these catches?

Similarly, instead of `allowMissing`, modify the list of files dynamically and have it not include the convolution report on Navi3x.

mlir/utils/jenkins/Jenkinsfile
krzysz00: ... never mind, thought this was the Python script

yiqian1:
> I'm not sure I like this approach. Wouldn't it make more sense to exclude the convolution perf testing on Navi3x cards instead of having all these catches?
> 
> Similarly, instead of `allowMissing`, modify the list of files dynamically and have it not include the convolution report on Navi3x.

Why do we want to exclude the convolution perf testing on Navi3x?

I think I can move the creation of html right after the performance test, similar to what "Test MLIR vs CK" does.

krzysz00 (approved):

PR #1278, opened 2023-10-09
Add support for first gemm scaling in attention kernels

manupak:
This commit adds support for scaling the first gemm output of attention kernels.
Additionally support for splat constant scale tensors are also added as they are quite common
with migraphx cases we are interested in supporting.

closes : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1094

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
@@ -1653,6 +1653,21 @@ struct GridwiseAttentionAccelRewritePattern
       }
       accelEmitterPtrGemm0->computeOutputConversion(
           rewriter, loc, accRegBufferGemm0, gemm0OutBuffer, forceUnroll);
+        if(Value scaleIn = op.getScale()){
+            Value rawBuffer;
+            std::tie(rawBuffer, std::ignore, std::ignore) = untransform(rewriter, scaleIn);
+            if(memref::GetGlobalOp constScale = rawBuffer.getDefiningOp<memref::GetGlobalOp>()){
+                FailureOr<TypedAttr> maybeSplatAttr = getSplatGlobalConstant(constScale);
+                if(failed(maybeSplatAttr)){
+                    return op.emitError("Only splat scale constant input is supported."); 
+                }
+                scaleFirstGemmSplat(rewriter, loc, gridCoordsGemm0, gemm0OutBuffer, gemm0OutSubTileViews, maybeSplatAttr.value());
+            }
+            else{
+                scaleFirstGemm(rewriter, loc, gridCoordsGemm0, gemm0OutBuffer, gemm0OutSubTileViews, scaleInBuffer, scaleIn);
+            }
+        }
+        
       bool hasPadding =
           op.getPrePadG0M().has_value() || op.getPrePadG0N().has_value();
       if (hasPadding) {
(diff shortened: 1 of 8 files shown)


mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1425,7 +1500,24 @@ struct GridwiseAttentionAccelRewritePattern
...
+                          "than gemm1LDSByteBufferSize.");
+    }
+    auto gemm1LDSByteBufferAType =
+        MemRefType::get({gemm1LDSByteBufferSize}, rewriter.getI8Type(),
+                        AffineMap{}, workgroupMemoryAddressSpace);
+    Value gemm1LDSByteBufferA = rewriter.create<memref::SubViewOp>(
+        loc, gemm1LDSByteBufferAType, ldsReductionWorkspaceByteBuffer,
+        ArrayRef<int64_t>{0}, ArrayRef<int64_t>{gemm1LDSByteBufferSize},
+        ArrayRef<int64_t>{1});
+    // Value gemm1LDSByteBufferA = ldsReductionWorkspaceByteBuffer;
```
krzysz00: Why commented out here?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1303,6 +1303,74 @@ struct GridwiseAttentionAccelRewritePattern
...
+  }
+
+  FailureOr<TypedAttr>
+  getSplatGlobalConstant(memref::GetGlobalOp getGlobalOp) const {
+    auto *symbolTableOp =
+        getGlobalOp->getParentWithTrait<OpTrait::SymbolTable>();
+    if (!symbolTableOp)
+      return failure();
+    auto global = dyn_cast_or_null<memref::GlobalOp>(
+        SymbolTable::lookupSymbolIn(symbolTableOp, getGlobalOp.getNameAttr()));
```
krzysz00: I think there's a way to look up the nearest symbol, you might want to use that?

krzysz00 (commented):
As an overall comment, what happens if I pass in a scale that isn't a splat? Does it just get silently ignored?

manupak:
> As an overall comment, what happens if I pass in a scale that isn't a splat? Does it just get silently ignored?

So currently we have an open issue about handling bound non-scalar constants : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1171

It is not still clear to me do we want to support non-scalar constants bound to kernels or leave them to host to be supplied.
MIGraphX will not bind these as of now.

No, it does not get silently ignored; it will just fail the pass as unsupported.
But non-scalar args will be handled as long as kernel is concerned.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
manupak: any pointers would be appreciated.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
manupak: forgot to delete

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: `lookupNearestSymbolFrom` is a static method in `SymbolTable`

manupak: thanks! now changed

krzysz00 (approved):

PR #1279, opened 2023-10-10
Fix missing directory creation that broke Makefile-based builds

krzysz00:
Turns out that when we build rocMLIR with `make`, the script to gather up the device libraries was running before the directory that the autogenerated file was meant to go in was created.

This commit fixes that by creating said directory at configure time.


yiqian1:
We need to define the cmake compiler in the MIGraphX stage. We were lucky that we didn't get an error until now.

krzysz00:
... yeah, looks like that was also an issue, have fixed

yiqian1 (approved):

PR #1280, opened 2023-10-12
Improve rpath handling, add workaround for ROCm asan builds

krzysz00:
1. Don't reset the build rpath in the main CMake file, but instead prepend the rocmlir library directory
2. Use an already defined LLVM_EXTERNAL_LIB_DIR variable to set up the build rpath before adding in the LLVM project
3. Pass the rpath back up from configuring llvm-project, thus ensuring any settings made in that include stick around for the rocmlir parts of the build so that we don't have to duplicate them.

Note that in all cases this is the *build* rpath, which isn't used when executables (or shared libraries) are installed, so this will only affect stuff like tablegen.


krzysz00:
... update, there's some extra fiddling needed here, I think, time to look at cmake scoping

jerryyin (approved):
Once this is merged, create another PR that backport this to `release/rocm-rel-6.0` branch.

PR #1285, opened 2023-10-18
[CI] remove pre-installed rocMLIR from MIGraphXDeps

manupak:
This commit removes the rocMLIR that get installed
as per the requirements of MIGraphX as this image
is being used for CI with tip of rocMLIR develop.


giuseros (approved):
LGTM

PR #1288, opened 2023-10-18
Don't set arch_support_mfma True if -mfma=off.

pcf000:


manupak:
@pcf000 first of all thanks for taking a stab at this!

Why limited to gfx90a ?

manupak (approved):
LGTM

PR #1290, opened 2023-10-20
Add -gz to compiler flags in debug builds

krzysz00:
As a tentative fix for
https://ontrack-internal.amd.com/browse/SWDEV-423466 , compress our debug symbols in debug or relwithdebinfo builds.

I haven't tested if this'll work, and I haven't tested this backported onto whatever commit MIOpen builds us with, so this is more here for testing.


manupak:
lets close this ?

krzysz00:
... yeah, we could close this

PR #1292, opened 2023-10-24
Add quick tuning to performance reports

djramic:


mlir/utils/performance/createPerformanceReports.py
```diff
@@ -13,11 +13,26 @@ def printAllPerformance(chip, lib='rocBLAS'):
...
+            f'Tuned/{lib}' : f'Quick Tuned/{lib}',
+            'PerfConfig' : 'Quick PerfConfig'
+        }
+        quickDf.rename(columns=quickColumns, inplace=True)
+        quickDf.drop(columns=['MLIR TFlops', f'{lib} TFlops (no MLIR Kernels)',f'MLIR/{lib}'], inplace=True)
+        if lib == 'rocBLAS':
+            df = pd.merge(df,quickDf, on=reportUtils.GEMM_TEST_PARAMETERS[:-2])
+        else:
+            df = pd.merge(df,quickDf, on=reportUtils.CONV_TEST_PARAMETERS[:-2])
+        df['Quick Tuned/ Tuned'] = df['Quick Tuned MLIR TFlops'] / df['Tuned MLIR TFlops']
```
krzysz00: Nit: extra space in the name here compared to the other columns

mlir/utils/performance/perfRegressionReport.py
```diff
@@ -12,10 +12,22 @@
 
 def loadMlirData(filename: str):
     df = pd.read_csv(filename, sep=',', header=0, index_col=False)
+    quickDf = pd.read_csv('quick_' + filename, sep=',', header=0, index_col=False)
```
krzysz00: Would it be possible to make this optional - or to make the filename of the quick tuning report a configurable option that defaults  to `quick_ + filename`?

That makes sure that the manual regression report generation flow doesn't break

Also, this breaks on filenames such as `../rocmlir-upstream/build/perf_upstream.csv`

mlir/utils/performance/perfRegressionReport.py
```diff
@@ -59,35 +71,47 @@ def computePerfStats(oldDf: pd.DataFrame, newDf: pd.DataFrame, oldLabel: str, ne
...
                          'Tuned MLIR TFlops_old': oldLabelTuned,
-                         'Tuned MLIR TFlops_new': newLabelTuned}, inplace=True)
+                         'Tuned MLIR TFlops_new': newLabelTuned,
+                         'Quick Tuned MLIR TFlops_old' : oldLabelQuickTuned,
+                         'Quick Tuned MLIR TFlops_new' : newLabelQucikTuned}, inplace=True)
     data['% change'] = 100.0 * (data[newLabel] - data[oldLabel]) / data[oldLabel]
     hasTuning = False
     if oldLabelTuned in data and newLabelTuned in data:
-        data['% change (tuned)'] = (data[newLabelTuned] - data[oldLabelTuned]) / data[oldLabelTuned]
+        data['% change (tuned)'] = 100.0 * (data[newLabelTuned] - data[oldLabelTuned]) / data[oldLabelTuned]
```
krzysz00: This'll need a second diff for if there's quick tuned data

mlir/utils/performance/perfRegressionReport.py
krzysz00: ... Heck, why isn't the quick tuned data being written out into the same table as the full tuned data by the perf runner?

It might be a good idea to extend the perf runner with `--quick-tuning-db=`

krzysz00 (commented):
Mostly comments and nitpicks, one broader suggestion

mlir/utils/performance/createPerformanceReports.py
```diff
@@ -13,11 +13,26 @@ def printAllPerformance(chip, lib='rocBLAS'):
 
     try:
         df = pd.read_csv(chip + '_' + reportUtils.PERF_REPORT_FILE[lib])
+        quickDf = pd.read_csv('quick_'  + chip + '_' + reportUtils.PERF_REPORT_FILE[lib])
```
krzysz00: Note remains:
1. This script must not fail if there is no quick tuning perf file.
2. These columns should be in the main dataframe anyway.
3. If you decide you have to have a seperate `quick_` file, that filename needs to be configurable.

mlir/utils/performance/createPerformanceReports.py
krzysz00: Nit again

mlir/utils/performance/createPerformanceReports.py
djramic: I changed the approach to expand the perfRunner with **--quick-tuning-db** . I'll update the PR soon.

PR #1295, opened 2023-10-26
Implement multibuffering pass

giuseros:
This is my stab at implementing a multibuffer pass in rocMLIR:

- I mostly added a transformation `RockMultiBuffer` inspired to the [memref multibuffering](https://mlir.llvm.org/doxygen/MultiBuffer_8cpp_source.html)
- I added an operation `rock.reinterpret_multibuffer` to mark explicitly when a raw `i8` allocation is used as a multibuffer
- I added two interfaces: an interface `RockAcceptingViewOp` which abstracts operations that accepts a view of the memory (i.e., memory that comes from a `rock.transform`). Another `RockWriterOp` interface to abstract operations that write into (any) memory
- I didn't change any of the main lowerings, because we don't have yet a pipeline pass yet, but I added two tests to show how things will pan out when this is applied to a standard (non-pipelined) loop. The transformation is exercised through a `TestRockMultibuffer` test pass 

Fixes: https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1206


mlir/include/mlir/Dialect/Rock/IR/RockAcceptingViewOpInterface.h
```diff
@@ -0,0 +1,24 @@
+//===- RockAcceptingViewOpInterface.h -
+//---===//
```
krzysz00: Can we drop some dashes to get rid of the word wrap here?

mlir/include/mlir/Dialect/Rock/IR/RockAcceptingViewOpInterface.td
```diff
@@ -0,0 +1,61 @@
...
+        /*desc=*/[{
+          "Return the extra indices used to index the operand accepting the view"
+        }],
+        /*retType=*/"::mlir::Operation::operand_range",
+        /*methodName=*/"getExtraIndices",
+        /*args=*/(ins),
+        /*methodBody=*/"",
+        /*defaultImplementation=*/""
+      >,
+    InterfaceMethod<
```
krzysz00, line 47: Might be better to return an `OpOperand` (or maybe it's `OpOperand*`) to the operand being viewed? Or if its operands, have a vector?

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -457,6 +459,26 @@ def Rock_GpuAllocOp:
   }];
 }
 
+// Wrap an allocOp as a multibuffer, i.e., a buffer with #n copies
+def Rock_ReinterpretMultiBufferOp:
+    Rock_Op<"reinterpret_multibuffer", []>,
+    Arguments<(ins Arg<MemRefRankOf<[I8], [1]>, "raw allocation", [MemRead]>:$input,
+                   I32Attr:$multibufferFactor)>,
```
krzysz00: IndexAttr?

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -457,6 +459,26 @@ def Rock_GpuAllocOp:
   }];
 }
 
+// Wrap an allocOp as a multibuffer, i.e., a buffer with #n copies
+def Rock_ReinterpretMultiBufferOp:
+    Rock_Op<"reinterpret_multibuffer", []>,
```
krzysz00: Implement `ViewLikeOp` from upstream here? Along with `Pure`, probably.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -457,6 +459,26 @@ def Rock_GpuAllocOp:
...
+    Arguments<(ins Arg<MemRefRankOf<[I8], [1]>, "raw allocation", [MemRead]>:$input,
+                   I32Attr:$multibufferFactor)>,
+    Results<(outs AnyMemRef:$output)> {
+  let summary = "Reinterpret raw memory as a multi-buffer";
+
+  let description = [{ Wraps a raw memory allocation as a multibuffer, i.e., a buffer with
+    multiple copies. The first dimension of the output type is the multi-buffering factor}];
+  let assemblyFormat = [{
+    $input attr-dict
+    `:` type($input) `->` type($output)
```
krzysz00: Conversions are usually spelled `to`, I've seen

mlir/lib/Dialect/Rock/IR/CMakeLists.txt
```diff
@@ -16,7 +18,8 @@ add_rocmlir_dialect_library(MLIRRockOps
   MLIRRockAttrDefsIncGen
   MLIRRockGemmWrapperInterfaceIncGen
   MLIRRockTuningParamAttrInterfaceIncGen
-  MLIRRockAccelTuningParamAttrInterfaceIncGen
```
krzysz00, line 19: Accidentally deleted a line?

mlir/include/mlir/Dialect/Rock/Transforms/Multibuffer.h
```diff
@@ -0,0 +1,29 @@
...
+//===----------------------------------------------------------------------===//
+
+#ifndef MLIR_DIALECT_ROCK_TRANSFORMS_MULTIBUFFER_H
+#define MLIR_DIALECT_ROCK_TRANSFORMS_MULTIBUFFER_H
+
+#include "mlir/Dialect/Rock/IR/Rock.h"
+
+namespace mlir {
+
+namespace rock {
```
krzysz00: Why's this header exist?

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -1423,6 +1449,18 @@ LogicalResult InBoundsStoreOp::verify() {
 //===-----------------------------------------------------===//
 // ThreadwiseReadIntoOp
 //===-----------------------------------------------------===//
+Operation *
+ThreadwiseReadIntoOp::cloneWithExtraIndices(OpBuilder &builder, Value view,
+                                            ArrayRef<Value> newExtraIndices) {
+  // OpBuilder builder(getContext());
```
krzysz00: Old comment?

mlir/lib/Dialect/Rock/Transforms/RockMultibuffer.cpp
```diff
@@ -0,0 +1,452 @@
...
+
+#define DEBUG_TYPE "rock-multibuffer"
+#define DBGS() (llvm::dbgs() << "[" DEBUG_TYPE "]: ")
+#define DBGSNL() (llvm::dbgs() << "\n")
+
+namespace {
+// Set of utility functions used during the transformation
+
+/// Find the alloc whose transformedValue belongs to
+FailureOr<rock::GpuAllocOp> findAlloc(Value transformedValue) {
```
krzysz00, line 45: In principle, you could return `rock::GpuAllocOp` here and use `nullptr` as the failure ... but I kinda like `FailureOr<>` as it makes it clearer what's up

mlir/lib/Dialect/Rock/Transforms/RockMultibuffer.cpp
```diff
@@ -0,0 +1,452 @@
...
+    if (auto transformOp = dyn_cast<rock::TransformOp>(curOp)) {
+      curOp = transformOp.getInput().getDefiningOp();
+    } else if (auto viewOp = dyn_cast<memref::ViewOp>(curOp)) {
+      curOp = viewOp.getSource().getDefiningOp();
+    } else {
+      return failure();
+    }
+    maybeAllocOp = dyn_cast<rock::GpuAllocOp>(curOp);
+  }
+  return maybeAllocOp;
```
krzysz00, line 61: You may need to test if that `dyn_cast` succeeded?

mlir/lib/Dialect/Rock/Transforms/RockMultibuffer.cpp
```diff
@@ -0,0 +1,452 @@
...
+}
+
+/// Find all the final users of a rockAllocOp. By final users we mean
+/// operations that write/load data from the alloc
+static std::vector<Operation *> findAllocUsers(rock::GpuAllocOp allocOp) {
```
krzysz00: Take a `SmallVectorImpl<Operation *>& ret` as a second parameter?

Or, if you don't care about the order, a `SmallPtrSetImpl<Operation *>&`. (might've spelled it wrong)

mlir/lib/Dialect/Rock/Transforms/RockMultibuffer.cpp
```diff
@@ -0,0 +1,452 @@
...
+  std::vector<Operation *> queue{allocOp.getOperation()};
```
krzysz00: Nit: usually this gets called `worklist`

mlir/lib/Dialect/Rock/Transforms/RockMultibuffer.cpp
```diff
@@ -0,0 +1,452 @@
...
+  std::set<Operation *> results;
```
krzysz00: ... Yeah, this is `[Small]PtrSet`

mlir/lib/Dialect/Rock/Transforms/RockMultibuffer.cpp
```diff
@@ -0,0 +1,452 @@
...
+
+  // Create and add the passthrough transform. Notice that logically
+  // every loop will have its own multibuffer. We already added
+  // a broadcast transformation on top of the transform stack
+  // to route the indices back to the real (smaller) multibuffer
+  SmallVector<int64_t, 4> multiBufferedUpperBounds{loopLength};
+  SmallVector<int64_t, 4> multiBufferedLowerBounds{loopLength};
+  ArrayRef<int64_t> originalUpperBounds = transform.getUpperBounds();
+  ArrayRef<int64_t> originalLowerBounds = transform.getLowerBounds();
+  llvm::append_range(multiBufferedUpperBounds, originalUpperBounds);
```
krzysz00, line 142: Neat function!

mlir/lib/Dialect/Rock/Transforms/RockMultibuffer.cpp
```diff
@@ -0,0 +1,452 @@
...
+  Value subview = rewriter.create<memref::SubViewOp>(loc, dstMemref, val,
+                                                     offsets, sizes, strides);
+  return subview;
+}
+
+/// This is replacing a generic operation op(buffer) with a sequence of
+/// - multibuffer_view = memref.subview(buffer)
+/// - op(multibuffer_view)
+Operation *subviewAndClone(RewriterBase &rewriter, Location loc, Operation *op,
+                           Value val, Value loopIndex, int64_t operandNumber,
```
krzysz00: I _think_ `operandNumber` can be handled by `OpOperand`

mlir/lib/Dialect/Rock/Transforms/RockMultibuffer.cpp
```diff
@@ -0,0 +1,452 @@
...
+
+/// Return if the operation accepts views for operand `operandNumber`
+bool acceptsViewAt(Operation *op, int64_t operandNumber) {
+  auto viewAcceptingOp = dyn_cast<RockAcceptingViewOpInterface>(op);
+  return (viewAcceptingOp != nullptr &&
+          viewAcceptingOp.acceptsViewAt(operandNumber));
+}
+
+/// Return true if the `transformInput` is the first of the transform stack
+bool isTopOfTransformStack(Value transformInput) {
```
krzysz00: I'd go with `isBottomOfTransformStack`

mlir/lib/Dialect/Rock/Transforms/RockMultibuffer.cpp
```diff
@@ -0,0 +1,452 @@
...
+  AffineExpr iv, lb, step;
+  bindDims(rewriter.getContext(), iv, lb, step);
+  Value bufferIndex = affine::makeComposedAffineApply(
+      rewriter, loc, ((iv - lb).floorDiv(step)) % multiBufferingFactor,
+      {ivVal, lbVal, stepVal});
+  LLVM_DEBUG(DBGS() << "--multi-buffered indexing: " << bufferIndex << "\n");
+  return bufferIndex;
+}
+
+/// Return true if the op fully overwrite the given `buffer` value.
```
krzysz00: Inaccurate comment

mlir/lib/Dialect/Rock/Transforms/RockMultibuffer.cpp
```diff
@@ -0,0 +1,452 @@
...
+    if (auto view = dyn_cast<memref::ViewOp>(owner)) {
+      auto newView = rewriter.create<rock::ReinterpretMultiBufferOp>(
+          loc, val, view.getType(), multiBufferingFactor);
+      replaceUsesAndPropagateType(rewriter, loc, view, newView, loop,
+                                  loopLength, multiBufferingFactor);
+    } else if (auto transform = dyn_cast<TransformOp>(owner)) {
+      // rock.transform case      Value toTransform = val;
+      Value toTransform = val;
+      if (isTopOfTransformStack(val) && val.getType().isa<ShapedType>()) {
+        // if this is the top of the transform stack (i.e., the input to the
```
krzysz00, line 282: I'm going to note that I've usually called this the bottom of the stack (see, for example, BottomUpTMBuilder taking the underlying shape as a starting point)

krzysz00 (commented):
I haven't seen any glaring logic bugs on my first read-through, but it also hasn't been the most careful inspection.

(Also, my questions about the header are withdrawn on account of how you're testing this pass)

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
giuseros: Only if you feel strongly about it :) . The nice thing with `I32Attr` (as far as I have seen) is that I can create the operator with a plain `int64_t`. If I use `indexAttr` I need an `APInt`.

manupak: APInt association is the plus side of IndexAttr. to be more precise, we wont hit numerical limit problems with IndexAttr unlike I32Attr as evidenced by @krzysz00's large tensor work.

mlir/include/mlir/Dialect/Rock/IR/RockAcceptingViewOpInterface.td
giuseros: I am not sure I follow. Operations like threadwise_writeall have an interface like `(raw_buffer, view, extraIndicesForTheView, additionalView,...)` . Are you suggesting to return an `OpOperand *operand` to `view` and get the extraIndices as `operand++`?

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
giuseros: Sorry, didn't know that. I switched to `IndexAttr`

mlir/lib/Dialect/Rock/Transforms/RockMultibuffer.cpp
giuseros: If I understood your comment correctly you are saying that I can pass simply the `OpOperand` instead of `{operation*, operandNumber}`, right? This is what I did, but if I misunderstood the comment, please let me know

mlir/include/mlir/Dialect/Rock/IR/RockAcceptingViewOpInterface.td
giuseros: (writing this interface was a bit painful anyway, so I am happy to better nail it down)

krzysz00: Do you need to do a `++` to get to the extraIndices at all?

If not, I'd just go for `SmallVectorImpl<OpOperand*>&` (or `OpOperand`) that's a list of the operands that've been viewed

giuseros: Yes, but that would get me the operands that accept views. I need the indices passed to the view. Basically the interface will be used like:
```
originalIndices = interface.getExtraIndices();
newIndices = originaIndices + additionalIndices;
newOp = interface.cloneWithNewIndices(newIndices, newView);
```

giuseros: Or are you saying that I can get the extra indices from the view?
```
SmallVectorImpl<OpOperand*>& operands = interface.getOperandsAcceptingView();
originaIndices = /*get indices from the operands*/;
newIndices = originaIndices + additionalIndices;
newOp = interface.cloneWithNewIndices(newIndices, newView);
```

krzysz00: Hold on, backing up, is it true that there's only one viewed argument that has extra indices?

giuseros: Yes, for now this is the case. 
- `threadwise_readInto/threadwise_writeAll` accept a single set of extra indices
- `threadwise_copy` does not accept any extra index

But I am happy to generalize this for a generic operation with a signature like: `(operand0, extraIndices0, ...., operandN-1, extraIndicesN-1)` : an idea could be:
```
originalIndices = interface.getExtraIndices(OperandIndex);
newIndices = originaIndices + additionalIndices;
newOp = interface.cloneWithNewIndices(OperandIndex, newIndices, newView);
```
But I am happy to hear more thoughts

krzysz00: Ok, so if there's only one operation, then it's as `OperandRange` for the extra indices an an `OpOperand` for the viewed value

giuseros: Yes, and this is why the `getExtraindices` returns an `operand_range`. I don't understand why I need the `OpOperand`

krzysz00: The point is that, instead of referring to operands by index, you should return the relevant `OpOperand`. Among other things, this'll save you an `O(n)` scan for which operand is the one that has the views applied to it

giuseros: Alright, done(I think)!

mlir/include/mlir/Dialect/Rock/IR/RockAcceptingViewOpInterface.td
```diff
@@ -0,0 +1,61 @@
...
+        /*desc=*/[{
+          "Return the extra indices used to index the operand accepting the view"
+        }],
+        /*retType=*/"::mlir::Operation::operand_range",
+        /*methodName=*/"getExtraIndices",
+        /*args=*/(ins),
```
krzysz00: This should take an `OpOperand*` to specify which set of extra views you want?

mlir/include/mlir/Dialect/Rock/IR/RockAcceptingViewOpInterface.td
```diff
@@ -0,0 +1,61 @@
...
+  let cppNamespace = "::mlir::rock";
+
+  let methods = [
+    InterfaceMethod<
+        /*desc=*/[{
+          "Clone the operation with a new set of extra indices"
+        }],
+        /*retType=*/"Operation *",
+        /*methodName=*/"cloneWithExtraIndices",
+        /*args=*/(ins "OpBuilder& ":$builder, "Value":$view, "ArrayRef<Value>":$newExtraIndices),
```
krzysz00: In general, pass in an `OpOparand*` to specify which value you're replacing?

mlir/lib/Dialect/Rock/Transforms/RockMultibuffer.cpp
krzysz00: That is `return maybeAllocOp ? maybeAllocOp ; failure();`

mlir/lib/Dialect/Rock/Transforms/RockMultibuffer.cpp
krzysz00: Note again

mlir/lib/Dialect/Rock/Transforms/RockMultibuffer.cpp
krzysz00: ... why resolved?

mlir/lib/Dialect/Rock/Transforms/RockMultibuffer.cpp
```diff
@@ -0,0 +1,456 @@
...
+                                         multiBufferingFactor);
+      if (newOp->getNumResults())
+        replaceUsesAndPropagateType(rewriter, loc, owner, newOp->getResult(0),
+                                    loop, loopLength, multiBufferingFactor);
+    }
+    opsToDelete.push_back(owner);
+  }
+  // Delete the tree that is not used
+  for (Operation *op : opsToDelete) {
+    rewriter.eraseOp(op);
```
krzysz00: Check if there are other uses - transforms can be CSE'd together.

krzysz00: ... or just let dead code elimination take care of this

mlir/lib/Dialect/Rock/Transforms/RockMultibuffer.cpp
```diff
@@ -0,0 +1,456 @@
...
+      return failure();
+    }
+    if (!skipOverrideAnalysis) {
+      /// Make sure there is no loop-carried dependency on the allocation.
+      if (!overrideBuffer(user, allocOp.getResult())) {
+        LLVM_DEBUG(DBGS() << "--Skip user: found loop-carried dependence\n");
+        continue;
+      }
+      // If this user doesn't dominate all the other users keep looking.
+      if (llvm::any_of(findAllocUsers(allocOp), [&](Operation *otherUser) {
```
krzysz00: Cache the `findAllUsers(allocOp)`

mlir/test/Dialect/Rock/integration/multibuffer/lit.local.cfg
```diff
@@ -0,0 +1,2 @@
+if "gfx1030" in config.arch:
```
krzysz00: I don't think this is the right way to phrase the exclude. Also, why is it here?

giuseros: This is because I have an `accel_gemm` operator which would not work for non accelerated GPUs

krzysz00: Ah

krzysz00: ... you could test for wmma or mfma in the default feature set instead?

giuseros: Done (I think)

krzysz00 (approved):

PR #1297, opened 2023-10-26
Fix handling of mid-tensor unit dims in transpose/collapse folding

krzysz00:
This commit fixes one of the failure cases in
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/issues/2365 (specifically, it handles the crash, not the numerical accuracy problems, which are still under investigation)

The cause of the crash was attempts to transpose merging on the following sequence of operations:

%t = transpose(%rawA) [0, 1, 3, 2] : tensor<1x1xKxM> to tensor<1x1xMxK>
%a = collapse_shape %t [[0, 1], [2], [3]] : tensor<1x1xMxK> to rock.gemm %c = %a * %b ...

The problem arose because the logic in the collapse shape folder didn't correctly handle the case where the unit dimensions it peeled out of the collapse.

The peeling code code was probably needed so that we could still, for example, combine transpose [0, 2, 3, 1] : tensor<Ax1xBxC> to tensor<AxBxCx1> with collapse_shape [[0, 1], [2], [3]], even though a naive remapping of the reassociation experessions would lead to non-continuous dimensions being merged.
The problem was that it created the set of reassociation expressions [[0], [2], [3]], which triggered assertions in tensor::CollapseShape that all the dimensions (other than the ones on the ends) had to be listed in order.

This should fix the issue in general, but it's possible we'll want to investegate all that code and rip it out if we do problem key v2.

(Also, update a utility script to make sure that static library builds actually rebuild the static library)

### DIFF
--- a/mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
+++ b/mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
@@ -501,6 +501,24 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
           dimMap[newDimsSorted[i]] = i;
           newReassocIndicesSorted.push_back(newReassocIdxMap[newDimsSorted[i]]);
         }
+        // HOTFIX: glue trailing unit dimensions onto collapses that need
+        // them. This is because a case like
+        // %t = transpose %aRaw [0, 1, 3, 2] : tensor<1x1xKxM> ->
+        // tensor<1x1xMxK> %a = collapse_shape [[0, 1], [2], [3]]
+        //    : tensor<1x1xMxK> -> tensor<1xMxK>
+        // will, with the above unit-dimension-removal logic, lead to the
+        // invalid reassociation [[0], [2], [3]], causing a crash.
+        // See MIGraphX bug #2365.
+        // The entire logic here should be reviewed, or at least made less
+        // complex if possible, but ... release-critical bug, what can we do?
+        for (size_t i = 0, e = newReassocIndicesSorted.size() - 1; i < e; ++i) {
+          ReassociationIndices &theseIndices = newReassocIndicesSorted[i];
+          const ReassociationIndices &nextIndices =
+              newReassocIndicesSorted[i + 1];
+          while (theseIndices.back() + 1 < nextIndices[0]) {
+            theseIndices.push_back(theseIndices.back() + 1);
+          }
+        }
         for (size_t i = 0; i < newDims.size(); i++) {
           newDims[i] = dimMap[newDims[i]];
         }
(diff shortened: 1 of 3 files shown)


mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -501,6 +501,25 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
...
+        // HOTFIX: glue trailing unit dimensions onto collapses that need
+        // them. This is because a case like
+        // %t = transpose %aRaw [0, 1, 3, 2] : tensor<1x1xKxM> -> tensor<1x1xMxK>
+        // %a = collapse_shape [[0, 1], [2], [3]]
+        //    : tensor<1x1xMxK> -> tensor<1xMxK>
+        // will, with the above unit-dimension-removal logic, lead to the
+        // invalid reassociation [[0], [2], [3]], causing a crash.
+        // See MIGraphX bug #2365.
+        // The entire logic here should be reviewed, or at least made less complex
+        // if possible, but ... release-critical bug, what can we do?
```
jerryyin: Ok... While I get the frustration there, I don't think this is the most professional way to put in the code base. How about create a ticket and then reference it as "Following logic to be reviewed in rocMLIR#<ticket id>"?

jerryyin (approved):

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -501,6 +501,25 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
...
+        for (size_t i = 0, e = newReassocIndicesSorted.size(); i < e; ++i) {
+          if (i == e - 1)
```
sjw36: why not make the loop bounds `i < e - 1`?

sjw36 (approved):
Looks good. Just one minor cleanup.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
krzysz00: To call attention to the fact that this doesn't affect the last item in the list. But yeah, this could just work on the loop bounds.

krzysz00: Fixed since I needed to clang-format anyway

manupak (approved):
LGTM!

Im hoping we could get rid of transpose folding with your shapes, strides representation -- which I think what you refer as problem-key v2.

manupak:
FYI : https://github.com/ROCmSoftwarePlatform/rocMLIR/discussions/1054 was created with that intent

krzysz00:
Problem key v2 is unrelated to NHWC / shapes + strides stuff

PR #1300, opened 2023-10-30
Remove broadcast handling from vectorization / contiguous merges

krzysz00:
We have ran into the following error case from an MIGraphX unit test (MLIR integration tests for this to follow Later (tm)):

- %aRaw : tensor<2x1x2x3>
- %aBcast = broadcast %aRaw to tensor<2x3x2x3>
- %aMergeBatch = merge{2, 3} %aBcast to tensor<6x2x3> rock.gemm ... %aMergeBatch * ...

The contiguos merge handling code would rewrite the merge{2, 3} to a merge{1, 6}, and then the optimization that v mod 1 == 0 would kick in, which would mean that the read would always read from the first batch (causing the second half of the matrix multiply to be incorrect).

Because of possibilities like this, I don't think we can have the clever "vectorize even in the presence of some broadcasts" setup we had before this commit, as I can't see any way to make it correct in general.

To put it another way, if you have the maps
(d0, d1, d2, d3) -> (d0, 0, d2, d3) # broadcast
and
(d0, d1, d2) -> (d0 / 3, d0 % 3, d1, d2) # merge
you can't optimize the merge to
(d0, d1, d2) -> (0, d0, d1, d2)
which is what our code was doing.

This commit therefor removes the isBroadcast field from all the vectorization tracking structs, and updates the unit tests.

### DIFF
--- a/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
+++ b/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
@@ -248,8 +248,6 @@ struct DimInfo {
   TransformMapAttr transformMap; // Map to which the dimension belongs
   TransformAttr transform;       // Transform to which the dimension belongs
   size_t positionInMerge; // Position of the dimension inside the transform
-  bool isBroadcast;       // Is this a broadcast dimension?
-  bool isPadded;          // Is this a padded dimension?
 
   // Utility function to return a transform pair <map,transform>
   std::pair<TransformMapAttr, TransformAttr> transformPair() {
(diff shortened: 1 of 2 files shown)


krzysz00:
I've checked the MIGraphX unit tests, they pass with this change (and the other backportable change).

This is a high-priority 6.0-critical PR

manupak (approved):
LGTM!

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -467,10 +456,7 @@ ContiguousMergesMap findContiguousGroups(ArrayAttr transforms,
         for (auto pair : llvm::zip(upperDims, lowerDims, params)) {
           uint32_t u = std::get<0>(pair);
           uint32_t l = std::get<1>(pair);
-          uint32_t p = std::get<2>(pair);
```
giuseros, line 470: This should be incorporated into the `passthrough/slice` case, I think

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -308,18 +307,9 @@ void findCountiguousGroupsUnmerge(const ArrayRef<uint32_t> upperDims,
...
         break;
 
-      // b) Unmerge parameters need to match merge parameters. Note that if
-      // the merge dimension goes through a Broadcast (before getting unmerged),
-      // this should be taken into account (i.e., the new parameter should be 1)
-      // However, if its padded, we should not be using 1 as the actual length
-      // becomes important.
-      int64_t dimJ = mergeJDims[posJ];
-      int64_t paramOrBroadcast =
-          (dimToMerge[dimJ].isBroadcast && !dimToMerge[dimJ].isPadded)
```
giuseros, line 318: I guess you can remove `isPadded` as well from the struct?

giuseros (approved):
I see the issue, and I think this looks similar to: https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1231 

Is it worth adding a TODO to better address this in  `collapseContiguousDims` ? 

Other than that and a couple of refactoring, this LGTM

ahsan-ca (approved):

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
krzysz00: Entirely possible, I'll take a look

krzysz00: Yep, done

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
krzysz00: Yep

PR #1301, opened 2023-10-30
[BACKPORT] Fix MIGraphX padding to TOSA padding

manupak:
Currently MIGraphX provides padding as
[hlow, wlow, hhigh, whigh] while TOSA expects
[hlow, hhigh, wlow, whigh]. However, our code
just copies it as it is.

This commit fixes this problem


krzysz00:
Let's not land the test, it depends on functionality that's not in the release branch (Yi's clone work)

krzysz00 (approved):
Approved, though we should take the test out

manupak:
removed them

PR #1302, opened 2023-10-31
[BACKPORT] Remove broadcast handling from vectorization / contiguous merges

jerryyin:
We have ran into the following error case from an MIGraphX unit test (MLIR integration tests for this to follow Later (tm)):

- %aRaw : tensor<2x1x2x3>
- %aBcast = broadcast %aRaw to tensor<2x3x2x3>
- %aMergeBatch = merge{2, 3} %aBcast to tensor<6x2x3> rock.gemm ... %aMergeBatch * ...

The contiguos merge handling code would rewrite the merge{2, 3} to a merge{1, 6}, and then the optimization that v mod 1 == 0 would kick in, which would mean that the read would always read from the first batch (causing the second half of the matrix multiply to be incorrect).

Because of possibilities like this, I don't think we can have the clever "vectorize even in the presence of some broadcasts" setup we had before this commit, as I can't see any way to make it correct in general.

To put it another way, if you have the maps
(d0, d1, d2, d3) -> (d0, 0, d2, d3) # broadcast
and
(d0, d1, d2) -> (d0 / 3, d0 % 3, d1, d2) # merge
you can't optimize the merge to
(d0, d1, d2) -> (0, d0, d1, d2)
which is what our code was doing.

This commit therefor removes the isBroadcast field from all the vectorization tracking structs, and updates the unit tests.

### DIFF
--- a/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
+++ b/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
@@ -248,7 +248,6 @@ struct DimInfo {
   TransformMapAttr transformMap; // Map to which the dimension belongs
   TransformAttr transform;       // Transform to which the dimension belongs
   size_t positionInMerge; // Position of the dimension inside the transform
-  bool isBroadcast;       // Is this a broadcast dimension?
   bool isPadded;          // Is this a padded dimension?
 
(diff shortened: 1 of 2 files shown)


manupak (approved):
Approving the backport

PR #1305, opened 2023-11-01
Update the fusion tests to use rocmlir-gen to create the clone harness logic

yiqian1:


krzysz00 (approved):
Approved given that you've done a nightly run

PR #1310, opened 2023-11-03
Fix blockwise_reduce reduction tree and initialization

manupak:
Currently the blockwise reduction tree was not getting the decreasing active thread bounds goind deep into the tree. Also, the max initialization was not getting proper neg inf. This commit fixes all that.

closes : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1242
closes : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1247


giuseros (approved):
LGTM, we should do something about `clang-tidy`

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -763,6 +761,13 @@ struct BlockwiseReduceRewritePattern
     }
   }
 
+  int64_t roundToNextPowerof2(int64_t n) const {
```
krzysz00: Please check the LLVM support libraries for this function, it should exist.

Also, this might be incorrect once you're in the range where `(double)(i) == (double)(i + 1)`?

krzysz00 (approved):
Approved for being a bugfix, but pleas check to make sure there's not a better way to do this.

(ex. nextPowerOfTwo feels like something that gets implemented with count leading zeroes or one of its relatives, and that's probably in the LLVM support libraries)

PR #1311, opened 2023-11-06
[Attention] Persist Q tile in regs if KPerBlock == head_dim

manupak:
The focus of the PR is that when the KPerBlock is equal to head_dim (K dim of the first gemm),
to prefetch the Q tile ( the tile of A tensor of first gemm) directly to regs outside the main (outermost)
loop of the attention kernel. This will remove the need to do repeatitve loading of Q in the main loop
of the attention kernel.

However, the current semantics of blockwise_gemm requires the operands to be in the LDS. Therefore,
this change required to break the first blockwise_gemm into its lowered version and lift the loading of
the first operand (LDS->regs) out of the loop.

This depends on : https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1310
closes : https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1232


mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -762,6 +761,13 @@ struct BlockwiseReduceRewritePattern
     }
   }
 
+  int64_t roundToNextPowerof2(int64_t n) const {
```
krzysz00: There's still an LLVM utility for this

mlir/include/mlir/Dialect/Rock/IR/Rock.h
```diff
@@ -40,6 +40,12 @@ class PatternRewriter;
 #include "mlir/Dialect/Rock/IR/ConvolutionDims.h"
 #include "mlir/Dialect/Rock/IR/GemmSize.h"
 
+// Following ifdef could be used to change
+// the attention operator to be a fused gemm-gemm
+// kernel for debugging purposes. This will also
+// adjust the test harness to verify the same as well
+// #define ROCK_DEBUG_ATTENTION_REMOVE_SOFTMAX
```
krzysz00, line 47: ... Why? Also, why here?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1368,6 +1380,24 @@ struct GridwiseAttentionAccelRewritePattern
         });
   }
 
+  void LoadGemmOperandsFromLDSToRegs(
```
krzysz00: Nit: capatilization. This is `mlir/`, not `llvm/`

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1368,6 +1380,24 @@ struct GridwiseAttentionAccelRewritePattern
...
+      PatternRewriter &rewriter, Location loc, Value ldsTileBuffer,
+      ArrayRef<Value> preAccelRegBuffers, StringRef dName, int64_t blockSize,
+      int64_t inDPerThread,
+      const std::unique_ptr<accel::AccelEmitter> &accelEmitterPtr) const {
```
krzysz00: Does this need to be a reference to the unique_ptr, or can you just `const AccelEmitter&` here?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1368,6 +1380,24 @@ struct GridwiseAttentionAccelRewritePattern
...
+    // Get current workitem ID.
+    auto tid = rewriter.create<WorkitemIdOp>(loc, rewriter.getIndexType());
+    rock::accel::AccelEmitterParams accelParams = accelEmitterPtr->getParams();
+    Value wrappedLDSBufferForLoad = accelEmitterPtr->wrapLDSBufferForLoad(
+        rewriter, loc, ldsTileBuffer, blockSize, inDPerThread, dName, false);
+    for (auto [idx, preAccelRegBuffer] : llvm::enumerate(preAccelRegBuffers)) {
+      Value m_i = rewriter.createOrFold<ConstantIndexOp>(loc, idx);
```
krzysz00: Nit on underscores

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1588,6 +1623,37 @@ struct GridwiseAttentionAccelRewritePattern
          elemTypeOut});
 
     zeroAccBuffer(rewriter, loc, attentionOutAccBuffer);
+#ifdef ROCK_DEBUG_ATTENTION_REMOVE_SOFTMAX
+    zeroAccBuffer(rewriter, loc, accRegBufferGemm1);
```
krzysz00, line 1636: ... Ok, yeah, I'm seeing the logic here, but wouldn't it make sense to make this an op attribute instead?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1588,6 +1623,37 @@ struct GridwiseAttentionAccelRewritePattern
...
+                 << "rock.attention: Prefetching Q tile into regs...\n");
+      Value zero = rewriter.createOrFold<ConstantIndexOp>(loc, 0);
+      LogicalResult statusLoadQTile = loadAndStoreGemmInputTile(
+          loc, inQ, /*kiter=*/zero, gridCoordsGemm1, fromGlobalRegBufferQ,
+          toLDSRegBufferQ, ldsByteBufferQ, "m", gemm0kpack, gemm0KpacksPerBlock,
+          gemm0MPerBlock, blockSize, gridSize, bidGridOrder,
+          gemm0BidGridLengths, forceUnroll, rewriter);
+      ldsTileBufferQ = viewBufferAs(rewriter, ldsByteBufferQ,
+                                    vectorTypeOrSelf(elemTypeQ, gemm0kpack));
+      if (failed(statusLoadQTile)) {
```
krzysz00, line 1654: If before applying views

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1769,9 +1876,9 @@ struct GridwiseAttentionAccelRewritePattern
             rewriter.getI32IntegerAttr(gemm1InMPerThread),
             rewriter.getI32IntegerAttr(gemm1InNPerThread),
             /*rotateMWithK=*/nullptr,
-            /*rotateNWithK=*/nullptr, preAccelRegBufferQxK, preAccelRegBufferV,
-            accRegBufferGemm1, op.getArchAttr(), op.getFeaturesAttr(),
-            op.getBlockSizeAttr(), gemm1TuningParams);
+            /*rotateNWithK=*/nullptr, preAccelRegBuffersQxK[0],
```
krzysz00: Why hardcoded [0] here?

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2437,8 +2437,11 @@ static func::FuncOp createCpuAttentionKernelWithMlir(ModuleOp module,
       builder, loc, elemType, expsTensor, reductionAxis);
   auto invExpsSums =
       createOpAndInfer<tosa::ReciprocalOp>(builder, loc, elemType, expsSums);
-  auto softmaxTensor = createOpAndInfer<tosa::MulOp>(
+  Value softmaxTensor = createOpAndInfer<tosa::MulOp>(
       builder, loc, elemType, expsTensor, invExpsSums, /*shift=*/0);
+#ifdef ROCK_DEBUG_ATTENTION_REMOVE_SOFTMAX
```
krzysz00, line 2442: Still kinda feels like it should be a flag somehow?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -930,24 +930,36 @@ struct GridwiseAttentionAccelRewritePattern
 
   // This fuction creates interrim register buffers to store data in once
   // loaded from the LDS before accelerator intrinsics are called
-  std::tuple<Value, Value>
+  std::tuple<SmallVector<Value>, SmallVector<Value>>
```
krzysz00: High-level question: how'll this integrate with pipelining/multibuffering?

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
manupak: this is not relevant to this PR; just inherited from https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1310 and its its fixed there. So it ll disappear when I merge develop after 1301 is merged.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
manupak: I thought about it... I dont want this to be a functional thing. My thoughts were it needed to be very global switch that changes the behaviour in two places in the code base (here and rocmlir-gen) purely for debugging -- which I found useful : for e.g. that is how I figured blockwise_reductions are doing something wrong.

Op attribute would kinda make it a functional thing and I would need to more plumbing in rocmlir-gen reference generation. I kind of have the philosophy debugging capabilities should be simple as possible; so that one does not need to debug the debugging utilities.

What do you think ? or more precisely why do you think op attribute would be better ?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
manupak: Moved to be repeat x ... memref instead now. so its more "compatible" with multibuffering

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
manupak: done

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
manupak: went away with higher-rank repeat x .. representation

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
manupak: done

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: I see your point about this being very much a debug switch, and thus something that's off in a source code edit ... though I might then take a page out of LLVM's book and make it a `ReallyHidden` CLI option instead.

I was thinking in the context of having GEMM-*-GEMM support in the future

manupak:
@krzysz00 I have addressed all you comment except for the debug switch.
I still prefer it to be an ifdef (which had been useful for debugging I ve done this week) and if you really dont like it -- I can remove it altogether (and contemplate later to add it back as an attribute)

krzysz00 (commented):
Ok, so, just to check, these are just changes to the attention kernel? Should they perhaps be incorporated to the gemm kernel as well?

manupak:
No its only related to attention kernels.

krzysz00 (approved):
Looks overall fine to me, though I'd like to see some of these attention utilities refactored into the gemm kernel too

PR #1312, opened 2023-11-06
[USE MERGE COMMIT] Improve MIGraphX IR dialect with migraphx.shaped type

krzysz00:
Fixes https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1057

Fixes https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1058

As planned in those tickets, this commit adds a `!migraphx.shaped` type, whose syntax is
`<[len1]x[len2]x...x[type], [stride1]x[stride2]x...>`

This type matches the current semantics of MIGraphX's types for their IR variables, and allows us to encode information about the expected strides of those variables. This will allow us to generate NHWC convolutions for MIGraphX, since they represent these convolutions using values whose lengths are in NCHW order but whose strides make the value NHWC.

While we're here, I have refactored the MIGraphX dialect to a more full-featured tree structure with IR/ and Transforms/ directories.

All the MIGraphX operations have been changed to take `!migraphx.shaped` inputs and outputs instead of `tensor` ones.

(Another side effect has been that the `migraphx.constant` operation, which was pretty weird and, IIRC, never lowered to anything, has been replaced by `migraphx.literal`.)

The MIGraphXToTosa conversion now proceeds in two phases. In the first, we convert the bodies of functions from the MIGraphX dialect to Tosa, translating all MIGraphX values to a tensor value whose type matches their logical shape (that is, we toss out the strides). At the boundaries of the body, we insert `migraphx.mlir.as_logical_shape` and `migraphx.mlir.as_underlying_shape` operations to preserve the type safety of function arguments and calls to `return`.

IN the second phase (which could, hypothenically, be decoupled into a different pass if some future extention required it), we convert function boundaries and those `as_{logical/underlying}_shape` operations introduced earlier. This time, however, MIGraphX shaped values are mapped to a `tensor<>` whose dimensions represent the underlying memory layout of the values in memory. The shape translation operations insert any transposes (and, when reading, slices) needed to translate this memory layout shape to the logical shape the body of the function expects.

For example, the value `!migraphx.shaped<4x3xf32, 1x4>` becomes `tensor<4x3>` when used in the body of a function, but `tensor<3x4>` as a function argument or return value, encoding the transposed natuer of the memory layout.

All MIGraphX-dialect tests have been updated to use the new type, with standard-shape strides inferred.

This commit also adds integration tests for the MHWC convolution.


mlir/include/mlir-c/Dialect/MIGraphX.h
```diff
@@ -22,10 +22,25 @@ extern "C" {
...
-#define MLIR_MIGRAPHX_DIALECT_API_VERSION 3
+// Version 4: The MLIR shaped type is added to better represent MIGRaphX's
+// native type
+#define MLIR_MIGRAPHX_DIALECT_API_VERSION 4
 
 MLIR_DECLARE_CAPI_DIALECT_REGISTRATION(MIGraphX, migraphx);
 
+// Types
+
+MLIR_CAPI_EXPORTED MlirTypeID rocmlirMIXRShapedTypeGetTypeId(void);
```
manupak, line 33: [naming] This becomes inconsistent with other APIs.
I feel we should make this mlir* for now and do a later change to rocmlir if need be.

mlir/include/mlir-c/Dialect/MIGraphX.h
```diff
@@ -22,10 +22,25 @@ extern "C" {
...
+
+MLIR_CAPI_EXPORTED bool rocmlirIsAMIXRShapedType(MlirType type);
```
manupak, line 35: [naming] Can we use MIGraphX rather than MIXR ? 
It becomes a bit confusing XMIR and anyother permutation of those characters.

manupak (commented):
I had a quick skim over it.... seems like most changes are noise from adopting the new type. It would be great if you can point out places you;d like a another pair of eyes on.

A very highlevel question and I ve looked at upstream -- and it seems a like problem upstream shares interest as well.

https://discourse.llvm.org/t/tensor-layout-in-mlir/70075/2
https://reviews.llvm.org/D99548

Any reason why we are diverging from that ?
(Not considering as a change for this PR; but I d like to understand why)

krzysz00:
Re upstream's tensor layout attribute
1. It seemed to be only used for sparse, and I didn't want to run into any weird hardcoded assumptions
2. Having our own type representation lets us do whatever we want to it in the future and control its translation more carefully.

But also I don't think there's any obvious reason _not_ to use a tensor encoding. However, it just felt better to have the MIGraphX dialect have a type that represents MIGraphX IR's types.

mlir/include/mlir-c/Dialect/MIGraphX.h
krzysz00: We already have some `rocmlir` APIs, if I remember right, and I wanted to set the trend going forward.

mlir/include/mlir-c/Dialect/MIGraphX.h
krzysz00: ... we probably could, yeah, it just turned the name rather long (but it turns out we don't say it too much)

So I'm good to change it if you'd like

krzysz00:
The places I'd like another pair of eyes on are `MIGraphX.cpp` and `MIGraphXToTosa[Pass].cpp`

The test changes are all noise

manupak:
> The places I'd like another pair of eyes on are `MIGraphX.cpp` and `MIGraphXToTosa[Pass].cpp`
> 
> The test changes are all noise

Cool I ll have another pass sometime today.

mlir/lib/Dialect/MIGraphX/IR/MIGraphX.cpp
```diff
@@ -247,29 +247,29 @@ void MIXRShapedType::getStridesToStandardShapePermutation(
...
-  // in the original strides array, which ensures that sorting by < leads
-  // to shapes like <1x2x1x3xf32, 6x3x3x1> are not permuted.
   SmallVector<std::pair<int64_t, size_t>, 4> indexedStrides;
   for (auto [idx, stride] : llvm::enumerate(strides)) {
     if (stride == 0) {
       // Broadcast dimensions stay put.
       ret[idx] = idx;
     } else {
-      indexedStrides.push_back({stride, n - idx});
+      indexedStrides.push_back({stride, idx});
```
pfultz2: Using `idx` or `n - idx` wont work correct for many cases.

Using `n - idx`, then cases like `reorder_shape({shape::float_type, {1, 24, 24, 1}}, {0, 3, 1, 2})` will have a permutation of `{0, 2, 1, 3}` instead of `{0, 3, 1, 2}`.

Using `idx`, is even less correct as cases with standard shape like `shape{shape::float_type, {64, 1, 24, 24}}` gives a permutation of `{1, 0, 2, 3}` instead of `{0, 1, 2, 3}`.

The algorithm that is used by both [migraphx](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/permutation.cpp#L44) and [miopen](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/include/miopen/tensor.hpp#L218) is to sort with the strides and dimensions(or `lens`):

```cpp
std::vector<int64_t> find_permutation(const shape &s) {
  std::vector<std::int64_t> result(s.lens().size());
  std::iota(result.begin(), result.end(), 0);
  std::stable_sort(result.begin(), result.end(), [&](auto x, auto y) {
    return std::make_tuple(s.strides()[x], s.lens()[x]) >
           std::make_tuple(s.strides()[y], s.lens()[y]);
  });
  return result;
}
```

This came from a long discusson with me and @jerryyin on how to handle many of these edge cases(see discussion [here](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/765#discussion_r596465551)). There are a lot of unit tests in migraphx to check the correctness of this implementation as well, so I strongly recommend using this well tested version over the version here that doesnt even have unit tests.

krzysz00: Ok, hold on, that doesn't make sense. How is the length of the dimension being sorted the tiebreaker? Wouldn't it be more reasonable to break ties between strides by position in the permuted shape?

Also, the permutations you posted are inverses of each other, so I wonder if some of the confusion here is due to us trying to generate two different permutation.

The function I'm writing is the one, that given a shape `s`, produces a permutation `p` such that `p(strides(s))` is in standard order and that is the identity on broadcasted dimensions

pfultz2: >  How is the length of the dimension being sorted the tiebreaker? 

Because the 1 dimensions repeats the strides of the previous dimensions. So when strides are equal you want the 1 dimension to come first and then the dimension that is not 1 to come after.

> Wouldn't it be more reasonable to break ties between strides by position in the permuted shape?

No because the shape has already been permutated so you dont know which dimension originally came first. 

> The function I'm writing is the one, that given a shape s, produces a permutation p such that p(strides(s)) is in standard order

Yes that is the inverse(Usually people refer to the permutation as we wrote it in migraphx(ie NHWC or 0231))

We compute the inverse permutation by using a sort with `std::less`(see [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/permutation.cpp#L39)), so it might be possible to get `find_permutation` to produce this permutation by flipping the `>` to a `<` instead in the sort comparator.

krzysz00: I'll move to your logic and find a good name for it

pfultz2: > We compute the inverse permutation by using a sort with std::less(see [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/permutation.cpp#L39)), so it might be possible to get find_permutation to produce this permutation by flipping the > to a < instead in the sort comparator.

So this probably wont work, I think it just reverses the permutation instead of inverting it. Maybe you could do something like `std::make_tuple(s.strides()[x], std::numeric_limits<int64_t>::max() - s.lens()[x])` in the comparator, but I think just inverting it would be simpler.

krzysz00: Ok, fixed, had to tweak a few tests but at least we caught the potential incompatibility

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -59,623 +101,1036 @@ static Type getShapedElementTy(Value v) {
...
-    }
-
-    // transpose the output back to NCHW so that it can match following
-    // operators.
-    auto top = getRank4TransposeOp(loc, cop, rewriter, NHWC2NCHW);
-    rewriter.replaceOp(op, top);
-    return success();
-  }
+  matchAndRewrite(ConvType op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override;
```
manupak, line 187: Why is the declaration and definition seperated like this now ?

manupak: (it makes it bit harder to spot the differences)

manupak (commented):
I did another pass at MIGraphXToTosa... my reading is there are not a substantial change there but just noise from your choice separating declaration and definitions in the same file.
While im not against such a change in principal, but that adds lot of unncessary noise in this PR where I think value is not there...

Overall, I d like to see a seperate file just for type conversion tests (apologies if I missed that in all the noise -- please point me at it) that covers important cases (not limited to) : 
1) transposed layout of differing strides
2) transposed layout of differing strides + unit dims
3) transposed layout with equal dim lengths 
4) a case where original ordering of strides is effectively used because of strides being equal

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
krzysz00: Fair. This is a "let's clean up style while we're here" change.

krzysz00:
@manupak The file you're looking for is `mlir/test/Conversion/MIGraphXToTosa/migraphx-to-tosa.mlir`, but it doesn't have all of the tests you asked for - I'll add them

manupak (approved):
Thanks for the tests! I dont have major concerns here...

PR #1313, opened 2023-11-08
Update perfConfigs list that incorporate all tier1 models

djramic:


krzysz00 (approved):
I'm approving because I assume you've tested to make sure that there aren't perf regressions

PR #1318, opened 2023-11-12
Add a code-coverage stage to Jenkinsfile

pcf000:
Add a code-coverage stage.
Add some agent and matrix framework.
Trying to make the uploader work.
Add proxy with -U for lockhart.
Make an html report, too, and archive the reports (anticipating codecov upload failures).


pcf000:
See https://confluence.amd.com/pages/viewpage.action?pageId=533483027

The new stage is controlled by a new parameter, runCodeCoverage, which defaults to false, at least until we figure out when we want to run it.  It does essentially the static-lib stage, but compiled for profiling, and then generates a few reports.  It uploads one to codecov in the recommended manner, but at the moment that silently fails while claiming to succeed;  I think there's a licensing issue.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -1057,5 +1058,68 @@ pipeline {
...
+
+        stage ("Code coverage") {
+            when {
+                beforeAgent true;
+                equals expected: true, actual: params.runCodeCoverage;
+            }
+            matrix {
+                axes {
+                    axis {
+                        name 'CHIP'
```
pcf000, line 1000: Might as well account for target differences.  Can be expanded or removed.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -1057,5 +1058,68 @@ pipeline {
...
+                        args dockerArgs()
+                        label getLabelFromChip("${CHIP}")
+                        alwaysPull true
+                    }
+                }
+                environment {
+                    PATH="/opt/rocm/llvm/bin:$PATH"
+                    HOME="${WORKSPACE}"
+                    // Note the %m to avoid issues with threads and dynamic libraries.
+                    LLVM_PROFILE_FILE="${WORKSPACE}/build/%m-%p.profraw"
```
pcf000, line 1016: The %m fills in a unique identifier for the binary;  %p is the process ID.  Otherwise, the default file is "default.profraw" and we get collisions from multiple simultaneous tests and also when both main program and .so are instrumented.  Also, by sticking them all in build, the *.profraw files are easier to find.

pcf000: Also, by sticking them all in build, the *.profraw files are easier to find.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -1057,5 +1058,68 @@ pipeline {
...
+                            sh 'rm -f build/*.profraw'
+                            buildProject('check-rocmlir-build-only',
+                                         '-DBUILD_FAT_LIBROCKCOMPILER=ON -DCMAKE_BUILD_TYPE=debug -DCMAKE_CXX_FLAGS_DEBUG="-O0 -fprofile-instr-generate -fcoverage-mapping"')
+                            dir ('build') {
+                                // Run tests.
+                                sh 'ninja check-rocmlir'
+                                // Profile processing.
+                                sh '/opt/rocm/llvm/bin/llvm-profdata merge -sparse ./*.profraw -o ./coverage.profdata'
+                                sh '/opt/rocm/llvm/bin/llvm-cov report --object ./bin/rocmlir-opt --object ./bin/rocmlir-driver --object ./bin/rocmlir-gen --instr-profile ./coverage.profdata --ignore-filename-regex=external/llvm-project > ./coverage.report'
+                                sh 'cat ./coverage.report'
```
pcf000: Summary report, percentages per file.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -1057,5 +1058,68 @@ pipeline {
...
+                                sh '/opt/rocm/llvm/bin/llvm-cov show --object ./bin/rocmlir-opt --object ./bin/rocmlir-driver --object ./bin/rocmlir-gen --instr-profile ./coverage.profdata --ignore-filename-regex=external/llvm-project -Xdemangler=/opt/rocm/llvm/bin/llvm-cxxfilt > ./coverage.info'
```
pcf000: Source files annotated with visit counts or highlighted if unvisited.  coverage.info is plain text and intended for upload, coverage.html is formatted and colorful.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -1057,5 +1058,68 @@ pipeline {
...
+                                sh '/opt/rocm/llvm/bin/llvm-cov show --object ./bin/rocmlir-opt --object ./bin/rocmlir-driver --object ./bin/rocmlir-gen --instr-profile ./coverage.profdata --ignore-filename-regex=external/llvm-project -Xdemangler=/opt/rocm/llvm/bin/llvm-cxxfilt --format=html > ./coverage.html'
+                                // Upload to codecov.
+                                withCredentials([string(credentialsId: 'codecov-token-rocmlir', variable: 'CODECOV_TOKEN')]) {
```
pcf000: The token can be found at https://app.codecov.io/gh/ROCmSoftwarePlatform/rocMLIR/settings and I've added it to rocmhead.amd.com:8080's credentials.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -1057,5 +1058,68 @@ pipeline {
...
+                                    sh '''
+                                       curl -Os https://uploader.codecov.io/latest/linux/codecov && chmod +x ./codecov
+                                       proxy_opt=""
+                                       if [ "${CHIP}" = "gfx90a" ]; then
+                                           proxy_opt='-U http://172.23.0.23:3128'
```
pcf000: Lockhart needs a proxy for everything.  I'm a little cranky about hard-coding the address, but I don't know if I can count on $http_proxy.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -1057,5 +1058,68 @@ pipeline {
...
+                                       fi
+                                       ./codecov -t ${CODECOV_TOKEN} --flags "${CHIP}" -f ./coverage.info ${proxy_opt}
```
pcf000: Using the target as a flag lets us see different plots for different targets, or would if uploads worked.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -1057,5 +1058,68 @@ pipeline {
...
+                                       '''
+                                }
+                            }
+                            archiveArtifacts artifacts: 'build/coverage.report, build/coverage.info, build/coverage.html', onlyIfSuccessful: true
```
pcf000: Archiving mostly to save the html so we at least have something to look at.  Some projects do further processing to make coverage diffs or to detect decreased coverage after pull requests, both of which would be future projects for us.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -1057,5 +1058,68 @@ pipeline {
...
+                    LLVM_PROFILE_FILE="${WORKSPACE}/build/%m-%p.profraw"
+                }
+                stages {
+                    stage ("body") {
+                        steps {
+                            // Build with profiling on, and just code-generation tests.
+                            sh 'rm -f build/CMakeCache.txt'
+                            sh 'rm -f build/*.profraw'
+                            buildProject('check-rocmlir-build-only',
+                                         '-DBUILD_FAT_LIBROCKCOMPILER=ON -DCMAKE_BUILD_TYPE=debug -DCMAKE_CXX_FLAGS_DEBUG="-O0 -fprofile-instr-generate -fcoverage-mapping"')
```
krzysz00: Check if there's something in the LLVM CMake files for this

mlir/utils/jenkins/Jenkinsfile
krzysz00: You should be able to count on it, and testing for gfx90a will break non-Lockharts

krzysz00: (That is, if `$http_proxy` is unset, don't use one)

krzysz00 (commented):
Some initial thoughts

mlir/utils/jenkins/Jenkinsfile
pcf000: There's -DLLVM_BUILD_INSTRUMENTED_COVERAGE, though it doesn't include the -O0.  Is that the sort of thing you had in mind?

mlir/utils/jenkins/Jenkinsfile
pcf000: No non-lockhart gfx90a in the pool at the moment.  I'll have to experiment before I believe I can exploit the presence or absence of $http_proxy.

mlir/utils/jenkins/Jenkinsfile
krzysz00: Yeah, though I'm not convinced we *need* the `-O0`

Or maybe we'd get it by just doing a `Debug` build too

pcf000:
Went with -DLLVM_BUILD_INSTRUMENTED_COVERAGE and that seems to be doing what I want.
It does appear that $http_proxy is unset except on lockhart, and I'm going with that too.  (At OldJob, everybody had $http_proxy set to the corporate gateway just to get out, thus my caution.)
And I finally differentiated the archived files by target name.

yiqian1:
Have you tested it?

yiqian1:
Have you tested it?

pcf000:
Yes.  See http://rocmhead.amd.com:8080/job/MLIR/job/mlir-pf-test/108/ for the most recent successful test.  It builds with profiling on, runs the static-lib tests, merges the profile data, and produces a summary report and annotated source as both text and html.  It does the upload to codecov.io but the upload doesn't stick.  $http_proxy ends up unset on gfx908 and set properly on gfx90a, which means no upload errors.

mlir/utils/jenkins/Jenkinsfile
krzysz00: Would it make sense to move this into the existing PR build stages, or do those not fit?

pcf000: I want it as a separate stage for now, till we determine where we want to run it.  First step, according to Jerry, is just to produce *some* code-coverage numbers.

After that, there are some projects that monitor coverage on pull requests, and even verify that coverage doesn't decrease.  Others just tally it daily or weekly.

pcf000: And I see in the document that Manupa just posted that we want to enable it for pull requests.  I think it still makes sense to have as a separate stage, since the compilation options are different than usual.

pcf000:
I've made extensive changes and this pull request no longer applies.

PR #1319, opened 2023-11-15
Allow threadwise copy to accept extra source/dest indices

giuseros:
I had to extend `threadwise_copy` to accept extra indices. The issue with this operator is that it is the only one accepting views on both input/output (which means more troubles). I ended up "normalizing" both views. 

The normalization process is as follows:
- If input view is `[extraIndicesSource, inputView]`
- If output view is `[extraIndicesDest, outputView]'
- The normalized input view is `[extrIndicesSource, extraIndicesDest, inputView]` and the normalized output view is `[extraIndicesSource, extraIndicesDest, outputView]`. 

The "dummy" indices will be dealt directly inside the transforming loop (bottom of the transform stack). I had to do this normalization for two reasons
a) `trnsforming_for` views need accept the same number of variables (same stride and same bounds)
b) I cannot simply `AddDim` the spurious dimensiona, because that would hinder the invertibility of the transform (hence its vectorization).

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -1334,11 +1334,13 @@ def Rock_InWarpTransposeOp :
 }
 
 def Rock_ThreadwiseCopyOp :
-    Rock_Op<"threadwise_copy">,
+    Rock_Op<"threadwise_copy", [AttrSizedOperandSegments]>,
     AllElementTypesMatch<["source", "dest"]>,
     Arguments<(ins
       Arg<MemRefOf<SupportedMemoryElems>, "source view", [MemRead]>:$source,
+      Variadic<Index>:$extraIndicesSource,
       Arg<MemRefOf<SupportedMemoryElems>, "detination view", [MemWrite]>:$dest,
+      Variadic<Index>:$extraIndicesDest,
       UnitAttr:$forceUnroll,
       UnitAttr:$useIndexDiffs
     )> {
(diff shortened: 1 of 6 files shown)


mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1351,8 +1353,14 @@ def Rock_ThreadwiseCopyOp :
...
   let assemblyFormat = [{
-    attr-dict $source  `->` $dest `:` type($source) `->` type($dest)
+    attr-dict $source (`[` $extraIndicesSource^ `]`)? `->` $dest (`[` $extraIndicesDest^ `]`)? `:` type($source) `->` type($dest)
   }];
+
+  let builders = [
+    // Custom builder to not specify the extra indices
+    OpBuilder<(ins "Value":$source, "Value":$dest, CArg<"bool","false">:$forceUnroll, CArg<"bool","false">:$useIndexDiffs),
+                    [{ build($_builder, $_state, source, {}, dest, {}, forceUnroll, useIndexDiffs); }]>
+  ];
```
krzysz00, line 1363: This reminds me, do we want to, at some point, fix up threadwise_write_all to not take a `length` parameter but to instead take a slice of the input?

mlir/lib/Dialect/Rock/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -288,33 +288,59 @@ LogicalResult ThreadwiseCopyRewritePattern::matchAndRewrite(
...
   ArrayRef<int64_t> rawStoreBufferShape =
       rawStoreBuffer.getType().cast<ShapedType>().getShape();
-  if (rawLoadBufferShape.size() != 1)
-    return op.emitOpError("Raw load buffers have to be flat.");
-  if (rawStoreBufferShape.size() != 1)
-    return op.emitOpError("Raw store buffers have to be flat.");
+  if (rawLoadBufferShape.size() > extraIndicesSourceSize + 1)
+    return op.emitOpError("Raw load buffers have to be flat or multi buffers.");
+  if (rawStoreBufferShape.size() > extraIndicesDestSize + 1)
+    return op.emitOpError("Raw store buffers have to be flat or muti buffers.");
```
krzysz00: *multi buffers

krzysz00 (approved):
I don't really see any issues here, just typos

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
giuseros: You mean something like this https://github.com/ROCmSoftwarePlatform/rocMLIR-internal/issues/1260 ?

PR #1320, opened 2023-11-16
Use Tuna for tuning in rocMLIR weekly and nightly.

pcf000:
Add --tflops option to include that in the .tsv file for Tuna.
Rearrange options slightly in conv-configs.
Initial unified Tuna script for CI.
Add export of results.
Running mysqld in user mode.
Quiet some logging.  Adjust tuna-script.sh to fit into CI better. Ready or not, here come the Tuna versions of the main tuning runs. Add toCommandLine methods to canonicalise configs to look up in tuningDb. Add multi_gpu to gfx908 labels to avoid single-GPU host. Improved local-mysql startup in tuna-script.sh, given libaio1 in the docker image. Make sure PATH and TUNA_DB_HOSTNAME are set when mysqld is already up. Quick-tuning support in the Tuna script.
Back to a development branch for quick-tuning and backward-data support. Tuning-space and load-factor support in tuna-script.sh. Results-to-stdout support in tuningRunner.py.
Use a temp directory instead of separate filenames to avoid rocprof stats collisions. More robust chdir for verify, support for attention op in tuna-script.sh. And turn on Tuna attention support in Jenkinsfile.


mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -245,17 +245,17 @@ static llvm::cl::opt<int>
 
 /// Matrix options
 static llvm::cl::opt<int64_t> gemmM("m",
-                                    llvm::cl::desc("M dimennsion of gemm()"),
```
pcf000, line 248: Typos.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -198,7 +198,7 @@ String getLabelFromChip(String chip) {
         case "gfx906":
             return getLabelFromCodepath("vanilla")
         case "gfx908":
-            return "mlir && gfx908"
+            return "mlir && gfx908 && multi_gpu"
```
pcf000, line 202: Tuna's chief benefit is parallelism.  We have four gfx908s in the rocmhead pool, with 6, 4, 4, and 1 GPU respectively;  don't use the 1-GPU one.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -546,25 +546,25 @@ pipeline {
                             dir('build') {
                                 timeout(time: 60, activity: true, unit: 'MINUTES') {
                                     // Tune gemms, fail if the DB is not created
-                                    sh """python3 ./bin/tuningRunner.py --operation gemm \
-                                            --configs_file=../mlir/utils/jenkins/ci-configs/selected-gemm-configs \
-                                            --output=tuning_gemm.tsv
+                                    sh """../mlir/utils/tuna/tuna-script.sh -o gemm \
```
pcf000, line 553: Replace invocation of tuningRunner.py with use of tuna-script.sh, which will set up for Tuna and then do the same tuning.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -546,25 +546,25 @@ pipeline {
...
-                                            --configs_file=../mlir/utils/jenkins/ci-configs/selected-conv-configs \
-                                            --output=tuning_conv.tsv
+                                    sh """../mlir/utils/tuna/tuna-script.sh -o convolution \
+                                            -c ../mlir/utils/jenkins/ci-configs/selected-conv-configs \
+                                            -f tuning_conv.tsv
                                           [ -f tuning_conv.tsv ]"""
-                                    sh """python3 ./bin/tuningRunner.py --operation attention \
-                                        --configs_file=../mlir/utils/jenkins/ci-configs/selected-attention-configs \
-                                        --output=tuning_attention.tsv --verify-mode none
-                                        [ -f tuning_conv.tsv ]"""
```
pcf000, line 560: Existing attention tuning was checking for wrong file, hence the longer diff.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -712,36 +712,29 @@ pipeline {
                             buildProject('check-rocmlir-build-only ci-performance-scripts', '')
                             dir('build') {
                                 // Tune gemms with default datatypes, fail if the DB is not created
-                                sh """python3 ./bin/tuningRunner.py --operation gemm \
-                                        --configs_file=../mlir/utils/performance/gemm-configs \
-                                        --output=mlir_tuning_${ARCH}.tsv
+                                // (Includes int8xint8->int8 for performance comparisons against CK.)
```
pcf000, line 722: In other words, this one tuna-script.sh invocation accounts for both tuningRunner.py invocations at this point.  I chose to have configuration-generation make all the TxT->T cases, and special-cased i8 to make both i8xi8->i8 and i8xi8->i32.

mlir/utils/performance/conv-configs
```diff
@@ -1,43 +1,43 @@
 # This section of the file comes from resnet50-miopen-configs
--n 256 -c 1024 -H 14 -W 14 -k 2048 -y 1 -x 1 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -t 1
```
pcf000, line 2: Swapped "-m conv" and "-g 1" when I was trying to make all the input and output configs match, since perfRunner.py did a simple string-compare.  Eventually made toCommandLine and canonicalised them all at runtime.

mlir/utils/performance/perfRunner.py
```diff
@@ -170,15 +170,18 @@ def read_tuning_db(path: Optional[str]) -> MaybeTuningDb:
...
                 if len(entries) == 3:
                     arch, config, perfConfig = entries
                     ret[arch, config] = perfConfig
-
                 # note: new format has 4 entries
-                if len(entries) == 4:
+                elif len(entries) == 4:
                     arch, _, config, perfConfig = entries
                     ret[arch, config] = perfConfig
+                # note: 5-entry form includes tflops at end
```
pcf000, line 181: I wanted TFlops in the tuningRunner.py output so I could include it in reports.

mlir/utils/performance/perfRunner.py
```diff
@@ -420,6 +426,16 @@ def fromCommandLine(cls, argv, arch, numCU):
             convStrideH, convStrideW, paddingH, paddingW, dilationH, dilationW,
                    group, arch, numCU)
 
+    def toCommandLine(self):
```
pcf000, line 429: Canonicalise so we can reliably compare them.

mlir/utils/performance/tuningRunner.py
```diff
@@ -46,10 +49,8 @@ def verifyModeFlags(verifyMode: str) -> str:
 
 #Run a gemm or conv config and verify it
 def verifyKernelWithPerfConfig(perfConfig, config, paths: Paths, options: Options) -> float:
-    print(f"Verifying with perfConfig = {perfConfig}")
+    print(f"Verifying with perfConfig = {perfConfig}", file=sys.stderr)
```
pcf000, line 52: Changed all the running-commentary prints to print to stderr, so I could optionally send the results to stdout.

mlir/utils/performance/tuningRunner.py
```diff
@@ -46,10 +49,8 @@ def verifyModeFlags(verifyMode: str) -> str:
...
     config.setPerfConfig(perfConfig.strip())
-    # remove the result file generated by rocprof in previous benchmarking
```
pcf000, line 51: The next few changes have the verify execution write its results to a temporary directory, which will automatically go away at the end, to ensure that multiple threads don't step on each others' output.  We don't need to delete the last run's output here, because it already happened.

mlir/utils/performance/tuningRunner.py
```diff
@@ -59,42 +60,48 @@ def verifyKernelWithPerfConfig(perfConfig, config, paths: Paths, options: Option
...
-    # get output.
-    try:
-        outs, errs = p3.communicate(timeout=600)
-        outs = outs.decode('utf-8')
-        if len(errs) > 0 or not CORRECT_RESULT_RE.search(outs):
-            print(f"""Verification failed:
+        print(rocmlirGenCommand, file=sys.stderr)
+
+    prevdir = os.getcwd()
+    with tempfile.TemporaryDirectory() as tmpdir:
```
pcf000, line 66: Make a temp directory, cd to it, execute using the usual filename, extract the nanoSeconds.  The 'with' removes the directory, the 'finally' ensures that we always cd back where we came from.

mlir/utils/performance/tuningRunner.py
```diff
@@ -329,15 +352,27 @@ def main(args=None):
...
+        print(allData, file=sys.stderr)
         allData.to_csv(f"{parsed_args.output}.debug", sep='\t')
 
     # Note, appending results here to allow multiple config sets
-    with open(parsed_args.output, 'a') as outFile:
-        print(f"# arch\tnumCUs\ttestVector\tperfConfig ({options.tuningSpaceKind})", file=outFile)
-        for testVector, perfConfig in winners.items():
-            print(f"Arch = {arch}({numCU} CUs), vector = '{testVector}', perfConfig = {perfConfig}")
-            print(f"{arch}\t{numCU}\t{testVector}\t{perfConfig}", file=outFile)
+    if parsed_args.output == '-':
```
pcf000, line 359: Writing results to stdout is another way to avoid collisions when multithreaded -- there's no file, thus no chance of identical filenames.

mlir/utils/tuna/tuna-script.sh
```diff
@@ -0,0 +1,155 @@
+#!/bin/bash
```
pcf000, line 1: New file, to manage the Tuna sequence.

A stable persistent Tuna setup has a database running, Tuna installed, and database tables initialised.  At the moment, we don't have a persistent database, and thus have to do all the setup on each run.

Each run then imports configs, starts a session, sets up jobs, executes those jobs, and exports the results to a .tsv file for later perfRunner.py use.

mlir/utils/tuna/tuna-script.sh
```diff
@@ -0,0 +1,155 @@
...
+
+# For installing mysql 8.0 for testing, or for running with an isolated database.
+function mysql_setup_generic
+{
+    # Latest "generic" version.  (Must have libaio1 installed already.)
+    # Note that all this happens without privileges.
```
pcf000, line 6: rocMLIR CI jobs run as a normal user.  Embedding mysqld in our docker image is (a) large and (b) painful.  The pain comes from not having systemd in docker-world and then needing root to start things up.  Instead, we install the "generic" mysql, which doesn't care about systemd, and then arrange to run it as a normal user, in part by using a socket instead of dealing with network ports and authorisation.

mlir/utils/tuna/tuna-script.sh
```diff
@@ -0,0 +1,155 @@
...
+    # Using this name should force socket access, which we want.
+    TUNA_DB_HOSTNAME=localhost
+    mysql --user root -e 'create database tuna;'
+}
+
+function tuna_setup
+{
+    startdir=`pwd`
+    rm -rf MITuna
+    git clone --branch pf-tuna-rocmlir-3 https://github.com/ROCmSoftwarePlatform/MITuna.git
```
pcf000: Clone Tuna and do python things.  My development branch right now, rather than wait for the Tuna pull request.  A lot of python modules are installed, but at least it only happens once and they stick around for subsequent tuna-script.sh invocations in the same docker container.  Add our database tables at the end.

mlir/utils/tuna/tuna-script.sh
```diff
@@ -0,0 +1,155 @@
...
+    if pgrep mysqld ; then
+        ${TUNA_DIR}/tuna/go_fish.py rocmlir --add_tables
+    fi
+}
+
+function clear_tables
+{
+    tablekind=$1
+
+    if [ "$tablekind" = "convolution" ]; then
```
pcf000, line 34: I was a little sloppy about naming.  The other ops only have one variation.

mlir/utils/tuna/tuna-script.sh
```diff
@@ -0,0 +1,155 @@
...
+    mysql --user root --database tuna -e "delete from rocmlir_${tablekind}_results;"
+    mysql --user root --database tuna -e "delete from rocmlir_${tablekind}_job;"
+    mysql --user root --database tuna -e "delete from rocmlir_${tablekind}_config;"
+}
+
+function tuna_run
+{
+    kind=$1
+    space=$2
+    baselabel=`date --iso-8601=minutes`
```
pcf000: Do the work.  Unfortunately, the tables are differentiated by config-type, so we have separate tables and separate jobs and separate runs for conv, gemm, and attention.  (Maybe we fix that later.)  Label starts with a timestamp to know when these ran.  Configs are global, but other parts are held under the umbrella of a "session" which includes the config-type and tuning-space;  jobs belong to a session and results are looked up by session.  Load-factor is an idea to explore how much parallelism we get and how much non-GPU overhead there is in a tuning run.

mlir/utils/performance/perfRunner.py
```diff
@@ -911,7 +932,8 @@ def runConfigWithMLIR(config: PerfConfiguration, paths: Paths, rocmlir_gen_flags
 # Benchmarking function.
 def benchmarkMLIR(commandLine, confClass, paths: Paths, arch, numCU, tuningDb: MaybeTuningDb, rocmlir_gen_flags):
     config = confClass.fromCommandLine(commandLine, arch, numCU)
-    configStr = ' '.join(commandLine)
+    #configStr = ' '.join(commandLine)
```
krzysz00: Don't leave stray commented code in here?

mlir/utils/performance/tuningRunner.py
krzysz00: Makes sense - I assume this auto-deletes after?

mlir/utils/performance/tuningRunner.py
```diff
@@ -59,42 +60,48 @@ def verifyKernelWithPerfConfig(perfConfig, config, paths: Paths, options: Option
...
     return nanoSeconds
 
 def getWinningConfig(tuningOutput, config, allData, options: Options):
     maxTFlops = -np.inf
     winningConfig = "None"
     for i, result in enumerate(tuningOutput):
         result = result.decode('utf-8').strip()
-        if i > 0 and i % 100 == 0:
-            print(f"Tested {i} configs, best perf {maxTFlops} TFlops on perf_config {winningConfig}")
+        if not options.quiet and i > 0 and i % 100 == 0:
```
krzysz00, line 101: One note: Jenkins relies on scripts printing out to stderr/stdout regularly to enforce timeouts - are you sure you're not breaking this here?

mlir/utils/performance/tuningRunner.py
```diff
@@ -329,15 +352,27 @@ def main(args=None):
...
-        print(f"# arch\tnumCUs\ttestVector\tperfConfig ({options.tuningSpaceKind})", file=outFile)
-        for testVector, perfConfig in winners.items():
-            print(f"Arch = {arch}({numCU} CUs), vector = '{testVector}', perfConfig = {perfConfig}")
-            print(f"{arch}\t{numCU}\t{testVector}\t{perfConfig}", file=outFile)
+    if parsed_args.output == '-':
+        outFile = sys.stdout
+    else:
+        outFile = open(parsed_args.output, 'a')
+
+    with outFile:
```
krzysz00, line 364: Does this work?

mlir/utils/tuna/tuna-script.sh
```diff
@@ -0,0 +1,155 @@
+#!/bin/bash
+
+# For installing mysql 8.0 for testing, or for running with an isolated database.
+function mysql_setup_generic
+{
+    # Latest "generic" version.  (Must have libaio1 installed already.)
+    # Note that all this happens without privileges.
+    pushd /tmp
+    wget https://dev.mysql.com/get/Downloads/MySQL-8.0/mysql-8.0.34-linux-glibc2.28-x86_64.tar.gz
```
krzysz00: ... all this goes in the Dockerfile, perhaps?

mlir/utils/tuna/tuna-script.sh
```diff
@@ -0,0 +1,155 @@
...
+    tar xf mysql-8.0.34-linux-glibc2.28-x86_64.tar.gz
+    mv mysql-8.0.34-linux-glibc2.28-x86_64 mysql
+    export PATH=$PATH:/tmp/mysql/bin
+    mysqld --initialize-insecure --datadir=/tmp/mysql/data
```
krzysz00: More accurately, I'd say the wget should be in the Dockerfile, and then we should create the database, either under /tmp/ or, more interestingly, under $WORKSPACE.

mlir/utils/tuna/tuna-script.sh
krzysz00: Can we move this into the Jenkinsfile so that it picks up on any weird config that Jenkins does?

mlir/utils/tuna/tuna-script.sh
```diff
@@ -0,0 +1,155 @@
...
+    rm -rf MITuna
+    git clone --branch pf-tuna-rocmlir-3 https://github.com/ROCmSoftwarePlatform/MITuna.git
+    cd MITuna
+    export TUNA_DIR=`pwd`
+
+    python3 -m venv myvenv
+    source myvenv/bin/activate
+
+    # --ignore-installed because of problems upgrading PyYAML.  See also -U.
+    python3 -m pip install -r requirements.txt --ignore-installed
```
krzysz00: ... See above: why not do this in the Dockerfile?

mlir/utils/tuna/tuna-script.sh
krzysz00: `[[` here if possible

... heck, run `shellcheck` on this script just to be safe

krzysz00 (commented):
I'm not sure I like how much setup work is leaking into the actual CI run process here.

mlir/utils/tuna/tuna-script.sh
pcf000: I hesitate, because then any desired Tuna update will require a docker image re-generation.

pcf000:
> I'm not sure I like how much setup work is leaking into the actual CI run process here.

Same here, but I was trying to get things in faster.  Anyway, I split the installation parts of setup out of tuna-script.sh and added them as final steps in Dockerfile;  the latter part is a separate pull request, because it'll have to be merged before this pull request will pass the tuning parts.

mlir/utils/tuna/tuna-script.sh
krzysz00: Re-raise nitpick / request for a `shellcheck`

pcf000: Ran shellcheck, and it didn't care about [ vs [[, though it did point out some useful places to double-quote and also urged $() over ``.

krzysz00 (approved):
Looks good to me

pcf000:
Reviewers, please let me do the merge.  All the checks may pass for bogus reasons while I work out the clone-MITuna details.

PR #1321, opened 2023-11-20
Implement a pipeline pass in rocMLIR

giuseros:
This is a first stab at implementing a `rock.pipeline` pass to automate the composition of the pipeline. 

The kernel developer can now split a for-loop in multiple stages and set the InitiationInterval.

So for instance, we can do:
```
 scf.for %arg3 = %c0 to %c16 step %c1 {
    rock.stage {
      rock.threadwise_read_into {forceUnroll, useIndexDiffs} [](%2) [%arg3, %9, %16, %18, %6] -> %view
      rock.threadwise_read_into {forceUnroll, useIndexDiffs} [](%4) [%arg3, %9, %16, %18, %6] -> %view_0 
      rock.threadwise_copy %22 -> %24
      rock.threadwise_copy %26 -> %28
      rock.yield
    } {name = "GlobalRead"}
    rock.stage {
      rock.threadwise_write_all features =  %view_1 -> [](%37) [%6]
      rock.threadwise_write_all features = %view_2 -> [](%41) [%6] 
      rock.yield
    } {name = "LDSWrite"}
    rock.stage {
      affine.for %arg4 = 0 to 1 {
        rock.threadwise_read_into [](%53) [%46, %arg4] -> %view_7
        affine.for %arg5 = 0 to 1 {
          rock.threadwise_read_into [](%57) [%46, %arg5] -> %view_8
          rock.accel_gemm %view_9 += %view_7[%arg4] * %view_8[%arg5] 
        }
      }
      rock.yield
    } {name = "MMA"}
} {__initiation_interval__ = 1 : index}
```
This will generate a pipeline where each iteration will contain a single stage, like:
```
// Prolog
"GlobalRead"(0)
"GlobalRead"(1)
"LDSWrite"(0)
// Kernel
"GlobalRead"(2)
"LDSWrite"(1)
"MMA"(0)
// Epilog
"LDSWrite"(2)
"MMA"(2)
"MMA"(1)
```
**Note0**: Since `GlobaRead`, `LDSWrite` and `MMA` will all execute at the same time (but in different iteration) we need to transform simple buffers in multi-buffers. 

**Note1**: the support for nested loops is still missing

**Note3**: the support for tuning/heuristic is still missing. For now we only pipeline GEMM in the same way we used to. This means that II=2 and the stages are as described above.


mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1364,4 +1364,15 @@ def Rock_ThreadwiseCopyOp :
   let hasVerifier = 1;
 }
 
+
+def Rock_StageOp :
```
manupak, line 1368: docs please. 
Think of a kernel developer, who would want to use this op..

mlir/include/mlir/Dialect/Rock/Passes.td
```diff
@@ -78,6 +78,14 @@ def RockCleanMathPass : Pass<"rock-clean-math", "::mlir::func::FuncOp"> {
   let summary = "Clean up math using integer range analysis and other MLIR passes";
 }
 
+def RockPipelinePass : Pass<"rock-pipeline", "::mlir::func::FuncOp"> {
```
manupak, line 81: docs

mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
```diff
@@ -42,6 +42,10 @@ struct RegsAsMatrixSubTiles {
   ArrayAttr threadSubTile;
 };
 
+// Attribute name to set the initiation interval of a pipelined loop
+constexpr llvm::StringLiteral kInitiationIntervalAttrName =
```
manupak: Should we use something like ? 
https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/4a5ca1a75edda5cfa31f71f462eedb8ce5b6a437/external/llvm-project/mlir/include/mlir/Dialect/GPU/TransformOps/GPUDeviceMappingAttr.td#L51-L65

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -0,0 +1,438 @@
...
+#include "mlir/Dialect/Rock/Passes.h.inc"
+} // namespace rock
+} // namespace mlir
+
+#define DEBUG_TYPE "rock-pipeline"
+#define DBGS() (llvm::dbgs() << "[" DEBUG_TYPE "]: ")
+
+using namespace mlir;
+
+enum class MemoryAccessType : uint32_t { READ = 1, WRITE = 2, UNKNOWN = 3 };
```
manupak, line 50: Cant we re-use MemoryEffect / MemoryEffectOpInterface of upstream here ?

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -0,0 +1,438 @@
...
+    rw.eraseOp(op);
+    return failure();
+  }
+};
+
+// Given an operation and its operand, find out what kind of access (if any)
+// the operation does on the operand
+MemoryAccessType getOperandAccessType(Operation *op, Value operand) {
+  if (hasEffect<MemoryEffects::Read>(op, operand)) {
+    return MemoryAccessType::READ;
```
manupak: ok so you do use that... but why the indirection into a custom enum here ? you can keep using mlir::SideEffects ?
Also you can get all the effect(s) using getEffects()

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -0,0 +1,438 @@
...
+  for (auto res : resources)
+    multiBuffers[res] = 1;
+
+  // Start building the schedules
+  //
+  // Since we accept the stages from the user, we don't need to do any
+  // analysis to determine what goes in each stage. We only have to group things
+  // in set of stages of length II.
+  //  For instance, if this is the unpipelined schedule decided by the user
+  //  +t\s+=== 0 ===+
```
manupak: Im not following the notation here.. can you explain ?

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -0,0 +1,438 @@
...
+  // and S3. If we increase II, we get to :
+  //  +t\s+=== 0 ===++=== 1 ===+
+  //  + 0 +== S0  ==++== S2  ==+
+  //  +===+=========++=========+
+  //  + 1 +== S1  ==++== S3  ==+
+  //  +===+=========++=========+
+  // Now we have only two time slots and 2 sets of parallel stages. conflicts
+  // can happen between S0 and S2 and between S1 and S3. The above is capture in
+  // the following algorithm. `t` is the time slot, i.e., the flowing of the
+  // time, and goes from 0 to II. `stageSet` is the set of stages that run in
```
manupak: 0 to II -1

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -0,0 +1,438 @@
...
+
+// Prune a dependency graph taking into account multi-buffers. Since
+// multi-buffers are logically different for each iteration, if the dependency
+// on a multi-buffer spans multiple iteration then it can be pruned
+DagType pruneGraph(DagType dag, DenseMap<rock::StageOp, int> &iterationMap,
+                   DenseMap<rock::GpuAllocOp, int> &factors) {
+  DagType prunedGraph;
+  // Multibuffers have the logical property of being unique for each iteration
+  // of the loop Hence, if we know we are dealing with a multi-buffer and the
+  // dependency concerns two different iteration. In other woeds, if stageA
```
manupak: words

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -0,0 +1,438 @@
...
+  for (auto [sink, edges] : dag) {
+    for (auto [source, deps] : edges) {
+      DenseSet<std::pair<rock::GpuAllocOp, DependencyType>> newDeps;
+      for (auto [alloc, type] : deps) {
+        // Add the dependency only if it's not over a multiBuffer (factor==1) or
+        // if it's over a multibuffer, if sink and source share the same
+        // iteration
+        if (factors[alloc] == 1 ||
+            (factors[alloc] > 1 &&
+             ((iterationMap[sink] % 2) == (iterationMap[source] % 2)))) {
```
manupak: I think this 2 should be "II". It should not be hardcoded to 2 to just to match our current implementation.

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -0,0 +1,438 @@
...
+  DenseMap<rock::GpuAllocOp, DenseMap<rock::StageOp, MemoryAccessType>>
+      resourceMapR;
+
+  // For each of the stages walk through the resources they are using. For now
+  // the only type of resource is memory
+  for (auto stage : stages) {
+    stage.walk([&](Operation *op) {
+      for (Value operand : op->getOperands()) {
+        MemoryAccessType accessType = getOperandAccessType(op, operand);
+        auto maybeAlloc = rock::findAlloc(operand);
```
manupak, line 158: This can potentially fail as well ? what happens if so...

manupak (commented):
I did an initial pass.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1208,9 +1208,9 @@ def Rock_AccelGemmOp:
     Rock_Op<"accel_gemm">,
     Arguments<(ins Index:$mRepeat,
                    Index:$nRepeat,
-                   MemRefRankOf<AccelArgTypes, [1]>:$matrixA,
-                   MemRefRankOf<AccelArgTypes, [1]>:$matrixB,
-                   MemRefRankOf<AccelResTypes , [1]>:$matrixC,
+                   Arg<MemRefRankOf<AccelArgTypes, [1]>, "source register A", [MemRead]>:$matrixA,
+                   Arg<MemRefRankOf<AccelArgTypes, [1]>, "source register B", [MemRead]>:$matrixB,
+                   Arg<MemRefRankOf<AccelResTypes , [1]>, "source register C", [MemWrite]>:$matrixC,
```
krzysz00: I think that C gets updated by this op?

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
krzysz00: Seconded

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -235,6 +236,23 @@ static LogicalResult checkLDSSize(OpT op, int64_t aBufferBytes,
   return success(ldsBytes <= 64 * 1024);
 }
 
+Value gpuAlloc(OpBuilder &b, Location loc, int bufferDim, Type elementType,
```
krzysz00: Perhaps `int64_t` or `size_t` for `bufferDim`

Also, instead of a weird bool, maybe an enum?

(Heck, maybe importing the address space enum with some `using`s so you can just say `Workgroup` and `Private`)

mlir/lib/Dialect/Rock/Transforms/RockMultibuffer.cpp
```diff
@@ -351,6 +337,28 @@ mlir::rock::multiBuffer(RewriterBase &rewriter, rock::GpuAllocOp allocOp,
...
+    return failure();
+  }
+  if (allocOp.getType().getShape().size() != 1) {
+    LLVM_DEBUG(DBGS() << "-- Not a flat buffer -> fail\n");
+    return failure();
+  }
+
+  bool isUsedByViews = true;
+  for (auto user : allocOp->getUsers()) {
+    if (!dyn_cast<memref::ViewOp>(user)) {
```
krzysz00: This could be an `llvm::all_of`

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
krzysz00: There might be other non-memory effects in the future, I think

krzysz00 (commented):
I haven't read the pipeline pass itself thoroughly, being low on time for doing this today, but I've got minor notes on the surrounding changes

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
giuseros: I think that having a simple `enum` has its benefits, mostly because enums are easier to manipulate. I can store the enum in a pair and define `dependencytype=pair<AccessType, AccessType>`, so that a RAR access can be expressed like `constexpr DependencyType RAR{MemoryAccessType::READ, MemoryAccessType::READ};` . I can easily compare enums, etc... I think I can find ways to do these things with the `mlir::SideEffects` , but it would make the code more complicated to read.

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
giuseros: Yes, sorry, I forgot to update. However, I was planning to replace that condition with a simpler `(iterationMap[sink] == iterationMap[source])`, because in theory if the LDS accesses happen in different logical loops, they should not interfere(right?)

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
giuseros: I tried to make better comments. The idea is to mimic the stages of the pipelines, with different iterations. The non-ascii version would look like:
image

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
giuseros: Well I check if `succeeded(maybeAlloc)` to add the dependency. So if I was not able to find the allocation, I am not adding the dependency.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
giuseros: You mean the operand is read and written? I don't think there is a `MemReadWrite` among the effects, but for my purposes, if something is read and the written, the `write` side effect is the one that matters.

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
giuseros: More discussion in the [correlated comment](https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/1321#discussion_r1400369639)

manupak (commented):
I dont have a major concern here..

However, I d like a simpler test or two to demo the pipeline.
Also for testing (just unit test covering the pass), can you add the following :
1) II=1 pipeline with #stages=3 or 4
2) II=#stages a.k.a unpipelined; this should work.
3) II=2 with #stages=4 -- here I want to see double buffering done stage 2 & 3

Also in the current pipeline, so we need a barrier to ensure stage 3 doesnt overlap with stage 2 to avoid a multi-buffer right? How is this inferred ? 
(As in between a barrier vs multi-buffer)

manupak:
Above can be simple global load , LDS storing , LDS read simple compute (elemwise LA generic blocks are fine for e.g.)

giuseros:
Hi @manupak , 
I tried to fix the barrier placement problem we discussed. I basically add dummy stages between all the real stages (plus an additional one after the last stage). Those dummy stages can contain barriers or not. I then double the II and create the schedule as usual. This works with the exception that I need to take care of back-to-back barriers and I need to remove the last barrier that does not belong to the loop. 

Also, I updated the test with simpler examples, so that it is more clear what the pass does.

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -0,0 +1,492 @@
...
+      return WalkResult::advance();
+
+    bool isNestedPipelining = false;
+    forOp.getBody()->walk([&](scf::ForOp nestedFor) {
+      if (nestedFor->hasAttr(rockPipelineAttrName))
+        isNestedPipelining = true;
+    });
+
+    if (isNestedPipelining)
+      return WalkResult::advance();
```
manupak: I think we should error out for now rather than silently ignoring.
So the user would not be mislead to believe the pipelining semantics the user provided are legit.

manupak (commented):
Thanks for the tests!
it is much clearer; I have one last comment.

manupak (approved):
LGTM!

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
krzysz00: I claim that the side effects should be `[MemRead, MemWrite]` then.

`MemWrite` is a pure store operation.

giuseros: Ah I see. For some reason, I thought that `MemWrite` meant `MemReadWrite`. Thanks!

PR #1323, opened 2023-11-24
Append quick tuning list for attention kernels

manupak:
This commit append quick tuning list with known
performant configs.


giuseros (approved):
LGTM!

PR #1328, opened 2023-11-27
Update tuning key support for fp8 convolutions

krzysz00:
Work in MIGraphX reveleased that our tuning key generator will fail (return -1) for fp8 data types. This commit fixes the issue in a rather short-term way.


manupak (approved):
LGTM

PR #1332, opened 2023-11-28
[DO NOT SQUASH] [mlir][ArithToAMDGPU] Add option for saturating truncation to fp8

krzysz00:
Many machine-learning applications (and most software written at AMD) expect the operation that truncates floats to 8-bit floats to be saturatinng. That is, they expect `truncf 256.0 : f32 to f8E4M3FNUZ` to yield `240.0`, not `NaN`, and similarly for negative numbers. However, the underlying hardware instruction that can be used for this truncation implements overflow-to-NaN semantics.

To enable handling this usecase, we add the saturate-fp8-truncf option to ArithToAMDGPU (off by default), which causes the requisite clamping code to be emitted. Said clamping code ensures that Inf and NaN are passed through exactly (and thus trancate to NaN).

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Conversion/Passes.td
+++ b/external/llvm-project/mlir/include/mlir/Conversion/Passes.td
@@ -105,6 +105,12 @@ def ArithToAMDGPUConversionPass : Pass<"convert-arith-to-amdgpu"> {
   }];
 
   let dependentDialects = ["amdgpu::AMDGPUDialect", "vector::VectorDialect"];
+
+  let options = [
+    Option<"saturateFP8Truncf", "saturate-fp8-truncf", "bool",
+           /*default=*/"false",
+           "Whether truncation to 8-bit float types should be saturating">,
+  ];
 }
 
 //===----------------------------------------------------------------------===//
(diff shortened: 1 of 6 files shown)


umangyadav:
can you add unit-tests somewhere to make sure clipping behaves as expected ?

krzysz00:
I could, in theory, add an integration test to see if that code does what I think it does, but all the other related code is just tested via this sort of lit test, which is generally sufficient.

external/llvm-project/mlir/include/mlir/Conversion/Passes.td
```diff
@@ -105,6 +105,12 @@ def ArithToAMDGPUConversionPass : Pass<"convert-arith-to-amdgpu"> {
   }];
 
   let dependentDialects = ["amdgpu::AMDGPUDialect", "vector::VectorDialect"];
+
+  let options = [
+    Option<"saturateFP8Truncf", "saturate-fp8-truncf", "bool",
+           /*default=*/"false",
```
umangyadav, line 111: is the default "False" ? 
I am mostly concerned about default for MIGraphX

krzysz00: For MIGraphX we set it to true, if you notice the changes to `Pipelines.cpp`

I'm defaulting it to false upstream for back-compat

umangyadav (approved):
lgtm.

sjw36 (approved):
Looks good.

PR #1333, opened 2023-11-28
Moving some setup from tuna-script.sh to Dockerfile (Dockerfile part)

pcf000:
Moved the installation of mysql and the Tuna-required python packages into Dockerfile from tuna-script.sh.

### DIFF
--- a/mlir/utils/jenkins/Dockerfile
+++ b/mlir/utils/jenkins/Dockerfile
@@ -130,6 +130,7 @@ RUN wget https://dev.mysql.com/get/Downloads/MySQL-8.0/mysql-8.0.34-linux-glibc2
 
 # python setup for tuna
 # --ignore-installed because of problems upgrading PyYAML.  See also -U.
-RUN git clone --branch pf-tuna-rocmlir-3 https://github.com/ROCmSoftwarePlatform/MITuna.git && \
-    python3 -m pip install -r MITuna/requirements.txt --ignore-installed && \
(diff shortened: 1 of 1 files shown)


mlir/utils/jenkins/Dockerfile
```diff
@@ -121,3 +121,15 @@ RUN apt-get update && rm /usr/share/keyrings/kitware-archive-keyring.gpg && \
...
+# mysql setup for tuna
+# Latest "generic" version.  (Must have libaio1 installed already.)
+RUN wget https://dev.mysql.com/get/Downloads/MySQL-8.0/mysql-8.0.34-linux-glibc2.28-x86_64.tar.gz && \
+    tar xf mysql-8.0.34-linux-glibc2.28-x86_64.tar.gz && \
+    mv mysql-8.0.34-linux-glibc2.28-x86_64 /usr/mysql
+
+# python setup for tuna
+# --ignore-installed because of problems upgrading PyYAML.  See also -U.
+RUN git clone --branch pf-tuna-rocmlir-3 https://github.com/ROCmSoftwarePlatform/MITuna.git && \
+    python3 -m pip install -r MITuna/requirements.txt --ignore-installed && \
```
krzysz00: 1. On the other hand, we usually put git clones in the Jenkinsfile
2. I suspect tuna may have different version requirements than, say, our performance script. Because of that, would you be willing to make this a `wget` of the `requirements.txt` and an installation into some sort of virtualenv?

krzysz00:
Oh, and, perhaps `/opt/mysql` for linux-ness or maybe even `/mysql` because we're in a container anyway?

krzysz00 (approved):

PR #1334, opened 2023-11-29
Subtree update for 6.1

krzysz00:
This required (probably because it's a small bit of catching up) very little changes on our end - one header moved and one function deprecated.

This is our first subtree merge with amd-mainline-open as opposed to llvm-project.

It's exposed some failures of the merge process for AMO, which I'm following up with for the compiler team while hacking-around them here.

We're not quite in a position where we can use comgr because of that hack for the s_clause hang.

The commits after the first one (the actual subtree merge) are small and should be easily reviewable.


sjw36 (approved):
Looks good. Let's also verify no perf regressions.

krzysz00:
@sjw36 I'm currently doing a nightly run on this PR and will compare those numbers against the ones from Giuseppe's PR fixing up pipelining or last night or such

krzysz00:
Perf is broadly flat

PR #1335, opened 2023-11-30
Fix barrier placement during the pipeline pass.

giuseros:
This is fixing the regression we are experiencing here: http://rocmhead.amd.com:8080/job/MLIR/job/mlir-nightly-all/1011/Performance_20report_20for_20gfx90a/

The high level idea of the fix is to instrument the pipeline so that the **backward lds barrier** is going to be at the beginning of the pipelined loop and not at the end. In this way we can push it down hosting non-LDS operation before barriering and we can have the global loads happening after the LDS writes, giving them time to hide their latency. 

While this "works", the choice of a different pipeline structure should be the subject of an analysis (but let's worry about that later :) ). 

I am attaching a comparison with the original performance from commit 27c0a1fb76df805440ed63800c593dc2aec3765f

[comparison.zip](https://github.com/ROCmSoftwarePlatform/rocMLIR/files/13516805/comparison.zip)

As we can observe it is mostly flat, with few spikes that are just glitches (I was not able to reproduce the only "red" we have on the report.

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
@@ -94,36 +94,6 @@ AddressSpace getAddressSpace(rock::GpuAllocOp alloc) {
   return gpu::AddressSpace::Global;
 }
 
-// Simple rewrite pass to remove the stages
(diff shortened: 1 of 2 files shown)


krzysz00 (approved):
Looks fine to me. I mean, we can work out something a bit more principled later, maybe, but it looks OK

PR #1336, opened 2023-12-03
allow quant_conv to take fp8 types and generate FP32 output

umangyadav:


krzysz00 (approved):
Approved, though you'll probably want to make sure this lowers through Tosa acceptably

PR #1337, opened 2023-12-04
Hoist the multi-buffer broadcast from the transform stack

giuseros:
We should hoist the `Broadcast` out of the multi-buffer transform stack, in this way we won't affect the stack invertibility. THis means that the `Broadcast` transform becomes a `i%multiBufferFactor` affine map.

### DIFF
--- a/mlir/test/Dialect/Rock/test_multi_buffer.mlir
+++ b/mlir/test/Dialect/Rock/test_multi_buffer.mlir
@@ -1,4 +1,5 @@
 // RUN: rocmlir-opt %s --rock-multibuffer-test | FileCheck %s
+// RUN: rocmlir-opt %s --rock-multibuffer-test --rock-blockwise-gemm-to-threadwise --rock-threadwise-gemm-lowering | FileCheck %s --check-prefix=LOWERING
 #map = affine_map<(d0, d1, d2) -> (d0, d2, d1)>
 #map1 = affine_map<(d0, d1, d2, d3, d4, d5, d6, d7) -> (d1, (d0 * 8 + d5) * 8 + d7, (d2 * 32 + d4) * 2 + d6)>
 #map2 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d1, d2, d3, d4 floordiv 8, d4 mod 8, d5 floordiv 8, d5 mod 8)>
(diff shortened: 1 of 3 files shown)


manupak (commented):
Code and the idea on high-level make sense to me!
Do you have a simple unit test to demo this ?

giuseros:
Yep, you are right. I amended the tests and added a check to make sure that the lowering to  `transforming_for`s is now producing the right (vectorized) strides

mlir/test/Dialect/Rock/test_multi_buffer.mlir
```diff
@@ -73,7 +74,11 @@ module attributes {mhal.arch = "amdgcn-amd-amdhsa:gfx90a"} {
     %18 = arith.divui %17, %14 : index
     rock.threadwise_read_into {forceUnroll, useIndexDiffs} [](%2) [%c0, %9, %16, %18, %6] -> %7 : memref<16x1x6x16x256x16xf16> -> memref<16xf16, #gpu.address_space<private>>
     rock.threadwise_read_into {forceUnroll, useIndexDiffs} [](%4) [%c0, %9, %16, %18, %6] -> %8 : memref<16x1x6x16x256x16xf16> -> memref<16xf16, #gpu.address_space<private>>
-    %19 = rock.alloc() : memref<16xf16, #gpu.address_space<private>>
+    %c0_0 = arith.constant 0 : index
+    // LOWERING: %[[multibuf0:.*]] = rock.alloc() : memref<64xi8, #gpu.address_space<private>>
+    %raw = rock.alloc(){__multibuffer__=2} : memref<32xi8, #gpu.address_space<private>>
```
krzysz00, line 79: Nit: Can you make this an inherent attribute on `rock.alloc` so it's harder to forget about?

krzysz00 (approved):
Seems fine

mlir/test/Dialect/Rock/test_multi_buffer.mlir
giuseros: We chatted about that, and the point is that `__multibuffer__` attribute is only used by the `--rock-multibuffer-test` pass.

manupak (approved):
LGTM!

PR #1338, opened 2023-12-05
Fix CK build steps in CI

giuseros:
This PR should bring CK build process back to life in our CI. Main changes:
- In the `Jenkinsfile` I build all CK now[. This is the suggested way to build](https://github.com/ROCmSoftwarePlatform/composable_kernel), and I think we should stick to it
- In the `CMakeLists.txt`  I had to specify a `device_gemm_operations` component for the `ck-benchmark-driver` target. Unfortunately, I was not able to add all the components to the target. It looks like they don't export a `composable_kernel_LIBS` or `composable_kernel_INCLUDE_DIRS`. Cmakers better than me can have an opinion on how to do that in a better way (cc @sjw36 , @yiqian1 , @krzysz00 , @manupak )

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -22,7 +22,7 @@ void buildCK(String cmakeOpts) {
                       -DCMAKE_C_COMPILER=/opt/rocm/llvm/bin/clang
                      ${cmakeOpts}
                      """
-    sh 'cd build; make -j $(nproc) device_operations'
+    sh 'cd build; make -j $(nproc)'
 }
 
 void buildMIGraphX(String cmakeOpts) {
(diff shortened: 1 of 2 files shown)


krzysz00 (approved):

krzysz00:
@giuseros I think the solution is to ask CK to set the variables you want or otherwise define a target

(I'll note that depending on a target often pulls in the libraries and include directories, so you don't need to add them by hand)

giuseros:
> (I'll note that depending on a target often pulls in the libraries and include directories, so you don't need to add them by hand) 

Do you mean I might not need this? `set(LIBS composable_kernel::device_gemm_operations hip::device benchmark-driver-utils)` . Because I tried to remove it, but  it didn't work

krzysz00:
Ah, no, yeah, that' you'll still need

But the point is that, having that target in `LIBS` is enough to set up all the include directories correctly

PR #1339, opened 2023-12-05
Fix errors in MIGraphX fusion tests

krzysz00:
When converting some nightly end-to-end tests, I mistakenly make the filter arguments to the convolutions in cases 3 and 5 breadcaseted tensors in the argument list (in addition to making the inttended bias tensor a broadcasted argument). This led to IR-level type mismatches, which caused test failures.

This commit resolves the issue.


giuseros (approved):
LGTM!

ravil-mobile (approved):
LGTM

PR #1341, opened 2023-12-06
enable FP8 for the quantizelinear and dequantizelinar

umangyadav:
With these changes, 
and MIGraphx branch "mobilenet_fp8" which has changes for the simplify_qdq changes to work with FP8 in MIGraphx, 
I am able to run mobilenet in FP8.

### DIFF
--- a/mlir/include/mlir/Dialect/MIGraphX/IR/MIGraphX.td
+++ b/mlir/include/mlir/Dialect/MIGraphX/IR/MIGraphX.td
@@ -196,9 +196,9 @@ def MIGraphX_QuantizeLinearOp :
 
 def MIGraphX_DeQuantizeLinearOp :
     MIGraphX_Op<"dequantizelinear", [AllElementTypesMatch<["scale", "output"]>]>,
-    Arguments<(ins MIXRShapedOf<[AnyInteger, F8E4M3FNUZ, F8E5M2FNUZ]>:$input,
+    Arguments<(ins MIXRShapedOf<[AnyInteger, AnyFloat, F8E4M3FNUZ, F8E5M2FNUZ]>:$input,
                    MIXRShapedOf<[AnyFloat]>:$scale,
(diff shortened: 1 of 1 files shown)


mlir/include/mlir/Dialect/MIGraphX/IR/MIGraphX.td
```diff
@@ -196,9 +196,9 @@ def MIGraphX_QuantizeLinearOp :
 
 def MIGraphX_DeQuantizeLinearOp :
     MIGraphX_Op<"dequantizelinear", [AllElementTypesMatch<["scale", "output"]>]>,
-    Arguments<(ins MIXRShapedOf<[AnyInteger]>:$input,
+    Arguments<(ins MIXRShapedOf<[AnyInteger, F8E4M3FNUZ, F8E5M2FNUZ]>:$input,
```
umangyadav: This needs to be `AnyFloat` I see instances where `float` is the input to dequantizelinear, Which is different compared to `onnx` though which doesn't allow float32 for input. 
e.g. 
https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/a09dc50215f91c5ec2c0d6b2afd9fa0ed45f931c/src/simplify_qdq.cpp#L186 

here input to dequantizelinear would be float for the `fp8` convolution/gemms.

krzysz00:
This isn't sufficient - you'll need to go add lowerings for `dequantizelinear` and `quantizelinear` over in `mlir/lib/Conversions/MIGraphXToTosa/MIGraphXToTosa.cpp` to make this work 

and a test for said lowerings

krzysz00:
(We can chat and I can give you a sense of what you'll need to do)
(Or we can take over the ticket)

umangyadav:
> (Or we can take over the ticket

Better if you can take over, if it doesn't take much of your time.

umangyadav:
superseded by #1342

PR #1344, opened 2023-12-07
Perform threadwise reductions first within blockwise reductions

manupak:
This commit makes the blockwise reductions to perform
threadwise reductions first within the thread and only
store the partial reductions to LDS for further inter thread
reductions.

closes : https://github.com/ROCm/rocMLIR-internal/issues/982


mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
```diff
@@ -40,6 +40,10 @@ struct RegsAsMatrixSubTiles {
   // Using just a iterative dix, this provides access to threadwise sub-tile
   // of the matrix.
   ArrayAttr threadSubTile;
+  // This is a [tid] to a 2D subtile view.
+  // It essentially provides access to as if sliced blockSubTile
```
krzysz00: Grammar and I'm still not entirely sure what this is.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -788,27 +978,57 @@ struct BlockwiseReduceRewritePattern
         lowerTr.getLowerBounds().asArrayRef();
     SmallVector<int64_t, 4> regTensorShape =
         llvm::to_vector<4>(lowerTrLowerBounds);
+    int64_t nonReductionDimSizeProduct =
+        calculateNonReductionDimProduct(regTensorShape, axis);
+
+    // 2DView is alwasy nrDim x rdim
+    constexpr size_t nrDim = 0;
+    constexpr size_t rDim = 1;
+    mlir::ArrayAttr inputThreadSubTile2dView =
```
krzysz00: I don't think you need the `mlir::` here

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -788,27 +978,57 @@ struct BlockwiseReduceRewritePattern
...
+        createInput2DView(loc, rewriter, op.getIterSubTileSliceView(), axis);
+    ArrayRef<int64_t> inputThreadSubTile2dShape =
+        getLowerShape(inputThreadSubTile2dView);
+    auto partialRedBufferType =
+        MemRefType::get(inputThreadSubTile2dShape[nrDim], elemType, AffineMap{},
+                        privateMemoryAddressSpace);
+    Value partialRedBuffer =
```
krzysz00: At the very least, "Read",

krzysz00 (commented):
I've got naming nitpicks ... but also the more fundamental question: what is this thread-wise to 2D tile thing for?

manupak:
> but also the more fundamental question: what is this thread-wise to 2D tile thing for?

If you mean [tid] --> 2D subtile, it is there to get the (strided) slice that changes with thread id.
i.e. dim1 x dim2 of that subtile == workgroup size. 
It is required to obtain a unique address in LDS to store each thread's partial reductions.

mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
manupak: I ve changed it along the lines of my recent answer (see below)

pfultz2:
> This commit makes the blockwise reductions to perform
threadwise reductions first within the thread and only
store the partial reductions to LDS for further inter thread
reductions.

If by threadwise you mean per lane(and not per wavefront), then it should load it into registers not LDS. And then you would reduce in the wavefront and then load that result into LDS. If the reduction size is smaller than or equal to the wavefront size, you can skip the LDS and just use `readlane` to get the result.

manupak:
wavewise optimization is a different ticket and it is not the scope here.

manupak:
I want to do it as a next step ; so we can clearly what amount of benefit it brings. 
Also Im not sure what readlane is ? is it an intrinsic ?

> If by threadwise you mean per lane(and not per wavefront), then it should load it into registers not LDS

Its already in the registers -- i.e. there is no loading into it -- its done by whatever subkernel produces result (i.e. mfma).
But yea, for inter-thread reductions we do it via LDS which could be optimized further by doing DPP in principle.

mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
```diff
@@ -40,6 +40,10 @@ struct RegsAsMatrixSubTiles {
   // Using just a iterative dix, this provides access to threadwise sub-tile
   // of the matrix.
   ArrayAttr threadSubTile;
+  // This is a [tid] to a 2D subtile view.
+  // i.e. [tid] --> [m_tid, n_tid]
+  // where |m_tid| x |n_tid| == workgroup size.
```
krzysz00, line 45: Add: "It is equivalent to removing all `iter`-dependent components from `blockSubTile`."

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -1911,11 +1925,15 @@ LogicalResult BlockwiseBroadcastReduceOp::verify() {
     return emitError("workspace LDS buffer should be flat");
   }
 
-  int64_t blockwiseInputTensorElements = 1;
-  for (int64_t dimSize : inputTensorShape) {
-    blockwiseInputTensorElements *= dimSize;
+  int64_t blockwiseInputPartRedTensorElements = 1;
+  for (auto [dim, dimSize] : llvm::enumerate(inputTensorShape)) {
+    if ((int64_t)dim == axis) {
+      blockwiseInputPartRedTensorElements *= inputPartRedTensorShape[axis];
```
krzysz00: Ah 'PartRed' -> 'PartialReduction'

krzysz00 (approved):
Overall, seems fine, barring documentation nitpicks

pfultz2:
> Also Im not sure what readlane is ? is it an intrinsic ?

Well the `v_readlane_b32` hardware instruction, but I think there is the [`__builtin_amdgcn_readlane`](https://github.com/ROCm/llvm-project/blob/f986706166c95f07b8fb1800b1d9346223b991c1/llvm/include/llvm/IR/IntrinsicsAMDGPU.td#L1979) intrinsic which will map to that instruction.

When doing reduction using DPP,  the last lane will have the reduction result, so readlane is used to get the last lane across all lanes. Plus, no barriers or synchronization is needed for this since it happens in the same wavefront.

PR #1346, opened 2023-12-11
Fixes a bug in attention padding

manupak:
This commit effectively fixes two things:
1) There was a bug in interpreting coordinates for padding in attention kernels
2) There was a wrong bound in init loop for reductions.

This fixes all that.
closes :  https://github.com/ROCm/rocMLIR-internal/issues/1281


sjw36 (approved):
LGTM!

krzysz00 (approved):
Looks good to me

PR #1351, opened 2023-12-13
[DO NOT SQUASH]Add support to group conv2d

yiqian1:
-Add a "group" attribute to tosa_conv2d op.

 -Preserve the group attribute in the migraphx-to-tosa pass

 -Apply rock.transform to 4-D tensors in the tosa-to-rock pass

 -Add two new linalg ops for grouped conv2d

 -Update test cases


external/llvm-project/mlir/test/Conversion/TosaToLinalg/tosa-to-linalg-named.mlir
```diff
@@ -511,7 +498,7 @@ func.func @conv2d_padded_f32(%input: tensor<1x47x40x28xf32>, %weights: tensor<28
   // CHECK: %[[C0:.+]] = arith.constant 0
   // CHECK: tensor.pad %arg0 low[0, 1, 1, 0] high[0, 1, 1, 0]
   // CHECK:   tensor.yield %[[C0]]
-  // CHECK: linalg.conv_2d_nhwc_hwcf
+  // CHECK: linalg.conv_2d_nhwgc_gfhwc
```
krzysz00: Yeah, this shouldn't be changed

mlir/lib/CAPI/Dialect/Rock.cpp
```diff
@@ -132,6 +133,9 @@ bool mlirRockTuningSetFromTable(MlirRockTuningTable perfTable,
 MLIR_CAPI_EXPORTED size_t mlirRockTuningGetKey(MlirModule module, char *buf,
                                                size_t bufLen) {
   auto mod = unwrap(module);
+  llvm::outs() << "in mlirRockTuningGetKey\n";
```
krzysz00: Debug print left in

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -67,15 +68,64 @@ static Value expandTensor(ConversionPatternRewriter &rw, Operation *op,
   }
   ArrayRef<int64_t> shape = oprType.getShape();
 
+  // operand.dump();
+  // llvm::outs()<<"\n";
```
krzysz00: Don't just leave commented out debug statements here. If you want to keep them, LLVM_DEBUG exists

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -699,7 +699,7 @@ LogicalResult GemmOp::verify() {
   if (inElems.isa<FloatType>() && !outElems.isa<FloatType>())
     return emitOpError(
         "float-valued inputs must have a floating-point output type");
-
+  // dump();
```
krzysz00: Stray debug prints again

external/llvm-project/mlir/test/Conversion/TosaToLinalg/tosa-to-linalg-named.mlir
```diff
@@ -430,14 +413,16 @@ func.func @conv2d_f32(%input: tensor<1x49x42x27xf32>, %weights: tensor<28x3x3x27
...
+  // CHECK: %[[INPUT:.+]] = "tosa.reshape"(%arg0) <{new_shape = array<i64: -9223372036854775808, 49, 42, 1, 27>}> : (tensor<?x49x42x27xf32>) -> tensor<?x49x42x1x27xf32>
+  // CHECK: %[[WEIGHT:.+]] = "tosa.reshape"(%arg1) <{new_shape = array<i64: 1, 28, 3, 3, 27>}> : (tensor<28x3x3x27xf32>) -> tensor<1x28x3x3x27xf32>
   // CHECK: %[[M_IN:.+]] = tensor.empty(%[[BATCH]])
   // CHECK: %[[CST:.+]] = arith.constant 0
   // CHECK: %[[FILL:.+]] = linalg.fill
-  // CHECK: %[[B_IN:.+]] = tensor.empty(%[[BATCH]])
-  // CHECK: %[[CONV:.+]] = linalg.conv_2d_nhwc_hwcf {dilations = dense<[2, 1]> : tensor<2xi64>, strides = dense<1> : tensor<2xi64>} ins(%arg0, %[[W]] : tensor<?x49x42x27xf32>, tensor<3x3x27x28xf32>) outs(%[[FILL]] : tensor<?x45x40x28xf32>)
-  // CHECK: %[[B:.+]] = linalg.generic {indexing_maps = [#[[$MAP1]], #[[$MAP2]], #[[$MAP2]]], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%arg2, %[[CONV]] : tensor<28xf32>, tensor<?x45x40x28xf32>) outs(%[[B_IN]] : tensor<?x45x40x28xf32>)
+  // CHECK: %[[OUT:.+]] = "tosa.reshape"(%[[FILL]]) <{new_shape = array<i64: -9223372036854775808, 45, 40, 1, 28>}> : (tensor<?x45x40x28xf32>) -> tensor<?x45x40x1x28xf32>
```
krzysz00: You probably shouldn't change these

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -67,15 +68,64 @@ static Value expandTensor(ConversionPatternRewriter &rw, Operation *op,
...
+  // }
+  // llvm::outs()<<"\n";
+  // oprType.dump();
+
   SmallVector<uint32_t, 8> endDims;
   SmallVector<uint32_t, 8> startDims;
+  SmallVector<StringRef, 8> startNames; // SmallString
+
+  // find the lower dimension that encodes the g dimension
+  uint32_t lowerIdx;
```
krzysz00: Slightly more descriptive variable names.

Also, instead of `bool found`, consider `std::optional<uint32_t> groupFoldedDim` or something like that.

And `lowerStride` could be some other name?

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -67,15 +68,64 @@ static Value expandTensor(ConversionPatternRewriter &rw, Operation *op,
...
+    if (layout[i] == lowerName[0]) {
+      lowerIdx = i;
+      found = true;
+    } else { // if ( layout[i] != "g" ) {
+      startDims.push_back(i);
+      endDims.push_back(found ? i + 1 : i);
+    }
   }
-  rock::BottomUpTMBuilder transform(rw, shape, loc);
+  lowerStride = g;
```
krzysz00: ... why define the variable up top instead of defining it here?

mlir/test/Conversion/MIGraphXToTosa/migraphx-to-rock-conv-nhwc.mlir
```diff
@@ -6,7 +6,7 @@
 // CHECK-SAME: memref<1x3x3x7xf32>
 func.func @convNHWC(%in: !migraphx.shaped<1x4x5x5xf32, 100x1x20x4>, %fil: !migraphx.shaped<7x4x3x3xf32, 36x1x12x4>) -> !migraphx.shaped<1x7x3x3xf32, 63x1x21x7> attributes {kernel, arch = "gfx1100", num_cu = 48 : i64} {
   // CHECK: rock.conv2d
-  // CHECK-SAME: filter_layout = ["k", "y", "x", "c", "g"]
+  // CHECK-SAME: filter_layout = ["k", "g", "y", "x", "c"]
```
krzysz00: Is this right? I could've sworn filters fold the group into the `k` dimension, so this should be `["g", "k", "c", "y", "x"]`

mlir/test/Conversion/TosaToRock/no-mixed-layout.mlir
```diff
@@ -1,6 +1,6 @@
 // RUN: rocmlir-driver -host-pipeline partition,highlevel -targets %arch %s | FileCheck %s
 
-// CHECK: rock.conv2d({{.*}}) {{.*}} {{{.*}}, filter_layout = ["k", "c", "y", "x", "g"], input_layout = ["ni", "ci", "hi", "wi", "gi"], output_layout = ["no", "ko", "ho", "wo", "go"]{{.*}}}
+// CHECK: rock.conv2d({{.*}}) {{.*}} {{{.*}}, filter_layout = ["k", "g", "c", "y", "x"], input_layout = ["ni", "ci", "gi", "hi", "wi"], output_layout = ["no", "ko", "go", "ho", "wo"]{{.*}}}
```
krzysz00: I think you've got this exactly backwards - it should be `NGCHW` not `NCGHW`

krzysz00 (changes requested):
I'm blocking pending confirmation that you've got the convolution layouts right here - the ones in the Rock tests don't match the ones in linalg.

Also, can you `[DO NOT SQUASH]` this and `git rebase -i` your branch into an external and internal part at some point?

mlir/test/Conversion/MIGraphXToTosa/migraphx-to-rock-conv-nhwc.mlir
yiqian1: Does it matter? ROCK can support any layouts.

krzysz00: Yes, yes it does, because that layout string represents the in-memory layout (as best it can)

More importantly, this matters because - I think - it means you've gotten your transformations backwards

yiqian1:
> I'm blocking pending confirmation that you've got the convolution layouts right here - the ones in the Rock tests don't match the ones in linalg.
> 
> Also, can you `[DO NOT SQUASH]` this and `git rebase -i` your branch into an external and internal part at some point?

Done.

external/llvm-project/mlir/include/mlir/Dialect/Linalg/IR/LinalgNamedStructuredOps.yaml
```diff
@@ -2695,6 +2695,243 @@ structured_op: !LinalgStructuredOpConfig
                 - !ScalarExpression
                   scalar_arg: K
 --- !LinalgOpConfig
+metadata: !LinalgOpMetadata
+  name: conv_2d_nhwgc_gfhwc
```
manupak, line 2699: Please remind me why are we adding a new op here and why we cannot use `linalg.conv_2d_ngchw_fgchw` for e.g.
The upstream has motivations to add these ops for different layout as they d want to perform optimizations in a linalg-based codegen approach (e.g. IREE) however our requirement is for verification.

Therefore, I dont see the reason why we cant use `linalg.conv_2d_ngchw_fgchw` with a transpose on the tensors.

external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalgNamed.cpp
```diff
@@ -199,6 +200,22 @@ class ConvConverter : public OpConversionPattern<TosaConvOp> {
...
+
+    if (auto convop = dyn_cast<tosa::Conv2DOp>(&op)) {
+      if (convop->getGroup().has_value())
+        group = convop->getGroup().value();
+    }
+
+    if (group > 1 && isConv2DOp &&
+        !std::is_same<LinalgConvOp, linalg::Conv2DNhwgcGfhwcOp>::value)
+      return rewriter.notifyMatchFailure(
+          op, "tosa.conv ops should map to grouped convolution ops");
```
manupak: I find this error message to be a bit cryptic.
Can we say something like "grouped tosa.conv ops should map to linalg::Conv2DNhwgcGfhwcOp" 

Please see my other comment above. I think we can do a transpose and not introduce a new linalg op.
Or at least I'd like to understand why such an approach wont work for our requirements.

external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalgNamed.cpp
```diff
@@ -199,6 +200,22 @@ class ConvConverter : public OpConversionPattern<TosaConvOp> {
...
+
+    if (group == 1 && isConv2DOp &&
+        !std::is_same<LinalgConvOp, linalg::Conv2DNhwcHwcfOp>::value)
```
manupak, line 217: Same here. Lets be specific about the error in the message

external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalgNamed.cpp
```diff
@@ -248,34 +263,66 @@ class ConvConverter : public OpConversionPattern<TosaConvOp> {
...
     Value zeroTensor = rewriter
                            .create<linalg::FillOp>(loc, ValueRange{zero},
                                                    ValueRange{emptyTensor})
                            .result();
+    if (isConv2DOp && group > 1) {
+      SmallVector<int64_t, 5> newResultShape{resultShape[0], resultShape[1],
+                                             resultShape[2], group,
+                                             resultShape[3] / group};
+      newResultTy = RankedTensorType::get(newResultShape, resultETy);
+      zeroTensor = rewriter.create<tosa::ReshapeOp>(
```
manupak: Why not adjust the resultShape around 
```
auto resultShape = resultTy.getShape();
```

and get rid of these lines. So the zero tensor is created in the right shape ?

manupak: i.e. no reshape is needed

external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalgNamed.cpp
```diff
@@ -307,9 +354,15 @@ class ConvConverter : public OpConversionPattern<TosaConvOp> {
...
           rewriter
               .create<LinalgConvQOp>(
-                  loc, resultTy, ValueRange{input, weight, iZpVal, kZpVal},
+                  loc, newResultTy, ValueRange{input, weight, iZpVal, kZpVal},
                   ValueRange{zeroTensor}, strideAttr, dilationAttr)
               ->getResult(0);
+      if (isConv2DOp && group > 1) {
+        conv = rewriter.create<tosa::ReshapeOp>(
+            loc, RankedTensorType::get(resultShape, resultETy), conv,
+            rewriter.getDenseI64ArrayAttr(resultShape));
```
manupak, line 362: Ok, you need this one. 
So I suppose we can use resultTy.getShape() here ?

external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalgNamed.cpp
```diff
@@ -318,10 +371,16 @@ class ConvConverter : public OpConversionPattern<TosaConvOp> {
 
     Value conv = rewriter
                      .create<LinalgConvOp>(
-                         loc, resultTy, ValueRange{input, weight},
+                         loc, newResultTy, ValueRange{input, weight},
                          ValueRange{zeroTensor}, strideAttr, dilationAttr)
                      ->getResult(0);
 
+    if (isConv2DOp && group > 1) {
```
manupak: nit : we just do a very early check on isConv2DOp and fail ; so that we dont have to keep checking whether its a conv2d

external/llvm-project/mlir/include/mlir/Dialect/Linalg/IR/LinalgNamedStructuredOps.yaml
```diff
@@ -2695,6 +2695,243 @@ structured_op: !LinalgStructuredOpConfig
...
+              scalar_fn:
+                kind: type
+                fn_name: cast_signed
+                type_var: U
+                operands:
+                - !ScalarExpression
+                  scalar_arg: K
+--- !LinalgOpConfig
+metadata: !LinalgOpMetadata
+  name: conv_2d_nhwgc_gfhwc_q
```
manupak, line 2798: ok quantized grouped conv is something needed and not there in upstream.

mlir/test/Conversion/MIGraphXToTosa/migraphx-to-rock-conv-nhwc.mlir
```diff
@@ -6,9 +6,9 @@
 // CHECK-SAME: memref<1x3x3x7xf32>
 func.func @convNHWC(%in: !migraphx.shaped<1x4x5x5xf32, 100x1x20x4>, %fil: !migraphx.shaped<7x4x3x3xf32, 36x1x12x4>) -> !migraphx.shaped<1x7x3x3xf32, 63x1x21x7> attributes {kernel, arch = "gfx1100", num_cu = 48 : i64} {
   // CHECK: rock.conv2d
-  // CHECK-SAME: filter_layout = ["k", "y", "x", "c", "g"]
-  // CHECK-SAME: input_layout = ["ni", "hi", "wi", "ci", "gi"]
-  // CHECK-SAME: output_layout = ["no", "ho", "wo", "ko", "go"]
+  // CHECK-SAME: filter_layout = ["g", "k", "y", "x", "c"]
+  // CHECK-SAME: input_layout = ["ni", "hi", "wi", "gi", "ci"]
+  // CHECK-SAME: output_layout = ["no", "ho", "wo", "go", "ko"]
```
manupak, line 11: Why are these changing ?

manupak (commented):
I did an initial pass.

My main concern is why are we adding a new linalg grouped convolution operator as opposed to using the existing one with a transpose -- though I see you need to add one for quantized. In an alternate universe, we could just add quantized one to complement already existing `[linalg.conv_2d_ngchw_fgchw](https://mlir.llvm.org/docs/Dialects/Linalg/#linalgconv_2d_ngchw_fgchw-linalgconv2dngchwfgchwop)`. Im not strongly against adding one but felt like if we didn;t add that upstream would maintain that lowering for us because we only need this in our reference flow.

Secondly, I dont think the existing tests should change as part of this ? any reason ?

external/llvm-project/mlir/include/mlir/Dialect/Linalg/IR/LinalgNamedStructuredOps.yaml
krzysz00: Because we're upstreaming this and we should give people the piece they might want to optimize over in the future.

external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalgNamed.cpp
krzysz00: I don't want this PR to minimally fit our requirements but to be something acceptable to MLIR as a whole.

mlir/test/Conversion/MIGraphXToTosa/migraphx-to-rock-conv-nhwc.mlir
krzysz00: Because we used to just tack the illusory group dimension at the end of the tensor, but now we're actually handling materializing a `g` dimension (if one exists) correctly.

krzysz00 (commented):
This looks much more like what I'd like to see here - I'll approve once Manupa is happy

external/llvm-project/mlir/lib/Conversion/TosaToLinalg/TosaToLinalgNamed.cpp
manupak: if we are eagerly upstreaming these, then it is fine.

 > but to be something acceptable to MLIR as a whole.

Yea, my comments were based off that this is just going to be rocMLIR thing downstream.

Its a shame that upstream cant fold linalg transposes when they are creating linalg.generic ops where I/O affine maps could nicely handle the folding. So we have to keep adding all permutations of layouts as linalg named ops.

external/llvm-project/mlir/include/mlir/Dialect/Linalg/IR/LinalgNamedStructuredOps.yaml
manupak: Good to know that we have plans to upstream these eagerly.

krzysz00: As far as I'm concerned, we have plans to upstream eagerly unless otherwise specified

manupak (approved):
Thanks for the changes! LGTM

mlir/test/fusion/pr-e2e/migraphx-group-conv2d.e2e.mlir
```diff
@@ -0,0 +1,10 @@
+// RUN: rocmlir-gen --clone-harness -arch %arch -fut mlir_convolution %s | rocmlir-driver  -kernel-pipeline migraphx | rocmlir-driver -host-pipeline migraphx,highlevel -targets %arch | rocmlir-gen -ph -verifier clone -fut mlir_convolution_wrapper - | rocmlir-driver -host-pipeline mhal -kernel-pipeline full | xmir-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_float16_utils%shlibext,%linalg_test_lib_dir/libmlir_c_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_async_runtime%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CLONE
+// ALLOW_RETRIES: 2
+// CLONE: [1 1 1]
+
+module {
+  func.func @mlir_convolution(%arg0: !migraphx.shaped<1x4x16x16xf32, 1024x256x16x1>, %arg1: !migraphx.shaped<4x1x3x3xf32, 36x9x3x1>) -> !migraphx.shaped<1x4x14x14xf32, 784x196x14x1> attributes {arch = "gfx1100", kernel = "mixr", num_cu = 42 : i64} {
```
manupak, line 6: Will this test work on non gfx1100 ? I think we need to remove attributes from here and your very own "--clone-harness" will add them back

yiqian1: Yes. It works on non-gfx1100. The rocmlir-gen --clone-harness will replace the arch attribute with the right one.

krzysz00: Huh, so we don't have to do the `sed/##ARCH##/%arch/g` thing anymore?

Also, might be worth removing `num_cu` here

krzysz00 (approved):
Approved now that Manupa's happy

PR #1352, opened 2023-12-13
Refactor accelGemmOp to accept views

giuseros:
Second attempt to refactor `threadwise_accel_gemm`. 

The high level idea is to pass views to the operator, in this way the operator is more expressive and can be used in different loop schedules (`i,j,k`, `k,i,j`, etc...) by simply changing the transformation maps. 

The operator now accepts the following views:
-  `[i,k]` for `A` 
-  `[j,k]` for `B` 
- `[i,j]` for `C` 
And will loop over `i,j,k` doing `C[i,j] += accel(A[i,k], B[j,k])`. 

This makes the operator very similar to the non-accel counterpart, which should facilitate the unification process between accel and non-accel paths (cc @djramic).


krzysz00 (approved):
I assume that you've checked assembly output and that this produces no functional change, so approved

giuseros:
Yeah, the difference is basically flat. Posting the perf comparison here for future reference:
[issue1154.zip](https://github.com/ROCm/rocMLIR/files/13674967/issue1154.zip)

PR #1353, opened 2023-12-13
Let CI build CK for the specific architecture of the node

giuseros:


sjw36 (approved):
Looks good! thx

stefankoncarevic:
If it should be built for gfx 1011, I don't see that this case is written for $CHIP. Is it gfx 1011 for sure?

giuseros:
> If it should be built for gfx 1011, I don't see that this case is written for $CHIP. Is it gfx 1011 for sure?

It needs to build for all architectures (gfx90a, gfx1030, gfx110, etc...)

PR #1354, opened 2023-12-14
[Attention] Use transposed/rotated attention kernel

manupak:
This requires https://github.com/ROCm/rocMLIR/pull/1344 to be merged.

Before this PR, the attention kernel would look like as follows : 
out = softmax(Q x K) x V

After this PR, it is now : 
out = transpose( transpose(V) x softmax( transpose(K) x transpose(Q) ) )

This layout is interesting because for two reasons:
1) The reductions axis of the softmax will be laid out in just two threads for MFMA (hence 1344 is important)
2) It introduces an opportunity (not done here) to avoid LDS-based swizzling to prepare the operands to MFMA as the transposed thread layout is what the output is produced after the first gemm.

Additionally, this PR also includes relaxation of blocksize >= 512 check just for attention because the performant tuning config for triton problems are just that and happens to work fine. In future we can investigate, whether that check is relevant now.

closes : https://github.com/ROCm/rocMLIR-internal/issues/1268


mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -542,8 +544,10 @@ void RockPipeline::runOnOperation() {
       }
     }
   }
-  if (isMultiBufferingFailed)
+  if (isMultiBufferingFailed) {
+    llvm::errs() << "Multi buffering failed to apply!\n";
```
krzysz00: `emitError()` please, so that we don't spam client error logs

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -552,8 +556,10 @@ void RockPipeline::runOnOperation() {
     gpuMemoryBytes[addressSpace] += alloc.getType().getShape().back();
   });
 
-  if (gpuMemoryBytes[AddressSpace::Workgroup] > size_t(64 * 1024))
+  if (gpuMemoryBytes[AddressSpace::Workgroup] > size_t(64 * 1024)) {
+    llvm::errs() << "LDS consumption is more than 64K!\n";
```
krzysz00: Same here

krzysz00 (approved):
Approved assuming the `llvm::errs() <<` is replaced with MLIR error facilities before merging

PR #1356, opened 2023-12-19
Cleanup duplicated code for threadwise accel gemm views

manupak:
We had code duplication to create views as required by ThreadwiseAccelGemm op.
(Im not 100% sure why ?)

This PR factors that code into AccelEmitters.

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
@@ -88,6 +88,44 @@ void AccelEmitter::computeOutputConversion(PatternRewriter &b, Location loc,
   }
 }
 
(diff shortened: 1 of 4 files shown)


giuseros:
I am not opposed, but are you sure that these are accelerator specific? It's more like they are operation specific. I mean, the way we might want to reorder the loops in gemm might be different to the way we might want to reorder the loops in FA.

manupak:
What is being refactored seems bit orthogonal to loops.
Its just the views that passthrough kBasePerThread and flattening of m_repeats x n_repeats. 

Loops are still defined in the op as you say...

giuseros:
Right but if we change the loop order from m,n,k to k,m,n (for instance) those views will also have to change (I think)

krzysz00 (approved):
Nice cleanup

PR #1358, opened 2023-12-31
Fix gemm and attention toCommandLine order to match Tuna output.

pcf000:
I tried to create a canonical order of options in the .tsv file produced by tuning and in the configs generated in perfRunner.py, so the configs would reliably be found in the .tsv -- it's a simple string match.  Unfortunately, I messed up the order in a couple of places and perf-configs were not found.  That ended up producing NaNs in the output when comparing against rocBLAS, which led to "FAILED" in the html report, and also caused occasional outright test failures.

### DIFF
--- a/mlir/utils/performance/perfRunner.py
+++ b/mlir/utils/performance/perfRunner.py
@@ -687,7 +687,7 @@ class GemmConfiguration(PerfConfiguration):
     def toCommandLine(self):
         return (f"-t {self.dataType} -out_datatype {self.outDataType} "
                 + f"-transA {str(self.transA).lower()} -transB {str(self.transB).lower()} "
-                + f"-g {self.g} -m {self.m} -k {self.k} -n {self.n}")
+                + f"-g {self.g} -m {self.m} -n {self.n} -k {self.k}")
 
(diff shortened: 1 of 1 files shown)


pcf000:
I do have an open issue to unify this area of code with the similar area in Tuna;  both have configuration objects with some overlap in interface, and removing duplication would help.

mlir/utils/performance/perfRunner.py
```diff
@@ -841,9 +841,10 @@ def fromCommandLine(cls, argv, arch, numCU):
         return cls(dtype, g, seq_len, head_dim, with_attn_scale, transQ, transK, transV, transO, arch, numCU, perf_config)
 
     def toCommandLine(self):
-        return (f"-transO {str(self.transO).lower()} -transV {str(self.transV).lower()} "
-                + f"-transK {str(self.transK).lower()} -transQ {str(self.transQ).lower()} "
-                + f"-t {self.dataType} -g {self.g} "
+        return (f"-t {self.dataType} "
```
krzysz00, line 844: Could this be a `' '.join([...])`

krzysz00 (approved):

sjw36 (approved):

PR #1361, opened 2024-01-04
Fold broadcasts in `GEMM([g,m,k], [broadcast(g), k,n])` on `m`

giuseros:
This is the pass I wrote to fold the broadcast on the batch to the `M` dimension. 

IR before:
```
A = <GxMxK>
B = <broadcast{1}(G)xKxN>
C = gemm(A, B)
```
IR after:
```
A = <G*MxK>
B = <KxN>
C = gemm(A, B)
```

I discussed with @manupak the possibility to compute blockSize and gridSize just before `gemmToGridwise` and created the ticket https://github.com/ROCm/rocMLIR-internal/issues/1319. For now I get away by re-running the `affixTuningParameters` pass once more after broadcasts dimensions are folded.


mlir/lib/Dialect/Rock/Transforms/FoldBroadcast.cpp
```diff
@@ -0,0 +1,174 @@
+//===- Regularize.cpp - rewrites to allow Rock kernel fusion  ------===//
```
manupak: copy-pasta :)

mlir/lib/Dialect/Rock/Transforms/FoldBroadcast.cpp
```diff
@@ -0,0 +1,174 @@
...
+#include "llvm/Support/Debug.h"
+
+namespace mlir {
+namespace rock {
+#define GEN_PASS_DEF_ROCKFOLDBROADCASTPASS
+#include "mlir/Dialect/Rock/Passes.h.inc"
+} // namespace rock
+} // namespace mlir
+
+#define DEBUG_TYPE "rock-canonicalize-gemm"
```
manupak: rock-fold-broadcast ?

mlir/lib/Dialect/Rock/Transforms/FoldBroadcast.cpp
```diff
@@ -0,0 +1,174 @@
...
+    // There is no batch, hence nothing that can be a broadcast
+    if (trMap.getUpperBounds().size() != 3)
+      return false;
+
+    auto ops = trMap.getOps();
+
+    // Look for a `Broadcast{1} at [0]` transformation
+    for (auto op : ops)
+      if (op.getType() == rock::TransformType::Broadcast &&
+          op.getUpperDims().back() == 0 && op.getParams().back() == 1)
```
manupak: Im a bit vary about this check... it is a bit fragile because unit dims / reshapes / transposes ? can easily break the ordering but it could be a still be a isBatchDimFoldable.

I think we need to backtrack a stride calculation on the batch dimension to see if its end up being zero on the actual memory layout.

mlir/test/fusion/pr-e2e/mixr-gemm/mixr-gemm-broadcast.e2d.mlir
```diff
@@ -0,0 +1,19 @@
+// RUN: rocmlir-driver -kernel-pipeline migraphx,highlevel %s | rocmlir-gen -ph -print-results -rand none - | rocmlir-driver -arch %arch -c  | mlir-cpu-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_float16_utils%shlibext --entry-point-result=void | FileCheck %s
```
manupak, line 1: Here is an example I d like to see working : 

```mlir
func.func @mlir_transpose_slice_dot_add_add_tanh_add_sigmoid_sub_mul_mul_add_real(%arg0: !migraphx.shaped<1x64xf32, 64x1>, %arg1: !migraphx.shaped<32x64xf32, 64x1>, %arg2: !migraphx.shaped<32x64xf32, 64x1>, %arg3: !migraphx.shaped<32x64xf32, 64x1>, %arg4: !migraphx.shaped<32x64xf32, 64x1>, %arg5: !migraphx.shaped<32x64xf32, 64x1>, %arg6: !migraphx.shaped<32x64xf32, 64x1>, %arg7: !migraphx.shaped<128x64xf32, 64x1>) -> !migraphx.shaped<32x64xf32, 64x1> attributes {kernel = "mixr", num_cu = 48 : i64} {
    %0 = migraphx.multibroadcast %arg0 {out_dyn_dims = [], out_lens = [32, 64]} : <1x64xf32, 64x1> -> <32x64xf32, 0x1>
    %1 = migraphx.transpose %arg7 {permutation = [1, 0]} : <128x64xf32, 64x1> -> <64x128xf32, 128x1>
    %2 = migraphx.slice %1 {axes = [1], ends = [128], starts = [64]} : <64x128xf32, 128x1> -> <64x64xf32, 64x1>
    %3 = migraphx.dot %arg6, %2 : <32x64xf32, 64x1>, <64x64xf32, 64x1> -> <32x64xf32, 64x1>
    %4 = migraphx.add %3, %0 : <32x64xf32, 64x1>, <32x64xf32, 0x1> -> <32x64xf32, 64x1>
    %5 = migraphx.add %arg1, %4 : <32x64xf32, 64x1>, <32x64xf32, 64x1> -> <32x64xf32, 64x1>
    %6 = migraphx.tanh %5 : <32x64xf32, 64x1> -> <32x64xf32, 64x1>
    %7 = migraphx.add %arg2, %arg3 : <32x64xf32, 64x1>, <32x64xf32, 64x1> -> <32x64xf32, 64x1>
    %8 = migraphx.sigmoid %7 : <32x64xf32, 64x1> -> <32x64xf32, 64x1>
    %9 = migraphx.sub %arg4, %8 : <32x64xf32, 64x1>, <32x64xf32, 64x1> -> <32x64xf32, 64x1>
    %10 = migraphx.mul %9, %6 : <32x64xf32, 64x1>, <32x64xf32, 64x1> -> <32x64xf32, 64x1>
    %11 = migraphx.mul %8, %arg5 : <32x64xf32, 64x1>, <32x64xf32, 64x1> -> <32x64xf32, 64x1>
    %12 = migraphx.add %10, %11 : <32x64xf32, 64x1>, <32x64xf32, 64x1> -> <32x64xf32, 64x1>
    return %12 : !migraphx.shaped<32x64xf32, 64x1>
  }
```

manupak (commented):
I think we need to do a bit more than tracking down a broadcast.
I have given a *suggestion* but happy to hear alternatives, however, I d like us to nail down the issue.

mlir/lib/Dialect/Rock/Transforms/FoldBroadcast.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+  // Determine if the first dimension of a view is a broadcast
+  bool isBatchDimFoldable(PatternRewriter &rw, Value aView) const {
+    auto [buffer, views, needs64BitIdxA] = untransform(rw, aView);
+
+    // There are no views, hence no broadcast is possible
+    if (views.empty())
+      return false;
+
+    auto trMap = views[0].cast<TransformMapAttr>();
```
manupak, line 125: It also can be on any views[N] not neccesarily on views[0]

mlir/lib/Dialect/Rock/Transforms/FoldBroadcast.cpp
giuseros: Just to be sure, you are saying that a broadcast might also not be upfront, right? It might be a passthrough/slice that leads to a broadcast.

manupak: yes -- any other transform to be precise.

giuseros: I am not sure what you mean with "any other transform"? I was thinking to "rerouting" transforms(like passthrough), because I am not sure how can I deal with `broadcast` buried behind a Merge/unmerge/embed

manupak: well the idea I had was to calculate stride of "g" dimension traversing up the trMaps.
Cant we handle the stride calculation in a merge/unmerge as the cumalative mulplication of faster changing dimensions in the merge set ? 

Embed is the tricky one but since we are only interested of figuring out stride is zero or not. We can possibly inherit the stride from bottom coords from the embed set

manupak: I think it is quite plausible that we may encounter shape collapsing and expansion around the broadcast -- which usually result in merge/unmerge 

Embeds occur in conv to gemm I think. Thats a good point! we should test a convolution too.

mlir/include/mlir/Dialect/Rock/IR/RockGemmWrapperInterface.td
```diff
@@ -156,6 +156,18 @@ def RockGemmWrapperInterface : OpInterface<"RockGemmWrapperInterface"> {
           $_op->setAttr($_op.getParamsAttrName(), params);
         }]
       >,
+    InterfaceMethod<
```
krzysz00, line 159: Didn't we factor this out into another interface? Or is that something we just planned to do?

krzysz00: Looks like it was just planned

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
```diff
@@ -124,14 +124,18 @@ void rock::buildKernelPipeline(OpPassManager &pm,
...
+   *   --rock-fold-broadcast --rock-affix-params --rock-gemm-to-gridwise
+   *   --rock-regularize  --rock-gridwise-gemm-to-blockwise
    */
   auto &funcPm = pm.nest<func::FuncOp>();
   funcPm.addPass(rock::createRockAffixTuningParametersPass(
       rock::RockAffixTuningParametersPassOptions{0, 0,
                                                  options.tuningFallback}));
   funcPm.addPass(rock::createRockConvToGemmPass());
+  funcPm.addPass(rock::createRockFoldBroadcastPass());
+  funcPm.addPass(rock::createRockAffixTuningParametersPass(
```
krzysz00: Could you run this conditionally from within `createRockFoldBroadcastPass()`?

mlir/lib/Dialect/Rock/Transforms/FoldBroadcast.cpp
```diff
@@ -0,0 +1,174 @@
+//===- Regularize.cpp - rewrites to allow Rock kernel fusion  ------===//
+//
+// Copyright 2022 Advanced Micro Devices.
```
krzysz00: Maybe also fix date?

mlir/lib/Dialect/Rock/Transforms/FoldBroadcast.cpp
krzysz00: I think the test is:

Follow the `G` dimension (this'll be an index, starts at 0) while looping through the transform stack.

PassThrough updates the index, as does Slice (given what condition we're looking for).

If, at some point, the transform that produced the current position of the G dimension is a Broadcast of a length-1 dimension, this is a foldable broadcast. Alternatively, if the G dimension is AddDim{}'d on, that's also a broadcast and so passes the check.

If the G dimension is created by Merge[*], Unmerge, or Embed, abort.

[*] There's a possible exception for Merge{1, 1, len(G), 1, 1}-type cases, but that's Effort and unlikely - unless we find it an example. Same goes for trivial Unmerge operations.

mlir/lib/Dialect/Rock/Transforms/FoldBroadcast.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+  // Select the 0th slice from a broadcast, de facto removing the broadcast
+  // dimension
+  Value unbroadcast(PatternRewriter &rw, Location loc,
+                    TypedValue<ShapedType> buffer) const {
+    auto shape = buffer.getType().getShape();
+    rock::TopDownTMBuilder unbroadcastBuilder(rw, {"d0", "d1"},
+                                              {shape[1], shape[2]}, loc);
+    unbroadcastBuilder.constDim({"g"}, 0, 0, shape[0]);
+    unbroadcastBuilder.passThrough({"d0"}, {1}, {"d0"});
```
krzysz00: Nit: you can combine the two `passThrough` calls

mlir/test/fusion/pr-e2e/mixr-gemm/mixr-gemm-broadcast.e2d.mlir
krzysz00: @manupak Can you clarify how the optimization we're discussing applies to that example? I don't see it here.

manupak: Yea -- it was not related to the optimization but just wanted to highlight this is also a foldable broadcast.

an example could be -- also that would also work with your logic (the traversal you mentioned earlier)

128 x f32
 |
V
4 x 32 x f32
 | bcast
V
16 x 4 x 32 x f32

mlir/lib/Dialect/Rock/Transforms/FoldBroadcast.cpp
manupak: That can also work in most cases until batches are collapsed/expanded after the broadcast -- which I agree is unlikely.
I was more speaking about strides thing was : dont we need that for problem key v2 ?. 
Therefore if we have that, it is just a matter of looking at the stride of "g".

giuseros: Yep this was a plan I had, but I didn't do it because I didn't know if I was being too general. I think my confusion is about: 

> If the G dimension is created by Merge[*], Unmerge, or Embed, abort.

@manupak is that what you meant? Or you were suggesting to go through these transformations as well?

manupak: So there is a reason we have the following code (it was done by Jungwook): 

https://github.com/ROCm/rocMLIR/blob/3202ef9b92a9278708b445a2611fff41750ffeea/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp#L297-L321

to me it indicates that we will probably add a reshape to batch dimension and it would be after the broadcast because the above case handle where there are multiple dimensions (can just be unit dims) and fold them to be a single batch dimension.

manupak: > @manupak is that what you meant? Or you were suggesting to go through these transformations as well?

I was suggesting to calculate/update a stride as we go through the stack and stop if and when it becomes zero (i.e. broadcast).

I ll point out that @krzysz00 's logic will work until we have a reshape that involves batch dimension after the broadcast. Now, thinking about it in the broadcast case, we might not commonly exercise that case that but I am a bit worried about unit dims.

Anyway, when we have problem key v2, we will have to calculate the strides. Therefore, whatever logic we are discussing could just go away and we can just take the decision looking at the stride ?

mlir/test/fusion/pr-e2e/mixr-gemm/mixr-gemm-broadcast.e2d.mlir
manupak: ok Im just realizing there is no batch dimension at all. sigh. nvm

manupak: anyways, my ask was to add something that does not have 'just'  a multibroadcast.

manupak: After having a second thought, this is actually a counter-example where there is a broadcast on the first dimension where we should *NOT* fold.

giuseros:
Ok, so I updated the pass :
- It now runs in the `highlevel` pipeline
- It now follows the dimensions through their transformations (I added also more complicated tests)

giuseros:
@manupak @krzysz00  a gentle reminder :)

krzysz00 (approved):
This all looks good to me

manupak (approved):
Looks great now!

PR #1362, opened 2024-01-04
[Attention] Replace 2nd blockwise_gemm with threadwise_gemm

manupak:
This commit replaces the second blockwise_gemm operation
with threadwise_gemm operation and friends.

This is a stepping stone to analyze the LDS traffic used to
swizzle the gemm operands into accelerator layouts (MFMA/WMMA)
and to potentially remove unnecessary LDS traffic in future.

closes : https://github.com/ROCm/rocMLIR-internal/issues/1304


krzysz00 (commented):
High-level question: what's different about the code you inlined here compared to the implementation of blockwise_gemm?

manupak:
Basically nothing is different.

1) Except now it allows me perf experiment by removing LDS read / write (incorrectly) of gemm operands + removing LDS allocation it self
2) Based on the speedup, I am considering removing the one LDS to/fro traffic as they could be omitted maybe at the expense of kpack in the current layout.

krzysz00:
Ah, so this is preliminary work for making the softmax -> gemm \#2 transition being in-registers/DPP/reading out of LDS in the gemm 2 layout ...

krzysz00 (approved):
Approved to get out of the way of perf experiments

PR #1364, opened 2024-01-08
[DO NOT SQUASH] Set matrix C to zero for GEMMs when `atomic_add save` is used

ravil-mobile:
This PR closes ROCm/rocMLIR-internal#1322


krzysz00 (changes requested):
I'm opposed to this sort of thing. It's an awkward special case that's easy to get wrong and that clients need to be very aware of, lest we have bug reports that we can't reproduce on our side because of this special-case. Also, this might not work for, for example, stuff coming from migraphx.

I claim we really should have a general solution to "this buffer has to be initialized". Even if it's not multiple kernels, what I'd want to see here is, perhaps, a pass that runs after `-kernel-pipeline=gpu` and hacks on the host code to add `memset`s. Or one that runs after final compilation, if we expect MIGraphX to read from that.

And then MIGraphX would similarly (perhaps we'd keep the needs-fill attribute around in some clever fashion) inspect the relevant intermediate and add in buffer fills.

manupak:
I think what we need is rocmlir-gen inspect the kernel function and insert prefills for whatever arg that need them.

> hacks on the host code to add memsets

I dont think we need a pass to do this.
This is about creating the right testing harness that would mimic what a client would do.

I think @krzysz00 's main complain is we are going off on '-store-method atomic_add' to insert prefills which I agree too.

Why cant we add the prefill arg attr to func arg and use rocmlir-gen to inspect function signature add prefills as neccessary ? 
Since we are modifying the kernel function signature (by hooking up arg attrs in the kernel generation), I think it is obvious to a client (or rocmlir-gen) to generate right host code.

manupak:
My suggestion is lift this : https://github.com/ROCm/rocMLIR/blob/192515b04fcdcc7c60fc3a8ae8d2bc146ac178a9/mlir/tools/rocmlir-gen/rocmlir-gen.cpp#L2910

out of just clone harness and run it somewhere s.t. it always run.

In the kernel gen, (affix tuning params or gemm to gridwise) just set the output argument to be prefilled.
We also need to error out in align-tiling if trailing elementwise ops are present.

krzysz00:
@manupak The problem I have is that, if I remember right, the reduction fusion can add a prefill attribute in the middle of the kernel compilation - that is, during align-tiling. And so we need to be able to handle prefill requests that come in during kernel compilation, which means a pass or some other transformation that runs _after_ kernel compilation.

manupak:
rocmlir-gen host harness generation is run after ~~kernel generation~~ highlevel for fusion tests though...

manupak:
Maybe in this scenario, we generate the host harness at the kernel generation time...

krzysz00:
I could've sworn we don't know we need a prefill until align-tiling?

krzysz00:
Which brings me to my original point, of making "add the host harness" something that happens (or is hacked on) later, so that both we and MIGraphX use the same logic for that

manupak:
> I could've sworn we don't know we need a prefill until align-tiling?

well its done in tosa-to-rock (i.e. in highlevel) because we know tosa.reduce_* op needs the output buffer prefilled.
so rocmlir-gen picks it up

manupak:
Can we not know whether it is going to be split-k after highlevel ?

ravil-mobile:
@manupak and @krzysz00, can you have a look at the code changes? I think it looks more generic than the initial solution

krzysz00:
@manupak Maybe, but that's logic I'd want to see created. Currently, for the convolution case, that lives in a combination of `Conv2DGenerator` and `rocmlir-gen` and `affix-tuning-parameters`.

From what I can tell, the decision about whether to split-K can and will be a function of the tuning parameters. So we'll need to pre-query (either late in the highlevel pipeline or during rocmlir-gen) whether or not we'll be doing a split-K.

Except that can't quite work because we don't get told what the tuning parameters will be until after the highlevel pipeline is over.

manupak:
> @manupak and @krzysz00, can you have a look at the code changes? I think it looks more generic than the initial solution

@ravil-mobile I think the main concern here is we need a host harness modification done a bit later in the pipeline because it seems like we cant lift affix-tuning-params to highlevel (can we ?)

While your solution is fine to support verification of '-store-method' arg of rocmlir-gen kernel generation -- which I can appreciate otherwise is broken -- we would needer a better solution to address when split-k decision is taken. In the short & current term, it would be on affix-tuning-params.. (On the longer term, I d like us to move that even down to gemm-to-gridwise).

The thing I dont like with @krzysz00's proposal is we doing a host harness modification (by running a pass) on the kernel pipeline. However, if we can run rocmlir-gen after the kernel pipeline (again) to insert such prefills that would be a clean seperation of harness modification that deployment scenarios wont exercise. What do you think @krzysz00  ?

krzysz00:
Might be interesting to move the host harness generation logic in general past kernel compilation. - something like `rock.exec_kernel` or one of the mhal ops.

ravil-mobile:
Hi @manupak  and @krzysz00, I generalized `Rock_ZeroInitKernelOp` and made `Rock_InitKernelOp` instead. It works much better with prefill.

ravil-mobile:
Hi, I was thinking about a solution. So, we don't want to modify harness logic and we want to delay the decision about prefill, etc. 
Because of such requirements, we need to some extra code generation inside `rocmlir-driver`.

I was playing with `rocmlir-gen` output and came up to the following; we may need some extra wrapping function which `rocmlir-driver` can operate and modify.

---

Here is the original code 
```mlir
  func.func @rock_gemm(%arg0: memref<1x1024x1024xf16>, %arg1: memref<1x1024x512xf16>, %arg2: memref<1x1024x512xf16> {rock.prefill = 0.000000e+00 : f32}) attributes {kernel, mhal.arch = "amdgcn-amd-amdhsa:gfx90a"} {
    rock.gemm %arg2 = %arg0 * %arg1 features =  mfma|dot|atomic_add storeMethod =  atomic_add {arch = "amdgcn-amd-amdhsa:gfx90a"} : memref<1x1024x512xf16> = memref<1x1024x1024xf16> * memref<1x1024x512xf16>
    return
  }

  func.func @prefill_0(%arg0: memref<1x1024x512xf16>) attributes {kernel, mhal.arch = "amdgcn-amd-amdhsa:gfx90a"} {
    rock.init_kernel %arg0 features =  mfma|dot|atomic_add {initValueAttr = 0.000000e+00 : f32} : memref<1x1024x512xf16>
    return
  }
  func.func @rock_gemm_gpu(%arg0: memref<1x1024x1024xf16>, %arg1: memref<1x1024x512xf16>, %arg2: memref<1x1024x512xf16>) {
    %c1 = arith.constant 1 : index
    %c10 = arith.constant 10 : index
    %c0 = arith.constant 0 : index
    %memref = gpu.alloc  () : memref<1x1024x1024xf16>
    gpu.memcpy  %memref, %arg0 : memref<1x1024x1024xf16>, memref<1x1024x1024xf16>
    %memref_0 = gpu.alloc  () : memref<1x1024x512xf16>
    gpu.memcpy  %memref_0, %arg1 : memref<1x1024x512xf16>, memref<1x1024x512xf16>
    %memref_1 = gpu.alloc  () : memref<1x1024x512xf16>
    gpu.memcpy  %memref_1, %arg2 : memref<1x1024x512xf16>, memref<1x1024x512xf16>
    scf.for %arg3 = %c0 to %c10 step %c1 {
      func.call @prefill_0(%memref_1) : (memref<1x1024x512xf16>) -> ()
      func.call @rock_gemm(%memref, %memref_0, %memref_1) : (memref<1x1024x1024xf16>, memref<1x1024x512xf16>, memref<1x1024x512xf16>) -> ()
    }
    gpu.memcpy  %arg0, %memref : memref<1x1024x1024xf16>, memref<1x1024x1024xf16>
    gpu.dealloc  %memref : memref<1x1024x1024xf16>
    gpu.memcpy  %arg1, %memref_0 : memref<1x1024x512xf16>, memref<1x1024x512xf16>
    gpu.dealloc  %memref_0 : memref<1x1024x512xf16>
    gpu.memcpy  %arg2, %memref_1 : memref<1x1024x512xf16>, memref<1x1024x512xf16>
    gpu.dealloc  %memref_1 : memref<1x1024x512xf16>
    return
  }
```

---

Here is my modify solution
```mlir
  func.func @prefill_0(%arg0: memref<1x1024x512xf16>) attributes {kernel, mhal.arch = "amdgcn-amd-amdhsa:gfx90a"} {
    rock.init_kernel %arg0 features =  mfma|dot|atomic_add {initValueAttr = 0.000000e+00 : f32} : memref<1x1024x512xf16>
    return
  }
  func.func @rock_gemm(%arg0: memref<1x1024x1024xf16>, %arg1: memref<1x1024x512xf16>, %arg2: memref<1x1024x512xf16> {rock.prefill = 0.000000e+00 : f32}) attributes {kernel, mhal.arch = "amdgcn-amd-amdhsa:gfx90a"} {
    rock.gemm %arg2 = %arg0 * %arg1 features =  mfma|dot|atomic_add storeMethod =  atomic_add {arch = "amdgcn-amd-amdhsa:gfx90a"} : memref<1x1024x512xf16> = memref<1x1024x1024xf16> * memref<1x1024x512xf16>
    return
  }
  func.func @rock_gemm_launch(%arg0: memref<1x1024x1024xf16>, %arg1: memref<1x1024x512xf16>, %arg2: memref<1x1024x512xf16>) {
    func.call @prefill_0(%arg2) : (memref<1x1024x512xf16>) -> ()
    func.call @rock_gemm(%arg0, %arg1, %arg2) : (memref<1x1024x1024xf16>, memref<1x1024x512xf16>, memref<1x1024x512xf16>) -> ()
    return
  }
  func.func @rock_gemm_gpu(%arg0: memref<1x1024x1024xf16>, %arg1: memref<1x1024x512xf16>, %arg2: memref<1x1024x512xf16>) {
    %c1 = arith.constant 1 : index
    %c10 = arith.constant 10 : index
    %c0 = arith.constant 0 : index
    %memref = gpu.alloc  () : memref<1x1024x1024xf16>
    gpu.memcpy  %memref, %arg0 : memref<1x1024x1024xf16>, memref<1x1024x1024xf16>
    %memref_0 = gpu.alloc  () : memref<1x1024x512xf16>
    gpu.memcpy  %memref_0, %arg1 : memref<1x1024x512xf16>, memref<1x1024x512xf16>
    %memref_1 = gpu.alloc  () : memref<1x1024x512xf16>
    gpu.memcpy  %memref_1, %arg2 : memref<1x1024x512xf16>, memref<1x1024x512xf16>
    scf.for %arg3 = %c0 to %c10 step %c1 {
      func.call @rock_gemm_launch(%memref, %memref_0, %memref_1) : (memref<1x1024x1024xf16>, memref<1x1024x512xf16>, memref<1x1024x512xf16>) -> ()
    }
    gpu.memcpy  %arg0, %memref : memref<1x1024x1024xf16>, memref<1x1024x1024xf16>
    gpu.dealloc  %memref : memref<1x1024x1024xf16>
    gpu.memcpy  %arg1, %memref_0 : memref<1x1024x512xf16>, memref<1x1024x512xf16>
    gpu.dealloc  %memref_0 : memref<1x1024x512xf16>
    gpu.memcpy  %arg2, %memref_1 : memref<1x1024x512xf16>, memref<1x1024x512xf16>
    gpu.dealloc  %memref_1 : memref<1x1024x512xf16>
    return
  }
```

---

Right now, the kernel pipeline will process only function with `kernel` attribute. Maybe we need to extend it and process functions which also attributed with `launcher` (e.g., `rock_gemm_launch`). A launcher function may contain a list of functions which we want to launch one after another. Additionally, our pipelines can modify/insert additional standalone kernels if it is needed (e.g., prefill).

---

Here is what I mean. Suppose that we initially generate the following code with `rocmlir-gen`
```mlir
  func.func @rock_gemm_launch(%arg0: memref<1x1024x1024xf16>, %arg1: memref<1x1024x512xf16>, %arg2: memref<1x1024x512xf16>)  attributes {launcher} {
    func.call @prefill_0(%arg2) : (memref<1x1024x512xf16>) -> ()
    func.call @rock_gemm(%arg0, %arg1, %arg2) : (memref<1x1024x1024xf16>, memref<1x1024x512xf16>, memref<1x1024x512xf16>) -> ()
    return
  }
```
---

Our pipeline can re-write `rock_gemm_launch` as follows
```mlir
  func.func @rock_gemm_launch(%arg0: memref<1x1024x1024xf16>, %arg1: memref<1x1024x512xf16>, %arg2: memref<1x1024x512xf16>) attributes {launcher} {
    func.call @prefill_0(%arg2) : (memref<1x1024x512xf16>) -> ()
    func.call @rock_gemm(%arg0, %arg1, %arg2) : (memref<1x1024x1024xf16>, memref<1x1024x512xf16>, memref<1x1024x512xf16>) -> ()
    return
  }
```

However, this approach will involve a considerable source code refactoring. But anyway, what do you think?

manupak:
> we need to some extra code generation inside rocmlir-driver.

rocmlir-gen has been the tool used for host harness generation generally. So it would equally can be in rocmlir-gen.
Alternatively if we want it in rocmlir-driver, we could think of a 'launcher' pipeline to assemble a multi-kernel launching cpu wrapper codegen pipeline. It is just that, I feel we should not overload the 'kernel' pipeline to generate CPU code.

> Right now, the kernel pipeline will process only function with kernel attribute. Maybe we need to extend it and process functions which also attributed with launcher (e.g., rock_gemm_launch). A launcher function may contain a list of functions which we want to launch one after another.

I d like us to keep the kernel pipeline simple -- s.t. functions with kernel attributes be codegen'd it to kernels.

>  A launcher function may contain a list of functions which we want to launch one after another. Additionally, our pipelines can modify/insert additional standalone kernels if it is needed (e.g., prefill).

Isn;t this what `mhal` exist for  @sjw36  ? (I see a significant overlap between this and what mhal infra is supposed to do) ?

ravil-mobile:
I am just posting here one important note. Initially, I modified the MLIR code as follows

```mlir
 func.func @rock_gemm(%arg0: memref<1x1024x1024xf16>, %arg1: memref<1x1024x512xf16>, %arg2: memref<1x1024x512xf16> {rock.prefill = 0.000000e+00 : f32}) attributes {kernel, mhal.arch = "amdgcn-amd-amdhsa:gfx90a"} {
    rock.init_kernel %arg2 features =  mfma|dot|atomic_add {initValueAttr = 0.000000e+00 : f32} : memref<1x1024x512xf16>    
    rock.gemm %arg2 = %arg0 * %arg1 features =  mfma|dot|atomic_add storeMethod =  atomic_add {arch = "amdgcn-amd-amdhsa:gfx90a"} : memref<1x1024x512xf16> = memref<1x1024x1024xf16> * memref<1x1024x512xf16>
    return
  }
```

But it led to incorrect numerical results.

Im not sure why but it seems to me that it will result in one kernel

```mlir
// -----// IR Dump After RockGemmToGridwisePass (rock-gemm-to-gridwise) //----- //
func.func @rock_gemm(%arg0: memref<1x1024x1024xf16>, %arg1: memref<1x1024x512xf16>, %arg2: memref<1x1024x512xf16> {rock.prefill = 0.000000e+00 : f32}) attributes {block_size = 64 : i32, grid_size = 16 : i32, kernel, mhal.arch = "amdgcn-amd-amdhsa:gfx90a", wave_size = 64 : i32} {
  %cst = arith.constant dense<0.000000e+00> : vector<8xf16>
  %c0 = arith.constant 0 : index
  %0 = rock.alloc() : memref<8xf16, #gpu.address_space<private>>
  %c0_0 = arith.constant 0 : index
  rock.in_bounds_store %cst -> %0[%c0_0] : vector<8xf16> -> memref<8xf16, #gpu.address_space<private>>, index
  %true = arith.constant true
  %1 = rock.workgroup_id : index
  %c64 = arith.constant 64 : index
  %c512 = arith.constant 512 : index
  %2 = rock.workitem_id : index
  %collapse_shape = memref.collapse_shape %arg2 [[0, 1, 2]] : memref<1x1024x512xf16> into memref<524288xf16>
  %3 = arith.muli %1, %c64 : index
  %4 = arith.addi %3, %2 : index
  %5 = arith.muli %4, %c512 : index
  %c0_1 = arith.constant 0 : index
  %c8 = arith.constant 8 : index
  scf.for %arg3 = %c0_1 to %c512 step %c8 {
    %7 = arith.addi %5, %arg3 : index
    rock.global_store  set %0[%c0] -> %collapse_shape[%7] if %true features = mfma|dot|atomic_add {canStoreOffEnd, length = 8 : index} : memref<8xf16, #gpu.address_space<private>> -> memref<524288xf16>
  }
  %6 = rock.transform %arg0 by <affine_map<(d0, d1, d2) -> (d0, d2, d1)> by [<PassThrough ["gemmG"] at [0] -> ["gemmG"] at [0]>, <PassThrough ["gemmK", "gemmM"] at [1, 2] -> ["gemmK", "gemmM"] at [2, 1]>] bounds = [1, 1024, 1024] -> [1, 1024, 1024]> : memref<1x1024x1024xf16> to memref<1x1024x1024xf16>
  %alloc = memref.alloc() : memref<1x1024x512xf32>
  rock.gridwise_gemm_accel(%6, %arg1, %alloc) storeMethod( atomic_add) features =  mfma|dot|atomic_add {arch = "amdgcn-amd-amdhsa:gfx90a", blockSize = 256 : i32, gridSize = 16 : i32, numCU = 110 : i32, params = #rock.xdlops_gemm_params<kpackPerBlock = 8, mPerBlock = 128, nPerBlock = 256, kpack = 4, mPerWave = 64, nPerWave = 128, forceUnroll = true>} : memref<1x1024x1024xf16>, memref<1x1024x512xf16>, memref<1x1024x512xf32>
  linalg.generic {indexing_maps = [affine_map<(d0, d1, d2) -> (d0, d1, d2)>, affine_map<(d0, d1, d2) -> (d0, d1, d2)>], iterator_types = ["parallel", "parallel", "parallel"]} ins(%alloc : memref<1x1024x512xf32>) outs(%arg2 : memref<1x1024x512xf16>) {
  ^bb0(%in: f32, %out: f16):
    %7 = arith.truncf %in : f32 to f16
    linalg.yield %7 : f16
  }
  return
}
```

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -1046,7 +1046,43 @@ static Value makeNDMemRef(OpBuilder &b, Value var, uint32_t ndim) {
...
+static func::FuncOp createPrefillKernel(const GenParams &params,
+                                        ModuleOp module, Value mem,
+                                        Attribute prefillAttr) {
+  MLIRContext *context = module.getContext();
+  auto loc = module->getLoc();
+  OpBuilder b(context);
+
+  static size_t funcCounter{0};
+  std::string funcName = "prefill_" + std::to_string(funcCounter);
+  ++funcCounter;
```
krzysz00: Inserting into a module automatically fixes up this sort of thing, IIRC

sjw36:
> > A launcher function may contain a list of functions which we want to launch one after another. Additionally, our pipelines can modify/insert additional standalone kernels if it is needed (e.g., prefill).
> 
> Isn;t this what `mhal` exist for @sjw36 ? (I see a significant overlap between this and what mhal infra is supposed to do) ?

mhal doesn't have this feature as yet. It can only reference a single kernel launch at this point. It should be enhanced to reference a sub-graph (func with `mhal.launch`'s) that can be inlined when selected.

ravil-mobile:
Hi @manupak  and @krzysz00, can you have a look at the changes. 

As @krzysz00 suggested, I completely removed all changes made in `rocmlir-gen` (I reverted this file to the `develop` branch).

Now, `GemmToGridwise` pass adds a `mhal::prefill` attribute

```cpp
  if (op.getStoreMethod() == StoreMethod::AtomicAdd) {
    auto func = llvm::cast<func::FuncOp>(op->getParentOp());
    func->setAttr(mhal::PrefillAttr::getMnemonic(),
                  mhal::PrefillAttr::get(b.getContext(), {2}));
  }

```

I added a pass to `mhal` which is going to 1) inspect each function, 2) walk through all `gpu::LaunchFuncOp`, 3) find their definitions, 4) examines attributes of all kernel parameters, 5) insert all necessary `gpu::MemsetOp` before the corresponding `gpu::LaunchFuncOp`s.

ravil-mobile:
I have to several complains/questions about the current solution.

1) The `mhal::prefill` attribute results in zero-initialization. Probably, we would like to extend it. Note, I cannot assign an attribute to a value (only to operation). In theory, multiple buffers may request `prefill`. Thus, I need to come up to some array like data structure which holds elements `(argIndex, initValue)`. This gets a bit complicated because `initValue` can be either `Float` or `Integer`. So, please, tell me if you have solution in mind.

2) We have `rock.prefill` attribute. Probably it can be deprecated now and replaced with `mhal.prefill` one. @manupak, what do you think?

3) Currently, I only handle `Gemm` operations. We also have `AttentionOp`. @manupak, do we need to add something there as well? 

4) Currently, my solution works only for `f32`. `f16` results in the following error

```
JIT session error: Symbols not found: [ mgpuMemset16 ]
Error: Failed to materialize symbols: { (main, { _mlir_rock_gemm_gpu, _memcpy_f32_f16_524288, _memcpy_f16_f32_524288, host_naive_gemm, _memcpy_f16_f32_1048576, _mlir_main, _mlir_rock_gemm_verify2, _mlir__memcpy_f16_f32_524288, rock_gemm_gpu, main, _mlir__memcpy_f32_f16_524288, _mlir_host_naive_gemm, rock_gemm_verify2, _mlir__memcpy_f16_f32_1048576 }) }
```

manupak:
> Note, I cannot assign an attribute to a value (only to operation).

If its in MLIR, it should be possible to assign to arguments of a function. why wouldn;t that work ?

> We have rock.prefill attribute. Probably it can be deprecated now and replaced with mhal.prefill one. @manupak, what do you think?

Im fine with it.

> Currently, I only handle Gemm operations. We also have AttentionOp. @manupak, do we need to add something there as well?

I dont see a requirement for it for now. I think we can keep it for gemms.

> Currently, my solution works only for f32. f16 results in the following error

I thought we were supposed to use your brand-new InitKernel here ?
Anyway, having mgpuMemset16 is not a bad idea too.

manupak:
> 4) examines attributes of all kernel parameters

You can also examine attributes on kernel func arguments.

krzysz00:
Yeah, we want `prefill` on the arguments to functions, which might require tracing through transforms/views/... .

Also, `mgpuMensetF16` is something we can create and upsstream.

ravil-mobile:
Hi @manupak and @krzysz00,

I did some changes as we discussed. I used arg attributes in our kernel pipeline for the prefill. During `RockToGPU` pass, I moved the arg attributes to the corresponding `GPUModule`. In short, we can prefill MemRefs with arbitrary values now.

I also fixed Memset16

```diff
--- a/external/llvm-project/mlir/lib/ExecutionEngine/RocmRuntimeWrappers.cpp
+++ b/external/llvm-project/mlir/lib/ExecutionEngine/RocmRuntimeWrappers.cpp
@@ -162,6 +162,15 @@ extern "C" void mgpuMemset32(void *dst, int value, size_t count,
                                         value, count, stream));
 }
 
+extern "C" void mgpuMemset16(void *dst, int short value, size_t count,
+                             hipStream_t stream) {
+  HIP_DIAG("mgpuMemset16(0x%lx, 0x%x, %ld, 0x%lx)\n", (uint64_t)dst, value,
+           count, (uint64_t)stream);
+  HIP_REPORT_IF_ERROR(hipMemsetD16Async(reinterpret_cast<hipDeviceptr_t>(dst),
+                                        value, count, stream));
+}
```

I would suggest to move the interfacing issue between MiGraphX and rocMLIR regarding the prefill to a dedicated.

My next step would be to 1) remove `rock.prefill` attribute and substitute it with `mhal.prefill` and 2) fill all CI broken tests. What do you think?

krzysz00:
Yeah, the MIGraphX interface is a different ticket.

I think these next steps are good

ravil-mobile:
Hi @manupak and @krzysz00 

I have one more question. Shall we remove the following?

https://github.com/ROCm/rocMLIR/blob/687507f3a310781fd8a93cd4fe307d1b9b3dbe83/mlir/tools/rocmlir-gen/rocmlir-gen.cpp#L2709-L2748

because, now, `mhal` is supposed to handle this case

manupak:
@ravil-mobile yea go ahead!

ravil-mobile:
> @ravil-mobile yea go ahead!

@manupak, removed

ravil-mobile:
Hi @manupak  and @krzysz00, 

Can you, please, review the PR? I think I fixed the failing tests and I hope that the pipelines are not going to fail.

external/mlir-hal/include/mlir/Dialect/MHAL/IR/MHALAttrDefs.td
```diff
@@ -79,4 +79,15 @@ def MHAL_KernelPackageAttr : MHAL_Attr<"KernelPackage"> {
...
 
+// Misc Attrs
+def MHAL_PrefillAttr : MHAL_Attr<"Prefill"> {
+  let mnemonic = "mhal.prefill";
+
+  let parameters = (ins
+    "uint32_t":$argIndex,
+    "uint32_t":$initValue
+  );
+  let assemblyFormat = "`<` params `>`";
```
krzysz00, line 99: Documentation?

external/mlir-hal/include/mlir/Dialect/MHAL/IR/MHALAttrDefs.td
```diff
@@ -79,4 +79,15 @@ def MHAL_KernelPackageAttr : MHAL_Attr<"KernelPackage"> {
     let hasCustomAssemblyFormat = 1;
 }
 
+// Misc Attrs
+def MHAL_PrefillAttr : MHAL_Attr<"Prefill"> {
+  let mnemonic = "mhal.prefill";
+
+  let parameters = (ins
+    "uint32_t":$argIndex,
+    "uint32_t":$initValue
```
krzysz00: `"TypedAttribute": $initValue` so we don't lose type information. MIGraphX will want it.

external/mlir-hal/lib/Dialect/MHAL/Transforms/Prefill.cpp
```diff
@@ -0,0 +1,140 @@
...
+using namespace mlir;
+namespace {
+class MHALPrefillPass
+    : public mhal::impl::MHALPrefillPassBase<MHALPrefillPass> {
+
+  Value createConstantOp(OpBuilder &builder, Type elementType,
+                         uint32_t initValue) {
+    Value constantOp;
+    auto loc = builder.getUnknownLoc();
+    if (llvm::isa<FloatType>(elementType)) {
```
krzysz00: Yeah, this'll all go away if the prefill attribute has a type on it.

external/mlir-hal/lib/Dialect/MHAL/Transforms/Prefill.cpp
```diff
@@ -0,0 +1,140 @@
...
+      APInt apValue(elementType.getIntOrFloatBitWidth(), initValue, true);
+      auto constValue = builder.getIntegerAttr(elementType, apValue);
+      constantOp =
+          builder.create<arith::ConstantOp>(loc, elementType, constValue);
+    }
+    return constantOp;
+  }
+
+  void insertPrefillOps(OpBuilder &builder, gpu::LaunchFuncOp &launchOp) {
+    auto func = llvm::cast<func::FuncOp>(launchOp->getParentOp());
```
krzysz00: Shouldn't need `llvm::` here. It should just be `cast` by way of a `using namespace mlir;`, which reexports `llvm` in some cases

external/mlir-hal/lib/Dialect/MHAL/Transforms/Prefill.cpp
```diff
@@ -0,0 +1,140 @@
...
+    auto module = llvm::cast<ModuleOp>(func->getParentOp());
+
+    auto kernel = launchOp.getKernel();
+    auto *callee = module.lookupSymbol(kernel);
+    assert(callee != nullptr && "expect to find the function defenition");
+
+    auto llvmFunc = llvm::cast<LLVM::LLVMFuncOp>(callee);
+    auto gpuModule = llvm::cast<gpu::GPUModuleOp>(llvmFunc->getParentOp());
+    assert(gpuModule != nullptr &&
```
krzysz00: `cast<>` already does this assert. `dyn_cast` is the version that doesn't.

external/mlir-hal/lib/Dialect/MHAL/Transforms/Prefill.cpp
```diff
@@ -0,0 +1,140 @@
...
+    for (auto attr : prefillAttrs) {
+      auto argIdx = attr.getArgIndex();
+      assert(argIdx < kernelOperands.size() &&
+             "provided arg index is out of bounds");
+
+      auto arg = kernelOperands[argIdx];
+      auto type = arg.getType().cast<MemRefType>();
+      auto elementType = type.getElementType();
+
+      builder.setInsertionPointToStart(arg.getParentBlock());
```
krzysz00: I don't think that's right. I'd set the insertion point before the launch?

Also, async tokens?

external/mlir-hal/lib/Dialect/MHAL/Transforms/Prefill.cpp
```diff
@@ -0,0 +1,140 @@
...
+    {
+      OpBuilder builder(&getContext());
+      OpBuilder::InsertionGuard insertGuard(builder);
+      for (gpu::LaunchFuncOp launchOp : ops) {
+        insertPrefillOps(builder, launchOp);
+      }
+    }
+  }
+};
+} // namespace
```
krzysz00: Style nit: move method implementations outside of the `namespace{}`

mlir/test/e2e/gemm_with_atomic_add.cfg
```diff
@@ -0,0 +1,2 @@
+if not 'atomic_add' in config.features:
+  config.unsupported = True
```
krzysz00: Nit: missing newline.

krzysz00 (commented):
Given discussions with MIGraphX about how they want the prefill API to work (they want an `Attribute` back), we'll need to fiddle with the design some.

mlir/test/e2e/gemm_with_atomic_add.cfg
ravil-mobile: done

external/mlir-hal/lib/Dialect/MHAL/Transforms/Prefill.cpp
ravil-mobile: It is good to know. Thanks!

external/mlir-hal/lib/Dialect/MHAL/Transforms/Prefill.cpp
ravil-mobile: Yes, you're right!

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
ravil-mobile: Strange. I don't know how you managed to select these lines because I already removed them a couple of days ago. The current implementation works as you described

external/mlir-hal/include/mlir/Dialect/MHAL/IR/MHALAttrDefs.td
ravil-mobile: @krzysz00 , are you sure that there is something like `TypedAttribute`. I search over `llvm-project/mlir` and I couldn't find something similar

ravil-mobile: There is `NamedAttribute` but I am not sure whether we want to use this one. I will look through some other dialects. Maybe I will find something

ravil-mobile: Maybe it is `TypedAttrInterface` ?

https://github.com/llvm/llvm-project/blob/bf716fb7169b500a7fe3f20a39ac0288fb13cac5/mlir/include/mlir/Dialect/Arith/IR/ArithOps.td#L180-L186

ravil-mobile: I guess I found. Let me try `TypedAttr`

ravil-mobile: @krzysz00, `TypedAttr` worked and it resulted in less code. Thanks!

external/mlir-hal/lib/Dialect/MHAL/Transforms/Prefill.cpp
ravil-mobile: yes, you're correct. The entire method went away.

external/mlir-hal/lib/Dialect/MHAL/Transforms/Prefill.cpp
ravil-mobile: done

external/mlir-hal/lib/Dialect/MHAL/Transforms/Prefill.cpp
ravil-mobile: Hi @krzysz00,

there are two places where I user `setInsertionPoint`, namely: 1) At the beginning of the parent block to insert a constant provided by the `PrefillAttr` and 2) to insert `memcpy` before a kernel launch. 

You highlighted the first case - i.e., inserting a constant to the beginning of the parent block.

The whole code looks like the following

```cpp
    builder.setInsertionPointToStart(arg.getParentBlock());
    auto initConstant = builder.create<arith::ConstantOp>(loc, elementType,
                                                          attr.getInitValue());
    builder.setInsertionPoint(launchOp->getBlock(),
                              --Block::iterator(launchOp));
    builder.create<gpu::MemsetOp>(loc, mlir::Type{}, mlir::ValueRange{}, arg,
                                  initConstant);
```

But, I agree one can just use one `setInsertionPoint`. For example, 

```cpp
    builder.setInsertionPoint(launchOp->getBlock(),
                              --Block::iterator(launchOp));
    auto initConstant = builder.create<arith::ConstantOp>(loc, elementType,
                                                          attr.getInitValue());
    builder.create<gpu::MemsetOp>(loc, mlir::Type{}, mlir::ValueRange{}, arg,
                                  initConstant);
```

Both solution will work. However, the second one results in less code. Ok, I will just update my implementation according to the second case which I wrote above.

external/mlir-hal/include/mlir/Dialect/MHAL/IR/MHALAttrDefs.td
ravil-mobile: It is a good point!

I added something like

```mlir
  let description = [{
      A user interface parameter for initializing specific memory buffers

      The attribute is a way to describe which specific kernel parameter
      (typically memref) must be initialized with specific value before
      a kernel invocation.

  }];
```

krzysz00 (approved):
This all looks good to me!

ravil-mobile:
@manupak , @krzysz00 , @sjw36 @giuseros 

The current problem with the CI is the following. `external/mlir-hal/include/mlir/Dialect/MHAL/Transforms/Passes.h` has been changed. We use `git diff`, compare changes with `origin/develop` branch and check only modified places.
 
I managed to reproduce the bug locally. Here is the command which is failing
 
```
clang-tidy -extra-arg=-std=c++17 external/mlir-hal/include/mlir/Dialect/MHAL/Transforms/Passes.h
```
 
The fix is 
```
clang-tidy -extra-arg=-std=c++17 external/mlir-hal/include/mlir/Dialect/MHAL/Transforms/Passes.h -- -I./external/mlir-hal/include -I./external/llvm-project/mlir/include/ -I./external/llvm-project/llvm/include 
``` 

The problem is that `Passes.h` is not in a compilation database because it is a header. Therefore, `clang-tidy` doesn't know which compile options to apply to this file (incling the list of include directories)


Here is what I can propose. I can add the following

```python
        if file_name.endswith('h') or file_name.endswith('hpp'):
          print(f'standalone header file is ignored `{file_name}`')
          continue

```

to 

https://github.com/ROCm/rocMLIR/blob/a71bfe561252f68079552831dd0e13701bdcef3a/mlir/utils/jenkins/static-checks/premerge-checks.py#L135-L144

What do you think?

krzysz00:
I don't think that's right, since we've had `clang-tidy` running successfully on header files before? So maybe there's something about the external directory or some other set of ignore settings that hasn't been updated.

krzysz00:
Overall, we might want to go poke upstream, now that they're using Github, and steal their clang-format checker.

ravil-mobile:
> I don't think that's right, since we've had `clang-tidy` running successfully on header files before? So maybe there's something about the external directory or some other set of ignore settings that hasn't been updated.

I don't know how it used to work before. Agree, it is strange that I've just bumped into this problem now.

I did one example: I took a random header file from our project and just add a declaration of a dummy function `foo` 

```bash
$ git diff mlir/include/mlir/Dialect/Rock/IR/Rock.h
diff --git a/mlir/include/mlir/Dialect/Rock/IR/Rock.h b/mlir/include/mlir/Dialect/Rock/IR/Rock.h
index 83c7481af799..8c72676dafca 100644
--- a/mlir/include/mlir/Dialect/Rock/IR/Rock.h
+++ b/mlir/include/mlir/Dialect/Rock/IR/Rock.h
@@ -61,6 +61,8 @@ ArrayAttr getIndexArrayAttr(Builder &b, ArrayRef<int64_t> values);
 } // end namespace rock
 } // end namespace mlir

+std::string foo();
+
 #include "mlir/Dialect/Rock/IR/RockAccelTuningParamAttrInterface.h"
 #include "mlir/Dialect/Rock/IR/RockTuningParamAttrInterface.h"
 #define GET_ATTRDEF_CLASSES
```

Then, I ran our `premerge` script and ended up with the same error

```bash
$ cd <rocmlir>
$ cp ./build/compile_commands.json .
$ python3 ./mlir/utils/jenkins/static-checks/premerge-checks.py
Running linters against base commit : origin/develop
clang-format did not modify any files

clang-tidy found error: mlir/include/mlir/Dialect/Rock/IR/Rock.h:16:10: error: 'mlir/Dialect/Func/IR/FuncOps.h' file not found [clang-diagnostic-error]
clang-tidy found 1 errors and 0 warnings.
```

mlir/utils/jenkins/static-checks/clang-tidy.ignore
```diff
@@ -4,4 +4,5 @@ mlir/examples/standalone
 mlir/test
 mlir/utils/performance/ck-benchmark-driver
 build/
+external
```
krzysz00: The suggestion was that you go manually fire off a build of the CI with `ignoreExternalLinting` checked.

ravil-mobile: Ok. reverted the change

PR #1365, opened 2024-01-08
Fix incorrect logic around stores to length-1 memrefs

krzysz00:
The code in #1211 was incorrect and led to colflicting stores to memref<1x1x...x1xT>s that were padded out because it erroniously ignored the validity bit.

This commit fixes the logic to always emit bounds checks in the presence of non-trivial validity logic, even if the final coordinate is 0D. This also improves the handling of such bounds checks by adding expand_shape calls to create a coordinate that the buffer bounds check trick can operate on.

(Honestly, it's surprising that this didn't come up with more failures.)

Fixes https://github.com/ROCm/rocMLIR-internal/issues/1321

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
@@ -903,6 +903,18 @@ static Value asGlobal(PatternRewriter &b, Value memref) {
                                                    memref);
 }
 
+/// The buffer coordinate oob trick requires having at least one coordinate, so
+/// this function takes 0D memrefs and adds a single unit-length dimension
+/// so all the IR validates.
+static Value zeroDMemrefAsOneD(PatternRewriter &b, Value memref) {
+  auto type = cast<MemRefType>(memref.getType());
+  auto oneDType = MemRefType::get({1}, type.getElementType(), nullptr,
+                                  type.getMemorySpace());
+  ArrayAttr expansions = b.getArrayAttr({});
+  return b.createOrFold<memref::ExpandShapeOp>(memref.getLoc(), oneDType,
+                                               memref, expansions);
+}
+
 struct GlobalLoadRewritePattern : public OpRewritePattern<GlobalLoadOp> {
   using OpRewritePattern<GlobalLoadOp>::OpRewritePattern;
   LogicalResult matchAndRewrite(GlobalLoadOp op,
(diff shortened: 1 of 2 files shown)


mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
```diff
@@ -935,8 +946,13 @@ struct GlobalLoadRewritePattern : public OpRewritePattern<GlobalLoadOp> {
     bool useBufferOps = !hasI64Idx && (numBytes.trunc(32).isNegative() ||
                                        emitOobChecks || op.getCanReadOffEnd());
 
-    if (!useBufferOps)
+    if (!useBufferOps) {
       source = asGlobal(b, source);
+    } else if (useBufferOps && emitOobChecks && coords.empty()) {
```
sjw36, line 951: nit: redundant check for useBufferOps

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
```diff
@@ -962,11 +978,8 @@ struct GlobalLoadRewritePattern : public OpRewritePattern<GlobalLoadOp> {
         Value zeroConstantOp = b.createOrFold<ConstantIndexOp>(loc, 0);
         for (Value &c : MutableArrayRef<Value>(coords).drop_back())
           c = b.create<arith::SelectOp>(loc, valid, c, zeroConstantOp);
-        if (!coords.empty()) {
-          Value &lastCoord = coords.back();
-          lastCoord =
-              b.create<arith::SelectOp>(loc, valid, lastCoord, numElems);
-        }
+        Value &lastCoord = coords.back();
```
sjw36, line 981: nit: Value is a pointer, so just copy.

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
```diff
@@ -1111,19 +1123,12 @@ struct GlobalStoreRewritePattern : public OpRewritePattern<GlobalStoreOp> {
...
-    if (isAtomic && coords.empty() &&
-        (!isAlwaysValid || op.getCanStoreOffEnd())) {
-      useBufferOps = false;
-      emitOobChecks = true;
-    }
-
-    if (!useBufferOps)
+    if (!useBufferOps) {
       dest = asGlobal(b, dest);
+    } else if (useBufferOps && emitOobChecks && coords.empty()) {
```
sjw36, line 1128: same as above

sjw36 (approved):
Looks good. Just a few nits.

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
krzysz00: I specifically want the mutability there.

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
krzysz00: I left the redundant check in for clarity and to make it easier to follow to the nested branch that relies on this property

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
sjw36: I don't follow. It is only used as an input to create.

PR #1367, opened 2024-01-09
FP8 truncation in software

krzysz00:
Rename the fp8-ext-to-tables to emulate-fp8-ext-trunc to be more accurate to what it's doing

fp8 truncation is implemented by generating fp8 truncation functions
for a given fp8 type following the IR created by running MIGraphX's
truncation functions (with template branches we don't want removed)
through -O3 on godbolt and manually inputting the LLVM IR as arith
dialect code while re-generalizing the constants.

I've commited all the intermediate sources so people can try to
reproduce this in the future.

(No, I don't understand what -O3 did, I just wanted something less
paintful than trying to SSA-ify some very imperative C++ by hand.)

This commit also includes integration tests that verify (on the host)
that the truncation function does what it's expected to do.

Fixes https://github.com/ROCm/rocMLIR-internal/issues/1271

And revert the old rocmlir-gen hotfix that disabled using random f8 values for our tests.

### DIFF
--- a/mlir/docs/fnuz-float-software-truncation-sources/fp8-trunc-O3.ll
+++ b/mlir/docs/fnuz-float-software-truncation-sources/fp8-trunc-O3.ll
@@ -1,0 +1,205 @@
+$unsigned char cast_to_f8<2u, 5u>(float) = comdat any
+
+$unsigned char cast_to_f8<3u, 4u>(float) = comdat any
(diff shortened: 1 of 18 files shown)


manupak (approved):
I trust you with LLVM-IR conversion ... and I dont see major issues!

PR #1372, opened 2024-01-15
Fix threadwise_read_into vector reads/writes

manupak:
Currently our threadwise_read_into lowering assumes that srcVecLen could only be >1 iff the source memref is vector typed.
This assumption is not correct because our vectorization widgets, will find infer a vector length for scalar memrefs.

This commit fixes this by correcting the bounds and amending the index to be as if a vector index.
Then, it will find a common vector length to transfer from reads and stores if both source and dest uses vectors.

closes : https://github.com/ROCm/rocMLIR-internal/issues/1328

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/ThreadwiseGemmLowering.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/ThreadwiseGemmLowering.cpp
@@ -656,7 +656,7 @@ LogicalResult ThreadwiseReadIntoRewritePattern::matchAndRewrite(
                 b.createOrFold<arith::ConstantIndexOp>(loc, commonVecLen * i);
             Value value = ifb.getResult(0);
             // Only need to perform slice extraction of vector-typed sources.
-            if(isSrcVectorBuffer){
+            if(vectorSrcLen > 1){
               value =
                   b.create<ExtractSliceOp>(loc, commonVecType, value, loadSliceStart);
             }
--- a/mlir/lib/Dialect/Rock/Transforms/ThreadwiseGemmLowering.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/ThreadwiseGemmLowering.cpp
@@ -562,7 +562,7 @@ LogicalResult ThreadwiseReadIntoRewritePattern::matchAndRewrite(
     dstVectorType = dstBufferType.getElementType().dyn_cast<VectorType>();
     vectorDstLen = dstVectorType.dyn_cast<VectorType>().getNumElements();
     numValues = numValues * vectorDstLen;
-    if(isSrcVectorBuffer){
+    if (isSrcVectorBuffer) {
       numValues = numValues / vectorSrcLen;
     }
   }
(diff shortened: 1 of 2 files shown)


mlir/lib/Dialect/Rock/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -547,6 +547,7 @@ LogicalResult ThreadwiseReadIntoRewritePattern::matchAndRewrite(
   if (isSrcVectorBuffer) {
     loadType = elementType;
     vectorSrcLen = elementType.dyn_cast<VectorType>().getNumElements();
+    elementType = elementType.dyn_cast<VectorType>().getElementType();
```
krzysz00, line 550: Refactoring nit: instead of these `isa`/`dyn_cast` pairs (which should be `cast` here anyway), we could instead go `VectorType maybeSrcBufferVectorType = dyn_cast<VectorType>(...)` above

krzysz00 (approved):
Having stared at this, I'm convinced that it is, in fact, a bugfix, so let's land it.

That being said, I just now realized that we aren't correctly handling reads from global memory into vector-typed destinations ... but that's not something we do so we might just want to add a quick "fail for being unsupported" message while we're here.

manupak:
"we aren't correctly handling reads from global memory into vector-typed destinations global memory into vector-typed destinations ... but that's not something we do" 

This is exactly what Im doing in (which I will PR tommorow) to load gemm operands directly into regs (bypassing LDS). So therefore Im super interested to know why -- which is in fact working in my branch. So Im happy to fix while Im here.

krzysz00:
Specifically, I claim that a threadwise_read_into from `%src : memref<...xT>` to `%dest : memref<Nxvector<KxT>>` will only write to indices `0, vectorLen(%src), 2*vectorLen(%src), ...` because it just lowers to a `global_load` to an `in_bounds_store`, and `in_bounds_store` doesn't do any sort of correction for vector-valued indices (unless transfer_write does, but I'd be surprised by that)

mlir/test/Dialect/Rock/lowering_threadwise_fusion_ops.mlir
```diff
@@ -157,3 +157,57 @@ func.func @threadwise_write_all_extra_idx(%source: memref<32xf32, #gpu.address_s
...
+    // CHECK-DAG: %[[baseElOffset2SliceOffset:.+]] = arith.addi %[[baseElOffset2]], %c4 : index
+    // CHECK-DAG: %[[destElOffset2:.+]] = arith.divui %[[baseElOffset2SliceOffset]], %c4 : index
+    // CHECK-DAG: memref.store %[[ldvalSlice2]], %[[dest]][%[[destElOffset2]]] : memref<8xvector<4xf16>, #gpu.address_space<private>>
+  rock.threadwise_read_into {forceUnroll, useIndexDiffs} [](%source)[] -> %dest : memref<4xvector<8xf16>, #gpu.address_space<workgroup>> -> memref<8xvector<4xf16>, #gpu.address_space<private>>
+  func.return
+}
+
+// CHECK-LABEL: func @threadwise_read_into_scalar_to_vec
+// CHECK-SAME: %[[source:.+]]: memref<32xf16, #gpu.address_space<workgroup>>, %[[dest:.+]]: memref<8xvector<4xf16>, #gpu.address_space<private>>
+func.func @threadwise_read_into_scalar_to_vec(%source: memref<32xf16, #gpu.address_space<workgroup>>, %dest: memref<8xvector<4xf16>, #gpu.address_space<private>>) {
```
manupak, line 182: look at the lowering here...

manupak (commented):
So it will load srcVectorLens from the global.
i.e. from  0, vectorLen(%src), 2*vectorLen(%src), ...

Depending on the destination (regs) vector length, it will write to dest in different ways, ranging from a slice of dest vector (i.e. using insert_slice) to multiple dest vectors.

 > because it just lowers to a global_load to an in_bounds_store, and in_bounds_store doesn't do any sort of correction for vector-valued indices (unless transfer_write does, but I'd be surprised by that)

Its not a in_bounds_store at all when the destination is vector typed -- it has not been that even before this PR. It results in a combination of insert_slice ops + multiple memref.store s to update the destintation vector elements -- which are all fine on the basis that these are going to get promoted to SSA eventually (reg memrefs)

manupak:
Ah you meant for *global memory*... we already have a check for that.
https://github.com/ROCm/rocMLIR/blob/acb77bee2e17f211fc1a12188be683476b9753bc/mlir/lib/Dialect/Rock/IR/RockDialect.cpp#L1513-L1515

krzysz00:
.. right, I forgot we already had that disabled

krzysz00 (approved):

PR #1373, opened 2024-01-16
Fix fold broadcast case when the broadcast is on A

giuseros:
Fix https://github.com/ROCm/rocMLIR-internal/issues/1337


krzysz00 (approved):
@giuseros This'll need to be bpckported to the release branch

PR #1374, opened 2024-01-16
[Attention] Load Q tile directly to regs

manupak:
Builds on https://github.com/ROCm/rocMLIR/pull/1372.

This PR enables the loading of tile of Q tensor directly into registers
in the respective accelerator operand layout. This will only be done
if kPerBlock == K where Q is prefetched into the CUs out of the loops.

closes : https://github.com/ROCm/rocMLIR-internal/issues/1262


mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
```diff
@@ -600,6 +600,275 @@ Value MfmaEmitter::wrapLDSBufferForLoad(OpBuilder &b, Location loc,
   return transform(b, buffer, ldsRead);
 }
 
+RegsAsMatrixSubTiles MfmaEmitter::createAccelGemmOperandTransforms(
```
krzysz00, line 603: Should this be used in the GridwiseGemmV2 lowering as well?

manupak: Maybe?

There are a bit of tradeoffs at play here : global vectoriztion, load coalescing vs lds traffic, bank conflicts, occupancy-affecting-wave level HW pipelining. For attention, its a bit LDS bound at the minute and Im not sure about gemms just yet.

krzysz00: Then I'm not understanding what this function does.

I thought it was the (tid, iter) -> (i, j) transform

manupak: This is just creating thread layout for the tensor tiles the accel gemm requires.

Yes it could be used in gemm, it is just there we do more steps : 
1) load into global vectorization & load coalescing friendly thread layout
2) swizzle it back to this layout.

So are you asking why cant we do just 2) with this?
Please file a ticket and I ll put in my pile.

manupak: I thought you are asking "why cant we bypass LDS for gemm kernels ?"

krzysz00: We've got a lot of these utilities that the attention pass uses that could be used in the general gemm pass, and I know there's already a general ticket for that refactoring.

manupak: Yep that is right.

Well at least this one is in a common place (AccelEmitter.cpp), so what is remaining is to do the code to use them in gemm lowering. 

Im a bit purposefully avoiding/delaying that work because Im constantly changing these utils until I have perf on attention kernels. Therefore I think it makes sense to do the refactor last when we have the attention kernels more or less ready -- so that we dont have to carry the burden of changing everything together, if that makes sense.

krzysz00: Yeah, entirely fair.

Which probably explains why the transform slicer hasn't happened yet.

krzysz00 (approved):
Approved as I don't see anything obviously wrong here

mlir/lib/Dialect/Rock/Transforms/AccelEmitter.cpp
manupak: that and I think @ravil-mobile is busy with split-k stuff. https://github.com/ROCm/rocMLIR-internal/issues/1303

PR #1378, opened 2024-01-17
Set llvm.inreg on our kernel arguments to enable argument preloading

krzysz00:
Fixes https://github.com/ROCm/rocMLIR-internal/issues/1338

Review of generated assembly via --arch gfx942 shows that there are no longer loads of arguments from global memory at the beginning of the kernel.


giuseros (approved):
LGTM

PR #1379, opened 2024-01-18
Improve multibuffering for private space

giuseros:
This closes https://github.com/ROCm/rocMLIR-internal/issues/1286

Main things I have done:
- Refactored the multi-buffer pass to use multiple buffers + select, instead of a single big buffer + indexing. While it is not super beneficial for performance, I think this is a nice clean-up and does not hurt performance. 
- Improved the pipeline pass. Now if there are RAW&&private dependencies in the parallel stages of the pipeline, we can swap the stages to make them WAR and avoid using multiple buffers. This fixes the GlobalRead -> LDSWrite issue, because the GlobalRead is acting as a writer of a buffer, while the LDS is acting like a reader of it. If we swap them, we don't need multiple buffers.

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
@@ -320,10 +320,10 @@ void createSchedule(SmallVector<rock::StageOp> &stages,
     // the stages. In this way, we don't need multibuffers (i.e., we read the
     // buffer first and then we write into it). From the point of view of the
     // stages, they don't care because they belong to different iterations. In
-    // theory this could be applied to any buffer, but for proper memory, this
+    // theory this could be applied to any buffer, but for LDS memory this
     // can be more expensive (i.e., you need barriers)
-    DenseMap<int, SmallVector<int>> swapCandidates;
-    DenseMap<int, SmallVector<int>> swapCandidatesR;
+    DenseMap<unsigned, SmallVector<unsigned>> swapCandidates;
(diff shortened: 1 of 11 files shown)


mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -293,33 +298,96 @@ void createSchedule(SmallVector<rock::StageOp> &stages,
...
+      thisMultiBuffers[alloc] = 1;
+    }
+
+    // Optimization: if there is a RAW register dependency (addrspace(5)) swap
+    // the stages. In this way, we don't need multibuffers (i.e., we read the
+    // buffer first and then we write into it). From the point of view of the
+    // stages, they don't care because they belong to different iterations. In
+    // theory this could be applied to any buffer, but for proper memory, this
+    // can be more expensive (i.e., you need barriers)
+    DenseMap<int, SmallVector<int>> swapCandidates;
```
krzysz00: Vague nit: might be a decent idea to go for `unsigned` here? Using `unsigned` for parameter indexing and the like seems pretty common in LLVM code. But that's not critical.

krzysz00 (approved):
This all seems reasonable to me.

Does this also lead to using multiple buffers for multi-buffered LDS?

manupak (approved):
LGTM!

manupak:
Is this enabled in our pipeline ? if so might worth doing a quick perf check with last weekly db.

giuseros:
> Is this enabled in our pipeline ? if so might worth doing a quick perf check with last weekly db.

No this is only enabled if stages share buffers, and in our default pipeline this is not the case

PR #1381, opened 2024-01-18
[DO NOT SQUASH] Legalize math operations on fp8 properly

krzysz00:
During 6.1, I did a quick hack to `-arith-emulated-unsupported-floats` to make stuff like `math.sin %x : f8E4M3FNUZ` work. Since then, I added a proper upstream pass (`-math-legalize-to-f32`, which matches what the GPU to ROCDL lowering would do anyway) and am now backporting it, including reverting the old hack.


giuseros (approved):
LGTM

PR #1383, opened 2024-01-19
Add --fused option to the benchmark driver

giuseros:
This is adding the possibility to test fusion in CK. In CK we need to specify which fusion we want at compile time. So, unless we don't want to go through the hurdle of jit compiling, we have to stick to a specific fusion. I picked the one whose tuning DB is used in MIGraphx: `FastGelu(A*B+D0+D1)`.


krzysz00 (commented):
Would it be reasonable to make the option name longer to include what the kernel is?

krzysz00 (approved):
Looks good to me!

PR #1387, opened 2024-01-26
Create CODEOWNERS file to resolve SWDEV-436548

jerryyin:
For detailed context please refer to https://ontrack-internal.amd.com/browse/SWDEV-436548

In this initial PR I'm including everyone. In the future we can add finer granular control according to example in [pytorch codeowners](https://github.com/ROCmSoftwarePlatform/pytorch/blob/main/CODEOWNERS)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @ravil-mobile @krzysz00 @jerryyin @giuseros @manupak @sjw36 @yiqian1 @djramic @pcf000 @stefankoncarevic


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @ravil-mobile @krzysz00 @jerryyin @giuseros @manupak @sjw36 @yiqian1 @djramic @pcf000 @stefankoncarevic
```
krzysz00, line 1: Let's add `/external/llvm-project` on the next line to indicate that we don't own upstream.

krzysz00 (approved):

.github/CODEOWNERS
jerryyin: If an external developer want to create a PR to our repo (either in /external or not), one of our "code owners" are still responsible for reviewing it. That's my understanding of how CODEOWNERS work, it is for poking right developers to do code reviews. Since we haven't assigned code owners to that granularity, I propose that we just skip this for now.

PR #1388, opened 2024-01-26
[Attention] Bypass LDS-based swizzling for second gemm's non-V input

manupak:
This PR conditionally enables the optimization that would avoid LDS-based swizzling for some cases of MFMA.
The detailed analysis is written here : https://github.com/ROCm/rocMLIR-internal/issues/1201

In a summary, this works only for MFMA in KReduction mode (because there is no KReduction mode in WMMA) where we could match the operand thread layout with the MFMA output thread layout -- so that they could be used back-to-back without needing a LDS-based swizzling in the middle.

In order to keep the op verifiers calculate correct LDS requirement and also to enable rock/utility/ to be used everywhere where Rock IR is used (i.e. including rock/IR itself), I have moved all the utilities into rock/IR folder. This was required for gridwise_attention op verifier to calculate the correct LDS usage.

closes : https://github.com/ROCm/rocMLIR-internal/issues/1201


krzysz00:
On the code organization front, I'd prefer that the LDS availability check be done in a lowering pass instead of going to `::verify`, which is a hook that gets called a lot.

Especially since having the utility directory being abstractions on top of the ops is pretty common MLIR practice.

manupak:
I can move the LDS check to lowering;
(that would remove the requirement of moving utilities at all)

However, I feel (not feeling strongly) we can keep the utilities in rock/IR -- well there are some arbitary utilities already there.
e.g. : MfmaInsnGroup.cpp -- just to keep them in one place.

krzysz00:
I think the MfmaInstnGroup ended up there because it was being used by tuning, which was historically getting run in a dialect-free context.

manupak:
A problem for another day! I ll move the LDS check back to lowering and revert the file moves

manupak:
ok moved utilities back to utilities/ as well as lds size check to lowering..
(I also did move AccelEmitter which to me is also a utility)

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -224,7 +224,7 @@ def Rock_AttentionOp :
   )>,
   Results<(outs Optional<TensorOf<[F32, F16]>>:$result)> {
   let summary = "Attention operation of transformer models";
-  let description = [{
+  let description = [{G
```
krzysz00: Stray character

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1535,6 +1541,57 @@ struct GridwiseAttentionAccelRewritePattern
...
+        op.getQueries().getType().cast<MemRefType>().getElementType();
+    Type elemTypeK = op.getKeys().getType().cast<MemRefType>().getElementType();
+    StringRef arch = op.getArch();
+    RockAccelTuningParamAttrInterface gemm0TuningParams = op.getParams0();
+    auto accelEmitterPtrGemm0 = rock::accel::AccelEmitter::select(
+        op.getFeatures(), elemTypeQ, elemTypeK, arch, gemm0TuningParams);
+    bool isMFMA = bitEnumContainsAny(op.getFeatures(), GemmFeatures::mfma);
+    if (isMFMA) {
+      auto mfmaEmitter =
+          static_cast<rock::accel::MfmaEmitter *>(accelEmitterPtrGemm0.get());
```
krzysz00: Instead of `static_cast` here, can we instead opt `MfmaEmitter` into the LLVM `cast<>` and `dyn_cast<>` infrastructure, please?

krzysz00 (commented):
This all looks reasonable overall! (Or, at least, it looks reasonable in the context of the attention code, which seems to have gained a lot of sprawl during all this perf iteration and could use a fresh pair of eyes once it's settled down)

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: good catch!

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
manupak: ok, adopted LLVM RTTI into AccelEmitter and friends.

mlir/include/mlir/Dialect/Rock/IR/AccelEmitter.h
```diff
@@ -153,10 +150,20 @@ struct AccelEmitter {
 
   virtual ~AccelEmitter() {}
 
+  enum AccelEmitterKind { AEK_MFMAEmitter, AEK_WMMAEmitter };
```
krzysz00, line 153: We could probably `enum class` this, but that's not a blocker.

krzysz00 (approved):
Overall, seems fine

PR #1389, opened 2024-01-27
Add type casts for mixed input types in GemmToGridwise

yiqian1:


krzysz00 (approved):
Looks good to me!

manupak:
Code looks good... though dont we need at least a single e2e test ?

yiqian1:
> Code looks good... though dont we need at least a single e2e test ?

I'll include an e2e test in my next PR that adds mixed-types to rocmlir-gen

PR #1390, opened 2024-01-30
Update rocBLAS benchmark driver

stefankoncarevic:
- Removed get_int8_x4_format function
- Set rocblas_flags value directly to rocblas_gemm_flags_none


giuseros (approved):
LGTM

PR #1391, opened 2024-01-30
Add the possibility to re-multibuffer

giuseros:
In theory this means we can start pipelining nested loops. 

Before:
- Create the schedule and collect the buffer factor
- Create multiple buffers
- Pipeline the loop

Now:
- Multibuffer by `1` all the allocs we can 
- for every loop
-- Create the schedule and collect the buffer factor
-- Pipeline the loop
- Create multiple buffers

I am attaching a perf diff, which is flat (only a glitch, that I was not able to reproduce)
[perf_diff.zip](https://github.com/ROCm/rocMLIR/files/14100463/perf_diff.zip)

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/Transforms/RockMultibuffer.h
+++ b/mlir/include/mlir/Dialect/Rock/Transforms/RockMultibuffer.h
@@ -16,23 +16,23 @@
 namespace mlir {
 
 namespace rock {
-FailureOr<SmallVector<rock::GpuAllocOp>> multiBuffer(RewriterBase &rewriter,
-                                                     rock::GpuAllocOp allocOp,
-                                                     unsigned multiplier,
-                                                     bool skipOverrideAnalysis);
-
-FailureOr<SmallVector<rock::GpuAllocOp>> multiBuffer(rock::GpuAllocOp allocOp,
-                                                     unsigned multiplier,
-                                                     bool skipOverrideAnalysis);
+LogicalResult multiBuffer(RewriterBase &rewriter, rock::GpuAllocOp allocOp,
+                          SmallVectorImpl<rock::GpuAllocOp> &newAllocs,
+                          unsigned multiplier, bool skipOverrideAnalysis);
 
 FailureOr<SmallVector<rock::GpuAllocOp>>
-updateMultiBuffer(RewriterBase &rewriter, Location loc,
-                  ArrayRef<rock::GpuAllocOp> multiBuffer,
-                  unsigned newMultiplier);
+multiBuffer(rock::GpuAllocOp allocOp,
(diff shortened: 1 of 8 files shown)


mlir/include/mlir/Dialect/Rock/Transforms/RockMultibuffer.h
```diff
@@ -16,14 +16,24 @@
...
 namespace rock {
-FailureOr<SmallVector<Value>> multiBuffer(RewriterBase &rewriter,
-                                          rock::GpuAllocOp allocOp,
-                                          unsigned multiplier,
-                                          bool skipOverrideAnalysis);
-
-FailureOr<SmallVector<Value>> multiBuffer(rock::GpuAllocOp allocOp,
-                                          unsigned multiplier,
-                                          bool skipOverrideAnalysis);
+FailureOr<SmallVector<rock::GpuAllocOp>> multiBuffer(RewriterBase &rewriter,
```
krzysz00: Nit: `LogicalResult` and `SmallVectorImpl<GpuAllocOp>&`  to make it more obvious this is allocated on the caller's stack?

mlir/lib/Dialect/Rock/Transforms/RockMultibuffer.cpp
```diff
@@ -276,25 +296,67 @@ mlir::rock::multiBuffer(RewriterBase &rewriter, rock::GpuAllocOp allocOp,
...
+      newAllocs.push_back(rewriter.create<rock::GpuAllocOp>(
+          loc, allocOp.getType(), ValueRange{}, allocOp->getAttrs()));
+    SmallVector<Value> startingValues;
+    for (auto alloc : newAllocs) {
+      startingValues.push_back(alloc);
+    }
+    replaceUsesAndPropagateType(rewriter, loc, allocOp, startingValues,
+                                nullptr);
+  } else if (allocs.size() > newMultiBufferingFactor) {
+    for (unsigned i = 0; i < allocs.size() - newMultiBufferingFactor; i++)
```
krzysz00: 1. Why the first allocations, not the last ones?
2. Style nit ` for (unsigned i = 0, e = allocs.size() - newMultiBufferingFactor; i < e; ++i)`

krzysz00 (approved):
This looks like a reasonable extension!

mlir/lib/Dialect/Rock/Transforms/RockMultibuffer.cpp
giuseros: Actually I forgot to add a test for that part, which contained a bug. So, thank you! I applied your change, fixed the bug, and added some more tests

PR #1392, opened 2024-01-31
Remove setting of http_proxy;  trust that Omkar will fix it.

pcf000:
Also update MITuna to come from ROCm org.


yiqian1:
Should http_proxy in "Code coverage" be removed as well?

pcf000:
It isn't setting it to a fixed address, so that isn't part of the problem.  It's counting on $http_proxy being set if needed, and passing it to the script.

yiqian1 (approved):

PR #1393, opened 2024-02-01
Moved the grid computation from AffixTuningParameters to a GemmToGridwise

ravil-mobile:
Closes ROCm/rocMLIR-internal#1319


manupak:
I think if you run this pass after GemmToGridwise, wouldnt the tranposes + padding be already done ? 
So just need to look at dimensions of inputs to Gridwise op.

That would also avoid duplicating the padding / transposing logic in yet a another place.

ravil-mobile:
> I think if you run this pass after GemmToGridwise, wouldnt the tranposes + padding be already done ? So just need to look at dimensions of inputs to Gridwise op.
> 
> That would also avoid duplicating the padding / transposing logic in yet a another place.

But, I am running this pass before `GemmToGridwise`. Or you want to put the new pass after `GemmToGridwise`?

manupak:
> But, I am running this pass before GemmToGridwise. Or you want to put the new pass after GemmToGridwise?

I dont see a requirement that it needs to run before GemmToGridwise. Is there any ? 
If not, we could just do this after with much simpler logic.

ravil-mobile:
> > But, I am running this pass before GemmToGridwise. Or you want to put the new pass after GemmToGridwise?
> 
> I dont see a requirement that it needs to run before GemmToGridwise. Is there any ? If not, we could just do this after with much simpler logic.

I can see your point. In theory, yes. It is possible to move `GridConfiguration` pass one step down the pipeline.

I have a concern regarding the software design. Currently, `GemmToGridwise` pass needs `gridSize` because it is a constructor parameter for `GridwiseGemmAccelOp` and `GridwiseGemmOp`

https://github.com/ROCm/rocMLIR/blob/b82208dd29ebfa8d07338180e6687cd460706b15/mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp#L188-L197

So, shall I remove `gridSize` input parameter from  `GridwiseGemmAccelOp` and `GridwiseGemmOp` and set `gridSize` later during `GridConfiguration` pass? How should I set `gridSize` then? Maybe just set it as an attribute to `GridwiseGemmAccelOp` and `GridwiseGemmOp`? What do you think?

The same reasoning goes to `GridwiseAttentionAccelOp`

https://github.com/ROCm/rocMLIR/blob/b82208dd29ebfa8d07338180e6687cd460706b15/mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp#L310-L326

manupak:
I think original idea was for this to be done "in" the `GemmToGridwise` pass : https://github.com/ROCm/rocMLIR-internal/issues/1319. (which I think is the simplest solution)

Im sorry if that was not clear (and I missed that you ve asked the similiar question in the issue 2 days ago coz I was not tagged in the ticket).

TL;DR

I personally dont like the idea of fixing up the op definition so that it could be lowered in subsequent passes. However, in the other hand knowing perf_config + wavesize, one could technically calculate the gridsize & blocksize at any point of time. There is another alternative is we can add interface methods to gridwise ops derive gridsize & blocksize using perf_config + wavesize. However, when we do the padding -- we are kind of "commiting" to a certain gridsize. So by the virtue of that earliest point of commiting to a gridsize is in `GemmToGridwise`

ravil-mobile:
@manupak @krzysz00 @giuseros @sjw36 

Can we decide whether `blockSizeOverride` and `gridSizeOverride` should be removed or not.

https://github.com/ROCm/rocMLIR/blob/1cd96bcbaed57256f173eb4b9bfd2691a43f8b0a/mlir/include/mlir/Dialect/Rock/Passes.td#L31-L34

I have some understanding why `blockSizeOverride` should be preserved. I think there might be some scenario where it can be useful. But `gridSizeOverride` looks completely unnecessary. I cannot see a scenario when a user wants to select some specific grid size. My opinion is to completely remove `gridSizeOverride`.

What do you think?

Btw, I haven't found any reference to `blockSizeOverride` or `gridSizeOverride` in rocMLIR.

ravil-mobile:
Hi, @manupak 

> I personally dont like the idea of fixing up the op definition so that it could be lowered in subsequent passes. However, in the other hand knowing perf_config + wavesize, one could technically calculate the gridsize & blocksize at any point of time. 

So, once you get `perfConfig` your `blockSize` is determined. I cannot see a point why some would want to change the block size. If for some reason one wants to change the block size then `perfConfig` must be updated accordingly, right? Thus, I think that it doesn't make sense to move computations of `blockSize` from `AffixTuningParameters` pass

The `gridSize` is a different beast. One may want to  adjust grid size for Split-K or Strem-K

manupak:
> So, once you get perfConfig your blockSize is determined. 

Hmmm.. that is based on the inferring knowledge about gridwise lowering at a very high level.

Also perfConfig should be interpretted alongside with gridwise_* op not with a higher level op that breaks it down to gridwise_* op -- though I admit we only have this conv to gemm so far, however, we might want to have new gridwise algorithms and preserving the right to interpret perfconfig and translate to that blocksize then and there would be preffered.

manupak:
> I have some understanding why blockSizeOverride should be preserved. 

The only thing I can think of is in when the client does not give a perf config -- we can pseudo-filter the list and pick one that matches the blocksize given.

When we are doing gridwise-gemm-to-blockwise, if the blocksize & perf config are incompatible -- whole thing just blows up.
Same could be said to gridSizeOverride.

ravil-mobile:
@manupak ,

> The only thing I can think of is in when the client does not give a perf config -- we can pseudo-filter the list and pick one that matches the blocksize given.

Yes, it is the exact reason. See, here
https://github.com/ROCm/rocMLIR/blob/1cd96bcbaed57256f173eb4b9bfd2691a43f8b0a/mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp#L220-L253

After line 235

ravil-mobile:
Hi @manupak , @krzysz00 and @giuseros ,

I fixed all failing tests. Can I ask you to review the changes?

ravil-mobile:
Hi @manupak , @krzysz00 and @giuseros ,

I removed the new pass and merged it with `GemmToGridwise`

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
```diff
@@ -309,16 +311,14 @@ class PopulateParamsAccel : public BasePopulateParams<InitParamsAccel> {
...
-                                       uint32_t &blockSize, uint32_t &gridSize,
-                                       int64_t &gemmKBlocks);
+                                       uint32_t &blockSize);
 
-  virtual LogicalResult obtainTuningParameters(
-      const PopulateParamsInfo &info, uint32_t blockSizeOverride,
-      const std::string &perfConfig, InitParamsAccel &validParams,
-      uint32_t &blockSize, uint32_t &gridSize, int64_t &gemmKBlocks);
+  virtual LogicalResult obtainTuningParameters(const PopulateParamsInfo &info,
+                                               const std::string &perfConfig,
```
krzysz00: Can we make this a `StringRef` while we're here?

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
```diff
@@ -223,6 +235,49 @@ GemmRewritePattern::matchAndRewrite(GemmOp op, GemmOpAdaptor adaptor,
...
+      if (!bwdOp.getKBlocks().has_value()) {
+        LLVM_DEBUG(llvm::dbgs() << "must have kBlocks set at lowering.\n");
+        return failure();
+      }
+      int64_t gemmKBlocks = bwdOp.getKBlocks()->getZExtValue();
+      gridSize *= gemmKBlocks;
+    }
+  } else {
+    auto tuningParams = params.cast<GeneralGemmParamsAttr>();
+    gridSize = obtainGridSize(tuningParams);
```
manupak: This should be already be part of G as per 
https://github.com/ROCm/rocMLIR/blob/fb40071f408a5b1665f6846876c58d0210e8c710/mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp#L586

This seems like over counting...

manupak: as in there is no need for specially accounting for kblocks here.

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
```diff
@@ -326,6 +385,43 @@ AttentionRewritePattern::matchAndRewrite(AttentionOp op,
...
 }
 
+LogicalResult
+AttentionRewritePattern::computeGridSize(ConversionPatternRewriter &rw,
+                                         AttentionOp op, Value queries,
+                                         Value keys, Value values) const {
+
+  RockAccelTuningParamAttrInterface accelParams0 =
+      op.getParams0Attr().cast<RockAccelTuningParamAttrInterface>();
+  //
```
manupak: stray

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -161,18 +169,8 @@ LogicalResult PopulateParams::calculateBlockGemmPerformanceParameters(
...
-      calculatePadding(params.gemmKPerBlock, params.gemmMPerBlock,
-                       params.gemmNPerBlock, gemmSize);
-  if (gemmExtraPad.has_value()) {
-    gemmSize.m += gemmExtraPad->m;
-    gemmSize.k += gemmExtraPad->k;
-    gemmSize.n += gemmExtraPad->n;
-  }
-
+LogicalResult
+PopulateParams::populateDerived(const InitParamsNonAccel &params) {
```
manupak, line 173: why do we need this function now ? It feels like redundant wrapper right ?

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -198,9 +195,7 @@ PopulateParams::getGemmParamsAttr(OpBuilder &b,
 LogicalResult
 PopulateParams::paramsProbablyValid(const PopulateParamsInfo &info,
                                     const InitParamsNonAccel &params) {
-  uint32_t gridSize;
-  GemmSize newGemmSize = info.gemmSize;
-  return populateDerived(params, newGemmSize, gridSize);
+  return populateDerived(params);
```
manupak, line 198: same could be said here... its just a wrapper ?

manupak (commented):
Broadly looking good except for one critical comment about gridsize calculation for split-k case.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
manupak: feel free to ignore on the basis... non_accel path has a short life time.

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
```diff
@@ -223,6 +235,49 @@ GemmRewritePattern::matchAndRewrite(GemmOp op, GemmOpAdaptor adaptor,
...
+    return (M / tuningParams.getMPerBlock()) *
+           (N / tuningParams.getNPerBlock()) * G;
+  };
+
+  int64_t gridSize{0};
+  if (isAccel(features)) {
+    auto tuningParams = params.cast<RockAccelTuningParamAttrInterface>();
+    gridSize = obtainGridSize(tuningParams);
+
+    if (auto bwdOp = llvm::dyn_cast<Conv2DBwdWeightOp>(*op)) {
```
krzysz00: Wait, how isn't trivially always false? `op` is always a `GemmOp`

I think at this point in the lowering the blocks have already been added to the `G` dimension so you don't need this.

mlir/test/Dialect/Rock/affix_tuning_params.mlir
```diff
@@ -18,10 +19,12 @@
 // CHECK-DAG: #[[$XDLOPS_PARAMS_6:.*]] = #rock.xdlops_gemm_params<kpackPerBlock = 16, mPerBlock = 4, nPerBlock = 64, kpack = 1, mPerWave = 4, nPerWave = 64, forceUnroll = true>
 // CHECK-DAG: #[[$XDLOPS_PARAMS_7:.*]] = #rock.xdlops_gemm_params<kpackPerBlock = 4, mPerBlock = 128, nPerBlock = 128, kpack = 8, mPerWave = 64, nPerWave = 128, forceUnroll = true>
 // CHECK-LABEL: @rock_conv2d
+// GRID: rock_conv2d
```
krzysz00: GRID-LABEL: @rock_conv2d

krzysz00 (commented):
Overall looks right, just a few minor notes

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
manupak: i just noticed its non-accel ... so we can ignore for now.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
krzysz00: 1. Validation
2. We might end up getting more derived parameters in the future, let's keep the structure in there.

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
ravil-mobile: Hi, yes. It makes total sense. I don't know why I forgot about it. Thanks for noticing!

Done.

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
ravil-mobile: Yes, makes sense. Thanks for noticing.

Done

mlir/test/Dialect/Rock/affix_tuning_params.mlir
ravil-mobile: done

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
ravil-mobile: Yes. Done

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
ravil-mobile: done

manupak (approved):
LGTM!

PR #1395, opened 2024-02-02
Force-add gfx11xx to rocprof in Dockerfile.

pcf000:
We have a problem with rocprof not supporting gfx1100, which is navi3x.  We can't work around it with the existing rocprof, but we can modify rocprof in our docker image.  This patch does that, adding gfx11xx to the list of supported GPUs, while we wait for an official patch.

(rocprofv2, btw, doesn't accept the "--stats" option that we use, and therefore isn't a drop-in replacement.)

### DIFF
--- a/mlir/utils/jenkins/Dockerfile
+++ b/mlir/utils/jenkins/Dockerfile
@@ -134,3 +134,6 @@ ADD "https://raw.githubusercontent.com/ROCmSoftwarePlatform/MITuna/develop/requi
 RUN python3 -m venv /tuna-venv && . /tuna-venv/bin/activate && \
     python3 -m pip install -r tuna-requirements.txt --ignore-installed && \
     python3 -m pip install scipy pandas
+
+# Workaround for 6.0.0 rocprof not supporting navi3x:  add it to list.
+RUN sed --in-place 's/gfx94x")/gfx94x","gfx11xx")/' /opt/rocm/bin/rocprof


mlir/utils/jenkins/Dockerfile
```diff
@@ -134,3 +134,6 @@ ADD "https://raw.githubusercontent.com/ROCmSoftwarePlatform/MITuna/develop/requi
 RUN python3 -m venv /tuna-venv && . /tuna-venv/bin/activate && \
     python3 -m pip install -r tuna-requirements.txt --ignore-installed && \
     python3 -m pip install scipy pandas
+
+# Workaround for 6.0.0 rocprof not supporting navi3x:  add it to list.
+RUN sed --in-place 's/gfx94x")/gfx94x","gfx11xx")/' /opt/rocm/bin/rocprof
```
krzysz00, line 139: Do we have any idea *why* or *to what extent* rocprof doesn't support Navi3x?

krzysz00 (approved):
Approved to unblock us but this needs to be followed up on

mlir/utils/jenkins/Dockerfile
pcf000: Personally, I think it's just an oversight.  The check for supported GPUs did not exist until October.  If it supported gfx11xx before then, it should support it after, unless they removed things.

Practically, that's a question I'll raise in the ticket, if/when I hear back.

PR #1397, opened 2024-02-06
Simplify threadwise_accel_gemm to accept the [i,j,k] compute coordinates

giuseros:
This should be the final change to allow me to decide how the microkernel [m,n,k] loop is going to be scheduled. 

This is stemming from a discussion I had with @manupak few weeks back. 

The idea is to have `threadwise_accel_gemm` to not generate any loop. We provide `viewA[i,k]`, `viewB[j,k]`, and `viewC[i,j]` and a triple of `computeIndices = [i0,j0,k0]`. Those indices are passed down to access the buffers.  and we generate `C[i0,j0] += A[i0,k0]*B[j0,k0]`. 

In this way the user of `threadwise_accel_gemm` can decide the scheduling order without having to change the transformations (and this also simplifies the lowering a bit).


krzysz00 (approved):
Looks good to me, approved!

sjw36 (approved):
Nice cleanup/simplify!

manupak (approved):
Nice!

PR #1399, opened 2024-02-07
[Attention] Implement a proper fusion between first gemm and softmax

manupak:
The current code for generating attention kernels only accepted two optional inputs
that could be used to mul (scale) and add (bias) between the first gemm and the softmax.

However, going forward we need squeeze in a arbitrary dequantization schemes that is
composed of element wise operations. Therefore, as a pre-requisite to that this PR
implements a generic fusion point between the first gemm and softmax.

The said tree is implemented as an inverted tree where the root of the tree is the softmax input
and one leaf being the first gemm output. Other leaves can reach farther as long as elementwise operations
go. Moreover, this PR add support to fused reshape-like ops as well because they are quite common
in graphs we get from MIGraphX that has rank > 3 and we do squeeze them to be rank = 3 around gemms because
tosa only support rank=3 matmuls. 

Future work : 

However, the code done here fundamentally can fuse any operator that lowers
to rock.transform, into the tree with 1 restriction : the path between the first gemm output leaf to the next branching point
in the tree should contain an invertible rock.transform.

closes : https://github.com/ROCm/rocMLIR-internal/issues/1368

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -203,50 +203,6 @@ def Rock_ReduceOp :
   }];
 }
 
-// def Rock_AttentionOp :
-//   Rock_Op<"attention", [AllElementTypesMatch<["queries", "keys", "values", "scale", "bias"]>,
-//                                              AttrSizedOperandSegments]>,
-//   Arguments<(ins
-//     Arg<TensorOrMemRefOf<[F32, F16]>, "queries", [MemRead]>:$queries,
-//     Arg<TensorOrMemRefOf<[F32, F16]>, "keys", [MemRead]>:$keys,
-//     Arg<TensorOrMemRefOf<[F32, F16]>, "values", [MemRead]>:$values,
-//     Arg<Optional<TensorOrMemRefOf<[F32, F16]>>, "scale", [MemRead]>:$scale,
-//     Arg<Optional<TensorOrMemRefOf<[F32, F16]>>, "bias", [MemRead]>:$bias,
-//     Arg<TensorOrMemRefOf<[F32, F16]>, "output", [MemRead, MemWrite]>:$out,
-//     UnitAttr:$qTransposed,
-//     UnitAttr:$kTransposed,
-//     UnitAttr:$vTransposed,
-//     UnitAttr:$oTransposed,
-//     StrAttr:$arch,
-//     Rock_GemmFeaturesAttr:$features,
-//     OptionalAttr<RockTuningParamAttrInterface>:$params0,
-//     OptionalAttr<RockTuningParamAttrInterface>:$params1
-//   )>,
-//   Results<(outs Optional<TensorOf<[F32, F16]>>:$result)> {
-//   let summary = "Attention operation of transformer models";
-//   let description = [{
-//     Performs the operation out = SOFTMAX((queries * keys) .* scale) * values.
-
-//     This operation performs attention mechanism of transformer models.
-
-//     Those creating a `rock.attention` must specify the GPU architecture being targetted
-//     and the number of compute units (numCu) available. The parameters
-//     `gridSize`, and `blockSize` are optional as they can be inferred by
-//     a tuning process or a heuristic, but they must be set before the `attention` is
-//     lowered into the `gridwise_attention` stage of the code generation pipeline.
-
-//     `features` specifies what hardware features can be used in the generated code.
-//   }];
-//   let hasVerifier = 1;
-//   let assemblyFormat = [{
-//     `(` operands `)` `features` `=` $features attr-dict
-//     `:` type(operands) (`->` type($result)^)?
-//   }];
-//   let extraClassDeclaration = [{
-//     ::mlir::OpOperand* getOutArgument() { return &(*this)->getOpOperands().back(); }
-//   }];
-// }
-
 def Rock_AttentionOp :
   Rock_Op<"attention", [AllElementTypesMatch<["queries", "keys", "values"]>]>,
   Arguments<(ins
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -247,6 +247,48 @@ def Rock_AttentionOp :
   }];
 }
 
+def Rock_AttentionOpV2 :
+  Rock_Op<"attentionv2", [AllElementTypesMatch<["queries", "keys", "values"]>]>,
+  Arguments<(ins
+    Arg<TensorOrMemRefOf<[F32, F16]>, "queries", [MemRead]>:$queries,
+    Arg<TensorOrMemRefOf<[F32, F16]>, "keys", [MemRead]>:$keys,
+    Arg<TensorOrMemRefOf<[F32, F16]>, "values", [MemRead]>:$values,
+    Variadic<AnyType>:$preSoftmaxElemWiseInputs,
+    Arg<TensorOrMemRefOf<[F32, F16]>, "output", [MemRead, MemWrite]>:$out,
+    UnitAttr:$qTransposed,
+    UnitAttr:$kTransposed,
+    UnitAttr:$vTransposed,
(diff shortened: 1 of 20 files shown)


mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -444,8 +442,9 @@ def Rock_GridwiseAttentionAccelOp :
   let description = [{
     The `rock.gridwise_attention_accel` op computes gridwise attention with acceleration.
   }];
+  let regions = (region AnyRegion:$preSoftmaxBody);
   let assemblyFormat = [{
-    `(` operands `)` `features` `=` $features attr-dict `:` type(operands)
+    `(` operands `)` `features` `=` $features `preSoftmaxOps` `=` $preSoftmaxBody attr-dict `:` type(operands)
```
manupak, line 454: Ive intentionally left this like this to get ideas for a better assembly format.

krzysz00: Hm ... what's the usual mathematical syntax for attention?

The first observation I have is that you can use optional groups, the syntax being something like
```
(`preSoftmaxArgs` $preSoftmaxArgs^)?
```
or even
```
(`(` $preSoftmaxArgs^`)`)?
```

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -715,6 +717,47 @@ struct CollapseExpandRewritePattern
...
+        tosa::LogicalNotOp,
+        tosa::NegateOp,
+        tosa::ReciprocalOp,
+        tosa::RsqrtOp,
+        tosa::SelectOp,
+        tosa::EqualOp,
+        tosa::GreaterOp,
+        tosa::GreaterEqualOp
+       >(op);
+  // clang-format on
```
krzysz00, line 758: ... I'm suspicious of this, but maybe this is the best we can do? Are we sure there isn't an interface somewhere?

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
sjw36: I like the linalg.generic like approach. Can you follow that format more closely

sjw36 (approved):
Nice generalization!

krzysz00 (commented):
... Huh, my comments didn't submit

mlir/lib/Dialect/Rock/Transforms/BufferizableOpInterfaceImpl.cpp
```diff
@@ -90,9 +91,16 @@ struct GemmLikeInterface
     if (!outBuffer) {
       return op->emitOpError("Couldn't find output argument\n");
     }
+    if (op->getNumRegions() == 1) {
+      auto newOp = cloneWithoutRegions(
+          rewriter, op, /*newResultTypes=*/TypeRange{}, bufferArgs);
+      rewriter.inlineRegionBefore(op->getRegion(0), newOp->getRegion(0),
```
krzysz00: Is this right? Don't we want to clone the region?

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
manupak: https://github.com/ROCm/rocMLIR/blob/4251df1dda5522598e5556179dd4359bd8f007c8/external/llvm-project/mlir/lib/Dialect/Tosa/Transforms/TosaPartition.cpp#L58-L106

I just borrowed it from here... I suspect @sjw36 might know if there is any better way to do this -- if any?

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -3584,17 +3625,18 @@ int main(int argc, char **argv) {
...
 
-      if (failed(pm.run(module))) {
-        llvm::errs() << "failed to apply rocm bufferize pipeline.\n";
-        exit(1);
-      }
+    rock::BufferizeOptions bufferizeOptions;
+    bufferizeOptions.disableRock = true;
+    rock::buildBufferizePipeline(pm, bufferizeOptions);
+
+    if (failed(pm.run(module))) {
```
krzysz00, line 3637: Why're we making this unconditional here?
I'm a little bit dubious about putting this in `rocmlir-gen`. I'd rather rip this bit of code out and stick it as `-host-pipeline bufferization` over in `rocmlir-driver`

mlir/lib/Dialect/Rock/Transforms/BufferizableOpInterfaceImpl.cpp
manupak: I was just basing this off from : https://github.com/ROCm/rocMLIR/blob/4251df1dda5522598e5556179dd4359bd8f007c8/external/llvm-project/mlir/lib/Dialect/Linalg/Transforms/BufferizableOpInterfaceImpl.cpp#L78-L81

but you maybe right, I ll try if we can remove inlineRegionBefore...

krzysz00: Huh, then that might just be it.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
manupak: Previously, we were only using tosa in host harness.. now we are using in rock ops themselves.
I dont see good reason why it needs to be conditional given that when we generating kernel or host code -- we rather prefer them to be buffered -- either by construction and/or lowering.

Im fine either way -- just wanted to keep the contract rocmlir-gen to always produce bufferized IR as it is currently.

krzysz00: Fair enough, reasonable point, let's not go aggressively refactoring things in this PR.

I'm thinking a future PR could be making `rocmlir-gen` output tensor code.

mlir/lib/Dialect/Rock/Transforms/BufferizableOpInterfaceImpl.cpp
manupak: it turned out i could just "clone" it.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: ok, I came up with the following : 

```mlir
    rock.attention{
     qk = %arg0 * %arg1 : memref<1x1024x32xf32>, memref<1x32x1024xf32>
     qk = elementwise otherIns(%arg3 : memref<1x1024x1024xf32>) {
    ^bb0(%arg5: memref<1x1024x1024xf32>, %arg6: memref<1x1024x1024xf32>, %arg7: memref<1x1024x1024xf32>):
      %0 = bufferization.to_tensor %arg5 restrict : memref<1x1024x1024xf32>
      %1 = bufferization.to_tensor %arg6 restrict : memref<1x1024x1024xf32>
      %2 = "tosa.mul"(%0, %1) <{shift = 0 : i32}> : (tensor<1x1024x1024xf32>, tensor<1x1024x1024xf32>) -> tensor<1x1024x1024xf32>
      %3 = bufferization.to_memref %2 : memref<1x1024x1024xf32>
      memref.copy %3, %arg7 : memref<1x1024x1024xf32> to memref<1x1024x1024xf32>
      rock.yield
    }
     %arg4 = softmax(qk) * %arg2 : memref<1x1024x32xf32> -> memref<1x1024x32xf32>
    } {arch = "amdgcn-amd-amdhsa:gfx90a:sramecc+:xnack-", features = #rock<GemmFeatures mfma|dot|atomic_add>}
```

I think it gives a clearer picture of the semantics of the op.

manupak:
I believe i have addressed the comments and Im taking this in to enable reviewing of https://github.com/ROCm/rocMLIR/pull/1411.

 So if you have more comments on this one, I m happy to address them in https://github.com/ROCm/rocMLIR/pull/1411 -- just to avoid the merging/rebasing pain from my side.

PR #1400, opened 2024-02-07
Add the full string of arhitecture to -DGPU_TARGETS.

stefankoncarevic:


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -145,10 +145,10 @@ void postProcessPerfRes(String chip) {
 def get_gpu_architecture() {
     try {
         def result = sh(script: 'rocminfo', returnStdout: true).trim()
-        def arch_pattern = /gfx\d+\w*/
+        def arch_pattern = /Name:\s+amdgcn-amd-amdhsa--(gfx\d+\w*(:sramecc[\+\-])(:xnack[\+\-]))/
```
jerryyin: A few comments on the regular expression:
- You need to make optional groups, it is possible that the string doesn't include sramecc or xnack, and has only one of them
- What is the last escape character '/' for?
- I think existing regular expression groups are defensive, but too restrictive
   - What if the ordering of xnack and sramecc are swapped? We've never seen them before but they'd still be valid strings
   - What if new feature string show up in later architectures?
   
Thinking of all above I might go with: `(gfx\d+\w*((:\w+[\+\-]))*)`

Besides, have you tested manually compile migraphx and test it on a gfx906 assuming you feed it with the right string? If it doesn't fix the issue, we have a bigger discussion to make with migraphx team.

stefankoncarevic: You're right about the order, there is a possibility that they might be swapped or that there is only one of them. So far, I have seen this arrangement, but the regular expression should definitely be more comprehensive, and I will try to adapt it.

jerryyin (approved):

PR #1403, opened 2024-02-08
Fix transpose/collapse_shape soundness issues in TosaToRock

### LINKED ISSUE #2692: ShuffleNet-V2 inaccuracy
[ShuffleNet-V2-12](https://github.com/onnx/models/blob/main/validated/vision/classification/shufflenet/model/shufflenet-v2-12.onnx) has an accuracy issue when running 
`migraphx-driver verify vision/classification/shufflenet/model/shufflenet-v2-12.onnx`

The output:
```bash
FAILED: vision/classification/shufflenet/model/shufflenet-v2-12.onnx
RMS Error: 0.320288
Max diff: 41.1116
Mismatch at 0: -0.151188 != 5.93148
```

With `MIGRAPHX_TRACE_EVAL=2` it is visible that that something goes wrong between the 4th and 5th concat operation, because the outputs of the `gpu` and `ref` are starting to differ from there : 
| |ref|gpu|
|-|-|-|
|4th concat|Output: 0, 0.152998, 0, 0, 0, ..., 0.314251, 0.0573865, 0.174285, 0.111327, 0.31034|Output: 0, 0.152998, 0, 0, 0, ..., 0.314251, 0.0573866, 0.174285, 0.111327, 0.310341|
|5th concat|Output: 0.803316, 0.818251, 0.739967, 0.778224, 0.67973, ..., 0, 0, 0, 0.00404376, 0|Output: 0.724681, 0.793088, 0.846303, 0.865964, 0.794026, ..., 0.721219, 0.661027, 0.640053, 0.680654, 0.321274|

Disabling rocMLIR with `MIGRAPHX_DISABLE_MLIR=1` fixes the issue, and the verification is succesfull.

The related part of the model from netron:
shufflenet_concat

krzysz00:
Fixes https://github.com/ROCm/AMDMIGraphX/issues/2692

Candidate for backport to 6.1.

The problematic transformation arises in the following context, simplified from the original issue slightly:

```
%0 = expand_shape(%arg0, ...) : tensor<7x6x5xf32> to tensor<7x3x2x5xf32>
%1 = transpose(%0, [0, 2, 1, 3]) : tensor<7x3x2x5xf32> to %tensor<7x2x3x5xf32>
%2 = collapse_shape(%1, ...) : tensor<7x2x3x5xf32> to tensor<7x6x5xf32>
```

Previously, our code for rewriting transpose(collapse_shape) to simply collapse_shape - which is important for ensuring we cleanly pattern-match even when we're adding and removing unit dimensions all over the place to match broadcasting requirments or operation specifications - would rewrite such the code such that this entire reshape/transpose/reshape chain was deleted.

That deletion led to incorrect results, because the within-dimension tranposition that that sequence of operations specifies is semantacially menaingful.

Due the complexity and pile of hacks that is
the collapse/trpnspose rewriter, I was unable to fully diagnose exactly how we end up making this mistake in a reasonable time.

However, based on the reasonable assumption that we have this rewrite to account for stray unit dimensions, I have simply added a guard such that collapse_shape operations that perform non-trivial concatentations (like the 2x3 -> 6 one above) cannot have transposes folden into them, as we have no way to ensure that those dimensions are not the results of some previous expand_shape or similar operation.

This change required updating an existing test that relied on the old behavior.

### DIFF
--- a/mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
+++ b/mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
@@ -479,11 +479,30 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
         llvm::SmallDenseSet<int64_t> preTpUnitDims;
         for (ReassociationIndices indices : reassocIndices) {
           ReassociationIndices newReassocIdx;
-          for (size_t i = 0; i < indices.size(); i++) {
+          size_t numNonUnitDimsMerged = 0;
+          for (size_t i = 0, e = indices.size(); i < e; ++i) {
             if (inShape[indices[i]] == 1) {
               preTpUnitDims.insert(dims[indices[i]]);
+            } else {
+              numNonUnitDimsMerged += 1;
             }
             newReassocIdx.push_back(dims[indices[i]]);
+          }
+          if (numNonUnitDimsMerged > 1) {
+            // Per MIGraphX bug #2692, this transpsoe/collaspe swap logic
+            // will be incorrect in cases like the following
+            //   %0 = expand_shape [[0], [1, 2], [3]] %arg0 : tensor<7x6x5xT> to
+            //   tensor<7x3x2x5xT> %1 = transpose %0, [0, 2, 1, 3] :
+            //   tensor<7x2x3x5xT> %2 = collapse_shape [[0], [1, 2], [2]] %1 :
+            //   tensor<7x2x3x5xT> to tensor<7x6x5xT>
+            // by way of creating a trivial expand/collapse pair that isn't
+            // correct.
+            //
+            // Therefore, as a sledgehammer fix, don't handle any cases where
+            // non-trivial collapses are performed.
+            return rewriter.notifyMatchFailure(
+                op, "abandoning attempt to interchange transpose and "
+                    "non-trivial collapse");
           }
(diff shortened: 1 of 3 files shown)


manupak (approved):
LGTM! 

I m thinking long-term solution to this annoying issue that might not be that much of breaking change (i.e. not change problem key).
* trMap utility that performs stride inference when doing untransform.
* then based on the strides, we sort the dimensions at the rock.* op.
   * For gemms: we use that to infer transposition.
   * For convs ; we use that to infer the layout.
* so rather than try to fold the transposes -- we can instead inject one just before rock.* op. so the semantics of the kernel is preserved.

WDYT ?

manupak:
Well I created a ticket to continue the discussion : https://github.com/ROCm/rocMLIR-internal/issues/1371

sjw36:
> LGTM!
> 
> I m thinking long-term solution to this annoying issue that might not be that much of breaking change (i.e. not change problem key).
> 
> * trMap utility that performs stride inference when doing untransform.
> * then based on the strides, we sort the dimensions at the rock.* op.
>   
>   * For gemms: we use that to infer transposition.
>   * For convs ; we use that to infer the layout.
> * so rather than try to fold the transposes -- we can instead inject one just before rock.* op. so the semantics of the kernel is preserved.
> 
> WDYT ?

This would be much cleaner. And rely on simple builtin transpose folding where possible.

PR #1404, opened 2024-02-08
[BACKPORT]  Fix transpose/collapse_shape soundness issues in TosaToRock

manupak:
Fixes https://github.com/ROCm/AMDMIGraphX/issues/2692

Original PR : https://github.com/ROCm/rocMLIR/pull/1403

Original Description :

Candidate for backport to 6.1.

The problematic transformation arises in the following context, simplified from the original issue slightly:

%0 = expand_shape(%arg0, ...) : tensor<7x6x5xf32> to tensor<7x3x2x5xf32>
%1 = transpose(%0, [0, 2, 1, 3]) : tensor<7x3x2x5xf32> to %tensor<7x2x3x5xf32>
%2 = collapse_shape(%1, ...) : tensor<7x2x3x5xf32> to tensor<7x6x5xf32>
Previously, our code for rewriting transpose(collapse_shape) to simply collapse_shape - which is important for ensuring we cleanly pattern-match even when we're adding and removing unit dimensions all over the place to match broadcasting requirments or operation specifications - would rewrite such the code such that this entire reshape/transpose/reshape chain was deleted.

That deletion led to incorrect results, because the within-dimension tranposition that that sequence of operations specifies is semantacially menaingful.

Due the complexity and pile of hacks that is
the collapse/trpnspose rewriter, I was unable to fully diagnose exactly how we end up making this mistake in a reasonable time.

However, based on the reasonable assumption that we have this rewrite to account for stray unit dimensions, I have simply added a guard such that collapse_shape operations that perform non-trivial concatentations (like the 2x3 -> 6 one above) cannot have transposes folden into them, as we have no way to ensure that those dimensions are not the results of some previous expand_shape or similar operation.

This change required updating an existing test that relied on the old behavior.

### DIFF
--- a/mlir/utils/performance/ck-benchmark-driver/ck-benchmark-driver.cpp
+++ b/mlir/utils/performance/ck-benchmark-driver/ck-benchmark-driver.cpp
@@ -12,6 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+#ifdef CK_FOUND
+
 // Include common utility functions
 #include "../common/benchmarkUtils.h"
 
@@ -234,3 +236,5 @@ int main(int argc, char **argv) {
   HIP_ABORT_IF_FAIL(hipFree(cDevice));
   return 0;
 }
+
+#endif
--- a/mlir/utils/performance/ck-benchmark-driver/ck-benchmark-driver.cpp
+++ b/mlir/utils/performance/ck-benchmark-driver/ck-benchmark-driver.cpp
@@ -63,7 +63,8 @@ struct GemmMemoryParameters {
 };
 
 // Main utility functions to run GEMM
-template <typename ALayout, typename BLayout, typename DT> struct GemmRunner {
+template <typename ALayout, typename BLayout, typename DT>
+struct GemmRunner {
   using D = GemmDeviceOp<ALayout, BLayout, DT>;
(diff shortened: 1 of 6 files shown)


krzysz00 (approved):
Approved

sjw36 (approved):

mlir/utils/performance/ck-benchmark-driver/ck-benchmark-driver.cpp
```diff
@@ -12,6 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+#ifdef CK_FOUND
```
manupak, line 15: Im bumping into :
`clang-tidy found error: mlir/utils/performance/ck-benchmark-driver/ck-benchmark-driver.cpp:19:10: error: 'ck/ck.hpp' file not found [clang-diagnostic-error]`
http://ml-ci-internal.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir/detail/PR-1404/8/pipeline/186

@giuseros I had to do this to avoid clang-tidy erroring out when ck headers are not found.
I am not quite sure how this ever worked coz looking at the jenkins pipeline we run get and build ck way later -- so the header is not going to be there. Maybe Im missing something?

PR #1405, opened 2024-02-09
Check blocksize during tuning

yiqian1:
Copy the blockSize check from the gridwise-to-blockwise pass into the rock-tuning-parameter phase to rule out invalid tuning parameters at an early stage.

### DIFF
--- a/mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
+++ b/mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
@@ -672,6 +672,17 @@ LogicalResult PopulateParamsXDL::isValidBlockwiseGemm(
     return failure();
   }
 
+  // Reject invalid blockSize
+  int64_t kPerBlock = param.gemmKPerBlock;
+  int64_t mPerBlock = param.gemmMPerBlock;
+  int64_t nPerBlock = param.gemmNPerBlock;
+  int64_t aCopyPerThread = (kPerBlock * mPerBlock) / blockSize;
+  int64_t bCopyPerThread = (kPerBlock * nPerBlock) / blockSize;
+  if (aCopyPerThread == 0 || bCopyPerThread == 0) {
+    LLVM_DEBUG(llvm::dbgs() << "tuning: Block size too large.\n");
+    return failure();
+  }
+
   // Sledgehammer hotfix because not unrolling sometimes makes the register
   // allocator break. This should be refined quickly.
   if (param.gemmAThreadCopyMoreGemmK == false) {
--- a/mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
+++ b/mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
@@ -676,9 +676,7 @@ LogicalResult PopulateParamsXDL::isValidBlockwiseGemm(
   int64_t kPerBlock = param.gemmKPerBlock;
   int64_t mPerBlock = param.gemmMPerBlock;
   int64_t nPerBlock = param.gemmNPerBlock;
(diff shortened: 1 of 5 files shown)


giuseros (commented):
Are you sure this is ok? Because  @ravil-mobile is trying to move the block-size computation down in `GemmToGridwise`  while you clearly need to compute it here to run this check.

manupak:
I think its fine... because that is there to reject perf config upfront.

I dont think this interferes with block-size computation movement if anything it just shows the coupling between blocksize and tuning param -- thus allowing it to reject it early.

manupak:
to clarify the requirement of this PR (and the ticket it closes) : is that to avoid our clients running a compilation that is doomed to fail when we can figure that out by isValid = f(op, perfConfig).

For e.g. we can 'guess' the blocksize (wavesize * mWaves * nWaves)  if we know the op and perfconfig when blocksize is not there. Yea that coupling is not nice but we d need a larger refactor move tuning param space definintion also down along with blocksize computation (See https://github.com/ROCm/rocMLIR-internal/issues/954)

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -672,6 +672,17 @@ LogicalResult PopulateParamsXDL::isValidBlockwiseGemm(
...
+  // Reject invalid blockSize
+  int64_t kPerBlock = param.gemmKPerBlock;
+  int64_t mPerBlock = param.gemmMPerBlock;
+  int64_t nPerBlock = param.gemmNPerBlock;
+  int64_t aCopyPerThread = (kPerBlock * mPerBlock) / blockSize;
+  int64_t bCopyPerThread = (kPerBlock * nPerBlock) / blockSize;
+  if (aCopyPerThread == 0 || bCopyPerThread == 0) {
+    LLVM_DEBUG(llvm::dbgs() << "tuning: Block size too large.\n");
+    return failure();
+  }
```
manupak, line 682: This needs to be in sync : https://github.com/ROCm/rocMLIR/blob/225f6b78ab260de8523a92f1c7a8b31426d2bc9d/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp#L433C3-L437C6

Therefore, how about refactor this into a function in a common util ?

krzysz00:
Also, @giuseros , block size, unlike grid size (which can be moved to a different pass, I'd argue) is very much a (sometimes derived) tuning parameter, so dealing with it in affix-tuning-parameters makes sense.

mlir/test/CAPI/mixr_full.c
```diff
@@ -229,7 +229,7 @@ static bool constructAndTraverseIr(MlirContext ctx) {
     mlirRockTuningSetParam(tuningClone, tuningParam);
     if (mlirLogicalResultIsFailure(
             mlirPassManagerRunOnOp(applicabilityPm, tuningCloneOp))) {
-      // CHECK: Perfconfig "4,64,4,4,64,1,1,1" is not applicable
+      // CHECK-NOT: Perfconfig "4,64,4,4,64,1,1,1" is not applicable
```
manupak: I think we can remove this line rather than having a CHECK-NOT; i suppose you are not getting any failures now ?

yiqian1: Right, no failures now. I think this CHECK-NOT would ensure an early rejection?

krzysz00: No, CHECK-NOT is for ensuring text doesn't appear.
I'd update this to `CHECK` for the first "not applicable" perf config in the tuning set.
But if there is none, you'll want to `CHECK-NOT: is not applicable`

manupak: `CHECK-NOT: is not applicable` is better.

manupak (approved):
LGTM!

PR #1406, opened 2024-02-09
Transform `extract(select(a,b))` into `select(extract(a),extract(b)))`

giuseros:
This is transforming the sequence:
```
a = select(struct0, struct1)
ptr = extractValue(a)
```
into
```
ptr0 = extractValue(struct0)
ptr1 = extractValue(struct1)
ptr = select(ptr0, ptr1)
```
In this way we can smoothly remove `memref`s before going into the translation. 

Before the PR:
```
$ mlir-to-asm.sh config | grep mask
v_cndmask_b32_e64 v64, 0, 1, s[2:3]
v_cndmask_b32_e64 v67, 0, 1, s[0:1]
.amdhsa_reserve_xnack_mask 1
```
After the PR:
```
$ mlir-to-asm.sh config | grep mask
.amdhsa_reserve_xnack_mask 1
```

Also, this PR will make SROA executing (but still failing). While before it was simply stopping because of the `insertvalue` left in the LLVM-IR.

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/RockPrepareLLVM.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/RockPrepareLLVM.cpp
@@ -24,6 +24,7 @@
 #include "mlir/Dialect/Rock/Passes.h"
 #include "mlir/IR/TypeUtilities.h"
 #include "mlir/Pass/PassManager.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 #include "mlir/Transforms/Passes.h"
 #include "llvm/ADT/SmallBitVector.h"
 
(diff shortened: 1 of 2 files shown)


mlir/lib/Dialect/Rock/Transforms/RockPrepareLLVM.cpp
```diff
@@ -39,6 +40,37 @@ namespace rock {
...
+//  %ptrA = llvm.extractvalue %structA
+//  %ptrB = llvm.extractvalue %structB
+//  %s = llvm.select %ptrA, %ptrB
+struct SelectExtractRewritePattern
+    : public OpRewritePattern<LLVM::ExtractValueOp> {
+  using OpRewritePattern<LLVM::ExtractValueOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(LLVM::ExtractValueOp op,
+                                PatternRewriter &rw) const override {
+    if (auto selectOp = dyn_cast_or_null<LLVM::SelectOp>(
```
krzysz00: Nit: invert the condition? If it's not a `selectOp`, return failure early?

krzysz00 (approved):
I don't know if this is profitable in general, but given our codebase, it makes sense as an unconditional rewrite.

Approved!

PR #1409, opened 2024-02-12
Add benchmark support for fp8

giuseros:
This is my stab at adding benchmark support for `fp8`. It comes with a bit of refactoring to make the drivers/tuners type deduction a bit easier. What I did:
- I added the `{ck,rocblas}-benchmark-driver` incantations to call the right routines
- I copied the `fp32_fp8` conversion routines from `/opt/rocm`. Not sure if that is the most elegant way
 - I added `perfRunner/tuningRunner` support for `fp8`
 - I introduced a new `getCType` in the `GemmWrapperInterface` that lets us remove the weird `isOut` flag that we were using to get `i8(input)->i32(output)`  working.


krzysz00 (approved):
Approved, with the side note that now there's ambiguity between perf runner and rocmlir-gen about whether `fp8_fp8` means "fp8 * fp8 -> f32 and I'm being precise" (rocmlir-gen) or "fp8 * fp8 -> fp8" (perf and tuning configs).

(This is because you can do `-t fp8_bf8` for example, to get `fp8 * bf8 -> f32` out of rocmlir-gen)

giuseros:
Yeah, I noticed the ambiguity, but on the perf-runner `type_type` always meant `input_type_output_type`. I agree that we will have to align those things at some point

PR #1410, opened 2024-02-12
Fix collapseContiguousMerges() being too  greedy with unit dimensions

### LINKED ISSUE #1407: Convolution hang/crash
This is trace from the crash that includes the mlir program and tuning problem and solution keys:

```
Problem: gfx90a:sramecc+:xnack- 110     conv -F 1 -f GNCHW -I HWNGC -O NGCHW -n 1 -c 512 -H 256 -W 256 -k 512 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -g 1
Benchmarking solution: 128,256,4,64,128,4,1,1
module {
  func.func @mlir_transpose_convolution_add(%arg0: !migraphx.shaped<1x512x256x256xf32, 0x1x0x0>, %arg1: !migraphx.shaped<256x256x1x512xf32, 131072x512x512x1>, %arg2: !migraphx.shaped<512x512x3x3xf32, 4608x9x3x1>) -> !migraphx.shaped<1x512x256x256xf32, 33554432x65536x256x1> attributes {arch = "", kernel = "mixr", num_cu = 0 : i64} {
    %0 = migraphx.transpose %arg1 {permutation = [2, 3, 0, 1]} : <256x256x1x512xf32, 131072x512x512x1> -> <1x512x256x256xf32, 512x1x131072x512>
    %1 = migraphx.convolution %0, %arg2 {dilation = [1, 1], group = 1 : i64, padding = [1, 1, 1, 1], padding_mode = 0 : i64, stride = [1, 1]} : <1x512x256x256xf32, 512x1x131072x512>, <512x512x3x3xf32, 4608x9x3x1> -> <1x512x256x256xf32, 33554432x65536x256x1>
    %2 = migraphx.add %1, %arg0 : <1x512x256x256xf32, 33554432x65536x256x1>, <1x512x256x256xf32, 0x1x0x0> -> <1x512x256x256xf32, 33554432x65536x256x1>
    return %2 : !migraphx.shaped<1x512x256x256xf32, 33554432x65536x256x1>
  }
}
```

### DoD
 - Need to be backported to ROCm 6.1

krzysz00:
Fixes https://github.com/ROCm/rocMLIR/issues/1407

The bug resolved here is one that comes from an overly aggressive approach to adding length-1 dimensions to contiguous groups when detecting Merge/Unmerge pairs, which led to unusual permutation sequences like the one abstracted into the unit test for this commit causing an overeager dimension collapse. This led to runtime page faults because the H*W index into the output was being scaled by the large multiplier for the batch (N) dimension of the result tensor, leading to out of bounds accesses.

We guard against such issues in the future by strictly limiting the unit dimension grabber to only include those unit dimensions that lie within the contiguous group to be merged (as in the original Merge{8, 1, 3}, Unmerge{8. 3} example).

In addition, we add rudimentary debug output for merge collapsing so that it's clear that it's occurring.

### DIFF
--- a/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
+++ b/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
@@ -321,19 +321,30 @@ void findCountiguousGroupsUnmerge(const ArrayRef<uint32_t> upperDims,
     if (groupCandidate.size() > 1 &&
         std::is_sorted(dimPosition.begin(), dimPosition.end())) {
 
-      uint32_t lowerDim = groupCandidate.back();
+      uint32_t fastestDim = groupCandidate.back();
+      size_t fastestDimPosInMerge = dimPosition.back();
+      size_t slowestDimPosInMerge = dimPosition.front();
       for (auto d : groupCandidate)
-        contiguousGroups[keyI.transformPair()].unionSets(lowerDim, d);
+        contiguousGroups[keyI.transformPair()].unionSets(fastestDim, d);
 
       // We also want to add the singleton dimensions of the merge the
       // groupCandidate belongs to. In this way we cover for situations like
-      // - Merge{8,1,3}, <AddDim at [1]>, <Unmerge{8,3}>
-      for (auto pair : llvm::zip(keyI.transform.getLowerDims(),
-                                 keyI.transform.getParams())) {
-        int64_t d = std::get<0>(pair);
-        int64_t p = std::get<1>(pair);
+      // - Merge{8,1,3}, <AddDim at [1]>, <Unmerge{8,3}>. However, it is
+      // unsound to collapse onto trailing unit dimensions not otherwise
+      // accounted for above, as they could be permuted out to
+      // arbitrary other positions. Therefore, we only examine unit dimensions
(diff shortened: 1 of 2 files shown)


mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -322,16 +322,27 @@ void findCountiguousGroupsUnmerge(const ArrayRef<uint32_t> upperDims,
         std::is_sorted(dimPosition.begin(), dimPosition.end())) {
 
       uint32_t lowerDim = groupCandidate.back();
+      size_t lowerDimPosInMerge = dimPosition.back();
```
giuseros: Nit: why not `fastestDimPosInMerge` (since you use ` slowestDimPosInMerge = dimPosition.front()`)?

giuseros (approved):

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
krzysz00: Existing code naming, but I might as well clean that up while I'm here

PR #1412, opened 2024-02-14
Update repository URLs in Dockerfile and Jenkinsfile to reflect ROCm-wide changes

stefankoncarevic:


krzysz00 (approved):
Edit title to "Update repository URLs in Dockerfile for ROCm-wide changes" or something like that

PR #1413, opened 2024-02-15
[BACKPORT] Fix collapseContiguousMerges() being too  greedy with unit dimensions

krzysz00:
Fixes https://github.com/ROCm/rocMLIR/issues/1407

The bug resolved here is one that comes from an overly aggressive approach to adding length-1 dimensions to contiguous groups when detecting Merge/Unmerge pairs, which led to unusual permutation sequences like the one abstracted into the unit test for this commit causing an overeager dimension collapse. This led to runtime page faults because the H*W index into the output was being scaled by the large multiplier for the batch (N) dimension of the result tensor, leading to out of bounds accesses.

We guard against such issues in the future by strictly limiting the unit dimension grabber to only include those unit dimensions that lie within the contiguous group to be merged (as in the original Merge{8, 1, 3}, Unmerge{8. 3} example).

In addition, we add rudimentary debug output for merge collapsing so that it's clear that it's occurring.

### DIFF
--- a/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
+++ b/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
@@ -321,19 +321,30 @@ void findCountiguousGroupsUnmerge(const ArrayRef<uint32_t> upperDims,
     if (groupCandidate.size() > 1 &&
         std::is_sorted(dimPosition.begin(), dimPosition.end())) {
 
-      uint32_t lowerDim = groupCandidate.back();
+      uint32_t fastestDim = groupCandidate.back();
+      size_t fastestDimPosInMerge = dimPosition.back();
+      size_t slowestDimPosInMerge = dimPosition.front();
       for (auto d : groupCandidate)
-        contiguousGroups[keyI.transformPair()].unionSets(lowerDim, d);
+        contiguousGroups[keyI.transformPair()].unionSets(fastestDim, d);
 
       // We also want to add the singleton dimensions of the merge the
       // groupCandidate belongs to. In this way we cover for situations like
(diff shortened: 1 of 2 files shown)


mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -321,19 +321,30 @@ void findCountiguousGroupsUnmerge(const ArrayRef<uint32_t> upperDims,
...
+      // unsound to collapse onto trailing unit dimensions not otherwise
+      // accounted for above, as they could be permuted out to
+      // arbitrary other positions. Therefore, we only examine unit dimensions
+      // which come earlier in the merge parameters than the first true
+      // collapsed dimension.
+      ArrayRef<uint32_t> thisMergeDims = keyI.transform.getLowerDims();
+      ArrayRef<int64_t> thisMergeParams = keyI.transform.getParams();
+      for (const auto [d, p] : llvm::zip(
+               thisMergeDims.slice(slowestDimPosInMerge,
+                                   fastestDimPosInMerge - slowestDimPosInMerge),
```
sjw36, line 342: Nit: for readability:
> int distance = fastestDimPosInMerge - slowestDimPosInMerge;
> thisMergDims.slice(slowestDimPosInMerge, distance);

sjw36 (approved):
Looks good.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
krzysz00: Think I should address that here? This is a backport PR, after all

manupak (approved):
Approving backport!

PR #1416, opened 2024-02-16
Update device library finding to prepend ROCM_PATH, use hipconfig

### LINKED ISSUE #1415: Build failure with ubuntu 22.04
There is this failure when migraphx builds with ubuntu 22.04:

```
Traceback (most recent call last):
  File "/usr/local/cget/build/tmp-25313f4698354e8987318405eb382ca9/rocMLIR-bf2911e5245d6452d194ee5dc41c9c2b2869e5c0/external/llvm-project/mlir/lib/Dialect/GPU/AmdDeviceLibsIncGen.py", line 53, in <module>
    generate(Path(sys.argv[1]), Path(sys.argv[2]), sys.argv[3:])
  File "/usr/local/cget/build/tmp-25313f4698354e8987318405eb382ca9/rocMLIR-bf2911e5245d6452d194ee5dc41c9c2b2869e5c0/external/llvm-project/mlir/lib/Dialect/GPU/AmdDeviceLibsIncGen.py", line 29, in generate
    with (bcPath / (lib + ".bc")).open("rb") as libFile:
  File "/usr/lib/python3.10/pathlib.py", line 1119, in open
    return self._accessor.open(self, mode, buffering, encoding, errors,
FileNotFoundError: [Errno 2] No such file or directory: '/usr/amdgcn/bitcode/ocml.bc'
```

krzysz00:
Fixes #1415

Because our static library build wants to avoid runtime dependencies on (the location of) ROCm, but we need the device libraries for math functions, we pull the device libraries into a #include-able file at configure time. A recent bug report showed the build system picking up the wrong value for ROCM_PATH, leading to the device libraries not being findable. This commit, hopefully, addresses that issue.

### DIFF
--- a/external/llvm-project/mlir/lib/Dialect/GPU/CMakeLists.txt
+++ b/external/llvm-project/mlir/lib/Dialect/GPU/CMakeLists.txt
@@ -167,6 +167,7 @@ if(MLIR_ENABLE_ROCM_CONVERSIONS)
   find_package(AMDDeviceLibs CONFIG)
   set(CMAKE_PREFIX_PATH "${REAL_CMAKE_PREFIX_PATH}")
   if(AMDDeviceLibs_FOUND)
+    message(STATUS "AMD device libs prefix: ${AMD_DEVICE_LIBS_PREFIX}")
     set(device_lib_targets ${AMD_DEVICE_LIBS_TARGETS})
     # Filter out control constants, since we set those inline
     list(FILTER device_lib_targets EXCLUDE REGEX "^oclc_")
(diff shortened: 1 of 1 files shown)


pfultz2:
Thanks, can this be backported to the release branch?

krzysz00:
@pfultz2 Does it work? IF so, yeah, I'd be happy to backport this

pfultz2:
> Does it work? 

No, that still doesnt fix the issue.

krzysz00:
... Ok, so now I'm thinking there's a system-level install of the AMD device libraries somewhere that's polluting CMake.

I've added a print statement to try and get a better sense of how we're getting into that error state.

krzysz00:
... Well, it turns out that this doesn't actually fix the underlying problem, but it might or might not be worth landing anyway?

krzysz00:
Closing - to the extent this is useful, it's hopelessly merge-conflicted and will be resubmitted otherwise

PR #1418, opened 2024-02-19
Implemeted Split-K GEMM with coord. trans.

ravil-mobile:
Hi,

Here is my implementation of `split-k` with coordinate transformations. 

I talked to @manupak and he mentioned that it would be better to avoid `embed` because it is an `irreversible` transformation.
Currently, I cannot come up to an implementation of `split-k` without `embed`.

Closes ROCm/rocMLIR-internal#1380

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
@@ -20,6 +20,7 @@
 //
 //===-----------------------------------------------------===//
 #include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/MHAL/IR/MHAL.h"
 #include "mlir/Dialect/Rock/IR/GemmSize.h"
 #include "mlir/Dialect/Rock/IR/Rock.h"
 #include "mlir/Dialect/Rock/IR/TransformMapBuilder.h"
@@ -258,10 +259,26 @@ std::tuple<Value, Value, Value>
 GemmRewritePattern::arrangeSplitKTransform(OpBuilder &builder, GemmOp op,
                                            Location loc, int64_t splitKFactor,
                                            Value a, Value b, Value c) const {
+  // adjust the store method
   auto storeMethod =
       builder.getAttr<rock::StoreMethodAttr>(rock::StoreMethod::AtomicAdd);
   op.setStoreMethodAttr(storeMethod);
 
+  // set the prefill attribute
+  auto func = llvm::cast<func::FuncOp>(op->getParentOp());
+  auto attrName = mhal::PrefillAttr::getMnemonic();
+  auto elementType = c.getType().cast<MemRefType>().getElementType();
+  Attribute zero;
+  if (llvm::isa<FloatType>(elementType)) {
+    zero = builder.getFloatAttr(elementType, 0.0);
+  } else {
+    assert(llvm::isa<IntegerType>(elementType) &&
+           "expecting `int` element type");
+    zero = builder.getIntegerAttr(elementType, 0);
+  }
+  func.setArgAttrs(2, builder.getNamedAttr(attrName, zero));
+
+  // perform coordinate transformations
   Value aNew{nullptr}, bNew{nullptr}, cNew{nullptr};
   ArrayRef<int64_t> aShape = a.getType().cast<MemRefType>().getShape();
   ArrayRef<int64_t> bShape = b.getType().cast<MemRefType>().getShape();
--- a/mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
@@ -174,18 +174,19 @@ GemmRewritePattern::matchAndRewrite(GemmOp op, GemmOpAdaptor adaptor,
   // Note, matrix dimension correctness is handled in the verifier
   GemmSize size(/*g=*/aShape[0], /*m=*/aShape[2], /*k=*/aShape[1],
                 /*n=*/bShape[2]);
-  GemmSize extraPad =
-      requiredPadding(params, size).value_or(GemmSize{0, 0, 0, 0});
 
+  GemmSize scale{1, 1, 1, 1};
   int64_t splitKFactor{1};
   if (auto attr = op->getAttr("split-k-factor")) {
     splitKFactor = attr.cast<IntegerAttr>().getInt();
+    scale.k = splitKFactor;
   }
 
-  a = padMatrix(a, rw, loc, "gemmK", extraPad.k * splitKFactor, "gemmM",
-                extraPad.m);
-  b = padMatrix(b, rw, loc, "gemmK", extraPad.k * splitKFactor, "gemmN",
-                extraPad.n);
+  GemmSize extraPad =
+      requiredPadding(params, size, scale).value_or(GemmSize{0, 0, 0, 0});
+
+  a = padMatrix(a, rw, loc, "gemmK", extraPad.k, "gemmM", extraPad.m);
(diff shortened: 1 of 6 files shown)


manupak:
Here is an example : 

https://github.com/ROCm/rocMLIR/blob/ebfe8e2259117eb624e6000f9ad6892de48819b4/mlir/lib/Dialect/Rock/utility/AccelEmitter.cpp#L492
...
https://github.com/ROCm/rocMLIR/blob/ebfe8e2259117eb624e6000f9ad6892de48819b4/mlir/lib/Dialect/Rock/utility/AccelEmitter.cpp#L531-L532
...
https://github.com/ROCm/rocMLIR/blob/ebfe8e2259117eb624e6000f9ad6892de48819b4/mlir/lib/Dialect/Rock/utility/AccelEmitter.cpp#L547

The trick is it needs to be done across > 1 transform.

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
```diff
@@ -235,6 +251,118 @@ GemmRewritePattern::matchAndRewrite(GemmOp op, GemmOpAdaptor adaptor,
...
+    SmallVector<StringRef> inputDimNames{"gemmG", "gemmM", "gemmN"};
+    BottomUpTMBuilder embedTransform(builder, inputDimNames, cShape, loc);
+
+    llvm::StringMap<uint32_t> embedDims = expandNamesInPlace(
+        inputDimNames, {{"gemmG", {"gemmGTilda", "gemmKSplit"}}});
+
+    const int64_t G = cShape[0];
+
+    BottomUpTMTopDimsWrapper embedWrap(embedTransform, std::move(embedDims));
+    embedWrap.embed({"gemmGTilda", "gemmKSplit"}, {G, splitKFactor}, "gemmG",
```
manupak: here this could be a unmerge that does : gemmG --> gemmGTilda x gemmKSplit

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
```diff
@@ -235,6 +251,118 @@ GemmRewritePattern::matchAndRewrite(GemmOp op, GemmOpAdaptor adaptor,
...
+    auto mergeTransform =
+        BottomUpTMBuilder::above(embedTransform, embedTransformAttr);
+
+    llvm::StringMap<uint32_t> mergeDimOutputNames =
+        expandNamesInPlace(inputDimNames, {});
+
+    BottomUpTMTopDimsWrapper mergeTransformWrapper(
+        mergeTransform, std::move(mergeDimOutputNames));
+
+    mergeTransformWrapper.merge("gemmG", {"gemmGTilda", "gemmKSplit"});
```
manupak: This can just be a passthrough : "gemmGTilda" --> "gemmG"
alongside : ignore("gemmKSplit").

manupak (commented):
here was my suggestion

ravil-mobile:
Thanks @manupak and @krzysz00. The power of the `TopDownBuilder` allowed me to eliminate the `embed` transform

manupak:
Would you be removing this https://github.com/ROCm/rocMLIR/blob/7861b5abe47990fd89862d2cc67b8bd0469faea2/mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp#L586C44-L586C50
here ?

manupak:
> Thanks @manupak and @krzysz00. The power of the `TopDownBuilder` allowed me to eliminate the `embed` transform

well... BottomUpTMBuilder::addDim can do the same.

ravil-mobile:
> Would you be removing this https://github.com/ROCm/rocMLIR/blob/7861b5abe47990fd89862d2cc67b8bd0469faea2/mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp#L586C44-L586C50 here ?


It is a very good question. Let's discuss whether it is better to remove it as a part of this PR or it is better to do it in a separate PR.

ravil-mobile:
@manupak , @krzysz00 @jerryyin @sjw36 

I just obtained some first performance results which I would like to share with you. My tested GEMM problem was `-g 1 -m 256 -k 4096 -n 256`. I used MI250 GPU which has 110 CUs.

I used the following performance config `--perf_config="64,64,4,32,32,4,1,1"`

Regarding the default (data parallel) implementation, there are 4x4 (16 blocks) tiles and each tile requires 256-k iterations. According to `rocprof`, execution took ~173us.

I compared it with the `split-k` variant with the split factor equal to `5`. It leads to 4x4x5 = 80 blocks and results in ~56us according to rocprof.

sjw36:
That's great @ravil-mobile .
Interested to see how `split-k` == `12` would do. Would be good to test multiple workgroups per CU.

manupak:
@ravil-mobile would you be able to measure a memset or rock.init_kernel overhead for that output ?
so we can kind of a get an idea about net benefit

ravil-mobile:
> @ravil-mobile would you be able to measure a memset or rock.init_kernel overhead for that output ? so we can kind of a get an idea about net benefit

Hi @manupak,

Here is the trace for the case when `splitk == 12`. Our implementation uses `hipMemset`

image

`FillBuffer` takes `3us`. 

I can see that we spend most of the time on `hipModuleLoadData`, which corresponds to the setting context for our generated kernel.

manupak:
Is the fill happening on CPU ? and copied over to device ?

ravil-mobile:
> Is the fill happening on CPU ? and copied over to device ?


No, it is call to `hipMemset`. It is a device kernel.

https://github.com/ROCm/rocMLIR/blob/da3df73b13c69d6b1d5ffec9c8281c261a0df257/external/mlir-hal/lib/Dialect/MHAL/Transforms/Prefill.cpp#L81-L82

https://github.com/ROCm/rocMLIR/blob/da3df73b13c69d6b1d5ffec9c8281c261a0df257/external/llvm-project/mlir/lib/ExecutionEngine/RocmRuntimeWrappers.cpp#L157-L171

sjw36:
> Here is the trace for the case when `splitk == 12`. 

What was the kernel runtime?

manupak:
Why is the fillBuffer not on the GPU ? IIRC, I saw fills in the same row as the gemm kernel.

ravil-mobile:
>That's great @ravil-mobile .
>Interested to see how `split-k` == `12` would do. Would be good to test multiple workgroups per CU.


Thanks @sjw36. Yes,  `split-k` == `12` is even better. It results in ~38us.

I did some sketch of the work balancing heuristics

image

What do you think? The one thing which is missing is the upper limit for the `split-k` range. Moreover, it doesn't consider increasing overheads of `attomicAdd` operations.

I decided to demonstrate the first performance results now because I was curious about performance improvements. A more elaborate analysis should be done in a separate PR.

ravil-mobile:
> Is the fill happening on CPU ? and copied over to device ?

Hi @manupak. I can see your point now

giuseros:
Chiming in late, but I think the code is in a good shape. I would be tempted to remove the split-k hack in conv2gemm, and simply setting the split-k-factor to bootstrap your transformation

giuseros:
Also, additional question about tuning. In your example, did you tune the split-k gemm?

ravil-mobile:
> Chiming in late, but I think the code is in a good shape. I would be tempted to remove the split-k hack in conv2gemm, and simply setting the split-k-factor to bootstrap your transformation

Hi @giuseros, I created a separate ticket for it. It is here: https://github.com/ROCm/rocMLIR-internal/issues/1388

ravil-mobile:
> Also, additional question about tuning. In your example, did you tune the split-k gemm?

@giuseros. No, not yet. It also has a dedicated ticket which is here: https://github.com/ROCm/rocMLIR-internal/issues/1389

ravil-mobile:
re-opening PR

PR #1419, opened 2024-02-19
Fix incorrect indexing inte Merge parameters in -rock-fold-broadcast

### LINKED ISSUE #1417: [Issue]: Assertion `Index < Length && "Invalid index!"' failed
### Problem Description

Build MIGraphX `develop` branch + https://github.com/ROCm/AMDMIGraphX/pull/2782 
I was using a ROCm 6.0.2 docker
I ran the migraphx-driver and hit the error listed in the "Additional Information" section of this issue.

### Operating System

20.04.6

### CPU

AMD EPYC 7702 64-Core Processor

### GPU

AMD Instinct MI100

### ROCm Version

ROCm 6.0.0

### ROCm Component

rocMLIR

### Steps to Reproduce

_No response_

### (Optional for Linux users) Output of /opt/rocm/bin/rocminfo --support

_No response_

### Additional Information

Trace collected from MIGRAPHX_TRACE_MLIR=1 /workspace/AMDMIGraphX/build/bin/driver perf /new-saved-models/tiny_llama/decoder_model.onnx --fill1 input_ids attention_mask --inpu
t-dim @input_ids 1 256 @attention_mask 1 256

[mlir_error_tinylama.txt](https://github.com/ROCm/rocMLIR/files/14316147/mlir_error_tinylama.txt)

krzysz00:
Fixes #1417

FoldBroadcast would crash when processing Merge operations like <Merge{4, 1} ["x"] at [1] -> ["a", "b"] at [1, 2]>, as it would index the merge parameters with 1 and 2 and not 0 and 1, causing out of bounds array accesses and thus assertion failures.

This commit resolves that issue.

In addition, while looking around nearby, I identifidied that the handling for Broadcast{} operations that handled multiple dimensions (like, say,
<Broadcast{1, 1} ["a", "b"] at [0, 1] -> ["a", "b"] at [0, 1]> ) was incorrect because the worklist update logic always checked the last underlying dimension for Broadcast{}. This commit fixes that issue.

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/FoldBroadcast.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/FoldBroadcast.cpp
@@ -74,8 +74,8 @@ struct FoldBroadcast : public OpRewritePattern<rock::GemmOp> {
             // anything, otherwise we need to be sure that the new
             // dimension is a single-length broadcast
             case rock::TransformType::Broadcast:
-              if (tr.getParams().back() != 1)
-                newWorkList.insert(tr.getLowerDims().back());
+              if (tr.getParams()[idx] != 1)
+                newWorkList.insert(tr.getLowerDims()[idx]);
               break;
             // AddDim and ConstDim are basically broadcasts. No
             // need to go further
@@ -92,8 +92,9 @@ struct FoldBroadcast : public OpRewritePattern<rock::GemmOp> {
             // we need to ensure that all their (lower) dimensions
             // bigger than 1 lead to broadcasts
             case rock::TransformType::Merge:
-              for (auto dim : tr.getLowerDims()) {
-                if (tr.getParams()[dim] != 1)
+              for (auto [length, dim] :
+                   llvm::zip(tr.getParams(), tr.getLowerDims())) {
+                if (length != 1)
                   newWorkList.insert(dim);
               }
               break;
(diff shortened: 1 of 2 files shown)


mlir/lib/Dialect/Rock/Transforms/FoldBroadcast.cpp
```diff
@@ -74,8 +74,8 @@ struct FoldBroadcast : public OpRewritePattern<rock::GemmOp> {
             // anything, otherwise we need to be sure that the new
             // dimension is a single-length broadcast
             case rock::TransformType::Broadcast:
-              if (tr.getParams().back() != 1)
-                newWorkList.insert(tr.getLowerDims().back());
+              if (tr.getParams()[idx] != 1)
+                newWorkList.insert(tr.getLowerDims()[idx]);
```
jerryyin, line 78: Ok orthogonal to this fix, I'm curious of how the original case condition works. Do we have any broadcast that falls into this if condition? My assumption is that `tr.getParams()[idx] == 1` should always satisfy because broadcast only supported from unit dimension.

krzysz00: I designed Broadcast to support the general case, and while I don't think we currently have any clients that call for that, it's - IMO - a very reasonable thing to support

jerryyin (approved):
Could you create a backport to release/rocm-rel-6.0?

PR #1420, opened 2024-02-20
Change vectorization and merge collapse APIs to take Value arguments

krzysz00:
This commit removes getMaxVectorization[forDataType]() as a function that takes a sequence of transform map attributes and replaces it with a version that takes a transformed Value. Similarly, collapseContiguousMerges() now takes the transformed Value and clones the IR containing said contiguous merges in place instead of taking a transform stack.

Furthermore, getMaxVectorization() now returns a VectorizationResult struct, which includes the maximum vectorization fot the underlying buffer along with information on how said underlying buffer is naturally vectorized (this is currently unused, but will be needed for rock.scalarize).

(While I was at it, I made the 128 bits / element type bitwidth thing the default behavior for vectorization, since we almost always use it.)

This change is nedeed:
1. To enable getMaxVectorization() to reason about - and through - the upcoming rock.scalarize operation, which isn't a coordinate transformation but sort of acts like one.
2. To adapt the API for future dynamic shape cases, where a TransformMapAttr might have symbols and so will need to, inherently, be part of a TransformOp.

As a side effect, the vectorization inferenece tests now have nice names and are split up into individual functions.


krzysz00:
Note: this isn't entirely ready to merge, since I `#if 0`'d the CollapseContiguousMerges unit tests out of the way temporarily, but I might as well get some review going.

krzysz00:
This now adds fusion traversal. It can be merged if people want it soon, but it's missing some of its test coverage.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -761,39 +778,119 @@ propagateVectorizationInfo(TransformMapAttr map, const VectorizationData &input,
...
+    while (auto trOp = currentVal.getDefiningOp<TransformOp>()) {
+      TransformMapAttr transformMap = trOp.getTransform();
+      LLVM_DEBUG(llvm::dbgs() << "Max vectorization data: ");
+      data.debugPrint();
+      LLVM_DEBUG(llvm::dbgs() << "Processing: " << transformMap << "\n");
+      data = propagateVectorizationInfo(transformMap, data, contiguousMerges);
+      currentVal = trOp.getInput();
+      currentUser = trOp.getOperation();
+    }
+    if (traverseFusions && currentVal.getDefiningOp<memref::AllocOp>()) {
```
manupak: [not a blocking comment] i think its much cleaner to write a wrapper function that returns currentVal and currentUser.

A function that would do :
So if its a TransformOp --> then return { trOp.getInput(), trOp.getOperation()}
If not && traverseFusions --> return findPostFusionTransforms(...)
(also can pass the fusionTraversalStatus get an update)

So you iterate until that wrapper function return success..

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -877,15 +969,25 @@ ArrayAttr mlir::rock::collapseContiguousMerges(ArrayAttr transforms,
...
     }
+    TransformMapAttr newMap = map;
     if (changed) {
-      auto newMap = TransformMapAttr::get(ops, map.getUpperBounds(),
-                                          map.getLowerBounds());
-      newTransformMaps.push_back(newMap);
+      newMap = TransformMapAttr::get(ops, map.getUpperBounds(),
+                                     map.getLowerBounds());
+      // From now on, we have to copy the transform stack.
+      needCloning = true;
```
manupak: [Clarification] Why do we started to need to clone *now* ?

manupak (commented):
Broadly looks fine.

I just had few stylistic comments.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
manupak: and get rid of this `while(true)`

krzysz00: The one tricky thing is that the transitions can sometimes involve affecting some other variables, but yeah, it could be a lambda.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
krzysz00: I'll update, but the point is we don't want to mutate the IR in place because there might be other users of the transform stack.

Though ... yeah, we could probably just replace one singular use ... yeah, now that I think about it, that's viable

krzysz00:
We've now migrated all the tests over

manupak (approved):
I dont have major concerns here...

mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
```diff
@@ -61,6 +61,14 @@ bool needs64BitIndices(TransformMapAttr map);
 /// Apply a chain of transforms on a memref and return the final view
 Value transform(OpBuilder &b, Value toBeTransformed, ArrayAttr transforms);
 
+/// Returns a version of `transformed` where all the `rock.transform`
+/// and `rock.scalarize` operations between `transformed` and the underlying
+/// memory have one user, cloning those intermediate operations if needed.
+/// This ensures that optimizations can edit those transformations in place
+/// without breaking any other uses that may have been merged together.
+/// If the transforms are already isolated, this function does nothing.
+Value isolateTransforms(OpBuilder &b, Value transformed);
```
manupak, line 70: Generally I feel this is a useful canonicalization for rock.transform ops because it is always going to be replicated at the ultimate user. WDYT ? 

No need to do anything here; just a comment.

manupak (approved):
Still looks good to me!

mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
krzysz00: On the other hand, transforms are `Pure`, should die when they have no users, etc., so this is temporarily breaking common subexpression elimination.

Though I'd at least consider whether `rock.transform` could take an array of transforms ...

PR #1423, opened 2024-02-20
Replacing ROCmSoftwarePlatform with ROCm

jerryyin:
Renaming performed by:

> find ./ -type f -exec sed -i -e 's/ROCmSoftwarePlatform/ROCm/g' {} \;
    find ./ -type f -exec sed -i -e 's/llvm-project-private/rocMLIR-internal/g' {} \;


manupak (approved):
Thanks for the cleanup! LGTM

PR #1424, opened 2024-02-20
[Attention] Fix cross-attention padding

manupak:
The padding application was wrong for
where gemm0 had : k0 x m * k0 x n0
and gemm1 had : n0 x m * n0 x n1

where m != n0 in cross-attention


sjw36 (approved):
Looks good.

PR #1411, opened 2024-02-13
[Attention] add support first-gemm only i8 kernels.

manupak:
This PR builds up on : https://github.com/ROCm/rocMLIR/pull/1399, hence that one needs merging first.

In terms of functionality additions : 
Given the functionality in https://github.com/ROCm/rocMLIR/pull/1399, This PR needed to handle tosa "implicit" broadcasting because that is what we usually end up getting for dequantization cases. Therefore, I have refactored the code in regularize of making a linalg generic to use identity index maps to a utility by pulling out non-idenity affine maps to their corresponding rock.transforms -- rest falls in place nicely. Note that, in tosa-to-rock, I have removed blocking of "implicit" broadcasting binary elementwise ops. 

Other non-test code is just making sure element type plumbing is sane in the lowering.

Testing code changes : 
* I have added rocmlir-gen to generate first-gemm only i8 rock.attention kernels and their corresponding host verification code.
* Added tests to cover i8 cases similiar to f16 cases.
* Added a combination of e2e test to cover : padding, quant scale and quant bias.

closes : https://github.com/ROCm/rocMLIR-internal/issues/953
closes : https://github.com/ROCm/rocMLIR-internal/issues/1349

PR #1428, opened 2024-02-21
[Attention] Generalize problem key to support cross attentions

manupak:
~~Builds on https://github.com/ROCm/rocMLIR/pull/1411. Thus that one needs merging first.~~

Currently,  we define the attention operation using primarily two parameters : sequence length and head dim (apart from number of heads which is a type of batching dimension)

However, this does not allow to define cross attention (See https://vaclavkosar.com/ml/cross-attention-in-transformer-architecture) which is found in stable diffusion models. Therefore,  this PR replace `seq_len` with `seq_len_q` +`seq_len_k` and `head_dim` with `head_dim_qk`, `head_dim_v`

closes : https://github.com/ROCm/rocMLIR-internal/issues/1386


mlir/test/fusion/nightly-misc-e2e/lit.local.cfg
```diff
@@ -1,2 +1 @@
-if not config.enable_rock_driver_e2e_test or config.no_AMD_GPU:
-  config.unsupported = True
+
```
manupak:
```suggestion
if not config.enable_rock_driver_e2e_test or config.no_AMD_GPU:
  config.unsupported = True
```

manupak:
I d appreciate a look here...

manupak:
This is merged as a side effect of merging : https://github.com/ROCm/rocMLIR/pull/1430

PR #1430, opened 2024-02-23
[Attention] Use head_dim_v as gemm1MPerBlock where possible

manupak:
Builds on https://github.com/ROCm/rocMLIR/pull/1428. Hence that one needs merging first.

This PR makes the gemm1 tuning param derivation always use head dimension of V as gemm1MPerBlock  so that second gemm will not be split across multiple workgroups.

Additionally, it adds a safe guard to not blow up if head_dim_v becomes too large resulting whole tuning space becoming invalid.

closes : https://github.com/ROCm/rocMLIR-internal/issues/1392


krzysz00 (approved):
This all feels reasonable to me

PR #1433, opened 2024-02-26
First step in N-dimensional convolution support.

pcf000:
First step in N-dimensional convolution support:  change the scalar width and height fields into elements of a vector of dimensions.


mlir/include/mlir/Dialect/Rock/Generator/ConvGenerator.h
```diff
@@ -1,17 +1,17 @@
-//===- Conv2dGenerator.h - MLIR to C++ option parsing ---------------===//
+//===- ConvGenerator.h - MLIR to C++ option parsing ---------------===//
```
pcf000, line 1: Lots of renaming "conv2d" to "conv."

mlir/include/mlir/Dialect/Rock/Generator/ConvGenerator.h
```diff
@@ -39,10 +41,10 @@ class Conv2dGenerator {
     std::string filterDataTypeStr;
     std::string inputDataTypeStr;
     std::string outputDataTypeStr;
-    int dilationHeight, dilationWidth;
-    int strideHeight, strideWidth;
-    int paddingHeightLeft, paddingHeightRight;
-    int paddingWidthLeft, paddingWidthRight;
+    SmallVector<int, 5> dilationDims;
```
pcf000: Arbitrary 5.  Krzysztof suggested 4 would be okay, as long as it's generalisable.  Should it be a symbolic constant?

mlir/include/mlir/Dialect/Rock/Generator/ConvGenerator.h
```diff
@@ -24,8 +24,10 @@ namespace mlir {
 namespace rock {
 struct ConvolutionDims;
 
-class Conv2dGenerator {
+class ConvGenerator {
 public:
+  enum DIM { HEIGHT = 0, WIDTH = 1 };
```
pcf000, line 29: Symbolic constants for dimensions within the vector.  Probably should have DEPTH, too, since it's common.

mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
```diff
@@ -135,8 +135,10 @@ void AffixTuningParameters::affixTuningParametersImpl(
         status = populateParamsAccelPtr->obtainTuningParameters(
             op, perfConfig, validParams, blockSize);
       }
-      if (failed(status))
+      if (failed(status)) {
         signalPassFailure();
+        return;
```
pcf000, line 140: The particular combination of cmake flags that I've been using has sent me down this path for a couple of tests.  I'm apparently doing combinations that CI doesn't usually do.

Most uses of signalPassFailure() return immediately, but these don't, leading to crashes later on.  With the returns, they just report errors, which I'll have to do something about, but at least it's cleaner.

mlir/utils/performance/common/benchmarkUtils.cpp
```diff
@@ -8,13 +8,13 @@
...
 //===----------------------------------------------------------------------===//
 
-#include "benchmarkUtils.h"
-
 #include <algorithm>
 #include <iostream>
 #include <string>
 #include <vector>
 
+#include "benchmarkUtils.h"
```
pcf000: benchmarkUtils.h was using something from one of those four system include files, and compilation complained.

mlir/include/mlir/Dialect/Rock/Generator/ConvGenerator.h
krzysz00: 1) While we're here, make it `int64_t`
2) I'd argue for 4 here because it's better aligned

mlir/include/mlir/Dialect/Rock/Generator/ConvGenerator.h
```diff
@@ -55,29 +57,29 @@ class Conv2dGenerator {
     SmallVector<int64_t, 5> inputDimension;
     SmallVector<int64_t, 5> outputDimension;
 
-    int filterHeight;
-    int filterWidth;
+    SmallVector<int, 5> filterDims;
```
krzysz00: This, however, is a good candidate for 5 or 6

mlir/utils/performance/common/benchmarkUtils.cpp
krzysz00: You'll want to rebase in a change I made - I added a `#include <string>` to `benchmarkUtils.h` and then you won't need this.

krzysz00 (approved):
Overall, looks like a good first step here

mlir/include/mlir/Dialect/Rock/Generator/ConvGenerator.h
pcf000: Making them int64_t had a lot of ripple effects.  See the last commit.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -70,9 +70,9 @@ class Rock_Conv2DOpBase<string mnemonic, list<Type> inputTypes=[F32, F16, BF16],
                    Rock_GemmFeaturesAttr:$features,
                    OptionalAttr<I32Attr>:$derivedBlockSize,
                    OptionalAttr<I32Attr>:$gridSize,
-                   I32ArrayLength<4>:$padding,
-                   I32ArrayLength<2>:$strides,
-                   I32ArrayLength<2>:$dilations,
+                   I64ArrayLength<4>:$padding,
```
krzysz00: Ok, I'm glad you flagged me down. Could you make these `index` please?

krzysz00: Or at least consider that - given that `index` is the local `ssize_t`, which seems like the right type for these attributes.

krzysz00 (commented):
Annoying nits the next.

PR #1436, opened 2024-02-29
[Attention] Fix G1MPerBlock Derivation

manupak:
Currently, if the G0MPerBlock is larger than
the upper bound heurstic used to increase G1MPerBlock
this will result in creating zeros.

This commit fixes this by restricting the guard
to ensure we only ever to do an increase in G1MPerBlock
by doing the whole calculation before assignment

closes : https://github.com/ROCm/rocMLIR-internal/issues/1396


krzysz00 (approved):
Looks fine to me. I might like a more descriptive variable name than `gemm1MPerBlockNew` but I wouldn't call that blocking on this bugfix.

manupak:
Noted. I ll slide something in a next PR or so.

PR #1439, opened 2024-03-01
Use `waves-per-eu` properly

manupak:
Currently waves-per-eu is calculated simply by blocksize / wavesize >= 2 heuristic.

I think this is not correct.
According to https://clang.llvm.org/docs/AttributeReference.html#amdgpu-waves-per-eu, we should only be doing this if there are sufficient parallelism within workgroup and/or grid constrained by the lds usage of the kernel. Otherwise, we will be losing scalar performance for non-existent occupancy.

I ll need to a perf eval shortly to see if the theory works.

This commit also removes bunch of side channelled info from higher up in the stack that were there to calculate lds usage where as in it could be properly calculated at `rock-to-gpu` before they are exported as symbols.

closes : https://github.com/ROCm/rocMLIR-internal/issues/1393


mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
```diff
@@ -616,3 +617,23 @@ mlir::rock::transposeSubTileViews(PatternRewriter &rewriter, Location loc,
...
+  }
+  StringAttr arch = op->getAttrOfType<StringAttr>("arch");
+  if (!arch)
+    arch = func->getAttrOfType<StringAttr>("mhal.arch");
+  if (!arch) {
+    auto mod = func->getParentOfType<ModuleOp>();
+    arch = mod->getAttrOfType<StringAttr>("mhal.arch");
+  }
+  return arch;
+}
```
giuseros: I think this line should be removed

giuseros (commented):
Could you run perf tests with the quick tuning space and see if there is any hit?

manupak:
> Could you run perf tests with the quick tuning space and see if there is any hit?

I just finished a run with last weekly tuning db and it seems flat.
[out.zip](https://github.com/ROCm/rocMLIR/files/14463571/out.zip)

that is old (best tuning for old) vs new (best tuning for old). So it could only equal or better if I use it with best tuning for new.

giuseros (approved):
LGTM!

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
```diff
@@ -616,3 +617,23 @@ mlir::rock::transposeSubTileViews(PatternRewriter &rewriter, Location loc,
...
+    func = op->getParentOfType<func::FuncOp>();
+    if (!func) {
+      func = op->getParentOfType<gpu::GPUFuncOp>();
+    }
+  }
+  StringAttr arch = op->getAttrOfType<StringAttr>("arch");
+  if (!arch)
+    arch = func->getAttrOfType<StringAttr>("mhal.arch");
+  if (!arch) {
+    auto mod = func->getParentOfType<ModuleOp>();
```
sjw36, line 640: Could also be a GPUModule.

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
```diff
@@ -616,3 +617,23 @@ mlir::rock::transposeSubTileViews(PatternRewriter &rewriter, Location loc,
                                 std::nullopt};
   }
 }
+
+StringAttr mlir::rock::getArch(Operation *op) {
```
sjw36: Nice utility.

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
```diff
@@ -330,6 +323,76 @@ void LowerRockOpsToGPUPass::runOnOperation() {
...
+        isUsingLessThanHalfLds = archInfo.maxSharedMemPerWG / ldsUsage >= 2;
+      }
+      // There is no point of instructing backend compiler to potentially reduce
+      // scalar code performance (via limiting VGPR usage) if we are already
+      // limited by the parallelsim within the kernel and/or lds usage.
+      if ((hasMoreThan2WavesPerBlock || hasMoreThan2WGsPerCU) &&
+          isUsingLessThanHalfLds) {
+        LLVM_DEBUG(llvm::dbgs() << "waves_per_eu:2"
+                                << "\n");
+        gpuFunc->setAttr("rocdl.waves_per_eu", b.getI32IntegerAttr(2));
```
sjw36: In Triton, there are cases where we set waves_per_eu to 3 and even with minor spilling, it produces better perf. But it is a tuning parameter, since it is difficult to guess how many registers will be used by the backend.

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
```diff
@@ -330,6 +323,76 @@ void LowerRockOpsToGPUPass::runOnOperation() {
...
+      return;
+    }
+    int64_t gridSize =
+        gpuFunc->getAttrOfType<IntegerAttr>("grid_size").getInt();
+    StringAttr arch = rock::getArch(gpuFunc);
+    if (arch) {
+      rock::AmdArchInfo archInfo = rock::lookupArchInfo(arch);
+      int64_t wavesPerBlock =
+          blockSize / (archInfo.numEUPerCU * archInfo.waveSize);
+      int64_t wgsPerCU = gridSize / archInfo.minNumCU;
```
sjw36: Is this assuming all WGs resident?

I think we should also have a `rock::getNumCUs(op)` lookup method (like `rock::getArch(op)`).

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
```diff
@@ -330,6 +323,76 @@ void LowerRockOpsToGPUPass::runOnOperation() {
...
+      LLVM_DEBUG(llvm::dbgs() << "  gridSize:" << gridSize << "\n");
+      LLVM_DEBUG(llvm::dbgs() << "  numCUs:" << archInfo.minNumCU << "\n");
+      LLVM_DEBUG(llvm::dbgs()
+                 << "maxSharedMemPerWG:" << archInfo.maxSharedMemPerWG << "\n");
+      LLVM_DEBUG(llvm::dbgs() << "ldsUsage:" << ldsUsage << "\n");
+      bool hasMoreThan2WavesPerBlock = wavesPerBlock >= 2;
+      bool hasMoreThan2WGsPerCU = wgsPerCU >= 2;
+      bool isUsingLessThanHalfLds = true;
+      if (ldsUsage > 0) {
+        isUsingLessThanHalfLds = archInfo.maxSharedMemPerWG / ldsUsage >= 2;
```
sjw36: This also doesn't account for maxSharedMemPerCU which is different than maxSharedMemPerWG on Navi. A WG could use all of maxSharedMemPerWG and still have 2 WGs per CU.

sjw36 (changes requested):
The main question I have is, does this account for lifetime analysis, such as in the example of FA where some LDS buffers do not overlap temporally?

manupak:
> does this account for lifetime analysis, such as in the example of FA where some LDS buffers do not overlap temporally

AFAIK, There is no such thing in our compiler as we are already in bufferized domain. So if you allocate LDS it will stay there...

If LDS buffers do not overlap temporally, then in *our* compiler we have to manage and share the same buffer. There is no intelligent bufferization processs that overlaps those later. I think you are stemming off how triton does things in their SSA compilation passes.

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
manupak: done

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
manupak: I d argue thats future work to investigate this -- with or without this PR the max we set is 2.
please create a ticket.

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
manupak: > Is this assuming all WGs resident?

No if WGs are resident, then gridSize / archInfo.minNumCU would be < 1 (zero).

> I think we should also have a rock::getNumCUs(op) lookup method (like rock::getArch(op)).

I disagree. numCUs are not a property of the op rather it is a property of the arch.

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
manupak: I see. so I should really be using maxSharedMemPerCU here...

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
sjw36: Yes, I meant this to be an FYI. Here is a ticket https://github.com/ROCm/rocMLIR-internal/issues/1406

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
manupak: done

sjw36:
> > does this account for lifetime analysis, such as in the example of FA where some LDS buffers do not overlap temporally
> 
> AFAIK, There is no such thing in our compiler as we are already in bufferized domain. So if you allocate LDS it will stay there...
> 
> If LDS buffers do not overlap temporally, then in _our_ compiler we have to manage and share the same buffer. There is no intelligent bufferization processs that overlaps those later. I think you are stemming off how triton does things in their SSA compilation passes.

Yes, Triton does this in an SSA form that is easy to deduce. We currently do this in gridwise-to-blockwise when we know everything. Does the gpu::WorkgroupAttributions method get the right LDS size for FA then?

manupak:
> > > does this account for lifetime analysis, such as in the example of FA where some LDS buffers do not overlap temporally
> > 
> > 
> > AFAIK, There is no such thing in our compiler as we are already in bufferized domain. So if you allocate LDS it will stay there...
> > If LDS buffers do not overlap temporally, then in _our_ compiler we have to manage and share the same buffer. There is no intelligent bufferization processs that overlaps those later. I think you are stemming off how triton does things in their SSA compilation passes.
> 
> Yes, Triton does this in an SSA form that is easy to deduce. We currently do this in gridwise-to-blockwise when we know everything. Does the gpu::WorkgroupAttributions method get the right LDS size for FA then?

Yes, now it does. We decide on the LDS size when we allocate it basically which happens in  gridwise-to-blockwise for all kernels. Then it stays the same until they are exported as global symbols prior to LLVM. So one could check LDS usage any point after they are allocated. If some optimization are happening then it should rewrite the allocation itself. Therefore, before convert to GPU, the right thing to do is calculate the LDS usage by looking at allocations at lowest most feasible point in the compilation pipeline.

sjw36:
> > > > does this account for lifetime analysis, such as in the example of FA where some LDS buffers do not overlap temporally
> > > 
> > > 
> > > AFAIK, There is no such thing in our compiler as we are already in bufferized domain. So if you allocate LDS it will stay there...
> > > If LDS buffers do not overlap temporally, then in _our_ compiler we have to manage and share the same buffer. There is no intelligent bufferization processs that overlaps those later. I think you are stemming off how triton does things in their SSA compilation passes.
> > 
> > 
> > Yes, Triton does this in an SSA form that is easy to deduce. We currently do this in gridwise-to-blockwise when we know everything. Does the gpu::WorkgroupAttributions method get the right LDS size for FA then?
> 
> Yes, now it does. We decide on the LDS size when we allocate it basically which happens in gridwise-to-blockwise for all kernels. 
Ok, sounds good. But just to clarify: for FA do we allocate 1 buffer for K and V to share? That is the larger of the two?

manupak:
> Ok, sounds good. But just to clarify: for FA do we allocate 1 buffer for K and V to share? That is the larger of the two?

Yes its by definition.
https://github.com/ROCm/rocMLIR/blob/d0016701ffa4749b5318d74fd420e69fd260e709/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp#L1743-L1753

https://github.com/ROCm/rocMLIR/blob/d0016701ffa4749b5318d74fd420e69fd260e709/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp#L934-L974

manupak:
@sjw36 is there anything else ?
(Thanks for the review btw, I wanted to run this through you)

manupak:
@sjw36 a gentle ping... I d like this moving. Are there more to be done here ?

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
sjw36: The arch only specifies the minimum CUs, see numCU on pretty much all the rock.ops. MIGraphX will tell us the actual num_cus and it will be added as a property along with arch.

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
```diff
@@ -330,6 +323,76 @@ void LowerRockOpsToGPUPass::runOnOperation() {
...
+      LLVM_DEBUG(llvm::dbgs()
+                 << "  numEUPerCU:" << archInfo.numEUPerCU << "\n");
+      LLVM_DEBUG(llvm::dbgs() << "  waveSize:" << archInfo.waveSize << "\n");
+      LLVM_DEBUG(llvm::dbgs() << "wgsPerCU:" << wgsPerCU << "\n");
+      LLVM_DEBUG(llvm::dbgs() << "  gridSize:" << gridSize << "\n");
+      LLVM_DEBUG(llvm::dbgs() << "  numCUs:" << archInfo.minNumCU << "\n");
+      LLVM_DEBUG(llvm::dbgs()
+                 << "maxSharedMemPerWG:" << archInfo.maxSharedMemPerWG << "\n");
+      LLVM_DEBUG(llvm::dbgs() << "ldsUsage:" << ldsUsage << "\n");
+      bool hasMoreThan2WavesPerBlock = wavesPerBlock >= 2;
```
sjw36: This should be wavesPerBlock > arch.numEUsPerCU. If there are 4 EUs per CU then a block must have more than 4 waves to cause multiple waves per EU.

In fact this is a hard factor in the `waves_per_eu` setting such that:

```
int32_t numWavesPerBlockPerEU = (wavesPerBlock + numEUPerCU) / numEUPerCU; // ceil

waves_per_eu = max(numWavesPerBlockPerEU, /* in the case of low LDS usage */ 2 ); // attr setting
```

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
```diff
@@ -330,6 +323,76 @@ void LowerRockOpsToGPUPass::runOnOperation() {
...
+      bool hasMoreThan2WGsPerCU = wgsPerCU >= 2;
```
sjw36: This factor doesn't make sense. For `waves_per_eu` the WG can factor in many ways. If a WG only has 1 wave, it would require more than 4 WGs scheduled in a CU simultaneously to need waves_per_eu. The max number of WGs doesn't mean they will all be resident simultaneously on the GPU. They can be run consecutively; i.e. 1 runs to completion and is evicted from the CU, free'ing up the CU for the next WG.

I don't think this factor should be considered but it may result in less application of `waves_per_cu`. 

The main factors are (1) LDS usage of a WG, and (2) register usage. We know LDS usage but we don't know register usage. I believe we can guestimate register usage based on LTA of private memref allocs.

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
krzysz00: Yeah, numCU's isn't an architectural property.
Remember, `gfx1100` covers multiple different models of Navi31, which have different core counts.

manupak: Done.

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
manupak: Adjusted to use the ceil but employed a different strategy. See the reply below.

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
manupak: I see your point, but Im not convinced number of WGs are totally unreleated.

What I did instead now is : I calculated totalWaves ( = wavesPerBlock * gridSize ) / totalEUs ( numEUPerCU * numCU ).
I think it addresses your point as well take into the account that we should only be using waves_per_eu if there are sufficient waves (in the whole grid not just within a WG) being launched.

I think register usage is a dependent variable -- which can be forced to go down with spills (but lemme confirm that with the compiler team). Therefore, we should only be using waves_per_eu when there is less usage of lds and when there is sufficient number waves being launched for the kernel
NOTE : thanks to your point, I realized it should be the number of waves in the whole grid rather than WGs.

manupak:
ok the final logic is 

```c++
      int64_t totalEUs = archInfo.numEUPerCU * numCU;
      int64_t wavesPerBlock = (blockSize / archInfo.waveSize);
      int64_t totalWaves = wavesPerBlock * gridSize;
      int64_t wavesPerEUPerBlock = wavesPerBlock / archInfo.numEUPerCU;
      int64_t wavesPerEUPerGrid = (totalWaves + totalEUs - 1) / totalEUs;
      int64_t wavesPerEU = std::max(wavesPerEUPerBlock, wavesPerEUPerGrid);
      if (ldsUsage > 0) {
        wavesPerEU =
            std::min(wavesPerEU, archInfo.totalSharedMemPerCU / ldsUsage);
      }
```

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
```diff
@@ -616,3 +618,62 @@ mlir::rock::transposeSubTileViews(PatternRewriter &rewriter, Location loc,
                                 std::nullopt};
   }
 }
+
+template <typename RetAttrType>
+static FailureOr<RetAttrType>
+getAttrFromOpOrParents(Operation *op, StringRef opAttr,
+                       std::optional<StringRef> maybeFuncAttr = std::nullopt,
+                       std::optional<StringRef> maybeModAttr = std::nullopt) {
```
sjw36: Since there are only 2 kinds of attr names, 
1) property/required attributes (not dialect qualified), and 
2) discardable attributes (which should always be dialect-qualified)

this can be simplified to have just one `maybeDialectAttr` for the discardable attr (on Func and Module).

manupak: done!

sjw36 (approved):
After discussions, this seems reasonable, however somewhat restrictive based on gridSize > numCUs (i.e. 1 WG per CU). The 2 main factors are LDS size (known before backend) and Register size (unknown before backend) based on lifetime analysis. Until we can guestimate register allocation, this solution suffice. But we may get better results by tuning this value.

manupak:
Thanks @sjw36! agreed that we need a handle/model to predict register allocation to avoid a forced-spill if we are to go beyond `waves-per-eu` > 2.

PR #1440, opened 2024-03-01
[Re-opened] Implemeted Split-K GEMM with coord. trans.

ravil-mobile:
The original PR is here: https://github.com/ROCm/rocMLIR/pull/1418

I don't know what happened today with github, but my PR didn't get updated from my branch. I though that closing and re-opening would help but it turned out that I didn't have enough rights to re-open a PR. In short, here is a new one. It has the same stuff as the original PR has

Closes https://github.com/ROCm/rocMLIR-internal/issues/1380

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
@@ -20,6 +20,7 @@
 //
 //===-----------------------------------------------------===//
 #include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/MHAL/IR/MHAL.h"
 #include "mlir/Dialect/Rock/IR/GemmSize.h"
 #include "mlir/Dialect/Rock/IR/Rock.h"
 #include "mlir/Dialect/Rock/IR/TransformMapBuilder.h"
@@ -258,10 +259,26 @@ std::tuple<Value, Value, Value>
 GemmRewritePattern::arrangeSplitKTransform(OpBuilder &builder, GemmOp op,
                                            Location loc, int64_t splitKFactor,
                                            Value a, Value b, Value c) const {
+  // adjust the store method
   auto storeMethod =
       builder.getAttr<rock::StoreMethodAttr>(rock::StoreMethod::AtomicAdd);
   op.setStoreMethodAttr(storeMethod);
 
+  // set the prefill attribute
+  auto func = llvm::cast<func::FuncOp>(op->getParentOp());
+  auto attrName = mhal::PrefillAttr::getMnemonic();
+  auto elementType = c.getType().cast<MemRefType>().getElementType();
+  Attribute zero;
+  if (llvm::isa<FloatType>(elementType)) {
+    zero = builder.getFloatAttr(elementType, 0.0);
+  } else {
+    assert(llvm::isa<IntegerType>(elementType) &&
+           "expecting `int` element type");
+    zero = builder.getIntegerAttr(elementType, 0);
+  }
+  func.setArgAttrs(2, builder.getNamedAttr(attrName, zero));
+
+  // perform coordinate transformations
   Value aNew{nullptr}, bNew{nullptr}, cNew{nullptr};
   ArrayRef<int64_t> aShape = a.getType().cast<MemRefType>().getShape();
   ArrayRef<int64_t> bShape = b.getType().cast<MemRefType>().getShape();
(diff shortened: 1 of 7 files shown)


mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
```diff
@@ -235,6 +266,127 @@ GemmRewritePattern::matchAndRewrite(GemmOp op, GemmOpAdaptor adaptor,
   return success();
 }
 
+std::tuple<Value, Value, Value>
+GemmRewritePattern::arrangeSplitKTransform(OpBuilder &builder, GemmOp op,
```
manupak, line 270: Can we add a single test to demo this behaviour (mostly as documentation) in https://github.com/ROCm/rocMLIR/blob/develop/mlir/test/Dialect/Rock/gemm_to_gridwise.mlir ?

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
```diff
@@ -170,13 +177,37 @@ GemmRewritePattern::matchAndRewrite(GemmOp op, GemmOpAdaptor adaptor,
   // Note, matrix dimension correctness is handled in the verifier
   GemmSize size(/*g=*/aShape[0], /*m=*/aShape[2], /*k=*/aShape[1],
                 /*n=*/bShape[2]);
+
+  GemmSize scale{1, 1, 1, 1};
```
manupak: Im just curious to know why are we are adding a generic scaling for all dims ?
(Id rather not add something like that unless we have some long term plan)

mlir/test/e2e/gemm_with_atomic_add.toml
```diff
@@ -9,7 +9,7 @@ prefix = "--store-method="
 
 [[axis]]
 name = "data type"
-values = ["f32", "f16"]
+values = ["f32", "f16 -RMS_threshold=1e-3"]
```
manupak, line 12: I thought we are not supporting f16

manupak (commented):
Mostly looking good. 

I may have missed some conversations around `scale`. I d like to understand why we need to scale all the dims (if not now, why in future?). I d rather not add a complexity that we are not going to test or use. However, i d like to learn your rationale here

ravil-mobile:
> I may have missed some conversations around `scale`. I d like to understand why we need to scale all the dims

Hi @manupak, It is a good question. So, we usually pad matrices based on block sizes. In the case of `split-k`, I need to pad matrices based on two factors - i.e., the block sizes and `split-k` factor. For example, let's consider `kPerBlock=64`, `splitKFactor=3` and `K=779`. If I pad as before, I get `K_{pad} = 832` but it is not divisible by `3`. So, the correct padding should be `960` - i.e., the padding factor `64 * 3 = 192`. In this case, I will have `15` blocks along the k-th dimension - i.e., `5` blocks per each `k`-split. The benefit is that the `gridwise-to-blockwise` pass becomes intact.

I use `scale` to account for the `split-K` factor - i.e., to scale block sizes

https://github.com/ROCm/rocMLIR/blob/d0009d38d297d21718099ef0a50d30b1b9ca4c34/mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp#L124-L134

mlir/test/e2e/gemm_with_atomic_add.toml
ravil-mobile: Currently, you can use `GEMMs` with `fp16` and the `atomic-add` as the store method. The split-k GEMM doesn't allow to do it. I hope that @giuseros has already figured out what was wrong and the fix is coming. This test doesn't fail. I would leave it because I am afraid that I would forget to add it again in the future.

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
ravil-mobile: Hi @manupak, here is the test:

https://github.com/ROCm/rocMLIR/blob/09b8348ed385c1c0f527a9738a0537c725cdbad6/mlir/test/Dialect/Rock/gemm_to_gridwise.mlir#L83-L99

`kpackPerBlock * kpack = 8`, `split-k = 3` and the original `K` length is equal to `238`. In this case, the correct padding should be `240`. The test confirms it

manupak (approved):
LGTM!

ravil-mobile:
> LGTM!

Thanks for the review!

PR #1442, opened 2024-03-04
Add support for tuningRunner to tune post highlevel modules

manupak:
This commit allows tuningRunner to tune modules that have
 rock operations in it -- usually what you see after highlevel.


krzysz00 (approved):
Looks fine to me

PR #1443, opened 2024-03-05
Fix Dockerfile.migraphx-ci

manupak:
We should only be changing the removal package
name once MIGraphX has changed the package name.
Currently, there is a mismatch between the package we try to remove and what MIGraphX installs - This commit fixes this.

### DIFF
--- a/mlir/utils/jenkins/Dockerfile.migraphx-ci
+++ b/mlir/utils/jenkins/Dockerfile.migraphx-ci
@@ -15,7 +15,9 @@ RUN rbuild prepare -d $PWD -s develop --cxx=/opt/rocm/llvm/bin/clang++ --cc=/opt
 # Remove rocMLIR version as per MIGraphX requirements
 # as this image is being used for CI with tip of rocMLIR
 # that will be built in the job.
-RUN cget remove -p $PWD ROCm/rocMLIR -y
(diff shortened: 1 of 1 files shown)


mlir/utils/jenkins/Dockerfile.migraphx-ci
```diff
@@ -15,7 +15,9 @@ RUN rbuild prepare -d $PWD -s develop --cxx=/opt/rocm/llvm/bin/clang++ --cc=/opt
 # Remove rocMLIR version as per MIGraphX requirements
 # as this image is being used for CI with tip of rocMLIR
 # that will be built in the job.
-RUN cget remove -p $PWD ROCm/rocMLIR -y
+# The rocMLIR package name should exactly match : https://github.com/ROCm/AMDMIGraphX/blob/develop/requirements.txt#L31
+# Otherwise, that package will not be uninstalled and we will be testing always against that.
+RUN cget remove -p $PWD ROCmSoftwarePlatform/rocMLIR -y
```
krzysz00, line 20: You should go poke MIGraphX to update their cget with the new repository name

krzysz00: Or update the commit we're pulling the requirements from, if that's hardcoded

manupak:
https://github.com/ROCm/AMDMIGraphX/pull/2904/files fixes it in MIGraphX. Thus no need for this.

PR #1447, opened 2024-03-06
Add support for `raw_buffer_atomic_fadd` in the fp16 case using packed logic

giuseros:
With this PR `raw_buffer_atomic_fadd` now supports (2-sized) vectors of fp16. 

Specific profile test case (using PR https://github.com/ROCm/rocMLIR/pull/1440):
```
a) --kernel-repeats 10   -t f16 -out_datatype f16 -g 1 -m 2 -k 1280 -n 4480 --perf_config=32,32,8,16,16,8,1,1
1.71972 TFlops
b) --kernel-repeats 10  --spilt-k 4 --transC=true  -t f16 -out_datatype f16 -g 1 -m 2 -k 1280 -n 4480 --perf_config=32,32,8,16,16,8,1,1
2.39983 TFlops
c) --kernel-repeats 10  --split-k 4 -t f16 -out_datatype f16 -g 1 -m 2 -k 1280 -n 4480 --perf_config=32,32,8,16,16,8,1,1
2.43473 TFlops
```
The difference between b) and c) is that in b) the matrix in a "good" vectorizable shape, while in c) the matrix is not and we pad every fp16 element in a vector of 2 elements to use the intrinsic (but the perf still seems good)

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
@@ -895,6 +895,7 @@ static Value computeMemRefNumElements(OpBuilder &b, Location loc,
 static void atomicFp16AddAligned(OpBuilder &b, Location loc, Value data,
                                  Value dest, ArrayRef<Value> coords) {
   const bool useBufferOobChecks = true;
+  const int packedVectorLen = 2;
   // Useful consts
   Value zero = b.create<arith::ConstantIntOp>(loc, 0, 32);
   Value one = b.create<arith::ConstantIntOp>(loc, 1, 32);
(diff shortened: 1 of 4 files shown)


giuseros:
This needs tests to be added, but for now let's have it here so that if MIGraphX integrates split-k they can enable it on the fp16 case with this

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
```diff
@@ -888,6 +888,42 @@ static Value computeMemRefNumElements(OpBuilder &b, Location loc,
...
+  // Extended packed data to use with the intrinsic
+  Value dataExt =
+      createZeroConstantOp(b, loc, vectorTypeOrSelf(b.getF16Type(), 2));
+  Value dataExt0 = b.create<vector::InsertElementOp>(loc, data, dataExt, zero);
+  Value dataExt1 = b.create<vector::InsertElementOp>(loc, data, dataExt, one);
+
+  // Manual alignment logic : if (addr % 2 != 0) step{AddressData}Back
+  Value stepBack = b.create<arith::SubIOp>(loc, coords[2], one);
+  Value alignment = b.create<arith::RemUIOp>(loc, coords[2], two);
+  Value notAligns =
```
krzysz00, line 923: As a note: when `threadwise_write_all` gets lowered, we're doing a vectorization query. That vectorization query also implicitly gives the alignment (I can probably adjust my new vectorization API to explicitly pass the final computed alignment back). So we only need this logic when we know we'll be in odd coordinates.

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
```diff
@@ -888,6 +888,42 @@ static Value computeMemRefNumElements(OpBuilder &b, Location loc,
...
+
+  // Extended packed data to use with the intrinsic
+  Value dataExt =
+      createZeroConstantOp(b, loc, vectorTypeOrSelf(b.getF16Type(), 2));
+  Value dataExt0 = b.create<vector::InsertElementOp>(loc, data, dataExt, zero);
+  Value dataExt1 = b.create<vector::InsertElementOp>(loc, data, dataExt, one);
+
+  // Manual alignment logic : if (addr % 2 != 0) step{AddressData}Back
+  Value stepBack = b.create<arith::SubIOp>(loc, coords[2], one);
+  Value alignment = b.create<arith::RemUIOp>(loc, coords[2], two);
```
krzysz00: Also, I don't think `[2]` is guaranteed to be the coordinate you want to query on.

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
krzysz00: ... that might be implicit in the fact that we've gone down this branch, never mind

krzysz00 (commented):
I'm seeing a few things that concern me in the fallback path

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
krzysz00: I think you'll need to find the last non-unit dimension

Consider, for example, `store %buf[%x, %c0] : memref<4x1xf16>`

krzysz00: Ping on that

giuseros: Done. I also added a couple of tests.

krzysz00 (approved):
Approved

(Looks like codecov's gotten annoyed about the lack of tests for the trailing unit dimension case but that's not blocking)

PR #1450, opened 2024-03-08
Fix the parse datatype selection in perfRunner

giuseros:
Fixes https://github.com/ROCm/rocMLIR-internal/issues/1408


jerryyin (approved):
Irrelevant with your PR I particularly doesn't like this 2 as hardcoded magic number

```python
        if len(dt) == 2:
            outMap[dt[0]] = dt[1]
```

PR #1451, opened 2024-03-08
Appended `perfConfig` string with `splitKFactor`

ravil-mobile:
Note, this PR generated two more tickets

- https://github.com/ROCm/rocMLIR-internal/issues/1410 (improvement)
- https://github.com/ROCm/rocMLIR-internal/issues/1411 (bug)

Closes ROCm/rocMLIR-internal#1389


mlir/include/mlir/Dialect/Rock/Tuning/Serializable.h
```diff
@@ -60,6 +61,15 @@ template <class Derived, char Seperator = ','> struct Serializable {
 
   bool deserialize(const std::string &s) {
     auto out = static_cast<const Derived &>(*this);
+
+    const auto numCommas =
```
krzysz00: Let's not do this hack.

Instead, could we, for example, make the perf config start with `v2:`?

If we see a perf config that doesn't start with `v2:`, we know it's an old perf config that doesn't have the split-K factor (so we can transparently add 1).

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -396,137 +396,137 @@ LogicalResult PopulateParamsAccel::obtainTuningParameters(
...
-  {32, 32, 8, 16, 16, 8, true, true},
-  {32, 32, 4, 16, 16, 4, true, true},
-  {16, 64, 8, 16, 16, 8, true, true},
-  {16, 64, 4, 8, 64, 4, true, true},
-  {16, 64 ,4, 16, 16, 8, true, true},
-  {16, 32, 4, 16, 16, 8, true, true},
-  {16, 32, 4, 16, 16, 4, true, true},
-  {16, 16, 4, 16, 16, 8, true, true},
-  {8, 64, 8, 8, 64, 8, true, true}
+  // M/block N/block K/block M/wave N/wave kPack splitKFactor forceUnroll bCopyMore
```
krzysz00, line 442: I'd like these to match the order they appear in in the perf config string (I think they do, just making sure)

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -55,6 +58,52 @@ void createAttnTuningRangeBF(TuningParamSet *newSpace, AttentionOp attnOp,
   }
 }
 
+auto getOptinalSplitKFactors(RockGemmWrapperInterface gemmOp,
```
krzysz00: No auto on return types please.

And perhaps `SmallVectorImpl<int64_t>& factors`

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -55,6 +58,52 @@ void createAttnTuningRangeBF(TuningParamSet *newSpace, AttentionOp attnOp,
...
+
+  struct LocalData {
+    int64_t splitKFactor{};
+    double workImbalance{};
+  };
+
+  const auto numCUs = gemmOp.getNumCU().value();
+  std::vector<LocalData> factors{};
+  for (int32_t splitKFactor = 1; splitKFactor < 100; ++splitKFactor) {
+    auto kPerBlock = math_util::lcm(gemmKPerBlock * kPack, splitKFactor);
```
krzysz00: LLVM style limits auto to ones that're obvious from context (ex, you'd be repeating the type name from a `cast<>` or where the underlying type is wordy and not that important. I'd like the integer types written out explicitly here.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -55,6 +58,52 @@ void createAttnTuningRangeBF(TuningParamSet *newSpace, AttentionOp attnOp,
...
+  }
+
+  struct LocalData {
+    int64_t splitKFactor{};
+    double workImbalance{};
+  };
+
+  const auto numCUs = gemmOp.getNumCU().value();
+  std::vector<LocalData> factors{};
+  for (int32_t splitKFactor = 1; splitKFactor < 100; ++splitKFactor) {
```
krzysz00: Why 100? And is a unit step the right move here?

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2229,7 +2229,8 @@ static func::FuncOp createGpuGemmKernel(ModuleOp module,
   if (!params.perfConfig.empty())
     gemm->setAttr("perf_config", b.getStringAttr(params.perfConfig));
 
-  gemm->setAttr("split-k-factor", b.getI32IntegerAttr(splitkFactor));
+  if (splitKFactor > 1)
```
krzysz00: Why? This'll be part of the perf config?

krzysz00 (changes requested):
I've got concerns about how we're doing this migration, especially in that, now that we're doing a migration, we might as well make it future-proof.

Also, check the parameter sweep scripts - they'll potentially be impacted by this.

krzysz00:
... also, while we're here, can we fix `kPerBlock` to actually be in units of K and not be a misnamed kpacks-per-block?

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: Ok. But let's not return a reference to a SmallVector.

krzysz00: A lot of functions take `SmallVectorImpl<T>& out` as a parameter so they can explicitly write onto tho caller's stack.

ravil-mobile:
Hi @krzysz00,

Thanks for your quick reply. I've just refactored the implementation. Found some mistakes and inefficiencies. Could you, please, have another look at it?

I still need to fix all the tests which make use of the `perfConfig` string. Meanwhile, @giuseros and I can start testing performance with this patch.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: I reduced to `32`. Regarding the unit step, I am doing filtering after that loop. I'm trying to select the variants which result in lowest work imbalance and then consider only 6 of them.

There is one more filter; the expected gain (based on the work imbalance regarding the default (data parallel) case) must be greater than 30%.

I would say these hard-coded constants are some hyper-parameters which require some consideration. We should discuss it.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
ravil-mobile: I was thinking that there may be a scenario where one wants to experiment with just matrix sizes and `splitKFactor` without bothering about tiling (via the `perfConfig` string).

What do you think?

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: Understood!

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: Ok. I though we were talking about the return parameter. I made it `SmallVector`. I usually rely on the return value optimization.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
ravil-mobile: So, `forceUnroll, bCopyMore` are optional for `NonAccel` type. Therefore, I decided to move `splitKFactor` to the 7th place because it appears in both `Accel` and `NonAccel` variants.

mlir/include/mlir/Dialect/Rock/Tuning/Serializable.h
ravil-mobile: Ok. Let me try it.

ravil-mobile: Hi @krzysz00 

What do you think about the following?

https://github.com/ROCm/rocMLIR/blob/f78a327f85979cd6f37599abe5d5c6f598acaede/mlir/include/mlir/Dialect/Rock/Tuning/Serializable.h#L77-L94

This implementation supports the old and new formats. It can be additionally extended if needed

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: TODO: add comment about the upper bound

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
ravil-mobile: Agree. Let's remove it

ravil-mobile: done

ravil-mobile:
> I've got concerns about how we're doing this migration, especially in that, now that we're doing a migration, we might as well make it future-proof.
> 
> Also, check the parameter sweep scripts - they'll potentially be impacted by this.

~~Hi @krzysz00, I looked through `parameter sweep` (https://github.com/ROCm/rocMLIR/blob/develop/mlir/utils/performance/parameterSweeps.py). It seems to me that it was designed to only test `ConvOp`s~~

Hi @krzysz00, I added the corresponding changes to `parameterSweeps.py`. Can you have a look?

ravil-mobile:
Hi @krzysz00, I reworked the PR based on our discussion on Monday. Could you, please, have another look at it?

mlir/include/mlir/Dialect/Rock/Tuning/Serializable.h
```diff
@@ -52,15 +62,50 @@ template <class Derived, char Seperator = ','> struct Serializable {
...
+
+    if (!checkVersionFormat(s)) {
+      // incorrect perf config format
+      return false;
+    }
+
     auto out = static_cast<const Derived &>(*this);
-    bool ok = true;
+
+    bool ok{true};
```
krzysz00: Can you put this back into `bool ok = true` since that's how these get written around here?

mlir/include/mlir/Dialect/Rock/Tuning/Serializable.h
```diff
@@ -52,15 +62,50 @@ template <class Derived, char Seperator = ','> struct Serializable {
...
+    const int32_t maxNumTokens = version == Version::V1 ? 8 : 9;
+    const int32_t maxNumSeperators = maxNumTokens - 1;
+    const int32_t minNumSeperators = maxNumSeperators - 2;
+    const auto numFoundSeperators = std::count_if(
+        s.begin(), s.end(), [](char c) { return c == Seperator; });
+    return numFoundSeperators >= minNumSeperators &&
+           numFoundSeperators <= maxNumSeperators;
+  }
+
+  bool deserialize(std::string s) {
```
krzysz00, line 82: While we're here, can we switch this to `StringRef`?

krzysz00: Or does the rest of the API not allow for that?

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -122,14 +122,14 @@ mlir::rock::requiredPadding(Attribute params, GemmSize gemmSize,
   int64_t kPerBlock, mPerBlock, nPerBlock;
   int64_t kPack = 1;
   if (auto generalParams = params.dyn_cast<GeneralGemmParamsAttr>()) {
-    kPerBlock = math_util::lcm(generalParams.getKPerBlock(), extraPadding->k);
-    mPerBlock = math_util::lcm(generalParams.getMPerBlock(), extraPadding->m);
-    nPerBlock = math_util::lcm(generalParams.getNPerBlock(), extraPadding->n);
+    kPerBlock = generalParams.getKPerBlock() * extraPadding->k;
```
krzysz00: Why this change?

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -15,7 +15,10 @@
 #include "mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h"
 #include "mlir/Dialect/Rock/Tuning/RockTuning.h"
 #include "mlir/Dialect/Rock/utility/AmdArchDb.h"
+#include "mlir/Dialect/Rock/utility/math.h"
 #include "llvm/ADT/SmallString.h"
+#include <algorithm>
+#include <vector>
```
krzysz00: Stray include?

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -15,7 +15,10 @@
 #include "mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h"
 #include "mlir/Dialect/Rock/Tuning/RockTuning.h"
 #include "mlir/Dialect/Rock/utility/AmdArchDb.h"
+#include "mlir/Dialect/Rock/utility/math.h"
 #include "llvm/ADT/SmallString.h"
+#include <algorithm>
```
krzysz00, line 23: Do we need this? Can you migrate to one of the `llvm::foo`s that takes a `Range` instead?

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -55,6 +59,96 @@ void createAttnTuningRangeBF(TuningParamSet *newSpace, AttentionOp attnOp,
   }
 }
 
+SmallVector<int64_t>
+computeOptimalSplitKFactors(RockGemmWrapperInterface gemmOp,
+                            int32_t gemmMPerBlock, int32_t gemmNPerBlock,
+                            int32_t gemmKPerBlock, int32_t kPack) {
+  auto info = PopulateParamsInfo::fromOp(gemmOp);
+  SmallVector<int64_t> splitKValues{1};
```
krzysz00: I'd prefer `= {1}` here

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -55,6 +59,96 @@ void createAttnTuningRangeBF(TuningParamSet *newSpace, AttentionOp attnOp,
...
+  constexpr double imbalaceThreshold{1.20};
+  if (dataPrallelGemmImbalabce < imbalaceThreshold) {
+    return splitKValues;
+  }
+
+  struct LocalData {
+    int64_t splitKValue{};
+    double workImbalance{};
+  };
+  SmallVector<LocalData> factors{};
```
krzysz00: Similar style nitpicks: we don't do this `{}` constructor in this codebase.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -55,6 +59,96 @@ void createAttnTuningRangeBF(TuningParamSet *newSpace, AttentionOp attnOp,
...
+    if (gain > minGain) {
+      factors.emplace_back(LocalData{splitKFactor, imbalance});
+    }
+  }
+
+  if (factors.empty()) {
+    return splitKValues;
+  }
+
+  std::sort(factors.begin(), factors.end(), [](LocalData &a, LocalData &b) {
```
krzysz00: I think there's a `llvm::sort` running around?

mlir/utils/performance/parameterSweeps.py
```diff
@@ -34,13 +34,30 @@ class Options:
     flags: list
     concurrent_tests: int
 
+class PerfConfig:
+    class Version(enum.Enum):
```
krzysz00, line 38: Given that this is an internal utility, I think we're good to just make this be `v2` by definition?

krzysz00 (commented):
Comments, mostly style

mlir/include/mlir/Dialect/Rock/Tuning/Serializable.h
ravil-mobile: Originally, it was `const std::string &s`. So, we used to provide the caller the guarantee that the input string is not going to be modified. Now, we need to extract the suffix from the `s` - i.e., everything after `v2:`. 

https://github.com/ROCm/rocMLIR/blob/cb3e1e6676c6636ba37e2c5fb6136972c9c4ec9b/mlir/include/mlir/Dialect/Rock/Tuning/Serializable.h#L89-L92

Even if we accept a reference to `s`, we need to write a suffix to a new string. We would need to allocate a new temporary variable for the suffix. All in all, we can also accept the input string just by value. it will result in the same complexity and grantees that user's input is not going to be modified.

mlir/include/mlir/Dialect/Rock/Tuning/Serializable.h
ravil-mobile: Done

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: removed them. Done

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: Done

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: done

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
ravil-mobile: The change was related to https://github.com/ROCm/rocMLIR-internal/issues/1411.

Now, I came up to a better way to do padding for the Split-K, which require neither `math_util::lcm` nor `generalParams.getKPerBlock() * extraPadding->k`. I reverted this changes to the state as it was before `Split-K`

mlir/utils/performance/parameterSweeps.py
ravil-mobile: I can remove `V1` from `PerfConfig` class but let's leave the class. At least, it knows how to print the `perfConfig` string correctly based on a requested version. Otherwise, we will end up with lot's magic constant in the code.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: Yes, you're right. Done

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: Yes, I used `llvm::for_each` which uses ranges,
done

manupak (changes requested):
I have not looked into detail yet -- which I may do at some point if required.

But there is a critical test required here that shows a rejection of a migraphx module that shows a fused cases with a split-k perfconfig is rejected.

manupak:
and it would be better if we dont return back the split-K perfConfigs for fusion cases until we have the other isSplitK API.

mlir/include/mlir/Dialect/Rock/Tuning/Serializable.h
krzysz00: My claim is that we can throw all this `visit` stuff out the window and just go to parsing on a `StringRef` because we shouldn't be modifying anything.

mlir/utils/performance/parameterSweeps.py
krzysz00: Fair enough, it's not an unreasonable restructuring

ravil-mobile: Updated

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -55,6 +58,78 @@ void createAttnTuningRangeBF(TuningParamSet *newSpace, AttentionOp attnOp,
...
+  const double numNTiles = gemmSize.n / gemmNPerBlock;
+
+  auto computeImbalance = [&](int32_t splitKFactor) -> double {
+    const double totalNumWorkGroups = numMTiles * numNTiles * splitKFactor;
+    const double maxWorkGroupsPerCU = std::ceil(totalNumWorkGroups / numCUs);
+    // imbalances = max. CU work / average work per CU
+    return (maxWorkGroupsPerCU * numCUs) / totalNumWorkGroups;
+  };
+
+  const auto dataPrallelGemmImbalabce = computeImbalance(1);
```
manupak: nit: spelling

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -55,6 +58,78 @@ void createAttnTuningRangeBF(TuningParamSet *newSpace, AttentionOp attnOp,
...
+  const double numCUs = gemmOp.getNumCU().value();
+
+  const InitParams params{gemmMPerBlock, gemmNPerBlock, gemmKPerBlock};
+  const GemmSize gemmSize =
+      calculatePaddedGemmSize(params, info.gemmSize, kPack);
+  const double numMTiles = gemmSize.m / gemmMPerBlock;
+  const double numNTiles = gemmSize.n / gemmNPerBlock;
+
+  auto computeImbalance = [&](int32_t splitKFactor) -> double {
+    const double totalNumWorkGroups = numMTiles * numNTiles * splitKFactor;
```
manupak: This is not right;

it should be gemmSize.g * numMTiles * numNTiles * splitKFactor

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -55,6 +58,78 @@ void createAttnTuningRangeBF(TuningParamSet *newSpace, AttentionOp attnOp,
...
+  SmallVector<LocalData> factors;
+  constexpr double minGain = 1.30;
+  // There are cases where perfect load balancing can be achieved with very
+  // high splitK values. However, experiments show that performance
+  // can considerably drop in such cases. Currently, we limit the `upperBound`
+  // on purpose because the current heuristics does not consider the overheads
+  // resulting from reducing partial solution along the split dimension.
+  // This needs to be improved in the future.
+  constexpr int32_t upperBound = 32;
+  for (int32_t splitKFactor = 2; splitKFactor < upperBound; ++splitKFactor) {
```
manupak, line 118: We need to revisit this in future. 
I think the termination critierion for this loop should be : kPerSplit < lambda * kPerBlock. The lambda should be emhprically decided based on occupancy & knee-point in kIters in the pipelined Kloop. Please create a ticket for this.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -55,6 +58,78 @@ void createAttnTuningRangeBF(TuningParamSet *newSpace, AttentionOp attnOp,
...
+    if (gain > minGain) {
+      factors.emplace_back(LocalData{splitKFactor, imbalance});
+    }
+  }
+
+  if (factors.empty()) {
+    return splitKValues;
+  }
+
+  llvm::sort(factors.begin(), factors.end(), [](LocalData &a, LocalData &b) {
```
manupak: FYI, one of the advantages of using llvm::sort is you dont need to supply begin and end iterators.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -55,6 +58,78 @@ void createAttnTuningRangeBF(TuningParamSet *newSpace, AttentionOp attnOp,
...
+    return a.workImbalance < b.workImbalance;
+  });
+
+  size_t maxVariants = 6;
+  maxVariants = factors.size() > maxVariants ? maxVariants : factors.size();
```
manupak: Who uses maxVariants ?

manupak (commented):
I had a look.

The major concern for me is this is a breaking change as it stands now.
We need a UnitAttr (that defaults to false) to enable split-k on `rock.gemm`.

This could be set via module attr (like num_cu) via the clients to indicate that they intend to tune for split-k and possible like to use split-k kernel if its winning. 

Then MIGraphX can set the boolean whenever they like to explore split-k where they provide non (limited) fusion cases. 

Additionally to that, we should set the said UnitAttr to false if you find fusion (linalg blocks trailing the rock.gemm).

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: Thanks @manupak. I've already created a ticket for it:

https://github.com/ROCm/rocMLIR-internal/issues/1410

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: Thanks! Done

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: Do you mean who can set it? At this point nobody. It is just a hard-coded value which is serving as limiting factor for the exhaustive tuning. Otherwise, it might take too long.

This PR is only about ennoblement the splitK through the `perfConfig` string.

The heuristic needs to be improved and the ticket is here: https://github.com/ROCm/rocMLIR-internal/issues/1410

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: Thanks for noticing it. You're correct!
Done

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: Good point. I've noticed the second variant, which takes only a container and a predicate. Thanks!
Done

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
manupak: No I mean from what I can read here maxVariants is dead code.
No one uses it ?

ravil-mobile:
Hi @krzysz00, I added the a guard to `AlignTiling` and a test which is expected to fail 


```mlir
// RUN: rocmlir-opt --rock-view-to-transform -rock-affix-params -rock-conv-to-gemm -rock-gemm-to-gridwise --rock-regularize --rock-gridwise-gemm-to-blockwise --rock-linalg-align %s -verify-diagnostics

#map0 = affine_map<(d0, d1, d2) -> (d0, d1, d2)>

module attributes {mhal.arch = "amdgcn-amd-amdhsa:gfx90a"} {

  func.func @rock_gemm(%arg0: memref<1x1024x1024xf32>, %arg1: memref<1x1024x512xf32>, %arg2: memref<1x1024x512xf32>) attributes {kernel, mhal.arch = "amdgcn-amd-amdhsa:gfx90a"} {
    %0 = memref.alloc() : memref<1x1024x512xf32>
    rock.gemm %0 = %arg0 * %arg1 features =  mfma|dot|atomic_add storeMethod = atomic_add {arch = "amdgcn-amd-amdhsa:gfx90a"} : memref<1x1024x512xf32> = memref<1x1024x1024xf32> * memref<1x1024x512xf32>

    // expected-error @+1 {{'linalg.generic' op lingalg generic ops are only allowed to operate with `Set` store method}}
    linalg.generic {indexing_maps = [#map0, #map0], iterator_types=["parallel", "parallel", "parallel"]} ins(%0 : memref<1x1024x512xf32>) outs(%arg2 : memref<1x1024x512xf32>) {
    ^bb0(%i: f32, %o: f32):
        %c0 = arith.constant 2.0 : f32
        %r = arith.addf %i, %c0 : f32
        linalg.yield %r : f32
    }
    return
  }
}
```

Hi @manupak, I added the corresponding `UnitAttr`. I did a few tests and can confirm that MIGraphX won't be able to pickup any perfConfig with splitK. The attribute is set in `rocmlir-gen` only which is enough for the development on our side.

manupak:
Summarizing the offline discussion with @ravil-mobile ,

Until we have discrepancy between what MIGraphX tune and tuningRunner.py tune, we need a way disable split-k tuning range from tuningRunner.py. Therefore a CLI arg has to be plumbed from tuningRunner.py -> rocmlir-gen to achieve this. I would prefer this to be an explicit enable (that by default is set to false) in the tuningRunner.

@krzysz00 see if you have objections to this request. I rather would have this until such a time MIGraphX uses split-k fully -- so that we dont have trouble reproducing perf issues reported by MIGraphX.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: Thanks for noticing it. Yep, I've reworked it based on our discussion

ravil-mobile:
> Summarizing the offline discussion with @ravil-mobile ,
> 
> Until we have discrepancy between what MIGraphX tune and tuningRunner.py tune, we need a way disable split-k tuning range from tuningRunner.py. Therefore a CLI arg has to be plumbed from tuningRunner.py -> rocmlir-gen to achieve this. I would prefer this to be an explicit enable (that by default is set to false) in the tuningRunner.
> 
> @krzysz00 see if you have objections to this request. I rather would have this until such a time MIGraphX uses split-k fully -- so that we dont have trouble reproducing perf issues reported by MIGraphX.


Thanks @manupak  for writing it down. I was thinking about it for a while. So, the requested feature is temporal. It seems to me that the flag is going affect many places in the code. The feature won't be needed after the split-K integration into MIGraphX. I am afraid that the additional flag (turn on/off split-k) will bring some ambiguity to the workflow. For example,

```
rocmlir-gen ... --enable-split-k --perf_config 64,64,16,32,32,4,1,1,1
``` 

is not going to generate the split-k variant. Another example,


```
rocmlir-gen ... --perf_config 64,64,16,32,32,4,10,1,1
``` 
will generate the code with split-k.

So, ` --enable-split-k` only enable/disable the corresponding attribute

https://github.com/ROCm/rocMLIR/blob/1d374392f30e077d416c2c8a54cf6476403746bb/mlir/tools/rocmlir-gen/rocmlir-gen.cpp#L2230-L2231

which is used only here

https://github.com/ROCm/rocMLIR/blob/1d374392f30e077d416c2c8a54cf6476403746bb/mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp#L74-L78

Again, my concern is only ambiguity and the fact that we need this flag ( ` --enable-split-k`) for a couple of months. 

I would suggest to add a compile-time switch through as a target definition which can be controlled by a CMake option. In this case, if somebody really needs to have tuning without split-k, then he/she can just re-compile rocMLIR with enabled CMake option. We will just use `#ifdef` macro at the top of `computeOptimalSplitKFactors` function. 

I consider this as an option for this PR because it will be easy to remove redundant code in the future -i.e., after the integration with MIGraphX.

ravil-mobile:
Hi @krzysz00 , one thing that I noticed while working on enabling fp16 for split-K (different PR). We use linalg generic for the conversion of the output type from LDS to HBM. For example, if we operate with `fp16` we accumulate results in `fp32` which then get converted back to `f16` while writing to HBM. 

So, in general, we cannot simply fail `AlignTiling` based on the store method

ravil-mobile:
Hi @krzysz00,

I think we need to change the guard in `AlignTiling` and implement it as follows

```cpp
  if (gemmStoreOp) {
    if (gemmStoreOp.getStoreMethod() != rock::StoreMethod::Set) {
      bool isLegal = true;
      for (auto &region : laGeneric->getRegions()) {
        for (auto &block : region) {
          for (auto &op : block) {
            bool isAllowed = llvm::isa<arith::TruncFOp>(op);
            isAllowed |= llvm::isa<arith::TruncIOp>(op);
            isAllowed |= llvm::isa<linalg::YieldOp>(op);
            if (!isAllowed) {
               isLegal = false;
               break;
           }
          }
        }
      }

      if (!isLegal)
        return laGeneric.emitOpError("lingalg generic ops are only allowed to "
                                     "operate with `Set` store method");
    }
  }
```

The reason that we can have a conversion `linalg.generic` when we use small data types (e.g., `f16`, `int8`). What do you think?


Or, it seems to me there is a better solution. We can put an attribute to those `linalg.generic` ops which we consider to be valid. For example,

https://github.com/ROCm/rocMLIR/blob/c6dafa1aaa93dc6711bea6028e20b2b94bc5d9ab/mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp#L243-L262

we can set an attribute:    ` op->setAttr("allowed", rw.getUnitAttr()); `

And then, in AlignTiling

we can do something like

```cpp

  if (gemmStoreOp) {
    if (gemmStoreOp.getStoreMethod() != rock::StoreMethod::Set) {
      if (!laGeneric->getAttr("allowed")) {
        return laGeneric.emitOpError("lingalg generic ops are only allowed to "
                                     "operate with `Set` store method");
      }
    }
  }
```

sjw36:
BTW, `isAllowed` is always true in the code above. The init value should be false.

ravil-mobile:
> BTW, `isAllowed` is always true in the code above. The init value should be false.

I mean we can do something

```cpp
  if (accumulator != c) {
    auto map = rw.getMultiDimIdentityMap(3);
    auto convertionOp = rw.create<linalg::GenericOp>(
        loc, ValueRange{accumulator}, ValueRange{c},
        ArrayRef<AffineMap>{map, map},
        ArrayRef<utils::IteratorType>{utils::IteratorType::parallel,
                                      utils::IteratorType::parallel,
                                      utils::IteratorType::parallel},
        /*doc=*/"", /*library_call=*/"",
        [](OpBuilder &builder, Location loc, ValueRange elems) {
          Value accumulator = elems[0], c = elems[1];
          Type cType = c.getType();
          if (cType.isa<IntegerType>()) {
            Value cElement =
                builder.create<arith::TruncIOp>(loc, cType, accumulator);
            builder.create<linalg::YieldOp>(loc, cElement);
          } else {
            Value cElement =
                builder.create<arith::TruncFOp>(loc, cType, accumulator);
            builder.create<linalg::YieldOp>(loc, cElement);
          }
        });
    convertionOp->setAttr("allowed", rw.getUnitAttr());
  }
```

Meaning, we allow this `linalg.generic` op to come with `atomicAdd` store method. In some sense, we mark this `linalg.generic` to be legal for `AlignTiling` pass.

ravil-mobile:
> BTW, `isAllowed` is always true in the code above. The init value should be false.

Yes, you're correct! Thanks for noticing. I've just corrected the snippet.

ravil-mobile:
Hi @krzysz00  and @manupak,

I tried to address the recent requests regarding the PR.

## 1. Improved the guard in `AlignTiling`

Now, we additionally check the result types of `arith::Trunc` operations

https://github.com/ROCm/rocMLIR/blob/79acd48afbb00624fb257fc4d085eb04b06712af/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp#L748-L774
https://github.com/ROCm/rocMLIR/blob/79acd48afbb00624fb257fc4d085eb04b06712af/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp#L808-L815


## 2. Additional guard

Added a forgotten guard which check whether the target machine supports `atomic_add`

https://github.com/ROCm/rocMLIR/blob/79acd48afbb00624fb257fc4d085eb04b06712af/mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp#L185-L188


## 3. Added Additional (temporary) Attribute 

https://github.com/ROCm/rocMLIR/blob/c6991eeebbd65e90ee30d2d7b11045ecd0ece7e6/mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td#L359-L363


## 4. Conditional tuning with Split-K

https://github.com/ROCm/rocMLIR/blob/79acd48afbb00624fb257fc4d085eb04b06712af/mlir/tools/rocmlir-gen/rocmlir-gen.cpp#L2236-L2239
https://github.com/ROCm/rocMLIR/blob/79acd48afbb00624fb257fc4d085eb04b06712af/mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp#L74-L78

There are two ways to turn off split-K in tuning

1) `./bin/tuningRunner.py --op=gemm --configs_file=<file> --output=bt.tsv --verify-mode=none --mlir-build-dir=$(realpath .) --rocmlir_gen_flags="--disable-split-k-for-tuning=true"`

2) `./bin/tuningRunner.py --op=gemm --configs_file=<file> --output=bt.tsv --verify-mode=none --mlir-build-dir=$(realpath .) --rocmlir_gen_flags="--atomic_add=off"`

manupak:
> For example,
> 
> ```
> rocmlir-gen ... --enable-split-k --perf_config 64,64,16,32,32,4,1,1,1
> ```
> 
> is not going to generate the split-k variant. Another example,
> 
> ```
> rocmlir-gen ... --perf_config 64,64,16,32,32,4,10,1,1
> ```
> 
> will generate the code with split-k.
> 
> So, ` --enable-split-k` only enable/disable the corresponding attribute
> 
> https://github.com/ROCm/rocMLIR/blob/1d374392f30e077d416c2c8a54cf6476403746bb/mlir/tools/rocmlir-gen/rocmlir-gen.cpp#L2230-L2231
> 
> which is used only here

Im afraid this rocmlir-gen example is slightly irrelevant to the point Im trying to make.

My concern was about **tuning** and in the **tuning** mode we dont provide a perf-config.
The **tuning** mode will populate the tuning range in rocmlir-tuning-driver looking at the rocmlir-gen'd module which has attribute `enable-split-k` (which is already present to handle migraphx integration).

That arg could be ignored/errored out if provided in a non-tuning context.

>> Again, my concern is only ambiguity and the fact that we need this flag (  --enable-split-k) for a couple of months.

Maybe... but I think until this feature is mature enough, we should have the flag at the disposal of developers of rocmlir.

> I would suggest to add a compile-time switch through as a target definition which can be controlled by a CMake option. In this case, if somebody really needs to have tuning without split-k, then he/she can just re-compile rocMLIR with enabled CMake option. We will just use `#ifdef` macro at the top of `computeOptimalSplitKFactors` function.
> 
> I consider this as an option for this PR because it will be easy to remove redundant code in the future -i.e., after the integration with MIGraphX.

I am (personally) not a fan of CMake option that makes a such a runtime change because it creates ambiguity between builds and branch changes where you'd have to clean build always.

I dont see the changes to make this runtime option are not very large.
As I have mentioned before and I repeat -- it should be two small changes in the two places : 
1) tuningRunner.py : add a new arg to 'enable-split-k-tuning' and pass into rocmlir-gen cmd line
2) rocmlir-gen : add llvm arg : 'enable-split-k-tuning' to take in the arg from tuningRunner.py from cli and set the already defined 'enable-split-k' flag in the module/func.

Above should be less than O(10) lines of code and Im not convinced that it is hard to remove these later on (if need be) as they are not invasive. These would save the trouble of recompiling and easy toggling of split-k enabled tuning.

manupak (approved):
LGTM! modulo @krzysz00 comments.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -741,6 +745,34 @@ static void reconfigureLAGeneric(LinalgAlignRewriter &b,
...
+static LogicalResult canFuseAcrossAtomic(LinalgAlignRewriter &b,
+                                         linalg::GenericOp laGeneric) {
+  bool isLegal = true;
+  for (auto &region : laGeneric->getRegions()) {
+    for (auto &block : region) {
+      for (auto &op : block) {
+        bool isTruncOp = llvm::isa<arith::TruncFOp, arith::TruncIOp>(op);
+        if (isTruncOp) {
+          OpResult result = op.getResult(0);
+          bool allowedType = result.getType() == b.getF32Type();
```
krzysz00: Why not put these two in one line?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -741,6 +745,34 @@ static void reconfigureLAGeneric(LinalgAlignRewriter &b,
...
 }
 
+static LogicalResult canFuseAcrossAtomic(LinalgAlignRewriter &b,
+                                         linalg::GenericOp laGeneric) {
+  bool isLegal = true;
+  for (auto &region : laGeneric->getRegions()) {
+    for (auto &block : region) {
+      for (auto &op : block) {
+        bool isTruncOp = llvm::isa<arith::TruncFOp, arith::TruncIOp>(op);
+        if (isTruncOp) {
```
krzysz00: 1. `trunci` needs a separate legality test - I don't think we're set up for atomics of i16 or less
2. Style wile, instead of those `isTruncOp`s, I'd've written
    ```c++
    if (auto truncfOp = dyn_cast<arith::TruncFOp>(op)) {
      ...
    }
    else if (auto truncIOp = dyn_cast<arith::TruncIOp>(op)) { ... }
    else if (isa<linalg::YieldOp>) { ... }
    else { isLegal = false; break; }
    ````
     (up to my hasty indentation)

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -741,6 +745,34 @@ static void reconfigureLAGeneric(LinalgAlignRewriter &b,
...
+
+        const bool isYieldOp = llvm::isa<linalg::YieldOp>(op);
+        if (!(isTruncOp || isYieldOp)) {
+          isLegal = false;
+          break;
+        }
+      }
+    }
+  }
+  return isLegal ? mlir::success() : mlir::failure();
```
krzysz00: 1. Don't need the `mlir::` here.
2. Neat little trick `return success(isLegal);` saves you the if operatior.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -55,6 +61,85 @@ void createAttnTuningRangeBF(TuningParamSet *newSpace, AttentionOp attnOp,
...
+                            int32_t gemmMPerBlock, int32_t gemmNPerBlock,
+                            int32_t gemmKPerBlock, int32_t kPack) {
+  auto info = PopulateParamsInfo::fromOp(gemmOp);
+  SmallVector<int64_t> splitKValues = {1};
+  if (!info.gemmAType.isF32()) {
+    return splitKValues;
+  }
+
+  // TODO[split-K]: remove after integrating split-K into MIGraphX
+  auto func = llvm::cast<func::FuncOp>(gemmOp->getParentOp());
```
krzysz00: `cast` is re-exported into `mlir` and so doesn't need a `llvm::`

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -55,6 +61,85 @@ void createAttnTuningRangeBF(TuningParamSet *newSpace, AttentionOp attnOp,
...
+  });
+
+  const size_t maxVariants = std::min(static_cast<size_t>(6), factors.size());
+  llvm::ArrayRef<LocalData> view(factors.data(), maxVariants);
+  llvm::for_each(view, [&](const LocalData &item) {
+    splitKValues.push_back(item.splitKValue);
+  });
+
+  return splitKValues;
+}
```
krzysz00, line 141: Since we're getting codecov warnings here that kinda make sense, can you add to or adapt `mlir/test/rocmlir-gen/emit-tuning-space.mlir` to test this split-K sorting heuristic?

mlir/test/rocmlir-driver/populate_pv_with_gpu.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: rocmlir-gen --arch gfx908  -p -mfma=on -dot=on -atomic_add=on -pv_with_gpu | FileCheck %s
+// RUN: rocmlir-gen --arch gfx908  -p -mfma=on -dot=on -atomic_add=on -pv_with_gpu -disable-split-k-for-tuning | FileCheck %s
```
krzysz00, line 2: Ah.

Note to future us: actually update these tests

krzysz00 (approved):
Approved with nitpicking - that is, I've got a few tiny irritations, but I don't see anything here I disagree with enough to stop this from merging (though I'd prefer the nitpicks get sorted)

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: done

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
ravil-mobile: Nice trick! done

mlir/test/rocmlir-driver/populate_pv_with_gpu.mlir
ravil-mobile: Good point. Done

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
ravil-mobile: Correct. done

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
ravil-mobile: Good point. I like the style. One thing, that I changed, was that I removed `break` from all branches and added the following to the end of the inner loop

```cpp
        if (!isLegal)
          break;
```

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: Yep. I added an additional test, which actually captured a small bug. Thanks!

PR #1445, opened 2024-03-05
Fix the blocksize check in the tuning stage

yiqian1:
Make the blocksize check in tuning to be aligned with what GridwiseGemmToBlockwise.cpp does.

Fix https://github.com/ROCm/rocMLIR/issues/1429

PR #1453, opened 2024-03-08
[BACKPORT] Fix the blocksize check in the tuning stage (#1445)

yiqian1:
* Fix the blocksize check in the tuning stage

* Update the test case

### DIFF
--- a/mlir/test/CAPI/mixr_full.c
+++ b/mlir/test/CAPI/mixr_full.c
@@ -194,7 +194,7 @@ static bool constructAndTraverseIr(MlirContext ctx) {
       mlirRockTuningSpaceCreate(module, RocmlirTuningParamSetKindFull);
   printf("Got tuning space,\n");
   unsigned fNum = mlirRockTuningGetNumParams(tuningSpace);
-  // CHECK: full set = 461
+  // CHECK: full set = 475
   printf("full set = %u\n", fNum);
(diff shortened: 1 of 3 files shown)


jerryyin (approved):
Let's get this merged by end of today so we will have a green on weekly CI.

mlir/test/CAPI/mixr_full.c
```diff
@@ -194,7 +194,7 @@ static bool constructAndTraverseIr(MlirContext ctx) {
       mlirRockTuningSpaceCreate(module, RocmlirTuningParamSetKindFull);
   printf("Got tuning space,\n");
   unsigned fNum = mlirRockTuningGetNumParams(tuningSpace);
-  // CHECK: full set = 357
+  // CHECK: full set = 461
```
yiqian1: There is a problem here. Due to some changes in GridwiseGemmParams.cpp not included in release6.1, this test case failed because it generated "full set = 475" here. Should I fix it in release6.1? 
@jerryyin @sjw36  @manupak @krzysz00

jerryyin: Oh just realized that this is a backporting PR. 

If you want to backport it, yah, this will have to be fixed to merge. I'd say just change the number to make test pass.

krzysz00 (approved):

PR #1454, opened 2024-03-11
Rework tuning to tune `mPerWave`,`mnPerXdl` and derive `nPerWave`

manupak:
Currently we tune for `mPerWave` and `nPerWave` (among other things) that
does two things : 
1) select number of waves -> hence blocksize
2) select mfma instructions (well; in CDNA context)

However, the issue with this approach is if we want to select a smaller
mnPerXDL mfma instruction (say 16x16) it will increase the number of
waves because dWaves = dPerBlock / dPerWave. This is not a true 
trade off -- i.e. one could have more repeats instead of more waves.

Moreover, in my experiments I found out the performant perfConfigs
most of the time contains around 4 waves. Thus, tuning `mPerWave` and `nPerWave`
and inferring/forcing repeats does not seem like the best way to do this.

Therefore, in this PR, I ve changed the tuning to tune `mPerWave` only and infer
'nPerWave' based on fixed parameter of maxWaves(=4). Then reusing (as well
as cutting down the tuning space) `nPerWave` -> `mnPerXdl`. The new tuning
space will offers the user to use `mPerWave` as a slider to balance the waves
between M & N dimensions while `mnPerXdl` offers to try out a variety of MFMA
instructions.

I ve tested it with our current gemm-configs on gfx90a and it improves performance
in many places 
(with two odd slight drops in cases where tflop count is already low).
See : 
[develop_vs_16x16_gfx90a.zip](https://github.com/ROCm/rocMLIR/files/14559558/develop_vs_16x16_gfx90a.zip)

More importantly this improves (mostly) performance on MI300X for sdxl gemm configs (well; what it used to be) as follows:

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/mkarunar/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/mkarunar/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">

  | develop (TFLOPS) | 16x16 forced (TFLOPs) | % improvement |   | maybe 16x16 + new tuning (TFLOPs) | % improvement | mfma type
-- | -- | -- | -- | -- | -- | -- | --
-t f16 -out_datatype f16 -transA false   -transB false -g 1 -m 154 -n 166400 -k 2048 | 199.1425414 | 211.3613268 | 6.135698205 |   | 216.3304601 | 8.630962829 | 16x16
-t f16 -out_datatype f16 -transA false   -transB false -g 1 -m 2 -n 1280 -k 2816 | 0.433282846 | 0.425583565 | -1.776964402 |   | 0.442348899 | 2.092409763 | 16x16
-t f16 -out_datatype f16 -transA false   -transB false -g 1 -m 2 -n 1280 -k 1280 | 0.369821116 | 0.369009009 | -0.219594595 |   | 0.360861186 | -2.422774026 | 16x16
-t f16 -out_datatype f16 -transA false   -transB false -g 1 -m 2 -n 1280 -k 320 | 0.202321561 | 0.204340234 | 0.997755051 |   | 0.228284799 | 12.83265982 | 16x16
-t f16 -out_datatype f16 -transA false   -transB false -g 1 -m 2 -n 13760 -k 1280 | 3.738852624 | 3.811881831 | 1.953251813 |   | 3.615663331 | -3.294842178 | 16x16
-t f16 -out_datatype f16 -transA true   -transB false -g 1 -m 8192 -n 640 -k 640 | 236.7489734 | 248.3489897 | 4.899711346 |   | 239.8115495 | 1.293596341 | 16x16
-t f16 -out_datatype f16 -transA false   -transB false -g 1 -m 8192 -n 1920 -k 640 | 374.2686496 | 419.1684197 | 11.99666875 |   | 429.5700429 | 14.77585508 | 16x16
-t f16 -out_datatype f16 -transA false   -transB false -g 1 -m 8192 -n 640 -k 640 | 246.3524247 | 243.6424049 | -1.100058089 |   | 236.5987308 | -3.95924413 | 32x32
-t f16 -out_datatype f16 -transA false   -transB false -g 1 -m 8192 -n 5120 -k 640 | 392.2344563 | 412.2830862 | 5.111389275 |   | 409.8564104 | 4.492709355 | 16x16
-t f16 -out_datatype f16 -transA false   -transB false -g 1 -m 8192 -n 640 -k 2560 | 305.7316614 | 307.9836345 | 0.736584862 |   | 338.6685373 | 10.77313213 | 32x32
-t f16 -out_datatype f16 -transA true   -transB false -g 1 -m 2048 -n 1280 -k 1280 | 208.4579381 | 212.962884 | 2.161081493 |   | 206.9089967 | -0.743047438 | 32x32
-t f16 -out_datatype f16 -transA false   -transB false -g 1 -m 2048 -n 3840 -k 1280 | 347.0430118 | 384.4384884 | 10.77545876 |   | 378.4192173 | 9.041013493 | 16x16
-t f16 -out_datatype f16 -transA false   -transB false -g 1 -m 2048 -n 1280 -k 1280 | 221.4156323 | 218.2473056 | -1.430940844 |   | 226.1918636 | 2.157133718 | 32x32
-t f16 -out_datatype f16 -transA false   -transB false -g 1 -m 2048 -n 10240 -k 1280 | 363.1677684 | 392.2975105 | 8.02101525 |   | 389.6693996 | 7.297352219 | 16x16
-t f16 -out_datatype f16 -transA false   -transB false -g 1 -m 2048 -n 1280 -k 5120 | 269.9797401 | 281.2049739 | 4.157805969 |   | 277.8173478 | 2.903035514 | 32x32

</body>

</html>

closes : https://github.com/ROCm/rocMLIR-internal/issues/1402


manupak:
Status : I have e2e tests all working but unit tests are yet to be fixed (and as it turns out there are conflicts as well). 

As this is somewhat a controversial change, I think we can start reviewing.

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
```diff
@@ -250,19 +250,53 @@ def Rock_GeneralGemmParamsAttr : Rock_Attr<"GeneralGemmParams", [RockTuningParam
...
+  let assemblyFormat = [{
+    `<` struct(params) `>`
+  }];
+}
+
+def Rock_XdlopsGemmDerivedParamsAttr : Rock_Attr<"XdlopsGemmDerivedParams", [RockTuningParamAttrInterface, RockAccelTuningParamAttrInterface]> {
+  let mnemonic = "xdlops_gemm_derived_params";
+  let description = [{
+    The tuning parameters for an xdlops-based matrix multiplication.
+  }];
```
krzysz00, line 291: Question: do you have thoughts on whether making the derived tuning parameters attribute take the underived one as an argument? I don't have strong thoughts on that, just wanted to ask.

krzysz00: ... never mind, the constructor approach is actually pretty nice

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
```diff
@@ -137,23 +143,23 @@ struct InitParamsAccel : InitParams, Serializable<InitParamsAccel> {
...
+        gemmMPerWave(attr.getMPerWave()),
+        gemmNPerWaveOrMnPerXdl(attr.getNPerWave()), gemmKPack(attr.getKpack()),
         gemmAThreadCopyMoreGemmK(attr.getForceUnroll()),
         gemmBThreadCopyMoreGemmKPack(false){};
 
   int64_t getKPack() { return gemmKPack; }
 
   int64_t gemmMPerWave;
-  int64_t gemmNPerWave;
+  int64_t gemmNPerWaveOrMnPerXdl;
```
krzysz00, line 165: ... why's this being named like that?

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
```diff
@@ -334,28 +341,19 @@ class PopulateParamsAccel : public BasePopulateParams<InitParamsAccel> {
 
   // Note that this is a method on the general class because the distinguishing
   // of MFMA and WMMA paths is handled under the hood in populateDerived().
-  LogicalResult paramsProbablyValid(const PopulateParamsInfo &info,
+  LogicalResult paramsProbablyValid(OpBuilder b, const PopulateParamsInfo &info,
```
krzysz00: `OpBuilder& b`

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -805,9 +827,11 @@ LogicalResult PopulateParamsWmma::isValidBlockwiseGemm(
                        int validMPerWave, validNPerWave, validKPerWave;
                        std::tie(validMPerWave, validNPerWave, validKPerWave) =
                            it;
-                       return (param.gemmMPerWave == validMPerWave) &&
-                              (param.gemmNPerWave == validNPerWave) &&
-                              (param.gemmKPerBlock % validKPerWave == 0);
+                       return (param.getMPerWave() == validMPerWave) &&
+                              (param.getNPerWave() == validNPerWave) &&
+                              (param.getKpackPerBlock() * param.getKpack() %
```
krzysz00: Note: I *think* `param.gennKPerBlock` was actually `kpackPerBlock` ... unless it wasn't?

mlir/utils/performance/sdxl-gemm-configs
```diff
@@ -0,0 +1,16 @@
...
+-t f16 -out_datatype f16 -transA true -transB false -g 1 -m 8192 -n 640 -k 640
+-t f16 -out_datatype f16 -transA false -transB false -g 1 -m 8192 -n 1920 -k 640
+-t f16 -out_datatype f16 -transA false -transB false -g 1 -m 8192 -n 640 -k 640
+-t f16 -out_datatype f16 -transA false -transB false -g 1 -m 8192 -n 5120 -k 640
+-t f16 -out_datatype f16 -transA false -transB false -g 1 -m 8192 -n 640 -k 2560
+-t f16 -out_datatype f16 -transA true -transB false -g 1 -m 2048 -n 1280 -k 1280
+-t f16 -out_datatype f16 -transA false -transB false -g 1 -m 2048 -n 3840 -k 1280
+-t f16 -out_datatype f16 -transA false -transB false -g 1 -m 2048 -n 1280 -k 1280
+-t f16 -out_datatype f16 -transA false -transB false -g 1 -m 2048 -n 10240 -k 1280
+-t f16 -out_datatype f16 -transA false -transB false -g 1 -m 2048 -n 1280 -k 5120
```
krzysz00: Nit: missing newline

krzysz00 (commented):
I'm not sure this is super controversial, I've got questions more than anything.

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
manupak: because of wmma; and Im not changing how thats tuned here. see L155.

sjw36:
So instead of more waves, we do more repeats and larger tile size?

Does this also help for i8/f8 where we have low LDS usage?

manupak:
Tile sizes are not affected. There are a function of dPerBlock, kPerBlock which are unaffected. Hence lds is unaffected.

manupak:
> Does this also help for i8/f8 where we have low LDS usage?

https://github.com/ROCm/rocMLIR/files/14559558/develop_vs_16x16_gfx90a.zip -- shows everything for gfx90a. MI300X is still running.

sjw36:
Wow, seems counter-intuitive but great find!

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
krzysz00: ... Ah

manupak:
> Wow, seems counter-intuitive but great find!

Hmmm... curious to know why you think its counter-intuitive?

Here is the thing -- we first decide on dPerBlock x kPerBlock. This defines the tile size we'll be fetching from global for both A & B tensors. Then we swizzle via LDS -- therefore requires dPerBlock x kPerBlock in LDS -- to MFMA layout. 

What we had been doing was we were launching more waves per workgroup if we had to select a smaller MFMA. As a side effect, we had to deal with relatively larger number of waves that has to collaborate to bring the same sized tile (defined by  dPerBlock x kPerBlock) and do the swizzling. Therefore, all the worgroup-level (workgroup & lds) barriers we ll need to keep this waves in sync for inorder to kernel to proceed. Hence, it turns out we were picking larger dPerWave to achieve that to balance to keep the tile sufficiently sized vs #waves that needed to be sync'd. IMO, which imposed artificial restriction to use smaller MFMAs.

However, if we have a smaller number of waves that does the same sized tiled (defined by  dPerBlock x kPerBlock), then we ll still enjoy the right amount of waves per WG but explore using variety of MFMAs. This would allow the tuner to select between whether to increase the gridsize or allow more repeats (if numCUs are not sufficient) -- where the former would allow more inter-wave pipelining whilst the latter allow intra-wave pipelining -- without incurring sync penalty.

giuseros:
Great work! I think at this point you could get also rid of :
https://github.com/ROCm/rocMLIR/blob/8a6e923a7ca0fae765678fec54ce0aea63dbf733/mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp#L439

This turned out important for wmma performance (where we use 16x16x16 as well). In this case, the problem is minor, because kbase=16/2=8 which is exactly the max vector length, but you neve know

mlir/utils/performance/sdxl-gemm-configs
manupak: done

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
manupak: yes you are right. fixed.

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
manupak: done

manupak:
> Great work! I think at this point you could get also rid of :
> 
> https://github.com/ROCm/rocMLIR/blob/8a6e923a7ca0fae765678fec54ce0aea63dbf733/mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp#L439
> 
> This turned out important for wmma performance (where we use 16x16x16 as well). In this case, the problem is minor, because kbase=16/2=8 which is exactly the max vector length, but you neve know

Good to know! can you create a seperate ticket please ? I d like that in a seperate PR with seperate measurements.

manupak:
ok, all tests are passing now!

manupak:
I d appreciate a look here..

sjw36:
@krzysz00 ping!

krzysz00 (approved):
Approved - this looks reasonable to me overall

manupak:
Thanks! it really helps getting this in to allow SDXL work to use our develop instead of this branch.

PR #1455, opened 2024-03-11
Fix the argument types of conv2d in a test case

yiqian1:


krzysz00 (approved):

pcf000:
Could swear I got them all, but clearly not.

pcf000 (approved):

PR #1457, opened 2024-03-15
Add the option to reverse the grid

manupak:
This commit adds the option to reverse the grid
of the all the kernels we produce by intersecting
how we lower to rock.workgroup id to gpu.blockID.

~~The reversal of K loop would be a seperate PR.~~
Its done here now.

closes : https://github.com/ROCm/rocMLIR-internal/issues/1403


krzysz00 (commented):
I'm not sure this is how I'd do this - I'd instead move the reversal into the grid tiling layout in gridwise gemm

manupak:
Why ? 

It feels like that d create a lot of room for error.

manupak:
Because if one is reversing the grid it has to be done in every place and the indirection present in rock.workgroup id --> gpu.block_id provides that nice abstraction for this to be done in a kernel agnostic manner.

krzysz00:
... but this isn't a kernel-agnostic transformation. This is specifically, "do the implicit gemm with the matrices swapped right-to-left (within their tiles)". 

I'd be worried about workgroup IDs getting changed on me in an easy-to-miss flag, especially when there're things like cooperative groups going around in the future.

Right, to be more specific about how this can shoot us in the foot eventually. Suppose we've got the ability to synchronize with or otherwise talk to some set of neighboring workgroups - for simplicity, WGs 0-3 can talk, 4-7 can talk, and so on.

If we just define `workgroup_id = [num_workgroups] - workgroup_id`, then the personholding the id "3" is actually workgroup number 4, and depending on how annoying the notional API is, having a value for `workgroup_id % [special constant]` that disagrees with the truth could be a problem.

That being said, this might be fine for now ... I'm just suspicious of it

mlir/include/mlir/Dialect/Rock/IR/Rock.h
```diff
@@ -58,6 +58,9 @@ namespace rock {
 ArrayAttr noTransformsArray(Builder &b, size_t n);
 
 ArrayAttr getIndexArrayAttr(Builder &b, ArrayRef<int64_t> values);
+
+const char reverseGridAttrName[] = "reverse_grid";
```
krzysz00: 1. Upstream merge'll make this nicer
2. Put this in the tablegen file with all the other `StringLiteral`s

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
```diff
@@ -119,6 +121,38 @@ struct MIIdRewritePattern : public OpRewritePattern<Tmi> {
...
+struct WorkgroupIdRewritePattern
+    : public OpRewritePattern<rock::WorkgroupIdOp> {
+  using OpRewritePattern<rock::WorkgroupIdOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(rock::WorkgroupIdOp op,
+                                PatternRewriter &b) const override {
+    Location loc = op.getLoc();
+    bool isReverseGrid = false;
+    auto maybeIsReverseGrid = rock::getReverseGrid(op);
+    if (succeeded(maybeIsReverseGrid)) {
```
krzysz00, line 133: This could just be an assignment?

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
```diff
@@ -119,6 +121,38 @@ struct MIIdRewritePattern : public OpRewritePattern<Tmi> {
...
+      FailureOr<IntegerAttr> maybeGridSize = rock::getGridSize(op);
+      if (failed(maybeGridSize)) {
+        return op->emitError("grid_size should ve been set by now.\n");
+      }
+      int64_t gridSize = maybeGridSize.value().getValue().getSExtValue();
+      AffineMap reverseMap = rock::getIdxReversalMap(b);
+      Value gridSizeVal = b.createOrFold<arith::ConstantIndexOp>(loc, gridSize);
+      Value blockIdVal = b.create<gpu::BlockIdOp>(loc, gpu::Dimension::x);
+      b.replaceOpWithNewOp<affine::AffineApplyOp>(
+          op, b.getIndexType(), reverseMap,
```
krzysz00, line 143: ... why the affine apply? Just `arith.sub` will be fine and won't complicate the lowering

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2578,8 +2595,22 @@ struct GridwiseGemmAccelRewritePattern
...
+        isReverseGrid = true;
+      }
+      // Purpose of reversing the grid is to exploit
+      // (if any) temporal locality between producers
+      // and consumers of data between kernels.
+      // Towards that goal, the kLoop has to be reversed
+      // to use latest producer.
+      if (isReverseGrid) {
+        AffineMap reverseMap = rock::getIdxReversalMap(b);
+        iv = b.createOrFold<affine::AffineApplyOp>(loc, reverseMap,
```
krzysz00, line 2610: ... why this? Why not just change the loop to be by -1 and switch the lower and upper bounds?

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
```diff
@@ -668,10 +679,25 @@ FailureOr<int64_t> mlir::rock::getNumCU(Operation *op) {
...
+  if (numCU.getValue().getSExtValue() < archInfo.minNumCU) {
     return op->emitError() << "num_cu=" << numCU
                            << " cannot be lower than arch minNumCU="
                            << archInfo.minNumCU;
   }
-  return numCU.getSInt();
+  return numCU.getValue().getSExtValue();
+}
+
+FailureOr<UnitAttr> mlir::rock::getReverseGrid(Operation *op) {
```
krzysz00, line 690: `FailureOr<UnitAttr>` could just be written `LogicalResult` but I see the genericness point.

krzysz00 (commented):
General notes aside, pulling in `AffineMap` for a subtraction is breaking how our code is generally structured and feels overcomplicated.

(But yeah, doing the grid reversal by changing up `workgroup_id`'s definition is probably fine for all practical purposes, so I'm not going to argue too hard about that)

manupak:
> pulling in AffineMap for a subtraction is breaking how our code is generally structured and feels overcomplicated.

Im fine with either; I actually started with a substraction then thought AffineMap is concise and clearer in the IR to read that there is a reversal going on seeing affine_map<-d0 + S0 -1> rather than observing a constantIdx op substraction.
Having said my reasons, I could be open about going back to a subtraction (which is what I started with). 
(Though I dont see how a index affine_apply affects our code structure)

manupak:
> ... but this isn't a kernel-agnostic transformation. This is specifically, "do the implicit gemm with the matrices swapped right-to-left (within their tiles)".
> 
> I'd be worried about workgroup IDs getting changed on me in an easy-to-miss flag, especially when there're things like cooperative groups going around in the future.
> 
> Right, to be more specific about how this can shoot us in the foot eventually. Suppose we've got the ability to synchronize with or otherwise talk to some set of neighboring workgroups - for simplicity, WGs 0-3 can talk, 4-7 can talk, and so on.
> 
> If we just define `workgroup_id = [num_workgroups] - workgroup_id`, then the personholding the id "3" is actually workgroup number 4, and depending on how annoying the notional API is, having a value for `workgroup_id % [special constant]` that disagrees with the truth could be a problem.
> 
> That being said, this might be fine for now ... I'm just suspicious of it

Yes you are right.. in that scenario, I would even doubt whether we could do any reversal (at all!) when the lowered code relies on workgroup interaction. This whole thing is based on the fact that workgroup ordering is independent.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
manupak: because : 
https://github.com/ROCm/rocMLIR/blob/598f7d1126088e8d3673780aa729d50a59f39974/external/llvm-project/mlir/lib/Dialect/Affine/IR/AffineOps.cpp#L1898

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
manupak: I d like to keep the pattern if thats alright.

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
manupak: done

mlir/test/Conversion/RockToGPU/misc_ops.mlir
```diff
@@ -30,3 +31,22 @@ module {
...
+
+// -----
+
+// CHECK: #[[MAP:.+]] = affine_map<(d0)[s0] -> (-d0 + s0 - 1)> 
+module {
+  func.func @misckernel(%arg0: memref<?xf32>, %arg1: memref<?xf32>)
+      attributes {kernel = 0 : i32, block_size = 64 : i32, grid_size = 900 : i32, reverse_grid} {
+    // CHECK: %[[BLOCKID:.+]] = gpu.block_id x
+    %bid = rock.workgroup_id : index
+    // CHECK: %[[REV_BLOCKID:.+]] = affine.apply #[[MAP]](%[[BLOCKID]])[%c900]
```
manupak, line 43: I (personally) found this clearer to see in the IR that is being reversed rather than a substraction.

manupak (commented):
Thanks @krzysz00 for reviews!

I have addressed them all except affine_apply comment.
I still struggle to understand why its breaks code structure -- to me its isomorphic to use the subtraction vs affine map apply -- and I found its clearer in the IR that there is reversal happening via the usage of the map.

As i said I can change this (back) if there is a solid reason and I could be convinced to do so.

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
manupak: See the comment below.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
krzysz00: Ah

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
krzysz00: Yeah, it's fine

mlir/test/Conversion/RockToGPU/misc_ops.mlir
krzysz00: But by the time we've lowered to `gpu` we're already in a low-level state where we have where we've even got `cf` instead of `scf` - that is, things like affine apply lowering have already happened.

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
krzysz00: `expandAffineMap()` type calls might be the right move here?

krzysz00 (commented):
Minor arguing about how the IR should look in the GPU dialect aside, we might want unit tests for kernel reversal in gridwise-gemm-to-blockwise for code coverage reasons.

But I don't have fundamental concerns here

pfultz2:
> Suppose we've got the ability to synchronize with or otherwise talk to some set of neighboring workgroups - for simplicity, WGs 0-3 can talk, 4-7 can talk, and so on.

I dont see how this is a problem. WGs 0-3 will become WGs 7-4(which can talk) and WGs 4-7 will become WGs 3-0(which can talk).

mlir/test/Conversion/RockToGPU/misc_ops.mlir
manupak: > things like affine apply lowering have already happened.

not really, see 
https://github.com/ROCm/rocMLIR/blob/develop/mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp#L208-L210

manupak:
> Minor arguing about how the IR should look in the GPU dialect aside, we might want unit tests for kernel reversal in gridwise-gemm-to-blockwise for code coverage reasons.
> 
> But I don't have fundamental concerns here

Thanks! I ve added all unit tests for gridwise lowering now..

Re:affine_apply, I think as long as memref s are around affine ops are very much legal. hence we have lower affine ops pass in the backend pipeline. expandAffineMap is actually what the lowering of affine_apply calls. I mean its trivial effort to change to that (now you ve suggested it)... however, I think it still not bring any benefit IMO.

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
krzysz00: My vague sense that we should expand the map out in place instead of having `affine.apply` sitting around in code that's set to get translated into LLVM IR soon aside, I don't have any real objections here.

krzysz00 (approved):
Overall seems fine to me.

kuhar:
@manupak BTW, in IREE we found that transposing the grid gives good performance on gfx942. Not sure how applicable this is to the code rocMLIR generates, though.

manupak:
Thanks @kuhar for taking a look.

Can you define what is 'transposing' the grid ?

krzysz00:
@kuhar We've got a somewhat strange 2-D tiling we took from the Triton people for our grid

kuhar:
@manupak 
> Can you define what is 'transposing' the grid ?

Instead of going left -> right and then down, you go top -> down ant then right. The implementation is here: https://github.com/openxla/iree/blob/089027d654fb98e8aa2dbf45519b31f37133e5b5/compiler/src/iree/compiler/Codegen/Common/GPU/WorkgroupReordering.cpp#L69-L80 . I tried reversing the order like in this PR (IIUC) and also some common patterns like Boustrophedon and the Z curve but they did not perform better, at least on UNET.

@krzysz00
> @kuhar We've got a somewhat strange 2-D tiling we took from the Triton people for our grid

Yeah, I'd expect this to depend on your codegen strategy. So just a datapoint from us.

manupak:
We actually do : 
image
(From : https://triton-lang.org/main/getting-started/tutorials/03-matrix-multiplication.html).

Its good to know that just reversing did not help :).
Just to confirm, when you say it did not help, are you saying alternating the order between kernels right ? 

And are you saying toggling 'tranposing' between kernels helped ?

manupak:
> The implementation is here: https://github.com/openxla/iree/blob/089027d654fb98e8aa2dbf45519b31f37133e5b5/compiler/src/iree/compiler/Codegen/Common/GPU/WorkgroupReordering.cpp#L69-L80 

Here it sounds like a static layout you d use through out all kernels.

kuhar:
I used to have some filtering to only apply this on matmuls/convs/etc, but at the end of the day I found that applying it everywhere worked best in practice.

> Just to confirm, when you say it did not help, are you saying alternating the order between kernels right ?

I'm saying that I tried other workgroup orders and they were either slower than the default or not better than the simple transpose.

kuhar:
> Just to confirm, when you say it did not help, are you saying alternating the order between kernels right ?

@krzysz00  That I didn't try, I don't quite know how to thread that through our pipeline. IE there are two angles to this: improve cache reuse within workgroups in a kernel or across kernels, and I'm only talking about our experience with the former.

manupak:
Right... our focus was was the latter.

PR #1458, opened 2024-03-15
Fix a bug when using atomicFadd in the fp16 scalar case

giuseros:
So there was a simple bug to fix, but the main issue was that we were not always selecting the atomicFadd intrinsic when doing an atomic Fp16/Fp32 addition (that's why I didn't catch the bug before). 

I enabled the buffer operation (i.e., the intrinsics) for all the atomics Fp16/Fp32 cases. @krzysz00 can you take a look? Is there a reason why we were not always enabling the intrinsic?

FYI, using atomics I always see accuracy degradation. 

```
./bin/rocmlir-gen --arch gfx90a -pv   --operation gemm -t f16 -g 1 -m 256 -k 512 -n 256  --perf_config 64,64,16,32,32,4,1,1,1 -split-k 2 | ./bin/rocmlir-driver -c | ../mlir/utils/widgets/rocm-run 
Number of elements: 65536
maxAbsDiff info: maxAbsDiff = 0.250000 (valNum = 309.50000, gpuNum = 309.25000), average absDiff = 5.6e-02
maxRelDiff info: maxRelDiff = 8.1e-04 (valNum = 309.5000000000, gpuNum = 309.2500000000), average relDiff = 1.8e-04
RMS = 3.8e-04
Histogram of relDiff: 
        relDiff = 0     : 50915/65536 (77.690125%)
     0 < relDiff < 1e-06: 0/65536 (0.000000%)
1e-06 < relDiff <= 1e-05: 0/65536 (0.000000%)
1e-05 < relDiff <= 1e-04: 0/65536 (0.000000%)
1e-04 < relDiff <= 1e-03: 14621/65536 (22.309875%)
1e-03 < relDiff <= 1e-02: 0/65536 (0.000000%)
1e-02 < relDiff <= 1e-01: 0/65536 (0.000000%)
1e-01 < relDiff <= 1e+00: 0/65536 (0.000000%)
1e+00 < relDiff < inf   : 0/65536 (0.000000%)
        relDiff = inf   : 0/65536 (0.000000%)
[0 1 1]
```
This happens with the CAS loop and with the atomic intrinsics.

### DIFF
--- a/external/llvm-project/mlir/lib/Target/LLVMIR/Dialect/ROCDL/ROCDLToLLVMIRTranslation.cpp
+++ b/external/llvm-project/mlir/lib/Target/LLVMIR/Dialect/ROCDL/ROCDLToLLVMIRTranslation.cpp
@@ -150,6 +150,15 @@ public:
       attrValueStream << value.getInt();
       llvmFunc->addFnAttr("amdgpu-waves-per-eu", llvmAttrValue);
     }
+    if (ROCDL::ROCDLDialect::getUnsafeFpAtomics() == attribute.getName()) {
+      auto func = dyn_cast<LLVM::LLVMFuncOp>(op);
+      llvm::Function *llvmFunc =
+          moduleTranslation.lookupFunction(func.getName());
+      llvm::SmallString<8> llvmAttrValue;
+      llvm::raw_svector_ostream attrValueStream(llvmAttrValue);
+      attrValueStream << "true";
+      llvmFunc->addFnAttr("amdgpu-unsafe-fp-atomics", llvmAttrValue);
+    }
 
     // Set reqd_work_group_size metadata
     if (ROCDL::ROCDLDialect::getReqdWorkGroupSizeAttrName() ==
--- a/external/llvm-project/mlir/lib/Target/LLVMIR/Dialect/ROCDL/ROCDLToLLVMIRTranslation.cpp
+++ b/external/llvm-project/mlir/lib/Target/LLVMIR/Dialect/ROCDL/ROCDLToLLVMIRTranslation.cpp
@@ -152,11 +152,15 @@ public:
     }
     if (ROCDL::ROCDLDialect::getUnsafeFpAtomics() == attribute.getName()) {
       auto func = dyn_cast<LLVM::LLVMFuncOp>(op);
+      if (!func)
+        return failure();
+      auto value = attribute.getValue().dyn_cast<BoolAttr>();
+
       llvm::Function *llvmFunc =
           moduleTranslation.lookupFunction(func.getName());
       llvm::SmallString<8> llvmAttrValue;
       llvm::raw_svector_ostream attrValueStream(llvmAttrValue);
-      attrValueStream << "true";
+      attrValueStream << value.getValue();
       llvmFunc->addFnAttr("amdgpu-unsafe-fp-atomics", llvmAttrValue);
     }
 
--- a/mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
+++ b/mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
@@ -223,6 +223,18 @@ void LowerRockOpsToGPUPass::runOnOperation() {
     b.setInsertionPointToEnd(&gpuFuncEntry);
     b.create<cf::BranchOp>(loc, clonedFuncEntry);
 
+    // Detect if we are using `atomic_rmw fadd` and ask LLVM to
+    // use intrinsics instead of CAS loops
+    funcBody.walk([&](memref::AtomicRMWOp op) -> WalkResult {
+      auto type = op.getMemRefType().getElementType();
+      if (op.getKind() == arith::AtomicRMWKind::addf &&
+          (type.isF32() || type.isF16())) {
+        gpuFunc->setAttr("rocdl.unsafe_fp_atomics", b.getUnitAttr());
+        return WalkResult::interrupt();
+      }
+      return WalkResult::advance();
+    });
+
     // Clone in global constants
     llvm::SmallDenseMap<SymbolRefAttr, FlatSymbolRefAttr> clonedConsts;
     WalkResult result = funcBody.walk([&](memref::GetGlobalOp op)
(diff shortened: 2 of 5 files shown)


giuseros:
@ravil-mobile this should get you going with split-k tests. Again, you will have to adjust the accuracy, because atomics are less accurate (so split-k will be less accurate). Maybe another reason to do locks ;) ?

sjw36:
An atomic reduction (accumulation in fp16) will generally result in less accuracy vs a full mfma reduction (accumulation in at least f32). I think the rel-diff is acceptable.

sjw36:
What is the accuracy degradation on f32?

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
```diff
@@ -1180,13 +1180,14 @@ struct GlobalStoreRewritePattern : public OpRewritePattern<GlobalStoreOp> {
...
     // dymanic shape case, we'll already be in the i64 case, so we don't set
     // this.
-    bool useBufferOps = !hasI64Idx && (numBytes.trunc(32).isNegative() ||
-                                       emitOobChecks || op.getCanStoreOffEnd());
-    bool useBufferOobChecks =
-        useBufferOps && (emitOobChecks || op.getCanStoreOffEnd());
-
     StoreMethod memoryOp = op.getStoreMethod();
     bool isAtomic = memoryOp != StoreMethod::Set;
+    bool isAtomicFadd = memoryOp == StoreMethod::AtomicAdd && elemTy.isF16();
```
sjw36: should this also allow f32?

sjw36 (approved):
Looks good overall, just 1 question.

ravil-mobile:
Hi @giuseros, I tried you patch and it works.

giuseros:
Hi @sjw36 , 
You are right, in the parallel data path we always accumulate in fp32. Indeed fp32 atomics work ok.

krzysz00 (changes requested):
1. We should be using global operations instead of buffer operations when possible, since they don't incur bounds checks
2. If that atomic is lowering to a CAS loop, try setting `"amdgpu-unsafe-fp-atomics"="true"` on the way to LLVM.
3. (If it isn't, we might as well always go to a buffer)

giuseros:
> 1. We should be using global operations instead of buffer operations when possible, since they don't incur bounds checks
> 2. If that atomic is lowering to a CAS loop, try setting `"amdgpu-unsafe-fp-atomics"="true"` on the way to LLVM.
> 3. (If it isn't, we might as well always go to a buffer)

Alright, I see what you mean. However, I don't see any difference performance-wise between global and buffer atomics. That said, let's agree on the logic. 

For fp16, I always need to write packed vectors (using the intrinsic). Does `atomic_rmw` supports vectors? If yes, the logic can be `use atomic_rmw everytime you don't need buffer checks, use buffer ops otherwise (+enable unsafe-fp-atomics).` If  `atomic_rmw` does not support vectors, we need to use that logic only for fp32, and let fp16 always use the buffer intrinsics.

krzysz00:
@giuseros For the vector case, you'll want `amdgcn_global_atomic_fadd` - it's an intrinsic you'll need to add `amdgpu` wrappers for, probably.

krzysz00:
Or least I'd expect that to be what you want there.

giuseros:
So, ok, adding `amdgcn_global_atomic_fadd` seems a bit too much for this PR, that wanted mostly to address a bug. We need to pipe it through `AMDGPUDialect` and then add it in ROCDL. I agree that this is what we should be doing, and I left a TODO in the code. Given that performance are very similar for now I would use buffer atomics whenever there is a fp16 (packed or unpacked).

ravil-mobile:
> So, ok, adding `amdgcn_global_atomic_fadd` seems a bit too much for this PR, that wanted mostly to address a bug. We need to pipe it through `AMDGPUDialect` and then add it in ROCDL. I agree that this is what we should be doing, and I left a TODO in the code. Given that performance are very similar for now I would use buffer atomics whenever there is a fp16 (packed or unpacked).

Maybe you can also immediately fire a ticket which with the TODO description?

external/llvm-project/mlir/lib/Target/LLVMIR/Dialect/ROCDL/ROCDLToLLVMIRTranslation.cpp
```diff
@@ -150,6 +150,15 @@ class ROCDLDialectLLVMIRTranslationInterface
       attrValueStream << value.getInt();
       llvmFunc->addFnAttr("amdgpu-waves-per-eu", llvmAttrValue);
     }
+    if (ROCDL::ROCDLDialect::getUnsafeFpAtomics() == attribute.getName()) {
+      auto func = dyn_cast<LLVM::LLVMFuncOp>(op);
+      llvm::Function *llvmFunc =
+          moduleTranslation.lookupFunction(func.getName());
+      llvm::SmallString<8> llvmAttrValue;
+      llvm::raw_svector_ostream attrValueStream(llvmAttrValue);
+      attrValueStream << "true";
```
krzysz00: Note: Let's make the attribute an actual boolean

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
```diff
@@ -223,6 +223,18 @@ void LowerRockOpsToGPUPass::runOnOperation() {
     b.setInsertionPointToEnd(&gpuFuncEntry);
     b.create<cf::BranchOp>(loc, clonedFuncEntry);
 
+    // Detect if we are using `atomic_rmw fadd` and ask LLVM to
+    // use intrinsics instead of CAS loops
+    funcBody.walk([&](memref::AtomicRMWOp op) -> WalkResult {
+      auto type = op.getMemRefType().getElementType();
+      if (op.getKind() == arith::AtomicRMWKind::addf &&
+          (type.isF32() || type.isF16())) {
+        gpuFunc->setAttr("rocdl.unsafe_fp_atomics", b.getUnitAttr());
```
krzysz00: I'd just set it unconditionally

krzysz00 (commented):
Agreed that we can do the global packed atomic in the future since it's not that big of a deal. It looks like this still needs flattening?

giuseros:
> Maybe you can also immediately fire a ticket which with the TODO description?
Done: https://github.com/ROCm/rocMLIR-internal/issues/1429

krzysz00 (approved):
Approved modulo maybe we should have some tests for the atomic add logic

giuseros:
> Approved modulo maybe we should have some tests for the atomic add logic
I added additional two tests to stress the flattening logic

PR #1459, opened 2024-03-15
Fix AlignTiling to reuse buffers when loading fusion arguments

yiqian1:
Fixes https://github.com/ROCm/rocMLIR-internal/issues/1419

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
@@ -864,13 +864,13 @@ LAGenericRewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
       b.setInsertionPoint(laGeneric);
     }
     Value newOutput = tileReadOp.getDest();
+    // Prevent SSA weirdness from register allocations introduced too late.
+    // Do this before calling addRegisterReadsForTiledOutput() as
+    // addRegisterReadsForTiledOutput() may move laGeneric.
+    b.moveBeforeIfNeeded(newOutput.getDefiningOp(), laGeneric);
     SmallVector<Value> newInputs;
     addRegisterReadsForTiledOutput(b, laGeneric, tileReadOp,
                                    globalCoordsToGenericViews, newInputs);
-
-    // Prevent SSA weirdness from register allocations introduced too late.
-    if (newOutput.getDefiningOp()->getBlock() == laGeneric.getBlock())
-      b.moveBeforeIfNeeded(newOutput.getDefiningOp(), laGeneric);
     reconfigureLAGeneric(b, laGeneric, newInputs, newOutput);
     b.eraseOp(tileReadOp);
     return success();
--- a/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
@@ -622,11 +622,11 @@ makeExtraInputTile(LinalgAlignRewriter &b, TiledOp tiledOp, Value src,
   LLVM_DEBUG(llvm::dbgs() << "Src type: " << src.getType()
                           << " tile type: " << tile.getType() << "\n");
 
-  // Reset the insertion point if laGeneric and tiledOp are in different blocks.
-  // Insert new transform operations and threadwiseReadIntoOp where
-  // linalg.generic is if linalg.generic reads from the output of tiledOp.
-  // Otherwise, if tiledOp reads from the output of linalg.generic, insert
-  // transformation and threadwiseReadIntoOp where tiledOp is.
+  // Reset the insertion point if laGeneric and tiledOp are in different blocks
+  // and tiledOp reads from the output of linalg.generic. In such cases, we
+  // should insert new transform operations and threadwiseReadIntoOp inside the
+  // block of tiledOp, as their arguments may depend on that control flow. Later
+  // we will also move linalg.generic into the block.
   if (laGeneric->getBlock() != tiledOp->getBlock() &&
       laGeneric.getOperation() != sink)
     b.setInsertionPoint(sink);
--- a/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
@@ -219,8 +219,7 @@ void LinalgAlignRewriter::moveBeforeIfNeeded(Operation *toMove,
   constexpr llvm::StringLiteral beforePrefix("   before  : ");
   logOpActivity(movePrefix, toMove);
   logOpActivity(beforePrefix, latestOp);
-  if (latestOp->getBlock() != toMove->getBlock() ||
-      latestOp->isBeforeInBlock(toMove))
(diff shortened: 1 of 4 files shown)


mlir/test/fusion/rock-gemm-align-tiling-multiple-args.mlir
```diff
@@ -0,0 +1,14 @@
+// RUN: rocmlir-driver --rock-gridwise-gemm-to-blockwise --rock-blockwise-gemm-to-threadwise --rock-linalg-align --verify-passes %s | rocmlir-opt
+
```
krzysz00, line 2: Insufficient testing

1. Please do some `FileCheck`-ing to ensure the output is what you expect
2. More generally, `--rock-linalg-align` needs unit tests: write up some example input fusions (with just `threadwise_read_into`s - that is, leave the `rock.threadwise_read_into`s but strip out the rest of the gemm

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -591,31 +618,19 @@ makeExtraInputTile(LinalgAlignRewriter &b, TiledOp tiledOp, Value src,
   LLVM_DEBUG(llvm::dbgs() << "Src type: " << src.getType()
                           << " tile type: " << tile.getType() << "\n");
 
+  // Reset the insertion point if laGeneric and tiledOp are in different blocks.
+  // Insert new transform operations and threadwiseReadIntoOp where
+  // linalg.generic is if linalg.generic reads from the output of tiledOp.
+  // Otherwise, if tiledOp reads from the output of linalg.generic, insert
+  // transformation and threadwiseReadIntoOp where tiledOp is.
```
krzysz00: I'm not sure `sink` is a good variable name here.

Also, this comment could use some motivation.

krzysz00 (changes requested):
As an overall note, one test I'd like to see is something like
```
%alloc0 = memref.alloc()
la.generic ins (%arg0, %arg1) outs(%alloc0) { ... }
%alloc1 = memref.alloc()
la.generic ins(%alloc0, %arg2) outs(%alloc1) { ... }
threadwise_read_into %alloc1[0, ...], ...
for %arg3 from 1 to 4 {
  threadwise_read_into %alloc1[%arg3, ...], ...
}
```
to make sure that acts how you expect it to

---

I'm also not sure that this is the change we want, since this creates a duplicate allocation between the copy of the generic in the prologue of the function and one for the one in the loop body. I'm not sure there's a good way around that, though, since identifying mergable buffers might be a somewhat expensive check.

mlir/test/fusion/linalg-generic-reuse-buffers.mlir
```diff
@@ -0,0 +1,60 @@
+// RUN: rocmlir-driver --rock-linalg-align %s |./external/llvm-project/llvm/bin/FileCheck %s
+#map1 = affine_map<(d0, d1, d2) -> (d0, d1, d2)>
+#map2 = affine_map<(d0, d1, d2, d3, d4, d5, d6, d7) -> (d1, d0 * 16 + d5 + d7, (d2 * 16 + d4) * 4 + d6)>
+#map3 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d1, d2, d3, d4 floordiv 16, d4 mod 16, d5, 0)>
+
+// CHECK: [[MAP:.*]] = #rock.transform_map<#map by [<PassThrough ["g_block"] at [1] -> ["g"] at [0]>, <Unmerge{4, 16, 1} ["k_loop", "k_thread", "k_iter"] at [0, 5, 7] -> ["k"] at [1]>, <Unmerge{1, 16, 4} ["m_block", "m_thread", "m_iter"] at [2, 4, 6] -> ["m"] at [2]>, <AddDim{1} ["n_block"] at [3] -> [] at []>] bounds = [4, 64, 1, 1, 16, 16, 4, 1] -> [64, 64, 64]>
```
krzysz00: You'll probably want to `#{{.*}}` instead of `#map` or `#map1` here. Also `CHECK-DAG` on these

mlir/test/fusion/linalg-generic-reuse-buffers.mlir
```diff
@@ -0,0 +1,60 @@
...
+    // CHECK: [[arg0_alloc:%.+]] = rock.alloc() : memref<4xf16
+    // CHECK: [[arg3_alloc:%.+]] = rock.alloc() : memref<4xf32
+    affine.for %arg4 = 1 to 4 {
+      // CHECK: [[arg0_tf:%.+]] = rock.transform %arg0 by [[MAP]]
+      // CHECK-NEXT: [[arg0_v:%.+]] = rock.transform [[arg0_tf]] by [[MAP1]]
+      // CHECK-NEXT: rock.threadwise_read_into {{.*}} []([[arg0_v]]) [%arg4, {{.*}}] -> [[arg0_alloc]]
+      // CHECK: [[arg3_tf:%.+]] = rock.transform %arg3 by [[MAP]]
+      // CHECK-NEXT: [[arg3_v:%.+]] = rock.transform [[arg3_tf]] by [[MAP1]]
+      // CHECK-NEXT: rock.threadwise_read_into {{.*}} []([[arg3_v]]) [%arg4, {{.*}}] -> [[arg3_alloc]]
+      // CHECK: linalg.generic {indexing_maps = [#map2, #map2, #map2], iterator_types = ["parallel"]} ins(%12, %13 : memref<4xf16, #gpu.address_space<private>>, memref<4xf32, #gpu.address_space<private>>) outs(%11 : memref<4xf32, #gpu.address_space<private>>) {
```
krzysz00: Ok, so, to ask very annoying questions: ... how is it your code ensures that the same temporary allocation gets used for both the generic in the loop and the generic outside of it? I didn't see anything in your changes that would've caused the two `rock::AllocOp`s to merge together.

One possibility I'm concerned by is that you're not calling `clone` early enough, and thus assuming that all the users of a given `linalg.generic` are tiled the same.

For example, let's do another test where we add a second `threadwise_read_into` that has a different tile size than the one outside the loop.

yiqian1: > Ok, so, to ask very annoying questions: ... how is it your code ensures that the same temporary allocation gets used for both the generic in the loop and the generic outside of it? I didn't see anything in your changes that would've caused the two `rock::AllocOp`s to merge together.
> 
I didn't merge them. The two threadwise_read_into in the original IR happen to write into the same allocation, so the two generic use the same allocation as well. If the original threadwise_read_into ops write into different allocations, the generic ops would use different temporary allocations, even if they have the same tile size.

krzysz00 (commented):
1. Might be good to have both the simple case and the complicated case as separate tests - more tests aren't a bad thing
1b. Those if statements you had to add when you'd modified the test show why I wanted that test, I claim
2. Overall, I think we're fine here
3. I would like to see a separate test (it can be in the same file, but a different function) with sequences of linalg.generic operations, given how much headache they caused us with output fusion

krzysz00:
... For some reason, my comment (which is already outdated) elaborating on the test I wanted showed up twice, please ignore it

krzysz00 (approved):
I'd say we're approved, (even though I still think the variable name `sink` could use some more thought)

PR #1460, opened 2024-03-20
Added `removeUpperDims` utility

ravil-mobile:
This PR adds `removeUpperDims` utility which can remove upper dimensions by indices or names and propagate the information through an entire `TransformMapAttr`. The implementation maybe not algorithmically-efficient but, based on the tests, it leads to correct solutions. Sometimes, it is difficult to decide whether the use of a vector, a map of a set is efficient or not. I would make a note that sizes of the involved containers are small in practice. Thus, the implementation is probably ok performance-wise.

Closes https://github.com/ROCm/rocMLIR-internal/issues/1303

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
+++ b/mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
@@ -230,12 +230,13 @@ Value addPassThroughIndices(OpBuilder &b, Value transformed,
 
 ArrayRef<int64_t> getLowerShape(ArrayAttr transformStack);
 
-ArrayAttr removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
-                          const SetVector<int64_t> &removeeIndices);
+FailureOr<ArrayAttr>
+removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
+                const SetVector<StringRef> &removeDimNamesSet);
 
-ArrayAttr removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
-                          const SetVector<StringRef> &removeeDimNames);
-
+FailureOr<ArrayAttr>
+removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
+                const SetVector<int64_t> &removeIndicesSet);
 } // end namespace rock
 } // end namespace mlir
 #endif
--- a/mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
+++ b/mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
@@ -230,6 +230,12 @@ Value addPassThroughIndices(OpBuilder &b, Value transformed,
 
 ArrayRef<int64_t> getLowerShape(ArrayAttr transformStack);
 
+ArrayAttr removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
+                          const SetVector<int64_t> &removeeIndices);
+
+ArrayAttr removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
+                          const SetVector<StringRef> &removeeDimNames);
+
 } // end namespace rock
 } // end namespace mlir
 #endif
--- a/mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
+++ b/mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
@@ -230,13 +230,12 @@ Value addPassThroughIndices(OpBuilder &b, Value transformed,
 
 ArrayRef<int64_t> getLowerShape(ArrayAttr transformStack);
 
-FailureOr<ArrayAttr>
-removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
-                const SetVector<StringRef> &removeDimNamesSet);
+FailureOr<ArrayAttr> removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
+                                     SetVector<int64_t> removeIndicesSet);
 
 FailureOr<ArrayAttr>
 removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
-                const SetVector<int64_t> &removeIndicesSet);
+                const SetVector<StringRef> &removeDimNamesSet);
 } // end namespace rock
 } // end namespace mlir
 #endif
--- a/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
+++ b/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
@@ -1755,9 +1755,170 @@ Value mlir::rock::addPassThroughIndices(OpBuilder &b, Value transformed,
   return ret;
 }
 
+SmallVector<int64_t>
+findPreservedPositions(ArrayRef<StringRef> names,
+                       const SetVector<StringRef> &removeSet) {
+  SmallVector<int64_t> positions;
+  for (auto name : llvm::enumerate(names)) {
+    if (!removeSet.contains(name.value())) {
+      positions.push_back(name.index());
+    }
+  }
+  return positions;
+}
+
+SmallVector<int64_t>
+generateNewUpperBounds(rock::TransformMapAttr trMap,
+                       const SetVector<StringRef> &removeDimNamesSet) {
+  size_t counter = 0;
+  auto oldUpperBounds = trMap.getUpperBounds().asArrayRef();
+  SmallVector<int64_t> newBounds;
+  for (auto tr : trMap.getOps()) {
+    for (auto upperName : tr.getUpperNames()) {
+      if (!removeDimNamesSet.contains(upperName)) {
+        newBounds.push_back(oldUpperBounds[counter]);
+      }
+      ++counter;
+    }
+  }
+  return newBounds;
+}
+
+SmallVector<int64_t>
+generateNewLowerBounds(rock::TransformMapAttr trMap,
+                       const SetVector<StringRef> &removeDimNamesSet) {
+  size_t lowerBoundCounter = 0;
+  auto oldLowerBounds = trMap.getLowerBounds().asArrayRef();
+  SmallVector<int64_t> newBounds;
+  for (auto tr : trMap.getOps()) {
+    for (auto lowerName : tr.getLowerNames()) {
+      if (!removeDimNamesSet.contains(lowerName)) {
+        newBounds.push_back(oldLowerBounds[lowerBoundCounter]);
+      }
+      ++lowerBoundCounter;
+    }
+  }
+  return newBounds;
+}
+
+template <typename T>
+SmallVector<T> getPreserved(ArrayRef<T> vector,
+                            const SmallVector<int64_t> &preservedPositions) {
+  SmallVector<T> result;
+  for (auto index : preservedPositions) {
+    result.push_back(vector[index]);
+  }
+  return result;
+}
+
+SmallVector<uint32_t> genNewDims(uint32_t start,
+                                 const SmallVector<int64_t> &positions) {
+  SmallVector<uint32_t> newDims = {};
+  if (!positions.empty()) {
+    const uint32_t lb = start;
+    const uint32_t ub = lb + positions.size();
+    newDims = llvm::to_vector_of<uint32_t>(llvm::seq(lb, ub));
+  }
+  return newDims;
+}
+
+FailureOr<ArrayAttr>
+mlir::rock::removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
+                            const SetVector<StringRef> &removeDimNamesSet) {
+  SmallVector<Attribute> results;
+
+  SetVector<StringRef> removeUpperDimSet = removeDimNamesSet;
+
+  for (auto map : transformAttrs) {
+    auto trMap = cast<rock::TransformMapAttr>(map);
+    SmallVector<TransformAttr> newOps;
+    SetVector<StringRef> removeLowerDimSet = {};
+
+    size_t processedUpperDims = 0;
+    size_t processedLowerDims = 0;
+    for (auto tr : trMap.getOps()) {
+
+      SmallVector<int64_t> upperPreservedPositions =
+          findPreservedPositions(tr.getUpperNames(), removeUpperDimSet);
+
+      const bool mustBePreserved =
+          upperPreservedPositions.size() == tr.getUpperNames().size();
+      const bool mustBeCompletelyRemoved = upperPreservedPositions.empty();
+      const bool mustBeModified = !mustBePreserved && !mustBeCompletelyRemoved;
+
+      SmallVector<int64_t> lowerPreservedPositions = {};
+      switch (tr.getType()) {
+      case TransformType::Merge: {
+        assert(!mustBeModified && "assume always one-to-many relation");
+        if (mustBePreserved) {
+          lowerPreservedPositions = llvm::to_vector_of<int64_t>(
+              llvm::seq(0L, static_cast<int64_t>(tr.getLowerDims().size())));
+        }
+        break;
+      }
+      case TransformType::PassThrough: {
+        lowerPreservedPositions = upperPreservedPositions;
+        break;
+      }
+      default:
+        return failure();
+      }
+
+      for (auto name : llvm::enumerate(tr.getLowerNames())) {
+        if (!llvm::is_contained(lowerPreservedPositions, name.index())) {
+          removeLowerDimSet.insert(name.value());
+        }
+      }
+
+      auto newUpperNames =
+          getPreserved(tr.getUpperNames(), upperPreservedPositions);
+      auto newLowerNames =
+          getPreserved(tr.getLowerNames(), lowerPreservedPositions);
+      SmallVector<uint32_t> newUpperDims =
+          genNewDims(processedUpperDims, upperPreservedPositions);
+      SmallVector<uint32_t> newLowerDims =
+          genNewDims(processedLowerDims, lowerPreservedPositions);
+
+      switch (tr.getType()) {
+      case TransformType::Merge: {
+        [[fallthrough]];
+      }
+      case TransformType::PassThrough: {
+        if (!mustBeCompletelyRemoved) {
+          auto passThrough = TransformAttr::get(
+              b.getContext(), tr.getType(), tr.getParams(), newUpperNames,
+              newUpperDims, newLowerNames, newLowerDims);
+          newOps.push_back(passThrough);
+        }
+        break;
+      }
+      default:
+        return failure();
+      }
+      // Note, we don't need to do anything in the case of
+      // `mustBeCompletelyRemoved`
+
+      processedUpperDims += upperPreservedPositions.size();
+      processedLowerDims += lowerPreservedPositions.size();
+    }
+
+    auto newUpperBounds = generateNewUpperBounds(trMap, removeUpperDimSet);
+    auto newLowerBounds = generateNewLowerBounds(trMap, removeLowerDimSet);
+
+    if (!newOps.empty()) {
+      auto newTrMap =
+          TransformMapAttr::get(newOps, newUpperBounds, newLowerBounds);
+      results.push_back(newTrMap);
+    }
+    removeUpperDimSet = removeLowerDimSet;
+  }
+
+  return b.getArrayAttr(results);
+}
+
 SetVector<StringRef>
 convertIndicesToDimNames(const ArrayAttr transformAttrs,
-                         const SetVector<int64_t> &removeeIndices) {
+                         const SetVector<int64_t> &removeIndicesSet) {
   SetVector<StringRef> dimNames = {};
   if (transformAttrs.empty()) {
     return dimNames;
@@ -1770,7 +1931,7 @@ convertIndicesToDimNames(const ArrayAttr transformAttrs,
     for (auto op : transformAttr.getOps()) {
       const auto tattr = op.cast<::rock::TransformAttr>();
       for (auto name : tattr.getUpperNames()) {
-        if (removeeIndices.contains(upperDimCounter)) {
+        if (removeIndicesSet.contains(upperDimCounter)) {
           dimNames.insert(name);
         }
         ++upperDimCounter;
@@ -1781,19 +1942,11 @@ convertIndicesToDimNames(const ArrayAttr transformAttrs,
   return dimNames;
 }
 
-ArrayAttr
+FailureOr<ArrayAttr>
 mlir::rock::removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
-                            const SetVector<int64_t> &removeeIndices) {
+                            const SetVector<int64_t> &removeIndicesSet) {
 
-  SetVector<StringRef> removeDimNames =
-      convertIndicesToDimNames(transformAttrs, removeeIndices);
-  return removeUpperDims(b, transformAttrs, removeDimNames);
-}
-
-ArrayAttr
-mlir::rock::removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
-                            const SetVector<StringRef> &removeeDimNames) {
-  SmallVector<Attribute> results;
-  // TODO (ravil):
-  return b.getArrayAttr(results);
+  SetVector<StringRef> removeDimNamesSet =
+      convertIndicesToDimNames(transformAttrs, removeIndicesSet);
+  return removeUpperDims(b, transformAttrs, removeDimNamesSet);
 }
--- a/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
+++ b/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
@@ -16,6 +16,7 @@
 #include "mlir/Dialect/Rock/IR/TransformMapBuilder.h"
 #include "mlir/Dialect/Rock/utility/loweringUtils.h"
 #include "mlir/Dialect/Rock/utility/math.h"
+#include "mlir/IR/BuiltinAttributes.h"
 #include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/EquivalenceClasses.h"
 #include "llvm/ADT/IndexedMap.h"
@@ -29,6 +30,7 @@
 #include <algorithm>
 #include <iterator>
 #include <numeric>
+#include <unordered_set>
 #include <utility>
 
 #define DEBUG_TYPE "rock-transform-map-utils"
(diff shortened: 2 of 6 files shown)


ravil-mobile:
Hi @manupak. I wanted to show you some of my intermediate results. Probably naming is incorrect, less accurate and needs to be improved. But what do you think in general?

mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
```diff
@@ -230,6 +230,12 @@ Value addPassThroughIndices(OpBuilder &b, Value transformed,
 
 ArrayRef<int64_t> getLowerShape(ArrayAttr transformStack);
 
+FailureOr<ArrayAttr> removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
+                                     SetVector<int64_t> removeIndicesSet);
```
krzysz00, line 234: I think the `SetVector` here could be by const reference?

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -29,6 +30,7 @@
 #include <algorithm>
 #include <iterator>
 #include <numeric>
+#include <unordered_set>
```
krzysz00: Note to me to find the user of this

ravil-mobile:
Hi @manupak and @krzysz00. Could you, please, review the PR?
Thanks in advance!

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -1740,3 +1741,294 @@ Value mlir::rock::addPassThroughIndices(OpBuilder &b, Value transformed,
...
+struct TransformAttrArgs {
+  rock::TransformType type;
+  std::pair<SmallVector<StringRef>, SmallVector<StringRef>> perservedNames;
+  std::pair<SmallVector<uint32_t>, SmallVector<uint32_t>> preservedDims;
+  SmallVector<int64_t> params;
+};
+
+template <DimType Type>
+SmallVector<uint32_t>
+getPerservedIndices(rock::TransformAttr tr,
```
krzysz00: ... does this need to be templated?

krzysz00: Also, typo "Perserved" (-> "Preserved")

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -1740,3 +1741,294 @@ Value mlir::rock::addPassThroughIndices(OpBuilder &b, Value transformed,
...
+  assert(names.size() == dims.size());
+  for (auto [idx, dim] : llvm::enumerate(dims)) {
+    if (llvm::is_contained(preservedDims, dim)) {
+      preseredNames.push_back(names[idx]);
+    }
+  }
+}
+
+template <DimType Type>
+SmallVector<uint32_t> getDifference(rock::TransformAttr tr,
```
krzysz00, line 1818: Maybe this is Rust brain talking, but why does this function take the entire struct and then pick a half of it to operate on based on a template parameter? You could just pass references to the sections you need?

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -1740,3 +1741,294 @@ Value mlir::rock::addPassThroughIndices(OpBuilder &b, Value transformed,
   }
   return ret;
 }
+
+enum DimType { Upper = 0, Lower = 1 };
+
+struct TransformAttrArgs {
+  rock::TransformType type;
+  std::pair<SmallVector<StringRef>, SmallVector<StringRef>> perservedNames;
```
krzysz00: Why not:
```
SmallVector<StringRef> preservedUpperNames;
SmallVector<StringRef> preservedLowerNames;
...
```
and then the utilities that need to be generic over upper or lower dimensions just take `ArrayRef<>` or `SmallVectorImpl<>&`?

This isn't a must-change sort of thing, but this tuples-and-templates structure just feels off

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
krzysz00: (redundant with above general structure notes)

mlir/test/Dialect/Rock/test_transformation_maps_utils.mlir
```diff
@@ -0,0 +1,119 @@
...
+
+func.func @transform_ex5(%arg0: memref<1x64x1024xf32>) -> memref<1x8x1024x64xf32>
+  attributes {
+    remove_dims_by_names = {
+      test0 = ["X"],
+      test1 = ["A", "B", "C"]
+    }
+  } {
+  %0 = rock.transform %arg0 by <affine_map<(d0, d1, d2) -> (d0, d2, d1)> by [<PassThrough ["A"] at [0] -> ["A"] at [0]>, <PassThrough ["B", "C"] at [1, 2] -> ["B", "C"] at [2, 1]>] bounds = [1, 1024, 64] -> [1, 64, 1024]> : memref<1x64x1024xf32> to memref<1x1024x64xf32>
+  %1 = rock.transform %0 by <affine_map<(d0, d1, d2, d3) -> (d0, d2, d3)> by [<AddDim{8} ["X"] at [1] -> [] at []>, <PassThrough ["A", "B", "C"] at [0, 2, 3] -> ["A", "B", "C"] at [0, 1, 2]>] bounds = [1, 8, 1024, 64] -> [1, 1024, 64]> : memref<1x1024x64xf32> to memref<1x8x1024x64xf32>
```
krzysz00: This is a very weird structure - assuming this even needs to be a test pass.

Why not

```
%0 = rock.transform ...
%1 = rock.transform ...
%2 = "drop_dimensions"(%1) {names = ["A"]} : (T) -> (T)
return %2
```
and then "drop_dimensions" is the signal to run the function you're trying to test

krzysz00 (commented):
I'm not going to give this whole thing a nitpick just yet - though I sure noticed a few spots.

I have the following observation:

This is a hairy enough piece of code that it needs a high-level comment explaining what it does and roughly summarizing how it works.

mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
ravil-mobile: No, `removeIndicesSet` it cannot be `const`. The set must be regularly updated while traversing from top to down.

https://github.com/ROCm/rocMLIR/blob/40dce5ee32d0f88f7d9a2bd1fb2831bb194522a3/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp#L1993-L1994

https://github.com/ROCm/rocMLIR/blob/40dce5ee32d0f88f7d9a2bd1fb2831bb194522a3/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp#L1816-L1820

https://github.com/ROCm/rocMLIR/blob/40dce5ee32d0f88f7d9a2bd1fb2831bb194522a3/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp#L1961-L1962

I accept `removeIndicesSet` by value in `removeUpperDims` because I don't want to distort user's data passed as the argument.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
ravil-mobile: removed. Done

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
ravil-mobile: Done

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
ravil-mobile:
```cpp
SmallVector<StringRef> preservedUpperNames;
SmallVector<StringRef> preservedLowerNames;
...
```
The above approach was my initial solution but it was too verbose. `std::pair` results in a more compact code which is  easy to read and, thus, to catch errors.  While working on this PR, I noticed that I made silly, typo-like mistakes quite often - e.g., using `preservedUpperNames` instead of `preservedLowerNames` and/or vice versa. The templated implementation is safer from this kind of errors. 

Conceptually, there is no difference. It is just a matter of taste but I would prefer the templated implementation because it is safer and more compact.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
ravil-mobile: My comment is below.

mlir/test/Dialect/Rock/test_transformation_maps_utils.mlir
ravil-mobile: So, `drop_dimensions` is an Op, right. So, I would need to handle it.

I define tests (i.e., which dimensions to remove) in each function attribute. For example, 

https://github.com/ROCm/rocMLIR/blob/9ed596f77f406852aceab376ccc259d226008e31/mlir/test/Dialect/Rock/test_transformation_maps_utils.mlir#L109-L115

In this particular example, `func.func @transform_ex5` is going to be used in two tests. The expected results of the tests are given about the function - i.e.,

https://github.com/ROCm/rocMLIR/blob/9ed596f77f406852aceab376ccc259d226008e31/mlir/test/Dialect/Rock/test_transformation_maps_utils.mlir#L102-L107

I didn't quite get your point about *"very weird structure"* and *"assuming this even needs to be a test pass"*

mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
krzysz00: Then why is the set of names in the StringRef version of the function passed by const reference?

krzysz00:
So, to give an example of what I'm thinking test wise
```mlir
// CHECK-LABEL: @delete_slow_dim
// %new doesn't need to be a function argument, it could be a "new_underlying"() : () -> memref<4xf32>
// CHECK-SAME: ([[orig:%.+]]: memref<12xf32>, [[new:%.+]]: memref<4xf32>)
func.func(%orig: memref<4x3xf32>) {
  // This could be a bunch of CHECK-SAMEs for readability
  // CHECK: [[new0:%.+]] = rock.transform [[new]] by <affine_map<(d0) -> (d0)> [<Unmerge>{4} ["x"] at [0]>] bounds = [4] -> [4]>
  // CHECK: "remove_dims"(%{{.*}}, [[new0]])
  // Or delete the "remove_dims" operator and add a
  // CHECK: "dims_removed"([[new0]])
  // Or any number of such things
  % = rock.transform %orig by <affine_map<(d0, d1) -> (d1 + 3 * d0)> [<Unmerge{4, 3} ["x", "y"] at [0, 1] -> ["raw"] at [0]>] bounds = [4, 3] -> [12]> : memref<12xf32> to memref<4x3xf32>
  // You could also stick `id` attributes on here if you ever think you need them
  "remove_dims"(%0) {names_to_drop = ["y"]} : (memref<4x3xf32>) -> ()
  // Note that you can have multiple "remove_dims" operations in one function
  return
}

File: mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h (hunk 1 above)
Comment (ravil-mobile) [reply mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h:234]: The algorithm works on the upper dim indices which need to be removed. But we provide an overload which can accept dim Names to be removed, convert it to dim indices and invoke the algorithm. To save an extra copy, I take `SetVector<StringRef>` by const reference and construct a new `SetVector<int64_t>` and invoke the algorithm.

**Comment by ravil-mobile** [comment]
> So, to give an example of what I'm thinking test wise
> 
> ```mlir
> // CHECK-LABEL: @delete_slow_dim
> // %new doesn't need to be a function argument, it could be a "new_underlying"() : () -> memref<4xf32>
> // CHECK-SAME: ([[orig:%.+]]: memref<12xf32>, [[new:%.+]]: memref<4xf32>)
> func.func(%orig: memref<4x3xf32>) {
>   // This could be a bunch of CHECK-SAMEs for readability
>   // CHECK: [[new0:%.+]] = rock.transform [[new]] by <affine_map<(d0) -> (d0)> [<Unmerge>{4} ["x"] at [0]>] bounds = [4] -> [4]>
>   // CHECK: "remove_dims"(%{{.*}}, [[new0]])
>   // Or delete the "remove_dims" operator and add a
>   // CHECK: "dims_removed"([[new0]])
>   // Or any number of such things
>   % = rock.transform %orig by <affine_map<(d0, d1) -> (d1 + 3 * d0)> [<Unmerge{4, 3} ["x", "y"] at [0, 1] -> ["raw"] at [0]>] bounds = [4, 3] -> [12]> : memref<12xf32> to memref<4x3xf32>
>   // You could also stick `id` attributes on here if you ever think you need them
>   "remove_dims"(%0) {names_to_drop = ["y"]} : (memref<4x3xf32>) -> ()
>   // Note that you can have multiple "remove_dims" operations in one function
>   return
> }
> ```

Thanks. I will work on it!

File: mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp [hunk 7]
```diff
@@ -1740,3 +1741,294 @@ Value mlir::rock::addPassThroughIndices(OpBuilder &b, Value transformed,
...
+    }
+    results.push_back(*newTrMap);
+  }
+
+  return b.getArrayAttr(results);
+}
+
+template <DimType Type>
+SetVector<int64_t>
+convertDimNamesToIndices(const ArrayAttr trAttrs,
```
manupak, line 2068: Why is there a template needed ? 
It d be much simpler to accept DimType as an argument ?

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -1740,3 +1741,294 @@ Value mlir::rock::addPassThroughIndices(OpBuilder &b, Value transformed,
...
+        llvm::SmallVector<int64_t>(trMap.getUpperBounds().asArrayRef());
+  }
+
+  for (auto map : transformAttrs) {
+    auto trMap = cast<rock::TransformMapAttr>(map);
+
+    assert(upperBounds.size() ==
+           static_cast<size_t>(trMap.getUpperBounds().size()));
+    llvm::SmallVector<int64_t> lowerBounds = {};
+    auto newTrMap = removeUpperDimsFromMap(b, trMap, removeIndicesSet,
```
manupak: [readability] please use the type here

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -1740,3 +1741,294 @@ Value mlir::rock::addPassThroughIndices(OpBuilder &b, Value transformed,
...
+    const bool mustBeCompletelyRemoved = preservedUpperDims.empty();
+    const bool mustBeModified = !mustBePreserved && !mustBeCompletelyRemoved;
+
+    // compute which lower dimensions must be preserved
+    switch (args.type) {
+    case TransformType::AddDim:
+    case TransformType::ConstDim:
+    case TransformType::Broadcast:
+    case TransformType::Merge: {
+      assert(!mustBeModified && "assume always one-to-many relation");
```
manupak: This is not a valid assumption in our codebase. 
Merge/Unmerge can happen as a passthrough from conditional stack creations.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -1740,3 +1741,294 @@ Value mlir::rock::addPassThroughIndices(OpBuilder &b, Value transformed,
...
+struct TransformAttrArgs {
+  rock::TransformType type;
+  std::pair<SmallVector<StringRef>, SmallVector<StringRef>> preservedNames;
+  std::pair<SmallVector<uint32_t>, SmallVector<uint32_t>> preservedDims;
+  SmallVector<int64_t> params;
+};
+
+template <DimType Type>
+SmallVector<uint32_t>
+getPreservedIndices(rock::TransformAttr tr,
```
manupak, line 1788: I think DimType could be an argument. no need for a template here.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -1740,3 +1741,294 @@ Value mlir::rock::addPassThroughIndices(OpBuilder &b, Value transformed,
...
+      preservedIndices.push_back(dim);
+    }
+  }
+  return preservedIndices;
+}
+
+template <DimType Type>
+void propagatePreservedNames(rock::TransformAttr tr, TransformAttrArgs &args) {
+  const auto &preservedDims = std::get<Type>(args.preservedDims);
+  auto &preseredNames = std::get<Type>(args.preservedNames);
```
manupak: nit : typo

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -1740,3 +1741,294 @@ Value mlir::rock::addPassThroughIndices(OpBuilder &b, Value transformed,
...
+  auto names = Type == DimType::Upper ? tr.getUpperNames() : tr.getLowerNames();
+  auto dims = Type == DimType::Upper ? tr.getUpperDims() : tr.getLowerDims();
+  assert(names.size() == dims.size());
+  for (auto [idx, dim] : llvm::enumerate(dims)) {
+    if (llvm::is_contained(preservedDims, dim)) {
+      preseredNames.push_back(names[idx]);
```
manupak: Is this empty  before ? 
If so this does not feel like a propogation but rather a population of TransformAttrArgs struct.

manupak: Can you add an assert make sure its empty ?

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -1740,3 +1741,294 @@ Value mlir::rock::addPassThroughIndices(OpBuilder &b, Value transformed,
...
+          preservedLowerDims.push_back(localLowerDims[idx]);
+        }
+      }
+      break;
+    }
+    default:
+      return failure();
+    }
+
+    propagatePreservedNames<DimType::Upper>(tr, args);
```
manupak: I dont like this design pattern here (feel free to push back if you think it has merit).

I d rather make TransformAttrArgs struct filled correctly from construction.
i.e. I dont think we need to create it so early in L1831 and make it legal on the go.
Instead, it could be made to accept preservedDims as a mandatory arg for the constructor and use it fill preservedNames --> and remove propagatePreservedNames functions.

I d also make it const here to reduce error surface..

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -1740,3 +1741,294 @@ Value mlir::rock::addPassThroughIndices(OpBuilder &b, Value transformed,
   }
   return ret;
 }
+
+enum DimType { Upper = 0, Lower = 1 };
+
+struct TransformAttrArgs {
```
manupak, line 1763: Please add docs for this.
I d also like a printer added here to support <<

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -1740,3 +1741,294 @@ Value mlir::rock::addPassThroughIndices(OpBuilder &b, Value transformed,
...
+    newRemoveIndicesSet.insert(dim);
+  }
+
+  remapDims<DimType::Upper>(argsVector, removedDims);
+  remapDims<DimType::Lower>(argsVector, removedDims);
+
+  // build new transformations based on the computer preserved data
+  for (auto &args : argsVector) {
+    auto newTr =
+        TransformAttr::get(b.getContext(), args.type, args.params,
```
manupak, line 1980: Having understood the code upto this point,

It feels like we only need to maintain `preservedDims` and `params`.
All the other args could be filled late -- maybe using TransformAttr::get(b.context(), params, preservedDims, tr).

I d also say lets make a constructor like that in the tablegen. Then we dont need to maintain this TransformAttrArgs struct or we can still maintain that struct but only have  `preservedDims` and `params` in it.

manupak (commented):
I did a pass on the implementation of utility.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
ravil-mobile: Hi @manupak,
This is the assumption within a single `TransformMapAttr`. It means that: `<newDim> -> <dimX, dimY, dimZ>, where `newDim` is a new upper dimension;  `dimX`, `dimY`, `dimZ` - some lower dimensions.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
ravil-mobile: Actually, we don't need to have this function generic. Probably, I can remove `DimType` at all

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
ravil-mobile: Do you really want to see `FailureOr<rock::TransformMapAttr>`. Maybe we can rename `newTrMap` to `maybeNewTrMapAttr`. What do you think?

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
ravil-mobile: Yes, but it is not going to make a huge difference. I just wanted to stick to one style in this part of the code. I mean there are several auxiliary functions where I need `template <DimType Type>`. So, I decided to keep doing it.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
ravil-mobile: Thanks! A nice catch!

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
ravil-mobile: I think we can do the following:

```cpp
template <DimType Type>
void propagatePreservedNames(rock::TransformAttr tr, TransformAttrArgs &args) {
  const auto &preservedDims = std::get<Type>(args.preservedDims);
  auto names = Type == DimType::Upper ? tr.getUpperNames() : tr.getLowerNames();
  auto dims = Type == DimType::Upper ? tr.getUpperDims() : tr.getLowerDims();
  assert(names.size() == dims.size());

  SmallVector<StringRef> preservedNames = {};
  for (auto [idx, dim] : llvm::enumerate(dims)) {
    if (llvm::is_contained(preservedDims, dim)) {
      preservedNames.push_back(names[idx]);
    }
  }
  std::get<Type>(args.preservedNames) = std::move(preservedNames);
}
```

It becomes clear that `preservedNames` is empty before entering the loop

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
ravil-mobile: > Having understood the code upto this point,
> 
> It feels like we only need to maintain `preservedDims` and `params`. All the other args could be filled late -- maybe using TransformAttr::get(b.context(), params, preservedDims, tr).
> 
> I d also say lets make a constructor like that in the tablegen. Then we dont need to maintain this TransformAttrArgs struct or we can still maintain that struct but only have `preservedDims` and `params` in it.

You also need `rock::TransformType` to build transforms. In theory, you could build `preservedNames` later but you would need a reference to the original transform. It it easy to do everything in the first for-loop.

So, the reason why one needs to delay the re-building processes is because of dimensions re-numbering. One needs to aggregate the whole information about the removees and only then to make a call to `remapDims`

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
krzysz00: Well, `Merge` is a one-to-(N >= 1) operation.

I think that's the assert not being phrased right.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
krzysz00: I think the templates, even if they're tripping our weird style sense, are reasonable.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
ravil-mobile: > Well, Merge is a one-to-(N >= 1) operation.

Yes. The assertion says that the transform must be either preserved or completely removed. So, there is no possible scenario to `modify` merge. I would suggest to just remove the string literal in the `assert`. I hope it will remove this ambiguity

ravil-mobile: Changed to `must be preserved or removed completely`

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
ravil-mobile: Renamed `propagatePreservedNames` -> `populatePreservedNames`

ravil-mobile:
Hi @krzysz00, could you review this PR again? I changed the test pass in a way as you described above

mlir/test/lib/Dialect/Rock/TestTransformationMapsUtils.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+  RewritePatternSet patterns(ctx);
+  patterns.add<RemoveDimTestPattern>(ctx);
+  FrozenRewritePatternSet frozenPatternSet(std::move(patterns));
+
+  PatternApplicator applicator(frozenPatternSet);
+  applicator.applyDefaultCostModel();
+
+  TransformMapRewriter rewriter(ctx);
+  LogicalResult result = applicator.matchAndRewrite(getOperation(), rewriter);
```
krzysz00, line 191: Huh, `PatternApplicator` is new. I always just went with the FoldGreedily thingy?. Seems useful

mlir/test/lib/Dialect/Rock/TestTransformationMapsUtils.cpp
```diff
@@ -0,0 +1,203 @@
...
+    for (auto trMapAttr : llvm::reverse(newTrMapAttrs)) {
+      auto trOp = b.create<rock::TransformOp>(
+          loc, input, trMapAttr.cast<TransformMapAttr>());
+      input = trOp.getOutput();
+    }
+    b.create<func::ReturnOp>(loc);
+
+    return success();
+  }
+};
```
krzysz00, line 173: Yeah, overall this test pass feels a lot better.

"Remove the function and replace it with the removed-names operations" and "add new function arguments" are both, IMO, reasonable approaches, so ... yeah, I don't have substantial complaints about the test code.

krzysz00 (commented):
I'm a lot happier with the test code

mlir/test/Dialect/Rock/test_transformation_maps_utils.mlir
ravil-mobile: done

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
ravil-mobile: Done, https://github.com/ROCm/rocMLIR/blob/851c359837641e3e66490414760e5b64b69a51d4/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp#L1766-L1788

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -1752,3 +1753,352 @@ Value mlir::rock::addPassThroughIndices(OpBuilder &b, Value transformed,
...
+                              const TransformAttrArgs &args) {
+  auto print = [&stream](StringRef name, auto &container) {
+    stream << name << " = [";
+    if (container.empty()) {
+      stream << "]\n";
+      return;
+    }
+
+    size_t i = 0;
+    for (; i < (container.size() - 1); ++i) {
```
krzysz00: This is `interleaveComma()`

krzysz00 (approved):
I think we're good to land this subject to there being plans to refactor some of the AccelEmitter code to use this guy.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
ravil-mobile: Thanks! It is good to know.
Done

PR #1462, opened 2024-03-20
Extending CAPI for the splitK GEMM scheme

ravil-mobile:
This PR is based on the [proposal](https://github.com/ROCm/AMDMIGraphX/discussions/2858).

Currently the APIs do not have implementations but it will be done as a part of this PR

Follow-up tickets:

- https://github.com/ROCm/rocMLIR-internal/issues/1461
- https://github.com/ROCm/rocMLIR-internal/issues/1460

Follow-up discussion:

- https://github.com/ROCm/rocMLIR/discussions/1481

Closes https://github.com/ROCm/rocMLIR-internal/issues/1391


mlir/include/mlir-c/Dialect/Rock.h
```diff
@@ -140,6 +133,33 @@ bool mlirRockTuningSetFromTable(MlirRockTuningTable perfTable,
...
+// TODO (ravil): document
+MLIR_CAPI_EXPORTED
+bool mlirIsModuleFusible(MlirModule module, MlirStringRef perfStr);
+
+MLIR_CAPI_EXPORTED
+size_t mlirGetNumPrefillArgs(MlirModule module);
+
+// TODO (ravil): document
+MLIR_CAPI_EXPORTED
+void mlirGetPrefillArgsInfo(MlirModule module, size_t *indices,
```
krzysz00, line 155: For my own paranoia, make this take a `size_t` giving the lengths of those arrays.

mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
```diff
@@ -0,0 +1,55 @@
...
+namespace mlir {
+
+struct BufferDependencyAnalysis {
+  BufferDependencyAnalysis(Operation *op);
+
+  static std::optional<memref::AllocOp> getAllocation(Value value);
+  static llvm::SmallVector<Operation *> getReaders(memref::AllocOp allocOp);
+  static llvm::SmallVector<Operation *> getWriters(memref::AllocOp allocOp);
+
+  struct Pair {
```
krzysz00: ... `std::pair` exists

mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
```diff
@@ -0,0 +1,55 @@
...
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/raw_ostream.h"
+
+namespace mlir {
+
+struct BufferDependencyAnalysis {
+  BufferDependencyAnalysis(Operation *op);
+
+  static std::optional<memref::AllocOp> getAllocation(Value value);
```
krzysz00: ... I don't think these should be static

krzysz00: Ok yeah no, looking around, these shouldn't be `static`

mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
```diff
@@ -0,0 +1,55 @@
...
+  static llvm::SmallVector<Operation *> getReaders(memref::AllocOp allocOp);
+  static llvm::SmallVector<Operation *> getWriters(memref::AllocOp allocOp);
+
+  struct Pair {
+    llvm::SmallVector<Operation *> readers;
+    llvm::SmallVector<Operation *> writers;
+  };
+  static Pair getReadersAndWriters(memref::AllocOp allocOps);
+
+  LogicalResult run();
```
krzysz00: This shouldn't exist, and should be in the constructor, probably.

mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
```diff
@@ -0,0 +1,55 @@
...
+  llvm::DenseMap<memref::AllocOp, llvm::SmallVector<Operation *>>
```
krzysz00: If you'll be returning this, const reference.

mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
```diff
@@ -0,0 +1,55 @@
...
+  // The function this analysis was constructed from.
+  func::FuncOp func;
+
+  llvm::DenseMap<memref::AllocOp, llvm::SmallVector<Operation *>> readersTable;
+  llvm::DenseMap<memref::AllocOp, llvm::SmallVector<Operation *>> writersTable;
+};
+
+} // namespace mlir
+
+#endif // ROCK_ANALYSIS_LOWERINGUTILS_H
```
krzysz00: Nit: newline?

mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
```diff
@@ -0,0 +1,55 @@
...
+  getWritersTable() {
+    return writersTable;
+  }
+  func::FuncOp getFuncton() const { return func; }
+
+private:
+  // The function this analysis was constructed from.
+  func::FuncOp func;
+
+  llvm::DenseMap<memref::AllocOp, llvm::SmallVector<Operation *>> readersTable;
```
krzysz00, line 49: There's an argument for going ahead and making this usable on any operation that has a `MemAlloc` effect, but that's future work stuff for if we decide to upstream this.

(Or, equivalently, if we decide to slide bufferization down the pipeline, this might end up needing to be used on tensors)

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -63,17 +63,17 @@ class IndexArrayLength<int n> : ConfinedAttr<IndexArrayAttr, [ArrayCount<n>]>;
...
-                   StrAttr:$arch,
-                   Rock_GemmFeaturesAttr:$features,
-                   OptionalAttr<I32Attr>:$derivedBlockSize,
-                   OptionalAttr<I32Attr>:$gridSize,
-                   IndexArrayLength<4>:$padding,
-                   IndexArrayLength<2>:$strides,
-                   IndexArrayLength<2>:$dilations,
-                   OptionalAttr<RockTuningParamAttrInterface>:$params);
+    dag commonConvArgs = (ins Arg<TensorOrMemRefRankOf<inputTypes, [5]>, "filter", [MemRead]>:$filter,
+                              Arg<TensorOrMemRefRankOf<inputTypes, [5]>, "input", [MemRead]>:$input,
```
krzysz00: This isn't correct.

The backwards data operation, for example, writes to the `"input"` tensor, while the backwards weight one writes to `"filter"`

mlir/include/mlir/Dialect/Rock/Tuning/RockTuning.h
```diff
@@ -73,6 +74,11 @@ LogicalResult tuningTableLookup(TuningTable *perfTable, ModuleOp &mod,
 LogicalResult tuningTableLookupByKey(TuningTable *perfTable,
                                      SmallVectorImpl<char> &out);
 
+bool isSplitKRequested(ModuleOp &mod, StringRef &perfConfig);
```
krzysz00: You don't need to take `mod` by reference here.

mlir/include/mlir/Dialect/Rock/Tuning/RockTuning.h
```diff
@@ -73,6 +74,11 @@ LogicalResult tuningTableLookup(TuningTable *perfTable, ModuleOp &mod,
...
+
+RocmlirSplitKSelectionLikelihood isSplitKFaster(int64_t gDim, int64_t mDim,
```
krzysz00, line 79: ... yeah, it's probably fine to expose the C enum here.

mlir/include/mlir/Dialect/Rock/utility/fusionUtils.h
```diff
@@ -0,0 +1,22 @@
...
+#define ROCK_UTILITY_FISION_H
+
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/Support/LogicalResult.h"
+
+namespace mlir {
+namespace rock {
+LogicalResult testFusibility(ModuleOp mod);
+LogicalResult testFusibility(func::FuncOp func);
```
krzysz00: I think this might be missing `testFusability(Operation* memoryOp, linalg::GenericOp laGeneric);`

Also, can we rename this guy `testFusionLegality();`?

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
```diff
@@ -0,0 +1,153 @@
...
+    if (auto viewOp = dyn_cast<ViewLikeOpInterface>(user)) {
+      findAllUsers(viewOp, users);
+    } else {
+      users.push_back({parent, user});
+    }
+  }
+}
+
+enum EffectType { read, write, unknown };
+EffectType getMemoryEffect(UserWrapper &user) {
```
krzysz00: It's a two pointer struct, might as well pass it by value, especially since you're not modifying it?

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
```diff
@@ -0,0 +1,153 @@
...
+  static inline constexpr EffectType type = EffectType::read;
+};
+
+struct Writers : public llvm::SmallVector<Operation *> {
+  static inline constexpr EffectType type = EffectType::write;
+};
+
+BufferDependencyAnalysis::BufferDependencyAnalysis(Operation *op) {
+  func = dyn_cast<func::FuncOp>(op);
+  assert(func &&
```
krzysz00: I don't think we actually need this requirement? This analysis can work on anything with a body.

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
```diff
@@ -0,0 +1,153 @@
...
+struct Readers : public llvm::SmallVector<Operation *> {
+  static inline constexpr EffectType type = EffectType::read;
+};
+
+struct Writers : public llvm::SmallVector<Operation *> {
+  static inline constexpr EffectType type = EffectType::write;
+};
+
+BufferDependencyAnalysis::BufferDependencyAnalysis(Operation *op) {
+  func = dyn_cast<func::FuncOp>(op);
```
krzysz00: This constructor is the thing that should actually do the analysis and stash it into the object.

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+  llvm::SmallVector<UserWrapper> users;
+  findAllUsers(allocOp, users);
+  for (auto &user : users) {
+    EffectType memoryEffect = getMemoryEffect(user);
+    assign(memoryEffect, user, impl, impls...);
+  }
+}
+
+struct Readers : public llvm::SmallVector<Operation *> {
```
krzysz00: I ... don't like this.

Just have the function take a `SmallVectorImpl<Operation*>&` and a `EffectType` as regular parameters?

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+  BufferDependencyAnalysis::Pair result;
+  result.readers = std::move(readers);
+  result.writers = std::move(writers);
+
+  return result;
+}
+
+std::optional<memref::AllocOp>
+BufferDependencyAnalysis::getAllocation(Value value) {
```
krzysz00: This should be cached in the analysis.

krzysz00: ... and part of me is tempted to have a nice table that also gives you all the views, so that `untransform()` is a analysis lookup, but that's a future PR.

mlir/lib/CAPI/Dialect/Rock.cpp
```diff
@@ -138,3 +141,69 @@ MLIR_CAPI_EXPORTED size_t mlirRockTuningGetKey(MlirModule module, char *buf,
...
+
+// TODO (ravil): document
+MLIR_CAPI_EXPORTED
+size_t mlirGetNumPrefillArgs(MlirModule module) {
+  auto mod = unwrap(module);
+  assert(mod.getRegion().getBlocks().size() == 1 &&
+         "expected a single block/function in a module");
+
+  std::optional<LLVM::LLVMFuncOp> func = std::nullopt;
+  mod.walk([&](LLVM::LLVMFuncOp op) { func = op; });
```
krzysz00, line 177: ... having looked at documentation some more, there're probably places where we called `walk()` where we just wanted https://mlir.llvm.org/docs/Tutorials/UnderstandingTheIRStructure/#filtered-iterator-getopsopty

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -882,5 +900,79 @@ LogicalResult tuningTableLookup(TuningTable *perfTable, ModuleOp &mod,
...
+  // M/block N/block K/block M/wave N/wave kPack
+  const std::vector<std::vector<uint32_t>> rangeGemmParams = {
+      {4, 8, 16, 32, 64, 128, 256},
+      {16, 32, 64, 128, 256},
+      {1, 2, 4, 8},
+      {1, 4, 8, 16}};
+
+  rock::GemmSize gemmSize(gDim, mDim, kDim, nDim);
+  llvm::SmallSetVector<int64_t, 10> splitKValues = {};
+  double minWorkImbalance = std::numeric_limits<double>::max();
```
krzysz00, line 948: We'll definitely want a comment making sure this is kept up-to-date with the tuning ranges.

Also, said tuning ranges are datatyp dependent.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -882,5 +900,79 @@ LogicalResult tuningTableLookup(TuningTable *perfTable, ModuleOp &mod,
...
+      for (uint32_t kPerBlock : rangeGemmParams[2]) {
+        for (uint32_t kPack : rangeGemmParams[3]) {
+          const double currWorkImbalance = computeWorkImbalance(
+              gemmSize, mPerBlock, nPerBlock, kPerBlock, kPack, numCUs);
+          minWorkImbalance = std::min(currWorkImbalance, minWorkImbalance);
+
+          llvm::SmallVector<int64_t> currSplitKValues =
+              computeOptimalSplitKFactors(gemmSize, mPerBlock, nPerBlock,
+                                          kPerBlock, kPack, numCUs);
+          std::for_each(
```
krzysz00: There's a `llvm::` version of this that doesn't want the explicit begin and end calls, almost certainly

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -882,5 +900,79 @@ LogicalResult tuningTableLookup(TuningTable *perfTable, ModuleOp &mod,
...
+              currSplitKValues.begin(), currSplitKValues.end(),
+              [&splitKValues](int64_t value) { splitKValues.insert(value); });
+        }
+      }
+    }
+  }
+
+  if (splitKValues.size() == 1) {
```
krzysz00, line 968: This'll never hit, I think, since each calculation returns, at minimum, a vector containing `1`.

Perhaps the `SmallVector` above should be some sort of binary heap situation?

mlir/lib/Dialect/Rock/utility/fusionUtils.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include "llvm/Support/Casting.h"
+
+using namespace mlir;
+using namespace mlir::rock;
+
+struct TestFusibilityPass
+    : public PassWrapper<TestFusibilityPass, OperationPass<func::FuncOp>> {
+  void runOnOperation() override {
+    BufferDependencyAnalysis &analysis =
+        getAnalysis<BufferDependencyAnalysis>();
```
krzysz00: Note that `getAnalysis()` implicitly passes in the result of `getOperation()` to the pass constructor.

mlir/test/CAPI/CMakeLists.txt
```diff
@@ -63,3 +64,20 @@ target_link_libraries(mlir-mixr-fullc-test
...
+target_link_libraries(mlir-mixr-split-k-test
+  PRIVATE
+  MLIRCAPIIR
+  MLIRCAPIRegisterRocMLIR
+  MLIRCAPIMIGraphX
+  MLIRCAPIRock
+  MLIRCAPIGPU
+  MLIRMIGraphXPipeline
+  MLIRRockPipeline
+)
```
krzysz00: Nit: newline

krzysz00 (commented):
Notes on why the CI might be blowing up, other comments, haven't really looked at the tests yet.

mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
ravil-mobile: Yes, I know. But, I don't usually use `std::pair` when elements of a pair have the same type because the semantical meaning of each pair element becomes not obvious. In this particular scenario, it is better to use a struct. Probably I shouldn't have called it `Pair`. I will rename it back to `Result`

mlir/include/mlir/Dialect/Rock/utility/fusionUtils.h
ravil-mobile: Yes, we can. I will rename it

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: Wait, `splitKValues` has type `llvm::SmallSetVector`. So, it should work

mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
krzysz00: Yeah, fair, whether to give the results meaningful names could go either way so I'm not going to complain too much.

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
ravil-mobile: I don't know. I was quite happy with this implementation. You extend basically `llvm::SmallVector<Operation *>` with just a static field. If I remember correctly the data object layout is not going to change. Plus, it is an internal structure regarding `BufferDependencyAnalysis.cpp` file. So, nobody else is suppose to use it.

You actually need to associate `SmallVectorImpl<Operation*>&` with some `EffectType` value to me the following work

https://github.com/ROCm/rocMLIR/blob/8fc61db4d3638e8b647519598273cdd2488554c2/mlir/lib/Analysis/BufferDependencyAnalysis.cpp#L73-L83

https://github.com/ROCm/rocMLIR/blob/8fc61db4d3638e8b647519598273cdd2488554c2/mlir/lib/Analysis/BufferDependencyAnalysis.cpp#L106-L117

The alternative would be:

```cpp
struct Reader {
  SmallVectorImpl<Operation*>& container;
  static inline constexpr EffectType type = EffectType::read;
}

struct Writer {
  SmallVectorImpl<Operation*>& container;
  static inline constexpr EffectType type = EffectType::write;
}
```
I don't like that `SmallVectorImpl<Operation*>& container;` gets repeated in both auxiliary structures

krzysz00: Why do we need this generic `assign()` at all? There's like, two or three cases here? We can just hand-route them to `readers` or `writers` as needed?

krzysz00: (To be clear, this isn't a "I won't merge the way you did it", it's a "hold on, the surrounding codebase doesn't pull this trick anywhere I've seen it, are we *sure* we want to pull in this new idiom?")

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
ravil-mobile: Makes sense. So, shall we assign `[MemRead, MemWrite]` for both `filter` and `input` or do you have another idea?

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
ravil-mobile: I've just look at some analyses from `external/llvm-project`. Yes, you're right. The logic of an analysis should be implemented in a constructor. I will rework.

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
manupak: Why do we even need EffectType here ? 

As in would we have a Reader that have some other effect type at all ?

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
```diff
@@ -0,0 +1,153 @@
...
+}
+
+template <typename Impl, typename... Impls>
+void assign(EffectType effect, UserWrapper &user, Impl &impl, Impls &...impls) {
+  assign(effect, user, impl);
+  assign(effect, user, impls...);
+}
+
+template <typename Impl, typename... Impls>
+void find(memref::AllocOp allocOp, Impl &impl, Impls &...impls) {
```
manupak: find is too generic of a name....
What does Impl and Impls mean here ?

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+template <typename Impl>
+void assign(EffectType effect, UserWrapper &user, Impl &impl) {
+  if (effect == Impl::type) {
+    impl.push_back(user.op);
+  }
+}
+
+template <typename Impl, typename... Impls>
+void assign(EffectType effect, UserWrapper &user, Impl &impl, Impls &...impls) {
```
manupak: What does Impl and Impls mean here ?

mlir/include/mlir-c/Dialect/Rock.h
ravil-mobile: take the sizes of array and add an assert

mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
ravil-mobile: `getAllocation` should be a table lookup. The logic is going to be impl. in the constructor; create a lookup table.

mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
ravil-mobile: inpl. `getReadersAndWriters`  as a table lookup

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
ravil-mobile: See `DeclareOpInterfaceMethods<MemoryEffectsOpInterface>`. Append `RockDialect.cpp`

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: Put a comment with the reference where the loops were taken

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+  return result;
+}
+
+std::optional<memref::AllocOp>
+BufferDependencyAnalysis::getAllocation(Value value) {
+  while (auto viewOp = dyn_cast<ViewLikeOpInterface>(value.getDefiningOp())) {
+    value = viewOp.getViewSource();
+  }
+  if (auto allocOp = dyn_cast<memref::AllocOp>(value.getDefiningOp())) {
```
manupak: What if traces back to func_arg and not an allocation ?

mlir/lib/Dialect/Rock/utility/fusionUtils.cpp
```diff
@@ -0,0 +1,86 @@
...
+        func.walk([&](rock::RockGemmWrapperInterface gemmOp) -> WalkResult {
+          auto gemmResult = gemmOp->getOperand(2);
+          auto allocOp = BufferDependencyAnalysis::getAllocation(gemmResult);
+          if (!allocOp.has_value()) {
+            return WalkResult::advance();
+          }
+
+          // make sure that no `linalg::GenericOp` reads from a gemm output
+          if (readersTable.contains(allocOp.value())) {
+            for (Operation *op : readersTable.at(allocOp.value())) {
```
manupak: Since you checks whether op is contained in the vector, vector is not the right choice of data structure.

mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
```diff
@@ -0,0 +1,55 @@
...
+struct BufferDependencyAnalysis {
+  BufferDependencyAnalysis(Operation *op);
+
+  static std::optional<memref::AllocOp> getAllocation(Value value);
+  static llvm::SmallVector<Operation *> getReaders(memref::AllocOp allocOp);
+  static llvm::SmallVector<Operation *> getWriters(memref::AllocOp allocOp);
+
+  struct Pair {
+    llvm::SmallVector<Operation *> readers;
+    llvm::SmallVector<Operation *> writers;
```
manupak: You might want to use SmallSet here instead, because later you want to check membership

manupak (commented):
I did a pass.

This relies on operations found on the kernel having MemoryEffectsOp Interface implemented.  Is that verified ?

Also what happens when memory effect is "unknown" ? 

What happens a operation that both reads and writes to a buffer ?

krzysz00:
To answer some of the questions (@manupak )

Yes, we're relying on MemoryEffectOpInterface . We know it's implemented right for upstream and to the extent it isn't for our ops, this PR (fixes/will fix) that.

Things that trace to function arguments will just not have an associated buffer, which is good news.

And a thing that reads and writes a buffer is a writer of that buffer for these purposes. (If it's updating a buffer in place in a more annoying way, like `rock.gemm %buf = %buf * %buf` ... then that's just not allowed, really)

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
ravil-mobile: Hi @manupak, I talked to @krzysz00 and we decided to change this implementation and get rid off this template parameter pack unpacking implementation

mlir/include/mlir/Dialect/Rock/Tuning/RockTuning.h
ravil-mobile: done

mlir/include/mlir/Dialect/Rock/utility/fusionUtils.h
ravil-mobile: I renamed `testFusability` to `testFusionLegality`. 

I would suggest to implement `testFusionLegality(Operation* memoryOp, linalg::GenericOp laGeneric)` in a dedicated PR when there will be a demand for such a feature.

manupak:
> struct Reader {
  SmallVectorImpl<Operation*>& container;
  static inline constexpr EffectType type = EffectType::read;
}

I think this is unnecessary.
Because by knowing the type is "Reader" you know the effect type to be read.
Therefore I dont see the additional information the member "type" have in this struct.

manupak:
> And a thing that reads and writes a buffer is a writer of that buffer for these purposes. (If it's updating a buffer in place in a more annoying way, like rock.gemm %buf = %buf * %buf ... then that's just not allowed, really).

LA generics will have in-place updates and it is one of the use cases destination passing style-based bufferization.
I understand based on the current implementation we are only interested in rock.gemm's output buffer readers and writers, however, since we have decided to pull out "BufferDependencyAnalysis" as a seperate generic pass rather than an implementation detail of TestFusability - I think we need to handle this.

Moreover, I think handling of such is not that hard -- just add it both readers and writers ?

manupak:
> Things that trace to function arguments will just not have an associated buffer, which is good news.

It does have a buffer; its not just an alloca one.
It is an entirely valid fusion IR to have the following :

```mlir
rock.gemm %arg2 = %arg0 * %arg1
linalg.elementwise relu ins(%arg2) outs(%arg2) 
```

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
ravil-mobile: Hi @krzysz00, Can I ask you to double check the following? 

https://github.com/ROCm/rocMLIR/blob/9e59554827a86467b0f99eb05d5767853820db1e/mlir/lib/Dialect/Rock/IR/RockDialect.cpp#L697-L730

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
ravil-mobile: Hi @manupak, the solution has been refactored. Now, there is not `assign` function

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
ravil-mobile: Hi @manupak, I talked to @krzysz00 about it. He said that we should only tract `allocOp` in this analysis. In the case which you describe, the function returns `std::nullopt`

mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
ravil-mobile: @krzysz00, I moved `getAllocation` to `fusionUtils.cpp`

mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
ravil-mobile: done. I removed `getReadersAndWriters` because we don't need it due to our lookup approach. I only left `getReaders` and `getWriters`

mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
ravil-mobile: done

mlir/lib/Dialect/Rock/utility/fusionUtils.cpp
ravil-mobile: done

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
ravil-mobile: yep, done

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
ravil-mobile: done

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
ravil-mobile: done

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: done

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
ravil-mobile: @manupak @krzysz00, I completely changed the implementation.

mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
ravil-mobile: done

ravil-mobile:
> > Things that trace to function arguments will just not have an associated buffer, which is good news.
> 
> It does have a buffer; its not just an alloca one. It is an entirely valid fusion IR to have the following :
> 
> ```mlir
> rock.gemm %arg2 = %arg0 * %arg1
> linalg.elementwise relu ins(%arg2) outs(%arg2) 
> ```

Hi @manupak, I think you are correct in general. But I think this analysis is meant to run after the `migraphx` pipeline. MIGraphX IR is based on SSA which will result in `AllocOp` after the bufferization pipeline

krzysz00:
@manupak The example you gave, while a potentially valid fusion in general, is not supported by any of our pipelines.

That is, I'm _pretty_ sure that your example won't work in rocmlir today

krzysz00:
Also, re readers+writers, we do have our gemms annotated as reading and writing the output buffer because split-K is a thing.

... so we probably should put it in both, maybe, but we'd need to take those intersections in stuff like vectorizations

krzysz00:
... I've just realized that when we're doing readers and writers, passing around `OpOperand`s instead of just `Operaiton*` (or only `OpOperand`s if they carry an `Operation*`)

mlir/lib/CAPI/Dialect/Rock.cpp
ravil-mobile: In theory, `getOps<LLVM::LLVMFuncOp> should work but for some reason it returns `nullptr` for me

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
ravil-mobile: done

mlir/lib/Dialect/Rock/utility/fusionUtils.cpp
ravil-mobile: No, I don't test whether an op in the vector. I test whether at least one reader of a given `allocOp` is `linalg::GenericOp`. I think that the data structure is correct in this case.

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
ravil-mobile: Make a dedicated ticket out of this

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
ravil-mobile: Conv2DBwdDataOp; getFilter (read, write)

Conv2DBwdWeightOp; if workspace set -> (read, write); filter -> () | otherwise no side effect for workspace () but `filter` (read, write)

mlir/include/mlir/Dialect/Rock/utility/fusionUtils.h
ravil-mobile: move to a dedicated ticket `testFusionLegality(Operation* memoryOp, linalg::GenericOp laGeneric)`

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
ravil-mobile: resolved

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
ravil-mobile: The corresponding tickets is [here](https://github.com/ROCm/rocMLIR-internal/issues/1460)

mlir/include/mlir/Dialect/Rock/utility/fusionUtils.h
ravil-mobile: The corresponding ticket has been created. See, [here](https://github.com/ROCm/rocMLIR-internal/issues/1461)

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
ravil-mobile: done

mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
```diff
@@ -0,0 +1,48 @@
...
+
+  const llvm::DenseMap<memref::AllocOp, llvm::SmallVector<Operation *>> &
+  getReadersTable() {
+    return readersTable;
+  }
+  const llvm::DenseMap<memref::AllocOp, llvm::SmallVector<Operation *>> &
+  getWritersTable() {
+    return writersTable;
+  }
+  Operation *getFuncton() const { return op; }
```
krzysz00: Is `getFunction()` still the right name here?

mlir/include/mlir/Dialect/Rock/utility/fusionUtils.h
```diff
@@ -0,0 +1,24 @@
...
+// Part of the MLIR Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===-----------------------------------------------------===//
+#ifndef ROCK_UTILITY_FISION_H
+#define ROCK_UTILITY_FISION_H
+
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
```
krzysz00: We can probably forward-declare the classes here.

mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
```diff
@@ -0,0 +1,48 @@
...
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/Support/LogicalResult.h"
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/raw_ostream.h"
+
+namespace mlir {
+
+struct BufferDependencyAnalysis {
```
giuseros, line 27: Could you add a bit of documentation to this structs (and its functions)?

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
```diff
@@ -0,0 +1,110 @@
...
+  return std::nullopt;
+}
+
+std::optional<llvm::SmallVector<Operation *>>
+BufferDependencyAnalysis::getWriters(memref::AllocOp allocOp) {
+  if (writersTable.contains(allocOp)) {
+    return writersTable[allocOp];
+  }
+  return std::nullopt;
+}
```
giuseros: I think also this needs more comments/doc

krzysz00 (approved):
LGTM

mlir/include/mlir/Dialect/Rock/utility/fusionUtils.h
```diff
@@ -0,0 +1,24 @@
...
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/Support/LogicalResult.h"
+
+namespace mlir {
+namespace rock {
+std::optional<memref::AllocOp> getAllocation(Value value);
+LogicalResult testFusionLegality(ModuleOp mod);
+LogicalResult testFusionLegality(func::FuncOp func);
```
giuseros: Could you add more docs here as well?

giuseros (changes requested):
Hi @ravil-mobile , would you mind adding a bit of comments/documentation in some parts of the code? I was going through it, but it took me some time to figure everything out :)

mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
ravil-mobile: Good point! I renamed it to `getOperaion`. Additionally, I added a comment similar to the one in mlir `Liveness` analysis.

done

mlir/include/mlir/Dialect/Rock/utility/fusionUtils.h
ravil-mobile: done

mlir/lib/Dialect/Rock/utility/fusionUtils.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include "mlir/Pass/Pass.h"
+#include "mlir/Pass/PassManager.h"
+#include "mlir/Support/LogicalResult.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/Casting.h"
+
+using namespace mlir;
+using namespace mlir::rock;
+
+std::optional<memref::AllocOp> mlir::rock::getAllocation(Value value) {
```
giuseros: This looks very similar to this:
https://github.com/ROCm/rocMLIR/blob/762fdc80e1be85c4124ff5abe827fd8aed5c0728/mlir/lib/Dialect/Rock/utility/loweringUtils.cpp#L519
Would it be possible to unify them?

giuseros (commented):
Thanks for applying the changes! I have only a comment about `findAlloc`/`getAlloc`

mlir/lib/Dialect/Rock/utility/fusionUtils.cpp
ravil-mobile: Thanks @giuseros . It is possible if you are ok to 1) template the function and 2) move the implementation to the header.  Do you think it is going to be ok?

ravil-mobile: Hi @giuseros, I managed to address your comment. Can you, please, re-review?

giuseros: Ah right, you are looking for memref allocs and I was looking for gpuallocs. I don't have strong opinions, so feel free to do it if you like , but I won't stop the approval :)

giuseros (approved):

mlir/lib/Analysis/BufferDependencyAnalysis.cpp
ravil-mobile: done

mlir/include/mlir/Dialect/Rock/utility/fusionUtils.h
ravil-mobile: done

mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
ravil-mobile: done

mlir/lib/Dialect/Rock/utility/fusionUtils.cpp
ravil-mobile: I managed to combine the implementations of both of them.
done

PR #1463, opened 2024-03-21
Clean up intermediate loop code, fix excess loop iterations

krzysz00:
1. Some of our transforms had single-argument Merge{} or Unmerge{} calls, which could've thrown off some analyses (looks like they didn't but they could've)
2. When a source buffer was made of vectors but the destination wasn't, the threadwise_read_into loop ran for too many iterations. I *think* this didn't cause a bug because the out of bounds stores were, after unrolling, trivially removable, but it probably didn't help our codegen.
3. Allow for eliminating single-iteration loops during SugarToLoops.

From what I can tell, these things don't affect our codegen any, but they were bugging me so.

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
@@ -167,6 +167,12 @@ struct TransformingForRewritePattern
     for (const auto &pair : llvm::zip(bounds, strides)) {
       int64_t bound, stride;
       std::tie(bound, stride) = pair;
+      // This is a one-iteration loop, so don't actually emit a loop so as to
+      // enable constant folding.
+      if (bound == stride && bound > 0) {
+        ivs.push_back(b.createOrFold<arith::ConstantIndexOp>(loc, 0));
+        continue;
+      }
       llvm::SmallVector<Value, 3> iterInits;
       if (loops.empty())
         llvm::copy(op.getIterInits(), std::back_inserter(iterInits));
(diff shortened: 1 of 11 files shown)


mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
```diff
@@ -241,7 +247,24 @@ struct TransformingForRewritePattern
...
+      auto yieldOp = cast<rock::YieldOp>(op.getBody()->getTerminator());
+      b.replaceAllUsesWith(op.getResults(), yieldOp.getOperands());
+      SmallVector<Value> blockArgValues;
+      for (auto [blockArg, initValue] :
+           llvm::zip(op.getIterArgs(), op.getIterInits())) {
+        cloneMap.map(blockArg, initValue);
+      }
+      blockArgValues.reserve(op.getBody()->getNumArguments());
+      for (auto blockArg : op.getBody()->getArguments())
+        blockArgValues.push_back(cloneMap.lookup(blockArg));
```
sjw36, line 261: Wouldn't this be the same as IterInits?

sjw36 (approved):
Looks good and fixes the amdgpu-promote-alloca problem in the backend!

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
krzysz00: No, the block arguments include things like the lower indices and the validity

PR #1465, opened 2024-03-22
Implement AMDGPU DPP operation in MLIR.

stefankoncarevic:
This commit adds support for the AMDGPU Data Parallel Primitives (DPP) operation in MLIR. The implementation focuses on the mov_dpp instruction, which allows efficient data movement between threads within a wavefront on AMD GPUs. This includes support for various DPP operations such as quad_perm, row_shl, row_shr, row_ror, wave_shl, wave_shr, wave_ror, wave_rol, row_mirror, row_half_mirror, row_bcast_15, and row_bcast_31 on all MI architectures. 
Operations that are not supported for navi2x and navi3x architectures is next:
wave_shl, wave_shr, wave_ror, wave_rol, row_bcast.
For now all are supported due to the reduction task how is next so when I finish that we can see if we need all the operations or some we can remove or see if there are other operations that can replace them for the navi2x or navi3x architectures.


external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
```diff
@@ -12,6 +12,7 @@
 include "mlir/Interfaces/SideEffectInterfaces.td"
 include "mlir/IR/EnumAttr.td"
 include "mlir/IR/OpBase.td"
+//include "mlir/Dialect/LLVMIR/LLVMOpBase.td"
```
stefankoncarevic: I forgot to remove.

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
```diff
@@ -411,6 +412,47 @@ def AMDGPU_RawBufferAtomicUminOp :
...
+                  OptionalAttr<I32Attr>:$row_shr,
+                  OptionalAttr<I32Attr>:$row_ror,
+                  OptionalAttr<I32Attr>:$row_mask,
+                  OptionalAttr<I32Attr>:$bank_mask,
+                  OptionalAttr<BoolAttr>:$bound_ctrl,
+                  OptionalAttr<BoolAttr>:$wave_shl,
+                  OptionalAttr<BoolAttr>:$wave_shr,
+                  OptionalAttr<BoolAttr>:$wave_ror,
+                  OptionalAttr<BoolAttr>:$wave_rol,
+                  OptionalAttr<BoolAttr>:$row_mirror,
```
krzysz00: With the schemes like `wave_ror` and `wave_shr`, you're only allowed to have one of them.

So, instead of all these separate optionals, I'd do the following:
1. Define an enum attribute, called something like `DPPPerm` (see `AMDGPU_MFMAPermB` in this same file for an example). Make the operation take a `DPPPerm:$kind`
2. Stick a `permArgument` on the operation, which'd be an optional `Attribute` (so either an ArrayAttr for quad_perm, an integer for the  cases that need one, and either unset or the unit attribute (I'd accept either) for the permutation types that take an argument).

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
```diff
@@ -411,6 +412,47 @@ def AMDGPU_RawBufferAtomicUminOp :
...
 
+def AMDGPU_DPP : AMDGPU_Op<"dpp", [SameTypeOperands, AllTypesMatch<["result", "src"]>]>,
+  Arguments<(ins AnyType:$src,
+                  OptionalAttr<ArrayAttr>:$quad_perm,
+                  OptionalAttr<I32Attr>:$row_shl,
+                  OptionalAttr<I32Attr>:$row_shr,
+                  OptionalAttr<I32Attr>:$row_ror,
+                  OptionalAttr<I32Attr>:$row_mask,
+                  OptionalAttr<I32Attr>:$bank_mask,
+                  OptionalAttr<BoolAttr>:$bound_ctrl,
```
krzysz00: `OptionalAttr<BoolAttr>` is redundant with either `OptionalAttr<UnitAttr>` or just `BoolAttr` or

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
```diff
@@ -411,6 +412,47 @@ def AMDGPU_RawBufferAtomicUminOp :
   let hasVerifier = 1;
 }
 
+def AMDGPU_DPP : AMDGPU_Op<"dpp", [SameTypeOperands, AllTypesMatch<["result", "src"]>]>,
```
krzysz00: This should be called DPPOp by convention, so that the operation is `AMDGPU::DPPOp`

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -830,6 +830,178 @@ LogicalResult PackedStochRoundFp8OpLowering::matchAndRewrite(
...
+struct AMDGPUDPPLowering : public ConvertOpToLLVMPattern<DPP> {
+  AMDGPUDPPLowering(LLVMTypeConverter &converter, Chipset chipset)
+      : ConvertOpToLLVMPattern<DPP>(converter), chipset(chipset) {}
+  Chipset chipset;
+
+  Value createConstantOrUseDefault(Operation *op,
+                                   ConversionPatternRewriter &rewriter,
+                                   StringRef attrName, Type type,
+                                   int defaultValue) const {
+    if (auto attr = op->getAttr(attrName)) {
```
krzysz00: Yeah, that this function exists kinda implies you want `DefaultValuedAttr`

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -830,6 +830,178 @@ LogicalResult PackedStochRoundFp8OpLowering::matchAndRewrite(
...
+      return rewriter.create<LLVM::ConstantOp>(op->getLoc(), type, attr);
+    } else {
+      return rewriter.create<LLVM::ConstantOp>(
```
krzysz00: As a side note, if you go look in ` llvm/include/llvm/IR/IntrinsicsAMDGPU.td`, you'll notice that all the flags (the control words, bounds_ctrl, that sort of thing) are all marked `ImmArg`, which means they're required to be constants.

As a matter of cleanup, I'd suggest going into `ROCDL.td` and changing the definition of the DPP intrinsic to take all those arguments as I32 or boolean attributes instead of as arguments.

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -830,6 +830,178 @@ LogicalResult PackedStochRoundFp8OpLowering::matchAndRewrite(
...
+
+    // Convert the source operand to the corresponding LLVM type
+    Location loc = op.getLoc();
+    Value src = adaptor.getSrc();
+    Type srcLLVMType = typeConverter->convertType(src.getType());
+    Type srcType = src.getType();
+    Value dppCtrl;
+
+    // If the source type is f16 or f32, use bitcast to convert it to i32
+    if (srcType.isF16()) {
```
krzysz00: 1. If `srcType` isn't an integer, then get its bitwidth and bitcast it to an integer of that same width.
2. Then, if the bitwidth of the integer is less than 32, extend it to 32.

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -830,6 +830,178 @@ LogicalResult PackedStochRoundFp8OpLowering::matchAndRewrite(
...
+      BCAST31 = 0x143,
+    };
+
+    // List of all attributes to check for dppCtrl.
+    llvm::ArrayRef<std::string> attributeNames = {
+        "quad_perm",  "row_shl",         "row_shr",      "row_ror",
+        "row_mirror", "row_half_mirror", "wave_shl",     "wave_shr",
+        "wave_ror",   "wave_rol",        "row_bcast_15", "row_bcast_31"};
+    int attrCount = 0;
+    for (const auto &attrName : attributeNames) {
```
krzysz00: ... Yeah, the fact that you can specify multiple DPP types and this'll just kinda pick one is a problem.

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -830,6 +830,178 @@ LogicalResult PackedStochRoundFp8OpLowering::matchAndRewrite(
...
+        createConstantOrUseDefault(op, rewriter, "bank_mask", llvmI32Type, 0xf);
+    auto boundCtrl =
+        createConstantOrUseDefault(op, rewriter, "bound_ctrl", llvmI1Type, 0);
+
+    // create a ROCDL_DPPMovOp instruction with the appropriate attributes
+    auto dppMovOp = rewriter.create<ROCDL::DPPMovOp>(
+        loc, llvmI32Type, src, dppCtrl, rowMask, bankMask, boundCtrl);
+
+    Value result = dppMovOp.getRes();
+    if (srcType.isF16()) {
```
krzysz00: And here you reverse the casts: truncate down if needed, then bitcast back.

krzysz00: ... That is to say, the fact that I want to use `bf16` here - especially since post upstream merge we'll actually be using LLVM's `bfloat` type probably - should require no additional code.

external/llvm-project/mlir/test/Conversion/AMDGPUToROCDL/dpp.mlir
```diff
@@ -0,0 +1,68 @@
+// RUN: mlir-opt -convert-amdgpu-to-rocdl=chipset=gfx908 %s | mlir-translate -mlir-to-llvmir | llc -mtriple=amdgcn-hsa -mcpu=gfx908 | FileCheck %s
```
krzysz00: ... You don't need to go all the way to assembly on this test. Actually, having looked at the CMake for the MLIR tests, you *can't* reliably call `llc` - it's not built when the LLVM build doesn't include the native target.

krzysz00: Heck, this directory shouldn't even be calling `mlir-translate`. You want to check for `rocdl.dpp_mov` operations here.

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
```diff
@@ -411,6 +412,47 @@ def AMDGPU_RawBufferAtomicUminOp :
...
+    - Row shift right by 1–15 lanes (`row_shr`)
+    - Row rotate right by 1–15 lanes (`row_ror`)
+    - Reverse within a row (`row_mirror`)
+    - Reverse within a half-row (`row_half_mirror`)
+    - Broadcast the 15th lane of each row to the next row (`row_bcast`)
+    - Broadcast lane 31 to rows 2 and 3 (`row_bcast`)
+  }];
+  let results = (outs AnyType:$result);
+  let assemblyFormat = [{
+    operands attr-dict `:` type($result)
```
krzysz00: Let's give this a slightly nicer assembly format, for instance, assuming a DPP scheme gets added:

```
$src $kind (`(` $permArgument^ `)`)? oilist (`row_mask` $rowMask | `bank_mask` $bankMask | `bound_ctrl` $boundCtrl) `:` type($src)
```

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
```diff
@@ -411,6 +412,47 @@ def AMDGPU_RawBufferAtomicUminOp :
   let hasVerifier = 1;
 }
 
+def AMDGPU_DPP : AMDGPU_Op<"dpp", [SameTypeOperands, AllTypesMatch<["result", "src"]>]>,
+  Arguments<(ins AnyType:$src,
```
krzysz00, line 438: While we're here, since I started looking at things the compiler itself does and recently read the public example we have of how to do a DPP reduction:
- Add an optional operand called  `old` . If `old` is present, lower to `update.dpp` instead of `mov.dpp`.
- (which means you'll need to crack the rocdl files open and provide a wrapper around that intrinsic and add the relevant tests, but that should be mostly a copy-paste job once the definition of mov.dpp's been fixed per below)

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -830,6 +830,178 @@ LogicalResult PackedStochRoundFp8OpLowering::matchAndRewrite(
...
+            attrValue |= num << (i * 2);
+          }
+        }
+        dppCtrl =
+            rewriter.create<LLVM::ConstantOp>(loc, llvmI32Type, attrValue);
+        attrCount++;
+      }
+    }
+
+    if (attrCount != 1) {
```
krzysz00: This check is why I wanted to do a DPP kind enum. Also, this sort of testing (or, with the change I want, checking of whether `permArgument` is the right type) should be in the operation verifier, not all the way down in the lowering.

krzysz00 (changes requested):
I have a highlevel question: did what I was wanting design-wise not come across well? It feels like something got lost in translation and maybe I should've provided some examples.

pfultz2:
DPP can be done with operations like `v_add` directly so `mov_dpp` should not be needed. Is the paln to support that in the future as well?

manupak:
> DPP can be done with operations like `v_add` directly so `mov_dpp` should not be needed. Is the paln to support that in the future as well?

No, we should not need to do this.

Its already a inst combine pass to do this.
Please refer : https://github.com/llvm/llvm-project/blob/a0e9a8da45402eea0db71976ab78d195a7287222/llvm/lib/Target/AMDGPU/GCNDPPCombine.cpp#L54

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
```diff
@@ -411,6 +411,50 @@ def AMDGPU_RawBufferAtomicUminOp :
...
+                  DefaultValuedAttr<I32Attr, "0xf">:$bank_mask, 
+                  DefaultValuedAttr<BoolAttr, "false">:$bound_ctrl,
+                  OptionalAttr<UnitAttr>:$wave_shl,
+                  OptionalAttr<UnitAttr>:$wave_shr,
+                  OptionalAttr<UnitAttr>:$wave_ror,
+                  OptionalAttr<UnitAttr>:$wave_rol,
+                  OptionalAttr<UnitAttr>:$row_mirror,
+                  OptionalAttr<UnitAttr>:$row_half_mirror,
+                  OptionalAttr<UnitAttr>:$row_bcast_15,
+                  OptionalAttr<UnitAttr>:$row_bcast_31)> {
```
stefankoncarevic: I have made a lot of modifications, changing the types for many attributes from bool to unit, utilized default values for attributes that are passed to the mov dpp instruction, limited possibilities for quad perm, and added a let verifier for certain checks.
With all these changes, the code now seems clearer and more optimized, hence I did not feel the need to use enums because the range of values is too wide.

external/llvm-project/mlir/lib/Dialect/AMDGPU/IR/AMDGPUDialect.cpp
```diff
@@ -326,6 +326,36 @@ LogicalResult MFMAOp::verify() {
...
+
+  if (attrCount != 1) {
+    std::string allAttributes = llvm::join(attributeNames, ", ");
+    return emitOpError("Expected exactly one of the following attributes: ")
+           << allAttributes << ", but found " << attrCount;
+  }
+
+  return success();
+}
+
```
stefankoncarevic: I have added checks for the types that are supported by the conversion and that only one attribute can be passed for mov dpp when I added the verify option.

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
krzysz00: ... No, the point of the enum is that the DPP control word can be exactly one of the options.

This is a matter of code structure and usability.

Or, to put another way, the choice of the control word should be a sum type, not a product type.

external/llvm-project/mlir/lib/Dialect/AMDGPU/IR/AMDGPUDialect.cpp
```diff
@@ -326,6 +326,36 @@ LogicalResult MFMAOp::verify() {
...
+// DPPOp
+//===----------------------------------------------------------------------===//
+LogicalResult DPPOp::verify() {
+  Type srcType = getSrc().getType();
+  if (srcType.getIntOrFloatBitWidth() > 32) {
+    return emitOpError("integer and floating point types larger than 32 bits "
+                       "are not supported");
+  }
+
+  llvm::ArrayRef<std::string> attributeNames = {
```
krzysz00: The core reason I want an enum for this is because someone who's just skimming the tablegen file will not understand that they're only allowed to pick one.

krzysz00 (changes requested):
I still have fundamental disagreements with the design.

external/llvm-project/mlir/lib/Dialect/AMDGPU/IR/AMDGPUDialect.cpp
stefankoncarevic: I understand your point, but I haven't seen any example to model an sum type that is not an plain C-like enum. Do you perhaps have one? The best way I could think of, as you see, is to enforce this constraint in validation.

krzysz00: Yeah, we're sadly going to need to enforce the constraints on the argument (if any) in `verify()` since neither C++ or MLIR have good support for tagged unions.

So I'm suggesting a C-like enum for the argument kind and then a second argument (which'd end up being an `OptionalAttr<Attribute>`) for any arguments that are relevant.

So, at the end of the day, the assembly format of the instruction would look something like
`%out = amdgpu.dpp %in row_shl(2 : i32)` or `%in quad_perm([1, 2, 3, 0])` or `%in row_mirror`

stefankoncarevic: Ok. It still seems cumbersome. Will try to support this syntax.

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -830,6 +830,151 @@ LogicalResult PackedStochRoundFp8OpLowering::matchAndRewrite(
...
+    // constants
+    auto rowMask =
+        createConstantOrUseDefault(DppOp, rewriter, "row_mask", llvmI32Type);
+    auto bankMask =
+        createConstantOrUseDefault(DppOp, rewriter, "bank_mask", llvmI32Type);
+    auto boundCtrl =
+        createConstantOrUseDefault(DppOp, rewriter, "bound_ctrl", llvmI1Type);
+
+    // create a ROCDL_DPPMovOp instruction with the appropriate attributes
+    auto dppMovOp = rewriter.create<ROCDL::DPPMovOp>(
```
krzysz00, line 952: Can we go clean up the definition of `ROCDL::DPPMovOp` to take those three attributes (row mask, bank mask, and bound control) as I32 attributes instead, to correspond to them being `immarg`s on the LLVM side? It's not critical, but since we're here, we might as well.

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -830,6 +830,151 @@ LogicalResult PackedStochRoundFp8OpLowering::matchAndRewrite(
...
+        loc, llvmI32Type, src, dppCtrl, rowMask, bankMask, boundCtrl);
+
+    Value result = dppMovOp.getRes();
+    if (srcType.getIntOrFloatBitWidth() < 32) {
+      result = rewriter.create<LLVM::TruncOp>(loc, llvmSrcIntType, result);
+    }
+
+    if (llvm::isa<FloatType>(srcType)) {
```
krzysz00: To allow for more general usage, let's make this "not an integer type", as opposed to "is a float type". That way, you can `dpp.mov` a `vector<4xi8>`

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -830,6 +830,151 @@ LogicalResult PackedStochRoundFp8OpLowering::matchAndRewrite(
...
+    auto permArgument = DppOp.getPermArgument();
+    int attrValue = 0;
+
+    switch (kind) {
+
+    case DPPPerm::quad_perm:
+      if (auto quadPermAttr = permArgument->dyn_cast<ArrayAttr>()) {
+        for (int i = 3; i >= 0; --i) {
+          auto intAttr = quadPermAttr[i].dyn_cast<IntegerAttr>();
+          int num = intAttr.getInt();
```
krzysz00: uint32_t num, probably

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -830,6 +830,151 @@ LogicalResult PackedStochRoundFp8OpLowering::matchAndRewrite(
...
+    auto kind = DppOp.getKind();
+    auto permArgument = DppOp.getPermArgument();
+    int attrValue = 0;
+
+    switch (kind) {
+
+    case DPPPerm::quad_perm:
+      if (auto quadPermAttr = permArgument->dyn_cast<ArrayAttr>()) {
+        for (int i = 3; i >= 0; --i) {
+          auto intAttr = quadPermAttr[i].dyn_cast<IntegerAttr>();
```
krzysz00: You want `cast<>` here, so it errors out when it's not integers

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -830,6 +830,151 @@ LogicalResult PackedStochRoundFp8OpLowering::matchAndRewrite(
...
+      WAVE_ROR1 = 0x13C,
+      ROW_MIRROR = 0x140,
+      ROW_HALF_MIRROR = 0x141,
+      BCAST15 = 0x142,
+      BCAST31 = 0x143,
+    };
+
+    auto kind = DppOp.getKind();
+    auto permArgument = DppOp.getPermArgument();
+    int attrValue = 0;
```
krzysz00: Yeah, this'll want to be a uint32_t.

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -830,6 +830,151 @@ LogicalResult PackedStochRoundFp8OpLowering::matchAndRewrite(
...
+
+    switch (kind) {
+
+    case DPPPerm::quad_perm:
+      if (auto quadPermAttr = permArgument->dyn_cast<ArrayAttr>()) {
+        for (int i = 3; i >= 0; --i) {
```
krzysz00: You want `for (auto elem : quadPermAttr.getAsRange<IntegerAttr>())`

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -830,6 +830,151 @@ LogicalResult PackedStochRoundFp8OpLowering::matchAndRewrite(
...
+    };
+
+    auto kind = DppOp.getKind();
+    auto permArgument = DppOp.getPermArgument();
+    int attrValue = 0;
+
+    switch (kind) {
+
+    case DPPPerm::quad_perm:
+      if (auto quadPermAttr = permArgument->dyn_cast<ArrayAttr>()) {
```
krzysz00: All these `if (auto x = dyn_cast<>())`s should be rewritten to, for example, `auto quadPermAttr = cast<ArrayAttr>(*permArgument);`. If that attribute isn't set up correctly by the time the control flow gets here, assertions should be failing loudly.

external/llvm-project/mlir/lib/Dialect/AMDGPU/IR/AMDGPUDialect.cpp
```diff
@@ -326,6 +326,62 @@ LogicalResult MFMAOp::verify() {
...
+//===----------------------------------------------------------------------===//
+LogicalResult DPPOp::verify() {
+  Type srcType = getSrc().getType();
+  if (srcType.getIntOrFloatBitWidth() > 32) {
+    return emitOpError("integer and floating point types larger than 32 bits "
+                       "are not supported");
+  }
+
+  DPPPerm kind = getKind();
+  auto permArgument = getPermArgument();
```
krzysz00: I'd explicitly write `std::optional<Attribute>` here if that's the return type.

Heck, I'd go for `Attribute permArgument = getPermArgument().value_or(Attribute{});`

external/llvm-project/mlir/lib/Dialect/AMDGPU/IR/AMDGPUDialect.cpp
```diff
@@ -326,6 +326,62 @@ LogicalResult MFMAOp::verify() {
...
+
+  switch (kind) {
+
+  case DPPPerm::quad_perm:{
+    auto quadPermAttr = permArgument->dyn_cast<ArrayAttr>();
```
krzysz00: You might be wanting `dyn_cast_or_null` here.

external/llvm-project/mlir/lib/Dialect/AMDGPU/IR/AMDGPUDialect.cpp
```diff
@@ -326,6 +326,62 @@ LogicalResult MFMAOp::verify() {
...
+
+  case DPPPerm::wave_shl:
+  case DPPPerm::wave_shr:
+  case DPPPerm::wave_rol:
+  case DPPPerm::wave_ror:
+  case DPPPerm::row_mirror:
+  case DPPPerm::row_half_mirror:
+  case DPPPerm::row_bcast_15:
+  case DPPPerm::row_bcast_31:
+
```
krzysz00: Let's defensively make sure no one's trying to pass non-trivial arguments to these (that is, if an attribute's present, it's a UnitAttr)

krzysz00 (commented):
I'm liking this! Now I'm mainly at trivialities and style observations

pfultz2:
> > DPP can be done with operations like `v_add` directly so `mov_dpp` should not be needed. Is the paln to support that in the future as well?
> 
> No, we should not need to do this.
> 
> Its already a inst combine pass to do this. Please refer : https://github.com/llvm/llvm-project/blob/a0e9a8da45402eea0db71976ab78d195a7287222/llvm/lib/Target/AMDGPU/GCNDPPCombine.cpp#L54

Probably should verify that this works, as I have never seen this work, which is why we use assembly for reductions in migraphx.

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
stefankoncarevic: @krzysz00 
Did you mean to change instead of I32 to I32Attr in ROCDL_DPPMovOp?
So instead of getting:
rocdl.mov.dpp %arg0 with %0, %1, %2, %3 : i32
getting this at the end:
rocdl.mov.dpp %arg0 with 320, 1, 1, true : i32

krzysz00: Yep

external/llvm-project/mlir/lib/Dialect/AMDGPU/IR/AMDGPUDialect.cpp
```diff
@@ -326,6 +326,64 @@ LogicalResult MFMAOp::verify() {
...
+        return emitOpError(
+            "Each element of quad_perm must be in the range [0, 3]");
+      }
+    }
+  } break;
+
+  case DPPPerm::row_shl:
+  case DPPPerm::row_shr:
+  case DPPPerm::row_ror: {
+    if (auto intAttr = permArgument.dyn_cast<IntegerAttr>()) {
```
krzysz00, line 365: Missing the case where the attribute isn't specified?

krzysz00 (approved):
Approved with minor nitpick

PR #1466, opened 2024-03-26
Using 1D arguments where possible

krzysz00:
Initial testing revealed that LLVM produces noticeably (as in, measurable performance improvements) better code when the input and output buffers are given as 1D and we explicitly have an Unmerge{} to combine the N-dimensional logical tensor indices we're using.

The code changes that I expect are helping

## Index diffs are computed on the flat index
We'd formerly do
```
%i0 = ...
%j0 = ...

...
%dI1 = ...
%dJ1 = ...
%i1 = add %i0, %dI1
%j1 = add %j0, %dJ1
%v = load %mem[%i1, %j1]
// Which is really
%tmp1_1 = mul %i1, J
%tmp2_1 = add %tmp1_1, %j1
%ptr_1 = gep %mem, %tmp2_1
%v = load %ptr_1
```

But now, with the explicit unmerge, we generate
```
%idx0 = %j0 + J * %i0
...
%dI1 = ...
%dJ1 = ...
%dI1_scaled = mul %dI1, J
%dIdx_1 = add %dI1_scaled, %dJ1
%idx1 = add %idx0, %didx_1
%v = load[%idx1]
```
which is a pattern that's much more legible to compilers (ex, you can hoist the gep for %idx0 outside the loop and the detector for that has a prayer of pattern matching)

This is especially notable in combination with
## Better handling for buffer operations

Currently, if we're trying to lower
```
%v = buffer_load %argN[%i0, %i1] if %c
```
we need to do (where L is the buffer extent)
```
%i0_if = select %c, %i0, 0
%i1_if = select %c, %i1, L
%idx = %i1_if + LEN(d1) * %i0_if
%desc = to_buffer_desc(%argN, L, ...) 
amdgpu.buffer_load %desc, %idx, ...
```

Note those selects that are inputs to arithmetic. This isn't an obvious pattern, so it doesn't codegen well.

Now, when arguments are flat, we merely have
```
%idx_if = select %c, %idx, L
%desc = ...
```
which can much more obviously be implemented as a clamp or some other such simple operation.

Examination of an interesting gemm shows that this change does actually carry all the  way through to final codegen

## What this PR does

1. Amends `rocmlir-gen` and `ConvGenerator.cpp` so that, when we generate a kernel for internal testing, the memref<>s passed in are 1-dimensional and then `rock.transform`'d to the shape that they used to have. Some care has been taken (except in attention) to ensure that the transforms have sensible dimension names
2. Changes the migraphx-to-tosa logic so that all tensor inputs and outputs that come from the MIGraphX dialect are 1-D and then `tosa.reshape`d back to their in-memory representation (which then goes to the existing logic for handling strides and slicing)
3. Removes the old C++ convolution verifier which is unused and would've been a pain to generalize.

### Impacts on tests
Built-in tests have been amended to expect 1-D memref printing when the underlying code was changed to 1D memrefs.

Existing tosa-level tests have not been 1-D-ified.

Existing Rock-level tests still work on N-D memrefs in most cases

Tests for the structure of the CPU convolution and for testing that the verifier produces expected output have been **restored** and strengthened.

## C++ verifier / -pv_with_cpp

The C++ verifier was our old infrastructure for verifying convolutions before we moved to generating the affine loop nests in rocmlir-gen. The verifier only worked for 2D convolutions.

It has been meaningfully dead (only tested by its own tests) for over a year. Refactoring that code to support the new calling convention (flat memrefs) would be a lot of work for zero gain, and thus, this commit goes and deletes the darn thing.


mlir/test/Conversion/MIGraphXToTosa/migraphx-to-rock-conv-nhwc.mlir
```diff
@@ -1,9 +1,9 @@
 // RUN: rocmlir-driver -kernel-pipeline migraphx,highlevel %s | FileCheck %s
 
 // CHECK-LABEL: @convNHWC
-// CHECK-SAME: memref<1x5x5x4xf32>
-// CHECK-SAME: memref<7x3x3x4xf32>
-// CHECK-SAME: memref<1x3x3x7xf32>
+// CHECK-SAME: memref<100xf32>
+// CHECK-SAME: memref<252xf32>
+// CHECK-SAME: memref<63xf32>
```
manupak, line 6: Why do we have to change the function signature ?
Why cant we collapse the shape as the first thing we do and use

manupak (commented):
I have not read it fully..

One thing that I did not expect is the function signature change.

Is that necessary ?  
What merit does it have over doing flattening as the first thing we do in our lowering ?

mlir/test/Conversion/MIGraphXToTosa/migraphx-to-rock-conv-nhwc.mlir
krzysz00: Because all the views are eventually written away.

The entire *point* is that the function signature - the ultimate source of truth for memrefs - is a 1D buffer.

krzysz00: Because, once everything's become coordinate transforms (or affine maps or who-knows-what), and we've computed the final index, the operations should be `load/store/... %argN[%idx]`.

Having a `memref<MxNxT>` running around as a real thing you really want to index causes the inefficiencies this change solves.

krzysz00: And the reason to do it in migraphx-to-tosa is that it makes stuff like the clone verifier behave reasonably, from where I'm standing

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2224,17 +2022,39 @@ static func::FuncOp createGpuGemmKernel(ModuleOp module,
...
+  expandedArgs.reserve(3);
+
+  ArrayRef<StringRef> dimLists[3] = {aNames, bNames, cNames};
+  for (auto [arg, t, names] :
+       llvm::zip(block->getArguments(), argTypes, dimLists)) {
+    auto logicalType = cast<MemRefType>(t);
+    rock::TopDownTMBuilder flattener(b, names, logicalType.getShape(), loc);
+    flattener.unmerge("raw", 0, names, logicalType.getShape());
+    rock::TransformMapAttr expandAttr = flattener.get();
+    expandedArgs.push_back(b.create<rock::TransformOp>(loc, arg, expandAttr));
```
giuseros: I see similar code to this used by conv2d generator and in another place in rocmlir-gen. I am wondering if there is a way to avoid duplication.

Also, I think this is kind of crucial to this PR (basically you have 1D memrefs and you provide a logical view that gets flattened, IIUC), so having this code in a function with a bit of comment would make things clearer.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2156,27 +1963,18 @@ createCPUConvFunc(ModuleOp module,
...
   }
 
   func =
       func::FuncOp::create(loc, funcName, b.getFunctionType(funcArgTypes, {}));
   module.push_back(func);
 
-  if (genValidation.getValue() == "mlir") { // -pv_with_mlir or -prc
-    createCPUConvWithMLIR(module, func, genConfig);
-  } else { // -pv_with_cpp
-    createCPUConvWithCPP(module, func, genConfig);
```
giuseros, line 2286: Is this related with the change? I mean the fact that we don't have structured memrefs allowed us to remove `createCPUConvWithCPP` (and a lot of other code)?

krzysz00: PR description updated

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
krzysz00: I think the means of getting the names are differint enough in the copies of this code that generality is being tricky.

Give ma a thought on the function signature if you've got one?

giuseros: Something on the line of:
```
// Provide a list of 1d memrefs `args`. Each `args[i]` will be reshaped like `types[i]` whose dimension's name will be contained in `names[i]`. The output will be contained in the `logicalViews` output parameter. 
void createLogicalViews(args, types, names, MutableArrayRef logicalViews){

}
```
Would that not be applicable throughout the three places?

manupak:
Cant we do the same with defining a MemRefLayoutAttrInterface that represent flattening (i.e d0, d1 -> d0*M + d1) ? (without losing the readability in the IR)

I cant still get over the fact we need to do this explicitly way higher up in our stack to get benefits. Initially, when this was proposed I thought we got the advantage of computing a single index when lowering our rock.transforms especially in the presence of padding. Because of that, I thought just flattening the memref as a view should get the same effect as padding will be done on top of the flattened memref (via rock.transforms).

manupak:
Also how much are the perf benefits ?

manupak:
https://mlir.llvm.org/docs/Dialects/Builtin/#layout

krzysz00:
@manupak Defining a layout isn't sufficient here - it still makes the flattening invisible to index diffs. That affine map will only be applied all the way down in MemRefToLLVM ... which is the problem.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
krzysz00: ... yeah, I can give it a shot

krzysz00: Though it gets fiddly with attention, perhaps

manupak:
> @manupak Defining a layout isn't sufficient here - it still makes the flattening invisible to index diffs. That affine map will only be applied all the way down in MemRefToLLVM ... which is the problem.

Can we at least carry arg attr with the original shape ?

krzysz00:
I specifically got rid of the arg attributes from when I did this as an experiment because they're clear from the transformations. Also, note that, as far as migraphx-to-tosa, the function arguments can already be weird versions of the original shapes because of strides

manupak (commented):
I see. OK.

I dont have major objections.
However i d like to see two things : 
1) rocmlir-gen option to disable this
2) perf (to kind of validate this infact is beneficial)

giuseros (approved):
LGTM, and thanks for the performance uplift!

PR #1314, opened 2023-11-09
[DO NOT SQUASH] Add a rock-check-residency pass to check occupancy/residency of a given kernel

sjw36:
This pass check various metrics collected during rock lowering and AMDGPU backend captured in the ISA metadata. 

Computes occupancy based on:
- Shared Buffer total size (rock lowering)
- VGPR allocation (amdgpu backend)
- SGPR allocation (amdgpu backend)

From these kernel requirements the rock-check-residency pass will compute blocks_per_cu occupancy within the arch target.

This PR also includes a new AttrName infrastructure to manage get/set utilities for NamedAttributes.
- Consolidated various rock NamedAttributes to use new infrastructure.

PR #1468, opened 2024-03-26
Change rock.conv2d to rock.conv, change y/x and h/w to 0/1 in layout keys, make some dimension scalars into arrays

pcf000:
Making a pull request for sanity check as much as for potential merging.

Changed "conv2d" to "conv" nearly everywhere, including the --operation flag to rocmlir-gen.  That's a set of straightforward diffs in the first commit.

Changing y/x and h/w to 0/1 and making some scalars into vectors are intermingled changes and spread over several commits, along with additional changes to make tests work.  The letters in the layout are used in several ways, and converted back and forth from layout strings to layout arrays to layout string-maps. I've tried to not affect outwardly-visible keys too much, which means (a) there are about four places that accept y/x/h/w and convert it to 0/1 and (b) the changes in tests are mostly in comparison strings and not in test commands.

I'll have a fair number of comments in the changes.  There are some places I want to highlight because they're either incomplete, will require further work to extend to 3-D, or raise questions for me.


mlir/include/mlir/Dialect/Rock/Generator/ConvGenerator.h
```diff
@@ -130,6 +130,13 @@ class ConvGenerator {
     std::transform(permutation.begin(), permutation.end(),
                    std::back_inserter(targetLayout),
                    [&targetSpec](int64_t p) { return targetSpec[p]; });
+
+    // +++pf:  update old key names.
```
pcf000, line 145: I used this same comment in all the places I have this kind of conversion code.  My intent is that we'll remove it as we finish the changeover.

mlir/include/mlir/Dialect/Rock/IR/ConvolutionDims.h
```diff
@@ -23,20 +25,19 @@ namespace rock {
 
 /// Structure for holding the dimensions of a convolution problem
 struct ConvolutionDims {
-  int64_t y;
-  int64_t x;
-  int64_t ho;
-  int64_t wo;
-  int64_t hi;
-  int64_t wi;
+  SmallVector<int64_t, 5> fil;
```
pcf000: These guys become arrays, which affects both how ConvolutionDims is constructed by its users and how the data is extracted and passed on from the struct.

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
```diff
@@ -50,23 +50,23 @@ class Rock_Attr<string name, list<Trait> traits = []> :
     AttrDef<Rock_Dialect, name, traits>;
 
 /// ConvOpType
-def ConvOp2DType : I32EnumAttrCase<"Fwd", 0, "conv2d">;
-def ConvOp2DBwdDataType : I32EnumAttrCase<"BwdData", 1, "conv2d_bwd_data">;
-def ConvOp2DBwdWeightType : I32EnumAttrCase<"BwdWeight", 2, "conv2d_bwd_weight">;
+def ConvOpType : I32EnumAttrCase<"Fwd", 0, "conv">;
```
pcf000, line 53: The heart of the conv2d-to-conv name change.  I could locate most of the ripples by making this change, compiling, and seeing what broke.

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
```diff
@@ -194,10 +194,7 @@ class BasePopulateParams {
       if (this->padding_amount < rhs.padding_amount) {
         return true;
       } else if (this->padding_amount == rhs.padding_amount) {
-        if (this->original_pos < rhs.original_pos) {
-          return true;
-        }
-        return false;
+        return (this->original_pos < rhs.original_pos);
```
pcf000, line 214: clangd pointed this one out to me.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -175,27 +173,14 @@ makeRockConv2D(ConversionPatternRewriter &rw, Operation *op, Value input,
   rock::GemmFeatures features;
   std::tie(arch, num_cu, features) = getArchAttributes(op, input.getType());
 
-  ArrayRef<int64_t> pad64 = pad;
```
pcf000, line 172: If we're getting an attribute array, we don't need to make intermediate arrays, to make another attribute array.

mlir/lib/Dialect/Rock/Generator/ConvGenerator.cpp
```diff
@@ -186,9 +186,9 @@ LogicalResult ConvGenerator::hasValidDimension() const {
   auto outDim = canonicalizeDims(config.outputDimension, config.outputLayout);
 
   // Note: hasDimensions() prints error messages
-  if (failed(hasDimensions(inDim, "ngchw", "input")) ||
-      failed(hasDimensions(filDim, "gkcyx", "filter")) ||
-      failed(hasDimensions(outDim, "ngkhw", "output"))) {
+  if (failed(hasDimensions(inDim, "ngc01", "input")) ||
```
pcf000, line 185: First update of layout strings.  Can't have something like "ngchw01" because the test cares about the length.

mlir/lib/Dialect/Rock/Generator/ConvGenerator.cpp
```diff
@@ -215,37 +215,37 @@ LogicalResult ConvGenerator::hasValidDimension() const {
   }
 
   int64_t expectedOutHeight = outputDim(
-      inDim["h"], filDim["y"], config.paddingLeftDims[DIM::HEIGHT],
+      inDim["0"], filDim["0"], config.paddingLeftDims[DIM::HEIGHT],
```
pcf000: inDim and filDim are StringMaps;  the key strings change.  DIM::HEIGHT is part of an enum defined in ConvGenerator.h, because I didn't want to hard-code 0 and 1 everywhere.

mlir/lib/Dialect/Rock/Generator/ConvGenerator.cpp
```diff
@@ -654,30 +654,38 @@ ConvGenerator::parseConvDims(int64_t batchSize, int64_t groupSize,
                              int64_t filterHeight, int64_t filterWidth) {
   config.filterDims[DIM::HEIGHT] = filterHeight;
   config.filterDims[DIM::WIDTH] = filterWidth;
-  static const std::string filterKeys = "kgcyx";
```
pcf000, line 658: The logic here bothered me.  Keys and Vals are parallel arrays -- given a layout key character, we'll find() it in the Keys array (string) to get an index, and use that index to look up the value in Vals.  Why not just a StringMap to associate the key and value directly?

mlir/lib/Dialect/Rock/Generator/ConvGenerator.cpp
```diff
@@ -654,30 +654,38 @@ ConvGenerator::parseConvDims(int64_t batchSize, int64_t groupSize,
...
-      static std::string nchw = "ngchw";
-      ii = nchw.find(keyl);
-      if (ii == std::string::npos)
+
+  llvm::StringMap<int64_t> filterMap = {{"k", outputChannel / groupSize},
+                                        {"g", groupSize},
+                                        {"c", inputChannel / groupSize},
+                                        {"y", filterHeight},
+                                        {"x", filterWidth},
+                                        {"0", filterHeight},
```
pcf000: Both old and new keys are present in the map, for now.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -421,8 +421,12 @@ ConvolutionDims ConvolutionDims::fromOp(Operation *op) {
 
     if (filterAttr.getValue() == "y") {
       y = filterShape[i];
+    } else if (filterAttr.getValue() == "0") {
```
pcf000, line 425: Another place that accepts both old and new keys.  I thought I could just convert old forms to new at a few entry points, but then there are places that read conv ops from a file and there's no place to convert before they call functions like fromOp().  It surprised me how much of Operation construction is completely generic.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -443,19 +451,23 @@ ConvolutionDims ConvolutionDims::fromOp(Operation *op) {
...
       wo = outputShape[i];
+    } else if (outputAttr.getValue() == "0o") {
+      ho = outputShape[i];
+    } else if (outputAttr.getValue() == "1o") {
+      wo = outputShape[i];
     }
   }
 
-  return ConvolutionDims(y, x, ho, wo, hi, wi, k, c, n, g);
+  return ConvolutionDims({y, x}, {ho, wo}, {hi, wi}, k, c, n, g);
```
pcf000, line 462: Because ConvolutionDims has arrays for those dimensions, we can aggregate them in the constructor call here.

The logic here will need brushing up to extend to 3-D.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -489,14 +501,14 @@ GemmSize GemmSize::fromConvolution(ConvOpType type,
   case ConvOpType::Fwd:
     gemmGSize = sizes.g;
     gemmMSize = sizes.k;
-    gemmKSize = sizes.c * sizes.y * sizes.x;
-    gemmNSize = sizes.n * sizes.ho * sizes.wo;
+    gemmKSize = sizes.c * sizes.fil[0] * sizes.fil[1];
```
pcf000, line 506: References to arrays now.  Not sure how some of these expressions should be extended for 3-D.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -555,8 +567,8 @@ static LogicalResult verifyConvOp(RockConvInterface convOp) {
     return (pos2 != pos1 + 1) && (pos1 != pos2 + 1);
   };
 
-  if (isDisjointed("filter_layout", "y", "x") ||
-      isDisjointed("input_layout", "hi", "wi"))
+  if ((isDisjointed("filter_layout", "y", "x") && isDisjointed("filter_layout", "0", "1")) ||
```
pcf000: Accepting both old and new.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -655,10 +667,10 @@ GemmSize Conv2DBwdDataOp::getGemmSize() {
   int64_t yTilda = strideH / gcdStrideDilationH;
```
pcf000: Have not turned some of these variables into arrays and thus have no loops for the dimensions.

mlir/lib/Dialect/Rock/IR/TransformMapBuilder.cpp
```diff
@@ -260,7 +260,10 @@ StringRef TransformMapBuilder::endName(uint32_t dim) {
 }
 
 uint32_t TransformMapBuilder::startIndex(StringRef name) {
-  assert(startIndices.count(name) == 1 && "Key not in starting set of names");
+  if (startIndices.count(name) != 1) {
+    llvm::errs() << "Key not in starting set of names: " << name << "\n";
```
pcf000: I wanted to see *which* key wasn't in the set.

mlir/lib/Dialect/Rock/Tuning/ConvContext.cpp
```diff
@@ -23,15 +23,26 @@ populateDimIndexAndSize(const ArrayAttr &layoutAttr,
   size_t dimValSize = layoutAttr.size();
   for (size_t i = 0; i < dimValSize; ++i) {
     auto key = layoutAttr.getValue()[i].cast<StringAttr>().getValue();
+
+    // +++pf: update old keys.
+    if (key == "y") key = "0";
+    if (key == "x") key = "1";
+    if (key[0] == 'h')
+      key = StringAttr::get(layoutAttr.getContext(),
```
pcf000, line 33: It is a pain in the butt to take a string attribute and replace a character.

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
```diff
@@ -66,6 +66,9 @@ LogicalResult mlir::rock::calculateKBlockNum(const int64_t batchSize,
 
   int64_t gemmKBlock = 1;
 
+  assert(gemmM > 0 && gemmN > 0 && gemmK > 0);
```
pcf000, line 72: One common failure mode was to divide by zero down here.

mlir/test/Dialect/Rock/conv_to_gemm_layout_regularization.mlir
```diff
@@ -1,20 +1,20 @@
-// RUN: rocmlir-gen --arch gfx908 -p -operation conv2d -fil_layout gkyxc -in_layout ngchw -out_layout hwgkn \
+// RUN: rocmlir-gen --arch gfx908 -p -operation conv -fil_layout gkyxc -in_layout ngchw -out_layout hwgkn \
```
pcf000, line 1: Changed the operation but not the layout, but changed the expected layout keys in the transform maps.  A bit inconsistent.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -793,6 +793,14 @@ static void correctConvParameters() {
   else if (inputLayoutValue.size() == 4)
     inputLayout = "g" + inputLayoutValue;
 
+  // +++pf:  update old key names.
```
pcf000, line 857: rocmlir-gen will accept either old or new keys, for now.

mlir/include/mlir/Dialect/Rock/IR/ConvolutionDims.h
krzysz00: I think you'll only need 3 or 4 here, since n, k, c, and g are held separately

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -1767,24 +1775,53 @@ createCPUConvWithMLIR(ModuleOp module, func::FuncOp &func,
...
       else
         layout = genConfig.outputLayout;
+      auto direction = genConfig.operation.value();
       for (auto c : layout) {
-        auto direction = genConfig.operation.value();
         if ((direction == rock::ConvOpType::Fwd ||
              direction == rock::ConvOpType::BwdWeight) &&
             tensor == INPUT) {
-          if (c == 'h') {
+          if (c == '0') {               // +++pf: may need adjustment to h/w.
```
pcf000: This whole section needs to be reworked.  It's connecting loopIVs by finding the right letter in the layout string, and needs to have both y/x and h/w present, which causes a collision since 0/1 could mean either.  At the moment, the string will have both 0/1 and y/x, but the better solution is to use a vector of strings like "0i" to allow for unambiguous keys.

There's also logic to override the key match in certain cases, which is necessary to make some test cases work but is confused a bit by the key renaming.  The set of changes here works but is not the final answer.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2000,6 +2037,8 @@ static void createCPUConvWithCPP(ModuleOp module, func::FuncOp &func,
   layoutConstOps['n'] = nConstantOp;
   layoutConstOps['h'] = hConstantOp;
   layoutConstOps['w'] = wConstantOp;
+  layoutConstOps['0'] = yConstantOp;
```
pcf000, line 2152: Need to accept both old and new keys.

mlir/lib/Dialect/Rock/Generator/ConvGenerator.cpp
krzysz00: This is a nice cleanup

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
krzysz00: These expressions extend naturally - they're the product of the array.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
krzysz00: Is this correct? It feels like the arms of the `&&` are mutually exclusive

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
krzysz00: These _should_ generalize out nicely? But that's a future PR, I think

mlir/lib/Dialect/Rock/IR/TransformMapBuilder.cpp
krzysz00: I think we've got `report_fatal_error` down in LLVM for this sort of thing?

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
krzysz00: And, right, this is a different file and we're probably looking at a struct that could be unified away

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -87,6 +87,12 @@ template <typename T>
 LogicalResult checkNames(ArrayRef<StringRef> actual,
                          ArrayRef<StringRef> expected, StringRef argName,
                          T op) {
+//   printf("checkNames({");
+//   for (auto name : actual) printf("%s,", name);
+//   printf("}, {");
+//   for (auto name : expected) printf("%s,", name);
```
krzysz00: Stray comments (that could be `LLVM_DEBUG()`?)

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -87,6 +87,12 @@ template <typename T>
 LogicalResult checkNames(ArrayRef<StringRef> actual,
                          ArrayRef<StringRef> expected, StringRef argName,
                          T op) {
+//   printf("checkNames({");
```
pcf000: Oops, debug code remaining.

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -134,22 +140,41 @@ LogicalResult getConvDimNames(T op, SmallVectorImpl<StringRef> &filterNames,
...
   outputNames.reserve(size);
 
+  auto update_old_name = [](StringAttr name)
+                         {
+#if 1
+                           auto ctx = name.getContext();
+                           if (name == "y") return StringAttr::get(ctx, "0");
+                           if (name == "x") return StringAttr::get(ctx, "1");
+                           if (name.strref().starts_with_insensitive("h")) {
+                             auto namestr = name.str();
```
krzysz00: Do you need the `.str()` here. And does that return a StringRef or? 

(and, on the third hand, does the string attr constructor take `Twine`?)

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -474,10 +499,14 @@ struct MatchLayoutsToInput final
     TypedValue<ShapedType> filter = op.getFilter(), output = op.getOutput();
     const llvm::StringMap<StringAttr> inputToFilter = {
         {"ci", b.getStringAttr("c")},
+        {"0i", b.getStringAttr("0")},
```
pcf000: Need both old and new forms.  Without this change, resnet18_blk.mlir has an RMS of 1e-3 instead of 1e-7.

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -598,7 +621,7 @@ LogicalResult backwardWeightAtomicAdd(Conv2DBwdWeightOp op,
     // is what's left
     llvm::StringMap<uint32_t> firstTransformOutDims = expandNamesInPlace(
         inputNames,
-        {{"ni", {"n0", "n1"}}, {"hi", {"hipad"}}, {"wi", {"wipad"}}});
+        {{"ni", {"n0", "n1"}}, {"0i", {"hipad"}}, {"1i", {"wipad"}}});
```
krzysz00: We'll likely want to refactor here so we get `0i_pad` , `1i_pad` and so on.

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -534,16 +563,10 @@ LogicalResult backwardWeightAtomicAdd(Conv2DBwdWeightOp op,
...
-  int64_t leftPadW = ctx.getPaddingVal()[2];
-  int64_t rightPadH = ctx.getPaddingVal()[1];
-  int64_t rightPadW = ctx.getPaddingVal()[3];
-
-  int64_t dilationH = ctx.getDilationVal()[0];
-  int64_t dilationW = ctx.getDilationVal()[1];
-  int64_t strideH = ctx.getStrideVal()[0];
-  int64_t strideW = ctx.getStrideVal()[1];
+  // Obtain convolution parameters: padding / dilation / stride.
+  auto pads = ctx.getPaddingVal();
```
pcf000, line 571: Intermediate arrays from the arrays in ConvolutionDims.

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -661,7 +683,7 @@ LogicalResult backwardWeightAtomicAdd(Conv2DBwdWeightOp op,
     firstWrap.passThrough("go");
     firstWrap.unmerge({"n0", "n1"}, "no",
                       {gemmKBlocks, convDims.n / gemmKBlocks});
-    firstWrap.passThrough({"ko", "ho", "wo"});
+    firstWrap.passThrough({"ko", "0o", "1o"});
```
pcf000: How does something like translate to 1-D and 3-D?

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -635,9 +657,9 @@ LogicalResult backwardWeightAtomicAdd(Conv2DBwdWeightOp op,
     auto gemmInputTransform =
         BottomUpTMBuilder::above(embedTransform, embedTransformAttr);
 
-    llvm::SmallVector<StringRef, 3> nonNHWDims = {"ci", "y", "x"};
+    llvm::SmallVector<StringRef, 3> nonNHWDims = {"ci", "0", "1"};
     matchUnderlyingOrder(nonNHWDims, gemmInputTransform);
-    llvm::SmallVector<StringRef, 3> nhwDims = {"n1", "ho", "wo"};
+    llvm::SmallVector<StringRef, 3> nhwDims = {"n1", "0o", "1o"};
```
krzysz00: Future work will be generating this list with a `for (i : numberOfImageDimensions)` and some calls to `Twine`.

... Heck, while we're here, we might go ahead and extend `TransformMapBuilder` to let it work on `StringAttr`s transparently.

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -775,15 +797,15 @@ LogicalResult backwardData(Conv2DBwdDataOp op, PatternRewriter &b) {
     // Embed y/x into {y/x}dot and {y/x}tilda (Why the
     // particular embed coefficients is in a presentation somewhere)
     llvm::StringMap<uint32_t> embedDims = expandNamesInPlace(
-        filterNames, {{"y", {"ydot", "ytilda"}}, {"x", {"xdot", "xtilda"}}});
+        filterNames, {{"0", {"0dot", "0tilda"}}, {"1", {"1dot", "1tilda"}}});
```
pcf000: Need to also make a loop that generates the "Ndot" and "Ntilda" strings, and can then make the tilda and slice scalars into arrays and make this section into loops.

mlir/lib/Dialect/Rock/Tuning/ConvContext.cpp
krzysz00: You shouldn't need to `std::string` here. `Twine()` should do you?

mlir/test/Conversion/RockToGPU/multiple_kernels.mlir
```diff
@@ -13,22 +13,22 @@
 
 module  {
   func.func @step1(%arg0: memref<1x512x16x3x3xf32>, %arg1: memref<512x1x16x32x32xf32>, %arg2: memref<512x1x512x30x30xf32>) attributes {kernel, arch = "amdgcn-amd-amdhsa:gfx906"} {
-    rock.conv2d(%arg0, %arg1, %arg2) features = dot {arch = "amdgcn-amd-amdhsa:gfx906", dilations = [1 : index, 1 : index], filter_layout = ["g", "k", "c", "y", "x"], input_layout = ["ni", "gi", "ci", "hi", "wi"], output_layout = ["no", "go", "ko", "ho", "wo"], padding = [0 : index, 0 : index, 0 : index, 0 : index], strides = [1 : index, 1 : index]} : memref<1x512x16x3x3xf32>, memref<512x1x16x32x32xf32>, memref<512x1x512x30x30xf32>
+    rock.conv(%arg0, %arg1, %arg2) features = dot {arch = "amdgcn-amd-amdhsa:gfx906", dilations = [1 : index, 1 : index], filter_layout = ["g", "k", "c", "y", "x"], input_layout = ["ni", "gi", "ci", "hi", "wi"], output_layout = ["no", "go", "ko", "ho", "wo"], padding = [0 : index, 0 : index, 0 : index, 0 : index], strides = [1 : index, 1 : index]} : memref<1x512x16x3x3xf32>, memref<512x1x16x32x32xf32>, memref<512x1x512x30x30xf32>
```
krzysz00: Let's go ahead and output the quoted strings here? Just so we have canonical-form IR in our tests.

mlir/test/Conversion/TosaToRock/no-mixed-layout.mlir
```diff
@@ -1,6 +1,6 @@
 // RUN: rocmlir-driver -host-pipeline partition,highlevel -targets %arch %s | FileCheck %s
 
-// CHECK: rock.conv2d({{.*}}) {{.*}} {{{.*}}, filter_layout = ["g", "k", "c", "y", "x"], input_layout = ["ni", "gi", "ci", "hi", "wi"], output_layout = ["no", "go", "ko", "ho", "wo"]{{.*}}}
+// CHECK: rock.conv({{.*}}) {{.*}} {{{.*}}, filter_layout = ["g", "k", "c", "y", "x"], input_layout = ["ni", "gi", "ci", "hi", "wi"], output_layout = ["no", "go", "ko", "ho", "wo"]{{.*}}}
```
krzysz00, line 3: Is updating TosaToRock a future PR?

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
krzysz00: This we might as well keep around for ergonomics

krzysz00 (commented):
Overall, seems fine, though I'd start to hitting our tests with the big search-replace hammer

mlir/test/Conversion/RockToGPU/multiple_kernels.mlir
pcf000: Not sure what you mean by "output the quoted strings";  something like "ngchw" instead of the list of strings?

mlir/test/Conversion/TosaToRock/no-mixed-layout.mlir
pcf000: I consider non-incidental Tosa stuff to be part of issue #1314.

pcf000:
Should I do the big test search-and-replace for this pull request?

krzysz00:
@pcf000 I'd prefer if we did

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -111,7 +111,7 @@ static Value expandTensor(ConversionPatternRewriter &rw, Operation *op,
 static std::tuple<StringAttr, std::optional<uint32_t>, rock::GemmFeatures>
 getArchAttributes(Operation *op, Type inputType) {
   auto func = op->getParentOfType<func::FuncOp>();
-  auto mod = func->getParentOfType<ModuleOp>();
+  //auto mod = func->getParentOfType<ModuleOp>();
```
krzysz00: ... this shouldn't be dead code, I think we can still have cases where the module is `arch`'d and the function isn't?

mlir/lib/Dialect/Rock/Generator/ConvGenerator.cpp
krzysz00: I think these are all loop-computable, subject to the awkward need to `formatv` or `Twine()` your loop variable.

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
```diff
@@ -132,18 +133,42 @@ mlir::rock::backwardDataKernelIds(int64_t strideHeight, int64_t strideWidth,
...
+    default:
+      llvm_unreachable("Only 2-D and 3-D have been implemented.");
+      break;
+    case 3:
+      divisor = filTilda[2];
+      iTilda[2] = kernelId % divisor;
+      [[fallthrough]];
+    case 2:
+      iTilda[1] = (kernelId % subproduct) / divisor;
+      iTilda[0] = kernelId / subproduct;
```
krzysz00, line 158: I think this can be generalized to N dimensions. 

... heck, you probably want `delinearize` from `external/llvm-project/mlir/include/mlir/Dialect/Utils/IndexingUtils.h`
, which makes it `iTilda = delinearize(kernelID, filDims);`

mlir/test/Dialect/Rock/integration/reduce/blockwise_reduce/blockwise_reducemax_nr_threads_lt_blocksize.mlir
```diff
@@ -1,11 +1,11 @@
 // RUN: rocmlir-gen -ph -print-results -rand fixed - < %s | rocmlir-driver -arch %arch -c  | mlir-cpu-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_float16_utils%shlibext --entry-point-result=void | FileCheck %s
```
krzysz00, line 1: I think you hit an overzealous sed on this file and can revert

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -607,25 +634,41 @@ LogicalResult backwardWeightAtomicAdd(Conv2DBwdWeightOp op,
...
     // The usual mapping of input space to dimensions such that filter elements
     // get multiplied by the right thing
-    llvm::StringMap<uint32_t> embedOutDims = expandNamesInPlace(
-        firstTransform, {{"hipad", {"y", "ho"}}, {"wipad", {"x", "wo"}}});
+    expansions.clear();
+    for (size_t i = 0; i < convDims.out.size(); i++) {
+      StringAttr key = b.getStringAttr(Twine(i) + "ipad");
+      StringAttr val1 = b.getStringAttr(Twine(i));
+      StringAttr val2 = b.getStringAttr(Twine(i) + "o");
+      expansions.insert({key, {val1, val2}});
```
krzysz00, line 665: Huh, does `StringMap` not own its keys?

krzysz00 (commented):
This here's coming along well (haven't really stared at all the exact details but they're probably good).

Do let me know when you think we've hit a point where hitting merge and moving to a followup PR is reasonable.

mlir/lib/Dialect/Rock/Generator/ConvGenerator.cpp
pcf000: Commit 4d3dff009 does that, along with some places where, eg, llvm::any_of() is more concise than std::any_of().

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
pcf000: I didn't write it, I just got tired of the unused-variable warning.

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
pcf000: The delinearize() function is computing something slightly different, which is giving me different results in practice.  Could be I need to call it differently, or could be that our needs just differ.

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
pcf000: Nope, it maps StringRefs to things.  I even witnessed strings changing out from under the StringMap.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -213,52 +213,82 @@ LogicalResult ConvConverter<ConvType>::matchAndRewrite(
...
+    cop = rewriter.create<tosa::Conv3DOp>(
+        loc, newOutTy,
+        ValueRange{
+            input, filter,
+            getZeroTensor(
+                loc, outputTy.getElementType(),
+                filter.getType().template cast<ShapedType>().getShape()[0],
+                rewriter)});
+    break;
+  default:
```
krzysz00, line 260: ... Add 1D?

mlir/lib/Dialect/Rock/Generator/ConvGenerator.cpp
```diff
@@ -499,36 +484,35 @@ LogicalResult ConvGenerator::parseConvConfig(OpBuilder &builder,
...
         "batchsize",   "groupsize",    "in_layout", "in_type",
         "in_channels", "in_h",         "in_w",      "out_layout",
         "out_type",    "out_channels", "out_h",     "out_w",
         "fil_layout",  "fil_type",     "fil_w",     "fil_h"};
-    if (!std::all_of(validKeys.cbegin(), validKeys.cend(),
-                     [&argMap](const std::string &key) {
-                       return argMap.count(key) > 0;
-                     })) {
-      return false;
+    if (argMap["in_layout"].length() > 5) { // Ie, 3-D.
```
krzysz00, line 492: Might be nice to have a second vector of keys to `all_of` on for the 3D case but that's me.

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
krzysz00: Could be, but yeah, this is a loop

```
SmallVector<int64_t> iTilda(0, filterDims.size());
int64_t remainingKernelId = kernelId;
for (auto [thisITilda, dim] : llvm::zip_equal(llvm::reverse(iTilda, filterDims)) {
  thisITilda = remainingKernelId % dim;
  remainingKernelId /= dim;
}
```

krzysz00: Which should be exactly `delinearize(kernelId, filterDims);`

krzysz00 (approved):
I think MIGraphX also wanted 1D convolution, but also,  I can't see any major issues here.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
pcf000: Not sure what to rewrite it into, since there isn't a tosa::Conv1DOp.

pcf000:
Replaced by #1505, which has been merged.

PR #1469, opened 2024-03-27
Modify the regularize pass for input fusion

yiqian1:
Fix https://github.com/ROCm/rocMLIR-internal/issues/1360


mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -836,6 +840,11 @@ LAGenericRewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
   // operation that reads from the output of this generic. If there is such an
   ThreadwiseReadIntoOp tileReadOp;
   if (!gemmStoreOp) {
+    if (!out.getDefiningOp<memref::AllocOp>()) {
+      LLVM_DEBUG(llvm::dbgs()
+                 << "generic output is not suitable for input fusion\n");
+      return success();
```
krzysz00: This one's a `notifyMatchFailure` and returning a `failure()`

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
```diff
@@ -196,26 +202,33 @@ struct PushTransformsUpRewritePattern
...
       Operation *useOp = use.getOwner();
       SmallVector<Operation *> chain;
       bool isWriter = collectChain(buffer, useOp, chain);
+
+      bool isFusor = isFusorOp(chain.back());
       if (isWriter) {
         assert(writer == nullptr);
         writer = useOp;
+      }
+      if (isFusor) {
```
krzysz00: I'm seeing a correctness issue, it goes like this

```
%alloc1 = alloc() : T
rock.gemm %alloc1 = ...
%alloc2 = alloc() : T
linalg.generic ins (%alloc1, ...) outs(%alloc2) { ... }
%v = view(%alloc2) : T to U
memref.copy %v, %argN : U to U
```

The regularization invariant for output fusions says this must be
```
%alloc1 = alloc() : T
gemm %alloc1 = ...
%alloc2_new = alloc() : U
%alloc2_like_old = [inverse view](%alloc2_new) : U to T
la.generic ins(%alloc1, ...) outs(%alloc2_like_old) { ... }
memref.copy %alloc2_new, %argN : U to U
```

Your definition of `fusor` won't cause the rewrite to fire like it used to.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -628,8 +629,8 @@ makeExtraInputTile(LinalgAlignRewriter &b, TiledOp tiledOp, Value src,
   // block of tiledOp, as their arguments may depend on that control flow. Later
   // we will also move linalg.generic into the block.
   if (laGeneric->getBlock() != tiledOp->getBlock() &&
-      laGeneric.getOperation() != sink)
-    b.setInsertionPoint(sink);
+      isa<ThreadwiseReadIntoOp>(tiledOp))
```
krzysz00, line 632: I can't help but feel like the logic we're actually looking here involves actually computing the CFG or dominator tree and asking "is the generic in a block that's outside of the one the tiled op is in" or a similar question.

That being said, this here's probably fine.

krzysz00 (commented):
I think this'll do, or at least let us make progress.

We still have the vectorizaition question, which is a whole can of worms.

And I suspect that the fusor/fusee analysis will get more generic once buffer dependency analysis lands.

Or at least I still am not convinced that `Regularize` shouldn't be working its way inside-out from the gemms to the associated generics and keeping track of whether something's an input or output fusion along the way.

But I don't think this breaks anything and it does maybe probably handle the typical cases so ... let me think about this a bit more (maybe all this needs is telling the pattern rewriter that you're meant to go bottom-up or top-down)

mlir/test/fusion/input-output-fusion-regularize.mlir
```diff
@@ -0,0 +1,47 @@
...
+      %c0 = arith.constant 2.0 : f32
+      %r = arith.addf %in, %c0: f32
+      linalg.yield %r : f32
+    }
+    %cp = rock.transform %alloc_1 by #transform_map2 : memref<1x32x32xf32> to memref<1x32x32xf32>
+    // CHECK: memref.copy %[[ALLOC_1]]
+    memref.copy %cp, %arg2 : memref<1x32x32xf32> to memref<1x32x32xf32>
+    return
+  }
+}
```
krzysz00, line 47: Update: yeah, this is the general case, so since this works as expected I'll approve

krzysz00 (approved):

PR #1471, opened 2024-04-02
Tune for MFMAs using `mnkPerXdl` + 16x16x4xf16 support.

manupak:
This PR fundamentally adds the capability sweep through all types of mfmas present in our table.
Additionally it adds 16x16x4xf16 entries to the table.

It is  `mnkPerXdl` is defined as :  `mnPerXdl` * 100 +  `kPerXdl` (making it somewhat human readable) to keep a single axis to sweep the mfmas.
So to keep it dense as opposed to defining two axes as `mnPerXdl` , `kPerXdl` where some points in the cross
product may only be valid. Moreover, it is easy to keep using single axis that is juxtaposing with 'nPerWave' with
WMMA while structures like `InitParamAccel' exists -- which IMO should be replaced with an AttrInterface in some point.

Moreover, it fills 'holes' present in the current set of keys in the table where it is possible to codegen
using MFMAs. 

Additionally, adds tests to cover the MFMA table such that every entry can validly generate code w/o
breaking.

Perf Analysis is ongoing (still running -- I ll share shortly) -- though after doing all the fixes, the sdxl perf seems to be flat and 16x16x4xf16 do not seem to bring much benefit. However, the code changes hopefully bring ease when adding experimenting with a new MFMAs in the future.

closes : https://github.com/ROCm/rocMLIR-internal/issues/1412
closes : https://github.com/ROCm/rocMLIR-internal/issues/1409


mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -28,36 +28,92 @@ namespace rock {
...
-                    cast<RockTuningParamAttrInterface>(params));
+  GemmFeatures features = attnOp.getFeatures();
+  if (bitEnumContainsAll(features, GemmFeatures::mfma)) {
+    // PopulateParamsXDL tuningInfo;
+    static const std::vector<std::vector<uint32_t>> validRangeAccelGemmParams =
+        {{4, 8, 16, 32, 64, 128, 256}, // gemmMPerBlock
+         {4, 8, 16, 32, 64, 128, 256}, // gemmNPerBlock
+         {2, 4, 8, 16, 32, 64},        // gemmKPerBlock
+         {16, 32, 64, 128, 256},       // gemmMPerWave
+         {401, 402, 404, 1601, 1602, 1604, 1608, 1616, 1632, 3201, 3202, 3204,
```
manupak, line 39: This axis is dynamically sparse due to only certain types of MFMAs are valid per each dtype.
I may make that explicit here.

manupak:
[gfx90a_16x16x4_compare.zip](https://github.com/ROCm/rocMLIR/files/14852988/gfx90a_16x16x4_compare.zip)

Here is the perf -- as I said it turned out flat as well.

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
```diff
@@ -198,73 +199,144 @@ auto getMfmaInsnGroupAttrMapAllArch = []() -> const MfmaInsnGroupMap & {
...
+          {{MfmaTypeId::Fp16TyId, 16, 4, 16, 64},
+           {ROCDL::mfma_f32_16x16x4f16::getOperationName(),
+            {{2, 0, MFMAPermB::none}}}}
+          //  {{MfmaTypeId::Fp16TyId, 16, 4, 64, 32},
+          //   {ROCDL::mfma_f32_16x16x4f16::getOperationName(), {{0, 0,
+          //   MFMAPermB::bcast_first_16}, {0, 0,
+          //   MFMAPermB::bcast_second_16}}}},
+          //  {{MfmaTypeId::Fp16TyId, 16, 4, 64, 16},
+          //   {ROCDL::mfma_f32_16x16x4f16::getOperationName(), {{0, 0,
+          //   MFMAPermB::bcast_first_16}}}}
```
giuseros, line 311: Spurious comments?

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
```diff
@@ -329,7 +330,10 @@ def Rock_XdlopsGemmDerivedParamsAttr : Rock_Attr<"XdlopsGemmDerivedParams", [Roc
       int64_t mPerWave = params.getMPerWave();
       int64_t mPerBlock = params.getMPerBlock();
       int64_t nPerBlock = params.getNPerBlock();
-      int64_t mnPerXdl = params.getMnPerXdl();
+      int64_t mnkPerXdl = params.getMnkPerXdl();
+      int64_t mnPerXdl = mnkPerXdl / 100;
+      int64_t kPerXdl = mnkPerXdl % 100;
+
```
giuseros, line 336: Not a strong opinion, but what about a "mn x k" string to represent different mfmas?

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
```diff
@@ -198,73 +199,144 @@ auto getMfmaInsnGroupAttrMapAllArch = []() -> const MfmaInsnGroupMap & {
...
+           {ROCDL::mfma_f32_4x4x4f16::getOperationName(),
+            {{4, 0, MFMAPermB::none}, {4, 1, MFMAPermB::none}}}},
+          {{MfmaTypeId::Fp16TyId, 4, 4, 8, 64},
+           {ROCDL::mfma_f32_4x4x4f16::getOperationName(),
+            {{4, 0, MFMAPermB::none}, {4, 1, MFMAPermB::none}}}},
+          {{MfmaTypeId::Fp16TyId, 4, 4, 4, 64},
+           {ROCDL::mfma_f32_4x4x4f16::getOperationName(),
+            {{4, 0, MFMAPermB::none}}}},
+          // f16 - 16x16x16
+          {{MfmaTypeId::Fp16TyId, 16, 0, 64, 64},
```
giuseros, line 270: Isn't this clashing with the other `{MfmaTypeId::Fp16TyId, 16, 0, 64, 64}` key ? Also, why do we need  a "default" case? I mean, in which circumstance it gets hit?

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -394,137 +395,179 @@ PopulateParamsAccel::obtainTuningParameters(RockGemmWrapperInterface op,
...
+  {32, 32, 4, 16, 1604, 8, 1, true, true},
+  {16, 64, 8, 16, 1604, 8, 1, true, true},
+  {16, 64, 8, 16, 1604, 4, 1, true, true},
+  {16, 32, 8, 16, 1604, 8, 1, true, true},
+  {16, 32, 8, 16, 1604, 4, 1, true, true},
+  {16, 16, 8, 16, 1604, 8, 1, true, true},
+  {16, 16, 4, 16, 1604, 4, 1, true, true},
+  {8, 64, 8, 4, 401, 8, 1, true, true},
+  {4, 64, 16, 4, 401, 1, 1, true, true},
+  {4, 64, 2, 4, 401, 8, 1, true, true}
```
krzysz00, line 438: ... could we just go ahead and declare `v3` instead of pulling this parameter packing trick?

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -394,137 +395,179 @@ PopulateParamsAccel::obtainTuningParameters(RockGemmWrapperInterface op,
...
+  {16, 64 ,4, 16, 1616, 8, 1, true, true},
+  {16, 32, 4, 16, 1616, 8, 1, true, true},
+  {16, 32, 4, 16, 1616, 4, 1, true, true},
+  {16, 16, 4, 16, 1616, 8, 1, true, true},
+  {8, 64, 8, 8, 404, 8, 1, true, true}
 };
 
 const InitParamsAccel
 PopulateParamsXDL::initParametersForward8Bit[
   PopulateParamsXDL::nInitParametersForward8Bit] = {
```
krzysz00, line 490: Do we have a justification for making the int8 quick tuning list twice as long? Paul's really not going lo like it and for good reason.

krzysz00 (changes requested):
(I'm leaving a blocker to prevent this landing by accident)

Given that
1. The packing of m, n, and k per xdl into one tuning parameter is ugly given that we *just* defined an extension mechanism for perf configs
2. This makes some very important quick tuning lists double the length without good perf justification and
3. That clash Giuseppe mentioned

I don't think we should land this the way it is, especially since it's not good for perf.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
manupak: Its not twice as long practically.
only half of it is valid for pre-MI300 and half of it valid for post-MI300.

manupak: (I can make two lists and pick based on arch)

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
manupak: We can... 

Just want to make sure -- you are asking for MFMA specific perfconfig format ?
i.e. v3mfma: xx, xx, ... , xx .

Then make v3wmma: xx, xx, ... , xx . to be seperate ? 
And also remove common abstractions like InitParamAccel ?

manupak:
> * The packing of m, n, and k per xdl into one tuning parameter is ugly given that we _just_ defined an extension mechanism for perf configs

Agreed, kind of resorted to avoid two different perfconfig formats for wmma and mfma.
But we dont have to.

> * This makes some very important quick tuning lists double the length without good perf justification and

No it does not. Its just we had a single list for pre-MI300 and post-MI300 i8 mfmas that picked different mfma s.

> * That clash Giuseppe mentioned

I can remove the default easily. just had that in the process of development to not to worry about the arch when wanting to pick a default config.

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
manupak: yes and no -- for some reason I got page faults only in attention kernels but worked for gemms.
I can remove them for now.

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
manupak: I think @krzysz00 wants to have a seperate perfconfig format for MFMAs.
Which also means we have to remove InitParamAccel like structures that are common between WMMAs and MFMAs.

giuseros: But why they need to be different? I mean wmma would be a 16x16 kind of config. Why having a separate string?

manupak: I have not studied the effects on WMMA oriented tuning to also bring the same kind of change to WMMAs.
i.e. fixing 4 waves etc, deriving nPerWave

Which to me seems like task of same propotion that needs to be conducted.
Until that is conducted, I thought of making this PR not affecting WMMA perf whatsoever.

manupak:
If we are going down the path of seperation of perf configs for MFMA vs WMMA (A1)  OR unifying the tuning between MFMA/WMMA (A2) -- then more pre-requsite work has to be done.

For A1) that would first need decouple perfConfig parsing for WMMAs vs MFMAs and remove common abstractions (i.e. InitParamAccel) first.
For A2) that would need a study to identifying bringing nPerWave inferencing mechanism is useful in WMMA context -- then do the similiar changes as MFMAs...

Therefore, if either of these are recommended as the course we want to take .. I ll probably close this PR for now as such effort cannot be really justified. I resorted to this 'hashing' mechanism to avoid all that.

mlir/lib/Dialect/Rock/IR/MfmaInsnGroup.cpp
manupak: yes it is and it was intentional -- just had the entry if some MFMA was needed to be picked w/o worrying about kPerXdl which can differ based on the arch. I can remove this entry if its confusing.

manupak:
Im closing this PR on the basis this is not worthwhile taking in *now*.

I ve created the pre-requisite ticket to decouple MFMAs and WMMAs : https://github.com/ROCm/rocMLIR-internal/issues/1447.
I think they fairly distinct to diverge in the tuning space in the future. So that when we truly have a use case that demands a seperate tuning space for either, it would be possible do so w/o doing a yet another refactor then and there.

PR #1472, opened 2024-04-03
Update sdxl configs

manupak:
This commit adds sdxl configs for
wider visbility of the team.


mlir/utils/performance/sdxl-conv-configs
```diff
@@ -0,0 +1,10 @@
+# python3 ./bin/tuningRunner.py --op conv --tuning-space=quick --verify-mode=none --config='<config>'
+convfp16 -F 1 -f GNCHW -I NGCHW -O NGCHW -n 2 -c 1280 -H 64 -W 64 -k 640 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -g 1
```
krzysz00: Do these actually need to specify NCHW? We might want to benchmark their NHWC versions.

krzysz00 (approved):
Approved though with some curiosity about the layout restriction

mlir/utils/performance/sdxl-conv-configs
manupak: make sense. removed them now.

PR #1473, opened 2024-04-03
[Attention] Increase upperbound of second gemm MPerBlock

manupak:
Empherically, it was found that there is slightly a bit more room to increase the imposed upperbound of second gemm's MPerBlock which is clamped or set to be gemm1M.

This helps larger head_dim cases.


krzysz00 (approved):
LGTM, thanks for noticing this sort of thing

manupak:
This is not needed anymore .. actually its not beneficial to increase mPerBlock when we can now iterate on M dimension: https://github.com/ROCm/rocMLIR/pull/1475

PR #1474, opened 2024-04-04
Adopt fp16 packed arithmetic

giuseros:
Closes https://github.com/ROCm/rocMLIR-internal/issues/1438

What this PR does:
- Once we convert linalg to affine, we vectorize the affine loops (this only will enable the backend to emit packed fp16 arithmetic)
- When we meet the `arith.trunc fp32 -> fp16` we replace that with the intrinsic `llvm.amdgcn.cvt.pkrtz`


giuseros:
I couldn't run the tests on the mi300x machine due to a cpu bug when we use the cpu-runner, but I expect we will have to lower the precision for some of the fp16 cases

external/llvm-project/mlir/include/mlir/Conversion/Passes.td
```diff
@@ -98,18 +98,23 @@ def ConvertAMDGPUToROCDL : Pass<"convert-amdgpu-to-rocdl"> {
...
+    done in two steps in order to avoid running a notional arith-to-rocdl and arith-to-llvm
     simultaniously.
   }];
 
   let dependentDialects = ["amdgpu::AMDGPUDialect", "vector::VectorDialect"];
 
   let options = [
+    Option<"convertFP8Arithmetic", "convert-fp8-arithmetic", "bool",
+           /*default=*/"true",
+           "Whether truncation to 8-bit float types should be saturating">,
```
krzysz00: Note: make this guy take `chipset` so you can switch on intrinsic availability for the fp8 conversions instead?

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
```diff
@@ -90,6 +90,24 @@ def AMDGPU_PackedTruncFp8x2Op :
   let hasVerifier = 1;
 }
 
+def AMDGPU_PackedTruncFp16x2Op :
```
krzysz00: I think you honestly don't need this one and can go for the rocdl intrinsic directly here - though I might go for `ub.poison` instead of LLVM's version

external/llvm-project/mlir/lib/Conversion/ArithToAMDGPU/ArithToAMDGPU.cpp
```diff
@@ -254,17 +264,90 @@ void TruncfToFloat8RewritePattern::rewrite(arith::TruncFOp op,
...
+      return failure();
+    if (outVecType.getShape().size() > 1)
+      // Multi-dimensional vectors are currently unsupported.
+      return failure();
+    outType = outVecType.getElementType();
+
+    // If the output type is a vector, we can assume the input type is a vector
+    // too
+    auto inputVecType = inputType.dyn_cast<VectorType>();
+    inputType = inputVecType.getElementType();
```
krzysz00: This one's `getElementTypeOrSelf()`

mlir/lib/Dialect/Rock/Transforms/VectorizeFusions.cpp
```diff
@@ -0,0 +1,125 @@
...
+        [](affine::AffineForOp nestedLoop) -> WalkResult {
+          return WalkResult::interrupt();
+        });
+
+    if (containsOtherLoops.wasInterrupted()) {
+      return WalkResult::advance();
+    }
+
+    // Only vectorize loops whose parent is the kernel itself
+    if (loop->getParentOp() != op.getOperation()) {
```
krzysz00: I'm not sure I'd go for this condition - it'll miss input fusions

mlir/lib/Dialect/Rock/Transforms/VectorizeFusions.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+void RockVectorizeFusionsPass::runOnOperation() {
+  func::FuncOp op = getOperation();
+  IRRewriter b(op.getContext());
+  op.walk([&](affine::AffineForOp loop) -> WalkResult {
+    // Try to isolate fusion loops
+
+    // Only vectorize loops without nested loops
+    WalkResult containsOtherLoops = loop.getLoopBody().walk(
+        [](affine::AffineForOp nestedLoop) -> WalkResult {
```
krzysz00: I'd go and make this a reasonably strict check - does the loop only contain `affine.load`, `affine.store`, `arith.*` or `math.*`

krzysz00: and so you can slide all the other data collection into the one walk through the loop body.

mlir/lib/Dialect/Rock/utility/AccelEmitter.cpp
```diff
@@ -72,7 +72,7 @@ void AccelEmitter::computeOutputConversion(PatternRewriter &b, Location loc,
       loc, ArrayRef<ValueRange>{{zeroConstantOp}, {zeroConstantOp}},
       ArrayRef<Attribute>{b.getArrayAttr({}), b.getArrayAttr(toRegCScalarAttr)},
       /*bounds=*/ArrayRef<int64_t>{mRepeats * nRepeats * nResultVectors},
-      /*strides=*/std::nullopt, forceUnroll, /*useIndexDiffs=*/true);
+      /*strides=*/std::nullopt, false, /*useIndexDiffs=*/true);
```
krzysz00: ... we probably want to delete support for type conversions here since they always should be going through the fusion flow

krzysz00 (commented):
Overall, good stuff, just nitpicks

mlir/lib/Dialect/Rock/Transforms/VectorizeFusions.cpp
giuseros: THe `arith.*/math.*` is problematic, I think. I guess the best thing is to leave this as a light check, and let Affine decide if the loop can be vectorized or not.

mlir/lib/Dialect/Rock/utility/AccelEmitter.cpp
giuseros: This seems an orthogonal refactor (which I agree with). Also, I am reverting to the original `forceUnroll`. This was a spurious test I did when I started working on this (to see if naively not unrolling the loop might have been beneficial)

giuseros:
Hi @krzysz00 a friendly reminder about this :)

external/llvm-project/mlir/include/mlir/Conversion/Passes.td
```diff
@@ -741,12 +746,12 @@ def ConvertMemRefToSPIRV : Pass<"convert-memref-to-spirv"> {
 def ConvertNVVMToLLVMPass : Pass<"convert-nvvm-to-llvm"> {
   let summary = "Convert NVVM dialect to LLVM dialect";
   let description = [{
-    This pass generates inline assembly for the NVVM ops which is not 
+    This pass generates inline assembly for the NVVM ops which is not
```
krzysz00, line 749: You'll probably want to revert this off as noise

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
```diff
@@ -196,14 +198,15 @@ void rock::buildBackendPipeline(OpPassManager &pm,
...
-    ArithToAMDGPUConversionPassOptions options;
-    options.saturateFP8Truncf = true;
-    gpuPm.addPass(createArithToAMDGPUConversionPass(options));
+  ArithToAMDGPUConversionPassOptions arithOptions;
+  arithOptions.chipset = options.chip;
+  arithOptions.allowPackedF16Rtz = true;
+  arithOptions.saturateFP8Truncf = true;
+  gpuPm.addPass(createArithToAMDGPUConversionPass(arithOptions));
+  if (archInfo.hasFp8ConversionInstrs)
     gpuPm.addPass(createArithToAMDGPUConversionPass());
```
krzysz00: Do we need this if statement anymore? We might want the else branch, but that's about it.

mlir/lib/Dialect/Rock/Transforms/VectorizeFusions.cpp
```diff
@@ -0,0 +1,132 @@
...
+        loop.getLoopBody().walk([&loopTypes](Operation *op) -> WalkResult {
+          if (auto affineLoad = dyn_cast<affine::AffineLoadOp>(op)) {
+            if (getAddressSpace(affineLoad.getMemref()) ==
+                AddressSpace::Private)
+              loopTypes.push_back(affineLoad.getType());
+          } else if (auto affineStore = dyn_cast<affine::AffineStoreOp>(op)) {
+            if (getAddressSpace(affineStore.getMemref()) ==
+                AddressSpace::Private)
+              loopTypes.push_back(affineStore.getMemRefType().getElementType());
+          } else if (dyn_cast<affine::AffineForOp>(op)) {
```
krzysz00: Nit: This is spelled `isa<>`

But also, we can
a) Iterate the immediate children of the loop instead of doing a `walk()`
b) We're probably wanting a dialect check because if it's not in `arith` or `math`, we shouldn't be vectorizing it.

mlir/lib/Dialect/Rock/Transforms/VectorizeFusions.cpp
```diff
@@ -0,0 +1,132 @@
...
+    LLVM_DEBUG(llvm::dbgs()
+               << "With vector length = " << vectorizationFactor << "\n");
+    DenseSet<Operation *> loops{loop};
+    affine::vectorizeAffineLoops(op, loops,
+                                 SmallVector<int64_t>{vectorizationFactor}, {});
+
+    // Make sure the transfer reads are in bounds
+    return WalkResult::advance();
+  });
+  op.walk([&](vector::TransferReadOp op) {
```
krzysz00: There's a widget something like `getFilteredOps` that saves us the recursion of `walk()` which we don't need

krzysz00 (approved):
Minor nitpicking and questions, but approved with those

mlir/lib/Dialect/Rock/Transforms/VectorizeFusions.cpp
giuseros: I didn't know you could do a `isa<dialect>` check. Good to know, thanks!

mlir/lib/Dialect/Rock/Transforms/VectorizeFusions.cpp
giuseros: I used `getOps<opType>`, but the vectorization is destructive, hence I still have to walk through the affineForOps (and then filter out the transfer{read,writes})

PR #1475, opened 2024-04-04
[Attention] Iterate on gemm1MBlocks

manupak:
Currently if gemm1MBlocks > 1 (which is rare) but is present in larger head_dim cases (i.e. 512),
we assign a single block in gemm1MBlocks  to a seperate workgroup. This result in duplication of
first gemm and softmax for such splits.

This change instead of assigning a different workgroup. this will iterate on gemm1MBlocks  in the same
workgroup. 

This shows speed ups of 3x for f32 case and 67% for f16 case that is of interest to SDXL work.
closes : https://github.com/ROCm/rocMLIR-internal/issues/1306


krzysz00 (commented):
I'm not entirely sure what's going on here - I'll take another look tomorrow, seeing as it's kinda late for me.

I feel the urge to ask - could that slice creation be phrased using `rock.extract_multibuffer()`?

manupak:
> I'm not entirely sure what's going on here - I'll take another look tomorrow, seeing as it's kinda late for me.

Thanks! it just changes the second gemm to iterate on M dimension rather than assignining different workgroups basically -- as I described in the PR description.

>  could that slice creation be phrased using rock.extract_multibuffer()?

'rock.extract_multibuffer' requires one to have a set of buffer objects and select one as op output which to me makes the IR bloated for no good reason here. I also need to flatten/collapse them to be able to have a single global tw_write_all afterwards and having to deal with multiple buffers makes it ugly.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1918,15 +1952,12 @@ struct GridwiseAttentionAccelRewritePattern
             loc, reverseMap, ValueRange{mLoopIV, mIterationsGemm0Val});
       }
       zeroAccBuffer(rewriter, loc, accRegBufferGemm0);
-#ifndef ROCK_DEBUG_ATTENTION_REMOVE_SOFTMAX
-      zeroAccBuffer(rewriter, loc, accRegBufferGemm1);
-#endif
       // The grid coordinates for gemm0 is almost simliar
```
giuseros: Is this comment still relevant?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2165,109 +2196,196 @@ struct GridwiseAttentionAccelRewritePattern
...
+          {numElementsAttnOut}, attentionOutAccBufferOutTypedElType,
+          AffineMap{}, privateMemoryAddressSpace);
+      auto reassociation =
+          getReassociationForFlattening(attentionOutAccBufferOutType);
+      attentionOutAccBufferOutTypedFlat =
+          rewriter.create<memref::CollapseShapeOp>(
+              loc, attentionOutAccBufferOutTypedFlatType,
+              attentionOutAccBufferOutTyped, reassociation);
+    }
+    // This map with create upper view [gblock, nblock, flatiter] -> [gblock,
```
giuseros: Is this a typo? "This map will create an upper view ..."

giuseros (approved):
Very (very) minor comments. I think I understand enough about what is going on to approve this!

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
manupak: no, removed it. thanks!

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
manupak: thanks for catching this.

PR #1476, opened 2024-04-05
Move rock-pipeline pass after rock-linalg-align to avoid redundant allocation

yiqian1:
Closes https://github.com/ROCm/rocMLIR-internal/issues/1437


giuseros (approved):
LGTM

krzysz00 (approved):
Yeah, this looks like it'll do.

Though I'd like @giuseros for if perhaps any extra argument reads needed for fusion should go in their own `stage{}` - though that's probably a future PR.

PR #1479, opened 2024-04-11
Avoid collapsing a dimension multiple times

giuseros:
It might happen that a singleton dimension can be reused multiple times generating larger groups and messing the collapse logic.


mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -360,8 +360,11 @@ void findCountiguousGroupsUnmerge(const ArrayRef<uint32_t> upperDims,
       uint32_t fastestDim = groupCandidate.back();
       size_t fastestDimPosInMerge = dimPosition.back();
       size_t slowestDimPosInMerge = dimPosition.front();
-      for (auto d : groupCandidate)
+      for (auto d : groupCandidate){
```
krzysz00: Nit: spacing on this brace (probably is a clang-format fail anyway but still)

krzysz00 (approved):
Good catch, nice bugfix, approved!

PR #1480, opened 2024-04-12
Filter out invalid WMMA perf configs (ex. m/nPerWave < 16)

krzysz00:
If mPerWave or nPerWave are less than  the input length of the WMMA, we can't cohenetly codegen usages of those WMMAs.

Similarly, while it wasn't an issue in this bug, if mPerWave or nPerWave aren't evenly divisible by the wmma's input length, reject that config as well.

Fixes https://github.com/ROCm/rocMLIR-internal/issues/1444

Move the tests

### DIFF
--- a/mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
+++ b/mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
@@ -54,6 +54,11 @@ FailureOr<WmmaInsn> WmmaInsn::select(mlir::Type elementTypeA,
   int64_t outLen = 8;
   int64_t outStride = 2;
 
+  if (mPerWave % inputLen != 0)
+    return failure();
+  if (nPerWave % inputLen != 0)
+    return failure();
+
   int64_t mRepeats = mPerWave / inputLen;
   int64_t nRepeats = nPerWave / inputLen;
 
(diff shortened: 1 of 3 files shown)


giuseros (approved):
LGTM!

ravil-mobile:
Hi @krzysz00 @giuseros, the CI fails at the same spot as mine CAPI PR fails.

```
Tested 9400 configs, best perf 0.13165714285714286 TFlops on perf_config v2:16,16,4,16,16,4,24,1,1

Traceback (most recent call last):

  File "./bin/tuningRunner.py", line 404, in <module>

    sys.exit(main())

  File "./bin/tuningRunner.py", line 374, in main

    winners, allData = tuneMLIRKernels(configs, confClass, paths, options)

  File "./bin/tuningRunner.py", line 159, in tuneMLIRKernels

    winningConfig, maxTFlops = getWinningConfig(tuningLoop.stdout, config, allData, options)

  File "./bin/tuningRunner.py", line 107, in getWinningConfig

    perfConfig, time = result.split('\t')

ValueError: not enough values to unpack (expected 2, got 1)

script returned exit code 1
```

It is either something with the runner or we need to patch something in our code. I suspect that there may be some problems with the multi-threading inside LLVM. In my case, I couldn't deterministically reproduce the error

manupak (approved):
Great catch! Thanks @krzysz00 !

PR #1484, opened 2024-04-17
Address accuracy issues introduced by fp16 packed arithmetic

giuseros:
Fixes https://github.com/ROCm/rocMLIR-internal/issues/1466

### DIFF
--- a/mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
+++ b/mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
@@ -484,10 +484,13 @@ void mcpuVerify(T *gpuResults, T *validationResults, long long dataSize,
 
     if (valNum == gpuNum) {
       hist_relDiff[0]++;
-    } else if (std::isinf(valNum)) {
-      // Let's not compare gpu results to infs
-      continue;
     } else {
+      // We know valNum != gpuNum. If valNum is inf, this branch will simply
+      // return nan. Let's instead represent infinite with max<fp16> and let's
+      // test for it
+      constexpr float fp16MaxVal = 65504;
+      if (std::isinf(valNum))
+        valNum = (valNum > 0 ? fp16MaxVal : -fp16MaxVal);
       float absDiff = fabs(valNum - gpuNum);
       // Update maxAbsDiff and its correspinding pair of values
       if (absDiff > maxAbsDiff) {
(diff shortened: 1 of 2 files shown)


manupak (dismissed):
LGTM on the basis that it is an accepted trade off to use packed f16 arithmetic!

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
```diff
@@ -484,6 +484,9 @@ void mcpuVerify(T *gpuResults, T *validationResults, long long dataSize,
 
     if (valNum == gpuNum) {
       hist_relDiff[0]++;
+    } else if (std::isinf(valNum)) {
```
manupak: If the validation result == inf, then we accept any error ?

manupak: I think solution should be in fp16, we should replace it with fp16 inf (== -65505) as valNum ?

krzysz00: I think the solution is entirely different: define the semantics of fp16 truncation to be saturating CPU-side.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2885,9 +2885,15 @@ static func::FuncOp createVerifierFunc(ModuleOp module, const KernelIF &kernel,
...
 
   if (testElemType.isa<FloatType>()) {
-    auto thr_RMS = getF32Val(RMSThreshold.getValue());
+    constexpr float defaultRMSThreshold(0.00003f);
+    constexpr float defaultRMSThresholdFP16(0.001f);
+    float RMSThresholdValue =
+        testElemType.isF16() ? defaultRMSThresholdFP16 : defaultRMSThreshold;
+    if (RMSThreshold)
+      RMSThresholdValue = RMSThreshold.getValue();
+    auto thr_RMS = getF32Val(RMSThresholdValue);
```
krzysz00: Can we not `auto` this?

mlir/lib/ExecutionEngine/conv-validation-wrappers.cpp
krzysz00: ... Wait no, I see yet another solution: when we generate FP16 verification kernels, explicitly set the rounding mode to `rtz` around the conversions

krzysz00 (commented):
I've got Thoughts here - maybe we land this anyway to unbreak CI but still

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
giuseros: What do you mean `auto`?

krzysz00: You're using `auto` in a context where it's easy and helps clarity to write the type out.

(The existing code used `Value` for one of these constants, and you even changed it to `auto`)

krzysz00 (approved):

PR #1486, opened 2024-04-17
Prevent calling LICM on removed loops

krzysz00:
Hopefully fix the crashes-after-a-while issues that hit CI.

Also add a missing #include to ArithToAMDGPU

PR #1487, opened 2024-04-18
Fixing include string issue in mlir Conversion files

urpetkov-amd:
Fixing compilation issue on Windows caused by including string which resulted in blockage of Windows build.


giuseros:
This has been already fixed here: https://github.com/ROCm/rocMLIR/pull/1486

PR #1490, opened 2024-04-19
Fix memory leaks in rocmlir-tuning-dirver, rocmlir-gen

krzysz00:
1. Move to OwningOpRef<ModuleOp> in rocmlir-tuning-driver and rocmlir-gen to prevent missed delete calls that cause memory leaks
2. When we create a tunig space, wrap it an a unique_ptr to ensure it doesn't leak.
3. Update certain tuning APIs to not take ModuleOp by reference
4. Update some functions that populate existing modules in rocmilr-gen to return void.


manupak (approved):
Great catch! LGTM!

PR #1493, opened 2024-04-24
Add support for update.dpp intrinsic

stefankoncarevic:


external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
stefankoncarevic, line 1: I was guided by this
https://github.com/llvm/llvm-project/blob/c1086532d4d5c0a261457dfb00e79fcb764e3d78/llvm/lib/Target/AMDGPU/VOP1Instructions.td#L1339
to only use update dpp.

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
```diff
@@ -461,7 +462,7 @@ def AMDGPU_DPPOp : AMDGPU_Op<"dpp", [SameTypeOperands, AllTypesMatch<["result",
   }];
   let results = (outs AnyType:$result);
   let assemblyFormat = [{
-    $src $kind (`(` $permArgument^ `)`)? attr-dict `:` type($result)
+    $src (`(` $old^ `:` type($old) `)`)? $kind (`(` $permArgument^ `)`)? attr-dict `:` type($result)
```
krzysz00:
```
src (`,'^ $old)?
```

Also, if you wait for upstream merge to land first, you'll get access to `OptionalTypesMatchWith`, so you don't have to repeat the type.

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
krzysz00: ... that's not what that says.

That's two rewrite patterns.

Can you please point me to some specific indication that the mov.dpp intrinsic shouldn't be used?

krzysz00 (changes requested):
I'd like to see evidence of this alleged deprecation

stefankoncarevic:
> I'd like to see evidence of this alleged deprecation

On this link https://github.com/llvm/llvm-project/pull/89233
from the upstream ticket there is a comment from Matt, that mov dpp is deprecated intrinsic, and should be replaced with update.dpp.
The link I posted above in the comment only gives an insight into how mov.dpp can be replaced with update.dpp to behave the same. As for me, I can separate in ROCDLOps.td that we have both mov.dpp and update.dpp so that we can choose which one we want to use.

krzysz00 (commented):
Per comments upstream, let's do what Matt said and always have an `old` argument and always use `update_dpp`.

However, while you're doing that, can you add some stuff to the documentation on mov.dpp - either in the LLVM documentation or in comments in `IntrinsicsAMDGPU.td` and maybe a deprecation on the builtin in clang? That'll be a separate commit but let's keep it in the same PR here.

krzysz00 (approved):
Approved!

PR #1496, opened 2024-04-26
Extract MIGraphX model subgraphs and create MLIR test cases

yiqian1:
Update the perl script to extract unique MLIR modules from a MIGraphX trace and generate test cases to verify that they can be successfully compiled.


mlir/utils/widgets/extractModulesCreateTest.pl
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env perl
```
krzysz00, line 1: ... hold on, do we want to merge this into the existing script, maybe? Or does this want to be its own thing?

mlir/utils/widgets/extractModulesCreateTest.pl
```diff
@@ -0,0 +1,99 @@
...
+            $toPrint = 0;
+        }
+        else {
+            # Extract the arch string
+            my $arch="%arch";
+            if ($_ =~ /arch\s*=\s*"([^"]*)"/) {
+                $arch = $1;
+            }
+            open($segment, ">:encoding(UTF-8)", "${prefix}_${count}.mlir") or die $!;
+            print $segment "// RUN: rocmlir-driver -kernel-pipeline migraphx,highlevel %s | rocmlir-gen -ph -print-results -rand none - | rocmlir-driver -arch $arch -c\n";
```
krzysz00: Also, I don't think this is the right test command.

You want `-verifier clone` and some `mlir-cpu-runner` in there?

mlir/utils/widgets/extractModulesCreateTest.pl
yiqian1: This script will replace the existing one. It includes the feature of the existing script. I'll remove the old one.

mlir/utils/widgets/extractModulesCreateTest.pl
yiqian1: These tests are for topological validation. So we've decided to just verify the extracted subgraphs can be compiled successfully, and not to run E2E tests.

mlir/test/migraphx_models/resnet50-v1-7-tests/resnet50-v1-7_1.mlir
```diff
@@ -0,0 +1,9 @@
+// RUN: rocmlir-driver -kernel-pipeline migraphx,highlevel %s | rocmlir-gen -ph -print-results -rand none - | rocmlir-driver -kernel-pipeline full  -arch gfx1100 --verify-passes | rocmlir-opt
```
krzysz00, line 1: Are we sure these shouldn't be e2e tests?

krzysz00 (approved):
The script looks fine but I'm really dubious about the point of these test cases

mlir/test/migraphx_models/resnet50-v1-7-tests/resnet50-v1-7_1.mlir
yiqian1: Yeah, I think compilation tests are OK for now. If we opt to have e2e tests later, It would be straightforward to change the script to do that.

PR #1497, opened 2024-04-29
[CI] Bump jenkinsfiles to use rocm6.1

manupak:


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -71,7 +71,7 @@ String dockerImage() {
 }
```
jerryyin, line 71: I think line 70 also needs change.

jerryyin (approved):
Could you address the comment as well as adding `mlir/utils/jenkins/Jenkinsfile.release`?

I'm giving approval to reduce back and forth

mlir/utils/jenkins/Jenkinsfile
manupak: done!

manupak:
Thanks for catching that

manupak:
So 6.1 data layout mismatch is still there -- I may have been comparing with a newer rocm that is being used in SDXL.
Additionally rocm 6.1 turns out not to have a clang-format binary shipped with it -- which is another problem.

Created a new PR to solve the issue differently : https://github.com/ROCm/rocMLIR/pull/1498.

manupak:
@jerryyin I ve now switched to base clang-format in our docker for now to get things going.

I think we need to do a bump our Dockerfile to match https://github.com/google/llvm-premerge-checks/blob/main/containers/buildbot-linux/Dockerfile in the next release. I d suggest we do not
do this *now* just for clang-format.

https://github.com/ROCm/rocMLIR-internal/issues/1476.

PR #1503, opened 2024-05-03
[DO NOT SQUASH] Fix build warnings

krzysz00:
1. Fix deprecated uses of `.isa<>` on `Value`s and anything else that pattern-matched in those files
2. Comment out the deprecation of `SerializeToHsaco` so we stop getting yelled at about it.


giuseros (approved):
LGTM

ravil-mobile (approved):

PR #1505, opened 2024-05-06
[DO NOT SQUASH] Rebased and compressed version of #1468

pcf000:
I couldn't figure out how to split pull request #1468 into two or more that would still run, and the recent main-line issues didn't help.  I rebased on develop to get past the conflicts from the upstream merge and I combined the commits into four of roughly useful contents.  Otherwise, it's the same as #1468.  I didn't change #1468 itself because I was using an experimental script for the rebase.

Do-not-squash to preserve the few separate commits.


mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -214,52 +214,82 @@ LogicalResult ConvConverter<ConvType>::matchAndRewrite(
   SmallVector<int64_t> NCHW2NHWC{0, 2, 3, 1};
   SmallVector<int64_t> NHWC2NCHW{0, 3, 1, 2};
 
+  int dims = outputTy.getShape().size() - 2;
```
krzysz00, line 217: Nit: `size_t`? Or `uint32_t`?

krzysz00 (approved):

PR #1509, opened 2024-05-08
Removed an out-of-date if-statement

ravil-mobile:


krzysz00 (approved):
s/redundant/out-of-date/ in the commit message, approved.

Note that might go in the commit message but is put here: general GEMM used to havne a different writeback operation that didn't support atomic add, this hasn't been the case for quite some time but we forgot to remove the guard (is the current theory of this PR)

ravil-mobile:
> s/redundant/out-of-date/ in the commit message, approved.

> Note that might go in the commit message but is put here: general GEMM used to havne a different writeback operation that didn't support atomic add, this hasn't been the case for quite some time but we forgot to remove the guard (is the current theory of this PR)

Thanks for the review!
Done

PR #1510, opened 2024-05-10
Updated the data layout string to sync with the current LLVM version.

stefankoncarevic:
I attempted to reproduce the warnings on gfx908 and gfx90a machines using the provided run line:
rocmlir-gen -operation attention -t f16 --arch gfx908:sramecc+:xnack- --num_cu 110 -g 768 -seq_len_q 384 -seq_len_k 384 -head_dim_qk 64 -head_dim_v 64 -with-attn-scale=False -transQ=False -transK=True -transQ=False -transK=False --kernel-repeats 5 --perf_config= | rocmlir-tuning-driver
but was unsuccessful. I updated the data layout in LowerGPUOpsToRocdl.cpp to match the one in clang/lib/Basic/Targets/AMDGPU.cpp and adjusted the test accordingly. Is there a way I can test this without making changes to get the warning message, to ensure the data layout is appropriate after add the change for address space 9?


krzysz00 (approved):
1. You won't see the warnings, we suppressed them
2. Approved
3. You'll need to `git clang-format develop` this because your string constant isn't correctly formatted

stefankoncarevic:
Is there an efficient way to validate the embedded dataLayout string to ensure it always stays in sync with the LLVM TargetMachine? Also, if there's a recommended approach to centralize this string to facilitate its maintenance and updates, I would appreciate any suggestions. If not, then let it stay like this for now.

PR #1514, opened 2024-05-14
Fix tuning-driver from crashing when tuning fused modules

manupak:
Currently the tuning-driver crashes if we a give a fused module to tune. This is because split-k applicability is not tested in applicability pipeline rather in the compilation pipeline. We dont want do AlignTiling in applicability pipeline either because it will fail compilation pipeline if it was run again on a module that has done AlignTiling (maybe this needs investigating/fixing in a future PR).

Therefore this commit, adds fusionLegality check on top of applicability check for tuning-driver to make it not crash when tuning fused modules.


ravil-mobile (commented):
Thanks @manupak. It looks good to me. I would also ask @krzysz00 to double check it.

krzysz00 (approved):
Seems fine as a fix

PR #1515, opened 2024-05-14
Expose fusability checks to rocmlir-gen

manupak:
This commit moves the implementation of fusability
checks from CAPI to tuningImpl mainly because all
the depedendent functions are there and make CAPI
a thin wrapper of this feature.

This allows us to verify fusability requests coming
from client and easily tests things from our side.


krzysz00 (approved):
Seems like a good refactor, let's go with it

PR #1516, opened 2024-05-14
Trim down ROCMLIR_EXTENSION_LIBS

djramic:
* Removed unused libraries
* Replaced registerAllExtensions with the necessarly extensions only


krzysz00 (approved):
Approved, lgtm, thanks!

krzysz00:
... you'll probably need to regenerate the static library dependency list though

PR #1518, opened 2024-05-16
Make tuning-driver get arch from func or mod

manupak:
Currently, the tuning-driver has a requirement
that the module needs to have the arch.
However, our clients may only set the arch
in the func.

This commit adds support to get get arch from
func or mod before failing the tuning-driver.

### DIFF
--- a/mlir/tools/rocmlir-tuning-driver/rocmlir-tuning-driver.cpp
+++ b/mlir/tools/rocmlir-tuning-driver/rocmlir-tuning-driver.cpp
@@ -22,6 +22,7 @@
 #include "mlir/Dialect/Rock/Pipelines/Pipelines.h"
 #include "mlir/Dialect/Rock/Tuning/RockTuning.h"
 #include "mlir/Dialect/Rock/utility/fusionUtils.h"
+#include "mlir/Dialect/Rock/utility/loweringUtils.h"
 #include "mlir/ExecutionEngine/RocmDeviceName.h"
 #include "mlir/IR/AsmState.h"
 #include "mlir/IR/BuiltinAttributes.h"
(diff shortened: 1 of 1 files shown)


krzysz00 (approved):
Seems reasonable to me.

krzysz00 (changes requested):
I don't think this is an accurate description of the changes: this now only gets `arch` from functions.

Can you keep the original logic?

That is,

1. If `mhal.arch` is set on the module, use that, otherwise
2. Walk the modules's functions to find an `mhal.arch`

manupak:
> If mhal.arch is set on the module, use that, otherwise
Walk the modules's functions to find an mhal.arch

Why does module arch take precedence ? 
or better yet do we even support if there is a mismatch ?

manupak:
Also does tuning-driver support multi-kernel in single module scenario ?

manupak:
>  this now only gets arch from functions.

no, this gets from func, if not module.
func takes precedence but also assumes there is one kernel func per module. hence the previous question

krzysz00:
@manupak  Yeah, multiple kernels in one module is supported, see backwards data.

And I'd want module arch taking precedence because it's easier to get to.

manupak:
>  I'd want module arch taking precedence because it's easier to get to.

FYI, This is not how we have been doing in the rest of the code base. See https://github.com/ROCm/rocMLIR/pull/1439/files#diff-f03bc66ec82c54629193060fb61d5427604f0f9842eeafe2d18ac9b08ad56aeb (even before my changes on 1439). and I d like it done same way thoughout the codebase. Practically, we never encountered a case where there was a conflict rather we encounter where one is missing (func vs module). 

To be clear, Im not opposing module taking preference; I just want to make sure we do it same throughout.
"because it's easier to get to" is not a good reason; here alternative is not that hard.

krzysz00 (approved):
Yeah, I see your point, approved

PR #1519, opened 2024-05-16
[AMDGPU] Extend llvm.amdgcn.update.dpp intrinsic to support f64 (#91190)

stefankoncarevic:
This is the PR needed for update dpp to support f64.


manupak (approved):
LGTM assuming that this is just a cherry pick

PR #1520, opened 2024-05-16
Minor infrastructure fixes factored out from flat memrefs

krzysz00:
- Make the tuning runner take problem configs to their canonical form to prevent suprious matches
- Add the printing options to rocmlir-gen and rocmlir-driver, enabling --mlir-print-local-scope to be used in more of our tools.


manupak (approved):
I quite like this change. 
Now we can have tests with proper labels w/o hoisted global attrs

PR #1521, opened 2024-05-16
Refactor the host convolution harness to make it N-dimensional

krzysz00:
Update the logic in the MLIR affine map based convolution test code to make it work correctly on inputs of arbitrary length in preparation for N-D convolution support in rocmlir-gen if we decide we want that.

(While we're here, make the upper bounds in the affine conditions the constants that they are instead of symbols for no good reason)

Note that most of what would test this commit actually lives on the flat memrefs branch since I had to restore it.


pcf000:
LGTM.  I'll admit that I didn't completely understand the original and I don't completely understand this version, but I do believe the corresponding parts are doing corresponding things, and some of the original complexity is replaced by something less so.

pcf000 (approved):

PR #1522, opened 2024-05-17
MIGraphX support for 3-D and 1-D convolutions

pcf000:
Pull request #1505 generalised the convolution code to handle N dimensions (for the most part).  This pull request covers the conversion from migraphx through tosa to rock.  There are a couple bits that aren't completely general yet, but it does handle (with a patch to AMDMIGraphX) "make check" and the python_3dunet example.

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOps.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOps.td
@@ -139,6 +139,7 @@ def Tosa_Conv3DOp : Tosa_InferShapedTypeOp<"conv3d"> {
     Tosa_IntArrayAttr6:$pad,
     Tosa_IntArrayAttr3:$stride,
     Tosa_IntArrayAttr3:$dilation,
+    OptionalAttr<I64Attr>:$group,
     OptionalAttr<Tosa_ConvOpQuantizationAttr>:$quantization_info,
     DefaultValuedOptionalAttr<BoolAttr, "false">:$local_bound
   );
@@ -1503,7 +1504,7 @@ def Tosa_ReduceSumOp : Tosa_InferTensorTypeOp<"reduce_sum"> {
 
   let hasFolder = 1;
   let hasVerifier = 1;
-  
+
   let extraClassDeclaration = [{
     /// Returns true when two result types are compatible for this op;
     /// Method used by InferTypeOpInterface.
@@ -1653,7 +1654,7 @@ def Tosa_ReverseOp: Tosa_Op<"reverse", [
 
   let hasFolder = 1;
   let hasVerifier = 1;
-  
+
   let assemblyFormat = "operands attr-dict `:` functional-type(operands, results)";
 }
 
@@ -1836,9 +1837,9 @@ def Tosa_CastOp: Tosa_Op<"cast", [Pure,
 
     | Mode                     | Input   | Output  |
     |--------------------------|---------|---------|
-    | signed 8 to bool         | int8    | Boolean | 
-    | signed 16 to bool        | int16   | Boolean | 
-    | signed 32 to bool        | int32   | Boolean | 
+    | signed 8 to bool         | int8    | Boolean |
+    | signed 16 to bool        | int16   | Boolean |
+    | signed 32 to bool        | int32   | Boolean |
     | bool to 8                | Boolean | int8    |
     | bool to 16               | Boolean | int16   |
     | bool to 32               | Boolean | int32   |
@@ -1852,8 +1853,8 @@ def Tosa_CastOp: Tosa_Op<"cast", [Pure,
     | float to signed 16       | float   | int16   |
     | signed 8 to float        | int8    | float   |
     | signed 16 to float       | int16   | float   |
-    | float 32 to float 64     | float32 | float64 | 
-    | float 64 to float 32     | float64 | float32 | 
+    | float 32 to float 64     | float32 | float64 |
+    | float 64 to float 32     | float64 | float32 |
   }];
 
   let arguments = (ins
--- a/external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOps.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOps.td
@@ -139,7 +139,6 @@ def Tosa_Conv3DOp : Tosa_InferShapedTypeOp<"conv3d"> {
     Tosa_IntArrayAttr6:$pad,
     Tosa_IntArrayAttr3:$stride,
     Tosa_IntArrayAttr3:$dilation,
-    OptionalAttr<I64Attr>:$group,
     OptionalAttr<Tosa_ConvOpQuantizationAttr>:$quantization_info,
     DefaultValuedOptionalAttr<BoolAttr, "false">:$local_bound
   );
@@ -1504,7 +1503,7 @@ def Tosa_ReduceSumOp : Tosa_InferTensorTypeOp<"reduce_sum"> {
 
   let hasFolder = 1;
   let hasVerifier = 1;
-
+  
   let extraClassDeclaration = [{
     /// Returns true when two result types are compatible for this op;
     /// Method used by InferTypeOpInterface.
@@ -1654,7 +1653,7 @@ def Tosa_ReverseOp: Tosa_Op<"reverse", [
 
   let hasFolder = 1;
   let hasVerifier = 1;
-
+  
   let assemblyFormat = "operands attr-dict `:` functional-type(operands, results)";
 }
 
@@ -1837,9 +1836,9 @@ def Tosa_CastOp: Tosa_Op<"cast", [Pure,
 
     | Mode                     | Input   | Output  |
     |--------------------------|---------|---------|
-    | signed 8 to bool         | int8    | Boolean |
-    | signed 16 to bool        | int16   | Boolean |
-    | signed 32 to bool        | int32   | Boolean |
+    | signed 8 to bool         | int8    | Boolean | 
+    | signed 16 to bool        | int16   | Boolean | 
+    | signed 32 to bool        | int32   | Boolean | 
     | bool to 8                | Boolean | int8    |
     | bool to 16               | Boolean | int16   |
     | bool to 32               | Boolean | int32   |
@@ -1853,8 +1852,8 @@ def Tosa_CastOp: Tosa_Op<"cast", [Pure,
     | float to signed 16       | float   | int16   |
     | signed 8 to float        | int8    | float   |
     | signed 16 to float       | int16   | float   |
-    | float 32 to float 64     | float32 | float64 |
-    | float 64 to float 32     | float64 | float32 |
+    | float 32 to float 64     | float32 | float64 | 
+    | float 64 to float 32     | float64 | float32 | 
   }];
 
   let arguments = (ins
--- a/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
+++ b/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
@@ -68,10 +68,10 @@ static bool canBeCalledWithBarePointers(gpu::GPUFuncOp func) {
 
 static constexpr StringLiteral amdgcnDataLayout =
     "e-p:64:64-p1:64:64-p2:32:32-p3:32:32-p4:64:64-p5:32:32-p6:32:32"
-    "-p7:160:256:256:32-p8:128:128-i64:64-v16:16-v24:32-v32:32-v48:64-v96:"
-    "128-v192:256-v256:256-v512:512-v1024:1024-v2048:2048-n32:64-S32-A5-"
-    "G1-ni:7:8";
-    
+    "-p7:160:256:256:32-p8:128:128-p9:192:256:256:32-i64:64-v16:16-v24:32-v32:"
+    "32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-v2048:2048-n32:"
+    "64-S32-A5-G1-ni:7:8:9";
+
 Value getLaneId(ConversionPatternRewriter &rewriter, Location loc,
                 const unsigned indexBitwidth) {
   auto int32Type = IntegerType::get(rewriter.getContext(), 32);
--- a/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
+++ b/external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
@@ -71,7 +71,7 @@ static constexpr StringLiteral amdgcnDataLayout =
     "-p7:160:256:256:32-p8:128:128-i64:64-v16:16-v24:32-v32:32-v48:64-v96:"
     "128-v192:256-v256:256-v512:512-v1024:1024-v2048:2048-n32:64-S32-A5-"
     "G1-ni:7:8";
-    
+
 Value getLaneId(ConversionPatternRewriter &rewriter, Location loc,
                 const unsigned indexBitwidth) {
   auto int32Type = IntegerType::get(rewriter.getContext(), 32);
--- a/mlir/lib/CAPI/Dialect/MIGraphX.cpp
+++ b/mlir/lib/CAPI/Dialect/MIGraphX.cpp
@@ -20,6 +20,7 @@
 #include "mlir/ExecutionEngine/OptUtils.h"
 #include "mlir/ExecutionEngine/RocmDeviceName.h"
 #include "llvm/ADT/StringRef.h"
+#include "llvm/Support/Process.h"
 #include "llvm/Support/TargetSelect.h"
 #include <mutex>
 #include <vector>
@@ -137,7 +138,7 @@ void mlirMIGraphXAddHighLevelPipeline(MlirPassManager pm) {
   passMan->setNesting(mlir::PassManager::Nesting::Implicit);
   mlir::migraphx::addHighLevelPipeline(*passMan);
   mlir::rock::buildBufferizePipeline(*passMan);
-  if (std::getenv("MIGRAPHX_MLIR_IR_PRINT")) {
+  if (llvm::sys::Process::GetEnv("MIGRAPHX_MLIR_IR_PRINT")) {
     passMan->getContext()->disableMultithreading();
     passMan->enableIRPrinting();
   }
@@ -174,7 +175,7 @@ MLIR_CAPI_EXPORTED bool mlirMIGraphXAddBackendPipeline(MlirPassManager pm,
   opts.optLevel = 3;
   mlir::rock::buildBackendPipeline(*passMan, opts);
 
-  if (std::getenv("MIGRAPHX_MLIR_IR_PRINT")) {
+  if (llvm::sys::Process::GetEnv("MIGRAPHX_MLIR_IR_PRINT")) {
     passMan->getContext()->disableMultithreading();
     passMan->enableIRPrinting();
   }
--- a/mlir/lib/CAPI/Dialect/MIGraphX.cpp
+++ b/mlir/lib/CAPI/Dialect/MIGraphX.cpp
@@ -137,6 +137,10 @@ void mlirMIGraphXAddHighLevelPipeline(MlirPassManager pm) {
   passMan->setNesting(mlir::PassManager::Nesting::Implicit);
   mlir::migraphx::addHighLevelPipeline(*passMan);
   mlir::rock::buildBufferizePipeline(*passMan);
+  if (std::getenv("MIGRAPHX_MLIR_IR_PRINT")) {
+    passMan->getContext()->disableMultithreading();
+    passMan->enableIRPrinting();
+  }
 }
 
 MLIR_CAPI_EXPORTED void
@@ -169,6 +173,11 @@ MLIR_CAPI_EXPORTED bool mlirMIGraphXAddBackendPipeline(MlirPassManager pm,
   opts.features = devName.getFeaturesForBackend();
   opts.optLevel = 3;
   mlir::rock::buildBackendPipeline(*passMan, opts);
+
+  if (std::getenv("MIGRAPHX_MLIR_IR_PRINT")) {
+    passMan->getContext()->disableMultithreading();
+    passMan->enableIRPrinting();
+  }
 
   return true;
 }
--- a/mlir/lib/CAPI/Dialect/MIGraphX.cpp
+++ b/mlir/lib/CAPI/Dialect/MIGraphX.cpp
@@ -20,7 +20,6 @@
 #include "mlir/ExecutionEngine/OptUtils.h"
 #include "mlir/ExecutionEngine/RocmDeviceName.h"
 #include "llvm/ADT/StringRef.h"
-#include "llvm/Support/Process.h"
 #include "llvm/Support/TargetSelect.h"
 #include <mutex>
 #include <vector>
@@ -138,10 +137,6 @@ void mlirMIGraphXAddHighLevelPipeline(MlirPassManager pm) {
   passMan->setNesting(mlir::PassManager::Nesting::Implicit);
   mlir::migraphx::addHighLevelPipeline(*passMan);
   mlir::rock::buildBufferizePipeline(*passMan);
-  if (llvm::sys::Process::GetEnv("MIGRAPHX_MLIR_IR_PRINT")) {
-    passMan->getContext()->disableMultithreading();
-    passMan->enableIRPrinting();
-  }
 }
 
 MLIR_CAPI_EXPORTED void
@@ -174,11 +169,6 @@ MLIR_CAPI_EXPORTED bool mlirMIGraphXAddBackendPipeline(MlirPassManager pm,
   opts.features = devName.getFeaturesForBackend();
   opts.optLevel = 3;
   mlir::rock::buildBackendPipeline(*passMan, opts);
-
-  if (llvm::sys::Process::GetEnv("MIGRAPHX_MLIR_IR_PRINT")) {
-    passMan->getContext()->disableMultithreading();
-    passMan->enableIRPrinting();
-  }
 
   return true;
 }
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -235,7 +235,7 @@ LogicalResult ConvConverter<ConvType>::matchAndRewrite(
   newShape.push_back(outShape[1]);
   Type newOutTy = RankedTensorType::get(newShape, outputTy.getElementType());
 
-  auto expandTo2D = [&rewriter, &loc](mlir::Value value) {
+  auto expandTo2D = [&rewriter, loc](mlir::Value value) {
     auto origShape = value.getType().cast<ShapedType>().getShape();
     std::vector<int64_t> expShape(origShape);
     expShape.insert(std::prev(expShape.end()), 1);
@@ -263,6 +263,7 @@ LogicalResult ConvConverter<ConvType>::matchAndRewrite(
                 loc, outputTy.getElementType(),
                 filter.getType().template cast<ShapedType>().getShape()[0],
                 rewriter)});
+    cop->setAttr("rock.expanded_from_1d", rewriter.getBoolAttr(true));
     break;
 
   case 2:
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -235,9 +235,36 @@ LogicalResult ConvConverter<ConvType>::matchAndRewrite(
   newShape.push_back(outShape[1]);
   Type newOutTy = RankedTensorType::get(newShape, outputTy.getElementType());
 
+  auto expandTo2D = [&rewriter, &loc](mlir::Value value) {
+    auto origShape = value.getType().cast<ShapedType>().getShape();
+    std::vector<int64_t> expShape(origShape);
+    expShape.insert(std::prev(expShape.end()), 1);
+    Value reshaped = rewriter.create<tosa::ReshapeOp>(
+        loc, value, rewriter.getDenseI64ArrayAttr(expShape));
+    return reshaped;
+  };
+
   // Construct a new Conv2DOp.
   Operation *cop;
+  Type new1DOutTy;
   switch (dims) {
+  case 1:
+    // Expand to do a conv2d, because there's no conv1d op.
+    newShape.insert(std::prev(newShape.end()), 1);
+    new1DOutTy = RankedTensorType::get(newShape, outputTy.getElementType());
+    input = expandTo2D(input);
+    filter = expandTo2D(filter);
+
+    cop = rewriter.create<tosa::Conv2DOp>(
+        loc, new1DOutTy,
+        ValueRange{
+            input, filter,
+            getZeroTensor(
+                loc, outputTy.getElementType(),
+                filter.getType().template cast<ShapedType>().getShape()[0],
+                rewriter)});
+    break;
+
   case 2:
     cop = rewriter.create<tosa::Conv2DOp>(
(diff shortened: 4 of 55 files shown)


pcf000:
That "Changing 'rock.conv2d'" commit came from the rebase-via-merge process, and contains merge conflicts.  Confusing, though, since an earlier commit exists with the same summary but more extensive contents.

mlir/lib/CAPI/Dialect/MIGraphX.cpp
```diff
@@ -137,6 +137,10 @@ void mlirMIGraphXAddHighLevelPipeline(MlirPassManager pm) {
   passMan->setNesting(mlir::PassManager::Nesting::Implicit);
   mlir::migraphx::addHighLevelPipeline(*passMan);
   mlir::rock::buildBufferizePipeline(*passMan);
+  if (std::getenv("MIGRAPHX_MLIR_IR_PRINT")) {
```
pcf000: MIGraphX likes environment variables to control things.  Also, there's no place to hang --mlir-ir-print-all in a python script.  I needed this to see what rocMLIR was doing underneath python_3dunet.  enableIRPrinting() requires disableMultithreading().

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -235,9 +235,36 @@ LogicalResult ConvConverter<ConvType>::matchAndRewrite(
   newShape.push_back(outShape[1]);
   Type newOutTy = RankedTensorType::get(newShape, outputTy.getElementType());
 
+  auto expandTo2D = [&rewriter, &loc](mlir::Value value) {
```
pcf000: migraphx.convolution becomes tosa.conv2d or tosa.conv3d.  There is no tosa.conv1d, so here we expand those into conv2d by adding an extra x1 dimension to the tensors and some dummy values to the attributes.  There are tosa.reshapes on the inputs and another after the result.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -287,6 +314,16 @@ LogicalResult ConvConverter<ConvType>::matchAndRewrite(
 
   // convolution config attributes
 
+  if (dims == 1) {
```
pcf000, line 325: Attributes for the dummy dimension.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -61,7 +61,7 @@ static bool isConstantZero(Value v) {
 }
 
 static Value expandTensor(ConversionPatternRewriter &rw, Operation *op,
-                          Value operand, SmallString<5> &layout,
+                          Value operand, SmallString<7> &layout,
```
pcf000: Bigger string for layouts with more dimensions.  I don't like this much, but I didn't find the correct ArrayRef or something else to generalise it.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -140,21 +140,27 @@ getArchAttributes(Operation *op, Type inputType) {
...
+             const DenseI64ArrayAttr &stride, const DenseI64ArrayAttr &dilation,
+             int64_t group) {
   Location loc = op->getLoc();
 
-  SmallString<5> filterLayout("kyxc");
+  SmallString<7> filterLayout("kyxc");
   if (auto attr = op->getAttrOfType<StringAttr>("filter_layout"))
     filterLayout = attr.getValue();
-  SmallString<5> inputLayout("nhwc");
+  else if (filter.getType().template cast<ShapedType>().getRank() > 4)
```
pcf000, line 155: Really not "> 4" as much as "== 5";  ie, not general.  We also have mixed-mode layouts here, with x/y/h/w for 2-D and 0/1/2 for 3-D.  Some existing rocMLIR tests failed if I used "k01c" and I didn't want to spend the time to track it down.  Will need a cleanup ticket later.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -266,6 +272,68 @@ class ConvConverter final : public OpConversionPattern<tosa::Conv2DOp> {
   }
 };
 
+class Conv3DConverter final : public OpConversionPattern<tosa::Conv3DOp> {
```
pcf000: Begging for abstraction.  Conv3DConverter is nearly identical to ConvConverter, except for the op type and the sizes of biasShape and reassociations.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -445,6 +513,7 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
                                              Value tOutput,
                                              ArrayRef<int32_t> dims,
                                              Value tInput) const {
+    SmallVector<Operation *> toBeErased;
```
pcf000: While debugging, I wandered down the address-sanitiser path.  Didn't find my problem that way, but I did see that nodes erased here were sometimes used after erasure.  Collect them and then erase at the end.

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -439,7 +439,10 @@ LogicalResult makeToLayoutLikeFromLayoutAlong(
   bool inOrder = true;
   size_t prevIndex = 0;
   for (StringAttr expected : expectedOrder) {
-    size_t thisIndex = toLayoutIdxs.find(expected)->getSecond();
+    auto foundp = toLayoutIdxs.find(expected);
+    assert(foundp != toLayoutIdxs.end());
```
pcf000: I had situations that wouldn't find the expected element in the toLayoutIdxs, causing the ->getSecond() to barf obscurely.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -451,19 +451,19 @@ TuningParamSet *createTunableParamSpace(ModuleOp mod, TuningParamSetKind kind) {
   newSpace = new TuningParamSet();
 
   // create range and heuristic
-  WalkResult findPrimary =
```
pcf000: No change here, just formatting that I'll fix.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -477,7 +477,8 @@ TuningParamSet *createTunableParamSpace(ModuleOp mod, TuningParamSetKind kind) {
     return WalkResult::interrupt();
   });
   if (!findPrimary.wasInterrupted() && !findAttention.wasInterrupted()) {
-    delete newSpace;
+    llvm_unreachable(
```
pcf000: If we don't find a matching op, newSpace is deleted and then returned, causing additional obscure barfs.  A proposed improvement is to change the interface so this function can return nullptr and the caller can detect and handle the situation.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -675,6 +672,14 @@ LogicalResult getTuningProblemStr(rock::RockGemmWrapperInterface gemmIF,
     SmallString<5> iLayout("#####");
     SmallString<5> oLayout("#####");
 
+    if (size > 5) {
```
pcf000: Code below relies on the layout strings being the correct length so it can overwrite characters.  This code doesn't handle 1-D, which migraphx doesn't care about but other paths might.

external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOps.td
```diff
@@ -139,6 +139,7 @@ def Tosa_Conv3DOp : Tosa_InferShapedTypeOp<"conv3d"> {
     Tosa_IntArrayAttr6:$pad,
     Tosa_IntArrayAttr3:$stride,
     Tosa_IntArrayAttr3:$dilation,
+    OptionalAttr<I64Attr>:$group,
```
krzysz00: We'll want to get this into a nice `[DO NOT SQUASH]` for for upstreaming

mlir/lib/CAPI/Dialect/MIGraphX.cpp
krzysz00: Ah. This is a slightly different bit of debug machinery though it is still meaningfully a subset of ticket 1227 (the debug flags one).

But also ... LLVM's got a utility for getenv, we might could use that here instead

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
krzysz00: Nit: no need to take `loc` by reference

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
krzysz00: Could you hang an attribute on the tosa.conv2d that's something like `rock.actually_1d`

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
krzysz00: Yeah, this should be `MutableArrayRef<char>`, probably

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -140,21 +140,27 @@ getArchAttributes(Operation *op, Type inputType) {
 }
 
 static FailureOr<rock::ConvOp>
-makeRockConv2D(ConversionPatternRewriter &rw, Operation *op, Value input,
-               Value filter, Value output, const DenseI64ArrayAttr &pad,
-               const DenseI64ArrayAttr &stride,
-               const DenseI64ArrayAttr &dilation, int64_t group) {
+makeRockConv(ConversionPatternRewriter &rw, Operation *op, Value input,
+             Value filter, Value output, const DenseI64ArrayAttr &pad,
+             const DenseI64ArrayAttr &stride, const DenseI64ArrayAttr &dilation,
```
krzysz00: While we're here, none of these attributes need to be by const reference

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
krzysz00: Yeah, integer template, no?

krzysz00: Or constructor arguments

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -266,6 +272,68 @@ class ConvConverter final : public OpConversionPattern<tosa::Conv2DOp> {
...
 };
 
+class Conv3DConverter final : public OpConversionPattern<tosa::Conv3DOp> {
+public:
+  using OpConversionPattern<tosa::Conv3DOp>::OpConversionPattern;
+
+  LogicalResult matchAndRewrite(tosa::Conv3DOp op,
+                                tosa::Conv3DOp::Adaptor adaptor,
+                                ConversionPatternRewriter &rw) const final {
+    auto operands = adaptor.getOperands();
```
krzysz00: Nit: much less auto

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -570,7 +639,7 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
           return failure();
         }
         if (op->use_empty())
-          rewriter.eraseOp(op);
+          toBeErased.push_back(op);
```
krzysz00: Why?

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
krzysz00: Put a string on the assert?

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
krzysz00: This is `emitFatalError()` or such more than unreachable

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -235,9 +235,37 @@ LogicalResult ConvConverter<ConvType>::matchAndRewrite(
   newShape.push_back(outShape[1]);
   Type newOutTy = RankedTensorType::get(newShape, outputTy.getElementType());
 
+  auto expandTo2D = [&rewriter, loc](mlir::Value value) {
+    auto origShape = value.getType().cast<ShapedType>().getShape();
+    std::vector<int64_t> expShape(origShape);
```
krzysz00: `SmallVector`

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -235,9 +235,37 @@ LogicalResult ConvConverter<ConvType>::matchAndRewrite(
   newShape.push_back(outShape[1]);
   Type newOutTy = RankedTensorType::get(newShape, outputTy.getElementType());
 
+  auto expandTo2D = [&rewriter, loc](mlir::Value value) {
+    auto origShape = value.getType().cast<ShapedType>().getShape();
```
krzysz00: `ArrayRef<int64_t> origShape`

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -287,6 +315,16 @@ LogicalResult ConvConverter<ConvType>::matchAndRewrite(
 
   // convolution config attributes
 
+  if (dims == 1) {
+    assert(dilations.size() == 1);
+    assert(strides.size() == 1);
```
krzysz00: Don't assert. `return emitError()` - this is a user error, not a failure of ours.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -235,9 +235,37 @@ LogicalResult ConvConverter<ConvType>::matchAndRewrite(
   newShape.push_back(outShape[1]);
   Type newOutTy = RankedTensorType::get(newShape, outputTy.getElementType());
 
+  auto expandTo2D = [&rewriter, loc](mlir::Value value) {
+    auto origShape = value.getType().cast<ShapedType>().getShape();
+    std::vector<int64_t> expShape(origShape);
+    expShape.insert(std::prev(expShape.end()), 1);
```
krzysz00: ... this is just `expShape.push_back(1);`

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -259,7 +287,7 @@ LogicalResult ConvConverter<ConvType>::matchAndRewrite(
                 rewriter)});
     break;
   default:
-    llvm_unreachable("Only 2-D and 3-D have been implemented.");
+    llvm_unreachable("Only 1-D, 2-D, and 3-D have been implemented.");
```
krzysz00: This is not `llvm_unreachable`, this is returning an error.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
krzysz00: This is still `MutableArrayRef<char>`

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
krzysz00: Same not on not naking attributes by const reference

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -140,21 +140,27 @@ getArchAttributes(Operation *op, Type inputType) {
...
-               const DenseI64ArrayAttr &stride,
-               const DenseI64ArrayAttr &dilation, int64_t group) {
+makeRockConv(ConversionPatternRewriter &rw, Operation *op, Value input,
+             Value filter, Value output, const DenseI64ArrayAttr &pad,
+             const DenseI64ArrayAttr &stride, const DenseI64ArrayAttr &dilation,
+             int64_t group) {
   Location loc = op->getLoc();
 
-  SmallString<5> filterLayout("kyxc");
+  SmallString<7> filterLayout("kyxc");
```
krzysz00: Let's go and make this 8 for good stack alignment

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
krzysz00: What tests, and failed how?

krzysz00, line 155: Could we use that `rock.actually_conv_1d` attribute here to drop those dummy dimensions?

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
krzysz00: Yeah no, let's actually debug that - but maybe in a future ticket

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -439,7 +439,10 @@ LogicalResult makeToLayoutLikeFromLayoutAlong(
...
+    size_t thisIndex = foundp->getSecond();
+    //    size_t thisIndex = toLayoutIdxs.find(expected)->getSecond();
```
krzysz00: Commented-out code?

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
```diff
@@ -453,9 +456,8 @@ LogicalResult makeToLayoutLikeFromLayoutAlong(
   /// And now we have to actually do the thing
   // Is just an attribute to allow array builder
   SmallVector<Attribute> newToLayout;
-  llvm::SmallDenseSet<StringAttr> permutedDimsSet;
-  for (StringAttr toPermute : expectedOrder)
-    permutedDimsSet.insert(toPermute);
+  llvm::SmallDenseSet<StringAttr> permutedDimsSet{expectedOrder.begin(),
```
krzysz00, line 459: Nice catch!

mlir/lib/Dialect/Rock/Transforms/ConvToGemm.cpp
krzysz00: ... Heck, this can be triggered by user input, it should be a `return failure()` instead

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
krzysz00: ... you could instead `SmallString<8> oLayout(size, '#');` above - I might have the constructor backwards

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -675,6 +673,14 @@ LogicalResult getTuningProblemStr(rock::RockGemmWrapperInterface gemmIF,
...
+      for (unsigned i = 0; i < size - 5; i++) {
```
krzysz00: This creates way too many `#`s also

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
pcf000: Actually, no.  It inserts the 1 as the next-to-last element.  I didn't want to mess with channels, just add a fake "w".

krzysz00: Ah, right, Tosa's NHWC

Then let's construct the shape by parts - a small-vector of everything but channels (via `drop_back()`), add a `1`, and then push the channels on

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
pcf000: On further study, I can't make these SmallStrings into MutableArrayRef in expandTensor(), because part of the expansion is to insert a "g" into the layout string.  As a MutableArrayRef, I can modify the existing elements, but since the buffer is owned by somebody else, I can't expand it.  Therefore, SmallString<8> all around.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
pcf000: Yes, but I'd prefer to leave that to the hypothetical cleanup pull-request, in the interests of not breaking anything during the review close to the deadline.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
pcf000: On further study, I can't make these SmallStrings into MutableArrayRef in expandTensor(), because part of the expansion is to insert a "g" into the layout string. As a MutableArrayRef, I can modify the existing elements, but since the buffer is owned by somebody else, I can't expand it. Therefore, SmallString<8> all around.

external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOps.td
manupak: We have lots of `if (group > 1 && isConv2DOp` added to external in TosaToLinalgNamed lowering that has not been upstreamed.
Therefore, I d like to request a MIGraphX e2e test with `--clone-harness` make sure this change is properly tested -- otherwise we ll have a hard time verifying fused modules from MIGraphX.

external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOps.td
```diff
@@ -1653,7 +1654,7 @@ def Tosa_ReverseOp: Tosa_Op<"reverse", [
 
   let hasFolder = 1;
   let hasVerifier = 1;
-  
+
```
manupak: please remove spurious unncessary changes to external/ which will affect upstream merges.

external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOps.td
```diff
@@ -1852,8 +1853,8 @@ def Tosa_CastOp: Tosa_Op<"cast", [Pure,
     | float to signed 16       | float   | int16   |
     | signed 8 to float        | int8    | float   |
     | signed 16 to float       | int16   | float   |
-    | float 32 to float 64     | float32 | float64 | 
-    | float 64 to float 32     | float64 | float32 | 
+    | float 32 to float 64     | float32 | float64 |
+    | float 64 to float 32     | float64 | float32 |
```
manupak: same here

external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
```diff
@@ -68,10 +68,10 @@ static bool canBeCalledWithBarePointers(gpu::GPUFuncOp func) {
...
 static constexpr StringLiteral amdgcnDataLayout =
     "e-p:64:64-p1:64:64-p2:32:32-p3:32:32-p4:64:64-p5:32:32-p6:32:32"
-    "-p7:160:256:256:32-p8:128:128-i64:64-v16:16-v24:32-v32:32-v48:64-v96:"
-    "128-v192:256-v256:256-v512:512-v1024:1024-v2048:2048-n32:64-S32-A5-"
-    "G1-ni:7:8";
-    
+    "-p7:160:256:256:32-p8:128:128-p9:192:256:256:32-i64:64-v16:16-v24:32-v32:"
+    "32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024-v2048:2048-n32:"
+    "64-S32-A5-G1-ni:7:8:9";
+
```
manupak: Why this change here?

mlir/lib/CAPI/Dialect/MIGraphX.cpp
```diff
@@ -137,6 +138,10 @@ void mlirMIGraphXAddHighLevelPipeline(MlirPassManager pm) {
   passMan->setNesting(mlir::PassManager::Nesting::Implicit);
   mlir::migraphx::addHighLevelPipeline(*passMan);
   mlir::rock::buildBufferizePipeline(*passMan);
+  if (llvm::sys::Process::GetEnv("MIGRAPHX_MLIR_IR_PRINT")) {
```
manupak: How is this different from using MIGRAPHX_MLIR_TRACE_MLIR?
If we are doing this change, I d like that documented here.

mlir/lib/CAPI/Dialect/MIGraphX.cpp
```diff
@@ -170,5 +175,10 @@ MLIR_CAPI_EXPORTED bool mlirMIGraphXAddBackendPipeline(MlirPassManager pm,
   opts.optLevel = 3;
   mlir::rock::buildBackendPipeline(*passMan, opts);
 
+  if (llvm::sys::Process::GetEnv("MIGRAPHX_MLIR_IR_PRINT")) {
```
manupak: Please refactor this to a single function to reduce error surface.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -235,9 +235,38 @@ LogicalResult ConvConverter<ConvType>::matchAndRewrite(
   newShape.push_back(outShape[1]);
   Type newOutTy = RankedTensorType::get(newShape, outputTy.getElementType());
 
+  auto expandTo2D = [&rewriter, loc](mlir::Value value) {
```
manupak, line 243: [stylistic] nit : I generally would use a lambda if it interacts heavily with variables in the scope where as in this one only uses rewriter and loc. I d consider defining a function with proper return types and some docs.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -235,9 +235,38 @@ LogicalResult ConvConverter<ConvType>::matchAndRewrite(
...
+
+    cop = rewriter.create<tosa::Conv2DOp>(
+        loc, new1DOutTy,
+        ValueRange{
+            input, filter,
+            getZeroTensor(
+                loc, outputTy.getElementType(),
+                filter.getType().template cast<ShapedType>().getShape()[0],
+                rewriter)});
+    cop->setAttr("rock.expanded_from_1d", rewriter.getBoolAttr(true));
```
manupak: please add this to RockAttrDefs and use here.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -658,40 +660,61 @@ LogicalResult getTuningProblemStr(rock::RockGemmWrapperInterface gemmIF,
     for (unsigned i = 0; i < size; ++i) {
       auto filterAttr =
           filterLayoutAttr.getValue()[i].template cast<StringAttr>();
-      fLayoutMap[filterAttr.getValue()] = i;
-    }
-    for (unsigned i = 0; i < size; ++i) {
+      auto fs = filterAttr.getValue();
```
manupak: nit: auto usage is only encouraged if its obvious as in the previous line in llvm/mlir project.
please use the type and expressive name here.

manupak (dismissed):
I think the tosa -> linalg plumbing is not done here and it is required if you are changing tosa. I d like that verified as a clone test.

[non blocking on this PR because we have already added 'group' but have not made an attempt of upstreaming]  As a secondary point, I d like at least raising the point of adding group dimension in the TOSA spec discourse that is a pre-requisite for upstreaming here : https://discuss.mlplatform.org/c/discussion-of-the-tosa-specification/5

external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOps.td
pcf000: What is involved in making such a test?  (I've never committed anything to MIGraphX.)  Would it be better to similarly test for "group > 1" in the migraphx conversion and leave the grouped ones out of the rock.conv path?

external/llvm-project/mlir/include/mlir/Dialect/Tosa/IR/TosaOps.td
pcf000: Trailing whitespace, the bane of my existence.

external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
pcf000: I did not change this file initially.  I had a clang-format failure because of a whitespace change.  When I synced that, I had a test failure because this layout variable changed on the develop branch while I still had the old version.

mlir/lib/CAPI/Dialect/MIGraphX.cpp
pcf000: MIGRAPHX_TRACE_MLIR dumps the IR at a couple of selected spots.  This variable enables the mechanism behind --print-ir-after-all, which dumps the IR after each pass.  For me, it was the difference between seeing migraphx.convolution and seeing where tosa.conv3d became rock.conv.

Krzysztof's issue #1227 will make this variable unnecessary by enabling a mechanism to take command-line options through an environment variable.  I'll go ahead and remove this in anticipation.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
pcf000: Thanks, I didn't know about that.

external/llvm-project/mlir/lib/Conversion/GPUToROCDL/LowerGpuOpsToROCDLOps.cpp
manupak: Yes I think if you merge the develop into your branch -- even when your branch dont have conflicts with develop -- this wont be needed because clang-format checks against origin/develop and if your branch is upto date with origin/develop clang-format wont even touch this file.

manupak (commented):
Only few things to go.

## Non-test code :

1) Please merge the latest develop into this branch --> that would make the clang-format changes go away. Then this PR would not need external/ changes and [DO NOT SQUASH]

## Test-code :

1) I d like at least a single e2e test. It should be trivial to add one. here is an example : https://github.com/ROCm/rocMLIR/blob/develop/mlir/test/fusion/pr-e2e/mixr-slice-convolution-add-small.mlir (note --clone-harness)
2) There are changes in TosaToRock -- there should be at least a single test in TosaToRock conversion.
3) There are changes in MIGraphXToTosa -  there should be at least a single test in MIGraphXToTosa conversion.

mlir/lib/CAPI/Dialect/MIGraphX.cpp
manupak: Cool thanks! I was going to suggest this passManager option setting should be done by MIGraphX who owns it.

pcf000:
> 1. Please merge the latest develop into this branch --> that would make the clang-format changes go away. 

Pretty to think so, but we'll see.

> Then this PR would not need external/ changes and [DO NOT SQUASH]

Also because the tosa.conv3d group attribute is deferred, the original reason for it.

> 1. I d like at least a single e2e test. It should be trivial to add one. here is an example : 

I had been leaning on the migraphx "make check" tests, but this is fair.

> (note --clone-harness)

Ah.  I'd thought that was an migraphx thing.

manupak (approved):
Thanks! I do not have major concerns here.

PR #1524, opened 2024-05-17
Add scripts for analyzing tuning data

djramic:
This commit adds scripts for analyzing tuning data. 
Here are the two scripts:
* **analizeDataAvarege.py** -> sorts the perfconfigs by the best averge value for tFlops
* **analizeDataCount.py** -> sorts the perfconfigs by the number of time they were the best choice

Both scripts expect the full tuning data (with a .debug extension, from tuningRunner.py) to be located at the relative paths _conv/tunedData/_ and _gemm/tunedData/_


mlir/utils/performance/analysis/analizeDataAvarege.py
```diff
@@ -0,0 +1,65 @@
...
+    for file in tsv_files:
+        df = pd.read_csv(file, sep='\t')
+        dfs.append(df)
+
+    final_df = pd.concat(dfs, ignore_index=True)
+
+    unique_data_types = final_df['DataType'].unique()
+
+    top_perfconfigs_by_dataype = {}
+
```
giuseros: Remove spacing

mlir/utils/performance/analysis/analizeDataAvarege.py
```diff
@@ -0,0 +1,65 @@
...
+usage: python3 analizeDataAvarge.py -o [output_file] 
+
+"""
+
+parser = argparse.ArgumentParser()
+parser.add_argument('-o', '--output', required=True, help="output file path")
+args = parser.parse_args()
+
+conv_files = glob.glob('conv/tunedData/*.debug')
+gemm_files = glob.glob('gemm/tunedData/*.debug')
```
giuseros: Can you add a comment about how those `*.debug` files are generated?

mlir/utils/performance/analysis/analizeDataAvarege.py
```diff
@@ -0,0 +1,65 @@
...
+                'mean_tflops': mean_tflops[config]
+            }
+            for config in sorted_top_perfconfigs
+        ]
+
+    for data_type, configs in top_perfconfigs_by_dataype.items():
+        f.write(f"\nDataType: {data_type}\n")
+        for entry in configs:
+            f.write(f"PerfConfig: {entry['config']} with avarge TFlops: {entry['mean_tflops']}\n")
+
```
giuseros: Remove the last `\n`

mlir/utils/performance/analysis/analizeDataAvarege.py
```diff
@@ -0,0 +1,65 @@
...
+import pandas as pd
+
+"""
+usage: python3 analizeDataAvarge.py -o [output_file] 
+
+"""
+
+parser = argparse.ArgumentParser()
+parser.add_argument('-o', '--output', required=True, help="output file path")
+args = parser.parse_args()
```
giuseros: Could you add a `main` function and do something like
```
if __name__=="__main__":
main(args)
```

mlir/utils/performance/analysis/analizeDataAvarege.py
```diff
@@ -0,0 +1,65 @@
...
+        for perfconfig, group_df in fgroups:
+        
+            problems_count = len(group_df)
+            not_nan_count = pd.notna(group_df['TFlops']).sum()
+            not_nan_counts[perfconfig] = not_nan_count
+
+            mean_tflops[perfconfig] = group_df['TFlops'].mean()
+
+        sorted_counts = sorted(not_nan_counts.items(), key=lambda x: x[1], reverse=True)
+        top_perfconfigs = [perfconfig for perfconfig, count in sorted_counts if count == problems_count]
```
giuseros: Can you add some comments here and get rid of too much spacing?

mlir/utils/performance/analysis/analizeDataCount.py
```diff
@@ -0,0 +1,68 @@
+import argparse
```
giuseros: Many of the comments to this file are similar to the comments on the other file

giuseros (commented):
It looks fine, other than stylistic issues. Another question: how do we generate the quick list tuning? Do we run this script and get the first 20 configs (per datatype/ per arch)?

djramic:
> It looks fine, other than stylistic issues. Another question: how do we generate the quick list tuning? Do we run this script and get the first 20 configs (per datatype/ per arch)?

Yes, I took the perfconfigs from the top (or near the top) in the outputs from both scripts, separatley for DataType/arch. Also I tried to cover as many different blocksizes as possible.

mlir/utils/performance/analysis/analizeDataAvarege.py
```diff
@@ -0,0 +1,68 @@
+import sys
+import argparse
+import glob
+import pandas as pd
+
+def analizeData(output):
+    # Data on the paths conv/tunedData, gemm/tunedData are obtained as output files
```
giuseros: Sorry about being pedantic, but can you write the exact command you used (or giving few examples of the commands you used)? My point is that engineers like @ethansaurusrex should be able to immediately reproduce your results by reading this script

giuseros:
> Yes, I took the perfconfigs from the top (or near the top) in the outputs from both scripts, separatley for DataType/arch. Also I tried to cover as many different blocksizes as possible.

 Can you write a comment about this in the python scripts? Possibly, if you can include a bash command to do that in the comment would be best. Something like:
```
# Quick tuning DB can be generated  by:
# $ export threshold=20
# $ python script.py -o out.txt
# $ head -n $threshold out.txt
```

mlir/utils/performance/analysis/analizeDataAvarege.py
djramic: There is a usage comment right below the main function

giuseros: IIUC that is the usage of the script. I am talking about the command to generate the `.debug` files.

djramic:
> > Yes, I took the perfconfigs from the top (or near the top) in the outputs from both scripts, separatley for DataType/arch. Also I tried to cover as many different blocksizes as possible.
> 
> Can you write a comment about this in the python scripts? Possibly, if you can include a bash command to do that in the comment would be best. Something like:
> 
> ```
> # Quick tuning DB can be generated  by:
> # $ export threshold=20
> # $ python script.py -o out.txt
> # $ head -n $threshold out.txt
> ```
Unfortunately, at this point scripts aren't suitable for fully automating the generation of perfconfigs. I can add a more detailed description of the generated output data as a coment in the script. Is that fine?

mlir/utils/performance/analysis/gatherTuningData.sh
```diff
@@ -0,0 +1,27 @@
...
+ fi
+
+ ../tuningRunner.py --op conv -d -c ../conv-configs --output=conv/tunedData/tuned_CI_conv.tsv
+
+counter=0
+while IFS= read -r line
+do
+    counter=$((counter + 1))
+    python3 ../tuningRunner.py --op conv --config="$line" -d --output=conv/tunedData/tuned_conv_$counter.tsv
+done <"specificConvConfigs"
```
giuseros: I think this needs to be more generic. Let the user specify a parameter `configs` (or similar) and use that (and give a usage example in this bash script). In this way you don't have to have those files hanging in the repo, but you can use any file

mlir/utils/performance/analysis/specificConvConfigs
```diff
@@ -0,0 +1,89 @@
+convfp16 -F 1 -f NCHWG -I NCHWG -O NCHWG -n 1 -c 3 -H 224 -W 224 -k 64 -y 7 -x 7 -p 3 -q 3 -u 2 -v 2 -l 1 -j 1 -g 1
```
giuseros: Remove those. You can simply specify a file name in `gather TUningData.sh`

mlir/utils/performance/analysis/analizeData.py
```diff
@@ -0,0 +1,141 @@
...
+# Main function.
+def main(args=None):
+    """
+    usage: python3 analizeData.py
+
+    """
+
+    counted_win_gemm = analizeData("gemm")
+    counted_win_conv = analizeData("conv")
+    avrg_tfops_per_datatype = averagePerformance()
```
giuseros: This needs to be more homogeneous. Can you specify a folder (e.g., `data_folder`) as a parameter and use that folder to locate the `debug` files?

Then this code would look like:
```
    counted_win = analizeData(data_folder)
    avrg_tfops_per_datatype = averagePerformance(data_folder)
```

manupak:
We have *some* scripts right now.. @djramic are these needed anymore ?

djramic:
> We have _some_ scripts right now.. @djramic are these needed anymore ?

This PR is deprecated since the scripts are already included in the new ones for which exists active PR.  I'll close it.

PR #1529, opened 2024-05-21
Manually set overflow flags when expanding indexing maps

krzysz00:
When we work with transform maps in the context of rocMLIR, we know the following properties that are not true of a general affine map:

1. We may assume that the inputs to the map are non-negative. If a previous computation (like subtracting off padding) would cause the inputs to potentially be negative, the validity bit will cause us to not use the final computation result, so we can have all the undefined behavior/unjustified optimizations/... in the rest of the address calculation - it won't matter.

To put it another way, we speculatively execute all the index calculatinos for the valid case, and then discard the result in the invalid case.

LLVM can't know that's what we meant, so we have to do that annocation ourselves.

2. During the index diffs mechanism, the individual diffs can be negative (ex. if our map is "x -> x / 3, x % 3" and we have the inital coordinates (0, 2), then moving to x + 1 gets us coordinates (1, 0) and diffs (1, -2)). We also can therefore see a diff becoming negative in general. However, we know that, unless we're in an operator that can impact validity (Pad{}, mainly, but also some Embed{}), the sum of the original index and the diff is always non-negative.

That non-negative-ness allows us to sprinkle `nuw` on most index diff updates, which is also a property that might be hard to infer in general.

--

This entire setup is implemented with a custom affine map expanded, which has the side benefit of letting us skip emitting all that super noise code for `mod`. That visitor looks for potentially negative expressions (that is, those with negative constants somewhere in their subtrees) and annotates all the adds and multiplies with nuw/nsw as appropriate for the non-negativity of the inputs.

(Note, because of the analysis up in AnalyzeMemoryUse, we have the invariant that signed overflow can't occur - if you ever need x int32_t_max as a map output, we go to 64-bit `index` instead).

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
@@ -28,11 +28,11 @@
 #include "mlir/Dialect/Affine/Utils.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/GPU/IR/GPUDialect.h"
-#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/Dialect/SCF/Utils/Utils.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
+#include "mlir/IR/AffineExprVisitor.h"
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/Matchers.h"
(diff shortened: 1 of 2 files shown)


mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
```diff
@@ -68,6 +68,100 @@ struct RockSugarToLoopsPass
...
+
+  template <typename OpTy>
+  std::pair<Value, bool> buildBinaryExpr(AffineBinaryOpExpr expr) {
+    auto [lhs, lhsNneg] = visit(expr.getLHS());
+    auto [rhs, rhsNneg] = visit(expr.getRHS());
+    auto op = builder.create<OpTy>(loc, lhs, rhs);
+    return std::make_pair(op.getResult(), lhsNneg && rhsNneg);
+  }
+
+  std::pair<Value, bool> visitAddExpr(AffineBinaryOpExpr expr) {
```
manupak, line 100: nit : + override ?

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
```diff
@@ -508,7 +599,15 @@ struct IndexDiffUpdateRewritePattern
...
-      return b.create<AddIOp>(loc, original, diff);
+      // If the operation is something like Embed{} or Pad{} that can produce
+      // negative results (but thereby cancel validity - that is, if a negative
+      // result means all the further operations are being speculatively
+      // executed), we conservatively don't say anything. However, in other
+      // cases, the original value must be non-negative and the index diff can't
+      // make it go below zero, so we can declare nuw.
+      return b.create<AddIOp>(loc, original, diff,
+                              canBeInvalid ? IntegerOverflowFlags::none
+                                           : (IntegerOverflowFlags::nuw));
```
manupak, line 610: Why is the discrepency here to not add nsw as well as compared affine map expanders ?

manupak (commented):
Broadly looking good. minor comment/questions

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
krzysz00: I think `5 + -1` isn't `nsw`, which can happen in index diffs and be valid but can't happen validly in the affine expansion

manupak (approved):
This all looks good. Can we have a test please before merge for SugarToLoops ?

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
manupak: makes sense.

PR #1536, opened 2024-05-28
[DO NOT SQUASH] Fix navi3x f16 atomics

manupak:
Currently the Split-K based tests are all broken in Navi3x. This is due to three reasons : 
1) We had manual thresholds in the tests that were stricter
2) F32 : this pipeline uses non-accel gemm path where the perfConfig format is different -- hence not supported (we should really double check these before blanket enablement throughout all archs)
3) F16 : buffer_atomic_fadd on f16s are not present in RDNA3.

Therefore, this commit:
1) Removed the manual thresholds.
2) Disabled tests on F32 for Navi3x where WMMAs are not available.
3) AtomicAdd is lowered to a atomic cmpswap.

Moreover, it breaks the split-k single e2e test suite into three test suites : split-k_f16 (nighlty), split-k_f32 (nighlty) and PrSplitK (on PRs just for F16). The idea is the latter one will hopefully catch these from landing in codebase.

closes : https://github.com/ROCm/rocMLIR-internal/issues/1488

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
@@ -222,7 +222,7 @@ def AMDGPU_RawBufferAtomicCmpswapOp :
       AttrSizedOperandSegments,
       AllTypesMatch<["src", "cmp", "value"]>,
       AllElementTypesMatch<["value", "memref"]>]>,
-    Arguments<(ins AnyTypeOf<[I32, I64, F32, F64]>:$src,
+    Arguments<(ins AnyType:$src,
                    AnyType:$cmp,
                    Arg<AnyMemRef, "buffer to operate on", [MemRead, MemWrite]>:$memref,
                    Variadic<I32>:$indices,
--- a/external/llvm-project/mlir/lib/Dialect/AMDGPU/Transforms/EmulateAtomics.cpp
+++ b/external/llvm-project/mlir/lib/Dialect/AMDGPU/Transforms/EmulateAtomics.cpp
@@ -12,7 +12,10 @@
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/ControlFlow/IR/ControlFlow.h"
 #include "mlir/Dialect/ControlFlow/IR/ControlFlowOps.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/Dialect/Vector/IR/VectorOps.h"
 #include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/TypeUtilities.h"
 #include "mlir/Transforms/DialectConversion.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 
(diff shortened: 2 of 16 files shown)


manupak:
NOTE : I have not added unit tests for EXTERNAL changes yet; I m holding that off until we all agree on the approach in EXTERNAL.

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
```diff
@@ -222,7 +222,7 @@ def AMDGPU_RawBufferAtomicCmpswapOp :
       AttrSizedOperandSegments,
       AllTypesMatch<["src", "cmp", "value"]>,
       AllElementTypesMatch<["value", "memref"]>]>,
-    Arguments<(ins AnyTypeOf<[I32, I64, F32, F64]>:$src,
+    Arguments<(ins AnyType:$src,
```
krzysz00, line 225: I'd call for a stricter validation here, since there's only a limited number of things you can CAS? But maybe we're punting that to the compiler

external/llvm-project/mlir/lib/Dialect/AMDGPU/Transforms/EmulateAtomics.cpp
```diff
@@ -113,9 +132,11 @@ LogicalResult RawBufferAtomicByCasPattern<AtomicOp, ArithOp>::matchAndRewrite(
   rewriter.setInsertionPointToEnd(loopBlock);
   Value prevLoad = loopBlock->getArgument(0);
   Value operated = rewriter.create<ArithOp>(loc, data, prevLoad);
+  dataType = operated.getType();
 
   SmallVector<NamedAttribute> cmpswapAttrs;
   patchOperandSegmentSizes(origAttrs, cmpswapAttrs, DataArgAction::Duplicate);
+  // invariantArgs = invariantArgs.drop_front();
```
krzysz00: Why this?

external/llvm-project/mlir/lib/Dialect/AMDGPU/Transforms/EmulateAtomics.cpp
```diff
@@ -150,6 +171,18 @@ void mlir::amdgpu::populateAmdgpuEmulateAtomicsPatterns(
       (chipset.majorVersion == 9 && chipset.minorVersion < 0x08)) {
     target.addIllegalOp<RawBufferAtomicFaddOp>();
   }
+  // gfx11 has no fp16 atomics
+  if (chipset.majorVersion == 11) {
+    target.addDynamicallyLegalOp<RawBufferAtomicFaddOp>(
+        [](RawBufferAtomicFaddOp op) -> bool {
+          Type elemType = op.getValue().getType();
```
krzysz00: This is `getElementTypeOrSelf`

krzysz00 (approved):
I've got nitpicking about the external changes but not enough to block this if it works

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
manupak: Yes, I entertained the thought of listing out possible vector elements that are legal then I found out following code is already there and would give a much more meaningful error message than what tablegen would ve given.

https://github.com/ROCm/rocMLIR/blob/3612396bca1139abf25e2ed0085fe481d275af89/external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp#L111-L135

external/llvm-project/mlir/lib/Dialect/AMDGPU/Transforms/EmulateAtomics.cpp
manupak: ah done!

external/llvm-project/mlir/lib/Dialect/AMDGPU/Transforms/EmulateAtomics.cpp
manupak: done

external/llvm-project/mlir/lib/Dialect/AMDGPU/Transforms/EmulateAtomics.cpp
```diff
@@ -86,6 +89,23 @@ static void patchOperandSegmentSizes(ArrayRef<NamedAttribute> attrs,
...
+                             Value val) {
+  if (auto vectorType = val.getType().dyn_cast_or_null<VectorType>()) {
+    int64_t bitwidth =
+        vectorType.getElementTypeBitWidth() * vectorType.getNumElements();
+    Type elemType = rewriter.getIntegerType(bitwidth);
+    SmallVector<int64_t, 2> flatVecTypeShape(1, vectorType.getRank());
+    auto flatVecType = VectorType::get(flatVecTypeShape, elemType);
+    auto bitcast = rewriter.create<vector::BitCastOp>(loc, flatVecType, val);
+    auto zero = rewriter.createOrFold<arith::ConstantIndexOp>(loc, 0);
+    auto ret = rewriter.create<vector::ExtractElementOp>(loc, bitcast, zero);
```
krzysz00, line 103: I think `ExtractElementOp` is semi-deprecated, but that's not for this critical fix

krzysz00 (approved):

krzysz00:
Also, I intend to land this once it's got a CI pass

PR #1537, opened 2024-05-29
[URGENT] Fix 1d arg and 3d-conv merge CI race condition

manupak:
This commit changes the new migraphx test added
in 3d conv work to conform with 1d arg lowering
we currently do.


jerryyin (approved):

pcf000:
Hmm, I could swear it worked when I wrote it.  I can see that dimensions might need to change, but where's the race?

manupak:
The race is between PR merging -- which is common in a large OSS projects (unlike our one). TLDR is this happens :).

So what happened is 3d-conv PR and 1d arg PR are individually verified prior to merge. Then we merge them together && they dont have source conflicts. However, since 1d arg PR changed how tosa lowering happens and the test in 3d-conv PR becomes outdated. so thats the race.

PR #1540, opened 2024-05-30
Add back relDiff and absDiff thresholds for navi3x nightly

manupak:
I shouldve not changed relDiff and absDiff thresholds rather only should ve changed RMS which was the one affected by packed fp16 changes (https://github.com/ROCm/rocMLIR/pull/1484).

This commit re-instates the manual relDiff and absDiff thresholds for conv_regression_fwd_navi3x.


manupak:
Cross-posting here for posterity for the "why" : 

Summary: 
1) we had conv_regression_fwd test suite for both mfma and wmma.
2) then due to accuracy issues on navi side -- we created a seperate conv_regression_fwd_navi3x suite with manual thresholds for absDiff. However, conv_regression_fwd test suite had a manual threshold of -relDiff_threshold 0.00001. Thats why they are passing.
3) then we did packed fp16 changes that had less accuracy in RMS metric.
4) in the meantime, we have made navi3x not required to pass in nightly and we never saw the failures.
5) Due to 3), the manual RMS threshold was stricter, and I removed ALL manual thresholds because I saw them passing with "fixed". However, what should've been done is only RMS threshold should ve been removed. Therefore, these were failing with "rand 1".
6) PR1540 (above) should re-instate relDiff and absDiff less strict manual thresholds for navi3x.

jerryyin (approved):

krzysz00 (approved):
LGTM, let's send this

PR #1543, opened 2024-06-05
Add check for block size being 1024 or less

krzysz00:
We were, in some cases, generating blocksize = 2048 because we got too aggressive with removing some limits on the blocksize.

This fixes the issue.

Also, fix a weird thing where attention perf configs could be invalid and get in to the tuning set.

Fixes https://github.com/ROCm/rocMLIR-internal/issues/1491

Must backport

### DIFF
--- a/mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
+++ b/mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
@@ -257,9 +257,9 @@ void createGemmTuningRangeBF(TuningParamSet *newSpace,
                                                splitKFactor, forceUnroll, true);
                     if (gemmMPerBlock >= gemmMPerWave &&
                         gemmNPerBlock >= gemmMnPerXdl) {
-                      if (kind == TuningParamSetKind::Exhaustive ||
-                          (succeeded(tuningInfo.paramsProbablyValid(
-                               b, info, gemmParams)) &&
+                      if (succeeded(tuningInfo.paramsProbablyValid(
+                              b, info, gemmParams)) &&
+                          (kind == TuningParamSetKind::Exhaustive ||
                            succeeded(
                                tuningInfo.couldBePerformant(info, gemmParams))))
                         newSpace->tuningRange.push_back(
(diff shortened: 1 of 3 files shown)


manupak (approved):

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -257,9 +257,9 @@ void createGemmTuningRangeBF(TuningParamSet *newSpace,
                                                splitKFactor, forceUnroll, true);
                     if (gemmMPerBlock >= gemmMPerWave &&
                         gemmNPerBlock >= gemmMnPerXdl) {
-                      if (kind == TuningParamSetKind::Exhaustive ||
-                          (succeeded(tuningInfo.paramsProbablyValid(
-                               b, info, gemmParams)) &&
+                      if (succeeded(tuningInfo.paramsProbablyValid(
+                              b, info, gemmParams)) &&
+                          (kind == TuningParamSetKind::Exhaustive ||
```
manupak, line 262: However, IIRC, TuningParamSetKind::Exhaustive was introduced to make all configs run through applicability.
Now this changes that to be a no op (for this release) because couldBePerformant is empty. See : 

https://github.com/ROCm/rocMLIR/blob/5261f9fb8cc5e0109cf59bbf4cb11cc6fb527209/mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp#L215-L221

Having said that Im onboard with the change on the premise TuningParamSetKind::Exhaustive is nearing the end of its usefulness for that purpose and I can see it being useful in this way.

krzysz00: No, `::Exhaustive` means to skip `couldBePerformant()`

You had it skipping "is this even a valid perf config"?

krzysz00: This also makes it match all the other cases, which always check `paramsProbablyValid()`

PR #1544, opened 2024-06-06
Unify accelerator and non-accelator paths

djramic:
This PR is created for tracking the unification paths ticket.
The current commit is focused on the threadwise unification.
_NormalizeView_ and _TransformingForOp_ components remain non-unified(hardcoded).
For normalizeView, the kPack parameter is problematic, whereas for TransfromingForOp the differing indexing methond is the concern.
I am currently investigating whether they should be integrated through the AccelEmitter or if there is a possibility for refactoring them through a blockwise path unification.

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/IR/AccelEmitter.h
+++ b/mlir/include/mlir/Dialect/Rock/IR/AccelEmitter.h
@@ -25,6 +25,7 @@
 #ifndef MLIR_LIB_DIALECT_ROCK_TRANSFORMS_MLIR_ACCEL_EMITTER_H
 #define MLIR_LIB_DIALECT_ROCK_TRANSFORMS_MLIR_ACCEL_EMITTER_H
 
+#include "mlir/Dialect/Rock/IR/FmaInsnGroup.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/Rock/IR/MfmaInsnGroup.h"
 #include "mlir/Dialect/Rock/IR/TransformMapBuilder.h"
@@ -86,7 +87,7 @@ struct AccelEmitter {
   /// Select the right accelerator based on the set of features and architecture
   static std::unique_ptr<AccelEmitter>
   select(GemmFeatures features, Type dataTypeA, Type dataTypeB, StringRef arch,
-         RockAccelTuningParamAttrInterface tuningParams);
+         RockTuningParamAttrInterface tuningParams);
 
   /// Emit the actual intrinsic in the threadwise operation
   virtual void emitThreadwiseLoop(OpBuilder &b, Location loc, Value argA,
@@ -150,15 +151,15 @@ struct AccelEmitter {
 
   virtual ~AccelEmitter() {}
 
-  enum AccelEmitterKind { AEK_MFMAEmitter, AEK_WMMAEmitter };
+  enum AccelEmitterKind { AEK_MFMAEmitter, AEK_WMMAEmitter, AEK_FMAEmitter};
 
   AccelEmitterKind getKind() const { return kind; }
 
 protected:
-  AccelEmitter(StringRef arch, RockAccelTuningParamAttrInterface tuningParams,
+  AccelEmitter(StringRef arch, RockTuningParamAttrInterface tuningParams,
                AccelEmitterParams accelEmitterParams, AccelEmitterKind kind);
 
-  RockAccelTuningParamAttrInterface tuningParams;
+  RockTuningParamAttrInterface tuningParams;
   AccelEmitterParams accelEmitterParams;
   int64_t waveSize;
 
@@ -170,7 +171,7 @@ private:
 struct MfmaEmitter : public AccelEmitter {
 
   MfmaEmitter(MfmaInsnGroup mfmaGroup, StringRef arch,
-              RockAccelTuningParamAttrInterface tuningParams);
+              RockTuningParamAttrInterface tuningParams);
 
   void emitThreadwiseLoop(OpBuilder &b, Location loc, Value argA, Value argB,
                           Value bufferC, ValueRange regCOffset) override;
@@ -208,7 +209,7 @@ private:
   /// Initialize the emitter parameters for mfma
   AccelEmitterParams
   initAccelEmitterParams(MfmaInsnGroup mfmaGroup,
-                         RockAccelTuningParamAttrInterface tuningParams);
+                         RockTuningParamAttrInterface tuningParams);
 
   MfmaInsnGroup mfmaGroup;
 };
@@ -217,7 +218,7 @@ private:
 struct WmmaEmitter : public AccelEmitter {
 
   WmmaEmitter(WmmaInsn wmmaInsn, StringRef arch,
-              RockAccelTuningParamAttrInterface tuningParams);
+              RockTuningParamAttrInterface tuningParams);
 
   void emitThreadwiseLoop(OpBuilder &b, Location loc, Value argA, Value argB,
                           Value bufferC, ValueRange regCOffset) override;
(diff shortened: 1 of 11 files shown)


krzysz00:
I'll note that kpack should actually be used in the non-accelerator path

krzysz00:
Like, while you're here, the FMA/dot product-based path should absolutely handle kpack

And if there isn't room for it in the perf config, a new perf config version might be in order

mlir/include/mlir/Dialect/Rock/IR/AccelEmitter.h
```diff
@@ -249,11 +250,50 @@ struct WmmaEmitter : public AccelEmitter {
...
+
+struct FmaEmitter : public AccelEmitter {
+
+  FmaEmitter(FmaInsn fmaInsn, StringRef arch,
+             RockTuningParamAttrInterface tuningParams);
+
+  void emitThreadwiseLoop(OpBuilder &b, Location loc, Value argA, Value argB,
+                          Value bufferC, ValueRange regCOffset) override;
+
+ virtual Value
```
krzysz00, line 269: You don't need to repeat `virtual` here

mlir/include/mlir/Dialect/Rock/IR/FmaInsnGroup.h
```diff
@@ -0,0 +1,39 @@
...
+  public:
+    static FailureOr<FmaInsn> select(Type elementTypeA, Type elementTypeB, StringRef arch);
+};
+
+
+
+} // namespace rock
+} // namespace mlir
+
+#endif // MLIR_FMA_INSN_GROUP_H
```
krzysz00, line 39: Nit: newline

krzysz00:
The fundamental observation, IMO, is that a a FMA is a `1 x kbase * kbase x 1` multiplication (I say K here because this could also be easily extended to dot products, so you'd have, for example, kbase = 2 for fp16.

Then, with that, you can use the general accelerator lowering, which is pretty independent of exactly what the accelerator does.

(For practical reasons, you might want to announce the accelerator as mPerThread x kbase x nPerThread ... but also, mRepeats and nRepeats exist for a reason and ought to cover most of this)

djramic:
I plan to start by unifying the lowering paths in the way they are currently implemented, in order to get a better understanding of each lowerting level. And to make the verification easier, using existing tests. After that, I can make make additional refactorisation and improvement(We can open a separate issue based on you comments above.) I spirit of my current aproach do you have any aditional feedback? I would wery much appreciate it. I'm not quite sure I'm on the right path.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1229,6 +1232,21 @@ def Rock_ThreadwiseAccelGemmOp:
...
+// threadwise_gemmv2
+def Rock_ThreadwiseGemmOpv2:
+    Rock_Op<"threadwise_gemmv2">,
+    Arguments<(ins Arg<MemRefOf<NativeMemoryOpTypes>, "source register view A", [MemRead]>:$matrixA,
+                   Arg<MemRefOf<NativeMemoryOpTypes>, "source register view B", [MemRead]>:$matrixB,
+                   Arg<MemRefOf<NativeMemoryOpTypes>, "dest register view C", [MemRead, MemWrite]>:$matrixC, Variadic<Index>:$computeIndices,
+                   StrAttr:$arch,
+                   Rock_GemmFeaturesAttr:$features,
+                   RockTuningParamAttrInterface:$params)> {
+  let assemblyFormat = [{
```
krzysz00, line 1244: This is a temporary thing, right? Why's this needed?

mlir/lib/Dialect/Rock/utility/AccelEmitter.cpp
```diff
@@ -130,27 +131,110 @@ Value AccelEmitter::generateThreadwiseViewBufferC(PatternRewriter &b,
   return viewC;
 }
 
+// **************************
```
krzysz00, line 134: Weird comment format

mlir/lib/Dialect/Rock/utility/AccelEmitter.cpp
```diff
@@ -130,27 +131,110 @@ Value AccelEmitter::generateThreadwiseViewBufferC(PatternRewriter &b,
...
+
+    //TO-DO
+
+    return params;
+}
+
+void FmaEmitter::emitThreadwiseLoop(OpBuilder &b, Location loc, Value argA, Value argB,
+                          Value bufferC, ValueRange regCOffset){
+
+      Type dataType = fmaInsn.argTypeA;
```
krzysz00, line 161: I think there's a `git clang-format main` in order

mlir/lib/Dialect/Rock/utility/AccelEmitter.cpp
krzysz00: ... though that's file style. Maybe worth fixing to the `//=---------------=//` style seen elsewhere in the codebase both here and below

mlir/lib/Dialect/Rock/utility/AccelEmitter.cpp
```diff
@@ -130,27 +131,110 @@ Value AccelEmitter::generateThreadwiseViewBufferC(PatternRewriter &b,
...
+  //TO-DO
+    }
+
+RegsAsMatrixSubTiles FmaEmitter::computeOutputTransforms(
+      OpBuilder &b, Location loc, int64_t mLen, int64_t nLen, int64_t blockSize,
+      ArrayRef<int64_t> bidGridLengths, int64_t inMPerThread,
+      int64_t inNPerThread, bool doSwapThreadIterSubDimsForM,
+      bool doSwapThreadIterSubDimsForN){
+
+  //TO-DO
```
krzysz00, line 211: ... Weird

krzysz00: This should be reasonably straightforward?

mlir/lib/Dialect/Rock/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -89,6 +89,168 @@ struct ThreadwiseReadIntoRewritePattern
...
+  TransformMapAttr normalizeView(OpBuilder &b, Location loc,
+                                 ArrayRef<StringRef> names,
+                                 ArrayRef<int64_t> sizes, bool ignoreStartShape,
+                                 DenseSet<StringRef> ignoreNames) const {
+    unsigned pos = 0;
+    unsigned newPos = 0;
+    TopDownTMBuilder td(b, names, sizes, loc);
+    // Convert the normalizedView to a real view by ignoring
+    // the names contained in `ignoreNames` and letting the rest pass through
+    for (pos = 0; pos < names.size(); pos++) {
```
krzysz00, line 108: Nit: move `unsigned pos` into the for loop.

... bigger nit: `auto [pos, name] : llvm::enumerate(names))`

mlir/lib/Dialect/Rock/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -89,6 +89,168 @@ struct ThreadwiseReadIntoRewritePattern
...
+    auto [rawBufferC, bufferViewC, dstNeeds64BitIdx] =
+          untransform(b, bufferC, normalizedViewC);
+    assert(!sourceANeeds64BitIdx && "Registers shouldn't need 64-bit indexing");
+    assert(!sourceBNeeds64BitIdx && "Registers shouldn't need 64-bit indexing");
+    assert(!dstNeeds64BitIdx && "Registers shouldn't need 64-bit indexing");
+
+    // Loop properties
+    auto computeStart = llvm::to_vector(op.getComputeIndices());
+
+    if(isMfma || isWmma){
```
krzysz00, line 195: Why? Don't WMMA and MFMA go down `threadwise_gemm`?

mlir/lib/Dialect/Rock/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -89,6 +89,168 @@ struct ThreadwiseReadIntoRewritePattern
...
+          loc, ArrayRef<ValueRange>{computeStart, computeStart, computeStart},
+          ArrayRef<Attribute>{bufferViewA, bufferViewB, bufferViewC}, dimensions,
+          /*strides=*/std::nullopt, /*forceUnroll=*/true,
+          /*useIndexDiffs=*/false);
+
+      {
+        OpBuilder::InsertionGuard guard(b);
+        b.setInsertionPointToStart(gemmLoop.getBody());
+
+        auto coordsA = gemmLoop.getLowerCoords(/*domain=*/0);
```
krzysz00, line 238: Why do you need this case? What prevents you from using `emitAccelLoop` here?

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
djramic: The idea is that this OP will replace threadwise_gemm and threadwise_accel_gemm. I am creating a new operation so I don't have to delete the existing code.

mlir/lib/Dialect/Rock/Transforms/ThreadwiseGemmLowering.cpp
djramic: This is temporary for parts that are not unified yet. The goal is to have it removed at the end of the unification paths.

mlir/lib/Dialect/Rock/Transforms/ThreadwiseGemmLowering.cpp
djramic: The problem is caused by different ways of creating gemmLoop and accelLoop TransformingForOp. I want to try to equalize these approaches by refactoring the blockwise level.

causten:
No longer required.

PR #1545, opened 2024-06-06
[backport] Add check for block size being 1024 or less

jerryyin:
https://github.com/ROCm/rocMLIR/pull/1543


manupak:
There is only one additional windows cmake fix commit in develop.. We might just want to use develop as 6.2 ?

jerryyin:
@manupak Done, I just made release branch caught up with develop

PR #1546, opened 2024-06-06
Use early-increment range for loop through uses that may erase an operation.

pcf000:
Observed a use-after-free error when building with -DLLVM_USE_SANITIZER=Address.  Looks like a node is being erased and its uses erased with it, before the loop has advanced to the next use.  The early-increment range exists to avoid that problem.


krzysz00 (approved):
Good catch, approved!

PR #1548, opened 2024-06-11
[rock] Switch to target attributes from serialization

fabianmcg:
This PR migrates rocMLIR to use GPU target attributes for compilation instead of the deprecated and removed `SerializeToHSA` pass.

Instead of `SerializeToHSA` the kernel pipeline now uses the passes `GpuROCDLAttachTarget` and `GpuModuleToBinaryPass`.

In addition, this PR also introduces a custom `gpu::TargetAttrInterface` for the
`ROCDLTargetAttr` attribute. This interface overrides the default behavior of:
`mlir::ROCDL::registerROCDLTargetInterfaceExternalModels`
By adding:
 - rocMLIR specific metadata in the binary.
 - Allowing to package device bitcode libs inside the rocMLIR binaries.
 - Use LLD API instead of calling the executable.

To register this interface one must call: `mlir::registerRocTarget`.

The idea of this custom interface is to minimize collisions with upstream, allowing to more customization on the rocMLIR side.

As consequence of this change, several parts of the infrastructure were updated to reflect the usage of `gpu.binary` ops intead of a GPU module with a binary string.


external/llvm-project/mlir/include/mlir/Target/LLVM/ROCDL/Utils.h
```diff
@@ -31,6 +31,15 @@ StringRef getROCMPath();
 /// strings. By default this class serializes into LLVM bitcode.
 class SerializeGPUModuleBase : public LLVM::ModuleToObject {
 public:
+  /// AMDGCN standard libraries.
+  typedef enum : uint32_t {
```
krzysz00: Could be an `enum class`?

krzysz00 (commented):
This seems pretty reasonable overall

external/llvm-project/mlir/include/mlir/Target/LLVM/ROCDL/Utils.h
fabianmcg: Sure, I can make it an enum class, the reason I didn't make it an enum class is because it's already scoped inside the `SerializeGPUModuleBase`, but I'll take it out of the class and make it an enum class.

krzysz00: ... Good point, it probably doesn't need to be an enum class because of the scoping.

I forget what upstream style on this sort of thing is

krzysz00:
Re the PR, could you split the change into commits that only touch external/llvm-project, external/mlir-hal, and everything else?

fabianmcg:
> Re the PR, could you split the change into commits that only touch external/llvm-project, external/mlir-hal, and everything else?

Yes, I'll do it like that.

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
```diff
@@ -221,10 +221,14 @@ void rock::buildBackendPipeline(OpPassManager &pm,
   llvmFuncPm.addPass(createCSEPass());
   llvmFuncPm.addPass(rock::createRockPrepareLLVMPass());
   if (options.compile) {
-    gpuPm.addPass(createGpuSerializeToHsacoPass(
-        options.triple, options.chip, options.features, options.optLevel,
-        options.suppressDiagnostic));
-    gpuPm.addPass(createRockCheckResidencyPass());
+    GpuROCDLAttachTargetOptions opts;
```
krzysz00, line 224: Note to future us: I wonder if it might be interesting to move this target attachment early in the pipeline

krzysz00 (commented):
I've got a few comments, haven't properly read the new translation widget yet

krzysz00 (commented):
Overall, the internal changes seem reasonable - I take it the copy-paste is somewhat inevitable here?

I'll ping @sjw36 for the MHAL changes

fabianmcg:
> I take it the copy-paste is somewhat inevitable here?

Are you referring to the LLVM upstream patches? or the Target attribute in rocMLIR?

manupak:
Only thing blocking the CI is fixed in https://github.com/ROCm/rocMLIR/commit/da8969573d2ad408c7ad129126679838d82d9350

Can you merge/rebase ?

manupak:
This PR has changes to two external repos and therefore it *should not* be squashed and merged.
Thus 22 commits is not something I d like to see in the tree.

Can you squash the commits to a sensible number while keeping the commits to each repo seperate.

fabianmcg:
I could potentially reduce it to 3 commits (llvm, mhal and roc), do we all agree on that?

manupak:
3 works for me; one each

krzysz00 (approved):
Approved here

PR #1549, opened 2024-06-12
Support for parsing options from ROCMLIR_DEBUG_FLAGS in library calls like MIGraphX

pcf000:
Call the necessary bits to initialise and parse command-line options from an environment variable, and do it from the MIGraphX interface functions.  Design decision request:  require a patch in MIGraphX that is slightly cleaner, or do it all in rocMLIR by being slightly unclean.

### DIFF
--- a/mlir/lib/CAPI/Dialect/MIGraphX.cpp
+++ b/mlir/lib/CAPI/Dialect/MIGraphX.cpp
@@ -134,7 +134,8 @@ MLIR_CAPI_EXPORTED bool mlirGetBinary(MlirModule module, size_t *size,
 MLIR_CAPI_EXPORTED
 void mlirMIGraphXAddHighLevelPipeline(MlirPassManager pm) {
   auto passMan = unwrap(pm);
-  applyPassManagerCLOptions(*passMan);
+  if (failed(applyPassManagerCLOptions(*passMan)))
+    llvm::errs() << "Failed to apply command-line options.\n";
   passMan->setNesting(mlir::PassManager::Nesting::Implicit);
   mlir::migraphx::addHighLevelPipeline(*passMan);
   mlir::rock::buildBufferizePipeline(*passMan);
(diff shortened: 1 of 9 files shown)


mlir/lib/CAPI/Dialect/MIGraphX.cpp
```diff
@@ -134,6 +134,7 @@ MLIR_CAPI_EXPORTED bool mlirGetBinary(MlirModule module, size_t *size,
 MLIR_CAPI_EXPORTED
 void mlirMIGraphXAddHighLevelPipeline(MlirPassManager pm) {
   auto passMan = unwrap(pm);
+  applyPassManagerCLOptions(*passMan);
```
pcf000: Have to call this one on the pass manager, therefore can't do it ahead of time.

mlir/lib/CAPI/RegisterRocMLIR/RegisterRocMLIR.cpp
```diff
@@ -10,11 +10,24 @@
...
 #include "mlir/InitRocMLIRPasses.h"
+#include "llvm/Support/CommandLine.h"
 
 void mlirRegisterRocMLIRDialects(MlirDialectRegistry registry) {
   mlir::registerRocMLIRDialects(*unwrap(registry));
 }
 
 void mlirRegisterRocMLIRPasses() { mlir::registerRocMLIRPasses(); }
+
+void mlirRegisterRocMLIROptions() {
```
pcf000, line 24: The open question is, to call this function from MIGraphX, in mlir.cpp where we also call the other two register functions in this file, or call it from one of the other two so we don't have to patch the other project.

krzysz00: I'd say patch MIGraphX.

mlir/lib/CAPI/RegisterRocMLIR/RegisterRocMLIR.cpp
```diff
@@ -10,11 +10,24 @@
...
+  const char *fakeArgv[] = {"rocMLIR-invoked-as-library",
+                            "--mlir-print-local-scope"};
+  mlir::registerAsmPrinterCLOptions();
+  mlir::registerMLIRContextCLOptions();
+  mlir::registerPassManagerCLOptions();
```
krzysz00: Could you extract these `mlir::register*CLOptions()` into a separate function and then also call that in all the various `mlir/tools/rocmlir-*/`s (except perhaps rocmlir-translate)? Might be a useful cleanup while we're here.

krzysz00 (approved):
No fundamental questions but a request

PR #1550, opened 2024-06-12
Pass config.rocmlir_gen_flags to get_arch_features to turn off support_mfma when '-mfma=off' is in the flags.

pcf000:
Design check:  should we pass the flags in to get_arch_features to turn the flag off directly, or should we test for -mfma=off in flags in the various lit.local.cfg files that will care?  This is for issue #1495;  its purpose is to stop the failures in the public CI, ml-ci.amd.com:21096.

### DIFF
--- a/mlir/test/common_utils/common.py
+++ b/mlir/test/common_utils/common.py
@@ -2,7 +2,7 @@ import subprocess
 
 # Helper function to decode arch to its features
 # Keep this in sync with mlir/lib/Dialect/Rock/Generator/AmdArchDb.cpp:mlir::rock::lookupArchInfo
-def get_arch_features(arch: str):
+def get_arch_features(arch: str, flags=""):
     chip_name = arch.split(':')[0]
     if(len(chip_name) < 5):
         return
@@ -29,7 +29,7 @@ def get_arch_features(arch: str):
     elif major == 'gfx11':
         arch_features = 'dot|atomic_add|atomic_fmax_f32|wmma'
     if arch_features and 'mfma' in arch_features:
-        support_mfma = True
+        support_mfma = not '-mfma=off' in flags
         pass
     elif arch_features and 'wmma' in arch_features:
         support_wmma = True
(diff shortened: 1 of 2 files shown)


krzysz00:
What we should do is stop setting `-mfma=off` explicitly in flags. rocmlir_gen_flags is a debug feature.

pcf000:
> What we should do is stop setting `-mfma=off` explicitly in flags. rocmlir_gen_flags is a debug feature.

We set it if the can-xdlops Jenkins parameter is turned off;  explicitly in Jenkinsfile.downstream;  and in parameterSweeps.py.  The first two came from Lixun in October 2022.  I could see arguing that it was left in Jenkinsfile.downstream by accident and should be removed, and indeed that's the one that causes the spurious public-CI failures in pull requests.  I can also see calling parameterSweeps.py a debugging tool.

But what about the can-xdlops flag?  Do we still have to worry about (a) non-xdlops targets (b) being able to turn off xdlops?

krzysz00:
The point of the `canXdlops` flag is to let us use the same Jenkinsfile on public and private CI. We probably should just rename it to `isExternalCI` and tell people to never touch it.

And that flag shouldn't pass in `-mfma=off`, I claim

pcf000:
I could swear I saw something that directly controlled -mfma= using canXdlops, but now I can't find it.  Maybe I'm losing it, maybe it's Friday.

Anyway, this problem may take care of itself:  it seems there is nothing in public CI lower than gfx90a, and we've already announced targets with mfma, so maybe we should just remove the ROCMLIR_GEN_FLAGS from Jenkinsfile.downstream (which is only used by pull requests).

krzysz00:
Yeah, we should do that

pcf000:
Lixun is fine with removing ROCMLIR_GEN_FLAGS from Jenkinsfile.downstream, which makes this whole pull request moot.  I'll kill it and make another with just the removal.

PR #1553, opened 2024-06-17
Also link ctest when installing cmake, since MIGraphX testing uses it.

pcf000:
The cmake build that I copied from llvm-premerge-checks installs into /etc/cmake/bin and then links to /usr/bin.  However, it doesn't like ctest, and verify-cmake-with-mlir barfed in the nightly.

Added the link, as the simplest change.  Could also use the ubuntu repo (cmake version 3.22.1) or the kitware repo with the necessary key, if one of those is preferred.


krzysz00 (approved):

jerryyin (approved):
Let's get this merge by the end of today so nightly can pick it up.

PR #1555, opened 2024-06-18
[DO NOT SQUASH] Remove mhal.prefill attr's tight coupling to `rock` dialect

manupak:
Currently, our lowering has a tight coupling with `rock` and `mhal` dialect
only via *prefill* attr. Now, this prohibits pulling in just the `rock` dialect
to third party projects without requiring pulling `mhal` dependency, especially
where the third part project has it own hal. Idea is we should be able to 
pull `rock` dialect to lower to upstream dialects.

This PR introduces a same concept of *prefill* attr in `rock` dialect and moves
the creation of `mhal` prefill attr to a conversion.

### DIFF
--- a/external/mlir-hal/include/mlir/Dialect/MHAL/Utility/Utils.h
+++ b/external/mlir-hal/include/mlir/Dialect/MHAL/Utility/Utils.h
@@ -1,0 +1,29 @@
+//=== Utils.h - functions that often come up during lowering
+//---------------===//
+//
+// Part of the MLIR Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+#ifndef MHAL_UTILITY_UTILS_H
+#define MHAL_UTILITY_UTILS_H
+
+#include "mlir/Dialect/MHAL/IR/MHAL.h"
+
+namespace mlir {
+
+namespace LLVM {
+class LLVMFuncOp;
+}
+
+namespace mhal {
+
+class PrefillAttr;
+
+// Return `mhal::PrefillAttr` attributes for a given function
+SmallVector<PrefillAttr> getStoredPrefillAttributes(LLVM::LLVMFuncOp func);
+
+} // namespace mhal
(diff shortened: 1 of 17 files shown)


ravil-mobile:
As far as I remember, the `prefill` was added to mhal because it fills in the output buffers with zero in the case of the split-k; it was needed for testing rocMLIR generated GEMM kernels.

We didn't want to create a special case in `rocmlir-gen` for split-k; we wanted to find a generic solution at that time. The solution, which we came up with, was to delegate the logic (see above) to mhal

manupak:
Yes, that aligns with my thinking, however, we should not have injected mhal concepts so much early into our lowering where as in we could carry them in rock and convert later on -- keeping the dialects a bit cleaner.

ravil-mobile:
> Yes, that aligns with my thinking, however, we should not have injected mhal concepts so much early into our lowering where as in we could carry them in rock and convert later on -- keeping the dialects a bit cleaner.

Maybe we can keep `prefill` in both rocMLIR and MHAL, and then (down the pipeline) make an attribute conversion - i.e., from `rock::prefill` to `mhal::prefill`. What do you think?

manupak:
> > Yes, that aligns with my thinking, however, we should not have injected mhal concepts so much early into our lowering where as in we could carry them in rock and convert later on -- keeping the dialects a bit cleaner.
> 
> Maybe we can keep `prefill` in both rocMLIR and MHAL, and then (down the pipeline) make an attribute conversion - i.e., from `rock::prefill` to `mhal::prefill`. What do you think?

Thats what is done in this PR :) . Note in RockToGPU the conversion happens.

ravil-mobile (approved):

ravil-mobile:
> > > Yes, that aligns with my thinking, however, we should not have injected mhal concepts so much early into our lowering where as in we could carry them in rock and convert later on -- keeping the dialects a bit cleaner.
> > 
> > 
> > Maybe we can keep `prefill` in both rocMLIR and MHAL, and then (down the pipeline) make an attribute conversion - i.e., from `rock::prefill` to `mhal::prefill`. What do you think?
> 
> Thats what is done in this PR :) . Note in RockToGPU the conversion happens.

Sorry, didn't noticed. LGTM

krzysz00 (approved):

PR #1557, opened 2024-06-28
Temporarily call registerMLIRCLOptions() from mlirRegisterRocMLIRPasses(), until we can call it from MIGraphX.

pcf000:
Short-term measure to get nightlies passing again.  test_gpu_mlir and the rest of "make check" all pass.


krzysz00 (approved):
Can you get rid of the `if(failed` we mentioned in chat earlier as well?

But also, approved

manupak (approved):
I can confirm this fixes the nightly! Thanks

PR #1558, opened 2024-07-02
Add `-dump-asm-to` to rocmlir-driver

manupak:
BUILD ON TOP OF : https://github.com/ROCm/rocMLIR/pull/1548

This commit adds support for rocmlir-driver
to dump assembly of the otherwise generated
code object to a user given filepath.

closes : https://github.com/ROCm/rocMLIR-internal/issues/1513


manupak:
This was done to aid with ATT vieweing but now the ATT view can work in "auto"mode where it doesn;t need to provided with an assembly file.
Moreover, we can get to these easily with `-debug-only=serialize-to-isa` option.

PR #1560, opened 2024-07-08
Rearrange alloc_tensor creation and CSE to fix input fusion crash

krzysz00:
Fixes https://github.com/ROCm/rocMLIR-internal/issues/1520

As mentioned in the comments in the code below, this PR fixes an issue that areses because `tensor.empty` operations were combined in a way that caused the regularization pass to trip an assert.

To put it another way, the rocMLIR fusion code relies on never needing to try to "fuse"

%a = alloc()
f(reads ..., writes %a)
%b = alloc()
g(reads ... + %a, writes %b)
// No other uses of %a until
h(reads ... + %b, writes %a)

This can happen if you combine two tensor.empty ops and the buferization analysis notices that you don't actually need two buffers for the two different empty tensors.

Therefore, to fix the assert, as a workaround, copy the translation from tensor.empty to bufferization.alloc_tensor before the call to CSE. We leave the existing call to that pass in place just to be safe.

This has a side effect of increasing the number of generics in the tosa transpsoe optimiization test.

### DIFF
--- a/mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
+++ b/mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
@@ -93,7 +93,7 @@ void rock::buildBufferizePipeline(OpPassManager &pm,
   funcPm2.addPass(rock::createRockFoldBroadcastPass());
 
   // bufferization
-  /* rocmlir-opt --canonicalize --cse -convert-tensor-to-linalg
+  /* rocmlir-opt --canonicalize -convert-tensor-to-linalg --cse
         --one-shot-bufferize="allow-return-allocs=1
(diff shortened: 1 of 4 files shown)


manupak:
Thanks for the quick fix.

Can we have a test following is not broken ?
(Because it might be easy for this to be broken again)

//, they can be CSE'd together, and then, if the bufferizer notices that the
// allocation that that empty tensor has two independent uses (that is,
// if op1 and op2 both have the "initial output" %x, and the values produces
// by op1 are dead by the time op2 rolls around), it'll reuse the buffer.
// This breaks rocMLIR's fusion code, which assumes allocations aren't reused
// like this.

manupak (approved):
LGTM!

PR #1562, opened 2024-07-09
Gfx12 support in rocMLIR

giuseros:
This is a  PR to add rocMLIR support to gfx12. Main changes in this PR
- In order for gfx12 to work, we need to fix a bug in the backend compiler (commit 1)
- Instead of issuing a `WaitcntOp(0);SBarrierOp()` (pre-gfx12) we need to issue: `WaitDscntOp(0); BarrierSignalOp(-1); BarrierWaitOp(-1);`. This happens in commit 2. 
- We need to update the wmma layout in the `AccelEmitter` class to comply to the new layout. This is commit 3.

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
+++ b/external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
@@ -89,11 +89,11 @@ class ROCDL_IntrPure1Op<string mnemonic> :
 
 class ROCDL_IntrOp<string mnemonic, list<int> overloadedResults,
   list<int> overloadedOperands, list<Trait> traits, int numResults,
-  int requiresAccessGroup = 0, int requiresAliasAnalysis = 0> :
+  int requiresAccessGroup = 0, int requiresAliasAnalysis = 0, list<int> immArgPositions = []> :
   LLVM_IntrOpBase<ROCDL_Dialect,  mnemonic,
     "amdgcn_" # !subst(".", "_", mnemonic), overloadedResults,
     overloadedOperands, traits, numResults, requiresAccessGroup,
-    requiresAliasAnalysis>;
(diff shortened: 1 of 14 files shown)


external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
```diff
@@ -247,6 +247,30 @@ def ROCDL_BarrierOp : ROCDL_Op<"barrier"> {
...
 
+def ROCDL_BarrierSignalOp : ROCDL_IntrOp<"s.barrier.signal", [], [], [], 0>,
+  Arguments<(ins I32Attr:$id)> {
+  let results = (outs);
+  let assemblyFormat = "$id attr-dict";
+  string llvmBuilder =
+    "createIntrinsicCall(builder, llvm::Intrinsic::amdgcn_s_barrier_signal,builder.getInt32(op.getId()));";
+}
+
+def ROCDL_BarrierWaitOp : ROCDL_IntrOp<"s.barrier.wait", [], [], [], 0>,
```
krzysz00, line 256: I'm going to note that `LLVM_IntrOpBase` from upstream has `immArgPositions`, which could be used to avoid constructing a builder yourself and allowing import from LLVM as well.

krzysz00 (approved):
Approved with the note from the other review and assuming that this is split into three commits for easy upstreaming

external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td
giuseros: Nice tip, thanks!

giuseros:
FYI, this is passing on my local gfx11 board (and I am not touching anything mfma related) - cc @jerryyin

jerryyin:
@giuseros  Please rebase. We fixed tuning and it shouldn't timeout on Navi3 any more.

PR #1564, opened 2024-07-09
Remove gfx906 from Jenkinsfile.  Most importantly, don't force gfx906 for vanilla codepath.

pcf000:
Forcing gfx906 for the vanilla codepath is what's causing the hangs on the public CI.

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -183,7 +183,7 @@ String getLabelFromCodepath(String codepath) {
         // and server nodes (gfx1030)
         label = 'mlir && ( gfx1030w || gfx1030 )'
     } else if (codepath == "vanilla"){
-        label = 'mlir && gfx906'
+        label = 'mlir'
     } else if (codepath == "navi3x") {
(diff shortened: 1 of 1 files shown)


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -183,7 +183,7 @@ String getLabelFromCodepath(String codepath) {
         // and server nodes (gfx1030)
         label = 'mlir && ( gfx1030w || gfx1030 )'
     } else if (codepath == "vanilla"){
-        label = 'mlir && gfx906'
+        label = 'mlir'
```
krzysz00, line 186: Given that both `gfx900` and `gfx906` don't exist in CI anymore, let's get rid of "vanilla" all together

pcf000: Jerry wants to keep it.

krzysz00: Why so? It's tested by navi2, no?

krzysz00 (approved):
Comment above aside, the broader removal of `vanilla` isn't a prerequisite for landing this, so it's approved

PR #1566, opened 2024-07-10
Implement emulated FP8 for the OCP formats.

pcf000:
Derived from migraphx cast_to_f8<> with appropriate parameters and some code tweaks.

Details from https://www.opencompute.org/documents/ocp-8-bit-floating-point-specification-ofp8-revision-1-0-2023-12-01-pdf-1

Rather than try to make the existing function do all the known formats, we do OCP with a second function.

Updated the tests to fit.  Hack needed for E5M2 NaNs, because APFloat forces signaling NaNs to quiet and breaks reversibility.


mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
```diff
@@ -31,6 +31,9 @@
 #include "llvm/ADT/TypeSwitch.h"
 #include "llvm/Support/raw_ostream.h"
 
+#include <bit>
```
pcf000: May not need these now, I'll double-check.

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
```diff
@@ -119,7 +127,18 @@ static Value getFloatValueTableFor(Type elementType, Operation *op,
   const auto &sem = type.getFloatSemantics();
   for (uint32_t i = 0; i < 256; ++i) {
     APFloat entry(sem, APInt(8, i));
-    tableElems.push_back(entry.convertToFloat());
+    float x = entry.convertToFloat();
```
pcf000, line 127: Originally I broke this up so I could print the table as it was generated, but then it became handy for the ugly hack.

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
```diff
@@ -365,14 +384,264 @@ static FlatSymbolRefAttr makeFp8TruncFunction(Location loc, FloatType outType,
   return symbolRef;
 }
 
+// Float8E5M2 and Float8E4M3FN
+static FlatSymbolRefAttr makeOCPFp8TruncFunction(Location loc, FloatType outType,
```
pcf000: New function, following the pattern of the other one.  These two types fit the OCP spec, which was handy.

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
```diff
@@ -365,14 +384,264 @@ static FlatSymbolRefAttr makeFp8TruncFunction(Location loc, FloatType outType,
...
+  uint32_t eBits = outType.getWidth() - 1 - mBits;
+  Value mWidth = i32Const(mBits);
+  Value eWidth = i32Const(eBits);
+
+  // Created here so we can branch to it, will be inserted last
+  Block *ret = new Block();
+  ret->addArgument(i8, loc);
+
+  Value bits = b.create<BitcastOp>(i32, in);
+  Value and22 = b.create<AndIOp>(bits, i32Const((1u << 23u) - 1));
```
pcf000, line 433: Numbers on the variables are the %-numbers from the mlir.

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
```diff
@@ -365,14 +384,264 @@ static FlatSymbolRefAttr makeFp8TruncFunction(Location loc, FloatType outType,
...
+
+  b.setInsertionPointToStart(bb1);
+  Value cmp37 = b.create<CmpIOp>(CmpIPredicate::eq, mWidth, i32Const(3));
+  b.create<cf::CondBranchOp>(cmp37, bb2, bb3);
+
+  b.setInsertionPointToStart(bb2);
+  Value trunc38 = b.create<TruncIOp>(i8, add33);
+  b.create<cf::BranchOp>(ret, trunc38);
+
+  // This block is a later edit, thus the numbers don't fit with bb6, etc.
```
pcf000, line 464: The later edit was to get e5m2 NaNs right.

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
```diff
@@ -365,14 +384,264 @@ static FlatSymbolRefAttr makeFp8TruncFunction(Location loc, FloatType outType,
...
+  b.create<cf::CondBranchOp>(cmp66, bb10, bb11);
+
+  b.setInsertionPointToStart(bb10);
+  Value min67 = b.create<MinUIOp>(bb9arg1, i32Const(31));
+  Value shr68 = b.create<ShRUIOp>(bb9arg2, min67);
+  b.create<cf::BranchOp>(bb12, ValueRange{shr68});
+
+  b.setInsertionPointToStart(bb11);
+  Value cmp69 = b.create<CmpIOp>(CmpIPredicate::eq, bb9arg1, i32Const(-1));
+  Value zext70 = b.create<ExtUIOp>(i32, cmp69);
```
pcf000, line 538: I believe that ExtUIOp turns into the llvm.zext I saw in the clang output (and similarly with sext and ExtSIOp) but I didn't trace the entire progression to confirm it.

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
```diff
@@ -365,14 +384,264 @@ static FlatSymbolRefAttr makeFp8TruncFunction(Location loc, FloatType outType,
...
+  Value add76 = b.create<AddIOp>(sub75, bb9arg1);
+  Value add77 = b.create<AddIOp>(add76, or74);
+  Value shl78 = b.create<ShLIOp>(i32Const(1), sub56);
+  Value add79 = b.create<AddIOp>(shl78, i32Const(-1));
+  Value and80 = b.create<AndIOp>(bb12arg0, shl78);
+  Value cmp81 = b.create<CmpIOp>(CmpIPredicate::eq, and80, i32Const(0));
+  Value select82 = b.create<SelectOp>(cmp65, cmp81, i1Const(false));
+  Value sext83 = b.create<ExtSIOp>(i32, select82);
+  Value add84 = b.create<AddIOp>(bb12arg0, sext83);
+  Value stoch = i1Const(false);         // Defaulted arguments.
```
pcf000: The C++ cast_to_f8<> that I used had stoch (for stochastic rounding) and rng (value used for that) as parameters with default values.  OCP spec mandates round-to-nearest-even but allows other rounding modes.  Hard-code the defaults here;  if we want to extend it, this is where to hook in the control.

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
```diff
@@ -365,14 +384,264 @@ static FlatSymbolRefAttr makeFp8TruncFunction(Location loc, FloatType outType,
...
+
+  return symbolRef;
+}
+
 LogicalResult Fp8TruncToCallPattern::match(TruncFOp op) const {
   if (failed(canBeConverted(op.getResult().getType())))
     return failure();
   Type resType = getElementTypeOrSelf(op.getOut().getType());
-  if (resType.isFloat8E4M3FN() && !f8E4M3FNUZFunc)
+  if (resType.isFloat8E4M3FNUZ() && !f8E4M3FNUZFunc)
```
pcf000, line 635: Typo in original?

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
```diff
@@ -365,14 +384,264 @@ static FlatSymbolRefAttr makeFp8TruncFunction(Location loc, FloatType outType,
...
     return failure();
   if (resType.isFloat8E5M2FNUZ() && !f8E5M2FNUZFunc)
     return failure();
+  if (resType.isFloat8E4M3FN() && !f8E4M3FNFunc)
```
pcf000, line 639: From here on down in this file just makes sure we address all four types instead of just the original two.

mlir/test/Conversion/EmulateFp8ExtTrunc/Integration/roundtrip.mlir
```diff
@@ -11,6 +11,18 @@
 // RUN: mlir-cpu-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_c_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_float16_utils%shlibext --entry-point-result=void | \
 // RUN: FileCheck %s --check-prefixes=CHECK,F8E5M2FNUZ
 
+// RUN: sed -e 's/##TYPE##/f8E4M3FN/g' %s | \
```
pcf000: Same test, new types.

mlir/test/Conversion/EmulateFp8ExtTrunc/Integration/roundtrip.mlir
```diff
@@ -36,6 +48,10 @@ func.func @printFail(%wanted : i8, %got : i8, %intermediate : f32) {
   call @printI64(%gotExt) : (i64) -> ()
   call @printComma() : () -> ()
   call @printF32(%intermediate) : (f32) -> ()
+  %interbits = arith.bitcast %intermediate : f32 to i32
```
pcf000, line 51: Print the intermediate float as bits -- NaNs were printing as "nan" and I needed to see exactly what was there.  Might as well keep the addition.

mlir/test/Conversion/EmulateFp8ExtTrunc/emulate-ocp-fp8-ext-trunc.mlir
```diff
@@ -0,0 +1,100 @@
+// RUN: rocmlir-opt -pass-pipeline='builtin.module(gpu.module(emulate-fp8-ext-trunc),emulate-fp8-ext-trunc)' -split-input-file %s | FileCheck %s
```
pcf000, line 1: Same intent as emulate-fp8-ext-trunc.mlir, but enough changes that I didn't want to merge with it.

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
```diff
@@ -31,6 +31,9 @@
 #include "llvm/ADT/TypeSwitch.h"
 #include "llvm/Support/raw_ostream.h"
 
+#include <bit>
+#include <cstdio>
```
krzysz00: You _definitely_ shouldn't need `cstdio`

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
```diff
@@ -119,7 +127,18 @@ static Value getFloatValueTableFor(Type elementType, Operation *op,
...
-    tableElems.push_back(entry.convertToFloat());
+    float x = entry.convertToFloat();
+    uint32_t u = llvm::bit_cast<uint32_t>(x);
+    // Ugly hack:  APFloat will turn signaling NaN into quiet Nan, so to get
+    // correct reversible expansion we force this special case.
+    if (type.isFloat8E5M2()) {
+      if (i == 0x7d)
+        u = 0x7fa00000;
+      if (i == 0xfd)
+        u = 0xffa00000;
```
krzysz00: Do we know that these are supposed to be qnans?

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
krzysz00: I claim that these two functions are separate enough that they might want to be separate functions?

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
krzysz00: Yep, that's the one

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
krzysz00: Yeah, good catch

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
```diff
@@ -460,13 +735,19 @@ void EmulateFp8ExtTruncPass::runOnOperation() {
...
+  FlatSymbolRefAttr f8E4M3FNTruncFunc = nullptr;
+  FlatSymbolRefAttr f8E5M2TruncFunc = nullptr;
+  SmallVector<Location> f8E4M3FNUZLocs, f8E5M2FNUZLocs, f8E4M3FNLocs, f8E5M2Locs;
   op->walk([&](TruncFOp op) {
     Type outElemType = getElementTypeOrSelf(op.getOut().getType());
     if (outElemType.isFloat8E4M3FNUZ())
       f8E4M3FNUZLocs.push_back(op->getLoc());
     else if (outElemType.isFloat8E5M2FNUZ())
       f8E5M2FNUZLocs.push_back(op->getLoc());
+    if (outElemType.isFloat8E4M3FN())
```
krzysz00: Nit: `else if` here as well

krzysz00 (commented):
I'm trusting the mechanical conversion and exhaustive testing - I assume the weirdness with NaN is part of the spec? If so, it's weird it wasn't in the MIGraphX code. Do they have a bug open for that change?

krzysz00:
Also, you can leave off stochiastic rounding support entirely - because calling that requires an extra argument (the random input), it can't come from an `arith.truncf`

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
krzysz00: ... Apparently they're not, never mind

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
pcf000: I went with the idea that the mantissa bits in 8-bit should become the mantissa bits in 32-bit, and for NaN the exponent is all-ones in both.  In OCP e5m2, the NaNs are 0x7d, 0x7e, and 0x7f, which would then become 0x7fa00000, 0x7fc00000, and 0x7fe0000.  APFloat was forcing 0x7fa00000 to 0x7fe00000, when meant that 0x7d, extended and then truncated, came back as 0x7f which failed the roundtrip.mlir test.

krzysz00: I'd go double-check with hardware / anyone else's details on what's supposed to happen here

pcf000:
Jerry pointed out some test areas I might have missed.  Added code to recognise either pair of F8 types just about everywhere, and made tests run through both pairs as well.  A couple of tests required adding a new --use-ocp-fp8-types option to rocmlir-gen, to generate the desired pair.

krzysz00:
For the OCP types, we might want to key the meaning of "fp8" and "bf8" off of `arch` with a `--force-float-type=[arch|nanoo|ocp]`

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/IR/AMDGPU.td
```diff
@@ -177,11 +177,11 @@ def AMDGPU_RawBufferLoadOp :
 def AMDGPU_RawBufferStoreOp :
     AMDGPU_Op<"raw_buffer_store", [AllElementTypesMatch<["value", "memref"]>,
       AttrSizedOperandSegments]>,
-    Arguments<(ins AnyTypeOf<[BF16, F16, F32, I32, I8, F8E5M2FNUZ, F8E4M3FNUZ,
+    Arguments<(ins AnyTypeOf<[BF16, F16, F32, I32, I8, F8E5M2FNUZ, F8E4M3FNUZ, F8E5M2, F8E4M3FN,
```
krzysz00: I went and removed the type restriction down in my int4 PR but that doesn't block this

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -524,40 +524,40 @@ static std::optional<StringRef> mfmaOpToIntrinsic(MFMAOp mfma,
       return ROCDL::mfma_f64_4x4x4f64::getOperationName();
   }
 
-  if (sourceElem.isFloat8E5M2FNUZ() && destElem.isF32() &&
-      chipset.minorVersion >= 0x40) {
+  if ((sourceElem.isFloat8E5M2FNUZ() || sourceElem.isFloat8E5M2()) &&
```
krzysz00: No

F8E45M2FNUZ is allowed on `minorVersion >= 0x40 && minorVersion < 0x50` and E5M2 and E4M3FN are on `minorVersion >= 0x50`

external/llvm-project/mlir/lib/Conversion/ArithToAMDGPU/ArithToAMDGPU.cpp
```diff
@@ -204,7 +205,8 @@ LogicalResult TruncFToFloat8RewritePattern::match(arith::TruncFOp op) const {
   if (inType && inType.getWidth() <= 8 && saturateFP8)
     // Conversion between 8-bit floats is not supported with truncation enabled.
     return failure();
-  return success(outType.isFloat8E5M2FNUZ() || outType.isFloat8E4M3FNUZ());
+  return success(outType.isFloat8E5M2FNUZ() || outType.isFloat8E4M3FNUZ() ||
```
krzysz00: I'd add chipset version guards here too

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
```diff
@@ -194,8 +194,8 @@ void rock::buildBackendPipeline(OpPassManager &pm,
   auto &gpuPm = pm.nest<gpu::GPUModuleOp>();
   gpuPm.addPass(amdgpu::createAmdgpuEmulateAtomicsPass({options.chip}));
   arith::ArithEmulateUnsupportedFloatsOptions floatEmuOpts;
-  SmallVector<std::string, 3> unsupportedFloats = {"bf16", "f8E4M3FNUZ",
-                                                   "f8E5M2FNUZ"};
+  SmallVector<std::string, 3> unsupportedFloats = {
```
krzysz00: Nit, let's make that `3` a `5`

mlir/test/Conversion/EmulateFp8ExtTrunc/Integration/roundtrip.mlir
```diff
@@ -53,7 +69,20 @@ func.func @testAllExtTrunc() {
     %o = arith.bitcast %oF8 : ##TYPE## to i8
     %mismatch = arith.cmpi ne, %i, %o : i8
     scf.if %mismatch {
-      func.call @printFail(%i, %o, %iFloat) : (i8, i8, f32) -> ()
+      // 7d, 7e, and 7f are all NaN.  They become 7f800001 as f32
```
krzysz00: That isn't correct for the old types - their NaN is `0x80`

mlir/test/Conversion/EmulateFp8ExtTrunc/Integration/roundtrip.mlir
```diff
@@ -67,32 +96,45 @@ func.func @main() {
...
+  // F8E5M2:     3, 57344
+  // F8E4M3FN:   3, nan
   call @testTruncExt(%id3, %in3) : (i64, f32) -> ()
 
   %id4 = arith.constant 4 : i64
   %in4 = arith.constant -57344.0 : f32
   // F8E5M2FNUZ: 4, -57344
   // F8E4M3FNUZ: 4, -240
+  // F8E5M2:     4, -57344
+  // F8E4M3FN:   4, -nan
```
krzysz00, line 134: Shouldn't it be `-inf`?

krzysz00 (commented):
Substantial notes on the upstream stuff, nitpicks otherwise

mlir/test/Conversion/EmulateFp8ExtTrunc/Integration/roundtrip.mlir
pcf000: OCP E4M3 doesn't have inf, and uses nan instead.

krzysz00: ... Ah, misread the test

external/llvm-project/mlir/lib/Conversion/ArithToAMDGPU/ArithToAMDGPU.cpp
pcf000: That turned out to be a bit more elaborate, since that pass didn't already keep chipset around, but the changes were mostly easily copied from the other pass.

mlir/test/Conversion/EmulateFp8ExtTrunc/Integration/roundtrip.mlir
pcf000: It only opens a window for a missed failure when input is 7d or 7e and output is 7f.  But I put a conditional in anyway.

external/llvm-project/mlir/lib/Conversion/ArithToAMDGPU/ArithToAMDGPU.cpp
```diff
@@ -212,7 +217,12 @@ LogicalResult TruncFToFloat8RewritePattern::match(arith::TruncFOp op) const {
   if (inType && inType.getWidth() <= 8 && saturateFP8)
     // Conversion between 8-bit floats is not supported with truncation enabled.
     return failure();
-  return success(outType.isFloat8E5M2FNUZ() || outType.isFloat8E4M3FNUZ());
+
+  return success(
+      (((outType.isFloat8E5M2FNUZ() || outType.isFloat8E4M3FNUZ()) &&
+        chipset.minorVersion >= 0x40 && chipset.minorVersion < 0x50) ||
```
krzysz00: Note: add an `isGfx940()` to `Chipset` that does `majorVersion == 9 && minorVersion >= 0x40 && majorVersion < 0x50`

Then perhaps add a `hasOcpFp8()` that does `(majorVersion == 9 && minorVersion >= 0x50) || majorVersion >= 12`

mlir/test/rocmlir-driver/fp8_ops.mlir
```diff
@@ -1,6 +1,7 @@
 // RUN: rocmlir-gen --arch gfx940 --operation gemm -t fp8 -p | rocmlir-driver --kernel-pipeline=gpu,rocdl | FileCheck %s --check-prefix=MFMA
 // RUN: rocmlir-gen --arch gfx940 --operation gemm -mfma=off -t fp8 -p | rocmlir-driver --kernel-pipeline=gpu,rocdl | FileCheck %s --check-prefix=MFMA_OFF
 // RUN: rocmlir-gen --arch gfx1100 --operation gemm -t fp8 -p | rocmlir-driver --kernel-pipeline=gpu,rocdl | FileCheck %s --check-prefix=GFX11
+// RUN: rocmlir-gen --arch gfx1100 --operation gemm -t fp8 -p | rocmlir-driver --kernel-pipeline=gpu,rocdl | FileCheck %s --check-prefix=GFX11_OCP
```
krzysz00: Something's wrong with your tests, because this is the exact same line as the `GFX11` prefix but yet seems to expect different results

krzysz00 (changes requested):
1. I'm seeing a bug in the FP8 emulation tests - not sure how you got a CI pass
2. The version checks aren't right - they cut off the gfx12s and will cause problems with, say, gfx1150

mlir/test/rocmlir-driver/fp8_ops.mlir
pcf000: It should obviously be forcing the types on one of them, and I bet it's making the old ones in both cases, for which the new types will match because they're substrings.  Weakness in the tests.  Anyway, two things to fix.

krzysz00 (approved):
Approved - though if you could `git rebase -i` and reorganize your commits to make upstreaming easy that'd be good

PR #1567, opened 2024-07-11
CHange definition of dequantizelinear to match MIGraphX, ONNX

krzysz00:
While debugging accuracy issues related to an MIGraphX change that caused the emission of ops like

    migraphx.dequantizelinear %x, %bias, %scale : <...xi8>, <...xi8>, <...xf32>

we discovered that our understanding of MIGraphX's semantics for dequnatizelinear were incorrect or out of date. The intended semantics, as seen in MIGraphX's quantiziation simplifier and, more importantly, in the ONNX reference at
https://github.com/onnx/onnx/blob/c7717bb39c684a6d86c82a5bb1d4c0e5d90353fe/onnx/reference/ops/op_dequantize_linear.py#L42 are that the input and bias have to be cast to the scale/output (they're required to match) element type before computation.

We didn't do this, causing validitaion errors due to 8-bit integer over/unferflow

This commit fixes is the issue. It isn't known if it needs a backport.

### DIFF
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -834,8 +834,11 @@ struct SoftmaxConverter final
 };
 } // namespace
 
-// MIGraphX pseudo code:
-// output[i] = static_cast<T>(input[i] - zero_pts[i]) * scales[i];
+// MIGraphX implements:
+// Let T = scale element type
+// output[i] = (convert<T>(input[i]) - convert<T>(zero_pts[i])) * scales[i];
+// For f32, this matches ONNX reference, dequantizing to f16, if it's ever done
+// will be less precise than the reference but that's probably fine.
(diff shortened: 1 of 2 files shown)


manupak:
Can we ask them to canonicalize the definition of dequantizelinear? 
(Then we dont have to maintain/deal with this and they can change)

@umangyadav @CharlieL7 do you think its possible to provide us with pointwise ops instead of dequantizelinear which has somewhat grey definition as we are seeing in this PR ?

manupak (approved):
LGTM. also thanks for cleaning up type creations.

CharlieL7:
> Can we ask them to canonicalize the definition of dequantizelinear? (Then we dont have to maintain/deal with this and they can change)
> 
> @umangyadav @CharlieL7 do you think its possible to provide us with pointwise ops instead of dequantizelinear which has somewhat grey definition as we are seeing in this PR ?

Yes, that is what I was suggesting for a later release. Directly pipe the pointwise ops to MLIR instead of the dequantizelinear. We have a PR that allows input fusions to MLIR https://github.com/ROCm/AMDMIGraphX/pull/3010.

PR #1568, opened 2024-07-12
Bumping Navi3 CI execution to use 4 threads

jerryyin:
In retrospect, bumping the Navi3 CI to use 1 thread is making the test running excruciating long. I'm attempt to bump the thread count back up again. I'm hoping this can move cost of time in CI back to reasonable range.


manupak:
Are we trading off timeouts for more hangs ?
I remember you mention that there is some hw feature we need to disable for hangs. Is that somewhat related ?

krzysz00 (approved):

jerryyin:
@manupak In my previous testing with Navi3 on two of the CI systems, they are fine with up to 4 threads. Although hardware folks casually mentioned about Navi3's in-ability to co-execute, looking at the state of CI spending couple of hours on just PR make me worried.

> there is some hw feature we need to disable for hangs.

This turns out to be not relevant. On my follow-up call, disabling CWSR is not a valid configuration for ROCm. Looks like we can't go down that path.

PR #1573, opened 2024-07-17
quickTunerGen

ethansaurusrex:
This is the 2nd part of the divided `quickTuner.py` script. This script lets the user pass in the large normalized dataframe from `quickTunerPreproc.py` and then run selected quick tuning perf config methods, printing or saving the data to `.qt` files.


mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -338,7 +338,7 @@ PopulateParamsAccel::paramsProbablyValid(OpBuilder &b,
 LogicalResult
 PopulateParamsAccel::couldBePerformant(const PopulateParamsInfo &info,
                                        const InitParamsAccel &params) {
-  return specificCouldBePerformant(params, info.gemmAType, info.gemmBType);
```
ethansaurusrex, line 341: This will be fixed in another PR, should not have been added

PR #1575, opened 2024-07-18
quickTunerPreproc.py, preprocessor script for quick tuner scripts

ethansaurusrex:
This script accepts the *.debug files that are generated from tuningRunner.py as input and concatenates them into a single tsv file for use in the next step of the quick tuning pipeline: generation.

### DIFF
--- a/mlir/utils/performance/quickTunerPreproc.py
+++ b/mlir/utils/performance/quickTunerPreproc.py
@@ -45,7 +45,7 @@ import pandas as pd
 import glob
 from sklearn.preprocessing import MinMaxScaler
 
-class qtPreprocessor(object):
+class qtPreprocessor():
     """
     class to process *.debug files into a single script
     """
@@ -97,29 +97,28 @@ class qtPreprocessor(object):
         print(os.path.join(input_dir, f"*.{file_ext}"))
 
         dfs = []
-        ct = 0
         for file in tsv_files:
-            df = pd.read_csv(file, sep='\t')
+            df = pd.read_csv(file, sep='\t', index_col=None)
+            df = df[df.columns[1:]]
             if normalize:
                 scaler = MinMaxScaler()
                 df['NormalizedTFlops'] = scaler.fit_transform(df[['TFlops']])
             dfs.append(df)
-            ct += 1
         if not dfs:
             return None
         new_df = pd.concat(dfs, ignore_index=True)
 
         if output_name:
-            new_df.to_csv(output_name, sep='\t')
+            new_df.to_csv(output_name, sep='\t', index=False)
             if debug:
                 print(f"Saved to {output_name}")
                               
         if debug:            
             # here output some stats about files
             if op == 'gemm':
-                qtPreprocessor.__get_stats_gemm(new_df, ct)
+                qtPreprocessor.__get_stats_gemm(new_df, len(tsv_files))
             elif op == 'conv':
-                qtPreprocessor.__get_stats_conv(new_df, ct)
+                qtPreprocessor.__get_stats_conv(new_df, len(tsv_files))
 
         return new_df
             
--- a/mlir/utils/performance/quickTunerPreproc.py
+++ b/mlir/utils/performance/quickTunerPreproc.py
@@ -103,7 +103,7 @@ class qtPreprocessor(object):
             if normalize:
                 scaler = MinMaxScaler()
                 df['TFlops'] = scaler.fit_transform(df[['TFlops']])
-                df = df.rename(columns={'TFlops':'NormalizedTFLops'})
+                df = df.rename(columns={'TFlops':'NormalizedTFlops'})
             dfs.append(df)
             ct += 1
         if not dfs:
--- a/mlir/utils/performance/quickTunerPreproc.py
+++ b/mlir/utils/performance/quickTunerPreproc.py
@@ -76,7 +76,7 @@ class qtPreprocessor():
         print(f"Number of unique Gemms: {num_gemms}")
         for _,row in unique_gemms.iterrows():
             tup = tuple(row)
-            print(f"{tup[0]},{tup[1]},{tup[2]},{tup[3]},{tup[4]},{tup[5]}")
+            print(','.join(map(str, tup)))
         
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


mlir/utils/performance/quickTunerPreproc.py
```diff
@@ -0,0 +1,169 @@
...
+
+class qtPreprocessor(object):
+    """
+    class to process *.debug files into a single script
+    """
+
+    def __init__(self, pargs):
+        self.input_dir = pargs.input_dir
+
+    @staticmethod
```
pcf000, line 56: Why staticmethod?  Not a criticism, I'm just less familiar with why that's done in python and I'm curious.

mlir/utils/performance/quickTunerPreproc.py
```diff
@@ -0,0 +1,169 @@
...
+        tsv_files = glob.glob(os.path.join(input_dir, f"*.{file_ext}"))
+        print(os.path.join(input_dir, f"*.{file_ext}"))
+
+        dfs = []
+        ct = 0
+        for file in tsv_files:
+            df = pd.read_csv(file, sep='\t')
+            if normalize:
+                scaler = MinMaxScaler()
+                df['TFlops'] = scaler.fit_transform(df[['TFlops']])
```
pcf000: Why one set of brackets on the left but two on the right??

mlir/utils/performance/quickTunerPreproc.py
ethansaurusrex: This was intended to be imported and called within the quickTunerGen.py set of classes but I could not rationalize each individual class having their own instance of this preprocessor class hanging out inside:
So, it was either have a standalone function or a class with a static method that would allow me to do something like:
```
qtPreprocessor.process( /* args */ )
```
From my experience with using the `@staticmethod` decorator, it has always been what to use when you want to package functions together that have similar functionality and can share data but do not share state.

mlir/utils/performance/quickTunerPreproc.py
ethansaurusrex: I think that was a mess up on my end. Fortunately (or unfortunately) the syntax is forgiving. 

Also, this will actually be changed to add a `NormalizedTFlops` column instead of overwriting the current one.
```
if normalize:
    scaler = MinMaxScaler()
    df['NormalizedTFlops'] = scaler.fit_transform(df[['TFlops']])
```

mlir/utils/performance/quickTunerPreproc.py
```diff
@@ -0,0 +1,169 @@
...
+
+    cmd = env_str + f" python3 ./bin/tuningRunner.py -q {special_args} \
+                     --config='{config_string}' --mlir-build-dir `pwd` \
+                     --output=- --tflops \
+                     --rocmlir_gen_flags='--device={self.gpu_id}' 2>/dev/null"
+
+
+to:
+    
+    import uuid
```
krzysz00, line 25: Could this instead be a patch to `rocmlir_worker`?

mlir/utils/performance/quickTunerPreproc.py
```diff
@@ -0,0 +1,169 @@
...
+"""
+
+import os
+import sys
+import argparse
+import pandas as pd
+import glob
+from sklearn.preprocessing import MinMaxScaler
+
+class qtPreprocessor(object):
```
krzysz00: Nit class case, and I'm pretty sure Python 3 doesn't require the `(object)` anymore

mlir/utils/performance/quickTunerPreproc.py
```diff
@@ -0,0 +1,169 @@
...
+
+        # num unique gemms in file:
+        cols = ['TransA', 'TransB', 'G', 'M', 'N','K']
+        unique_gemms = df[cols].drop_duplicates()
+        
+        num_gemms = len(unique_gemms)
+        print(f"Number of unique Gemms: {num_gemms}")
+        for _,row in unique_gemms.iterrows():
+            tup = tuple(row)
+            print(f"{tup[0]},{tup[1]},{tup[2]},{tup[3]},{tup[4]},{tup[5]}")
```
krzysz00: This is, I'd think, print(`,`.join(row)`)

mlir/utils/performance/quickTunerPreproc.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        staticmethod process() function that compiles output files into a single dataframe and saves to tsv file
+        """
+
+        tsv_files = glob.glob(os.path.join(input_dir, f"*.{file_ext}"))
+        print(os.path.join(input_dir, f"*.{file_ext}"))
+
+        dfs = []
+        ct = 0
+        for file in tsv_files:
```
krzysz00, line 118: You can `emumerate()` instead of maintaining a `ct`

mlir/utils/performance/quickTunerPreproc.py
krzysz00: ... This could also be `to_csv` on the dataframe with `skip_index=True` or some such thing

krzysz00 (commented):
Python nitpicking

mlir/utils/performance/quickTunerPreproc.py
pcf000: Not in its current form;  it cuts off the usual functionality of passing results to stdout and we don't have code to read results from the files and send it on.  I do have an open issue about collecting all the results into the database and this may become part of it.

mlir/utils/performance/quickTunerPreproc.py
ethansaurusrex: > This is, I'd think, print(`,`.join(row)`)

If the row contains non string entries, then it will complain and require conversion outside of the `.join()` function.

mlir/utils/performance/quickTunerPreproc.py
ethansaurusrex: I will replace this with a `len()` call since it is not needed within the loop only for generating stats.

mlir/utils/performance/quickTunerPreproc.py
krzysz00: `','.join(map(str, tup))`

mlir/utils/performance/quickTunerPreproc.py
```diff
@@ -0,0 +1,168 @@
...
+"""
+
+import os
+import sys
+import argparse
+import pandas as pd
+import glob
+from sklearn.preprocessing import MinMaxScaler
+
+class qtPreprocessor():
```
krzysz00, line 48: Standard complaining re style on class names

krzysz00 (approved):
Broadly approved

PR #1577, opened 2024-07-18
Mfma prune

ethansaurusrex:
Try pruning configs that take too many registers as likely non-performant. This adds a tentative check in the XDLOPs `specificCouldBePerformant()` function to see if the output size (`numOutputVectorElements`) is greater than the number of available registers.

### DIFF
--- a/mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
+++ b/mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
@@ -341,14 +341,6 @@ PopulateParamsAccel::paramsProbablyValid(OpBuilder &b,
 LogicalResult
 PopulateParamsAccel::couldBePerformant(OpBuilder &b, const PopulateParamsInfo &info,
                                        const InitParamsAccel &params) {
-  //int64_t mRepeats = ;
-  //int64_t nRepeats = ;e
-  // look ABBEgg
-  // OpBuilder b
-  // info.gemmAType, info.gemmBtype
-  // info.arch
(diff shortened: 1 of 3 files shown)


mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -693,12 +696,33 @@ PopulateParamsXDL::getTuningParameters(KernelType opType, Type dataTypeA,
...
+
+  if (!accelEmitterPtr)
+    return failure();
+
+  rock::accel::AccelEmitterParams accelParams = accelEmitterPtr->getParams();
+
+  int64_t numOutputVectorElements = accelParams.numOutputVectorElements();
+
+  // would be best to have register count be a part of arch, is not necessarily totalVGPRPerEu
+  if(numOutputVectorElements > 256) {
```
krzysz00, line 721: Let's go ahead and make that an arch field, then

krzysz00 (commented):
... I thought I'd sent the review in already

This seems fine - but do we have data that this is a good condition to impose? Are there cases where the number of outputs is > 256 but performance is good?

ethansaurusrex:
> ... I thought I'd sent the review in already
> 
> This seems fine - but do we have data that this is a good condition to impose? Are there cases where the number of outputs is > 256 but performance is good?

I am doing another tuning run this weekend and testing it further, atm cannot tell. I should have results by Monday or Tuesday for that question.

krzysz00:
Did we ever get the experiment run?

PR #1578, opened 2024-07-18
[DO NOT SQUASH] Subtree merge from llvm-project upstream, 2024-07-17

pcf000:
Latest subtree merge from amd-staging.  The four commits are the big squash, the merge commit, and follow-up changes divided into external and internal parts.  Only three file conflicts in the subtree pull.

Have passed check-mlir, check-rocmlir, and pull-request CI on lockhart, with ignore-external-linting ON.  Haven't yet tried the premerge checks (ie, clang-format), a full weekly, or other targets.

It went so smoothly (aside from troubles with AddMLIR.cmake) that I wonder if I've forgotten something, so feel free to point anything out.


external/llvm-project/mlir/include/mlir/Dialect/GPU/IR/CompilationAttrInterfaces.td
```diff
@@ -47,7 +47,8 @@ def GPUTargetAttrInterface : AttrInterface<"TargetAttrInterface"> {
         meant to be used for passing additional options that are not in the
         attribute.
       }], "::mlir::Attribute", "createObject",
-        (ins "const ::llvm::SmallVector<char, 0>&":$object,
+        (ins "::mlir::Operation*":$module,
```
pcf000: I think this is the only subtree-pull conflict where I needed to take our side instead of the incoming side.  All the uses of this function merged cleanly.

external/llvm-project/mlir/include/mlir/ExecutionEngine/RocmDeviceName.h
```diff
@@ -22,19 +22,19 @@ namespace mlir {
 // RocmDeviceName handles decomposition of gcnArchName
 class RocmDeviceName {
 public:
-  LogicalResult parse(StringRef devName);
+  LogicalResult parse(llvm::StringRef devName);
```
pcf000: StringRef and LogicalResult moved around and needed more namespace qualifiers.

external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt
```diff
@@ -384,6 +384,9 @@ if(LLVM_ENABLE_PIC)
       intrinsics_gen
     )
 
+    set_property(TARGET obj.MLIRRocmExecutionEngineUtils
```
pcf000: Hack.  The one file in this library, RocmSystemDetect.cpp, was not getting the -D__HIP_PLATFORM_AMD__=1 that it needed to compile.  Setting this property here is the smaller cleaner alternative to using the old AddMLIR.cmake instead of the upstream one.  I should probably pull it from hip::host instead of hard-coding.

It seems to indicate an issue with how properties are propagated with the recent cmake changes that prepend "obj." to target names, but I haven't tracked it further than that yet.

external/llvm-project/mlir/lib/ExecutionEngine/CpuSystemDetect.cpp
```diff
@@ -33,11 +33,7 @@ CpuSystemDetect::CpuSystemDetect() {
   // collect all CPU
   llvm::SmallString<32> targetTriple(llvm::sys::getProcessTriple());
   llvm::SmallString<8> cpu(llvm::sys::getHostCPUName());
-  llvm::StringMap<bool> features;
-  if (!llvm::sys::getHostCPUFeatures(features)) {
-    // System detect can't fail but we somehow missed the features;
-    features.clear();
-  }
+  llvm::StringMap<bool> features(llvm::sys::getHostCPUFeatures());
```
pcf000: Interface change:  getHostCPUFeatures() returns its StringMap instead of updating the reference parameter.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -731,48 +730,49 @@ GemmSize ConvBwdWeightOp::getGemmSize() {
 
 void ConvOp::getEffects(
     SmallVectorImpl<MemoryEffects::EffectInstance> &effects) {
-  effects.emplace_back(MemoryEffects::Read::get(), getOutput(),
```
pcf000: Interface change in MemoryEffects constructor.

mlir/test/rocmlir-driver/pipelines.mlir
```diff
@@ -9,7 +9,7 @@
 // COM: Do not put a leading space between the colon and the pass you're looking for
 // MIGRAPHX:Kernel pipeline:
 // MIGRAPHX-NEXT:builtin.module(func.func(migraphx-transform),
-// MIGRAPHX-NEXT:func.func(canonicalize{  max-iterations=10 max-num-rewrites=-1 region-simplify=true test-convergence=false top-down=true}),
+// MIGRAPHX-NEXT:func.func(canonicalize{  max-iterations=10 max-num-rewrites=-1 region-simplify=normal test-convergence=false top-down=true}),
```
pcf000: Attribute changed from true/false to disabled/normal/aggressive.

external/llvm-project/mlir/include/mlir/Dialect/GPU/IR/CompilationAttrInterfaces.td
krzysz00: Check if this hasn't been updated upstream recently - this looks like a @fabianmcg in-progress PR.

(Or, also, @fabianmcg , please advice if we should take upstream's side of this)

external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt
krzysz00: Could it be the case that you're missing a `target_link_libraries` here?

external/llvm-project/mlir/test/Target/LLVMIR/rocdl.mlir
```diff
@@ -27,7 +27,7 @@ llvm.func @rocdl_special_regs() -> i32 {
   // CHECK: call i64 @__ockl_get_num_groups(i32 2)
   %12 = rocdl.grid.dim.z : i64
 
-  // CHECK: call range(i32 0, 64) i32 @llvm.amdgcn.workitem.id.x()
+  // CHECK: call noundef range(i32 0, 64) i32 @llvm.amdgcn.workitem.id.x()
```
krzysz00: Ah - this is the result of an ugly hack on our side that can be lemoved (after double-checking that `IntrinsicsAMDGPU.td` in the LLVM copy has `noundef`)

mlir/lib/Conversion/RockToGPU/RockToGPU.cpp
```diff
@@ -201,13 +201,13 @@ void LowerRockOpsToGPUPass::runOnOperation() {
     if (auto attr = theFunc->getAttr("block_size")) {
       gpuFunc->setAttr("block_size", attr);
       blockSize = cast<IntegerAttr>(attr).getInt();
-      gpuFunc->setAttr(gpu::GPUFuncOp::getKnownBlockSizeAttrName(),
+      gpuFunc->setAttr("known_block_size",
```
krzysz00: This can be just `gpuFunc.setKnownBlockSize()`

krzysz00 (commented):
Minor notes on the reviewable bits

external/llvm-project/mlir/include/mlir/Dialect/GPU/IR/CompilationAttrInterfaces.td
fabianmcg: Take downstream, that's a patch that hasn't gotten into upstream yet.

external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt
pcf000: target_link_libraries is there, target_compile_definitions is there, but somewhere the property is not communicated to the right place.

external/llvm-project/mlir/test/Target/LLVMIR/rocdl.mlir
pcf000: Not yet, it seems.

external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt
krzysz00: I think the `target_link_libraries` is going to a different target. Pay attention to which library the system name stuff is part of.

external/llvm-project/mlir/test/Target/LLVMIR/rocdl.mlir
krzysz00: That is, take a look at the file, it'll define the intrinsics to have `noundef` as one of their inherent properties (so it wouldn't show up in the call like this)

external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt
pcf000: I think you're right.  It seems I need target_link_libraries for both MLIRRocmExecutionEngineUtils and obj.MLIRRocmExecutionEngineUtils -- one to get the -D__HIP_PLATFORM_AMD__ for compilation and one to get HIP stuff for linking.  With those, I don't need this hack.  I'll see how small a change I can get away with and push it.

external/llvm-project/mlir/test/Target/LLVMIR/rocdl.mlir
pcf000: Many intrinsics in that file do have NoUndef<RetIndex>, but the only "workitem" definition I see is this:

  `defm int_amdgcn_workitem_id : AMDGPUReadPreloadRegisterIntrinsic_xyz;`

which has no NoUndef.  Since it's defm instead of def, I don't know if there's even a place to hang it.

krzysz00: Check if we have
```
class AMDGPUReadPreloadRegisterIntrinsic
  : DefaultAttrsIntrinsic<[llvm_i32_ty], [], [NoUndef<RetIndex>, IntrNoMem, IntrSpeculatable]>;
```
above that and the comment block. Note the `NoUndef<RetIndex>`.

If that's present, we can remove our `noundef` hack

pcf000: In the file is

```
   class AMDGPUReadPreloadRegisterIntrinsic
     : DefaultAttrsIntrinsic<[llvm_i32_ty], [], [NoUndef<RetIndex>, IntrNoMem, IntrSpeculatable]>;
```

which suggests we can remove it, but on the other hand we're producing that "noundef" in the asm -- oh, is that the hack, to force it in the printer?  Either way, where is the hack that I could remove?

krzysz00: Yeah. Go look in the translations to LLVM IR on `ROCDLOps.td` and the surrounding C++ source code - it'll be a patch we're carrying

krzysz00: ... Yeah `external/llvm-project/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td` line 111 is where I put the hack in

krzysz00:
Nits on reviewing the patch diff:
- In `external/llvm-project/mlir/include/mlir/Conversion/Passes.td ` at `def ConvertNVVMToLLVMPass ` : spurious change to `description` field
- At `external/llvm-project/mlir/include/mlir/Dialect/Func/IR/FuncOps.td`, spurious include of `AttrTypeBase.td`
- In `AMDGPUToROCDL.cpp` , in the WMMA lowering, we have an `if` where upstream has an `else if` - might be worth cleaning up for consistency / I'm pretty sure that would've been a code review change
- Grab upstream's `llvm/lib/Target/AMDGPU/AMDGPUPromoteAlloca.cpp`  and `llvm/lib/Target/AMDGPU/SIISelLowering.cpp` and `llvm/lib/Target/ARM/ARMISelLowering.cpp` - those are changes to a patch we cherry-picked out of review

Other notes, not for this PR but for we should get to these:
- ... someone (@giuseros ?) should upstream the packed RTZ thing
- What's the status of the DPP op upstreaming?
- Someone needs to get `unsafe_fp_atomics` up
- @sjw36 Why do we still have our changes to CallOpInterface? Do we still need them?
- Note to make a ticket: we might not need those `truncf` rewrites because we should be able to use LLVM's `bfloat` these days
- ... Can we ditch the Tosa partitioner? I don't think that whole system ever went anywhere

pcf000:
> In external/llvm-project/mlir/include/mlir/Conversion/Passes.td at def ConvertNVVMToLLVMPass : spurious change to description field

Giuseppe, last April, in a packed-fp16 patch.  Don't know why he changed it.

> At external/llvm-project/mlir/include/mlir/Dialect/Func/IR/FuncOps.td, spurious include of AttrTypeBase.td

Simon, in 2022 when he added getReadAccessAttrName() and getWriteAccessAttrName().

> In AMDGPUToROCDL.cpp , in the WMMA lowering, we have an if where upstream has an else if - might be worth cleaning up for consistency / I'm pretty sure that would've been a code review change

I read that the other way around, that we have "else if" and upstream has "if" -- but style guidelines would say to not have "else" anyway, because all the clauses are returns.  Looks like it was a clang-tidy fix from Mehdi Amini in October 2023;  why didn't clang-tidy complain about the others?

krzysz00:
Re the AttrTypeBase include, if it's not doing anything, we can get rid of it

Re Giuseppe's patch ... guess we can keep it for now

Re WMMA ... good point, I misread the diff, let's take upstream's version

krzysz00 (approved):
... apparently I didn't send in the approval for some reason

PR #1579, opened 2024-07-18
Fix reduction fusion with trailing views and/or memcpy

manupak:
Currently, reduction fusion assumed there would be same ranked output buffer
as the gemm it get fused with.
Additionally, it did not account for a trailing memcpy to be present
(This feature was done in a time where memcpy s were removed early on... *I think*)

This commit fixes to : 
1) Handle memcpy leading to output buffer of reduce_op
2) If there were trailing views following reductions, they'll be inverted and stacked onto the output buffer.

closes: https://github.com/ROCm/rocMLIR-internal/issues/1541

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
+++ b/mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
@@ -252,6 +252,8 @@ FailureOr<ArrayAttr> removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
 FailureOr<ArrayAttr>
 removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
                 const SetVector<StringRef> &removeDimNamesSet);
+
+ArrayAttr getArrayAttr(OpBuilder &b, ArrayRef<TransformMapAttr> transforms);
 } // end namespace rock
 } // end namespace mlir
 #endif
--- a/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
@@ -936,17 +936,17 @@ LAGenericRewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
 }
 
 template <typename WriteOp>
-static LogicalResult rewriteMemCpy(memref::CopyOp copy,LinalgAlignRewriter &b, function_ref<void(WriteOp , Value)> destUpdateFunc) {
+static LogicalResult
+rewriteMemCpy(memref::CopyOp copy, LinalgAlignRewriter &b,
+              function_ref<void(WriteOp, Value)> destUpdateFunc) {
   auto src = copy.getSource();
   auto target = copy.getTarget();
   Location loc = copy.getLoc();
 
   // See if the copy source is transformed.
   // If so we need to reverse them on the destination.
-  auto [rawSrc, transformsOnCpySrc, isBig] =
-  untransform(b, src);
+  auto [rawSrc, transformsOnCpySrc, isBig] = untransform(b, src);
   std::ignore = isBig;
-
 
   Operation *gemmStoreOp = nullptr;
   SmallVector<TransformMapAttr> views;
@@ -967,16 +967,14 @@ static LogicalResult rewriteMemCpy(memref::CopyOp copy,LinalgAlignRewriter &b, f
         return checkResult;
       }
       if (!isUniqueReader) {
-        gemmStoreOp =
-            static_cast<WriteOp>(b.clone(*twop.getOperation()));
+        gemmStoreOp = static_cast<WriteOp>(b.clone(*twop.getOperation()));
       } else {
         gemmStoreOp = twop;
       }
     }
   }
   if (gemmStoreOp) {
-    if (WriteOp writeOp =
-            dyn_cast<WriteOp>(gemmStoreOp)) {
+    if (WriteOp writeOp = dyn_cast<WriteOp>(gemmStoreOp)) {
       b.moveAfterIfNeeded(writeOp, copy);
 
       // 1. replace memref.copy with rock.threadwise_write_all
@@ -996,19 +994,22 @@ static LogicalResult rewriteMemCpy(memref::CopyOp copy,LinalgAlignRewriter &b, f
 LogicalResult
 MemcpyRewritePattern::matchAndRewrite(memref::CopyOp copy,
                                       LinalgAlignRewriter &b) const {
-  LogicalResult updateWriteAllOp = rewriteMemCpy<ThreadwiseWriteAllOp>(copy, b, [](ThreadwiseWriteAllOp writeOp, Value newDest){
-    writeOp.getDestMutable().assign(newDest);
-  }); 
-  if(succeeded(updateWriteAllOp)){
+  LogicalResult updateWriteAllOp = rewriteMemCpy<ThreadwiseWriteAllOp>(
+      copy, b, [](ThreadwiseWriteAllOp writeOp, Value newDest) {
+        writeOp.getDestMutable().assign(newDest);
+      });
+  if (succeeded(updateWriteAllOp)) {
     return success();
   }
-  LogicalResult updateReduceOp = rewriteMemCpy<ReduceOp>(copy, b, [](ReduceOp writeOp, Value newDest){
-    writeOp.getOutMutable().assign(newDest);
-  }); 
-  if(succeeded(updateReduceOp)){
+  LogicalResult updateReduceOp =
+      rewriteMemCpy<ReduceOp>(copy, b, [](ReduceOp writeOp, Value newDest) {
+        writeOp.getOutMutable().assign(newDest);
+      });
+  if (succeeded(updateReduceOp)) {
     return success();
   }
-  LLVM_DEBUG(llvm::dbgs() << "memCpy did not trace back to a WriteAll or Reduce op");                                
+  LLVM_DEBUG(
+      llvm::dbgs() << "memCpy did not trace back to a WriteAll or Reduce op");
   return failure();
 }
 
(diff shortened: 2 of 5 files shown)


mlir/include/mlir/Dialect/Rock/utility/transformMapUtils.h
```diff
@@ -252,6 +252,8 @@ FailureOr<ArrayAttr> removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
 FailureOr<ArrayAttr>
 removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
                 const SetVector<StringRef> &removeDimNamesSet);
+
+ArrayAttr getArrayAttr(OpBuilder &b, ArrayRef<TransformMapAttr> transforms);
```
krzysz00, line 256: Let's name this something less generic? Maybe `transformsToArrayAttr`?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -935,53 +935,81 @@ LAGenericRewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
...
+rewriteMemCpy(memref::CopyOp copy, LinalgAlignRewriter &b,
+              function_ref<void(WriteOp, Value)> destUpdateFunc) {
   auto src = copy.getSource();
   auto target = copy.getTarget();
   Location loc = copy.getLoc();
 
+  // See if the copy source is transformed.
+  // If so we need to reverse them on the destination.
+  auto [rawSrc, transformsOnCpySrc, isBig] = untransform(b, src);
+  std::ignore = isBig;
```
krzysz00, line 949: You can put the `std::ignore` in the `[]`s above?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -935,53 +935,81 @@ LAGenericRewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
...
+
   Operation *gemmStoreOp = nullptr;
   SmallVector<TransformMapAttr> views;
-  if (auto allocOp = src.getDefiningOp<memref::AllocOp>()) {
-    if (auto twop = dyn_cast_if_present<ThreadwiseWriteAllOp>(
-            traceToWriter(src, views))) {
+  LLVM_DEBUG(llvm::dbgs() << "copy source = " << rawSrc << "\n");
+  if (auto allocOp = rawSrc.getDefiningOp<memref::AllocOp>()) {
```
krzysz00, line 954: "No transforms on readers" is an invariant of the output fusion chain.

If `rawSrc != source`, then you need to go up to `Regularize.cpp` and fix it

krzysz00 (changes requested):
The appropriate fix is to change the regularization pass to move the transforms from the source of the copy to the destination of the reduce

manupak:
> The appropriate fix is to change the regularization pass to move the transforms from the source of the copy to the destination of the reduce

That could be an optimization. to this date, I quite do not get why we need "regularization".
So I disagree. 

If this passes get this IR, it should be dealt with. 
But if you think you can optimize it better above in regularization, I dont see why we cant do both either.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: I m just not going to do that now.
So I dont accept this as a reason to block this PR merely because this make this pass more robust if the situation arises.

However, you are more than welcome to fail/handle in regularization.

manupak:
>  I quite do not get why we need "regularization".

i.e. why we had split that functionality of moving tr stacks of readers into dests.
That created an artificial coupling between these two passes due to partitioning of the machinery of "fusion".

I am convinced this is good enough for now and unblocks @umangyadav by fixing the bug.
If you have better reasons with your approach -- feel free to fix it up in "regularize".

krzysz00:
We need regularization because otherwise we can't analyze vectorization in gridwise-gemm-to-blockwise or in places like that

That is, if I need the ultimate, real vectorization (which, say, the LDS buffers thing might want) we need to be able to traverse `linalg.generic` chains

And to do that well, we'd need to take inverses of maps it a bunch of random spots in the code

And so on and so forth - it's easier to establish the "no transforms on readers" invariant

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
krzysz00: No, I'm blocking this PR because "the generic/copy/[thing that gets fused away] does not take a view of the intermediate `alloc`" is an invariant of our fusion infrastructure

manupak:
Readers are readers of the output buffer of the gemm, So that are trailing operations of the gemm.

> we can't analyze vectorization in gridwise-gemm-to-blockwise

Only thing we do in gridwise-gemm-to-blockwise is figuring out vectorization dim of loads.
view stacks on readers of the output buffer of the gemm is not relevant to that.

> That is, if I need the ultimate, real vectorization (which, say, the LDS buffers thing might want) we need to be able to traverse linalg.generic chains

The real vectoriztion happens when threadwise_write_all get lowered that happens after rock-linalg-align. So I dont understand the relevance of lifting this up here.

> And to do that well, we'd need to take inverses of maps it a bunch of random spots in the code

This is the bit I dont get; why we had to lift them up out of linalg-align into a new pass called "regularization".

> And so on and so forth - it's easier to establish the "no transforms on readers" invariant

Sure, but I dont see why that has to be before rock-linalg-align.

## Summary

I still think fixing this in this pass is the last line of defense -- while for some optimization reasons that I dont get -- it could also be done above in regularization.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: > "the generic/copy/[thing that gets fused away] does not take a view of the intermediate alloc"

This is an artificial restriction that would otherwise could be solved here.

krzysz00:
Being able to query the correct vectorization up in gridwise-gemm-to-blockwise or something up there like the LDS swizzles pass can be important. Currently, it's something we only need for input fusion, but there have been/might be times where we need a `getMaxVectorization()` on the output buffer pre-fusion.

And there were correctness issues that become a lot simpler with the regularizations in place - the main one is to do with extra arguments

Namely
```
f(c) = gemm(a, b)
y = g(c) + h(d)
z = p(y) * q(e)
return z
```
for view functions `f, g, h, p, q`

Regularization will rewrite this to
```
f(c) = gemm(a, b)
inv_p(y') = g(c) + h(d)
z = y' * q(e)
return z
```
and then
```
f(inv_g(c')) = gemm(a, b)
inv_p(y') = c' + h(d)
z = y' * q(e)
return z
```

If we don't do this, we have (or perhaps had) a problem where we can't map in register coordinates accurately.

That is, the registers of the GEMM output corresponds to a concrete set of memory locations in `c'`, the regularized buffer. Then, to get the corresponding values from `d`, all we need to do is apply `d` to the outputs of that set of transforms.

If we didn't do this, to get coordinates that are an acceptable input to `h` (which are in `y'-space pre-regularization), we'd need to take our `c`-space coordinates and get them into `y`-space. That is, we have a `(bid, tid, iter) => [point in C]` function, applied in one go, and need to construct `(bid, tid, iter) => [point in D]`, which is `h(inv_g(f(bid, tid, iter)))` .

These days, with `threadwise_write_all` being what it is, it might not suck too much to do that, but back when there was a `transforming_for` loop and all we had was `[point in C]` and no inverse utilities, there'd be a lot of surgery required.

So we went and decreed "the thing that'll get fused in doesn't transform the buffer it's reading from" in order to massively simplify AlignTiling.cpp , and that's still a useful invariant to maintain.

manupak:
Im not a fan of making unnecessary assumption on the state of the IR which is a source of problems and inflexibility.

Input fusion is the only thing I have not thought about w.r.t to rationale of the existence of the regularization pass.
That might actually need something different than "no transforms on readers" invariant and needs to be done before rock.gemm lowering. 

Let me highlight the requirements needed by the fusions : 
R1) [Output Fusion] We need identity views between gemm output buffer and "readers" who read it.
R2) [Input Fusion] We need identity views between gemm input buffer and the "writers" who write it.

If the above is wrong. Please correct me before reading anything following this point.

We created "regularization" for R1.
The trailing compute operations that might read the output buffer of the gemm via different view -- that violates the R1 requirement -- were changed such that "readers" have an identity while views transferred to the writing buffer.

Now for R2, I m not sure whether  "no transforms on readers" invariant is still applicable.
The preceding compute operations might write to the input buffer of the gemm via different view that might violate the R2 requirement. Therefore it needs to have a  "no transforms on writers" ? 
i.e. we need to get the IR into a form as follows:

```
threadwise_read_into [view_stack_1] %input1 into %buffer1
linalg.generic %buffer1 into %buffer2
rock.gemm %buffer3 =  %buffer2 * .... 
```

which then gets expanded into

```
threadwise_read_into [view_stack_1] %input1 into %buffer1
linalg.generic %buffer1 into %buffer2
threadwise_read_into [gemm_tile_view_stack] %buffer2 into %gemm_A_buffer
```

^ Here assumption is views encoded in affine maps is pushed out into inputs using `makeLinalgGenericWithIdentityAffMaps` API.

So for the input fusion to happen, we need to make the linalg.generic -- the "writer" into `%buffer2` have no transforms on the writer. So that the two `threadwise_read_into ` ops could be combined ? 

leading to  

```
threadwise_read_into [view_stack_1, gemm_tile_view_stack] %input1 into %buffer1
linalg.generic %buffer1 into %gemm_A_buffer
```

To me the transformation is anchor op centric and its not about "readers" centric to buffer allocation. 
We need a transformation that pushes out transform stacks outwards from the anchor op and place proper verifiers to make sure that is in that state.

To this end, I think we should add rock.fusable_gemm (and rock.fusable_attention) with a `OpTrait` that has verifiers which verifies identity views on :
1) writers to its input buffers
2) readers of its output buffers

manupak:
Moving the discussion : https://github.com/ROCm/rocMLIR/discussions/1594

PR #1580, opened 2024-07-19
swizzle the MFMA outputs via LDS

dhernandez0:
swizzle the MFMA outputs via LDS:
1. locate threadwise_write_all operation that writes to global memory
2. rewrite it with a LDS-based swizzle to improve the thread layout for global store vectorization
3. reuse LDS allocated for loading A and B for this

We do not run this pass if:
- the LDS space required > LDS allocated for A and B matrices
- vectorization of iter in the original map is good

### sdxl-nhwc-conv (https://github.com/ROCm/rocMLIR-internal/issues/1421)
|    | baseline TFlops | PR TFlops | speed up |
|----|-----------------|-----------|----------|
| 0  | 98.77           | 100.711   | 1.02     |
| 1  | 122.27          | 126.0887  | 1.03     |
| 2  | 114.93          | 118.9413  | 1.03     |
| 3  | 97.42           | 95.2313   | 0.98     |
| 4  | 117.50          | 119.8767  | 1.02     |
| 5  | 97.52           | 90.02081  | 0.92     |
| 6  | 115.35          | 117.1437  | 1.02     |
| 7  | 99.72           | 100.609   | 1.01     |
| 8  | 5.92            | 5.809598  | 0.98     |
| 9  | 116.90          | 119.5479  | 1.02     |
| 10 | 22.51           | 22.8614   | 1.02     |
| 11 | 106.58          | 106.4475  | 1.00     |
| 12 | 124.72          | 127.9261  | 1.03     |
| 13 | 98.27           | 105.4659  | 1.07     |
| 14 | 111.36          | 116.6155  | 1.05     |
| 15 | 95.71           | 98.36484  | 1.03     |
| 16 | 117.19          | 120.0904  | 1.02     |
| 17 | 126.01          | 131.9751  | 1.05     |

### sdxl-conv (utils/performance folder)
|   | baseline TFlops | PR TFlops | speed up |
| - | --------------- | --------- | -------- |
| 0 | 93.29           | 103.639   | 1.11     |
| 1 | 96.02           | 105.6677  | 1.10     |
| 2 | 103.29          | 114.5144  | 1.11     |
| 3 | 93.94           | 101.262   | 1.08     |
| 4 | 73.43           | 77.04453  | 1.05     |
| 5 | 83.57           | 83.15381  | 1.00     |
| 6 | 72.22           | 81.47444  | 1.13     |
| 7 | 99.55           | 100.166   | 1.01     |
| 8 | 97.51           | 109.3014  | 1.12     |

### attention (utils/performance folder)
|    | baseline TFlops | PR TFlops | speed up |
| -- | --------------- | --------- | -------- |
| 0  | 99.41           | 99.02092  | 1.00     |
| 1  | 101.26          | 103.0688  | 1.02     |
| 2  | 101.35          | 107.7713  | 1.06     |
| 3  | 102.55          | 108.7795  | 1.06     |
| 4  | 104.76          | 110.9894  | 1.06     |
| 5  | 15.49           | 15.48772  | 1.00     |
| 6  | 79.92           | 83.23795  | 1.04     |
| 7  | 75.67           | 75.21676  | 0.99     |
| 8  | 23.41           | 24.44899  | 1.04     |
| 9  | 91.43           | 93.25776  | 1.02     |
| 10 | 27.76           | 29.06839  | 1.05     |
| 11 | 67.68           | 69.59236  | 1.03     |
| 12 | 25.32           | 25.22538  | 1.00     |

### sdxl-gemm (utils/performance folder)
|    | baseline TFlops | PR TFlops | speed up |
| -- | --------------- | --------- | -------- |
| 0  | 99.83           | 100.2244  | 1.00     |
| 1  | 0.52            | 0.480751  | 0.93     |
| 2  | 0.86            | 0.88      | 1.02     |
| 3  | 0.16            | 0.159502  | 1.02     |
| 4  | 1.93            | 1.889785  | 0.98     |
| 5  | 94.30           | 96.86614  | 1.03     |
| 6  | 107.63          | 108.9562  | 1.01     |
| 7  | 122.68          | 123.8708  | 1.01     |
| 8  | 95.72           | 98.13533  | 1.03     |
| 9  | 57.02           | 57.77229  | 1.01     |
| 10 | 121.62          | 123.1618  | 1.01     |
| 11 | 91.18           | 87.78368  | 0.96     |
| 12 | 105.45          | 106.6885  | 1.01     |
| 13 | 114.47          | 116.3354  | 1.02     |
| 14 | 120.36          | 123.1254  | 1.02     |
| 15 | 94.17           | 93.00009  | 0.99     |
| 16 | 38.12           | 37.67769  | 0.99     |

https://github.com/ROCm/rocMLIR-internal/issues/1511


mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,563 @@
...
+using mlir::gpu::AddressSpace;
+
+namespace {
+struct RockGemmOutputSwizzlePass
+    : public rock::impl::RockGemmOutputSwizzlePassBase<
+          RockGemmOutputSwizzlePass> {
+  void runOnOperation() override;
+};
+
+bool hasPrivateMemoryAddressSpace(MemRefType type) {
```
krzysz00: Don't we have this function somewhere?

Also, LLVM style nit: utility functions generally go outside the `namespace{}` and are `static`

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,563 @@
...
+              rewriter.createOrFold<arith::ConstantIndexOp>(loc, offset);
+
+          auto newView = rewriter.create<memref::ViewOp>(
+              loc, view.getType(), ldsByteBuffer, byteOffset,
+              view.getDynamicSizes());
+          rewriter.replaceAllUsesWith(view->getResults(),
+                                      newView->getResults());
+          rewriter.eraseOp(view);
+          ++numViews;
+        } else {
```
krzysz00: This is just `cast<>`

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,563 @@
...
+        if (auto view = dyn_cast<memref::ViewOp>(owner)) {
+          Value byteOffset =
+              rewriter.createOrFold<arith::ConstantIndexOp>(loc, offset);
+
+          auto newView = rewriter.create<memref::ViewOp>(
+              loc, view.getType(), ldsByteBuffer, byteOffset,
+              view.getDynamicSizes());
+          rewriter.replaceAllUsesWith(view->getResults(),
+                                      newView->getResults());
+          rewriter.eraseOp(view);
```
krzysz00: `replaceAllUsesWith` + `eraseOp` is `replaceOp`. Heck, this sequence is `replaceOpWithNewOp<memref::viewOp>`

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2880,4 +2850,4 @@ void RockGridwiseGemmToBlockwisePass::runOnOperation() {
                                     std::move(patterns)))) {
     signalPassFailure();
   }
-}
+}
```
krzysz00: Nit: newline

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
```diff
@@ -745,3 +750,32 @@ mlir::rock::getReassociationForFlattening(ShapedType srcTp) {
...
+  auto memoryAddressSpaceAttr =
+      b.getAttr<gpu::AddressSpaceAttr>(memoryAddressSpace);
+
+  auto rawMemType =
+      MemRefType::get({bufferDim * getByteWidth(elementType)}, b.getI8Type(),
+                      AffineMap{}, memoryAddressSpaceAttr);
+  auto buffer = b.create<GpuAllocOp>(loc, rawMemType);
+
+  return viewBufferAs(b, buffer, elementType);
+}
```
krzysz00: Nit: newline

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,563 @@
...
+  } else if (writes.empty()) {
+    LLVM_DEBUG(llvm::dbgs() << "No ThreadwiseWriteAllOp writes to "
+                               "global memory, skipping pass\n");
+  } else {
+    // Rewrite
+    RewritePatternSet patterns(&getContext());
+    patterns.add<ThreadwiseWriteAllRewritePattern>(&getContext());
+
+    GreedyRewriteConfig config;
+    config.strictMode = GreedyRewriteStrictness::ExistingOps;
```
krzysz00: Huh, what's that for?

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,563 @@
...
+    for (Operation *op : writes) {
```
krzysz00: ... you don't need to do this

You can `applyPatternsAndFoldGreedily` or the like to the entire function - the usual style - or, if you have a list of operations, you can pass them all in to one call.

Doing it this way is expensive

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,563 @@
...
+
+    // Prepare some useful constants.
+    Value convertedC = op.getSource();
+    Value matC = op.getDest();
+    Type destType = op.getDest().getType().getElementType();
+
+    // Convert from reg -> memory transform to reg -> block
+    int64_t mPerBlock, nPerBlock;
+    ArrayAttr idToLDS;
+    FailureOr<std::tuple<int64_t, int64_t, ArrayAttr>> maybeTuple =
```
krzysz00: Better variable naming?

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,563 @@
...
+          ArrayRef<int64_t>{numElements}, ArrayRef<int64_t>{1});
+      rewriter.replaceOp(alloc, subView);
+    }
+
+    offset += numElements;
+    i++;
+  }
+}
+
+static FailureOr<std::tuple<int64_t, int64_t, ArrayAttr>>
```
krzysz00: std::optional is probably fine here

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,563 @@
...
+        getMaxVectorization(matC, dimensionM, /*inputDimLen=*/
+                            dataPerThread, matC.getDefiningOp());
+    int64_t mVectorLen = mVectorRes.max;
+    VectorizationResult nVectorRes =
+        getMaxVectorization(matC, dimensionN, /*inputDimLen=*/
+                            dataPerThread, matC.getDefiningOp());
+    int64_t nVectorLen = nVectorRes.max;
+
+    int64_t dim;
+    if (mVectorLen > nVectorLen) {
```
krzysz00: This could be `int64_t dim = (mVectorLen > nVectorLen ? dimensionM : dimensionN`

And it might make sense to let `dim` be `unsigned` but that's not important

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,563 @@
...
+               << " iter: " << iter << "\n");
+    float row;
+    if (dim == dimensionM) {
+      LLVM_DEBUG(llvm::dbgs() << "dim = M\n");
+      row = mPerBlock;
+    } else {
+      row = nPerBlock;
+      LLVM_DEBUG(llvm::dbgs() << "dim = N\n");
+    }
+    float rowsPerWave = (elementsWrittenPerThread * 64) / row;
```
krzysz00: Why's this a float? Also, why 64? (Did you want ceiling or floor division?)

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,563 @@
...
+    LLVM_DEBUG(llvm::dbgs()
+               << "row: " << row << " rowsPerWave: " << rowsPerWave << "\n");
+
+    Value finalC =
+        gpuAlloc(b, loc, dataPerThread, destType, AddressSpace::Private);
+
+    TopDownTMBuilder tidIterMerge(b, {"tid", "iter"},
+                                  {blockSize, dataPerThread});
+    SmallVector<StringRef, 5> throughDims{"tid"};
```
krzysz00: You shouldn't need this SmallVector

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,563 @@
...
+    toMatrixC.unmerge("gemmM", 1, {"m_block", "block_m"}, {mBlocks, mPerBlock});
+    toMatrixC.unmerge("gemmN", 2, {"n_block", "block_n"}, {nBlocks, nPerBlock});
+    TransformMapAttr toMatrixCAttr = toMatrixC.get();
+
+    SmallVector<Attribute> transformAttrsStore{
+        createMblockAttr, tidIterMergeMemAttr, tidIterFlattenMemAttr,
+        flattenToBlockCoordAttr, toMatrixCAttr};
+    ArrayAttr idToMatrixCMaps = b.getArrayAttr(transformAttrsStore);
+
+    b.create<ThreadwiseWriteAllOp>(loc, finalC, matC, idToMatrixCMaps,
```
krzysz00: Same nits about replaceOp

... Also, note to future me that getting rid of that `extraAttrs` thing might require some more work here

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
```diff
@@ -165,6 +165,8 @@ void rock::buildKernelPipeline(OpPassManager &pm,
       funcPm.addPass(createConvertLinalgToAffineLoopsPass());
       funcPm.addPass(rock::createRockVectorizeFusionsPass());
     }
+    funcPm.addPass(rock::createRockGemmOutputSwizzlePass());
```
krzysz00: Actually, could you put this before `alignTiling`? That way, all the reads for fusions will vectorize nicely.

mlir/include/mlir/Dialect/Rock/utility/loweringUtils.h
```diff
@@ -16,6 +16,7 @@
 #include "mlir/Dialect/Utils/ReshapeOpsUtils.h"
 #include "mlir/Support/LLVM.h"
 #include "llvm/ADT/SmallVector.h"
+#include "mlir/Dialect/GPU/IR/GPUDialect.h"
```
krzysz00: Might be better to forward-declare `gpu::AddressSpace` here

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,563 @@
...
+                      workgroupMemoryAddressSpace);
+
+  rewriter.setInsertionPoint(allocs[0]);
+  Location loc = allocs[0]->getLoc();
+  auto ldsByteBuffer = rewriter.create<GpuAllocOp>(loc, ldsMemRefBufferType);
+
+  // Rewrite all GpuAllocs as SubViewOp
+  int64_t offset = 0;
+  size_t i = 0;
+  for (GpuAllocOp alloc : allocs) {
```
krzysz00: No need for an `i`, `llvm::enumerate` exists - that is, `for (auto [i, alloc] : llvm::enumerate(allocs))`

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,563 @@
...
+    }
+
+    if (replaceWithSubView) {
+      auto subView = rewriter.create<memref::SubViewOp>(
+          loc, ldsByteBuffer, ArrayRef<int64_t>{offset},
+          ArrayRef<int64_t>{numElements}, ArrayRef<int64_t>{1});
+      rewriter.replaceOp(alloc, subView);
+    }
+
+    offset += numElements;
```
krzysz00: I think this should be `+= elementBytes`?

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
krzysz00: ... We have way too many versions of this function and should make it a utility, but that might not be this PR

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,563 @@
...
+  int64_t lastSize = 0;
+  func.walk([&](GpuAllocOp gpuAlloc) {
+    auto type = gpuAlloc.getOutput().getType();
+    auto memSpaceValue =
+        dyn_cast_or_null<gpu::AddressSpaceAttr>(type.getMemorySpace())
+            .getValue();
+    if (memSpaceValue == gpu::GPUDialect::getWorkgroupAddressSpace()) {
+      lastSize = type.getNumElements() * getByteWidth(type.getElementType());
+      totalSize += lastSize;
+      allocs.push_back(gpuAlloc);
```
krzysz00: Structural note: maybe make `allocs` a vector of `tuple<GpuAllocOp, int64_t>` where the int64_t is the previous value of `totalSize`? That saves you from needing to recompute `offset` later.

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,563 @@
...
+    rewriter.setInsertionPointAfter(alloc);
+    Location loc = alloc->getLoc();
+
+    if (i == allocs.size() - 1) {
+      offset = 0;
+    }
+    LLVM_DEBUG(llvm::dbgs() << "offset: " << offset
+                            << " numElements: " << numElements << "\n");
+
+    bool replaceWithSubView = offset == 0;
```
krzysz00: I'd argue that unconditional replacement with a `memref::ViewOp` is the right call here for uniformity's sake. That preserves the invariant that LDS allocations are byte buffers that are `memref.view`'d into their actual type

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,563 @@
...
+  }
+  ArrayAttr idToLDS = maybeIdToLDS.value();
+
+  int64_t mPerBlock = getLowerShape(idToLDS)[0];
+  int64_t nPerBlock = getLowerShape(idToLDS)[1];
+
+  return std::make_tuple(mPerBlock, nPerBlock, idToLDS);
+}
+
+struct ThreadwiseWriteAllRewritePattern
```
krzysz00: If you want the overall condition for when to apply this - if you take the vectorization of `iter` in the original maps and it's bad, you'll want to do this swizzle.

Which brings me to my standing refactoring of getting rid of that "extra maps" argument on threadwise_write_all, but if we did that here we'd need to agree on a marker of some sort for where the GEMM rewrites start - which could just be an arbitrary attribute.

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,563 @@
...
+void reuseDeadLDS(func::FuncOp &func) {
+  // Retrieve all GpuAlloc operations targeting LDS and
+  // calculate the total memory space required for all operations
+  // except the last one. Additionally, compute the memory space
+  // required by the last GpuAlloc operation.
+  SmallVector<GpuAllocOp> allocs;
+  int64_t totalSize = 0;
+  int64_t lastSize = 0;
+  func.walk([&](GpuAllocOp gpuAlloc) {
+    auto type = gpuAlloc.getOutput().getType();
```
krzysz00: Might be good to spell out `MemRefType` here - upstream style (which we broadly try to hold to) says to use auto only in cases where the type is either obvious from context (like in those `dyn_cast<>`s below) or long and pointless to name (ex. various iterators)

krzysz00 (commented):
Ok, so, large amount of trivialities aside:

1. Good stuff!
2. I'd like to commend your defensive programming style throughout
3. Could you add a test to make sure this pass is doing what you think it does?

(Also, you'll need to update the `pipelines.mlir` test)

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
dhernandez0: I understand you mean just after createRockBlockwiseGemmToThreadwisePass()? 

When I tried that there was an issue for fp16. The code generated did not use "global_store_dwordx4". I think the issue is that "destType" is fp32, so "elementsWrittenPerThread = 128 / destType.getIntOrFloatBitWidth()" is 4, after that there is a linalg that truncates the result to fp16, so moving it after linalg makes it work and generate "global_store_dwordx4" stores.

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
dhernandez0: In GPUDialect, however I was told not to introduce that dependency here? should I move those utility functions somewhere else or leave this for a refactor PR?

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
dhernandez0: Sorry, I'm not sure how to do this, how can I pass a list of operations in one call?

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
krzysz00: ... GPUDialect is already a dependency?

mlir/include/mlir/Dialect/Rock/Passes.td
```diff
@@ -127,4 +127,9 @@ def RockCheckResidencyPass : Pass<"rock-check-residency"> {
   let dependentDialects = ["LLVM::LLVMDialect"];
 }
 
+def RockGemmOutputSwizzlePass : Pass<"rock-gemm-output-swizzle", "::mlir::func::FuncOp"> {
+  let summary = "This pass changes ThreadwiseWriteAllOp to swizzle on LDS before writing to GPU memory";
+  let dependentDialects = ["rock::RockDialect", "affine::AffineDialect", "gpu::GPUDialect", "vector::VectorDialect", "memref::MemRefDialect", "linalg::LinalgDialect", "scf::SCFDialect"];
```
krzysz00: You only need to depend on a dialect if you're referring to it in the pass and can't rely on it already being present. So you'll want `memref::MemRefDialect` and `rock::RockDialect` but you don't need `SCF`, say. (Heck, check your `#include`s to see if you're missing any)

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
krzysz00: Ah

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
dhernandez0: you're right!

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -523,12 +543,16 @@ void RockGemmOutputSwizzlePass::runOnOperation() {
     config.strictMode = GreedyRewriteStrictness::ExistingOps;
     if (failed(applyOpPatternsAndFold(writes, std::move(patterns), config))) {
       signalPassFailure();
+      return;
```
krzysz00: Weird little note that you can write `return signalPassFailure();`

mlir/test/Dialect/Rock/lowering_output_swizzle.mlir
```diff
@@ -0,0 +1,40 @@
...
+  return
+}
```
krzysz00: Nit: newline

mlir/test/Dialect/Rock/lowering_output_swizzle.mlir
```diff
@@ -0,0 +1,40 @@
...
+  %19 = arith.muli %18, %c22 : index
+  %20 = arith.subi %c20, %19 : index
+  %21 = arith.minui %20, %c22 : index
+  %22 = arith.remui %17, %21 : index
+  %23 = arith.addi %19, %22 : index
+  %24 = arith.remui %17, %c352 : index
+  %25 = arith.divui %24, %21 : index
+
+  // CHECK: rock.alloc() : memref<24576xi8, #gpu.address_space<workgroup>>
+  // CHECK: memref.view
```
krzysz00: Could we get some `CHECK-SAME: memref<... x ...>`s (and maybe some `CHECK-DAG %[[cN:.+]] = arith.constant N` above) to confirm that the views are what you think they should be?

krzysz00 (commented):
From a code level, this is approved

Can we perf this? (I can give you the "how I perf a PR" tutorial son if no one else does)

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
dhernandez0: this function returns void, should I still do that?

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
krzysz00: The catch with not doing this is that all the fusions load in the less coalesced layout, which might matter

So what might make sense is to go crack open the vectorizer to track changes to `destType`

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,556 @@
...
+      if (auto elemVecType = dyn_cast<VectorType>(destElemType)) {
+        LLVM_DEBUG(llvm::dbgs() << "ThreadwiseWriteAllOp saves a vector type"
+                                << ", skipping swizzle\n");
+      } else {
+        size_t extraIdxCount = threadwiseWriteAll.getExtraIndices().size();
+        VectorizationResult vectorRes =
+            getMaxVectorization(destView, extraIdxCount);
+        int64_t vectorLenBits =
+            vectorRes.max * destElemType.getIntOrFloatBitWidth();
+        if (vectorLenBits == 128) {
```
krzysz00: Yeah, I'm not sure this is what we want for fp16? Maybe `>= 4` here? Worth checking - though I would'n re-tune to test this

krzysz00 (commented):
A few notes, including on placing this pass relative to align-tiling

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,571 @@
+//===- GemmOutputSwizzle - MLIR Rock ops lowering passes -----===//
```
manupak: naming: Its not just "Gemm".  This should generally work for any kernel that writes to global memory in the end.

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,571 @@
...
+
+static std::optional<std::tuple<int64_t, int64_t, ArrayAttr>>
+getIdToLDS(ThreadwiseWriteAllOp &op, OpBuilder &b) {
+  ArrayAttr srcTransform = op.getExtraViewsAttr();
+  SetVector<StringRef> dimensionsToRemove;
+  dimensionsToRemove.insert("g_block");
+  dimensionsToRemove.insert("m_block");
+  dimensionsToRemove.insert("n_block");
+  // for attention:
+  dimensionsToRemove.insert("gblock");
```
manupak: feel free to unify the names.

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,571 @@
...
+
+    // Obtain critical matrix dimensions.
+    ArrayRef<int64_t> cShape;
+    cShape = op.getDest().getType().getShape();
+    if (cShape.size() != 3) {
+      return failure();
+    }
+    int64_t G = cShape[0];
+    int64_t M = cShape[1];
+    int64_t N = cShape[2];
```
manupak: Why do you need to know G.M and N ?

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,571 @@
...
+    }
+    FailureOr<IntegerAttr> maybeBlockSize = getBlockSize(op);
+    if (failed(maybeBlockSize)) {
+      return failure();
+    }
+    int64_t blockSize = maybeBlockSize.value().getValue().getSExtValue();
+    int64_t gridSize = maybeGridSize.value().getValue().getSExtValue();
+    bool useIndexDiffs = true;
+    bool forceUnroll = true;
+    int64_t ldsRequiredBytes = mPerBlock * nPerBlock * getByteWidth(destType);
```
manupak: This is not the right way to get this. You should already be getting this properly somewhere in "reuseLDS" code.
This is certainly not right for attention.

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,571 @@
...
+    constexpr int64_t dimensionM = 1;
+    constexpr int64_t dimensionN = 2;
+    int64_t dataPerThread = (nPerBlock * mPerBlock) / blockSize;
+    if ((nPerBlock * mPerBlock) % blockSize != 0) {
+      return failure();
+    }
+
+    VectorizationResult mVectorRes =
+        getMaxVectorization(matC, dimensionM, /*inputDimLen=*/
+                            dataPerThread, matC.getDefiningOp());
```
manupak: You should not need to give dataPerThread here. Thus, all the dependent information extracted above should not be needed.

manupak: but feel free to explain if I am wrong.

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,571 @@
...
+
+    // LDS barrier.
+    // This barrier is needed because we reuse A and B buffers
+    b.create<LDSBarrierOp>(loc);
+
+    // Store C results to LDS.
+    b.create<ThreadwiseWriteAllOp>(loc, convertedC, ldsBufferMNToRaw,
+                                   /*extraViews=*/idToLDS,
+                                   /*extraIndices=*/ValueRange{tid},
+                                   op.getFeatures(), StoreMethod::Set,
```
manupak: dont hardcode the `StoreMethod::Set`

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,571 @@
...
+    int64_t blockSize = maybeBlockSize.value().getValue().getSExtValue();
+    int64_t gridSize = maybeGridSize.value().getValue().getSExtValue();
+    bool useIndexDiffs = true;
+    bool forceUnroll = true;
+    int64_t ldsRequiredBytes = mPerBlock * nPerBlock * getByteWidth(destType);
+
+    // Decide register vectorization.
+    constexpr int64_t dimensionM = 1;
+    constexpr int64_t dimensionN = 2;
+    int64_t dataPerThread = (nPerBlock * mPerBlock) / blockSize;
```
manupak: Please extract this from `threadwise_write_all` maps not like this. This is not general enough.

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
manupak: ok after reading a bit more you need dataPerThread  for something else... but please extract properly from `threadwise_write_all`

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,571 @@
...
+      LLVM_DEBUG(llvm::dbgs() << "dim = M\n");
+    } else {
+      LLVM_DEBUG(llvm::dbgs() << "dim = N\n");
+    }
+
+    Value finalC =
+        gpuAlloc(b, loc, dataPerThread, destType, AddressSpace::Private);
+
+    TopDownTMBuilder tidIterMerge(b, {"tid", "iter"},
+                                  {blockSize, dataPerThread});
```
manupak: So yea -- `dataPerThread` should be extracted from original `threadwise_write_all` for stability of this

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,571 @@
...
+      return failure();
+    }
+    bool isAttention = op.getExtraIndices().size() == 3;
+    LLVM_DEBUG(llvm::dbgs() << "isAttention: " << isAttention << "\n");
+    if (isAttention) {
+      // attention
+      dataPerThread = dataPerThread * mBlocks;
+      mBlocks = 1;
+      bidGridLengths = {G, nBlocks, blockSize, dataPerThread};
+      bidGridOrder = {"g_block", "n_block", "tid", "iter"};
```
manupak: These are more than grid dimensions. Therefore I d not name like "bidGrid*".

These are upperView of threadwise_write_all. Therefore, I dont see the requirement to construct it like this.
Can we just extract them instead?

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,571 @@
...
+    // Store C results to LDS.
+    b.create<ThreadwiseWriteAllOp>(loc, convertedC, ldsBufferMNToRaw,
+                                   /*extraViews=*/idToLDS,
+                                   /*extraIndices=*/ValueRange{tid},
+                                   op.getFeatures(), StoreMethod::Set,
+                                   /*forceUnroll=*/forceUnroll,
+                                   /*useIndexDiffs=*/useIndexDiffs);
+
+    // Load from LDS to registers.
+    int64_t elementsWrittenPerThread = 128 / destType.getIntOrFloatBitWidth();
```
manupak: There is an underlying assumption this always succeed for dwordx4 vectorization.
I d feel safer to use maxVecLen instead here.

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
```diff
@@ -0,0 +1,571 @@
...
+        TopDownTMBuilder::below(createMblock, createMblockAttr);
+    tidIterMergeMem.passThrough({"g_block", "m_block", "n_block", "tid"});
+    tidIterMergeMem.merge({"iter", "numElements"}, {4, 5}, "iter",
+                          {iter, elementsWrittenPerThread});
+    auto tidIterMergeMemAttr = tidIterMergeMem.get();
+
+    TopDownTMBuilder tidIterFlattenMem =
+        TopDownTMBuilder::below(tidIterMergeMem, tidIterMergeMemAttr);
+    tidIterFlattenMem.passThrough({"g_block", "m_block", "n_block"});
+    tidIterFlattenMem.unmerge("flattenBlock", 3, {"iter", "tid", "numElements"},
```
manupak: Will it be possible factor out that does the transformation of `tid, iter --> iterVec, tid, VecItem` + flattenting?
I think logic would be to passthrough everything else.

manupak (dismissed):
Thanks for great work!
I really appreciate tackling of this somwhat complex task as a starter.

I basically only reviewed the Pass you added. I ll take a look at non-Pass changes in the next round.

Im putting a blocker just because I see few critical things that needs to be fixed but majority of the comments are for non-critical things.

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
manupak: i.e. walk the code and fetch the allocs

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
manupak: i take that back this is for LDS.

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
dhernandez0: I get mPerBlock, nPerBlock and destType from ThreadwiseWriteAllOp. For mPerBlock, nPerBlock I use removeUpperDims and getLowerShape to get the right m and n per block. 

The goal of this check is to see if there is enough LDS memory for this pass: saving the block results into LDS. 

Sorry, maybe there's something I'm missing?

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
dhernandez0: I think this is related to a previous comment, dataPerThread = (nPerBlock * mPerBlock) / blockSize. And nPerBlock, mPerBlock are extracted from threadwise_write_all. Is this correct?

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
dhernandez0: nPerBlock, mPerBlock are extracted from threadwise_write_all. Is this correct?

manupak: I mean dataPerThread could be extracted directly...

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
krzysz00: I figure this is for checking if the pass should be applied?

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
krzysz00: These are for testing whether the pass should be applied, and `dataPerThread` is "how muth data would be copied per thread if this pass were applied"

mlir/lib/Dialect/Rock/Transforms/GemmOutputSwizzle.cpp
krzysz00: Isn't the point here to reconstruct the view to account for the swizzles?

mlir/lib/Dialect/Rock/Transforms/OutputSwizzle.cpp
```diff
@@ -0,0 +1,579 @@
...
+    if (!maybeBlockInfo.has_value()) {
+      return failure();
+    }
+    std::tie(dim0PerBlock, dim1PerBlock, idToLDS) = maybeBlockInfo.value();
+
+    // Obtain critical matrix dimensions.
+    ArrayRef<int64_t> cShape;
+    cShape = op.getDest().getType().getShape();
+    if (cShape.size() != 3) {
+      return failure();
```
manupak, line 255: add a debug when this happens

mlir/test/Dialect/Rock/lowering_output_swizzle.mlir
```diff
@@ -0,0 +1,60 @@
+// RUN: rocmlir-opt --rock-output-swizzle %s | FileCheck %s
```
manupak: I d like to see another test that runs passes upto & including -rock-threadwise-gemm-lowering which lowers `threadwise_write_all` where we can see the vectorization has actually changed. Will that be a possible test to add ?

manupak (commented):
Code-wise looks good to me.
I d like a test that confirms the vectorization -- as a way to make sure this is not broken in future. I ve added details in the inline comment.

manupak (commented):
looks good to me. Before I hit the button -- can we do quick check in best-5 and worst-5 configs to see the perf still holds ?

manupak (approved):
LGTM! thanks for the all hard work.

krzysz00 (approved):

PR #1581, opened 2024-07-19
Add reduce as a fusor in regularization

krzysz00:
rock.reduce is an operation that can write intermediate buffers and get fused in to things, just like linalg.generic. However, it wasn't being treated as a fusor by the regularization pass, breaking the "no transforms on readers" invariant that the gridwise-gemm-to-blockwise ..= align-tiling pipeline relies on.

This patch fixes this issue.

Thanks to @manupak in PR #1579 for the initial approach and the tests

Fixes https://github.com/ROCm/rocMLIR-internal/issues/1541


manupak (approved):
LGTM! especially on the premise of not blocking a bug fix to go in.

but we can have that chat about why we need "regularization" once im back from holidays.
I still disagree (or do not understand) about the existence of this pass and what it claims to be solving. (I have further comments in my PR).

PR #1582, opened 2024-07-19
Fix `mhal::PrefillAttr` data retrieval after switching to target attributes

fabianmcg:
This patch updates `mhal::getStoredPrefillAttributes`, `mlirGetNumPrefillArgs`, and `mlirGetPrefillArgsInfo` to work on GPU binaries.

Previously `mhal::PrefillAttr` attributes were stored on the GPU module discardable attributes. However, now they are stored in the property dictionary of `#gpu.object` attributes.

This patch fixes: https://github.com/ROCm/rocMLIR-internal/issues/1543
However, the tests remain de-activated.


krzysz00 (approved):
Approved

PR #1584, opened 2024-07-23
[DO NOT SQUASH!] Non-MIGraphX-related changes for int4 support

krzysz00:
The details are in individual commit messages, but, to summarize:
- Upstream extensions to `EmulateNarrowType` in `Memref` that we needed
- A pass in `mhal` that runs said upstream narrow type emulation on host/test code
- `rock-emulate-narrow-type` that runs a more comprehensive (that is, it handles vectors) set of rewrites on GPU code (and non-MHAL host code)
- Tests, including a basic integration test for an `i4 * f16` gemm

Overall, we handle i4 by just letting i4 memrefs happen and then rewriting indexing on i4 to indexing on i8 using the patterns upstreamed last year to enable exactly this sort of thing

Next steps are hooking up `migraphx.pack` to this (probably by having it rewrite function arguments to `i4` type and adding a generic that does the `extui` at the right spot)


mlir/test/Dialect/MHAL/emulate-narrow-type.mlir
```diff
@@ -0,0 +1,31 @@
+// RUN: rocmlir-opt -mhal-emulate-narrow-type -canonicalize \
```
manupak, line 1: This change is in wrong commit. Please move it to the MHAL commit

mlir/lib/Dialect/Rock/Transforms/EmulateNarrowType.cpp
```diff
@@ -0,0 +1,256 @@
...
+                                                  boundaryPatterns);
+  mhal::populateMHalNarrowTypeEmulationBoundaryPatterns(typeConverter,
+                                                        boundaryPatterns);
+  if (failed(applyPartialConversion(op, boundaryTarget,
+                                    std::move(boundaryPatterns))))
+    return signalPassFailure();
+
+  RewritePatternSet patterns(ctx);
+  memref::populateMemRefNarrowTypeEmulationPatterns(typeConverter, patterns);
+  vector::populateVectorNarrowTypeEmulationPatterns(typeConverter, patterns);
```
manupak, line 243: Why cant we run the seperate passes ?

mlir/lib/Dialect/Rock/Transforms/EmulateNarrowType.cpp
```diff
@@ -0,0 +1,256 @@
...
+  mhal::populateMHalNarrowTypeEmulationPatterns(typeConverter, patterns);
```
manupak, line 244: Can we run this later in the kernel pipeline ? I d like to keep the codegen part out of mhal as much as possible.
In other words, what is importance that we need to run this now ?

manupak (commented):
Overall implementation-wise I dont have issue.

However, I m not liking globbing emulate narrow type patterns inside a -rock pass.
Why cant we do them seperately? (especially mhal one)

mlir/test/Dialect/MHAL/emulate-narrow-type.mlir
krzysz00: It's not in an MHAL commit because it's not being applied to the MHAL directory

mlir/lib/Dialect/Rock/Transforms/EmulateNarrowType.cpp
krzysz00: ... because they all share a common type converter and, more fundamentally, there _isn't_ a separate pass. (There's a test pass, but, well, that's for testing)

This is us embracing the developing expectation that downstream projects build their own pass - or at least that makes sense given that all the code came from IREE

mlir/lib/Dialect/Rock/Transforms/EmulateNarrowType.cpp
krzysz00: 1. We need to run this before we hit `gpu.func` so we don't have to go writing rewrites for all the GPU operations
2. This only runs on the kernels - we're calling MHAL to get the conversions for function boundaries

manupak:
> This is us embracing the developing expectation that downstream projects build their own pass - or at least that makes sense given that all the code came from IREE.

I cant help but feel the odd how this is all designed to be use. .(maybe its just me)

So basically this seems like cross-dialect transformation.
Therefore, why cant we have an OpInterface -- say EmulatableNarrowTypeOpInterface -- that allows operations in different dialects to "subscribe" to be part of it. 

Then have an upstream pass to that works on EmulatableNarrowTypeOpInterface to perform the rewrite thoughout. Only argument against that would be if we do not want some ops that implements EmulatableNarrowTypeOpInterface  not to be rewritten -- is that a thing ?

cc: @MaheshRavishankar

MaheshRavishankar:
> > This is us embracing the developing expectation that downstream projects build their own pass - or at least that makes sense given that all the code came from IREE.
> 
> I cant help but feel the odd how this is all designed to be use. .(maybe its just me)
> 
> So basically this seems like cross-dialect transformation. Therefore, why cant we have an OpInterface -- say EmulatableNarrowTypeOpInterface -- that allows operations in different dialects to "subscribe" to be part of it.
> 
> Then have an upstream pass to that works on EmulatableNarrowTypeOpInterface to perform the rewrite thoughout. Only argument against that would be if we do not want some ops that implements EmulatableNarrowTypeOpInterface not to be rewritten -- is that a thing ?
> 
> cc: @MaheshRavishankar

Having an interface would be a good idea. If someone sends patches upstream to refactor this into interfaces i'd be happy to review (but I dont know the shape of the interface). Matter of preference maybe. End of the day the only difference is how you  dispatch to a particular line of code. Interfaces I think should be designed a bit more carefully cause it is harder to wind them back, and hard to do in a pinch.

manupak:
> Having an interface would be a good idea. If someone sends patches upstream to refactor this into interfaces i'd be happy to review (but I dont know the shape of the interface). Matter of preference maybe. End of the day the only difference is how you dispatch to a particular line of code. Interfaces I think should be designed a bit more carefully cause it is harder to wind them back, and hard to do in a pinch.

Thanks makes sense. I ll see if I can gather some thoughts on a upstream discussion to test the inertia of the idea.

Im of the opinion when the interfaces are done for a very specific singular purpose (far away of hooks like "cannonicalize" where it could ne very ambigous) we could improve the core components bit more generically. I feel this is a good candidate for that, especially since there are examples of different dialects implementing the transformation similarly. 

So yea having said that I am understanding this is indeed how this is intended to be consumed as of now.

mlir/test/Dialect/MHAL/emulate-narrow-type.mlir
manupak: yep thats the problem... the pass definition belongs to /external/mhal ?
Why do we have the test in rocMLIR

mlir/lib/Dialect/Rock/Transforms/EmulateNarrowType.cpp
manupak: deferring this to later based on the recent discussion on how upstream could be doing this.

manupak:
@krzysz00 This all looks good except for that fact that we have a test on mhal pass on rocmlir. Any reasons that Im missing ?

krzysz00:
Re the MHal test, https://github.com/ROCm/rocMLIR-internal/issues/1564 is the ticket for giving MHal a test suite, which it doesn't currently have. I'd rather not pull "add a test suite and hook it up to our CI" into this commit

manupak (approved):
Thanks! LGTM

PR #1585, opened 2024-07-24
Remove other traces of gfx906, missed last time.

pcf000:
Missed this bit, and it causes errors in the public weekly, where we turn canXdlops off.  Don't force the name to contain "gfx906," just unstash what's there.

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -76,7 +76,7 @@ String dockerImageCIMIGraphX() {
 
 void preMergeCheck(String codepath) {
     // Only do static check on mfma codepath during PR CI
-    if ( (params.nightly == false) && (codepath == "mfma") ) {
+    if ( !params.nightly && (codepath == "mfma") ) {
         echo "Performing Static Test (preMergeCheck)"
         sh '''
         if [ ! -f ./build/compile_commands.json ];  then
@@ -100,7 +100,7 @@ void preMergeCheck(String codepath) {
 
 void preMergeCheckPackage(String codepath) {
     // Only do static check on mfma codepath during PR CI
-    if ( (params.nightly == false) && (codepath == "mfma") ) {
+    if ( !params.nightly && (codepath == "mfma") ) {
         echo "Checking if the fat library target list is accurate"
         dir('build') {
             sh '../mlir/utils/jenkins/static-checks/get_fat_library_deps_list.pl > ./librockcompiler_deps.cmake.new'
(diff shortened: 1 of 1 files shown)


krzysz00 (approved):

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -76,7 +76,7 @@ String dockerImageCIMIGraphX() {
 
 void preMergeCheck(String codepath) {
     // Only do static check on mfma codepath during PR CI
-    if ( (params.nightly == false) && (codepath == "mfma") ) {
```
pcf000, line 79: Gratuitous change to all "== false" and the like, to make them concise.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -316,31 +316,33 @@ void parameterSweep(String CONFIG, String codepath) {
...
-    if (params.canXdlops && (params.disableNavi3x == false) && (codepath == "navi3x")) {
+    if (params.canXdlops && !params.disableNavi3x && (codepath == "navi3x")) {
         return true
     }
     return false
 }
 
 boolean shouldRunFromChip(String chip) {
     switch (chip) {
-        case "gfx906":
```
pcf000, line 340: Summary:  default is vanilla, gfx90a is allowed for either mfma or vanilla (because that's all we have on the public side, where vanilla is run), and gfx906 isn't mentioned at all.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -354,10 +356,14 @@ boolean shouldRunFromChip(String chip) {
 void archivePerfDB() {
     // Note: add additional architectures here
     dir ('build/perfDB') {
-        unstash name: "MLIR-PerfDB-${params.canXdlops ? 'gfx908' : 'gfx906'}"
-        unstash name: "MLIR-PerfDB-${params.canXdlops ? 'gfx90a' : 'gfx906'}"
-        if (params.disableNavi21 == false)
-            unstash name: "MLIR-PerfDB-${params.canXdlops ? 'gfx1030' : 'gfx906'}"
+        if (params.canXdlops) {
```
pcf000, line 359: Unfortunately there isn't a conditional unstash;  if it isn't present, it errors.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -772,6 +778,10 @@ pipeline {
                 beforeAgent true;
                 equals expected: true, actual: params.weekly;
                 equals expected: true, actual: params.staticLib;
+                anyOf {
```
pcf000, line 781: Clause added so the stash and unstash conditions are the same.

PR #1586, opened 2024-07-25
Change BufferDependencyAnalysis to use OpOperand* in its internals

krzysz00:
I'm refactoring the regularization pass to correctly handle multiple readers from a fusor's buffer. As part of this, I realized I should go ahead and migrate that pass to BufferDependencyAnalysis as planned.

While thinking about that, I realized that the reader and writer lists should actually consist of OpOperand*'s (that is, effectively, an operation and a position within that operation's operands - or, equivalently, a Value and a way to identify a specific use of that value). This allows users of the analysis to, for example, pick out the (probably transformed) value that a reader or writer is operating on, which'd allow for, for example, calls to isolateTransforms() on that value.

This commit also adds the analyze() method so that when new buffers are added or existing buffers have their use lists modified, the results of the analysis can be recomputed.

This commit doesn't directly fix any bugs but does pave the way to fixes for our reduction fusion issues.

### DIFF
--- a/mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
+++ b/mlir/include/mlir/Analysis/BufferDependencyAnalysis.h
@@ -27,16 +27,16 @@ namespace mlir {
 struct BufferDependencyAnalysis {
   BufferDependencyAnalysis(Operation *op);
 
-  std::optional<llvm::SmallVector<Operation *>>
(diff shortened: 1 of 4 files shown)


manupak (approved):
LGTM!

ravil-mobile (approved):

PR #1590, opened 2024-07-29
Fix multi-output fusion bugs by reworking regularization

krzysz00:
Fixes https://github.com/ROCm/rocMLIR-internal/issues/1550

fixes https://github.com/ROCm/rocMLIR-internal/issues/1546

Depends on https://github.com/ROCm/rocMLIR/pull/1586

This commit rewrites --rock-regularize to have it handle more cases. Now, the regularizer directly uses the new buffer dependency analysis and looks through that information in order to not have to grow its own idea fo when to fuse things.

Secondly, the regularizer now fully collects which linalg.generic opesaritos are input fusions and which ones are output fusions, so that we can correctly move transforms from the generic to its readers when we're in an input fusion and move transforms from the readers to the writing generic in an output fusion. The old ad-hoc fusor detector wasn't really working.

Third, we extend the buffer dependency analysis to allow for updates and to allow the inverse lookup (given an OpOperand, determine which buffer it reads or writes).

We also now correctly handle the case where we have

```
%buf = alloc()
%c = f(%buf)
%d = g(%buf)
%e = h(%buf)
rock.gridwise_gemm_* %c = ...
linalg.generic ins(%d, ...)
linalg.generic ins(%e, ...)
```

by inserting memref.copy operations as nedede to isolate the two transform chains. These copies will be eliminated during AlignTiling if possible (sometimes they are legitimate copies, like, for exaple, "write out the GEMM results (with regular writes)" and "write a copy of those results through this reduction view and atomic_add") - those'll get cleaned up by mem2reg.

Overall, by not trying to rely on the pattern rewriter and doring the IR transformations we care about in place, we've cleaned up the regularizer.

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
@@ -182,7 +182,63 @@ static void annotateGenericOp(linalg::GenericOp lgop) {
 ////  Classify buffer writes into input and output fusion chains
 ////////////////////////////////////////////////////////////////////////
 
+/// The overall goal of this pass is to make sore you can go from the operands
+/// of the fusion root (usually a GEMM or attention) to function arguments
+/// through any intervening fusable operations (generally linalg.generic
+/// operations or reductions) without needing to invert any transformations.
+/// This allows us to accurately (assuming the heuristic for multiple-output
+/// input fusions isn't meaningfully incorrect) analyze the memory layout of the
+/// underlying memory despite any intervening fusion code. This is important,
+/// especially in the case of input fusions, as we rely on those vectorization
+/// analyses in order to correctly select elements of our schedule, like whether
+/// to do a bank conflict avoidance swizzle.
+///
+/// To achaive this, we traverse the writeOperands starting from the output
+/// buffer of each "fusion root" (the C algument of GEMM, the output of
+/// attention, etc) and record all the buffer-writing operands on the way from
+/// that argument to a function result (such as the outputs of an intervening
+/// linalg.generic) and classify them as output fusion writes. That is, output
+/// fusion writes are all those that will be replaced by writes to a register
+/// buffer during fusion. We then classify all the buffer writes between
+/// funciton inputs and the read operands of a fusion root as input fusion
+/// writes, along with the writes needed to produce any elementwise arguments to
+/// operations in the output fusion path (ex. if the bias elements are
+/// constructed by a linalg.generic, the output operand of that generic is an
+/// input fusion write). Equivalently, input fusion writes are those that become
+/// reads into a register buffer after fusion.
+///
+/// Once we have done this, we regularize the function as follows:
+/// - If the operand that points at a buffer write is on the output path,
+/// all the readers of that buffer must refer to that buffer directly and not
+/// through transforms. If there are such transforms, we invert them and replace
+/// the buffer with one whose type is the one expected by the reading operation.
+/// Shoulld there be multiple such operations that disagree about what the type
+/// of that buffer should be, we copy the original output buffer so that the
+/// inversions can be performed independently.
+/// - Conversely, for operands on the input path, the operation that writes
+/// to the buffer has to refer to the bufffer directly, so any views on that
+/// writing operand are inverted and those inverses become part of the view
+/// stack for each operation reading the buffer. This lets us traverse from
+/// the writing operand to its input during a vectorization analysis without
+/// having to invert the maps on the output.
+
 namespace {
+/// Information needed while regularizing the view chains to allow vectorization
+/// and permit fusion.
+/// `worklist` is the set of memref.alloc ops that haven't been processed yet.
+/// When we do a non-trivial regularization, we always add the new buffer
+/// to the worklist to ensure that the regularization invariants are mantained.
+/// `inputFusionWrites` is the set of `OpOperand*`s that point at arguments
+/// of operations that write to a buffer where the underlying operation is on
+/// an input fusion chain.
+/// `outputFusionWrites` is similarly the set of operands indicating a
+/// potentially transformed buffer that's written during the main output fusion
+/// chain. These sets are needed to resolve the ambiguity abut how to regularize
+/// a `linalg.generic` - that is, can its `out` have a view on it or not? Note
+/// that this is mutable state because, in the case of conflicting
+/// transformations in an output fusion, we'll need to copy the intermediate
+/// buffer so we can invert the conflicting transformations onto the writes
+/// of each copy.
 struct TransformPushState {
   llvm::SmallVector<memref::AllocOp> worklist;
   llvm::SmallPtrSet<OpOperand *, 8> inputFusionWrites;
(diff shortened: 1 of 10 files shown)


mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
```diff
@@ -147,187 +146,452 @@ struct RegularizeGenericRewritePattern
...
-        auto arg = dyn_cast<BlockArgument>(inp);
-        auto shape = cast<ShapedType>(inp.getType());
-        int64_t argSize = shape.getNumElements();
-        if (inputIdx == 0 || argSize > majorTensorSize ||
-            (argSize == majorTensorSize && argIdx > arg.getArgNumber())) {
-          majorTensorIdx = inputIdx;
-          majorTensorSize = argSize;
-          argIdx = arg.getArgNumber();
-        }
+static void annotateGenericOp(linalg::GenericOp lgop) {
```
manupak, line 149: nit : style

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
```diff
@@ -147,187 +146,452 @@ struct RegularizeGenericRewritePattern
...
-      result = top.getResult();
-      if (!result.hasOneUse()) {
-        // currently restricted to 1 reader
-        LLVM_DEBUG(llvm::dbgs() << "multiple readers on transform\n");
-        return false; // TODO: fix when encountered
-      }
-      chain.push_back(forwOp);
-      forwOp = (*result.getUses().begin()).getOwner();
+namespace {
+struct TransformPushState {
```
manupak, line 242: some docs here please

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
```diff
@@ -147,187 +146,452 @@ struct RegularizeGenericRewritePattern
...
+  llvm::SmallVector<memref::AllocOp> worklist;
+  llvm::SmallPtrSet<OpOperand *, 8> inputFusionWrites;
+  llvm::SmallPtrSet<OpOperand *, 8> outputFusionWrites;
+};
+} // end namespace
+
+/// Given an operand that reads a buffer, collect the writer of thet buffer and,
```
manupak: What does an operand that reads a buffer mean here ?

In bufferized IR, we have ops with OpOperands that are usually buffers. Therefore I struggle follow this docs what does a operand that reads a buffer mean ?

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
```diff
@@ -147,187 +146,452 @@ struct RegularizeGenericRewritePattern
...
+/// recursively, all the writers of the buffers that writer reads, and so on.
+/// This is meant to collect all the buffer writes involved in an input fusion.
+static void collectInputFusionWrites(OpOperand *reader,
+                                     const BufferDependencyAnalysis &bufferDeps,
+                                     TransformPushState &state) {
+  std::optional<memref::AllocOp> maybeBuffer = bufferDeps.getReadBuffer(reader);
```
manupak: so yea, whose operand that would not be a buffer in this case ?

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
```diff
@@ -147,187 +146,452 @@ struct RegularizeGenericRewritePattern
...
-      while (auto top = dyn_cast<rock::TransformOp>(useOp)) {
-        result = top.getResult();
-        useOp = (*result.getUses().begin()).getOwner();
+/// Given the operand that writes to a buffer, collect the writes of its reader,
+/// and the writes of that reader, and so on until you're no longer writing to a
+/// buffer. If any auxiliary inputs come from buffers themselves, collect
+/// their writes as input fusions, because, in a fusion context, those other
+/// writers will be tiled by a threadwise_read_into, not a threadwise_write_all.
+static void
+collectOutputFusionWrites(OpOperand *writer,
```
manupak: Same confusion here around how an OpOperand be a writer... what am I missing ?

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
```diff
@@ -147,187 +146,452 @@ struct RegularizeGenericRewritePattern
...
+                              TransformPushState &state) {
+  bool foundFusionRoot = false;
+  kernel.walk([&](Operation *op) {
+    llvm::TypeSwitch<Operation *>(op)
+        .Case(
+            [&](memref::AllocOp allocOp) { state.worklist.push_back(allocOp); })
+        .Case([&](GridwiseGemmOp gemmOp) {
+          foundFusionRoot = true;
+          collectOutputFusionWrites(&gemmOp.getCMutable(), bufferDeps, state);
+          collectInputFusionWrites(&gemmOp.getAMutable(), bufferDeps, state);
```
manupak: See all these are buffers; not readers or writers. Therefore, collectXFusionWrites API is a bit confusing...

manupak (commented):
I did a pass. I have a question that kind of blocks me from understanding few functions used in the overall functionality.

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
```diff
@@ -147,187 +146,452 @@ struct RegularizeGenericRewritePattern
...
+/// opposite of what the fusor does. Before this pass runs, both the fusor and
+/// the fusee may access that intermediate buffer (which will be rewritten away
+/// by fusion) through some sequence of view operations.
+///
+/// After these rewrites, we have the following invariants, which make the
+/// fusion code itself (it's down in AlignTiling.cpp) much simpler and enables
+/// analysis of vectorization within the parts of the pipeline between this pass
+/// and AlignTiling (which can be important for key performance queries).
+///
+/// - The fusee always has a direct reference to the buffer
```
manupak: please define buffer. 

This should be the common buffer between fusee and fusor.

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
```diff
@@ -147,187 +146,452 @@ struct RegularizeGenericRewritePattern
...
+///
+/// Furthermore, if multiple fusees have such a non-direct reference, the
+/// intermediate buffer will be duplicated so that each transform stack can be
+/// inverted in isolation.
+
+/// Invert the transform stack between `operand` and `oldAlloc`, replacing all
+/// users but that transform stack with a new allocation whose type is the type
+/// of the front of the transform stack. If the original transforms (and
```
manupak: The usage of phrase "front of the transform stack" is ambigous and confusing.

I d suggest to reword this to mean: 
1) we invert view stack between operand and oldAlloc
2) create a newAlloc with no views between operand.
3) append the inverted stack to connect rest of the users of oldAlloc to newAlloc

manupak (commented):
Some more comments to clarify the code.

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
```diff
@@ -147,187 +146,452 @@ struct RegularizeGenericRewritePattern
...
+}
+
+LogicalResult findFusionRoots(func::FuncOp kernel,
+                              const BufferDependencyAnalysis &bufferDeps,
+                              TransformPushState &state) {
+  bool foundFusionRoot = false;
+  kernel.walk([&](Operation *op) {
+    llvm::TypeSwitch<Operation *>(op)
+        .Case(
+            [&](memref::AllocOp allocOp) { state.worklist.push_back(allocOp); })
```
manupak: This will capture all the `memref::AllocOp` s in the function. yes ? 
If so why do I see worklist.push_back of allocs in fusionwrites collection functions ?

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
```diff
@@ -147,187 +146,452 @@ struct RegularizeGenericRewritePattern
...
+        .Case([&](GridwiseGemmOp gemmOp) {
+          foundFusionRoot = true;
+          collectOutputFusionWrites(&gemmOp.getCMutable(), bufferDeps, state);
```
manupak: Can we add a trait (rely on the memory effects) or an OpInterface like `RockFusorOp`  to call collectOutputFusionWrites or collectInputFusionWrites. For RMW, we can do collectOutputFusionWrites? 

Maybe an OpInteface might be better that could return getRegularizeInputs() & getRegularizeOutputs() ?
(Could be a future PR though)

manupak: also that include checkRegularizeInputs() and checkRegularizeOutputs() as a verifier to verify the "invariant"

manupak (commented):
Some further comments...

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
krzysz00: What's the style nit?

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
krzysz00: "Operand that reads a buffer" is the operand (Value + place in arg list + operation that's using it) that refers to a buffer that it'll read through some number of views

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
krzysz00: Function argument isn't a buffer for these purposes

Might need a s/buffer/alloc/ but also `BufferDependencyAnalysis` is focused on allocs so

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
krzysz00: The `writer` is, for example, the `c` argument on a GEMM or the output of a linalg.generic

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
krzysz00: For example, in a GEMM, the A and B arguments are readers - they have MemRead effects - and C is a writer

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
krzysz00: The extra push_back()s are guards against bugs - if you've already corrected the issue, you should be able to re-run the analysis again and have it be a noop. I remember leaving comments about that

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
manupak: What Im saying is operand itself is the buffer (or a view of it) and its not a reader or writer to the buffer itself.

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
manupak: A,B, and C are all buffers or view of them. They are not readers or writers.

To me readers/writers are somethings like : 
```mlir
%buffer1 = memref.alloc()
%buffer2 = memref.alloc()
threadwise_read_into %buffer1 -> %buffer2
threadwise_write_all %buffer2 -> %arg
```

So `threadwise_read_into` is a reader of `%buffer1` and a writer of `%buffer2`
The first OpOperand of  `threadwise_read_into` is ` %buffer1` that is actually a buffer or a view of it.
The second OpOperand of  `threadwise_read_into` is ` %buffer2` that is actually a buffer or a view of it.
None of the OpOperands, I would not classify as readers and writers.
Hence my confusion.

manupak:
Having slept on this...

I have higher order questions about the design of the pass. As this fixes the immediate bugs, we can defer implementation but I d like to understand why went onto create stateful walker based approach vs Visitor/PatternRewrite way of doing things.

I d started a discussion on this here : https://github.com/ROCm/rocMLIR/discussions/1594.
But I ll keep that seperate to the review of this version of implementation.

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
manupak: my bad. you are actually fixing the style

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
```diff
@@ -147,187 +146,502 @@ struct RegularizeGenericRewritePattern
...
+/// written, recurse to collect that write. This classifies the intermediate
+/// buffers for the output fusion chain as those that will be read from (their
+/// tile size will come from a threadwise_read_into) and those that will be
+/// written (their tile size will come from a threadwise_write_all).
+static void
+collectOutputFusionWrites(OpOperand *writeOperand,
+                          const BufferDependencyAnalysis &bufferDeps,
+                          TransformPushState &state) {
+  // This is, broadly, a cached untransform() except that it fails when you hit
+  // a function argument, which is what we want here.
```
manupak, line 302: thanks for this

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
```diff
@@ -147,187 +146,502 @@ struct RegularizeGenericRewritePattern
...
+                              const BufferDependencyAnalysis &bufferDeps,
+                              TransformPushState &state) {
+  bool foundFusionRoot = false;
+  kernel.walk([&](Operation *op) {
+    if (auto allocOp = dyn_cast<memref::AllocOp>(op)) {
+      state.worklist.push_back(allocOp);
+      return;
+    }
+
+    if (op->hasTrait<OpTrait::rock::FusionRoot>()) {
```
manupak, line 580: Nice! much cleaner :)

manupak (approved):
LGTM!

PR #1591, opened 2024-07-30
[EXTERNAL] Add missing llvm builders for ROCDL intrinsics

manupak:
Currently there were two missing instrinsic
builders for some new ROCDL intrinsics.

This commit fixes that.


krzysz00 (approved):
Approved, though these should've been autogenerated by LLVM_IntrOpBase - we should check for upstream differences or otherwise work out why this wasn't caught

manupak:
I agree but I rather do the change for all of ROCDL translations as a seperate PR rather than having this inconsistency.

krzysz00:
No, as in "it's weird that the builders exist upstream"

Unless part of the problem is that when these got added upstream, we forgot to add tests for the LLVM translation

PR #1596, opened 2024-08-03
MIGraphX changes for int4

krzysz00:
[Define migraphx.unpack operator, add unsigned integer conversions](https://github.com/ROCm/rocMLIR/commit/7b87361e9111d5b8f3b252d3a5a75f68480ff3f5)

We define migraphx.unpack to mirror the operaton MIGraphX uses to
decompress int4 buffers.

We add a `zeroExtend` option to `migraphx.convert` that doesn't exist
in MIGraphX so that we can represent the unpacking in a way that makes
more sense to us.

[Add rewrites for migraphx.unpack to the rocMLIR pipeline](https://github.com/ROCm/rocMLIR/commit/2aa63c8afe9eb48ff481faada685805cf7c368b5)

This commit handles MIGraphX's unpack operation (which is used to
unfold tensors of packed 4-bit integers into 8-bit integer tensors) by
exposing the underlying int4 types. This means that future operatons
don't have to deal with an expanding fusion, that the type system
knows the real type of all tensors, and that the upstream int4
load/store emulation works correctly.

To accomplish this, the unpack operations gets rewriten into on that
unpacks from a byte buffer into a tensor of int4 (adjusting the shape
nad stride appropriately) which is then turned into the underlying
byte buffer with a zero-extension (a newly added attribute) and a
reshape.

Then, patterns fold away unpack, turning transpose(unpack(x)) into
unpack(transpose(x)) (and similarly for reshape) until the unpack()
reaches the function border. There, the byte-tensor argument is
replaced with one that uses an int4 tensor.

Note that this is the first commit, I haven't written all the tests
yet, but I want to unblock MIGraphX.


mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -971,6 +981,42 @@ LogicalResult QuantizeLinearConverter::matchAndRewrite(
...
+  MIXRShapedType shapedResultType = op.getOutput().getType();
+  Type outElemType = shapedResultType.getElementType();
+  Type outType = getTypeConverter()->convertType(shapedResultType);
+  Value empty = rewriter.create<tensor::EmptyOp>(loc, outType,
+                                                 /*dynamicSize=*/ValueRange{});
+  SmallVector<AffineMap> iterationMaps(
+      2, rewriter.getMultiDimIdentityMap(shapedResultType.getRank()));
+  SmallVector<utils::IteratorType> iteratorKinds(shapedResultType.getRank(),
+                                                 utils::IteratorType::parallel);
+  auto converter = rewriter.create<linalg::GenericOp>(
```
manupak: Please dont do this in MIGraphXToTosa. 
Please add a new conversion pass : MIGraphXToLinalg and do it there.

Alternatively why cant you use `tosa.rescale` ?

mlir/lib/Dialect/MIGraphX/Transforms/RealizeInt4.cpp
```diff
@@ -0,0 +1,200 @@
+//===- RealizeInt4.cpp ------------------------------------===//
+//
+// Part of the rocMLIR Project, under the Apache License v2.0 with LLVM
+// Exceptions. See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (c) 2024 Advanced Micro Devices
+//
+//===----------------------------------------------------------------------===//
```
manupak, line 9: Please add more docs what is pass is supposed to be doing here.

manupak (dismissed):
Please either use `tosa.rescale` -- though Im not sure whether it works with int4.
If not please add a new conversion : MIGraphXToLinalg that runs after MIGraphXToTosa.

Dont do it under MIGraphXToTosa.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
krzysz00: I'm pretty sure those all are implicitly signed conversions

And ... I don't think tosa.rescale is the type change op, that's tosa.cast

... Also ... like, we need to implement zero-extend _somehow_, and the only tosa way to do it is to add some unrealized casts to `ui4`.

_and_ the Shaped -> Tensor type converter lives in MIGraphXToTosa, and that conversion needs to be done in one shot.

mlir/lib/Dialect/MIGraphX/Transforms/RealizeInt4.cpp
krzysz00: It's in the tablegen

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
krzysz00: (That is, strong disagree about not squeezing this in here)

manupak: > I'm pretty sure those all are implicitly signed conversions
> 
> And ... I don't think tosa.rescale is the type change op, that's tosa.cast
> 
> ... Also ... like, we need to implement zero-extend _somehow_, and the only tosa way to do it is to add some unrealized casts to `ui4`.

https://github.com/llvm/llvm-project/blob/0b2d50c08d2eadf32f685fcd5b5468bc2ec7a16d/mlir/test/Conversion/TosaToLinalg/tosa-to-linalg.mlir#L1146

yep they do it by unrealized casts. They kind of have support for that but I agree its wonky.

> the Shaped -> Tensor type converter lives in MIGraphXToTosa

Clearly this does not need to live here. It could live somewhere common so we have more lowerings from MIGraphX dialect.

> that conversion needs to be done in one shot.

Not really.. its fine to partially convert it TOSA. you just need pull out the Shaped -> Tensor type converter

mlir/lib/Dialect/MIGraphX/Transforms/RealizeInt4.cpp
manupak: oh ok.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
manupak: Shaped -> Tensor type converter is using tensor dialect, I presume ?

mlir/include/mlir/Conversion/RocMLIRPasses.td
```diff
@@ -37,6 +37,23 @@ def ConvertRockToGPUPass : Pass<"convert-rock-to-gpu", "ModuleOp"> {
 // TosaToRock
 //===----------------------------------------------------------------------===//
 
+def RocmlirTosaToLinalgPass : Pass<"rocmlir-tosa-to-linalg", ""> {
+  let summary = "Lower TOSA to Rock";
+  let description = [{
+    Pass that converts TOSA operations to bufferized Rock operations.
```
manupak: shouldn;t this be "Pass that converts custom TOSA operations to **linalg** operations"?

manupak (approved):
This all looks good to me. however, please change description of the pass and maybe the pass name to be `rock-custom-tosa-to-linalg` ?

mlir/include/mlir/Conversion/RocMLIRPasses.td
krzysz00: x.x

krzysz00:
@manupak I'm not sure Rock is the right name there because they're the RocMLIR custom tosa operations and not ones related to Rock (the kernel dialect)

But I'm not exttemely strongly attached to the name

manupak:
Ah I was emphasizing on the addition of `custom` not on `rock`.
Apologies here is what I meant then : `rocmlir-tosa-custom-to-linalg`

krzysz00:
Makes sense and can do

PR #1599, opened 2024-08-06
[CI] Add support for Navi4x architecture in nightly CI pipeline. 

stefankoncarevic:
This was initially added to nightly CI, as an optional check, until we see how stable the test execution will be on the navi4x machine. Only testing rock-e2e tests with -j1 for stability.


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -506,6 +536,11 @@ pipeline {
                                         check_RockE2ETests_Navi3x(true)
                                     }
                                 }
+                                if ( "${CODEPATH}" == 'navi4x' ) {
```
pcf000: I think we need an "else if" here, or navi3x could run in its section just above and then run again in the else just below.

stefankoncarevic:
Included Navi4x architecture in Nightly CI pipeline only for Shared Library E2E tests (fixed and random) to run with max threads.

dhernandez0 (approved):

PR #1600, opened 2024-08-07
Move ops inside k-Loop into pipeline stages

manupak:
Currently, we have grid-reversal outside a
stage. It caused a silencable error when
grid-reversal is used.

This commit fixes that by moving it inside the
stage.

closes : https://github.com/ROCm/rocMLIR-internal/issues/1565


krzysz00 (approved):
Looks good to me!

dhernandez0 (approved):

PR #1602, opened 2024-08-07
Update CAPI tests to include C++ tests

fabianmcg:
This patch updates the test config to include C++ test. It also deprecates several tests that fail due to being outdated. For example, `mixr_cobj` fails because it uses tensor types instead of `!migraphx.shaped`.


manupak:
As we discussed.. can you add an archived note inside those .cc file.?

mlir/test/CAPI/CMakeLists.txt
```diff
@@ -21,34 +21,34 @@ target_link_libraries(mlir-mixr-capi-test
   MLIRCAPIRegisterRocMLIR
 )
 
-add_llvm_executable(mlir-tosa-miir-test
-  EXCLUDE_FROM_ALL
-  tosa_miir.cpp
-)
-llvm_update_compile_flags(mlir-tosa-miir-test)
+# add_llvm_executable(mlir-tosa-miir-test
```
manupak, line 25: Also add a note to say these are archived

fabianmcg: Apologies, I missed that comment, I'll add the note.

krzysz00 (approved):

PR #1603, opened 2024-08-07
Move threadwise_copy ops in gridwise_gemm_accel, pipeline non-accel

krzysz00:
1. The threadwise_copy ops were mistakenly placed near the threadwise_read_into ops when the pipelined loop was set up for gridwise_gemm_accel. Fix this by moving them into the right spot in the pipelined loop so that we don't have `shufflevector`s forcing global loads to complete

2. Change the non-accel lowering path tho use the software pipelining scheme so that it matches the accel path, making future fusion optimizations easier.

The attached perf report shows that, for GEMM on gfx90a, this change does somewhere between nothing and a few cases of perf improvement on "good" configs (ones where we're already getting many TFlops) with potential 'losses'/noise on "bad" configs (where we already had 0.07 TFlops). All that could be benchmarking noise, so it's possible that all this change does is make things look better and provide more opportunities to do `sched_group_barrier` stuff in the future.
[improve-pipelining-gemm-perf-gfx90a.html.txt](https://github.com/user-attachments/files/16533952/improve-pipelining-gemm-perf-gfx90a.html.txt)


manupak (approved):
LGTM!

PR #1607, opened 2024-08-09
use removeUpperDims if possible

dhernandez0:
In this PR we use "removeUpperDims" utility function wherever possible to clean up thread layout view creation functions.

There are some cases where it's not possible due to this bug: https://github.com/ROCm/rocMLIR-internal/issues/1562

ticket: https://github.com/ROCm/rocMLIR-internal/issues/1529

### DIFF
--- a/mlir/lib/Dialect/Rock/utility/AccelEmitter.cpp
+++ b/mlir/lib/Dialect/Rock/utility/AccelEmitter.cpp
@@ -383,35 +383,73 @@ llvm::FailureOr<RegsAsMatrixSubTiles> MfmaEmitter::computeOutputTransforms(
   }
 
   {
-    // Create views for tid slice of blockwise sub-tile of C
-    SetVector<StringRef> dimensionsToRemove;
-    dimensionsToRemove.insert("g_block");
-    dimensionsToRemove.insert("m_block");
-    dimensionsToRemove.insert("n_block");
-    dimensionsToRemove.insert("item");
-    FailureOr<ArrayAttr> maybeBlockSubTileTidSlice =
-        removeUpperDims(b, ret.gridSubTile, dimensionsToRemove);
+    // TODO: we can't use "removeUpperDims" because of bug #1562
+    // (https://github.com/ROCm/rocMLIR-internal/issues/1562)
+    // Create views for tid slice of blockwise sub-tile of C    // Create views
+    // for tid slice of blockwise sub-tile of C
+    TopDownTMBuilder splitMemoryCoords(b, {"tid"}, {blockSize}, loc);
+    splitMemoryCoords.merge(
+        {"wave", "m_tid", "n_tid"}, {0, 1, 2}, "tid",
+        {wavesInKernelBlock, waveSize / inputSpanLen, inputSpanLen});
+    TransformMapAttr splitMemoryCoordsAttr = splitMemoryCoords.get();
+    auto toRowsAndCols =
+        TopDownTMBuilder::below(splitMemoryCoords, splitMemoryCoordsAttr);
+    // "blkMajor" and "blkMinor" are placeholder names because we don't know
+    // if they'll be column or row until we check for broadcast-ness.
+    llvm::StringMap<uint32_t> rowsAndColsIdxs =
+        expandNamesInPlace(splitMemoryCoords, {{"wave", {"wave_m", "wave_n"}}});
+    TopDownTMBottomDimsWrapper rowsAndColsWrap(toRowsAndCols, rowsAndColsIdxs);
+    rowsAndColsWrap.merge({"wave_m", "wave_n"}, "wave",
+                          {wavesInKernelBlock / nWaves, nWaves});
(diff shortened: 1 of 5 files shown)


mlir/lib/Dialect/Rock/utility/AccelEmitter.cpp
```diff
@@ -1194,96 +1036,24 @@ RegsAsMatrixSubTiles WmmaEmitter::createAccelGemmOperandTransforms(
...
-    TopDownTMBuilder splitWaveId =
-        TopDownTMBuilder::below(splitTid, splitTidAttr);
-    {
-      splitWaveId.passThrough({"kpack"}, {0}, {"kpack"});
-      splitWaveId.merge({"wave_m", "wave_n"}, {1, 2}, "wave_id",
-                        {mWaves, nWaves});
-      splitWaveId.passThrough({"lane_id", "d_iter", "k_iter"}, {3, 4, 5},
-                              {"lane_id", "d_iter", "k_iter"});
+    SetVector<StringRef> dimensionsToRemove;
+    dimensionsToRemove.insert("k_loop");
```
krzysz00, line 1092: Is there no `append` or other method that takes an initializer list?

krzysz00 (commented):
This all seems reasonable - I'll let Manupa take a look too but nothing jumps out as me as incorrect - or at least the sort of incorrect that'll slip past testing

mlir/lib/Dialect/Rock/utility/AccelEmitter.cpp
dhernandez0: I didn't find any. We could change removeUpperDims to take llvm::StringSet instead of SetVector<StringRef>, it provides a constructor that takes an initializer list and seems the right fit because the inputs are a set of strings.

mlir/test/Dialect/Rock/gridwise_attention_accel_lowering.mlir
```diff
@@ -144,13 +144,19 @@
...
   // Viewing another set of register with kPack packing
   // CHECK: %[[G1AregsKpackTr0:.+]] = rock.transform %[[G1AregsKpack:.+]] by
   // CHECK-DAG: %[[G1AregsKpackTr1:.+]] = rock.transform %[[G1AregsKpackTr0]] by
   // CHECK-DAG: %[[G1AregsKpackTr2:.+]] = rock.transform %[[G1AregsKpackTr1]] by
+  // CHECK-DAG: %[[G1AregsKpackTr3:.+]] = rock.transform %[[G1AregsKpackTr2]] by
+  // CHECK-DAG: %[[G1AregsKpackTr4:.+]] = rock.transform %[[G1AregsKpackTr3]] by
+  // CHECK-DAG: %[[G1AregsKpackTr5:.+]] = rock.transform %[[G1AregsKpackTr4]] by
 
-  // CHECK-DAG: rock.threadwise_copy %[[gemm0NormExpTr2]] -> %[[G1AregsKpackTr2]]
+  // CHECK-DAG: rock.threadwise_copy %[[gemm0NormExpTr5]] -> %[[G1AregsKpackTr5]]
```
manupak: Are these the same views being appended on either side?

manupak:
> There are some cases where it's not possible due to this bug: https://github.com/ROCm/rocMLIR-internal/issues/1562

Can we add comments in the code where this blocks from doing the change ? 
Also in the DoD of the ticket 1562, please link the locations of the code that needs to be changed as part of 1562

mlir/test/Dialect/Rock/gridwise_attention_accel_lowering.mlir
dhernandez0: Using "removeUpperDims" generates more rock.transforms, even though they are just passthroughts of everything. There are three more rock.transform for each side.

dhernandez0:
> > There are some cases where it's not possible due to this bug: [ROCm/rocMLIR-internal#1562](https://github.com/ROCm/rocMLIR-internal/issues/1562)
> 
> Can we add comments in the code where this blocks from doing the change ? Also in the DoD of the ticket 1562, please link the locations of the code that needs to be changed as part of 1562

done

mlir/test/Dialect/Rock/gridwise_attention_accel_lowering.mlir
dhernandez0: sorry, this is not needed as I had to revert some cases where I was using removeUpperDims after running tests on lockhart.

manupak (approved):
LGTM! Thanks for cleaning up the code -- much appreciated.

mlir/lib/Dialect/Rock/utility/AccelEmitter.cpp
krzysz00: Yeah, we might want to go to `StringSet`.

... which is reminding me that we might want to move the transform builders to `StringSaver`s

krzysz00 (approved):
Approved here too - and noting that it's mergable (CodeCov doesn't need to pass)

PR #1608, opened 2024-08-09
Buildbot improvements and fixes.

pcf000:
Couple of things that have popped up and been fixed in the past weeks.

### DIFF
--- a/mlir/utils/buildbot/build-run.sh
+++ b/mlir/utils/buildbot/build-run.sh
@@ -71,4 +71,4 @@ fi
 docker build ${BUILDPROXY} -t "${IMAGE_NAME}:latest" .
 # Use "--group-add root" to work around container putting /dev/kfd and /dev/dri
 # in root instead of the proper video or render.
-docker run -d --net=host --device=/dev/kfd --device=/dev/dri --group-add video --group-add root --name buildbot --restart always ${RUNPROXY} ${VOLUMES} ${ARGS} "${IMAGE_NAME}:latest" ${CMD}
+docker run -d --net=host --device=/dev/kfd --device=/dev/dri --group-add video --group-add render --group-add root --name buildbot --restart always ${RUNPROXY} ${VOLUMES} ${ARGS} "${IMAGE_NAME}:latest" ${CMD}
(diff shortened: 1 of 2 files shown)


mlir/utils/buildbot/build-run.sh
```diff
@@ -71,4 +71,4 @@ fi
 docker build ${BUILDPROXY} -t "${IMAGE_NAME}:latest" .
 # Use "--group-add root" to work around container putting /dev/kfd and /dev/dri
 # in root instead of the proper video or render.
-docker run -d --net=host --device=/dev/kfd --device=/dev/dri --group-add video --group-add root --name buildbot --restart always ${RUNPROXY} ${VOLUMES} ${ARGS} "${IMAGE_NAME}:latest" ${CMD}
+docker run -d --net=host --device=/dev/kfd --device=/dev/dri --group-add video --group-add render --group-add root --name buildbot --restart always ${RUNPROXY} ${VOLUMES} ${ARGS} "${IMAGE_NAME}:latest" ${CMD}
```
pcf000, line 74: Somehow forgot the render group, for months, and got away with it until recently.

mlir/utils/buildbot/mlir-rocm-mi200/Dockerfile
```diff
@@ -17,13 +17,9 @@ FROM rocm/mlir:rocm6.2-latest
...
 
-# LTS releases often bundle obsolete pip versions that cannot access newest
-# Linux binary wheels.
-RUN python3 -m pip install --upgrade pip && python3 -m pip install --ignore-installed PyYAML
-
-# Refer to mlir/lib/Bindings/Python/requirements.txt. Listed explicitly here
-# and version pinned for consistency as this is a bot.
-RUN python3 -m pip install numpy 'pybind11>=2.8.0' dataclasses
+# Accounts for PyYAML, numpy, pybind11, ml_dtypes
+ADD "https://raw.githubusercontent.com/llvm/llvm-project/main/mlir/python/requirements.txt" requirements.txt
```
pcf000, line 21: A new test from upstream requires the python module ml_dtypes, which we don't install anywhere.  Rather than add it to the explicit list, let's use the requirements.txt file that changed to include it.

Grabbing from upstream rather than rocMLIR because the buildbot is sent upstream code to build and we need it to keep up.  Next time they add a module, we just have to kill and restart the buildbot and it'll pick it up (I think).

krzysz00 (approved):

PR #1609, opened 2024-08-13
[DO NOT SQUASH][EXTERNAL] fix plumbing of rocdl attrs: waves_per_eu & unsafe_atomics

manupak:
Currently, the rocdl attrs: waves_per_eu & unsafe_atomics are not plumbed to LLVM.
Not sure how that happened and therefore this results in selecting `cmpswap` instructions
instead of `atomic_fadd` in the backend -- leading to perf degradation in cases where `raw_buffer_*` ops are not used.

closes : https://github.com/ROCm/rocMLIR-internal/issues/1558


krzysz00 (approved):
Approved

How much of this needs to be sent upstream, or is this all already up there and just got lost in a merge?

manupak:
None of those are there in upstream : https://github.com/llvm/llvm-project/blob/54d868991ab79abc5c52cf38939f0a45292e7506/mlir/include/mlir/Dialect/LLVMIR/ROCDLOps.td.

krzysz00:
Then let's fix that after this patch lands

(I do also know there're plans to turn atomic safety from a per-function thing to a per-operation thing, but that's future us)

PR #1610, opened 2024-08-13
Fix tests where no GPU is present

krzysz00:
When I built rocMLIR on a system that had its GPU
out (rocm_agent_enumerator just reports a gfx000) various tests didn't run because %arch was the empty string.

Also, the get_agents() function wasn't properly updated for the utf-8 decodes that are now done on subprocess output, so gfx000 wasn't necessarily being filtered out properly.

### DIFF
--- a/mlir/test/fusion/pr-e2e/reductions/atomic_add/lit.local.cfg
+++ b/mlir/test/fusion/pr-e2e/reductions/atomic_add/lit.local.cfg
@@ -1,2 +1,2 @@
-if not 'atomic_add' in config.features:
+if not config.enable_rock_driver_pr_e2e_test or config.no_AMD_GPU or not 'atomic_add' in config.features:
   config.unsupported = True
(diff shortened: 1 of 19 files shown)


mlir/test/fusion/pr-e2e/reductions/atomic_add/lit.local.cfg
```diff
@@ -1,2 +1,2 @@
-if not 'atomic_add' in config.features:
+if not config.enable_rock_driver_pr_e2e_test or config.no_AMD_GPU or not 'atomic_add' in config.features:
```
manupak, line 1: If there is no gpu, shouldn;t we have an empty config.features ?

manupak (commented):
Thanks for the cleanup.

I have a question w.r.t e2e tests.

mlir/test/fusion/pr-e2e/reductions/atomic_add/lit.local.cfg
krzysz00: The problem is is that if there's no GPU, then `config.features` is `None`, which means `'atomic_add' in config.features` is `'atomic_add` in None` which throws an exception and prevents any tests from running at all

Maybe the fix is changing how `config.features` is defined, but a `no_AMD_GPU` check is more in line with the rest of the test suite

manupak: I see ok. I thought no_AMD_GPU is thing introduced in this PR; it was there...

manupak (approved):
LGTM

PR #1612, opened 2024-08-14
[DO NOT SQUASH] Fp8 support for gfx12

giuseros:
This PR adds support for fp8 in rocMLIR. As per usual, I split the PR in two:
- Changes to external/
- Changes to our internal compiler
 
Given the status of our CI, I was not able to run a full set of tests, but I tried a couple of examples, and it looked like it was working. We are trying to fix this, but can I have a first round of reviews to see if those changes are reasonable?

### DIFF
--- a/external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
+++ b/external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
@@ -393,6 +393,7 @@ static void wmmaPushInputOperand(ConversionPatternRewriter &rewriter,
                                  Location loc,
                                  const TypeConverter *typeConverter,
                                  bool isUnsigned, Value llvmInput,
+                                 Value mlirInput,
                                  SmallVector<Value, 4> &operands) {
   Type inputType = llvmInput.getType();
   auto vectorType = dyn_cast<VectorType>(inputType);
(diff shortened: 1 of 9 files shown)


mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -519,12 +518,17 @@ GemmSize GemmSize::fromConvolution(ConvOpType type,
 }
 
 static LogicalResult verifyGemmTypes(Operation *op, GemmFeatures features,
-                                     Type elemTypeA, Type elemTypeB,
-                                     Type elemTypeC) {
+                                     StringRef arch, Type elemTypeA,
+                                     Type elemTypeB, Type elemTypeC) {
+  bool isGfx12 = arch.contains("gfx12");
```
krzysz00: Let's make this a test for gfx11 instead? Since future chips should have fp8 support

mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
```diff
@@ -84,6 +84,10 @@ FailureOr<WmmaInsn> WmmaInsn::select(mlir::Type elementTypeA,
     insn = ROCDL::wmma_f32_16x16x16_bf16::getOperationName();
   } else if (elementTypeA.isInteger(8)) {
     insn = ROCDL::wmma_i32_16x16x16_iu8::getOperationName();
+  } else if (elementTypeA.isFloat8E4M3FN()) {
+    // insn = ROCDL::wmma_
```
krzysz00: Comments?

krzysz00 (commented):
Overall, seems fine

Please leave a big shouty comment on the gfx12 info saying that this isn't accurate and needs to be adjusted after hardware release

mlir/lib/Dialect/Rock/IR/WmmaInsnGroup.cpp
giuseros: Sorry, somehow this was overridden. Should be fine now

giuseros:
> Please leave a big shouty comment on the gfx12 info saying that this isn't accurate and needs to be adjusted after hardware release

I am not sure I understand where the bit shouty comment should go :)

krzysz00:
Sorry, in AmdArchDb on our side - it could do with a reminded that, say, numCUs isn't right

krzysz00 (approved):

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -401,28 +402,30 @@ static void wmmaPushInputOperand(ConversionPatternRewriter &rewriter,
...
 
+  auto mlirInputType = dyn_cast<VectorType>(mlirInput.getType());
+  if (mlirInputType.getElementType().isInteger(8)) {
+    // if element type is 8-bit signed or unsigned, ignore the isUnsigned flag
+    bool localIsUnsigned = isUnsigned;
+    if (elemType.isUnsignedInteger(8)) {
+      localIsUnsigned = true;
+    } else if (elemType.isSignedInteger(8)) {
+      localIsUnsigned = false;
+    }
```
umangyadav, line 418:
```suggestion
    localIsUnsigned = elemType.isUnsignedInteger(8)
```

Can this work ?

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -401,28 +402,30 @@ static void wmmaPushInputOperand(ConversionPatternRewriter &rewriter,
...
+    Value sign = createI1Constant(rewriter, loc, !localIsUnsigned);
+    operands.push_back(sign);
```
umangyadav, line 420: If you are flipping sign with `!localIsUnsigned` then perhaps better to make it `localIsigned`

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
giuseros: No, because it can be signedless, and in that case it would take the value of `isUnsigned`. Your logic would invalidate that.

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
giuseros: So the point is that the wmma Operation has a `getUnsigned()` method. So at some point we need to to that flip to reconcile LLVM vs MLIR operations.

umangyadav (approved):

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
krzysz00: More generally, the WMMA op has been rigged to work whether you specify the signedness on the type or on the operation (the latter is more preferred, but I left the type-level version in case people want to go that route)

PR #1615, opened 2024-08-16
[Attention] Fixed preSoftmaxElementwiseRegion input ordering

manupak:
Currently, there were two (wrong) assumptions:
1) elementwise ops at most have two inputs
2) preSoftmax elementwise regions have gemm input as the first

The above get violated when migraphx provides a ternary
elementwise ops (such as `where`) and it also had the
case where gemm input is in the middle.

Therefore, this commit fixes all that.
closes: https://github.com/ROCm/rocMLIR-internal/issues/1575


mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -221,7 +221,8 @@ def Rock_AttentionOp :
     Rock_GemmFeaturesAttr:$features,
     OptionalAttr<I32Attr>:$numCU,
     OptionalAttr<RockTuningParamAttrInterface>:$params0,
-    OptionalAttr<RockTuningParamAttrInterface>:$params1
+    OptionalAttr<RockTuningParamAttrInterface>:$params1,
+    I32Attr:$firstGemmIdx
```
krzysz00, line 225: `IndexAttr` maybe?

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -893,17 +895,25 @@ struct AttentionRewritePattern : public OpRewritePattern<tosa::MatMulOp> {
...
                               Block *block, SmallVector<Value> &elemwiseArgs,
                               std::optional<Location> loc = std::nullopt,
-                              bool doRewrite = false) const {
+                              bool doRewrite = false, int recDepth = 0) const {
     PatternRewriter::InsertionGuard guard(regionBuilder);
     regionBuilder.setInsertionPointToEnd(block);
     // If the matmul is found, we return this information to the
     // root.
+    LLVM_DEBUG(llvm::dbgs()
+               << std::string(recDepth, '\t')
```
krzysz00, line 904: Why `recDepth`

krzysz00 (approved):
Minor nitpicks, seems fine

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
manupak: gets a nice log of how recursion went with `-debug-only`

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: I ll sneak it in a next PR or do a follow up. I think same could be said for numCU as well.

krzysz00:
... in post-PR realizations, that `std::string` call for the tabs is `os.indent()` and `os.unindent()`, I think

PR #1619, opened 2024-08-20
[DO NOT SQUASH][EXTERNAL] Add a scheduling barrier guard around inlineAsm lds.barrier

manupak:
This commit adds a scheduling regions around the inlineAsm
to guard against possible complications arising from them
interfering with the backend scheduler / register allocation.

closes : https://github.com/ROCm/rocMLIR-internal/issues/1579


krzysz00 (approved):
Please update the `lds_barrier` tests so we can upstream this. Also, does this cause issues for gfx1100?

manupak:
> Please update the lds_barrier tests so we can upstream this. Also, does this cause issues for gfx1100?

It doesn't for the case @pcf000 pointed out.

Im not sure whether we want to upstream this. 

I d be more favor of removing this if we dont care about performance on gfx908 & gfx1100. Following the blame, I understood this was done to remove vmcnt on waitcnt that would otherwise be inserted by the backend compiler. More reasoning here : https://ontrack-internal.amd.com/browse/SWDEV-302242

While I think putting this into basicblock by its own will be less prone to these kind of errors but we should not need to use inlineAsm in the first place.

manupak:
Oh this has been upstreamed... Im realizing that now.. 
Why did we do that ? I feel this as a tradeoff between performance vs error "proneness" for gfx908 & gfx1100.

ok Im conflicted now... it feels like a tactical decision rather than a technical one.  @jerryyin WDYT?

PR #1621, opened 2024-08-21
Fix crash arising from insufficient guards in WMMA instruction selector

krzysz00:
When serecting WMMA instructions, we get the number of repeats by dividing by dPerAccel. We used to guard this against division by 0 by checking that mPerWave and nPerWave were divisible by inputVectorLen, which was equal to dPerAccel ... until gfx12, where dPerAccel becase double the length of inputVectorLen (because WMMA changed to re-use both halves of the wave). This caused 0 coefficients in transform maps, which led to a bunch of weird crashes.

Fixes https://github.com/ROCm/rocMLIR-internal/issues/1582 (by making the input problem invalid) but doesn't actually fix the assertion crash we saw in that ticket because I got a different crash on develop.

### DIFF
--- a/mlir/lib/Dialect/Rock/utility/AccelEmitter.cpp
+++ b/mlir/lib/Dialect/Rock/utility/AccelEmitter.cpp
@@ -843,7 +843,7 @@ AccelEmitterParams WmmaEmitter::initAccelEmitterParams(
   // and we want to do a(16x16) * b(16x16), 16 threads are loading a vector
   // of 16 Ks and the other 16 threads are replicating those values.
   int64_t waveSize = rock::lookupArchInfo(arch).waveSize;
-  if (arch.contains("gfx12")) {
+  if (!isGfx11) {
     // Post-gfx12 each thread is loading a partial set of values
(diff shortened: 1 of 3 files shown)


manupak (approved):

mlir/lib/Dialect/Rock/utility/AccelEmitter.cpp
```diff
@@ -843,7 +843,7 @@ AccelEmitterParams WmmaEmitter::initAccelEmitterParams(
   // and we want to do a(16x16) * b(16x16), 16 threads are loading a vector
   // of 16 Ks and the other 16 threads are replicating those values.
   int64_t waveSize = rock::lookupArchInfo(arch).waveSize;
-  if (arch.contains("gfx12")) {
+  if (!isGfx11) {
```
umangyadav, line 846: Gfx10 would also fall in branch. is `gfx10` supported ?

krzysz00: It doesn't have WMMA

PR #1622, opened 2024-08-22
Collected Jenkinsfile tweaks for reliability.

pcf000:
1.  Default disableNavi21 to true.  Retain the code just in case.
2.  Use gfx1101 instead of gfx1100, because they don't hit GPU issues as often.
3.  Add disable parameters for gfx908, gfx90a, gfx942.
4.  Check out MITuna with git verb instead of sh, avoids lockhart proxy issues.
5.  Remove "&& multi_gpu" condition for gfx908, fewer gfx908s now.

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -185,7 +185,7 @@ String getLabelFromCodepath(String codepath) {
     } else if (codepath == "vanilla"){
         label = 'mlir'
     } else if (codepath == "navi3x") {
-        label = 'mlir && ( gfx1100 || gfx1101 )'
+        label = 'mlir && gfx1101'
     } else {
         echo "${codepath} is not supported"
         label = 'wrongLabel'
@@ -199,7 +199,7 @@ String getLabelFromChip(String chip) {
         case "gfx906":
             return getLabelFromCodepath("vanilla")
         case "gfx908":
-            return "mlir && gfx908 && multi_gpu"
+            return "mlir && gfx908"
         case "gfx90a":
             return "mlir && gfx90a"
         case "gfx942":
@@ -341,10 +341,12 @@ boolean shouldRunFromChip(String chip) {
             return shouldRunFromCodepath("vanilla")
         case "gfx90a":
             // Special case because all our "vanilla" hosts are gfx90a.
-            return shouldRunFromCodepath("mfma") || shouldRunFromCodepath("vanilla")
+            return params.disable90a == false &&
+                   (shouldRunFromCodepath("mfma") || shouldRunFromCodepath("vanilla"))
         case "gfx908":
+            return params.disable908 == false && shouldRunFromCodepath("mfma")
         case "gfx942":
-            return shouldRunFromCodepath("mfma")
+            return params.disable942 == false && shouldRunFromCodepath("mfma")
         case "gfx1030":
             return shouldRunFromCodepath("navi21")
         case "gfx1100":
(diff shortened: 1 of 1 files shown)


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -185,7 +185,7 @@ String getLabelFromCodepath(String codepath) {
     } else if (codepath == "vanilla"){
         label = 'mlir'
     } else if (codepath == "navi3x") {
-        label = 'mlir && ( gfx1100 || gfx1101 )'
```
pcf000, line 188: First controversial part.  In my recent experience, gfx1101 has been less likely to have problems than gfx1100.  Let's experiment with using it.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -199,7 +199,7 @@ String getLabelFromChip(String chip) {
         case "gfx906":
             return getLabelFromCodepath("vanilla")
         case "gfx908":
-            return "mlir && gfx908 && multi_gpu"
```
pcf000, line 202: We have fewer gfx908, let's not limit ourselves if it can help.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -341,10 +341,12 @@ boolean shouldRunFromChip(String chip) {
             return shouldRunFromCodepath("vanilla")
         case "gfx90a":
             // Special case because all our "vanilla" hosts are gfx90a.
-            return shouldRunFromCodepath("mfma") || shouldRunFromCodepath("vanilla")
+            return params.disable90a == false &&
```
pcf000, line 344: Added disable switches for the other architectures, if we want to focus on a single one for debugging.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -427,12 +429,18 @@ pipeline {
                choices: ['default', 'mfma', 'navi21', 'navi3x', 'vanilla'],
                description: 'Choose the codepath to test')
         // option to disable navi21 cells in case nodes are offline
-        booleanParam(name: 'disableNavi21', defaultValue: false,
```
pcf000, line 430: Second controversial part:  cut out navi21 by disabling it rather than removing the code, in case we have some navi21-specific bug and need to verify it.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -544,9 +552,11 @@ pipeline {
                         }
                         steps {
                             buildProject('ci-performance-scripts', '')
+                            // How to check out into specific directory, according to stackoverflow.
```
pcf000, line 555: Check out MITuna this way rather than as a shell command.  I've been seeing lockharts fail with apparent proxy problems, and I think this'll avoid them.  Also it's arguably cleaner.

krzysz00 (approved):
Are the perf runs still gfx1100?

But also ... sure, approved

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -802,7 +813,7 @@ pipeline {
                 axes {
                     axis {
                         name 'CHIP'
-                        values 'gfx908', 'gfx90a', 'gfx1030', 'gfx1100'
```
pcf000, line 805: More gfx1101.  If we tracked performance over time, we might see a discontinuity with this change.

pcf000:
> Are the perf runs still gfx1100?

No, also gfx1101.

PR #1625, opened 2024-08-23
Use separate call to check for `gfx11`

umangyadav:
Earlier it was using `isGfx11` flag which gets initialized after call to base class constructor `AccelEmitter`

Therefore it can't be used inside the constructor of the base class. 

Fixes GemmVariants tests on Navi3x


krzysz00 (approved):
Let's add a comment pointing out the weirdness here.

Approved, thanks for the fast debugging!

umangyadav:
> Let's add a comment pointing out the weirdness here.

Done

pcf000:
Appears to fix my sanity_wmma.mlir problem, too.

PR #1629, opened 2024-08-26
Handle non-zero-preserving input fusions, make read_into track validity

krzysz00:
Apologies for the mega-commit. (this one also includes a cherry-pick from upstream)

Previously, if you input-fused an elementwise funciton like
(x) => (x + 4), this would cause issues, because the 4 would be added
both to elements of the input value but also to padding values
introduces during tiling, which would cause incorrect results.

To fix this, we:
1. Make threadwise_read_into optionally return a boolean vector of
validities, recording whether or not an element in the register tile
actually came from memory or not
2. Give threadwise_read_into a dynamic validities argument, where the
validity data from the coordinate transforms is ANDed with the
validity vectors from those arguments to allow applying validitiies
that don't come from coordinate transforms.
3. Update all the threadwise_read_into operations that can have input
fusions on them to generate a validity record.
4. It a linalg.generic is being tiled by a threadwise_read_into and
that read records validity (that is, it came from an input
fusion) *and* the result of the elementwise function on all-zero
inputs isn't zero, we propagate the validities.

That is, we make the threadwise_read_into ops that read each input to
the generic record their validities, and then do a register ->
register threadwise_read_into (which would ordinarily be a memcpy())
with dynamic validities taken from those reads to re-apply the 0 mask
after the generic runs.

**Note:** While I was here, I noticed that non-xdlops gemms were producing two sets of theradwise_read_into ops, so I fixed that, which broke a few tests

Also, fix up the validator on `rock.alloc` so it doesn't reject int4 buffers ( @dhernandez0 )

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -856,7 +856,7 @@ def Rock_IndexDiffUpdateOp :
   let hasVerifier = 1;
 }
 
-defvar SupportedMemoryElems = [F32, F16, BF16, I<4>, I8, I16, I32,
+defvar SupportedMemoryElems = [F32, F16, BF16, I1, I<4>, I8, I16, I32,
                                F8E5M2FNUZ, F8E4M3FNUZ, F8E5M2, F8E4M3FN];
 defvar NativeMemoryOpTypes = [F32, F16, BF16, I<4>, I8, I16, I32,
                            F8E5M2FNUZ, F8E4M3FNUZ, F8E5M2, F8E4M3FN,
@@ -987,19 +987,34 @@ def Rock_GlobalStoreOp :
   let hasVerifier = 1;
 }
 
+defvar SameShapeVectorOfI1 = [{
+  ::mlir::VectorType::get(::llvm::cast<::mlir::ShapedType>($_self).getShape(), ::mlir::IntegerType::get($_ctxt, 1))
+}];
 // threadwise_read_into
 def Rock_ThreadwiseReadIntoOp :
     Rock_Op<"threadwise_read_into",
     [DeclareOpInterfaceMethods<RockAcceptingViewOpInterface>,
-     DeclareOpInterfaceMethods<RockWriterOpInterface>]>,
-    AllElementTypesMatch<["source", "dest"]>,
+     DeclareOpInterfaceMethods<RockWriterOpInterface>, AttrSizedOperandSegments,
+     // This wasn't being enforced and broke buffers of vectors.
+     //AllElementTypesMatch<["source", "dest"]>,
+     OptionalTypesMatchWith<"validity record is a vector of one boolean per dest entry",
+      "dest", "validityRecord", SameShapeVectorOfI1>,
+     TypesMatchWith<"all dynamic validities are a vector of one boolean per dest entry",
+      "dest", "dynamicValidities", SameShapeVectorOfI1,
+      [{
+        [](::mlir::Type lhs, ::mlir::TypeRange rhs) -> bool {
+          return llvm::all_of(rhs, [lhs](::mlir::Type t) { return t == lhs; });
+        }
+      }]>]>,
     Arguments<(ins Arg<MemRefOf<NativeMemoryOpTypes>, "source view", [MemRead]>:$source,
       Arg<MemRefOf<NativeMemoryOpTypes>, "destination registers", [MemWrite]>:$dest,
+      Variadic<VectorOf<[I1]>>:$dynamicValidities,
       TransformMapArrayAttr:$extraViews,
       Variadic<Index>:$extraIndices,
       UnitAttr:$forceUnroll,
       UnitAttr:$useIndexDiffs
-    )> {
+    )>,
+    Results<(outs Optional<VectorOf<[I1]>>:$validityRecord)> {
   let summary = "Read values from transformed source into destination";
 
   let description = [{
@@ -1040,13 +1055,37 @@ def Rock_ThreadwiseReadIntoOp :
     If extraIndices are used, the [extraViews]source must have the form
     (extraIdx0, ... , extraIdxN, iteration_number).
 
-    This is used during fusion to represent loads from addition arguments like
-    bias tensors.
+    If any `dynamicValidities` are passed in, they must have the same shape
+    as the output registers, and the booleans in each of these validities will be
+    ANDed together with the validity implied by the coordinate transformations
+    to determine whether the read is to actually be performed.
+
+    Note that using dynamic validities forces the vectorization width to 1.
+
+    If `validityRecord` is specified, it will be filled with a boolean indicating
+    whether each of the values in `dest` was actually read from memory or if it
+    fell into some padded or otherwise invalid (ex. a coordinate went negative)
+    region. This is used to allow manually applying re-applying padding after
+    an input fusion that could cause the padded region to be non-zero.
+
+    This operation is also used during fusion to represent loads from additional
+    arguments like bias tensors.
   }];
+
+  let builders = [
+    // Alias for the common case where we aren't filling a buffer of validity checks
+    OpBuilder<(ins "Value":$source, "Value":$dest,
+                   "ArrayAttr":$extraViews, "ValueRange":$extraIndices,
+                   "bool":$forceUnroll, "bool":$useIndexDiffs),
+      [{
+        build($_builder, $_state, TypeRange{}, source, dest, ValueRange{}, extraViews, extraIndices, forceUnroll, useIndexDiffs);
+      }]>
+  ];
 
   let assemblyFormat = [{
     attr-dict $extraViews `(` $source `)` (`[` $extraIndices^ `]`)? `->` $dest
-    `:` type($source) `->` type($dest)
+    (`if` ` ` `[` $dynamicValidities^ `]`)?
+    `:` type($source) `->` type($dest) (`,` type($validityRecord)^)?
   }];
   let hasVerifier = 1;
 }
--- a/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
+++ b/mlir/include/mlir/Dialect/Rock/IR/RockOps.td
@@ -1060,7 +1060,10 @@ def Rock_ThreadwiseReadIntoOp :
     ANDed together with the validity implied by the coordinate transformations
     to determine whether the read is to actually be performed.
 
-    Note that using dynamic validities forces the vectorization width to 1.
+    Note that using dynamic validities forces the vectorization width to 1 (but
+    that merely requesting a validity vector does not). There isn't any inherent
+    reason for this limitation other than that support for post-vectorization
+    masking wasn't needed in the initial usecase for this feature.
 
     If `validityRecord` is specified, it will be filled with a boolean indicating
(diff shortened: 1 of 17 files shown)


mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -772,29 +798,100 @@ static void reconfigureLAGeneric(LinalgAlignRewriter &b,
...
+  return success(foldedToZero);
+}
+
+/// If this generic doesn't preserve zero (ex, it's x => x + 1) and if the
+/// validity of the tiling operation was being tracked (this indicates input
+/// fusion), then:
+/// - Clone the output tile
+/// - Set up a register->register threadwise_read_into between this cloned tile
+///   and the original output tile, with dynamic validities drawn from the
+///   validity validity results of each read.
```
dhernandez0: "validity" is written twice.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -772,29 +798,100 @@ static void reconfigureLAGeneric(LinalgAlignRewriter &b,
...
+  // using that record, we don't need to propagate the validity records from the
+  // inputs.
+  if (!oldTwRead.getValidityRecord() ||
+      oldTwRead.getValidityRecord().use_empty()) {
+    if (validityRecords.empty())
+      return Value{};
+    if (succeeded(knownToPreserveZero(reconfiguredGeneric, b)))
+      return Value{};
+  }
+  Value originalTile = reconfiguredGeneric.getOutputs()[0];
```
dhernandez0, line 885: can this have more than one output?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -915,8 +908,50 @@ struct GridwiseAttentionAccelRewritePattern
     return ldsByteBuffer;
   }
 
-  // This function will create fromGlobalRegsBuffer, toLDSRegBuffer and
-  // ldsTileBuffer for a gemm input
+  std::tuple<SmallVector<Value>, int64_t>
+  createSharedLDSByteBufferRefs(PatternRewriter &rewriter, Location loc,
```
dhernandez0: I think this function is not used. This was removed with the "reuse lds" pass.

mlir/lib/Dialect/Rock/Transforms/ThreadwiseGemmLowering.cpp
```diff
@@ -525,6 +525,13 @@ LogicalResult ThreadwiseReadIntoRewritePattern::matchAndRewrite(
     loadType = vectorTypeOrSelf(elementType, vectorSrcLen);
   }
 
+  // Force the dynamic validity case down to a vectorizaiton of 1
```
dhernandez0: vectorizaiton -> vectorization

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -875,15 +972,25 @@ LAGenericRewritePattern::matchAndRewrite(linalg::GenericOp laGeneric,
...
     if (newOutput.getDefiningOp()->getBlock() == laGeneric->getBlock())
       b.moveBeforeIfNeeded(newOutput.getDefiningOp(), laGeneric);
     reconfigureLAGeneric(b, laGeneric, newInputs, newOutput);
-    b.eraseOp(tileReadOp);
+    Value newValidityRecord =
+        reapplyPaddingIfNeeded(laGeneric, newValidityRecords, tileReadOp, b);
+    if (!newValidityRecord)
+      b.eraseOp(tileReadOp);
+    else
+      b.replaceOp(tileReadOp, newValidityRecord);
```
dhernandez0: shouldn't this be the ThreadwiseReadIntoOp instead of its validityRecord?

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -1040,13 +1055,37 @@ def Rock_ThreadwiseReadIntoOp :
...
     (extraIdx0, ... , extraIdxN, iteration_number).
 
-    This is used during fusion to represent loads from addition arguments like
-    bias tensors.
+    If any `dynamicValidities` are passed in, they must have the same shape
+    as the output registers, and the booleans in each of these validities will be
+    ANDed together with the validity implied by the coordinate transformations
+    to determine whether the read is to actually be performed.
+
+    Note that using dynamic validities forces the vectorization width to 1.
```
manupak: Why ? do we that for reading padded regions ?

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -1661,6 +1661,15 @@ LogicalResult ThreadwiseReadIntoOp::verify() {
           "Vector buffers vector's lengths need to be evenly divisible");
   }
 
+  if (!getDynamicValidities().empty() &&
+      srcType.getElementType() != destType.getElementType()) {
+    return emitOpError("dynamic validities are incompatible with making the "
```
manupak: Why this ? 
In my mind, vectorization should be unaffected. 
You just need to code zero out the (partial) vectors based on validity

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
manupak: sounds like noise from a merge

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
krzysz00: In full generality, yes, in the context of the fusion code, no

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
krzysz00: Nah, it's replacing an op by a set of `Value`s

external/llvm-project/mlir/lib/Dialect/Math/Transforms/LegalizeToF32.cpp
```diff
@@ -65,10 +65,11 @@ void mlir::math::populateLegalizeToF32TypeConverter(
 
 void mlir::math::populateLegalizeToF32ConversionTarget(
     ConversionTarget &target, TypeConverter &typeConverter) {
-  target.addDynamicallyLegalDialect<MathDialect>(
-      [&typeConverter](Operation *op) -> bool {
-        return typeConverter.isLegal(op);
-      });
+  target.markUnknownOpDynamicallyLegal([&typeConverter](Operation *op) -> bool {
```
manupak: a weird change... but I take it its not yours ?

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: as in we still read vectors from padded regions before this change.. so after this change will we switch back to scalars if there is padding ?

mlir/include/mlir/Dialect/Rock/utility/builderUtils.h
```diff
@@ -39,6 +39,10 @@ Type getFlattenedType(Type type);
 Value getAsTensor(OpBuilder &builder, Location loc, mlir::Value value,
                   bool isWritable = false);
 
+// Return the type of a boolean vector whose shape is the same as the shape of
+// `v` except that its elements are booleans.
+Type validityVectorShapedLike(Value v);
```
manupak: maybe a better name : "getBoolVectorOfShape(Value v)" ? 
Point being .. we could be using this elsewhere its not "validity" in future.
(not a strong comment)

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -772,29 +798,100 @@ static void reconfigureLAGeneric(LinalgAlignRewriter &b,
...
+  // using that record, we don't need to propagate the validity records from the
+  // inputs.
+  if (!oldTwRead.getValidityRecord() ||
+      oldTwRead.getValidityRecord().use_empty()) {
+    if (validityRecords.empty())
+      return Value{};
+    if (succeeded(knownToPreserveZero(reconfiguredGeneric, b)))
+      return Value{};
+  }
+  Value originalTile = reconfiguredGeneric.getOutputs()[0];
```
manupak, line 885: Can we error out explicitly, please ?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -772,29 +798,100 @@ static void reconfigureLAGeneric(LinalgAlignRewriter &b,
...
+                                    LinalgAlignRewriter &b) {
+  // If the old read isn't recording its validity anywhere, or if we're not
+  // using that record, we don't need to propagate the validity records from the
+  // inputs.
+  if (!oldTwRead.getValidityRecord() ||
+      oldTwRead.getValidityRecord().use_empty()) {
+    if (validityRecords.empty())
+      return Value{};
+    if (succeeded(knownToPreserveZero(reconfiguredGeneric, b)))
+      return Value{};
```
manupak, line 880: maybe lets do FailureOr<Value> as the return type.
It d play nicely with emitting errors below...

manupak (commented):
I did a pass.

I only have a single concern.. that is that we drop vectorization (to 1) if there is 0 != f(0) input fusion. The fact that you need to explicitly force it to be 1 in the presense of validities makes me think it worked with "padding".

So the question is why cant we `and` read vector with a mask in your second `threadwise_read_into` ?

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
krzysz00: The thing is, the drop to 1 happens in the register->register copy

external/llvm-project/mlir/lib/Dialect/Math/Transforms/LegalizeToF32.cpp
krzysz00: Yeah, needed pulled in to fix a crash (which I also hit)

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
krzysz00: Error out when?

manupak: when #outputs > 1

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
manupak: ah okay .. if so can we explicitly assert that? i.e. the drop to 1 does not happen elsewhere looking at src & dest memory spaces.

krzysz00: Well, no, we need the drop to 1 if we were doing KV cache stuff

manupak (approved):
Thanks! LGTM

PR #1630, opened 2024-08-26
Fix test for input vectorization traversal, use types correctly, add …

krzysz00:
…tests

1. The test for input vectorization traversal was backwards
2. The majorTensorNumber was being set incorrectly on single-arugment linalg.generic operations
3. Now, the vectorizer looks at the ultimate type of the buffer and not the type of the intermeditae allocation when deciding the maximum possible vectorization.

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
@@ -153,7 +153,7 @@ static void annotateGenericOp(linalg::GenericOp lgop) {
   size_t argIdx = -1;
   if (lgop.getInputs().size() == 1) {
     lgop->setAttr("rock.majorTensorNumber",
-                  IntegerAttr::get(IndexType::get(ctx), 1));
+                  IntegerAttr::get(IndexType::get(ctx), 0));
     return;
   }
   for (auto [inputIdx, inp] : llvm::enumerate(lgop.getInputs())) {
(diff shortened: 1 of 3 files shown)


mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
```diff
@@ -153,7 +153,7 @@ static void annotateGenericOp(linalg::GenericOp lgop) {
   size_t argIdx = -1;
   if (lgop.getInputs().size() == 1) {
```
dhernandez0, line 158: does this function work if lgop.getInputs().size() == 0? (not sure if that can happen)

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -960,7 +962,7 @@ VectorizationResult mlir::rock::getMaxVectorization(
   // warp will issue, if possible, a global_load_dwordx4 instruction
   if (!ignoreDataType) {
     constexpr int64_t maxVectorLenBits = 128;
-    int64_t bwidth = upperType.getElementTypeBitWidth();
+    int64_t bwidth = outputType.getElementTypeBitWidth();
```
dhernandez0, line 965: I think this fixes the issue where getMaxVec returns the final type? (for example fp16 insetad of fp32)? Nice!

manupak (approved):
LGTM!

mlir/lib/Dialect/Rock/Transforms/Regularize.cpp
krzysz00: Point, I should check on that

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
krzysz00: Yep, went and fixed that here - so maybe we can move the output swizzle creation before align-tiling!

dhernandez0 (approved):

PR #1631, opened 2024-08-27
Fix removeUpperDims

manupak:
Currently the implementation of removeUpperDims is too simple and not useful handle the layouts we have.
It primarily had two defiencies/bugs:

1) It assumed dimensions will not be collapsed and expanded.

Say we if we had : `A1 x A2 x A3` --> `B` --> `C1 X C2 X C3`.
Then we if we remove A2 for e.g., then Cs should be scaled down accordingly.
This is handled in the PR by introducing the notion of sub-dimensions that is removed that carried over transform stack.
The sub-dimensions will always be recorded relative to the main dimension that will get merged/unmerged/passthrough -- thus, they will also undergo the same transformations.

2) The dimensions remapping only looked at a single transformation in trOps.

Say if we had many passthroughs and some of the dimensions on those are removed.
The remapping were done locally within a single trOp. These led to "holes" when we
are supposed to create a "dense" dim indices.

The above are fixed in this PR.
closes : https://github.com/ROCm/rocMLIR-internal/issues/1562


mlir/lib/Dialect/Rock/utility/AccelEmitter.cpp
```diff
@@ -383,73 +383,35 @@ llvm::FailureOr<RegsAsMatrixSubTiles> MfmaEmitter::computeOutputTransforms(
...
-                          {wavesInKernelBlock / nWaves, nWaves});
-    rowsAndColsWrap.passThrough({"m_tid", "n_tid"});
-    TransformMapAttr toRowsAndColsAttr = toRowsAndCols.get();
-    auto toMatrixC = TopDownTMBuilder::below(toRowsAndCols, toRowsAndColsAttr);
-    toMatrixC.unmerge("gemmM", 0, {waveM.name, mTid.name},
-                      {waveM.size, mTid.size});
-    toMatrixC.unmerge("gemmN", 1, {waveN.name, nTid.name},
-                      {waveN.size, nTid.size});
+    // Create views for tid slice of blockwise sub-tile of C
+    SetVector<StringRef> dimensionsToRemove;
```
dhernandez0: suggestion: we could change SetVector<StringRef> to StringSet. It has an std::initializer_list constructor, so the code would be shorter.

dhernandez0 (approved):

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -2071,13 +2241,14 @@ mlir::rock::removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
     llvm::SmallVector<int64_t> lowerBounds = {};
     FailureOr<rock::TransformMapAttr> maybeNewTrMapAttr =
         removeUpperDimsFromMap(b, trMap, removeIndicesSet, upperBounds,
-                               lowerBounds);
+                               lowerBounds, preservedSubDims);
```
dhernandez0, line 2244: it looks like "preservedSubDims" is not used here, can't it be an internal variable of removeUpperDimsFromMap?

manupak: its used between calls same as "removeIndicesSet".

mlir/lib/Dialect/Rock/utility/AccelEmitter.cpp
manupak: done now.

PR #1632, opened 2024-08-28
Report error when a library fails to load

dhernandez0:
Small PR to report an error when a library fails to load. This would have saved me some time :D


manupak (approved):

krzysz00 (approved):
Approved

Can you check the upstream version of this fire and see if that change needs to be made there too?

dhernandez0:
> Approved
> 
> Can you check the upstream version of this fire and see if that change needs to be made there too?

do you mean llvm/mlir/lib/Target/LLVM/ROCDL/Target.cpp?
I think that one looks good.

umangyadav (approved):

krzysz00:
@dhernandez0 Yeah, that's the file I had in mind

PR #1633, opened 2024-08-28
Fix verifyGemmTypes for fp8

dhernandez0:


manupak (approved):
LGTM! but I d double check with @krzysz00

umangyadav (approved):

krzysz00 (approved):
LGTM - though, IIRC, the WMMA operations don't support miked fp8-bf8, so we should check for that?

dhernandez0:
> LGTM - though, IIRC, the WMMA operations don't support miked fp8-bf8, so we should check for that?

do you mean checking if elemTypeA == elemTypeB? Is that something that should hold for all cases or just fp8-bf8?

PR #1634, opened 2024-08-29
Handle the new F8 types in RockTuningImpl.cpp too.  Oops.

pcf000:
Handle the new F8 types in RockTuningImpl.cpp too.
Add an F8-types emit-tuning-key test to an existing test.
Canonicalise F8 types from fp8/bf8 to full type names early in rocmlir-gen.
Restrict the F8 type canonicalisation to runs when --arch is specified.

Aside from  forgetting to handle the OCP types in the tuning code, it turns out that both rocmlir-gen.cpp and ConvGenerator.cpp have code to convert "fp8" and "bf8" to types.  We should merge that code at some point, but the rocmlir-gen one has access to architecture and the --force-f8-types flag and the ConvGenerator one doesn't, and it runs second.

Rather than try the merging, I made rocmlir-gen.cpp convert the type strings "fp8" and "bf8" to the strings naming the correct type for the chip and flag, eg, "f8M4E3FN".  ConvGenerator then won't have to figure it out.

One minor quirk is that it's possible to run rocmlir-gen without specifying --arch in some circumstances, and that's used in several tests (which usually insert it in the next stage with rocmlir-driver).  I don't try to convert the type strings if there's no --arch, because then I can't really tell the chip.

### DIFF
--- a/mlir/lib/Conversion/TosaToRock/CMakeLists.txt
+++ b/mlir/lib/Conversion/TosaToRock/CMakeLists.txt
@@ -12,7 +12,7 @@ add_rocmlir_conversion_library(MLIRTosaToRock
   LINK_LIBS PUBLIC
   MLIRIR
   MLIRLinalgDialect
-  MLIRRockConv2dGenerator
+  MLIRRockConvGenerator
   MLIRRockTransforms
   MLIRLinalgUtils
   MLIRPass
--- a/mlir/lib/Dialect/Rock/Generator/ConvGenerator.cpp
+++ b/mlir/lib/Dialect/Rock/Generator/ConvGenerator.cpp
@@ -350,8 +350,10 @@ static Type strToType(StringRef dataTypeStr, OpBuilder &builder) {
           .Case("bf16", builder.getBF16Type())
           .Case("i32", builder.getI32Type())
           .Case("i8", builder.getI8Type())
-          .Cases("f8E5M2FNUZ", "bf8", builder.getFloat8E5M2FNUZType())
-          .Cases("f8E4M3FNUZ", "fp8", builder.getFloat8E4M3FNUZType())
+          .Case("f8E5M2", builder.getFloat8E5M2Type())
+          .Case("f8E4M3FN", builder.getFloat8E4M3FNType())
+          .Case("f8E5M2FNUZ", builder.getFloat8E5M2FNUZType())
+          .Case("f8E4M3FNUZ", builder.getFloat8E4M3FNUZType())
           .Default(std::nullopt);
   if (!type) {
     llvm::errs() << "Unknown data type: " << dataTypeStr << "\n";
--- a/mlir/lib/Dialect/Rock/Generator/ConvGenerator.cpp
+++ b/mlir/lib/Dialect/Rock/Generator/ConvGenerator.cpp
@@ -1,5 +1,5 @@
-#include "mlir/Dialect/AMDGPU/Utils/Chipset.h"
 #include "mlir/Dialect/Rock/Generator/ConvGenerator.h"
+#include "mlir/Dialect/AMDGPU/Utils/Chipset.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/Rock/IR/GemmSize.h"
 #include "mlir/Dialect/Rock/IR/Rock.h"
--- a/mlir/lib/Dialect/Rock/Generator/ConvGenerator.cpp
+++ b/mlir/lib/Dialect/Rock/Generator/ConvGenerator.cpp
@@ -1,3 +1,4 @@
+#include "mlir/Dialect/AMDGPU/Utils/Chipset.h"
 #include "mlir/Dialect/Rock/Generator/ConvGenerator.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/Rock/IR/GemmSize.h"
@@ -547,6 +548,13 @@ LogicalResult ConvGenerator::parseConvConfig(OpBuilder &builder,
   config.chipFeatures = splitter.getFeaturesForBackend();
   config.triple = splitter.getTriple().str();
 
+  FailureOr<amdgpu::Chipset> maybeChipset = amdgpu::Chipset::parse(config.chip);
+  if (failed(maybeChipset)) {
+    emitError(UnknownLoc::get(builder.getContext()),
+              "Invalid chipset name: " + config.chip);
+    exit(1);
+  }
+
   strToStr("perf_config", config.perfConfig);
   strToInt("num_cu", config.num_cu);
   strToInt(rock::ReverseGridAttrAttr::getMnemonic().str(), config.reverseGrid);
@@ -557,12 +565,21 @@ LogicalResult ConvGenerator::parseConvConfig(OpBuilder &builder,
     return failure();
   }
 
-  auto canonicalizeDataType = [](const std::string &type) {
+  auto canonicalizeDataType = [&](const std::string &type) {
     if (type == "fp32")
       return std::string("f32");
     if (type == "fp16")
       return std::string("f16");
-    // +++pf: chipset-based expansion of fp8/bf8 here?
+    if (type == "fp8") {
+      if (maybeChipset->hasOcpFp8())
+        return std::string("f8E4M3FN");
+      return std::string("f8E4M3FNUZ");
+    }
+    if (type == "bf8") {
+      if (maybeChipset->hasOcpFp8())
+        return std::string("f8E5M2");
+      return std::string("f8E5M2FNUZ");
+    }
     return type;
   };
   config.operation = op.value();
--- a/mlir/lib/Dialect/Rock/Generator/ConvGenerator.cpp
+++ b/mlir/lib/Dialect/Rock/Generator/ConvGenerator.cpp
@@ -562,10 +562,7 @@ LogicalResult ConvGenerator::parseConvConfig(OpBuilder &builder,
       return std::string("f32");
     if (type == "fp16")
       return std::string("f16");
-    if (type == "f8E5M2FNUZ")
-      return std::string("bf8");
-    if (type == "f8E4M3FNUZ")
-      return std::string("fp8");
+    // +++pf: chipset-based expansion of fp8/bf8 here?
     return type;
   };
   config.operation = op.value();
--- a/mlir/lib/Dialect/Rock/IR/RockDialect.cpp
+++ b/mlir/lib/Dialect/Rock/IR/RockDialect.cpp
@@ -532,6 +532,10 @@ GemmSize GemmSize::fromConvolution(ConvOpType type,
   return GemmSize(gemmGSize, gemmMSize, gemmKSize, gemmNSize);
 }
 
+static bool isFloat8Type(Type type) {
+  return isa<FloatType>(type) && type.getIntOrFloatBitWidth() == 8;
+}
+
 static LogicalResult verifyGemmTypes(Operation *op, GemmFeatures features,
                                      StringRef arch, Type elemTypeA,
                                      Type elemTypeB, Type elemTypeC) {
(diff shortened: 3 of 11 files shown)


pcf000:
Oh, and I haven't been able to reproduce the reported error on gfx90a.  Could be specific to navi4x, could be I'm building MIGraphX wrong.

mlir/lib/Dialect/Rock/Generator/ConvGenerator.cpp
```diff
@@ -350,8 +350,10 @@ static Type strToType(StringRef dataTypeStr, OpBuilder &builder) {
           .Case("bf16", builder.getBF16Type())
           .Case("i32", builder.getI32Type())
           .Case("i8", builder.getI8Type())
-          .Cases("f8E5M2FNUZ", "bf8", builder.getFloat8E5M2FNUZType())
-          .Cases("f8E4M3FNUZ", "fp8", builder.getFloat8E4M3FNUZType())
```
pcf000, line 356: Remove "bf8" and "fp8" from the cases because the intent is that the caller has converted them.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -643,6 +644,14 @@ LogicalResult getTuningProblemStr(rock::RockGemmWrapperInterface gemmIF,
   KernelType opType = gemmIF.getKernelType();
   Operation *gemmOp = gemmIF.getOperation();
 
+  auto f8TypeStr = [](const Type &type) -> const char * {
```
pcf000: Abstraction, for forming the op field of the tuning key.

mlir/test/rocmlir-gen/gemm-misc-options.mlir
```diff
@@ -6,5 +6,7 @@
 // CONV: amdgcn-amd-amdhsa:gfx900   {{.*}}     conv -F 1 -f GNC01 -I NGC01 -O NGC01 -n 128 -c 8 -H 32 -W 32 -k 128 -y 3 -x 3 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -g 1
 // RUN: rocmlir-gen --emit-tuning-key -p -t fp8_fp8 --arch gfx900 | FileCheck %s --check-prefix=CONVFP8
 // CONVFP8: amdgcn-amd-amdhsa:gfx900   {{.*}}     convfp8_fp8 -F 1 -f GNC01 -I NGC01 -O NGC01 -n 128 -c 8 -H 32 -W 32 -k 128 -y 3 -x 3 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -g 1
+// RUN: rocmlir-gen --emit-tuning-key -p -t fp8_fp8 --arch gfx1201 | FileCheck %s --check-prefix=CONVOCPFP8
```
pcf000, line 9: gfx1201 to bring in the OCP types.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -1327,7 +1327,9 @@ static Type typeFromString(StringRef name, MLIRContext *ctx) {
   std::optional<Type> result =
       llvm::StringSwitch<std::optional<Type>>(name)
           .Case("f32", Float32Type::get(ctx))
+          .Case("fp32", Float32Type::get(ctx))
```
pcf000, line 1298: When I was thinking of merging typeFromString() and strToType() -- the latter includes these two cases.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -3746,6 +3748,31 @@ int main(int argc, char **argv) {
   llvm::cl::ParseCommandLineOptions(argc, argv,
                                     "MLIR Rock Dialect host generation\n");
 
+  if (!arch.getValue().empty()) {
```
pcf000, line 3720: Important for fixing the bug -- the typeFromString/strToType changes aren't enough, because the latter runs second and isn't aware of architecture.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
krzysz00: I don't like this.

I'd do a `std::optional<StringLiteral> toShortFp8Name()` and then `llvm::formatv("conv{0}_{1}", ...)` to form the name.

If nothing else, `const char*` doesn't get used much in this codebase

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -3746,6 +3748,31 @@ int main(int argc, char **argv) {
...
+    SmallVector<std::string>f8e4m3TypeNames{"", "f8E4M3FNUZ", "f8E4M3FN"};
+    SmallVector<std::string>f8e5m2TypeNames{"", "f8E5M2FNUZ", "f8E5M2"};
+    int archIndex = llvm::to_underlying(F8TypesChoice::Arch);
+    if (archChip().substr(0, 5) == "gfx12") {
```
krzysz00: I don't like this approach either.

Could you instead, when `arch` is the fp8 type interpretation, just look at the architecture and then go to an if statement for nanoo/ocp? This hacking on arrays thing doesn't feel right. Especially given the amount of `to_underlying()` involved, which just feels wrong

krzysz00 (commented):
I've got nitpicks

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
pcf000: formatv() is what I needed.  Formatted output is better than << for anything half-complicated.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
pcf000: Went with a map<> to avoid the to_underlying() and ?: in the init list to avoid map-hacking.  Trying to simplify the levels of the conditionals.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -643,6 +645,14 @@ LogicalResult getTuningProblemStr(rock::RockGemmWrapperInterface gemmIF,
   KernelType opType = gemmIF.getKernelType();
   Operation *gemmOp = gemmIF.getOperation();
 
+  auto f8TypeStr = [](const Type &type) -> std::optional<StringRef> {
+    if (type.isFloat8E4M3FNUZ() || type.isFloat8E4M3FN())
```
krzysz00, line 642: Note, you might want this to be `std::optional<StringLiteral>` to make the lifetime situation clearer

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -3746,6 +3748,30 @@ int main(int argc, char **argv) {
   llvm::cl::ParseCommandLineOptions(argc, argv,
                                     "MLIR Rock Dialect host generation\n");
 
+  if (!arch.getValue().empty()) {
+    bool archPrefersOCP = (archChip().substr(0, 5) == "gfx12");
```
krzysz00: ... this isn't the test. gfx950

Can you maybe pull `amdgpu::Chipset` in here?

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -3746,6 +3748,30 @@ int main(int argc, char **argv) {
...
+    std::map<F8TypesChoice, std::string> f8e4m3TypeNames{
```
krzysz00: LLVM's types? `DenseMap<F8TypesChoice, StringLiteral>` or something like that?

krzysz00 (approved):
Nitpicking, approved otherwise

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
pcf000: DenseMap yes, StringLiteral no -- there's no implicit constructor for StringLiteral from "", and also the map is used to set some cl::opts that are themselves std::string.

krzysz00: Fair enough, so we're probably good as is

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
pcf000: Hmm.  If I contemplate amdgpu::Chipset, I start looking at TargetInfo and ArchInfo, and now wonder if AmdArchInfo::hasFp8ConversionInstrs is an adequate flag to indicate prefers-OCP.  (At the moment only true for gfx12, but there is no gfx950 in AmdArchDb.cpp.)

pcf000:
Don't merge yet, there are a couple more files that need discussing.

mlir/lib/Conversion/TosaToRock/CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ add_rocmlir_conversion_library(MLIRTosaToRock
   LINK_LIBS PUBLIC
   MLIRIR
   MLIRLinalgDialect
-  MLIRRockConv2dGenerator
+  MLIRRockConvGenerator
```
pcf000, line 17: Belated cleanup of this name.

mlir/lib/Dialect/Rock/Generator/ConvGenerator.cpp
```diff
@@ -547,6 +548,13 @@ LogicalResult ConvGenerator::parseConvConfig(OpBuilder &builder,
   config.chipFeatures = splitter.getFeaturesForBackend();
   config.triple = splitter.getTriple().str();
 
+  FailureOr<amdgpu::Chipset> maybeChipset = amdgpu::Chipset::parse(config.chip);
```
pcf000, line 552: Use the chipset to choose the appropriate 8-bit float types for the architecture.  Here in parseConvConfig(), we won't see the --force-f8-types option because that's in rocmlir-gen.cpp, and we do have a --arch option and use the architecture that's given.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -532,6 +532,10 @@ GemmSize GemmSize::fromConvolution(ConvOpType type,
   return GemmSize(gemmGSize, gemmMSize, gemmKSize, gemmNSize);
 }
 
+static bool isFloat8Type(Type type) {
```
pcf000, line 535: No specific predicate for float-type-width.  This one is used in two places that could accept any of the four possibilities.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -118,7 +118,7 @@ GemmFeatures mlir::rock::AmdArchInfo::getDefaultFeatures(Type dataType) {
   }
   bool isMfma = bitEnumContainsAll(theseFeatures, GemmFeatures::mfma);
   if (isMfma && !hasFp8ConversionInstrs) {
-    if (dataType.isFloat8E4M3FNUZ() || dataType.isFloat8E5M2FNUZ())
+    if (isa<FloatType>(dataType) && dataType.getIntOrFloatBitWidth() == 8)
```
pcf000, line 121: Another 8-bit-floats acceptance point, but not worth making a function for the one place.

pcf000:
All the changes should be in now.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
krzysz00: `hasFp8ConversionInsts` isn't the right flag - gfx94x have FP8 conversion instructions. You'll need to add a new flag

krzysz00 (approved):
Approved, thank you

PR #1635, opened 2024-08-29
disable occupancy warnings

umangyadav:
While lowering to ROCDL, rocmlir sets `amdgpu-waves-per-eu` attribute to guide backend code generation. 
When it sets this attribute, it may not be possible to always accurately determine range of `waves-per-eu` in advance and therefore backend code generation may not be able to meet constraints set by `waves-per-eu`

Suppressing warnings as they are unnecessary for consumers for rocMLIR.

Fixes https://github.com/ROCm/rocMLIR-internal/issues/1584

### DIFF
--- a/external/llvm-project/llvm/lib/Target/AMDGPU/AMDGPUAsmPrinter.cpp
+++ b/external/llvm-project/llvm/lib/Target/AMDGPU/AMDGPUAsmPrinter.cpp
@@ -1083,7 +1083,7 @@ void AMDGPUAsmPrinter::getSIProgramInfo(SIProgramInfo &ProgInfo,
   ProgInfo.Occupancy = AMDGPUMCExpr::createOccupancy(
       STM.computeOccupancy(F, ProgInfo.LDSSize), ProgInfo.NumSGPRsForWavesPerEU,
       ProgInfo.NumVGPRsForWavesPerEU, STM, Ctx);
-
+  /*
   const auto [MinWEU, MaxWEU] =
       AMDGPU::getIntegerPairAttribute(F, "amdgpu-waves-per-eu", {0, 0}, true);
   uint64_t Occupancy;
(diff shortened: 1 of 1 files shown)


krzysz00 (commented):
I would substantially prefer if we had a way to silence (this particular) `DiagnosticInfoOptimizationFailure` through some LLVM configuration mechanism, but I'll accept this patch if there's thonging else we can do.

umangyadav:
> I would substantially prefer if we had a way to silence (this particular) `DiagnosticInfoOptimizationFailure` through some LLVM configuration mechanism, but I'll accept this patch if there's thonging else we can do.

There is a way but it requires more changes than commenting out. 

`DiagnosticInfoOptimizationFailure` is always enabled by definition. 

https://github.com/ROCm/rocMLIR/blob/7f83159a3252c47931f119190f83de4a430df09f/external/llvm-project/llvm/include/llvm/IR/DiagnosticInfo.h#L1004

https://github.com/ROCm/rocMLIR/blob/7f83159a3252c47931f119190f83de4a430df09f/external/llvm-project/llvm/include/llvm/IR/DiagnosticInfo.h#L987

Because its "Severity" is set to "DS_Warning" during construction.

https://github.com/ROCm/rocMLIR/blob/7f83159a3252c47931f119190f83de4a430df09f/external/llvm-project/llvm/lib/IR/DiagnosticInfo.cpp#L365

I would need to change `DiagnosticInfoOptimizationFailure` to `DiagnosticInfoIROptimization` instead to set Severity manually.

https://github.com/ROCm/rocMLIR/blob/7f83159a3252c47931f119190f83de4a430df09f/external/llvm-project/llvm/include/llvm/IR/DiagnosticInfo.h#L668

krzysz00 (approved):
Near the comment, could you add a note about why this is commented out? 

Otherwise, approved

umangyadav:
@krzysz00  can this go in ? CI is all green

PR #1637, opened 2024-09-03
Fix overly-strict guards in LLVM conversions for fp8 intrinsic

krzysz00:
After fp8 conversions were lowered to AMDGPU dialect ops, those operations were not being converted to the LLVM intrinsics they correspond to because the rewrite patterns were still checking for gfx940+.

As part of this, factor out tests for type-match isto isNativeFp8() and isNativeBf8() functions in the AMDGPUToRocdl rewrites.

Also, fix a typo in isGfx940() that caused it to be true for gfx950.

Finally, test all these OCP format conversions by duplicating the gfx940 tests.

### DIFF
--- a/external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/Utils/Chipset.h
+++ b/external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/Utils/Chipset.h
@@ -19,7 +19,7 @@ struct Chipset {
   static FailureOr<Chipset> parse(StringRef name);
 
   bool isGfx940() const {
-    return majorVersion == 9 && minorVersion >= 0x40 && majorVersion < 0x50;
+    return majorVersion == 9 && minorVersion >= 0x40 && minorVersion < 0x50;
   }
   bool hasOcpFp8() const {
     return (majorVersion == 9 && minorVersion >= 0x50) || majorVersion >= 12;
--- a/external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
+++ b/external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
@@ -462,6 +462,20 @@ static void wmmaPushOutputOperand(ConversionPatternRewriter &rewriter,
   }
 }
 
+/// Return true if `type` is the E5M2 variant of an 8-bit float that is
+/// supported by the `_bf8` instructions on the given `chipset`.
+static bool isNativeBf8(Chipset chipset, Type type) {
+  return (chipset.isGfx940() && type.isFloat8E5M2FNUZ()) ||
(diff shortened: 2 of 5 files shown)


external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/Utils/Chipset.h
```diff
@@ -19,7 +19,7 @@ struct Chipset {
   static FailureOr<Chipset> parse(StringRef name);
 
   bool isGfx940() const {
-    return majorVersion == 9 && minorVersion >= 0x40 && majorVersion < 0x50;
+    return majorVersion == 9 && minorVersion >= 0x40 && minorVersion < 0x50;
```
pcf000, line 22: Argh.  I need to fix that in my upstream pull request, too.

krzysz00: I think you'll want to pull all of this into your upstream PR?

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -758,7 +766,7 @@ LogicalResult ExtPackedFp8OpLowering::matchAndRewrite(
     ExtPackedFp8Op op, ExtPackedFp8OpAdaptor adaptor,
     ConversionPatternRewriter &rewriter) const {
   Location loc = op.getLoc();
-  if (chipset.majorVersion != 9 || chipset.minorVersion < 0x40)
+  if (!(chipset.isGfx940() || chipset.hasOcpFp8()))
```
dhernandez0, line 769: suggestion: might be easier to have a chipset.hasFp8() ?

external/llvm-project/mlir/include/mlir/Dialect/AMDGPU/Utils/Chipset.h
```diff
@@ -19,7 +19,7 @@ struct Chipset {
   static FailureOr<Chipset> parse(StringRef name);
 
   bool isGfx940() const {
```
dhernandez0, line 21: are we sure some arch in the future won't have support for E5M2FNUZ and E4M3FNUZ? If that happens we will have to change the name of the function to something else. What if we call this hasNonOcpFp8() or something like that?

manupak (approved):
An observation : the APIs within chipset is not ergonomic to use as a general comment, however thats not the scope of the PR.

this would ve been easier for a user of `chipset` to have chipset.hasNativeSupport(Type t) API rather than querying chipset.IsXGPU() && type.isWhatWeKnowAsNativelySupportedInXGPU().

this is approved on the basis of choices that has already been made on upstream already.

dhernandez0 (approved):

PR #1639, opened 2024-09-05
Add support for `pad` + removeSubDims in `removeUpperDims`

manupak:
Currently, the `pad` is not properly handled when removing upper dims,
especially in presense of dimensions collapse and expansion happenning.

This PR adds support for that in terms of the two cases that be there:
1) If a part of the dimension that is being removed does not involve a padded region : 
this means we can do the padding as usual as the padding is contained within the subtile.
2) If a part of the dimension that is being removed does involve a padded region : 
From the subtile's perspective, this means we are taking a tile out of a dimension where edges are padded. Therefore, as long as the subtile is considered the padding does not affect coordinates within the subtile because they would be materialized with the padded (zero most of the time) value. Thus, we ll ignore the padding.

closes : https://github.com/ROCm/rocMLIR-internal/issues/1590

### DIFF
--- a/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
+++ b/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
@@ -2118,12 +2118,30 @@ static FailureOr<rock::TransformMapAttr> removeUpperDimsFromMap(
       case TransformType::Pad: {
         for (auto [idx, upperDim] : llvm::enumerate(tr.getUpperDims())) {
           LLVM_DEBUG(llvm::dbgs() << "pad upper dim = " << upperDim << "\n");
-          assert(!removedSubDims.contains(upperDim));
+          if(removedSubDims.contains(upperDim)){
+            // If the padded dimension is being sub-tiled, then the padding is meaningless
+            // because it happens at the either boundary and only some tiles will have the effect.
+            // For all GPU usecases, we would materialize the padded region in sub tiles.
+            // Thus, we ignore the padding in sub-tile views.
+            LLVM_DEBUG(llvm::dbgs() << "Some parts of the padded dimension is to be removed.\n");
+            const auto lowerDim = tr.getLowerDims()[idx];
+            origLowerBounds[lowerDim] = origUpperBounds[upperDim];
+            args.params.append({0, 0});
+          }
+          else{
+            args.params.append({tr.getParams()[idx * 2], tr.getParams()[idx * 2 + 1]});
+          }
         }
+        break;
+      }
+      case TransformType::Broadcast:
+      case TransformType::AddDim:
+      case TransformType::ConstDim: {
         llvm::copy(tr.getParams(), std::back_inserter(args.params));
         break;
       }
       default: {
+        LLVM_DEBUG(llvm::dbgs() << "Unsupported coordinate transform:" << tr << "\n");
         return failure();
       }
       }
--- a/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
+++ b/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
@@ -1953,6 +1953,7 @@ static FailureOr<rock::TransformMapAttr> removeUpperDimsFromMap(
     case TransformType::AddDim:
     case TransformType::ConstDim:
     case TransformType::Broadcast:
+    case TransformType::Pad:
     case TransformType::Merge: {
       assert(!mustBeModified && "must be preserved or removed completely");
       [[fallthrough]];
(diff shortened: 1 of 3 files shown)


mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -2107,14 +2108,46 @@ static FailureOr<rock::TransformMapAttr> removeUpperDimsFromMap(
...
+                                  << " to:" << upperToLower[dim] << "\n");
           newRemovedSubDims[upperToLower[dim]] = subDimInfo;
         }
-        [[fallthrough]];
+        llvm::copy(tr.getParams(), std::back_inserter(args.params));
+        break;
       }
-      default: {
+      case TransformType::Pad: {
+        for (auto [idx, upperDim] : llvm::enumerate(tr.getUpperDims())) {
```
krzysz00, line 2119: ... why do we need this case?

Also, I'm dubious about leaving this rather specific assumption in a generically-named function without any warnings that you're about to cause problems.

krzysz00 (commented):
I'm somewhat hazy on the usefulness of the complicated case here - or at least I'd like to see it made explicit in the caller that we're blowing padding away

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -2107,14 +2108,46 @@ static FailureOr<rock::TransformMapAttr> removeUpperDimsFromMap(
...
+        break;
+      }
+      case TransformType::Broadcast:
+      case TransformType::AddDim:
+      case TransformType::ConstDim: {
         llvm::copy(tr.getParams(), std::back_inserter(args.params));
         break;
       }
+      default: {
+        LLVM_DEBUG(llvm::dbgs()
```
dhernandez0, line 2147: llvm::errs() ?

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
manupak: a sub-tile that d tile a padded region will need to index into the padded region as it was unpadded -- i.e. fully materialized region with padded value.

I dont see any problem with that given how our padding works. Im happy to entertain an example if you have one that is problematic.

Since this is blocking any reduction blockwise fusion, I can put a option to enable handling the padding to the function and use it explicitly

manupak: > ... why do we need this case? 

When we tile a partial reduction axis, we need to pad it such that it has a exact split in block & thread tile. When that happens, we d be tiling the padded region. As you know, block & thread tile it is irrelevant that grid coordinates had padding in them -- as we would have tile sizes as if they were unpadded.

manupak:
> I'm somewhat hazy on the usefulness of the complicated case here - or at least I'd like to see it made explicit in the caller that we're blowing padding away.

Thanks for taking a look.

Im happy to put the explicit trigger to enable padding unless you get around to understand it or make me understand your concern -- to keep this moving. I ve replied inline in the comment.

Though, I can only see myself not using that option where I know for sure padding is not encountered (that is all current usecases where we know this is case by construction because we create the maps statically for all layouts)

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
manupak: hmmm we dont know whether this is an error... yet. It depends on the caller. Caller could also test the function and handle if it fails.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -2107,14 +2108,46 @@ static FailureOr<rock::TransformMapAttr> removeUpperDimsFromMap(
...
+        llvm::copy(tr.getParams(), std::back_inserter(args.params));
+        break;
       }
-      default: {
+      case TransformType::Pad: {
+        for (auto [idx, upperDim] : llvm::enumerate(tr.getUpperDims())) {
+          LLVM_DEBUG(llvm::dbgs() << "pad upper dim = " << upperDim << "\n");
+          if (removedSubDims.contains(upperDim)) {
+            // If the padded dimension is being sub-tiled, then the padding is
+            // meaningless because it happens at the either boundary and only
```
krzysz00, line 2123: "at the either boundary"?

krzysz00 (approved):
Ok, so, to rephrase things, what I'm worried about is someone hitting the "pretend block IDs don't exist" button and suddenly seeing padded regions turn valid where that isn't expected.

Maybe this is more of a "I'd like a warning on the function" sort of thing.

This is partly because I thought of `removeUpperDims()` as something like `createSubtileByErasing()` ... where ... yeah, I see your point that if you want to handle `Pad{}`, this is the only reasonable way to do it.

So approved pending a note to help the next person trying to figure out if they're using the function wrong or something

manupak:
ok added docs to describe what to expect if `pad` is encountered.

PR #1644, opened 2024-09-11
Use cmake -E touch for cross-platform compatibility

stefankoncarevic:
This PR replaces the platform-specific touch command with cmake -E touch to ensure compatibility across different operating systems. The cmake -E touch command works on both Windows and Linux, eliminating the need for conditional branches in the CMake script.


stefankoncarevic:
I believe this change will help improve cross-platform compatibility by using cmake -E touch instead of platform-specific commands. This should simplify the CMake script and ensure consistent behavior across different operating systems.
https://github.com/ROCm/rocMLIR-internal/issues/1609

krzysz00 (approved):

PR #1646, opened 2024-09-11
[DO NOT SQUASH] Stop treating bf16 as i16

krzysz00:
* Remove all the hacks in rocmlir-gen for not using bfloat constants
because LLVM supports those now
* Remove some old header files that aren't used anywhere
* Update the list of types that get passed to
arith-emulate-unsupported-floats because you can now ask LLVM for a
bfloat add (for example) and the compiler will do the right thing
* [external part, to be upstreamed] Get the code out of the ROCDL lowering that converts bf16 to i16 instead of keeping it intact (so that it'll become LLVM's `bfloat`)


external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
```diff
@@ -694,18 +694,25 @@ struct WMMAOpLowering : public ConvertOpToLLVMPattern<WMMAOp> {
...
     Location loc = op.getLoc();
-    Type outType = typeConverter->convertType(op.getDestD().getType());
+    auto outType =
+        cast<VectorType>(typeConverter->convertType(op.getDestD().getType()));
 
     if (chipset.majorVersion != 11 && chipset.majorVersion != 12)
       return op->emitOpError("WMMA only supported on gfx11 and gfx12");
 
+    // The WMMA operations represent vectors of bf16s as vectors of i16s, so we
+    // need to bitcast bfloats to i16 and then bitcast them back.
```
dhernandez0, line 704: can't we change the ROCDL WMMA to take bf16?

dhernandez0 (approved):

external/llvm-project/mlir/test/Conversion/AMDGPUToROCDL/wmma.mlir
```diff
@@ -16,6 +16,7 @@ func.func @mfma_to_rocdl(%arg0 : vector<16xf16>, %arg1 : vector<8xf32>, %arg2 :
   // CHECK: rocdl.wmma.f16.16x16x16.f16{{.*}}: (vector<16xf16>, vector<16xf16>, vector<8xf16>, i1) -> vector<8xf16>
   amdgpu.wmma %arg0 * %arg0 + %arg4 {subwordOffset = 0 : i32}: vector<16xf16>, vector<16xf16>, vector<8xf16>
   // CHECK: rocdl.wmma.bf16.16x16x16.bf16{{.*}}: (vector<16xi16>, vector<16xi16>, vector<16xi16>, i1) -> vector<16xi16>
+  // CHECK-NEXT: llvm.bitcast {{.*}} : vector<16xi16> to vector<16xbf16>
   amdgpu.wmma %arg3 * %arg3 + %arg3 {subwordOffset = 1 : i32}: vector<16xbf16>, vector<16xbf16>, vector<16xbf16>
   // CHECK: rocdl.wmma.bf16.16x16x16.bf16{{.*}}: (vector<16xi16>, vector<16xi16>, vector<8xi16>, i1) -> vector<8xi16>
```
dhernandez0, line 21: same test here as well?

external/llvm-project/mlir/lib/Conversion/AMDGPUToROCDL/AMDGPUToROCDL.cpp
krzysz00: Go talk to the compiler folks about that, but ... probably not, there's non-trivial backwards compatibility issues at play there, last I heard

external/llvm-project/mlir/test/Conversion/AMDGPUToROCDL/wmma.mlir
krzysz00: Point, might want to get on that

PR #1649, opened 2024-09-12
Fix aliasing for LDS

dhernandez0:
In this PR we add alias information for LDS loads and writes. First, we revert the GpuAllocOp per color in reuse LDS pass, instead we create a single GpuAllocOp. 

Then, we use a NoAliasView to annotate that each color does not alias with each other. NoAliasView is just a view (it's lowered into a view) that we know does not alias with any other NoAliasView.

In the PropagateAlias pass we find the operations with alias information (AliasAnalysisOpInterface) and try to track their last argument to a NoAliasView, if that's possible, we set the right alias_scope and noalias_scope.

For this to work, we also add AliasAnalysisOpInterface to memref.load, memref.store, vector.load and vector.store.

TODO:
- small performance regression in one attention test due to unrolling issues, trying to fix it
- upstream PR?

closes ticket: https://github.com/ROCm/rocMLIR-internal/issues/1581


external/llvm-project/mlir/include/mlir/Dialect/MemRef/IR/MemRefOps.td
```diff
@@ -1221,9 +1224,29 @@ def LoadOp : MemRef_Op<"load",
   let arguments = (ins Arg<AnyMemRef, "the reference to load from",
                            [MemRead]>:$memref,
                        Variadic<Index>:$indices,
-                       DefaultValuedOptionalAttr<BoolAttr, "false">:$nontemporal);
+                       DefaultValuedOptionalAttr<BoolAttr, "false">:$nontemporal,
+                       OptionalAttr<LLVM_AliasScopeArrayAttr>:$alias_scopes,
```
krzysz00, line 1228: ... that feels like it'll cause screaming.

We might want to RFC up an MLIR version of these annotations

krzysz00: My other theory here is to let these be discardable attributes - use `llvm.alias_scopes = ...`

That should be more reasonable and we can argue that, hey, these are advisory annotations and if a pass doesn't preserve them, well, that's life until someone designs a better scheme for this

external/llvm-project/mlir/lib/Dialect/MemRef/IR/MemRefDialect.cpp
```diff
@@ -73,3 +73,7 @@ std::optional<Operation *> mlir::memref::findDealloc(Value allocValue) {
   }
   return dealloc;
 }
+
+SmallVector<Value> LoadOp::getAccessedOperands() { return {getMemRef()}; }
+
+SmallVector<Value> StoreOp::getAccessedOperands() { return {getMemRef()}; }
```
krzysz00, line 79: Nit: newline

krzysz00 (commented):
Hm ... I can't tell what I think about `rock.noalias_view` as a name, so it's probably fine ... maybe something like `rock.disjoint_slice` but that's also not quite it ... eh, that's bikeshedding, never mind.

---

Overall, my thoughts are that the minimal change to make here would to be to make those LLVM-specific attributes discardable as an interim measure, and then we can have a long chat with everyone about how to abstract them into something MLIR-y that'd work for other backends, since I can tell that people will Complain (tm) about a MemRef <-> LLVM coupling like that.

Anyhow, thanks for keeping up on this!

dhernandez0:
This PR makes two changes:
1. Creates a single GpuAllocOp in the reuse LDS pass (instead of one per color).
2. Adds aliasing information.

After implementing the first change, I observed some performance regressions. Upon further investigation, it became clear that adding aliasing information does not resolve the issue. The actual cause is that the LLVM loop unrolling pass fails to unroll certain loops after the first change. I can address this by manually invoking mlir::affine::loopUnrollFull() at the MLIR level for the affected loops.

Given this, I don’t think the aliasing information PR is necessary at this time. Additionally, since we're in maintenance mode, it's unclear whether implementing both (1) and manual loop unrolling in MLIR is worthwhile, as the main benefit is slightly cleaner code for the reuse LDS pass.

manupak:
Thanks @dhernandez0 for getting me upto speed. 
I agree with you on all points...

I feel aliasing info propogation from memref to llvm should be discussed and done properly in upstream.
If they are not helping to resolve an existing issue (perf or otherwise), I d rather not pull that into rocMLIR at this point.

PR #1650, opened 2024-09-12
[DO NOT SQUASH] Use half-precision math library calls

dhernandez0:
In this PR we add f16 math library calls. 
- Refactor LegalizeToF32 to get supported float types and target type as arguments. 
- Add f16 math library calls to MathToROCDL
- Update tests 

TODO:
- ~test MIGraphX model (accuracy)~ 
- ~check if there's any performance improvement~ roughly same performance for bert_base_cased_1_fp16_gpu, distilgpt2_1_fp16_gpu and resnet50-v1-7 (--fp16)

- upstream PR?

closes ticket: https://github.com/ROCm/rocMLIR-internal/issues/1439


external/llvm-project/mlir/lib/Dialect/Math/Transforms/ExtendToLLVMTypes.cpp
```diff
@@ -0,0 +1,175 @@
...
+  patterns.add<ExtendToLLVMTypesRewritePattern>(typeConverter,
+                                                patterns.getContext());
+}
+
+void ExtendToLLVMTypesPass::runOnOperation() {
+  Operation *op = getOperation();
+  MLIRContext *ctx = &getContext();
+
+  // Parse target type
+  std::optional<Type> maybeTargetType = parseFloatType(ctx, targetTypeStr);
```
dhernandez0, line 143: I copied this from EmulateUnsupportedFloats.cpp. Where can I move this function so that it can be used by both passes?

external/llvm-project/mlir/lib/Dialect/Math/Transforms/ExtendToLLVMTypes.cpp
```diff
@@ -0,0 +1,175 @@
...
+  using math::impl::MathExtendToLLVMTypesBase<
+      ExtendToLLVMTypesPass>::MathExtendToLLVMTypesBase;
+
+  void runOnOperation() override;
+};
+} // namespace
+
+/// Map strings to float types. This function is here because no one else needs
+/// it yet, feel free to abstract it out.
+static std::optional<FloatType> parseFloatType(MLIRContext *ctx,
```
dhernandez0: I copied this from EmulateUnsupportedFloats.cpp. Where can I move this function so that it can be used by both passes?

external/llvm-project/mlir/include/mlir/Dialect/Math/Transforms/Passes.h
```diff
@@ -56,11 +56,13 @@ void populateMathPolynomialApproximationPatterns(
 void populateUpliftToFMAPatterns(RewritePatternSet &patterns);
 
 namespace math {
-void populateLegalizeToF32TypeConverter(TypeConverter &typeConverter);
-void populateLegalizeToF32ConversionTarget(ConversionTarget &target,
-                                           TypeConverter &typeConverter);
-void populateLegalizeToF32Patterns(RewritePatternSet &patterns,
-                                   TypeConverter &typeConverter);
+void populateExtendToLLVMTypesTypeConverter(TypeConverter &typeConverter,
```
krzysz00: I don't think `LLVM` is the right naming here - there's nothing LLVM specific about this. Perhaps `ExtendToSupportedTypes` and then `const SetVector<Type> & extraSupportedTypes`?

(With it being documented that `f32` and `f64` are always implicitly supported)

external/llvm-project/mlir/include/mlir/Dialect/Math/Transforms/Passes.td
```diff
@@ -28,11 +28,18 @@ def MathLegalizeToF32 : Pass<"math-legalize-to-f32"> {
...
+    the original semantics while enabling lowering. The supported floating-point
+    types for the target are passed as arguments.
 
     As an exception, this pass does not legalize `math.fma`, because
     that is an operation frequently implemented at low precisions.
   }];
+  let options = [
+    ListOption<"sourceTypeStrs", "source-types", "std::string",
+      "MLIR types with arithmetic support on a given target">,
+    Option<"targetTypeStr", "target-type", "std::string", "\"f32\"",
```
krzysz00, line 41: ... On the other hand, I can see the logic to having the target type being an option. We can probably keep that

external/llvm-project/mlir/lib/Conversion/GPUCommon/OpToFuncCallLowering.h
```diff
@@ -78,7 +78,15 @@ struct OpToFuncCallLowering : public ConvertOpToLLVMPattern<SourceOp> {
 private:
   Value maybeCast(Value operand, PatternRewriter &rewriter) const {
     Type type = operand.getType();
-    if (!isa<Float16Type>(type))
+    if (isa<ShapedType>(type)) {
```
krzysz00: I think the condition might want to be `!isa<FloatType>(type))`?

Also, no need for braces here, I think

external/llvm-project/mlir/lib/Dialect/Math/Transforms/ExtendToLLVMTypes.cpp
krzysz00: I'd go looking for some sort of utilities library or somewhere under `IR/`.

Or, really, `Arith/Utils` might work, since `math` depends on `arith` last I checked.

krzysz00 (commented):
I've got some naming/design complaints, but overall don't see too much wrong here and I'm glad you put in the work for this

external/llvm-project/mlir/include/mlir/Dialect/Arith/Utils/Utils.h
```diff
@@ -130,6 +130,10 @@ namespace arith {
 Value createProduct(OpBuilder &builder, Location loc, ArrayRef<Value> values);
 Value createProduct(OpBuilder &builder, Location loc, ArrayRef<Value> values,
                     Type resultType);
+
+// Map strings to float types.
+std::optional<FloatType> parseFloatType(Builder &b, StringRef name);
```
krzysz00: ... if we're doing this, we could probably rig it to just take a `MLIRContext*`, at the cost of a somewhat uglier implementation

krzysz00 (approved):
My two observations:
- There's a discrepancy between the `populate()` calls and the command-line setup which could be confusing. We might want to resolve that, and I don't have a strong opinion on which direction to go with it. (That is, if we want the populate()s to take an extra set, we'd hardcode F32 and F64 into the type converter). But that's up to you
- We'll want to do some `git rebase -i`ing and a `[DO NOT SQUASH]` to split up the math legalization change and the rocdl change - those will be separate PRs upstream.

That being said, this here's fundamentally fine, so approved pending whatever else you want to do here

dhernandez0:
> * There's a discrepancy between the `populate()` calls and the command-line setup which could be confusing. We might want to resolve that, and I don't have a strong opinion on which direction to go with it. (That is, if we want the populate()s to take an extra set, we'd hardcode F32 and F64 into the type converter). But that's up to you

I understand you mean calling addConversion() for f64 and f32 (and ShapedTypes of f64 and f32)? That seems like duplicating the code to me. I'm not sure how to do this cleanly, another way would be to pass MLIRContext to populateExtendToSupportedTypesConversionTarget() and there just copy the extraSupported vector and add f64 and f32.
I'm not even sure why those populate functions are public, they seem internal stages of the pass?

PR #1652, opened 2024-09-14
[DO NOT SQUASH] func.read_access/func.write_access -> mhal.*

krzysz00:
Rename `func.read_access` and `func.write_access` to `mhal.read_access` and `mhal.write_access` and move setting them from the Tosa partitioner to an MHAL pass


external/mlir-hal/include/mlir/Dialect/MHAL/Transforms/Passes.td
```diff
@@ -22,6 +22,15 @@ def MHALPackageTargetsPass : Pass<"mhal-package-targets", "ModuleOp"> {
   let summary = "apply target implementation to host kernel funcs";
 }
 
+def MHALAnnotateAccessKindsPass : Pass<"mhal-annotate-access-kinds", "func::FuncOp"> {
+  let summary = "Annotate arguments for read and write access";
+  let description = [{
+    Annotate arguments and results of functions to preserve readonly/writeonly/read-write
```
umangyadav:
```suggestion
    Annotate arguments and results of functions to preserve read-only/write-only/read-write
```

This Looks nicer

umangyadav (approved):

PR #1653, opened 2024-09-17
[DO NOT SQUASH] Fix vector conversion (ExtendToSupportedTypes)

dhernandez0:
Fix a bug I recently introduced for vector conversion. Due to actually using __ocml_erf_f16, a test changes slightly from 0.999 to 1 and -0.999 to -1.


manupak:
This needs to be two commits (one for external changes and one for rocmlir changes) and [DO NOT SQUASH] otherwise approved!

krzysz00 (approved):

PR #1655, opened 2024-09-18
Do the ROCMLIR_DEBUG_FLAGS registration entirely in rocMLIR.

pcf000:
Paul Fultz pointed out that I could do the registration on the rocMLIR side.  That's easier, let's do it that way.

### DIFF
--- a/mlir/include/mlir/InitRocMLIRCLOptions.h
+++ b/mlir/include/mlir/InitRocMLIRCLOptions.h
@@ -16,6 +16,7 @@
 #include "mlir/IR/AsmState.h"
 #include "mlir/IR/MLIRContext.h"
 #include "mlir/Pass/PassManager.h"
+#include "llvm/Support/CommandLine.h"
 
 namespace mlir {
 
@@ -23,6 +24,16 @@ inline void registerMLIRCLOptions() {
   mlir::registerAsmPrinterCLOptions();
   mlir::registerMLIRContextCLOptions();
   mlir::registerPassManagerCLOptions();
+}
+
+inline void registerRocMLIROptions() {
+  const char *fakeArgv[] = {"rocMLIR-invoked-as-library",
+                            "--mlir-print-local-scope"};
+  mlir::registerMLIRCLOptions();
+  llvm::cl::ParseCommandLineOptions(
+      sizeof(fakeArgv) / sizeof(const char *), fakeArgv,
+      "Fake 'command line' for MIGraphX library debugging", nullptr,
+      "ROCMLIR_DEBUG_FLAGS");
 }
 
 } // namespace mlir
--- a/mlir/lib/CAPI/Dialect/MIGraphX.cpp
+++ b/mlir/lib/CAPI/Dialect/MIGraphX.cpp
@@ -19,6 +19,7 @@
 #include "mlir/Dialect/Rock/Pipelines/Pipelines.h"
 #include "mlir/ExecutionEngine/OptUtils.h"
 #include "mlir/ExecutionEngine/RocmDeviceName.h"
+#include "mlir/InitRocMLIRCLOptions.h"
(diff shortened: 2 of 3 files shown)


mlir/include/mlir/InitRocMLIRCLOptions.h
```diff
@@ -25,6 +26,16 @@ inline void registerMLIRCLOptions() {
   mlir::registerPassManagerCLOptions();
 }
 
+inline void registerRocMLIROptions() {
```
pcf000, line 29: Moved from the original CAPI file and made C++.

mlir/lib/CAPI/Dialect/MIGraphX.cpp
```diff
@@ -136,6 +137,7 @@ MLIR_CAPI_EXPORTED bool mlirGetBinary(MlirModule module, size_t *size,
 MLIR_CAPI_EXPORTED
 void mlirMIGraphXAddHighLevelPipeline(MlirPassManager pm) {
   auto passMan = unwrap(pm);
+  mlir::registerRocMLIROptions();
```
pcf000, line 140: Register the options here, which is called from MIGraphX mlir.cpp's run_high_level_pipeline().

krzysz00: I'm not sure we should be doing this, since you shouldn't be re-parsing the command line multiple times

krzysz00: That is, `registerMLIRCLOptions()` should be called exactly once, which is why we want MIGraphX to do it - just like they need to call dialect registration once

(Like, it might be *safe* to call it repeatedly, but ... why)

mlir/include/mlir/InitRocMLIRCLOptions.h
krzysz00, line 29: `registerRocMLIRLibClOptions()` or some other name that makes it look less like an upstream function

mlir/lib/CAPI/Dialect/MIGraphX.cpp
pcf000: Well, color me confused.  Way back at the beginning, I tried doing everything at the one-time-registration point in MIGraphX, and it wouldn't do print-ir, which led me to doing it in the add-pipeline functions.

Following your comments, I decided to trace that problem more thoroughly, and now it works the way I want it to -- calling the registration function at the same time we register passes and dlalects does in fact lead to recognising --mlir-print-ir-after-all.

I'm going to sit on this for a bit, wipe out all my test installations and caches, and try again from scratch, to make sure that's what's really happening.  Either I'll be able to do it one time, where the other registrations happen, or I'll understand better why I can't.

BTW, registration for command-line options isn't parsing, it's just creation and initialisation.  It's probably innocuous to repeat it, but it's cleaner not to.

krzysz00: So let's get rid of this bit?

pcf000: Yes.  The cheating way is to replace the call to mlir::registerMLIRCLOptions() in mlirRegisterRocMLIRPasses() with a call to mlirRegisterRocMLIROptions() (or whatever we rename it too).  In other words, sneak it into a function that's already called only once.  The ostensibly cleaner way is to call mlirRegisterRocMLIROptions() from get_dialect_registry() in mlir.cpp alongside the other registration functions.  Assuming it continues to work.

krzysz00: Yeah, I'd rather call it in the dialect registry getter or something else that runs as part of library initialization

pcf000:
Superseded by pull request #1665 .

PR #1661, opened 2024-09-25
Update comment regarding barrier gfx908 hack

dhernandez0:
In https://github.com/llvm/llvm-project/pull/109678 we learned this is caused by a backend bug: https://github.com/llvm/llvm-project/issues/109294 
Therefore, we can remove the scheduling barrier when that bug is fixed. Here I'm updating the comment with a TODO.


manupak (approved):
Thanks for cross checking with upstream...

PR #1604, opened 2024-08-08
[CI] Created new CI job pipeline for Navi4x architecture

stefankoncarevic:
This PR adds a new CI pipeline specifically for the Navi4x architecture. Most configurations and functionalities have been adapted from the main Jenkins file, with adjustments for Navi4x. Created stages for setting system properties, environment setup, and executing PR CI tests and E2E tests. 
Please review the Jenkins file and provide any suggestions for further improving the testing process.

PR #1664, opened 2024-09-26
Collected improvements to CI performance testing

pcf000:
These changes fix problems that either caused silent failures or produced unnecessary FAILED entries in performance reports.

perfRunner.py:
 - Force split-k-factor to 1 in tuningDb inside benchmarkFusionKernels.
 - Abstract process-pipeline creation and exception-handling into runPipeline.
 - Copy solution to "rocm_agent_enumerator -name" from pull request #1604.
 - Use toCommandLine to index into tuningDb for fusion tests.
 - Use same column order in the config-not-found case as in the run-config case.

reportUtils.py
 - Simple output tweak for perfRunner.py reports.

Jenkinsfile:
 - Don't do attention perfRunner.py on gfx110x.
 - Don't run the CK benchmarking for gfx110x, because ck-benchmark-driver won't compile.

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -400,6 +400,10 @@ boolean shouldRunBuildAndTest(String codepath) {
     }
 }
 
+boolean isNotNavi3x(String chip) {
+    return "${CHIP}" != 'gfx1100' && "${CHIP}" != 'gfx1101'
+}
+
 
 pipeline {
     agent none
@@ -914,6 +918,11 @@ pipeline {
                         }
                     }
                     stage("Test Attention") {
+                        when {
+                            beforeAgent true;
+                            // gfx110x lacks mfma/wmma necessary for attention.
+                            expression { return isNotNavi3x("${CHIP}") }
+                        }
                         steps {
                             dir('build') {
                                 // Run attention benchmarks
@@ -927,6 +936,9 @@ pipeline {
                         when {
                             beforeAgent true;
                             equals expected: true, actual: params.checkCK;
+                            // ck-benchmark-driver requires CK xdl functions, which are
+                            // only built for gfx9xx.
+                            expression { return isNotNavi3x("${CHIP}") }
                         }
                         steps {
                             catchError (buildResult: null) { // This is an optional stage
--- a/mlir/utils/performance/perfRunner.py
+++ b/mlir/utils/performance/perfRunner.py
@@ -209,6 +209,26 @@ def getMilliseconds(output):
 
     return float(result.group(1))
 
+def runPipeline(proc_specs):
+    procs = []
+    for proc in proc_specs:
+        prev_stdout = procs[-1].stdout if procs else subprocess.DEVNULL
+        po = subprocess.Popen(proc, stdin=prev_stdout, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        procs.append(po)
+    try:
+        for p in procs:
+            p.wait()
+            if p.returncode != 0:
+                raise OSError(str(p.stderr))
+        outs, errs = p.communicate()
+        return outs, errs
+    except Exception as err:
+        print(f"Error:  {err}")
+        print(f"Failing command:  {' '.join(p.args)}")
+        print(f"Failing pipeline:  {' | '.join([' '.join(proc) for proc in proc_specs])}")
+        outs, errs = p.communicate()
+    return outs, errs
+
 class PerfConfiguration:
     TABLE_COLUMNS = []
 
@@ -232,7 +252,7 @@ class PerfConfiguration:
         raise NotImplementedError()
 
     @classmethod
-    def benchmarkExternal(cls, commandLine, paths: Paths, arch, numCU, envs={}):
+    def benchmarkExternal(cls, commandLine, paths: Paths, arch, numCU):
         raise NotImplementedError()
 
     EXTERNAL_NAME = "unknown"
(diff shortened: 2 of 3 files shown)


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -400,6 +400,10 @@ boolean shouldRunBuildAndTest(String codepath) {
     }
 }
 
+boolean isNotNavi3x(String chip) {
```
pcf000, line 403: Couldn't manage to put this condition in a "when" clause.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -914,6 +918,11 @@ pipeline {
                         }
                     }
                     stage("Test Attention") {
+                        when {
+                            beforeAgent true;
+                            // gfx110x lacks mfma/wmma necessary for attention.
```
pcf000, line 923: Compiling the tests would fail with an error.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -927,6 +936,9 @@ pipeline {
                         when {
                             beforeAgent true;
                             equals expected: true, actual: params.checkCK;
+                            // ck-benchmark-driver requires CK xdl functions, which are
```
pcf000, line 939: Building the tool would fail.

mlir/utils/performance/perfRunner.py
```diff
@@ -209,6 +209,26 @@ def getMilliseconds(output):
 
     return float(result.group(1))
 
+def runPipeline(proc_specs):
```
pcf000, line 212: Abstract the common idiom of making a chain of processes, waiting for the resuts, and handling errors and returns.  Also made a slight change to allow cases which print to stderr but with a zero return code to still complete normally -- some tests emit warnings that don't indicate a problem, for instance MIOpenDriver.

mlir/utils/performance/perfRunner.py
```diff
@@ -232,7 +252,7 @@ def toCommandLine(self):
         raise NotImplementedError()
 
     @classmethod
-    def benchmarkExternal(cls, commandLine, paths: Paths, arch, numCU, envs={}):
```
pcf000, line 235: envs was never used.

mlir/utils/performance/perfRunner.py
```diff
@@ -1008,6 +973,8 @@ def generatePerformanceResults(configs, confClass, paths: Paths, arch, numCU, tu
                            suffixes=('', f" ({externalName})"))
     externalTFlopsCol = f"{externalName} TFlops (no MLIR Kernels)"
     df.rename(columns={'TFlops': 'MLIR TFlops', f"TFlops ({externalName})": externalTFlopsCol}, inplace=True)
+#     if tuned_df is None and quick_tuned_df is None:
```
pcf000, line 976: Oops, test code that should be removed.

mlir/utils/performance/perfRunner.py
```diff
@@ -1212,15 +1166,16 @@ def benchmarkFusionKernels(test_dir, paths: Paths, arch, numCU, tuningDb: MaybeT
         # Find the best perf_config
         bestPerf =""
         if tuningDb:
-            if (arch, testVector) in tuningDb:
-                bestPerf = tuningDb[arch, testVector]
+            configStr = config.toCommandLine()
```
pcf000, line 1169: Without the toCommandLine call, testVector lacks the "-m conv" and "-t 1" that tuningDb has in the keys for conv entries.  Then it won't find a perfconfig and gives up.

mlir/utils/performance/perfRunner.py
```diff
@@ -1212,15 +1166,16 @@ def benchmarkFusionKernels(test_dir, paths: Paths, arch, numCU, tuningDb: MaybeT
...
+            if (arch, configStr) in tuningDb:
+                bestPerf = tuningDb[arch, configStr]
                 config.setPerfConfig(bestPerf)
             else: # Tuning DB present but doesn't contain config, add a NaN entry
                 if not testVector in perfResults:
                     oneEntry = config.tableEntry(np.nan)
-                    oneEntry['FileName'] = filename
                     oneEntry['MLIR TFlops'] = np.nan
                     oneEntry['Fusion/MLIR'] = np.nan
+                    oneEntry['FileName'] = filename
```
pcf000, line 1178: Add the columns in the same order that they are in the running case.  Doesn't matter in practice, but it bugged me in diffs.

mlir/utils/performance/reportUtils.py
```diff
@@ -70,7 +70,7 @@ def setCommonStyles(styler: 'pd.io.formats.style.Styler', speedupCols: list, col
         {'selector': 'tbody tr:nth-child(even)', 'props': [('background-color', '#eeeeee')]},
         {'selector': 'table', 'props': [('background-color', '#dddddd'), ('border-collapse', 'collapse')]},
         {'selector': 'th, td', 'props': [('padding', '0.5em'), ('text-align', 'center'), ('max-width', '150px')]}])
-    styler.format(precision=ROUND_DIGITS, na_rep="FAILED")
```
pcf000, line 73: Trying to print something more neutral than "FAILED" for data that we just don't have, like MIOpenDriver and LDSBankConflicts (which it doesn't report).

mlir/utils/performance/perfRunner.py
```diff
@@ -887,28 +897,15 @@ def benchmarkExternal(cls, commandLine, paths: Paths, arch, numCU, envs={}):
...
-            if len(errs) > 0:
-                print("Test printed errors: ", errs.decode('utf-8'))
-                print("Failing command line: ", profilerCommand)
-                if p.returncode != 0:
-                    raise OSError(errs.decode('utf-8'))
-        except subprocess.TimeoutExpired:
-            print("Test timed out: ", profilerCommand)
-            p.kill()
-            outs, errs = p.communicate()
+        outs,errs = runPipeline([profilerCommand])
```
krzysz00, line 900: Nit: space

mlir/utils/performance/perfRunner.py
krzysz00: Why, or is that your PR?

mlir/utils/performance/reportUtils.py
krzysz00: Except that we do probably want our own perf results to `FAILED` if we don't successfully benchmark something but ... yeah, that's a niche case, approved

krzysz00 (approved):
Overall seems fine, good refactorings, approved!

Can you go take a look at the weekly parameter sweep thing to see if it can use your utility?

mlir/utils/performance/reportUtils.py
pcf000: It wants more nuance.  MIOpenDriver never reports LDSBankConflict, so that shouldn't be FAILED, but true failures should.  Meanwhile, it's weird to report a missing perfconfig as "nan".  Not solved here, or anywhere yet.

mlir/utils/performance/perfRunner.py
pcf000: Trying to clean up the display when there isn't a PerfConfig to display, but other report-generating code depends on that column existing even if it's empty.  Will have to approach it differently.

PR #1667, opened 2024-10-02
Fix crash with invalid !migraphx.shaped types

krzysz00:
asFlatMemoryTensor() was being invoked in the MIGraphX type conversions without checking if the underlynig call to asMemoryLayoutTensor() succeeded. This caused crashes when trying to check the rank of a null pointer.

This commit restores us to merely failing the pass.

Fixes https://github.com/ROCm/rocMLIR-internal/issues/1640

### DIFF
--- a/mlir/lib/Dialect/MIGraphX/IR/MIGraphX.cpp
+++ b/mlir/lib/Dialect/MIGraphX/IR/MIGraphX.cpp
@@ -251,6 +251,8 @@ RankedTensorType MIXRShapedType::asMemoryLayoutTensor() const {
 
 RankedTensorType MIXRShapedType::asFlatMemoryTensor() const {
   RankedTensorType memoryTensorType = asMemoryLayoutTensor();
+  if (!memoryTensorType)
+    return nullptr;
   return memoryTensorType.clone(memoryTensorType.getNumElements());
 }
 


mlir/lib/Dialect/MIGraphX/IR/MIGraphX.cpp
```diff
@@ -251,6 +251,8 @@ RankedTensorType MIXRShapedType::asMemoryLayoutTensor() const {
 
 RankedTensorType MIXRShapedType::asFlatMemoryTensor() const {
   RankedTensorType memoryTensorType = asMemoryLayoutTensor();
+  if (!memoryTensorType)
+    return nullptr;
```
umangyadav, line 255: are there any unit-tests for such kind of methods/functions ?

krzysz00: Nope

I'd've added a `not --crash` test, if it weren't for the fact that we're still going to fail the example that triggered the crash.

manupak (approved):
Approving the fix. 
As a code base, we kind of lack the negative testing @umangyadav if thats what you meant. I ll let @krzysz00 decide whether to add one here..

mlir/lib/Dialect/MIGraphX/IR/MIGraphX.cpp
krzysz00: That is to say, we could have a test for this invalid input producing an error, but it's trickier (given that, in general, testing via `lit` is preferred over unit tests) it's harder to express "fails in this specific way but doesn't crash" (probably due to LLVM having a propensity to `report_fatal_error`)

PR #1668, opened 2024-10-03
Use `blockwise_broadcast_reduce` in reduction fusions.

manupak:
~~[TODO] : add more e2e tests and unit tests for passes.~~

Currently, we handle reductions that are being fused with gemm-like operations
by using atomic stores to the destination buffer. This can be cripplingly slow when
most of the output is being reduced as evidenced in layer_norm cases.

This PR adds the ability to `blockwise_broadcast_reduce` on the block sub-tiles of 
of gemm output.
However, in-order to that we need to make sure the reduction dimension is uniformly
distributed across the blocks. This is achieved by : 
1) Firstly, this PR introduces a utility where for a given set of upper dimensions, it can
    traverse a transform stack and produce a list of sub-dimensions per each lower dimension
    where the upper reduction axes are mapped to.
2) Then, this PR introduces `ShuffleGemmForReductions` pass, which will split and transpose
    the parallel dimension of the gemm such that reduction dimension is uniformly
    distributed across the blocks.
3) Then at AlignTiling pass, we extract the block subtile when fusing in the `rock.reduce` operator. 
    Then perform a `blockwise_broadcast_reduce` on the block subtile.
4) Since we only want to write the partial reductions per block, we pad out broadcasted part of the subtile.
    (We rely on any block coordinate that goes to the padded region within the block will not be written out)
5) Then we need to do `Recombine` the modified sub-tile coordinate maps with grid-only coordinates maps.
     a) Here, we drop all the upper-dimensions except `g_block`, `m_block` and `n_block` and obtain the grid-only transform map 
         stack.
     b) In parallel, we re-use getLowerSubDimensions utility to figure out which sub-dimension gets mapped with the above 
         grid-only dimensions.
     c) Then we extract of those sub-dimensions in a bottom up fashion and stitch it up with the said  grid-only transform map 
         stack.

I ll try to create some slides to explain all these. 

In the cases I ve tests, this yields two orders of magnitude (~100x) gains over the pure atomics approach of doing reduction fusions.


mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
```diff
@@ -151,15 +151,22 @@ void rock::buildKernelPipeline(OpPassManager &pm,
   funcPm.addPass(rock::createRockConvToGemmPass());
   funcPm.addPass(rock::createRockGemmToGridwisePass());
   funcPm.addPass(rock::createRockRegularizePass());
+  funcPm.addPass(rock::createRockShuffleGemmForReductions());
   funcPm.addPass(rock::createRockGridwiseGemmToBlockwisePass());
-  funcPm.addPass(rock::createRockBlockwiseGemmToThreadwisePass());
+  // We want to delay blockwise lowering in the fusion cases
+  // until after linalg align pass because with reduction fusion
+  // it may introduce blockwise_reductions.
+  if (!options.enableFusion) {
```
krzysz00, line 159: Nit: we could move `rock-pipeline` and friends out of the `enable-fusion` if statement, so that we can just move `-blockwise-gemm-to-threadwise` after `-linalg-align`

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -405,8 +406,10 @@ void LinalgAlignRewriter::notifyMatchFailure(
 
 static Value applyViewsOnDest(LinalgAlignRewriter &rewriter, Location loc,
                               Value dest, ArrayRef<TransformMapAttr> views) {
-  for (TransformMapAttr trMap : llvm::reverse(views)) {
-    dest = rewriter.create<TransformOp>(loc, dest, trMap);
+  if (!views.empty()) {
```
krzysz00: Why this if statement? Shouldn't the loop be trivial if views is empty?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1145,21 +1148,399 @@ ReduceRewritePattern::matchAndRewrite(rock::ReduceOp reduceOp,
...
+  // Obtain blockwise subtile.
+
+  // This has < block dimensions ... > x tid x iter to Gemm Dimensions.
+  ArrayAttr extraViews = threadwiseWriteOp.getExtraViews();
+  ArrayAttr destTrs;
+  Value dest;
+  std::tie(dest, destTrs, std::ignore) =
+      untransform(rewriter, threadwiseWriteOp.getDest());
+  ArrayAttr toBeReducedViews = prependUpperViews(rewriter, extraViews, destTrs);
+  TransformMapAttr firstCoordTransform =
```
krzysz00, line 1382: `return notifyMatchFailure()` or some such if there are no transforms at all so we don't crash here?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1145,21 +1148,399 @@ ReduceRewritePattern::matchAndRewrite(rock::ReduceOp reduceOp,
...
+      cast<TransformMapAttr>(toBeReducedViews[0]);
+  int upperRank = firstCoordTransform.getUpperBounds().size();
```
krzysz00: This might want to be an `int64_t` to match the others

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1145,21 +1148,399 @@ ReduceRewritePattern::matchAndRewrite(rock::ReduceOp reduceOp,
...
+  removeIndicesSet.clear();
+  for (int64_t i = 0; i < upperRank - 1; i++) {
+    removeIndicesSet.insert(i);
+  }
+  FailureOr<ArrayAttr> threadSubTileViews =
+      removeUpperDims(rewriter, toBeReducedViews, removeIndicesSet);
+
+  // Extract grid-only dims
+  removeIndicesSet.clear();
+  for (int64_t i = 3; i < upperRank; i++) {
```
krzysz00: Why `3` not `upperRank - 2`?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1145,21 +1148,399 @@ ReduceRewritePattern::matchAndRewrite(rock::ReduceOp reduceOp,
...
+      SmallVector<StringRef> nameRefs;
+      for (size_t i = 0; i < blockTileShape.size(); i++) {
+        SmallString<8> dimName(Twine("dim" + Twine(i)).str());
+        names.push_back(dimName);
+        nameRefs.push_back(names.back());
+      }
+      TopDownTMBuilder toReducedView(rewriter, nameRefs, blockTileShape);
+      for (unsigned i = 0; i < blockTileShape.size(); i++) {
+        if (blockReductionAxis == i) {
+          toReducedView.pad({nameRefs[i]}, {0, blockTileShape[i] - 1});
```
krzysz00: ... hold on, is this using `Pad{}` to take a length-1 dimension to a length-`blockTileShape[i]` one?

This feels like it could be `Broadcast{}`?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1145,21 +1148,399 @@ ReduceRewritePattern::matchAndRewrite(rock::ReduceOp reduceOp,
...
+    {
+      SmallVector<TransformMapAttr> transformAttrs;
+      ArrayRef<int64_t> lowerShapeGridOnly =
+          getLowerShape(gridOnlyDims.value());
+      size_t lowerGridOnlyRank = lowerShapeGridOnly.size();
+      for (auto [idx, attr] : llvm::enumerate(paddedReducedTrStack)) {
+        TransformMapAttr trMapAttr = cast<TransformMapAttr>(attr);
+        SmallVector<TransformAttr> trAttrs;
+        SmallVector<int64_t> gridUpperShape;
+        SmallVector<int64_t> gridLowerShape;
```
krzysz00: Do these need to be vectors?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1145,21 +1148,399 @@ ReduceRewritePattern::matchAndRewrite(rock::ReduceOp reduceOp,
...
+      llvm::SmallDenseMap<int64_t, SmallVector<int64_t>> gridSubDims;
+      llvm::SmallDenseMap<int64_t, SmallVector<int64_t>> blockSubDims;
+      TransformMapAttr lastMerge;
+      {
+        llvm::SmallDenseMap<int64_t, SmallVector<SmallString<8>>> names;
+        llvm::SmallDenseMap<int64_t, SmallVector<StringRef>> nameRefs;
+        llvm::SmallDenseMap<int64_t, SmallVector<unsigned>> upperDims;
+        int64_t dimInsertionPoint = 0;
+        // The three dimensions correspond to G, M and N here.
+        for (unsigned dim = 0; dim < 3; dim++) {
```
krzysz00: Same note re hardcoded 3

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
```diff
@@ -0,0 +1,595 @@
...
+ArrayAttr reverse(ArrayAttr attrs) {
+  SmallVector<Attribute> attrsReversed = llvm::to_vector(llvm::reverse(attrs));
+  IRRewriter rewriter(attrs.getContext());
+  return rewriter.getArrayAttr(attrsReversed);
+}
+
+ArrayAttr getAllViewsFromSource(OpOperand *operand) {
+  Value val = operand->get();
+  SmallVector<Attribute> attrs;
+  while (TransformOp trOp = dyn_cast<TransformOp>(val.getDefiningOp())) {
```
krzysz00: This is `untransform()` no?

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -264,13 +299,16 @@ void createGemmTuningRangeBF(TuningParamSet *newSpace,
...
+                                                        gemmNPerBlock) ||
+                        failed(tuningInfo.couldBePerformant(info, gemmParams));
                     if (gemmMPerBlock >= gemmMPerWave &&
                         gemmNPerBlock >= gemmMnPerXdl) {
                       if (succeeded(tuningInfo.paramsProbablyValid(
                               b, info, gemmParams)) &&
                           (kind == TuningParamSetKind::Exhaustive ||
-                           succeeded(
-                               tuningInfo.couldBePerformant(info, gemmParams))))
+                           !nonExhaustiveSkipCondition))
```
krzysz00: Could you implement this into `couldBePerformant()` please? Maybe extend that function to take a `bool isReduction`?

krzysz00: ... This is also, IMO, adding a `has_reduction` to the problem key, so that MIGraphX doesn't cache reduction and non-reduction GEMMs together

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -2319,3 +2374,148 @@ mlir::rock::removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
...
+        nextSubDimInfo = getNextSubDimInfo(subDimInfo);
+    if (failed(nextSubDimInfo))
+      return failure();
+    subDimInfo = nextSubDimInfo.value();
+  }
+  if (subDimInfo.empty()) {
+    return failure();
+  }
+  return subDimInfo;
+}
```
krzysz00, line 2522: Out of curiosity, could this SubDimInfo stuff be used to simplify `collapseContiguousMerges()`?

(Not for this PR, just checking my understanding)

krzysz00 (commented):
First round of looking:
1. I've got the inevitable stash of nitpicks
2. The sub-dim gathering utility looks correct, and I have a solid amount of trust in `removeUpperDims()` at this point
3. The rats' nest that is constructing and manipulating all the transforms is calling out for some refactoring and utilities ... but that's not this PR

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -2033,7 +2033,8 @@ LogicalResult BlockwiseBroadcastReduceOp::verify() {
   }
   if (blockwiseInputPartialReductionTensorElements > wsShape[0]) {
     return emitError(
-        "workspace should be at least the size of elements per block");
+               "workspace should be at least the size of elements per block ")
+           << blockwiseInputPartialReductionTensorElements << " " << wsShape[0];
```
dhernandez0: print variable names as well? or remove if this was for debugging purposes?

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
```diff
@@ -0,0 +1,595 @@
...
+  // Only run this pass on GPU kernel functions.
+  if (!func->hasAttr("kernel")) {
+    return;
+  }
+  ReduceOp largestReductionOp;
+  int64_t currReductionDimSize = 0;
+  func.walk([&](ReduceOp rOp) -> WalkResult {
+    TypedValue<ShapedType> rIn = rOp.getIn();
+    int64_t reduceDimSize =
+        rIn.getType().getShape()[rOp.getAxis().getZExtValue()];
```
dhernandez0, line 613: could the type be vector? in that case we might want to multiply by the vector size as well

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
```diff
@@ -0,0 +1,595 @@
...
+    if (reduceDimSize > currReductionDimSize) {
+      largestReductionOp = rOp;
+    }
+    return WalkResult::advance();
+  });
+  if (largestReductionOp) {
+    auto &bufferDeps = getAnalysis<BufferDependencyAnalysis>();
+    rearrangeGemmParallelDimsForReduction(largestReductionOp, bufferDeps);
+  }
```
dhernandez0, line 628: suggestion: add a debug message if the pass doesn't run

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
```diff
@@ -0,0 +1,595 @@
...
+rearrangeGemmParallelDimsForReduction(ReduceOp rOp,
+                                      const BufferDependencyAnalysis &deps) {
+  FailureOr<std::tuple<ArrayAttr, Operation *>> res =
+      obtainGemmToReduceViews(rOp, deps);
+  if (succeeded(res)) {
+    auto [views, gemmOp] = res.value();
+    LLVM_DEBUG(llvm::dbgs() << "gemmToReduceViews=" << views << "\n");
+    FailureOr<MNPerBlock> mnPerBlock = getMNPerBlock(gemmOp);
+    if (failed(mnPerBlock)) {
+      return;
```
dhernandez0: suggestion: return failure() and print debug message from the caller (the pass will not run).

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -2319,3 +2374,148 @@ mlir::rock::removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
...
+mlir::rock::getLowerSubDimensions(OpBuilder &b, ArrayAttr transformAttrs,
+                                  int64_t dim) {
+  return getLowerSubDimensions(b, transformAttrs, ArrayRef<int64_t>{dim});
+}
+
+FailureOr<llvm::SmallDenseMap<int64_t, SmallVector<SubDimInfo>>>
+mlir::rock::getLowerSubDimensions(OpBuilder &b, ArrayAttr transformAttrs,
+                                  ArrayRef<int64_t> dims) {
+  llvm::SmallDenseMap<int64_t, SmallVector<SubDimInfo>> subDimInfo;
+  TransformMapAttr topMap = cast<TransformMapAttr>(transformAttrs[0]);
```
dhernandez0, line 2389: check transformAttrs is not empty

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -2319,3 +2374,167 @@ mlir::rock::removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
...
+            if (upperToLower.contains(dim)) {
+              LLVM_DEBUG(llvm::dbgs() << "remapping:" << dim << " to "
+                                      << upperToLower[dim] << "\n");
+              nextSubDimInfo[upperToLower[dim]] = subDimInfo;
+            }
+          }
+        } break;
+        case TransformType::Merge: {
+          SmallVector<int64_t> subDimStrides = getStrides(trAttr.getParams());
+          for (size_t subDim = 0; subDim < trAttr.getParams().size();
```
krzysz00: This could be a `llvm::zip()` or an `llvm::enumerate()` if you do need the index

krzysz00 (commented):
Very tiny note that might generalize

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
manupak: Having read the code again, here we have `rock-pipeline` in enableFusion!
Hence, I believe we dont do `enableFusion` for non-fusion cases -- its kind of broken isn;t it? 

In retrospect, do we really need to have `enableFusion` ? The fact that nothing breaks, indicates its not used. WDYT ? 
If agreed, we can remove `enableFusion` option in a followup ?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: extraViews can never be empty by definition because those correspond to g_block x m_block x n_block x tid x iter -> G X M X N. (though destTrs could be empty, thus concat of the two will also never be empty)

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
krzysz00: Yeah, let's lose `enableFusion`. I think it was added so the MIOpen people wouldn't see the overhead

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
krzysz00: In the future, we might get rid of `extraViews`, for example

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1145,21 +1146,406 @@ ReduceRewritePattern::matchAndRewrite(rock::ReduceOp reduceOp,
...
+          // in the matrix coordinates. Here we split out matrix dims
+          // into sub-dims that are related to the said blocking dimensions.
+          SmallVector<SubDimInfo> subDims = lowerSubDims.value()[dim];
+          llvm::sort(subDims, [](const SubDimInfo &L, const SubDimInfo &R) {
+            return L.stride > R.stride;
+          });
+          SmallVector<int64_t> splitSizes;
+          int64_t currSize = toBeReducedShape[dim];
+          for (const SubDimInfo &subDim : subDims) {
+            assert(currSize % (subDim.size * subDim.stride) == 0);
```
krzysz00: Why assert? Maybe fail here?

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
```diff
@@ -0,0 +1,597 @@
...
+      continue;
+    }
+    SmallVector<MemoryEffects::EffectInstance> effects;
+    writeOp.getEffects(effects);
+    for (const MemoryEffects::EffectInstance &effect : effects) {
+      OpOperand *readOperand = effect.getEffectValue<OpOperand *>();
+      LLVM_DEBUG(llvm::dbgs()
+                 << "readOperand = " << readOperand->get() << "\n");
+      // Test against the write operand to guard against [MemRead, MemWrite]
+      if (readOperand && readOperand != writerOperand &&
```
krzysz00, line 102: Doesn't `BufferDependencyAnalysis` do this?

krzysz00: Or have this cached?

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -2066,32 +2075,61 @@ static FailureOr<rock::TransformMapAttr> removeUpperDimsFromMap(
...
                     std::max(removedSubDimInfo.stride, subDimStrides[subDim]);
-                diff = (subDimStrides[subDim] * tr.getParams()[subDim]) /
-                       rhsBoundForRemoval;
+                if (maxStrideSubDim % rhsBoundForRemoval != 0) {
+                  LLVM_DEBUG(
+                      llvm::dbgs()
+                      << "non divisible subDim removal found. aborting..\n");
+                  return failure();
+                }
+                // assert(maxStrideSubDim % rhsBoundForRemoval == 0);
```
krzysz00: Stray comment

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -2319,3 +2374,172 @@ mlir::rock::removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
...
+    if (topMap.getUpperBounds()[dim] == 1) {
+      continue;
+    }
+    LLVM_DEBUG(llvm::dbgs()
+               << "creating subDim of <size=" << topMap.getUpperBounds()[dim]
+               << ",stride=" << 1 << "> @ " << dim << "\n");
+    subDimInfo[dim].push_back({topMap.getUpperBounds()[dim], 1});
+  }
+
+  for (Attribute attr : transformAttrs) {
```
krzysz00: You can do a `.getAsRange<TransformMapAttr>()`

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: that would mean they would go to destTrs.. right? 
coz somewhere there needs to hold the maps that starts with g_block x m_block x n_block x tid x iter -> G X M X N
So the concat would still be not empty.. 

(I can put an assert)

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
manupak: Just to keep the bisectability, I ll do that in a seperate PR if thats ok..

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
manupak: This is heavily inspired by your code here : https://github.com/ROCm/rocMLIR/blob/82378ac44b8627ecbaf9078353fb53588090fe28/mlir/lib/Dialect/Rock/Transforms/Regularize.cpp#L282-L290.

I think its safer to have here as well.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: removed

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
manupak: I moved into couldBePerformant() and I d defer problem key change to be in a different PR for bisectability.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
manupak: potentially yes...

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
manupak: thanks! it did

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: refactored into a function and a failure return

manupak:
Thanks @krzysz00 for all the reviews on this big PR. I really appreciate it.
I have addressed all of them

I have not added the unit tests for passes yet and aling-tiling yet... which I will do next.
@dhernandez0 would you be able to take a look here ?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1097,6 +1098,440 @@ MemcpyRewritePattern::matchAndRewrite(memref::CopyOp copy,
...
+      untransform(rewriter, threadwiseWriteOp.getDest());
+
+  ArrayAttr toBeReducedViews = prependUpperViews(rewriter, extraViews, destTrs);
+  TransformMapAttr firstCoordTransform =
+      cast<TransformMapAttr>(toBeReducedViews[0]);
+  int64_t upperRank = firstCoordTransform.getUpperBounds().size();
+  SetVector<int64_t> removeIndicesSet;
+  // We only want to keep tid x iter in the maps
+  // which is the last two for block subtile
+  for (int64_t i = 0; i < upperRank - 2; i++) {
```
dhernandez0, line 1388: I guess rank is always at least 2? tid, iter?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1097,6 +1098,440 @@ MemcpyRewritePattern::matchAndRewrite(memref::CopyOp copy,
...
+      removeUpperDims(rewriter, toBeReducedViews, removeIndicesSet);
+  if (failed(blockSubTileViews)) {
+    LLVM_DEBUG(llvm::dbgs() << "blockSubTileViews creation using "
+                               "removeUpperDims is unsuccesful.\n");
+    return failure();
+  }
+  // We only want to keep tid in the maps
+  // which is the last two for block subtile tid
+  removeIndicesSet.clear();
+  for (int64_t i = 0; i < upperRank; i++) {
```
dhernandez0: we can skip the loop and just add iter:
removeIndicesSet.insert(upperRank-1);
The rest of the indices are already inserted

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1097,6 +1098,440 @@ MemcpyRewritePattern::matchAndRewrite(memref::CopyOp copy,
...
+      removeUpperDims(rewriter, toBeReducedViews, removeIndicesSet);
+  if (failed(blockSubTileTidSliceViews)) {
+    LLVM_DEBUG(llvm::dbgs() << "blockSubTileTidSliceViews creation using "
+                               "removeUpperDims is unsuccesful.\n");
+    return failure();
+  }
+  // We only want to keep iter in the maps
+  // which is the last one.
+  removeIndicesSet.clear();
+  for (int64_t i = 0; i < upperRank - 1; i++) {
```
dhernandez0: same here, just remove upperRank-1 and add upperRank-2

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1097,6 +1098,440 @@ MemcpyRewritePattern::matchAndRewrite(memref::CopyOp copy,
...
+  if (failed(gridOnlyDims)) {
+    LLVM_DEBUG(
+        llvm::dbgs()
+        << "gridOnlyDims creation using removeUpperDims is unsuccesful.\n");
+    return failure();
+  }
+
+  FailureOr<llvm::SmallDenseMap<int64_t, SmallVector<SubDimInfo>>>
+      lowerSubDims =
+          getLowerSubDimensions(rewriter, toBeReducedViews, {0, 1, 2});
```
dhernandez0: should we iterate until upperRank-2 as in the previous code?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1097,6 +1098,440 @@ MemcpyRewritePattern::matchAndRewrite(memref::CopyOp copy,
...
+  blockReductionAxis =
+      blockSubTileTidSliceRank - 1 - blockReductionAxisFromLeft;
+  int64_t partialReductionsPerThread =
+      blockSubTileTidSliceShape[blockReductionAxis];
+  int64_t ldsWorkspaceSize = 1;
+  for (auto [idx, size] : llvm::enumerate(blockLowerShape)) {
+    if (idx == (size_t)blockReductionAxis) {
+      ldsWorkspaceSize *= partialReductionsPerThread;
+    } else {
+      ldsWorkspaceSize *= size;
```
dhernandez0, line 1476: we could assert that memory required is not bigger than GPU LDS here. Reuse LDS pass would do it anyway, but it might be easier to debug if we add an assert here if this issue comes up.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1097,6 +1098,440 @@ MemcpyRewritePattern::matchAndRewrite(memref::CopyOp copy,
...
+                                      src.getType().getElementType(),
+                                      gpu::AddressSpace::Workgroup);
+
+  rewriter.create<BlockwiseBroadcastReduceOp>(
+      loc, src, ldsWorkspace, broadcastReducedSrc,
+      /*extraOut=*/nullptr, rewriter.getIndexAttr(blockReductionAxis),
+      reduceOp.getReduceMethodAttr(), blockSubTileViews.value(),
+      blockSubTileTidSliceViews.value(), threadSubTileViews.value(),
+      /*extraViews=*/nullptr,
+      getBlockSize(reduceOp->getParentOfType<func::FuncOp>()).value());
```
dhernandez0, line 1499: could we get this internally in the BlockwiseBroadcastReduceOp lowering instead?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1097,6 +1098,440 @@ MemcpyRewritePattern::matchAndRewrite(memref::CopyOp copy,
...
+
+  ViewLikeOpInterface viewOp =
+      ldsWorkspace.getDefiningOp<ViewLikeOpInterface>();
+  rewriter.create<GpuDeallocOp>(loc, viewOp.getViewSource());
```
dhernandez0, line 1503: nit: viewOp.getViewSource() -> ldsWorkspace

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -2319,3 +2371,167 @@ mlir::rock::removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
...
+}
+
+FailureOr<llvm::SmallDenseMap<int64_t, SmallVector<SubDimInfo>>>
+mlir::rock::getLowerSubDimensions(OpBuilder &b, ArrayAttr transformAttrs,
+                                  ArrayRef<int64_t> dims) {
+  llvm::SmallDenseMap<int64_t, SmallVector<SubDimInfo>> subDimInfo;
+  TransformMapAttr topMap = cast<TransformMapAttr>(transformAttrs[0]);
+  for (int64_t dim : dims) {
+    // No point of tracing size 1 dimensions
+    if (topMap.getUpperBounds()[dim] == 1) {
```
dhernandez0, line 2392: could this cause issues if the code calling getLowerSubDimensions expects the dictionary to have this dim info?

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -2319,3 +2371,167 @@ mlir::rock::removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
...
+            }
+          }
+        } break;
+        case TransformType::Unmerge: {
+          SmallVector<int64_t> subDimStrides = getStrides(trAttr.getParams());
+          int64_t lowDim = trAttr.getLowerDims()[0];
+          for (size_t subDim = 0; subDim < trAttr.getParams().size();
+               subDim++) {
+            int64_t upperDim = trAttr.getUpperDims()[subDim];
+            if (currSubDimInfo.contains(upperDim)) {
```
dhernandez0, line 2484: I thought it currSubDimInfo.contains(upperDim) should always be true, when can this be false?

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -2319,3 +2371,167 @@ mlir::rock::removeUpperDims(OpBuilder &b, ArrayAttr transformAttrs,
...
+                                      << upperToLower[dim] << "\n");
+              nextSubDimInfo[upperToLower[dim]] = subDimInfo;
+            }
+          }
+        } break;
+        case TransformType::Merge: {
+          SmallVector<int64_t> subDimStrides = getStrides(trAttr.getParams());
+          for (auto [lowDim, subDimStride, param] : llvm::zip(
+                   trAttr.getLowerDims(), subDimStrides, trAttr.getParams())) {
+            int64_t upperDim = trAttr.getUpperDims()[0];
```
dhernandez0: nit: move out of the loop

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1097,6 +1098,440 @@ MemcpyRewritePattern::matchAndRewrite(memref::CopyOp copy,
...
+  // Create partial reduction views
+  ArrayAttr paddedReducedTrStack;
+  {
+    SmallVector<Attribute> transformAttrs;
+    ArrayRef<int64_t> blockTileShape = getLowerShape(blockSubTileViews.value());
+    SmallVector<SmallString<8>> names;
+    SmallVector<StringRef> nameRefs;
+    for (size_t i = 0; i < blockTileShape.size(); i++) {
```
dhernandez0: nit: llvm::enumerate()

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1097,6 +1098,440 @@ MemcpyRewritePattern::matchAndRewrite(memref::CopyOp copy,
...
+    threadwiseWriteOp.getSourceMutable().assign(broadcastReducedSrc);
+    LLVM_DEBUG(llvm::dbgs() << "transformAttrs = "
+                            << "\n";
+               llvm::interleaveComma(transformAttrs, llvm::dbgs());
+               llvm::dbgs() << "\n");
+    TypedValue<ShapedType> reduceOut = reduceOp.getOut();
+    reduceOut = cast<TypedValue<ShapedType>>(
+        applyViewsOnDest(rewriter, loc, reduceOut, transformAttrs));
+    threadwiseWriteOp.getDestMutable().assign(reduceOut);
+    threadwiseWriteOp.setStoreMethodAttr(stMethod);
```
dhernandez0: is there any case where we could get rid of atomics? maybe leave a TODO as a possible optimization if that's the case.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1097,6 +1098,440 @@ MemcpyRewritePattern::matchAndRewrite(memref::CopyOp copy,
...
+        toReducedView.passThrough({nameRefs[i]}, {i}, {nameRefs[i]});
+      }
+    }
+    transformAttrs.push_back(toReducedView.get());
+    ArrayAttr arrayTransformAttrs = rewriter.getArrayAttr(transformAttrs);
+    paddedReducedTrStack = prependUpperViews(
+        rewriter, blockSubTileViews.value(), arrayTransformAttrs);
+  }
+
+  // Recombine block dimensions
```
dhernandez0, line 1532: could you explain a bit more here? 
nit: consider moving it to a function

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1097,6 +1098,440 @@ MemcpyRewritePattern::matchAndRewrite(memref::CopyOp copy,
...
+                 llvm::dbgs() << "\n"; llvm::dbgs() << "upperBounds = ";
+                 llvm::interleaveComma(upperBounds, llvm::dbgs());
+                 llvm::dbgs() << "\n"; llvm::dbgs() << "lowerBounds = ";
+                 llvm::interleaveComma(lowerBounds, llvm::dbgs());
+                 llvm::dbgs() << "\n");
+      TransformMapAttr newTrMap =
+          TransformMapAttr::get(trAttrs, upperBounds, lowerBounds);
+      transformAttrs.push_back(newTrMap);
+    }
+    ArrayRef<int64_t> currLowerShape =
```
dhernandez0: rock::getLowerShape

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1097,6 +1098,440 @@ MemcpyRewritePattern::matchAndRewrite(memref::CopyOp copy,
...
+      SmallVector<StringRef> nameRefs;
+      for (size_t i = 0; i < currLowerShape.size(); i++) {
+        SmallString<8> dimName(Twine("d" + Twine(i)).str());
+        names.push_back(dimName);
+        nameRefs.push_back(names.back());
+      }
+      TopDownTMBuilder toAddMissingBlockDims(rewriter, nameRefs,
+                                             currLowerShape);
+      {
+        toAddMissingBlockDims.passThrough({0, 1, 2}, {0, 1, 2});
```
dhernandez0: could we do 0 to upperRank - 2?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1097,6 +1098,440 @@ MemcpyRewritePattern::matchAndRewrite(memref::CopyOp copy,
...
+    llvm::SmallDenseMap<int64_t, SmallVector<int64_t>> blockSubDims;
+    TransformMapAttr lastMerge;
+    {
+      llvm::SmallDenseMap<int64_t, SmallVector<SmallString<8>>> names;
+      llvm::SmallDenseMap<int64_t, SmallVector<StringRef>> nameRefs;
+      llvm::SmallDenseMap<int64_t, SmallVector<unsigned>> upperDims;
+      int64_t dimInsertionPoint = 0;
+      for (unsigned dim = 0; dim < toBeReducedShape.size(); dim++) {
+        // The lower subDims contain sub-dimensions where blocking
+        // indices -- namesly g_block, m_block and n_block -- maps to
```
dhernandez0: namesly -> namely

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1097,6 +1098,440 @@ MemcpyRewritePattern::matchAndRewrite(memref::CopyOp copy,
...
+                   llvm::interleaveComma(splitSizes, llvm::dbgs());
+                   llvm::dbgs() << "\n");
+        toMatrixView.unmerge(nameRefs[dim], upperDims[dim],
+                             reduceLowerShapeNameRefs[dim], splitSizes);
+      }
+      lastMerge = toMatrixView.get();
+    }
+    LLVM_DEBUG(llvm::dbgs() << "lastMerge=" << lastMerge << "\n");
+    // The above view contains splitted sub-dims that are either associated
+    // with grid and non-grid dimensions. Then, we concat them into 6
```
dhernandez0: "6 dimensions", I think 6 is not hardcoded here

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -211,12 +215,41 @@ PopulateParams::paramsProbablyValid(OpBuilder &b,
   return populateDerived(params);
 }
 
+static LogicalResult couldFusedReductionBePerformant(const GemmSize &gemmSize,
+                                                     int64_t mPerBlock,
+                                                     int64_t nPerBlock) {
+  // 16 is practically lowest m in MFMAs/WMMAs
+  // that could be performant. If the gemm sizes
+  // are not divisible by that, then we definitely
+  // need padding. Therefore, it cant use blockwise
```
dhernandez0: cant -> can't

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -211,12 +215,41 @@ PopulateParams::paramsProbablyValid(OpBuilder &b,
...
+  // reductions. Thus, it becomes a competition among
+  // atomic_store based reduction kernels.
+  if (gemmSize.m % 16 != 0) {
+    return success();
```
dhernandez0, line 231: shouldn't this be failure()?

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -211,12 +215,41 @@ PopulateParams::paramsProbablyValid(OpBuilder &b,
...
+  }
+  if (gemmSize.n % 16 != 0) {
+    return success();
```
dhernandez0, line 234: same here

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -211,12 +215,41 @@ PopulateParams::paramsProbablyValid(OpBuilder &b,
   return populateDerived(params);
 }
 
+static LogicalResult couldFusedReductionBePerformant(const GemmSize &gemmSize,
```
dhernandez0, line 218: nit: I'd say this is not a failure/success function but a bool type

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -364,7 +364,7 @@ void createQuickTuningRange(TuningParamSet *newSpace,
              tuningInfo.getTuningParameters(info.kernelType, info.gemmAType,
                                             info.gemmBType, info.arch),
              info.gemmSize)) {
-      if (succeeded(tuningInfo.paramsProbablyValid(b, info, param)))
+      if (succeeded(tuningInfo.couldBePerformant(info, param)))
```
dhernandez0: should we check if params are valid anyway?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: yes... they must be there.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: a better solution might be we add that to the verifier of `rock.gpu_alloc` ?

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
```diff
@@ -0,0 +1,594 @@
...
+// This function will attempt to shuffle M & N dimensions of the gemm so that
+// reductions sub-dimensions within it could be split to blocks equally.
+// However, for that to work the transform stack needs to be invertible and
+// sub-dimension should be discoverable using "getLowerSubDimensions". If one of
+// those fail, we bail and not attempt to use blockwise_reductions in such
+// fusions.
+static void
+rearrangeGemmParallelDimsForReduction(ReduceOp rOp,
+                                      const BufferDependencyAnalysis &deps) {
+  FailureOr<std::tuple<ArrayAttr, Operation *>> res =
```
dhernandez0: nit: res -> reduceInfo? or some other more descriptive variable name

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: yep a good suggestion.. but I d do that seperately as a NFC change for bisectability reasons.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: no right... ldsWorkspace is view op with the right element type.. here we want the raw i8 memref.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: I dont need the element from blockTileShape. 
I just need the idx. 
enumerate will provide both here.

manupak: I ll remove the whole loop as I ve created a utility to do this now.

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
```diff
@@ -0,0 +1,594 @@
...
+    }
+    IRRewriter rewriter(rOp.getContext());
+    ArrayAttr invertedViews = invertTransforms(rewriter, rOp.getLoc(), views);
+    LLVM_DEBUG(llvm::dbgs()
+               << "inv(gemmToReduceViews)=" << invertedViews << "\n");
+    if (!invertedViews || invertedViews.empty()) {
+      return;
+    }
+    FailureOr<llvm::SmallDenseMap<int64_t, SmallVector<mlir::rock::SubDimInfo>>>
+        subDimensions = getLowerSubDimensions(rewriter, invertedViews, 2);
```
dhernandez0: use a variable name for dimension to make it clear what it is

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
```diff
@@ -0,0 +1,594 @@
...
+  return {additionalViewsA, additionalViewsB};
+}
+
+ArrayAttr generateShuffledGemmOutputViews(
+    OpBuilder &builder, int64_t g, int64_t m, int64_t mPerBlock, int64_t n,
+    int64_t nPerBlock,
+    llvm::SmallDenseMap<int64_t, SmallVector<SubDimInfo>> reductionSubDims) {
+  // Split the reduction and non-reduction splits
+  int64_t totalReductionSizeM = 1;
+  for (const SubDimInfo &sdInfo : reductionSubDims[1]) {
```
dhernandez0: we should not assume reductionSubDims contains 1 or 2

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
```diff
@@ -0,0 +1,594 @@
...
+    for (Attribute trMap : additionalViewsB) {
+      trGemmInB = rewriter.create<TransformOp>(rOp.getLoc(), trGemmInB,
+                                               cast<TransformMapAttr>(trMap));
+    }
+    ArrayAttr additionalOutputViews = generateShuffledGemmOutputViews(
+        rewriter, g, m, mnPerBlock.value().MPerBlock, n,
+        mnPerBlock.value().NPerBlock, subDimensions.value());
+    rewriter.setInsertionPointAfterValue(gemmOut);
+    Value trGemmOut = gemmOut;
+    Value firstUse = nullptr;
```
dhernandez0: I don't think this is used anywhere

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
manupak: we will not have vectors for rock.reduce op as elements.

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
dhernandez0: also, should we also consider G? for batch norm?

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
dhernandez0: also, use reduction dim from reduceop

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
manupak: caller cant have a such an expectation.
it is an information provided by this util whether it contains or not.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
manupak: as I said before, not all dimensions need to contain the interested subdimensions that we are tracing.

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
```diff
@@ -0,0 +1,594 @@
...
+  auto toSplitOriginalSubDims =
+      BottomUpTMBuilder::above(toCombinedReductionDim, combinedReduction);
+  int64_t nSubDimStartPoint = -1;
+  {
+    toSplitOriginalSubDims.passThrough("G");
+    SmallVector<SubDimInfo> mReductionSubDimInfo = reductionSubDims[1];
+    SmallVector<SubDimInfo> nReductionSubDimInfo = reductionSubDims[2];
+
+    unsigned dimInsertionPoint = 1;
+    {
```
dhernandez0, line 347: instead of doing it for m, then n. We could make it generic and iterate over m and n (and g if we consider doing that as well).

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
manupak: im being consistent with "couldBePerformant" API.

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
```diff
@@ -0,0 +1,594 @@
...
+  // Recombine into original M & N
+  auto toRecombineMN =
+      BottomUpTMBuilder::above(toSplitOriginalSubDims, splitOriginalSubDims);
+  {
+    toRecombineMN.passThrough("G");
+    SmallVector<StringRef, 4> startNames;
+    toRecombineMN.getStartNames(startNames);
+
+    // M
+    {
```
dhernandez0, line 467: nit: same here, make it generic and iterate over dims?

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
```diff
@@ -0,0 +1,594 @@
...
+
+std::tuple<ArrayAttr, ArrayAttr> generateShuffledGemmInputViews(
+    OpBuilder &builder, int64_t g, int64_t m, int64_t mPerBlock, int64_t k,
+    int64_t n, int64_t nPerBlock,
+    llvm::SmallDenseMap<int64_t, SmallVector<SubDimInfo>> reductionSubDims) {
+  BottomUpTMBuilder toReductionSplitA(builder, {"G", "K", "M"}, {g, k, m});
+  ArrayAttr additionalViewsA = builder.getArrayAttr({});
+  if (!reductionSubDims[1].empty()) {
+    additionalViewsA = reorderReductionDims(
+        toReductionSplitA, reductionSubDims[1], "M", m, mPerBlock);
```
dhernandez0: reductionSubDims might not contain 1

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
```diff
@@ -0,0 +1,594 @@
...
+  }
+  BottomUpTMBuilder toReductionSplitB(builder, {"G", "K", "N"}, {g, k, n});
+  ArrayAttr additionalViewsB = builder.getArrayAttr({});
+  if (!reductionSubDims[2].empty()) {
+    additionalViewsB = reorderReductionDims(
+        toReductionSplitB, reductionSubDims[2], "N", n, nPerBlock);
```
dhernandez0: reductionSubDims might not contain 2

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
manupak: We only shuffle dimensions where we do tiling and G is not a tiling dimension.
I ll add a comment.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
manupak: no ... 
If the size is not divisible by 16 then it means all the configs will have gemm padding for the tile sizes we support.
Thus, the tuning is done among atomic based solutions.  Therefore, all the solutions must be tuned.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
manupak: see the reply above

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: no its not a ArrayAttr (yet) to use that

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: ended up adding here as we want this merged before release.

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
manupak: added more docs as discussed.

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
manupak: I prefer to keep this very specific to M & N not be general because it relies on things like MPerBlock and NPerBlock present in the gridwise_gemm_accel op.

Also, it would be helpful, to see them in IR which dims are being split explicitly as "M" or "N".

In principle I agree a general solution might work here, but I d delay that until this is mature enough. 
So that it allows me to inspect easily to handle bugs. WDYT ?

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1332,25 +1334,23 @@ static LogicalResult insertBlockwiseReduction(
...
+      SmallVector<StringRef> nameRefs = getStringRefsFor(names);
       TopDownTMBuilder toAddMissingBlockDims(rewriter, nameRefs,
                                              currLowerShape);
       {
-        toAddMissingBlockDims.passThrough({0, 1, 2}, {0, 1, 2});
+        SmallVector<unsigned> gridOnlyDimIdxs;
+        for (unsigned i = 0; i < upperRank - 2; i++) {
+          gridOnlyDimIdxs.push_back(i);
+        }
+        toAddMissingBlockDims.passThrough(gridOnlyDimIdxs, {0, 1, 2});
```
dhernandez0: toAddMissingBlockDims.passThrough(gridOnlyDimIdxs, gridOnlyDimIdxs);

manupak: my bad.

mlir/lib/Dialect/Rock/Transforms/ShuffleGemmForReductions.cpp
dhernandez0: Fair enough, let's keep it like this. In the future we could make it generic but still write M/N in the IR, hardcoding dimStr = {"m", "n"}.

dhernandez0 (approved):

mlir/test/fusion/pr-e2e/mixr-multi-reduce-mo-4d.mlir
```diff
@@ -0,0 +1,29 @@
...
+    %7 = migraphx.reshape %6 {dims = [2, 32, 40960, 1]} : <2x32x10x64x64xf32, 1310720x40960x4096x64x1> -> <2x32x40960x1xf32, 1310720x40960x1x1>
+    %8 = migraphx.reduce_sum %7 {axes = [2]} : <2x32x40960x1xf32, 1310720x40960x1x1> -> <2x32x1x1xf32, 32x1x1x1>
+    %9 = migraphx.reshape %8 {dims = [2, 32, 1, 1, 1]} : <2x32x1x1xf32, 32x1x1x1> -> <2x32x1x1x1xf32, 32x1x1x1x1>
+    %10 = migraphx.multibroadcast %0 {out_dyn_dims = [], out_lens = [2, 32, 10, 64, 64]} : <1xf32, 0> -> <2x32x10x64x64xf32, 0x0x0x0x0>
+    %11 = migraphx.mul %4, %4 : <2x32x10x64x64xf32, 1310720x40960x4096x64x1>, <2x32x10x64x64xf32, 1310720x40960x4096x64x1> -> <2x32x10x64x64xf32, 1310720x40960x4096x64x1>
+    %12 = migraphx.mul %11, %10 : <2x32x10x64x64xf32, 1310720x40960x4096x64x1>, <2x32x10x64x64xf32, 0x0x0x0x0> -> <2x32x10x64x64xf32, 1310720x40960x4096x64x1>
+    %13 = migraphx.reshape %12 {dims = [2, 32, 40960]} : <2x32x10x64x64xf32, 1310720x40960x4096x64x1> -> <2x32x40960xf32, 1310720x40960x1>
+    %14 = migraphx.reduce_sum %13 {axes = [2]} : <2x32x40960xf32, 1310720x40960x1> -> <2x32x1xf32, 32x1x1>
```
dhernandez0, line 25: nit: add tests where the axis of reduction is not 2. Also, tests where m/n is not reduced at all. This is to test the fixes introduced today.

manupak: Thanks. will add them.
I also need to fix/add unit test for align-tiling/rock-shuffle-gemm-for-reductions

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
manupak: I ve refactored recombine into a new function + docs.

manupak:
@dhernandez0 I ve added the promised tests now!

dhernandez0:
> @dhernandez0 I ve added the promised tests now!

Thanks! I was wondering if we should add a test where the reduction occurs along the G axis only for completeness. I understand it would only involve atomics, but it might be good to include it just in case. I've already approved the PR, so whatever you decide.

PR #1669, opened 2024-10-04
Don't construct Embed{}s for 1x1 filters in convolutions

krzysz00:
Fixes https://github.com/ROCm/rocMLIR-internal/issues/1642

The [dN] + [dN]o (that is, index in the filter (from the K dimension)
+ index in the output (from the N dimension)) embedding in our conv2gemm is needed.

However, in the case of 1x1 filters, that should simplify to just [dN]o (possibly scaled by the stride) that Embed{} operation was throwing off collapseContiguousMerges(), leading to inefficient codegen compared to the mathematically equivalent GEMM operation as described in the ticket.

This commit resolves the issue and adds tests.


manupak (approved):
LGTM!

PR #1670, opened 2024-10-04
[CI] External CI mainline build support

jayhawk-commits:


krzysz00 (approved):
Approved, untagging `skip-ci` since that's not "land without a CI pass" but "don't run CI, I'm still working on it"

PR #1672, opened 2024-10-08
Use rocprofv2 instead of rocprof.

pcf000:
Use rocprofv2 instead of rocprof.
Account for .MLIR_N_REPEATS in rocprofv2 results, which don't include it.
Account for nrepeats in a smarter way -- count the rows, while verifying.
getFusionTestInfo and runFusionKernel turn out to be mostly the same.
Invent --rocprof-version to switch between rocprof and rocprofv2.
Change default to rocprofv2.

### DIFF
--- a/mlir/utils/performance/perfRunner.py
+++ b/mlir/utils/performance/perfRunner.py
@@ -27,7 +27,7 @@ from perfCommonUtils import Operation, GEMMLibrary
 BENCHMARKINGV1_RESULT_FILE_NAME = 'results.stats.csv'
 BENCHMARKINGV1_METRICS_FILE_NAME = 'results.csv'
 ROCPROFV1 = '/opt/rocm/bin/rocprof'
-ROCPROFV1_OPTS = ['--stats', '-o', BENCHMARKINGV1_METRICS_FILE_NAME]
+ROCPROFV1_OPTS = ['--stats', '-o', BENCHMARKINGV1_RESULT_FILE_NAME]
 BENCHMARKINGV2_RESULT_FILE_NAME = './pmc_1/results_stats.csv'
 BENCHMARKINGV2_METRICS_FILE_NAME = './pmc_1/results_stats.csv'
 ROCPROFV2 = '/opt/rocm/bin/rocprofv2'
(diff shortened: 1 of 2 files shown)


mlir/utils/performance/perfRunner.py
```diff
@@ -23,11 +24,20 @@
 from perfCommonUtils import Operation, GEMMLibrary
 
 # global variables.
-ROCPROF = '/opt/rocm/bin/rocprof'
-MIOPENDRIVER = '/opt/rocm/bin/MIOpenDriver'
-BENCHMARKING_RESULT_FILE_NAME = 'results.stats.csv'
-BENCHMARKING_METRICS_FILE_NAME = 'results.csv'
+BENCHMARKINGV1_RESULT_FILE_NAME = 'results.stats.csv'
```
pcf000, line 27: Of course things move around for rocprofv2, and there's no way I found to make them the same.  In particular, that "pmc_1" directory inserts itself because of either --kernel-trace or the stats in -i, I forget which.

mlir/utils/performance/perfRunner.py
```diff
@@ -129,6 +139,9 @@ def create_paths(config_file_path, mlir_build_dir_path) -> Paths:
 
 # utility functions.
 def getNanoSeconds(fileName):
+    pass
```
pcf000: I'm going to assign V1 or V2 to getNanoSeconds.  I don't really need this "pass" implementation.

mlir/utils/performance/perfRunner.py
```diff
@@ -153,21 +197,46 @@ def getMetricArgsForRocprof(arch):
 # Bank conflict functions.The percentage of GPUTime LDS is stalled by bank
 # conflicts. Value range: 0% (optimal) to 100% (bad).
 def getBankConflict(fileName):
+    pass
+
+def getBankConflictV1(fileName):
     if not os.path.exists(fileName):
-        result = "NaN"
-        return result
+        return np.nan
```
pcf000: We had not been consistent and used "nan", "NaN", and np.nan in different places.

mlir/utils/performance/perfRunner.py
```diff
@@ -153,21 +197,46 @@ def getMetricArgsForRocprof(arch):
...
     with open(fileName, 'r') as csv_file:
         reader = csv.DictReader(csv_file, delimiter = ',')
         header = reader.fieldnames
         if 'LDSBankConflict' not in header:
             return np.nan
-
-        result = []
+        sum = 0
```
pcf000: Counting the rows and accumulating the sum, vs accumulating a list and then calling sum and len on it.

mlir/utils/performance/perfRunner.py
```diff
@@ -209,10 +278,10 @@ def getMilliseconds(output):
 
     return float(result.group(1))
 
-def runPipeline(proc_specs):
+def runPipeline(proc_specs, initial_stdin=subprocess.DEVNULL):
```
pcf000, line 312: "initial_stdin" exists to send some mlir text into the first stage, see below.

mlir/utils/performance/perfRunner.py
```diff
@@ -1048,50 +1117,7 @@ def findRunCommand(filename):
     print("WARNING: cannot find valid RUN command in ", filename)
     return None, None
 
-# Extract testVector and test function name from the test file
-def getFusionTestInfo(filename, paths: Paths):
```
pcf000, line 1052: I had missed getFusionTestInfo when I updated everything to use runPipeline.  When I started updating it, I noticed that everything up to the tuningKey process was identical with runFusionKernel, so I abstracted the common part into makeBasicFusionPipeline.

Then I realised that it was doing redundant work, because we'd collect tests and call getFusionTestInfo on them, then loop through the collected tests and call runFusionKernel on those, and that would do the basic pipeline twice.  I recast it to do the basics once and save the mlir, and merged the collection and running loops so it doesn't have to save the mlir for very long.  More notes below.

mlir/utils/performance/perfRunner.py
```diff
@@ -1108,28 +1134,27 @@ def runFusionKernel(filename, rocmlirGenArgs, paths: Paths):
     else:
         rocmlirDriverCommand = [paths.mlir_paths.rocmlir_driver_path, '-host-pipeline', 'highlevel', '-targets', getChip()]
         commands.append(rocmlirDriverCommand)
+    return commands
 
+def runFusionKernel(mlirfile, rocmlirGenArgs, paths: Paths):
```
pcf000, line 1173: Now takes as input a file of the mlir from the basic-fusion-pipeline.

mlir/utils/performance/perfRunner.py
```diff
@@ -1140,63 +1165,75 @@ def benchmarkFusionKernels(test_dir, paths: Paths, arch, numCU, tuningDb: MaybeT
...
-
-        print("Profiling:", filename)
-        # Sanity check
-        if not testVector:
-            print("\tCannot find a test vector")
-            continue
-        if not futName:
-            print("\tCannot find rocmlir-gen with -fut")
-            continue
+    # Prepare test cases
```
pcf000, line 1201: Merged the two loops into one, with the split-k hack moved up first.  Go through each .mlir file, extract the useful RUN: command if present, and do makeBasicFusionPipeline with it to make the initial mlir code.  We have a temp file to hold that code, because I couldn't make a good way to save it as a string and then send it to another pipeline later without having a file.

mlir/utils/performance/perfRunner.py
```diff
@@ -1140,63 +1165,75 @@ def benchmarkFusionKernels(test_dir, paths: Paths, arch, numCU, tuningDb: MaybeT
...
-                    perfResults[testVector] = oneEntry
+            commandLine = testVector.split(sep=' ')
+            if commandLine[0].startswith('conv'):
+                op = 'conv'
+                config = ConvConfiguration.fromCommandLine(commandLine, arch, numCU)
+            else:
+                op = 'gemm'
+                config = GemmConfiguration.fromCommandLine(commandLine, arch, numCU)
+
+            # Find the best perf_config
```
pcf000, line 1235: Look for the perf-config for the kernel configuration, or make a dummy NaN one.

mlir/utils/performance/perfRunner.py
```diff
@@ -1140,63 +1165,75 @@ def benchmarkFusionKernels(test_dir, paths: Paths, arch, numCU, tuningDb: MaybeT
...
+                else: # Tuning DB present but doesn't contain config, add a NaN entry
+                    if not testVector in perfResults:
+                        oneEntry = config.tableEntry(np.nan)
+                        oneEntry['MLIR TFlops'] = np.nan
+                        oneEntry['Fusion/MLIR'] = np.nan
+                        oneEntry['FileName'] = filename
+                        perfResults[testVector] = oneEntry
+                    continue
+
+            # Run fusion test
```
pcf000, line 1251: Run the kernel, reusing the mlir in the temp file, and record its time.  We anticipate duplicates -- eg, in the bert tests there are 24 .mlir file but only eight unique kernels -- and just take the best-performing.  Given natural variation, times will be close but the winning file can be different every time.

mlir/utils/performance/perfRunner.py
```diff
@@ -1140,63 +1165,75 @@ def benchmarkFusionKernels(test_dir, paths: Paths, arch, numCU, tuningDb: MaybeT
...
-
-        # Run gemm or conv op with the same configuration
-        runConfigWithMLIR(config, paths, arch, '')
-        # Get nanoseconds of gemm/conv
-        nanoSeconds = getNanoSeconds(BENCHMARKING_RESULT_FILE_NAME)
-        oneEntry['MLIR TFlops'] = config.computeTFlops(nanoSeconds)
-        oneEntry['Fusion/MLIR'] = oneEntry['TFlops']/oneEntry['MLIR TFlops']
-        oneEntry['FileName'] = filename
-        perfResults[testVector] = oneEntry
+            # Run gemm or conv op with the same configuration
```
pcf000, line 1262: Run generated kernel in the usual way for reference.

mlir/utils/performance/perfRunner.py
```diff
@@ -1220,26 +1257,12 @@ def tuneMLIRKernels(configs, arch, numCU):
...
-                _, errs = p1.communicate()
+            runPipeline([MIOpenDriverCommand])
 
 def is_xdlops_present() -> bool:
     """This function checks whether a GPU with xdlops support is present"""
     xdlop_supported_gpus = ['gfx908', 'gfx90a', 'gfx942']
-    xdlop_supported_gpus_str = xdlop_supported_gpus[0]
-    for gpu in xdlop_supported_gpus[1:]:
-        xdlop_supported_gpus_str += '|' + gpu
+    xdlop_supported_gpus_str = '|'.join(xdlop_supported_gpus)
```
pcf000, line 1298: Idiom.

mlir/utils/performance/perfRunner.py
```diff
@@ -1220,26 +1257,12 @@ def tuneMLIRKernels(configs, arch, numCU):
...
-            # get output.
-            try:
-                _, errs = p1.communicate(timeout=300)
-                if len(errs) > 0 and p1.returncode != 0:
-                    raise OSError(errs.decode('utf-8'))
-            except subprocess.TimeoutExpired:
-                p1.kill()
-                print("MIOpen tuning timed out")
-                _, errs = p1.communicate()
+            runPipeline([MIOpenDriverCommand])
```
pcf000, line 1293: Missed another one.  Last one, I swear.

mlir/utils/performance/perfRunner.py
```diff
@@ -1285,7 +1308,7 @@ def getNumCU(chip):
         rocminfo = subprocess.check_output("/opt/rocm/bin/rocminfo",
                                            stderr=subprocess.PIPE)
     except subprocess.CalledProcessError as e:
-        print(e.stderr.decode('utf-8'))
+        print(f"Process error: {e.stderr.decode('utf-8')}")
```
pcf000, line 1344: Still trying to identify the cause of the intermittent rocminfo failures.  Highlight this case in the log file.

mlir/utils/performance/perfRunner.py
```diff
@@ -1447,8 +1470,32 @@ def main(args=None):
          help='Force a set of datatypes'
     )
 
+    parser.add_argument("--rv", "--rocprof-version", choices=['1', '2'],
+        default='2',
+        help="Use rocprof (V1) or rocprofv2.")
+
     parsed_args = parser.parse_args(args)
 
+    global getNanoSeconds, getBankConflict, ROCPROF, ROCPROF_OPTS
```
pcf000: Swap functions, options, and filenames based on the option.

mlir/utils/performance/perfRunner.py
```diff
@@ -894,6 +964,7 @@ def benchmarkExternal(cls, commandLine, paths: Paths, arch, numCU):
         benchmarkArgs = config.generateMlirDriverCommandLine("")
         # remove the result file generated by rocprof in previous benchmarking
         os.system("rm -f "+BENCHMARKING_RESULT_FILE_NAME)
+        os.system("rm -f "+BENCHMARKING_METRICS_FILE_NAME)
```
krzysz00, line 998: We can call `rm` just once

mlir/utils/performance/perfRunner.py
krzysz00: ... why are we keeping rocprof v1 support?

krzysz00 (commented):
(Haven't fully read the PR, minor thoughts)

pcf000:
Keeping rocprof V1 in case we have discrepancies to investigate (though since we don't keep much history, that's probably not a big concern) and in case rocprofv2 stops supporting an architecture before we do.  Mostly because it was pretty easy to do and might be helpful.

mlir/utils/performance/perfRunner.py
```diff
@@ -137,9 +150,40 @@ def getNanoSeconds(fileName):
         result = 0
         for row in reader:
             result += int(row['AverageNs'])
-        csv_file.close()
         return result
 
+def getNanoSecondsV2(fileName):
```
krzysz00: Overall comment: perhaps `class Profiler:` is in order here, instead of hot-swapping functions onto variables?

causten:
Moved to v3

PR #1674, opened 2024-10-08
Fix alignment constraints not being correctly imposed in certain vect…

krzysz00:
…orizations

The test this commit adds shows the issu, which is that if, for example, you're vectorizingEmbed{2, 1}(...@4, align(1), 4@1 align(4)), the resulting vectorization should be 4@1 align(2), not align(4). This matters for determining how much a future Pad{} call reduces the vectorization by.

The logic I replaced came from a similar incident (fixed in PR #1199 - the need for alignment tracking in the first place was PR #772, a bug somewhat like this one). The issue was that the post-vectorization alignment fixup code assumed that all inputs to the Embed{} that had vectorizaiton data already had their alignment data incorporated, which isn't the case in the example (we'd derive 4@1 align(4) and then stop because 4 * 1 != 2, without incorporating the factor of 2 * 1 that needs to be applied to the alignment). This change refines the earlier logic by explicitly tracking which vectorization candidates have had their alignments looked at.

(The real-world case that the test was simplified from is, in relevant part
Max vectorization data: 1@8 align(1) ?@? align(?) 1@12 align(1) 1@12 align(1) 1@8 align(1) 1@4 align(1) 4@1 align(4)
Processing: #rock.transform_map<affine_map<(d0, d1, d2, d3, d4, d5, d6) -> (d0, d1, d2, d4, d5 * 2 + d6)> by [<PassThrough ["ni", "gi", "ci"] at [0, 1, 2] -> ["ni", "gi", "ci"] at [0, 1, 2]>, <AddDim{1} ["0"] at [3] -> [] at []>, <PassThrough ["0o"] at [4] -> ["0ipad"] at [3]>, <Embed{2, 1} ["1", "1o"] at [5, 6] -> ["1ipad"] at [4]>] bounds = [1, 1, 1, 1, 1, 3, 8] -> [1, 1, 1, 1, 12]>
Max vectorization data: 1@8 align(1) ?@? align(?) 1@12 align(1) 1@8 align(1) 4@1 align(4) )

Fixes https://github.com/ROCm/rocMLIR-internal/issues/1643

### DIFF
--- a/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
+++ b/mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
@@ -666,21 +666,18 @@ propagateVectorizationInfo(TransformMapAttr map, const VectorizationData &input,
       // of 1
       bool hasNegativeCoefficients = false;
       std::optional<VectorizationInfo> ourResult;
-
-      // When there are equal embed coefficients
-      // we should always refer to the one with lowest
-      // alignment as the alignment
-      llvm::SmallDenseMap<int64_t, int64_t> coeffToAlignment;
-      for (auto [coefficient, upperDim] : data) {
-        if (input[upperDim].has_value()) {
-          if (coeffToAlignment.count(coefficient)) {
-            coeffToAlignment[coefficient] = math_util::gcd(
-                coeffToAlignment[coefficient], input[upperDim]->alignment);
-          } else {
-            coeffToAlignment[coefficient] = input[upperDim]->alignment;
-          }
-        }
-      }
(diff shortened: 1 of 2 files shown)


krzysz00:
Have re-tested the original failure case, this does fix the issue

pcf000:
Looks reasonable from here, but another reviewer would be good.

manupak (approved):
Ah ha! 

right, we irrespective of coefficient being same -- we need to gcd all of coefficient * alignment that gets embed'd into the single dimension. 

Thanks for catching this! approved.

PR #1675, opened 2024-10-09
Separate quick-tuning lists by conv and gemm

djramic:
This commit separates the quick-tuning lists by conv and gemm operations. We are doing this to reduce the number of configs for quick-tuning, and consequently, the tuning time.
Currently, it is implemented only for xdl-ops. If we agree that it make sense to proceed with this approach I will add it for WMMA as well.

closes : https://github.com/ROCm/rocMLIR-internal/issues/1641

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
+++ b/mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
@@ -385,19 +385,32 @@ protected:
 // Xdlops interface
 //
 class PopulateParamsXDL : public PopulateParamsAccel {
-  static constexpr size_t nInitParameters = 40;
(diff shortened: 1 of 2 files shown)


manupak (changes requested):
The perf drop is too steep.

As we discussed offline, we need anwsers to following : 
1) Why 20 configs ?
2) If 20 configs, why did we decide to drop what we have dropped here? 

I d propose following steps : 
1) Lets find a number between 20 - 40 that does not result in a perf drop.
2) In the absence of better understanding what Ethan's scripts are doing, we should at least brute force dropping one perf config at a time from the above 40 (or less) configs and decide on dropping of which configs can do the least damage. If we can do this, then we can sort them based on this order.

djramic:
Outdated, since there is a new PR: #1689

PR #1677, opened 2024-10-09
Fix not enabling fp8 WMMA on Navi4 by default

krzysz00:
getDefaultFeatures() in the architecture database hadn't been told that sometimes there's a fp8 WMMA, so it wasn't sending those operations down that path.

This commit fixes the issue and tests that enablement.


stefankoncarevic (approved):
Thanks for resolving this. Approved.

PR #1666, opened 2024-09-30
Use target attached on gpu module when lowering

umangyadav:
Moves `rocdl-attach-target` before `convert-gpu-to-rocdl` pass.  `rocdl-attach-target` would add target attributes on gpu module which are then read during `convert-gpu-to-rocdl`

Solves Task 1 of https://github.com/ROCm/rocMLIR-internal/issues/1534

PR #1679, opened 2024-10-09
Use convert-to-llvm pass for conversions to LLVMIR Dialect 

umangyadav:
This is more of a refactor and WIP. Would require some discussions with upstream.
Depends on https://github.com/ROCm/rocMLIR/pull/1666

cc: @fabianmcg


external/llvm-project/mlir/lib/Conversion/ConvertToLLVM/ConvertToLLVMPass.cpp
```diff
@@ -6,12 +6,28 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "mlir/Analysis/DataLayoutAnalysis.h"
 #include "mlir/Conversion/ConvertToLLVM/ToLLVMInterface.h"
 #include "mlir/Conversion/ConvertToLLVM/ToLLVMPass.h"
+#include "mlir/Conversion/GPUCommon/GPUCommonPass.h"
```
krzysz00: You can remove this

krzysz00 (commented):
Overall, approved

However, I'm not checking the box because I think this set of changes should be sent upstream first and then backported to us

external/llvm-project/mlir/lib/Conversion/ConvertToLLVM/ConvertToLLVMPass.cpp
umangyadav: Done.

external/llvm-project/mlir/lib/Conversion/ConvertToLLVM/ConvertToLLVMPass.cpp
```diff
@@ -69,31 +75,102 @@ class ConvertToLLVMPass
...
+    }
+
     RewritePatternSet tempPatterns(context);
     auto target = std::make_shared<ConversionTarget>(*context);
     target->addLegalDialect<LLVM::LLVMDialect>();
-    auto typeConverter = std::make_shared<LLVMTypeConverter>(context);
+    auto typeConverter = std::make_shared<LLVMTypeConverter>(context, options);
+
+    DenseMap<Attribute, uint64_t> addressSpaceMap;
+    if (DataLayoutOpInterface iface = dyn_cast<DataLayoutOpInterface>(op)) {
```
krzysz00, line 129: I think this is another usage of `getAtOrAbove()`

krzysz00 (commented):
Re the utility function I was considering, now that I've looked at the structure of the code, I'm seeing another approach, one which'll make `ConvertToLLVM` less special.

Specifically, take that `addTypeAttributeConversion()` call and move it into the constructor for `LLVMTypeConverter` ... and have it take the `Attribute -> uint32_t` map as an element of `LowerToLLVMOptions` .

(While doing that, I'd go ahead and extend the `DataLayout` class with a `getEntries()` sort of iterator so you can loop through all the entries in that cache)

That way, anyone constructing their `LowerToLLVMOptions()` with a `DataLayoutAnalysis` automatically gets this functionality.

external/llvm-project/mlir/lib/Conversion/ConvertToLLVM/ConvertToLLVMPass.cpp
krzysz00: ... no, it isn't, but it could be, if `DataLayout` (the class) grew a method.

umangyadav:
CLosing this one for now since i am not going to work on this for a while

PR #1682, opened 2024-10-15
Fix int4 loads to be vector typed always

manupak:
Currently, when we load int4, we rely on the
views + input for it be vectorizable to be
barely functional. However, this assumption
will not hold when we have breaking views that breaks
the vectorization.

Therefore, this commit introduces code to load
2xi4 always and select the right element to mimic
single i4 element load.

closes: https://github.com/ROCm/rocMLIR-internal/issues/1645

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
@@ -1097,6 +1097,9 @@ static Value zeroDMemrefAsOneD(PatternRewriter &b, Value memref) {
                                                memref, reassociationRef);
 }
 
+// A helper to obtain coordinates of a global load and loaded type
+// Note if there is scalar i4 load, we still would load vector<2xi4>
+// element and select later on.
 std::tuple<SmallVector<Value>, Type> getCoordsAndType(PatternRewriter &b,
                                                       GlobalLoadOp op) {
   MemRefType srcType = op.getSource().getType();
@@ -1111,22 +1114,25 @@ std::tuple<SmallVector<Value>, Type> getCoordsAndType(PatternRewriter &b,
   if (srcType.getElementType().getIntOrFloatBitWidth() >= 8 ||
       originalLoadVecLen != 1) {
     return {coords, originalLoadedType};
-  } else if (srcType.getElementType().getIntOrFloatBitWidth() == 4) {
+  } else {
+    assert(srcType.getElementType().getIntOrFloatBitWidth() == 4 &&
+           "we only support 4bits in narrow types");
     ArrayRef<int64_t> shape = srcType.getShape();
     Value flatAddress = flattenCoords(b, loc, coords, shape);
     Type coordType = flatAddress.getType();
+    Value minusOne = getConstIntOrIndexValue(b, loc, -1, coordType);
     Value one = getConstIntOrIndexValue(b, loc, 1, coordType);
-    flatAddress = b.createOrFold<arith::ShRUIOp>(loc, flatAddress, one);
-    flatAddress = b.createOrFold<arith::ShLIOp>(loc, flatAddress, one);
+    Value lsbMask = b.createOrFold<arith::XOrIOp>(loc, minusOne, one);
+    flatAddress = b.createOrFold<arith::AndIOp>(loc, flatAddress, lsbMask);
     SmallVector<Value> newCoords;
     unflattenCoords(b, loc, flatAddress, shape, newCoords);
     Type loadedType = VectorType::get({2}, srcType.getElementType());
     return {newCoords, loadedType};
-  } else {
-    llvm_unreachable("less than 4 bit element types are not implemented");
   }
 }
 
+// A helper to select the right i4 element if it was supposed to
+// be a scalar i4 load.
 Value selectDataIf4b(PatternRewriter &b, GlobalLoadOp op, Value loadedVec) {
   MemRefType srcType = op.getSource().getType();
   Type originalLoadedType = op.getResult().getType();
@@ -1140,7 +1146,9 @@ Value selectDataIf4b(PatternRewriter &b, GlobalLoadOp op, Value loadedVec) {
   }
   if (originalLoadVecLen != 1) {
     return loadedVec;
-  } else if (srcType.getElementType().getIntOrFloatBitWidth() == 4) {
+  } else {
+    assert(srcType.getElementType().getIntOrFloatBitWidth() == 4 &&
+           "we only support 4bits in narrow types");
     assert(isa<VectorType>(loadedVec.getType()));
     Location loc = op.getLoc();
     SmallVector<Value, 5> coords(op.getSourceCoord());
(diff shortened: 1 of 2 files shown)


mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
```diff
@@ -1085,6 +1097,64 @@ static Value zeroDMemrefAsOneD(PatternRewriter &b, Value memref) {
...
+  if (srcType.getElementType().getIntOrFloatBitWidth() >= 8 ||
+      originalLoadVecLen != 1) {
+    return {coords, originalLoadedType};
+  } else if (srcType.getElementType().getIntOrFloatBitWidth() == 4) {
+    ArrayRef<int64_t> shape = srcType.getShape();
+    Value flatAddress = flattenCoords(b, loc, coords, shape);
+    Type coordType = flatAddress.getType();
+    Value one = getConstIntOrIndexValue(b, loc, 1, coordType);
+    flatAddress = b.createOrFold<arith::ShRUIOp>(loc, flatAddress, one);
+    flatAddress = b.createOrFold<arith::ShLIOp>(loc, flatAddress, one);
```
dhernandez0: nit: looks like you shift right and then left? can't we do a "bitwise and" operation?

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
```diff
@@ -1085,6 +1097,64 @@ static Value zeroDMemrefAsOneD(PatternRewriter &b, Value memref) {
...
+    return loadedVec;
+  }
+  int64_t originalLoadVecLen = 1;
+  if (VectorType originalLoadVecType =
+          dyn_cast_if_present<VectorType>(originalLoadedType)) {
+    originalLoadVecLen = originalLoadVecType.getNumElements();
+  }
+  if (originalLoadVecLen != 1) {
+    return loadedVec;
+  } else if (srcType.getElementType().getIntOrFloatBitWidth() == 4) {
```
dhernandez0: "else if (srcType.getElementType().getIntOrFloatBitWidth() == 4)" -> "assert(srcType.getElementType().getIntOrFloatBitWidth() == 4)" and remove last else

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
```diff
@@ -1085,6 +1097,64 @@ static Value zeroDMemrefAsOneD(PatternRewriter &b, Value memref) {
...
+  Type originalLoadedType = op.getResult().getType();
+  int64_t originalLoadVecLen = 1;
+  if (VectorType originalLoadVecType =
+          dyn_cast<VectorType>(originalLoadedType)) {
+    originalLoadVecLen = originalLoadVecType.getNumElements();
+  }
+  if (srcType.getElementType().getIntOrFloatBitWidth() >= 8 ||
+      originalLoadVecLen != 1) {
+    return {coords, originalLoadedType};
+  } else if (srcType.getElementType().getIntOrFloatBitWidth() == 4) {
```
dhernandez0: "else if (srcType.getElementType().getIntOrFloatBitWidth() == 4)" -> "assert(srcType.getElementType().getIntOrFloatBitWidth() == 4)" and remove last else

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
```diff
@@ -1085,6 +1097,64 @@ static Value zeroDMemrefAsOneD(PatternRewriter &b, Value memref) {
...
+    return loadedVec;
+  } else if (srcType.getElementType().getIntOrFloatBitWidth() == 4) {
+    assert(isa<VectorType>(loadedVec.getType()));
+    Location loc = op.getLoc();
+    SmallVector<Value, 5> coords(op.getSourceCoord());
+    ArrayRef<int64_t> shape = srcType.getShape();
+    Value flatAddress = flattenCoords(b, loc, coords, shape);
+    Type coordType = flatAddress.getType();
+    Value one = getConstIntOrIndexValue(b, loc, 1, coordType);
+    Value lsb = b.createOrFold<arith::AndIOp>(loc, flatAddress, one);
```
dhernandez0: nit: In getCoordsAndType we already compute "flattenCoords". We could compute lsb in getCoordsAndType and pass it here to save some instructions?

dhernandez0:
would this affect GlobalStore as well? or LDS load/store?

manupak:
> would this affect GlobalStore as well? or LDS load/store?

Just global load for now..

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
manupak: not really me thinks...
So the instruction that d remain after folding would be two `and` ops.

1) Above: i ve changed the above (I ll push them next) to use lsbMask which is not the same as exracting the lsb. (e.g. : ... 1110)
2) Below : here it would be just extracting lsb (e.g 0001)

So at least there need to be two instructions.

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
```diff
@@ -1085,6 +1097,70 @@ static Value zeroDMemrefAsOneD(PatternRewriter &b, Value memref) {
...
+  Type originalLoadedType = op.getResult().getType();
+  int64_t originalLoadVecLen = 1;
+  if (VectorType originalLoadVecType =
+          dyn_cast<VectorType>(originalLoadedType)) {
+    originalLoadVecLen = originalLoadVecType.getNumElements();
+  }
+  if (srcType.getElementType().getIntOrFloatBitWidth() >= 8 ||
+      originalLoadVecLen != 1) {
+    return {coords, originalLoadedType};
+  } else {
```
dhernandez0: nit: no need for else here, if returns

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
```diff
@@ -1085,6 +1097,70 @@ static Value zeroDMemrefAsOneD(PatternRewriter &b, Value memref) {
...
+    return loadedVec;
+  }
+  int64_t originalLoadVecLen = 1;
+  if (VectorType originalLoadVecType =
+          dyn_cast_if_present<VectorType>(originalLoadedType)) {
+    originalLoadVecLen = originalLoadVecType.getNumElements();
+  }
+  if (originalLoadVecLen != 1) {
+    return loadedVec;
+  } else {
```
dhernandez0: nit: same here

mlir/lib/Dialect/Rock/Transforms/SugarToLoops.cpp
dhernandez0: Nice, thanks! I wasn't aware of how folding works, completely ignored the fold part of createOrFold :D

dhernandez0 (approved):

krzysz00 (commented):
This shouldn't be needed - the selection logic is already implemented in the int4 emulation code. Why did we have to add this?

manupak:
You can try running the test case in the PR and also see the discussion in the issue.
From what I can tell is that if maxVecLen for i4 reads are < 2, then it just fails. You were going with assumption i4's are always vectorizable. yes?

PR #1683, opened 2024-10-16
Fix gated threadwise_write_all

dhernandez0:
The scf.if condition was using the wrong validity domain when threadwise_write_all writes to non-global memory.


manupak (approved):
Thanks for fixing this!
nit: add a newline at the end of the test.

PR #1684, opened 2024-10-16
navi4x tests fail with mixed types bf8_fp8

dhernandez0:
Fix when wmma is enabled by rocmlir-gen (and ConvGenerator) for mixed types.


dhernandez0:
should we have a function in AmdArchDb.cpp similar to getDefaultFeatures that takes the two types?

manupak (approved):
LGTM!

PR #1686, opened 2024-10-17
Collected small Jenkinsfile improvements

pcf000:
1.  catchError doesn't catch timeout exceptions, but try/catch does.  Use try/catch in the codecov stage.  This is a precursor to using "retry" to automatically catch and retry after infrastructure issues.
2.  check-rocmlir-build-only with profiling on does make some *.profraw files.  Delete them before check-rocmlir.
3.  Use df to check for enough disk space before running codecov, because it'll generate around 125 GB of *.profraw files.
4.  Abstract the bulk of the codecov shell script into a function.


krzysz00 (approved):
Approved, thanks!

PR #1687, opened 2024-10-18
Fix hardcoded arguments and results ids for prefill

dhernandez0:
In this PR we trace back the kernel argument (or return result) instead of using hardcoded indices for the prefill flag. This is modified in TosaToRock for reduce ops and in GemmToGridwise for splitk.

Additionally, rocmlir-gen will fill output tensors without "prefill" flag to 100 for the kernel launcher. This is to make sure non-prefill outputs are init to different values for the gpu vs cpu. Therefore, if an output is incorrectly not set to prefill we would fail.

### DIFF
--- a/mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
+++ b/mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
@@ -215,73 +215,37 @@ makeRockConv(ConversionPatternRewriter &rw, Operation *op, Value input,
   return cop;
 }
 
-static bool isTosaConv(Operation *op) {
-  return isa<tosa::Conv2DOp, tosa::Conv3DOp, tosa::DepthwiseConv2DOp,
-             tosa::FullyConnectedOp>(op);
-}
-
-static bool isTosaUnary(Operation *op) {
-  return isa<tosa::AvgPool2dOp, tosa::ArgMaxOp, tosa::MaxPool2dOp,
-             tosa::RFFT2dOp, tosa::ClampOp, tosa::SigmoidOp, tosa::TanhOp,
-             tosa::ErfOp, tosa::AbsOp, tosa::BitwiseNotOp, tosa::CeilOp,
-             tosa::ClzOp, tosa::CosOp, tosa::ExpOp, tosa::FloorOp, tosa::LogOp,
-             tosa::LogicalNotOp, tosa::NegateOp, tosa::ReciprocalOp,
-             tosa::RsqrtOp, tosa::SinOp, tosa::PadOp, tosa::ReshapeOp,
-             tosa::ReverseOp, tosa::SliceOp, tosa::TileOp, tosa::TransposeOp,
-             tosa::GatherOp, tosa::ScatterOp, tosa::ResizeOp, tosa::CastOp,
-             tosa::RescaleOp, tosa::IdentityOp>(op);
-}
-
-static bool isTosaBinary(Operation *op) {
-  return isa<
-      tosa::MatMulOp, tosa::FFT2dOp, tosa::AddOp, tosa::ArithmeticRightShiftOp,
-      tosa::BitwiseAndOp, tosa::BitwiseOrOp, tosa::BitwiseXorOp, tosa::IntDivOp,
-      tosa::LogicalAndOp, tosa::LogicalLeftShiftOp, tosa::LogicalRightShiftOp,
-      tosa::LogicalOrOp, tosa::LogicalXorOp, tosa::MaximumOp, tosa::MinimumOp,
-      tosa::MulOp, tosa::PowOp, tosa::SubOp, tosa::TableOp, tosa::EqualOp,
-      tosa::GreaterOp, tosa::GreaterEqualOp>(op);
-}
-
 static bool isTosaReduction(Operation *op) {
   return isa<tosa::ReduceMaxOp, tosa::ReduceSumOp, tosa::ReduceMinOp,
              tosa::ReduceProdOp, tosa::ReduceAllOp, tosa::ReduceAnyOp>(op);
 }
 
-static Value traceToRes(Value tensor, func::FuncOp func,
-                        DenseMap<Value, Value> &cache, Value expectedTensor) {
+static Value traceToRes(Value tensor, DenseMap<Value, Value> &cache,
+                        Value expectedTensor) {
   auto cached = cache.find(tensor);
   if (cached != cache.end())
     return cached->second;
 
   Value res = nullptr;
-  if (auto view = tensor.getDefiningOp<ViewLikeOpInterface>())
-    res = traceToRes(view.getViewSource(), func, cache, expectedTensor);
-  else if (auto collapse = tensor.getDefiningOp<tensor::CollapseShapeOp>())
-    res = traceToRes(collapse.getSrc(), func, cache, expectedTensor);
-  else if (isTosaConv(tensor.getDefiningOp())) {
-    auto *op = tensor.getDefiningOp();
-    assert(op->getNumOperands() >= 3 &&
-           "Expected conv operation to have at least three inputs");
-    res = traceToRes(op->getOperand(0), func, cache, expectedTensor);
-    if (!res)
-      res = traceToRes(op->getOperand(1), func, cache, expectedTensor);
-    if (!res)
-      res = traceToRes(op->getOperand(2), func, cache, expectedTensor);
-  } else if (isTosaUnary(tensor.getDefiningOp())) {
-    auto *op = tensor.getDefiningOp();
-    assert(op->getNumOperands() >= 1 &&
-           "Expected unary operation to have at least one input");
-    res = traceToRes(op->getOperand(0), func, cache, expectedTensor);
-  } else if (isTosaBinary(tensor.getDefiningOp())) {
-    auto *op = tensor.getDefiningOp();
-    assert(op->getNumOperands() >= 2 &&
-           "Expected binary operation to have at least two inputs");
-    res = traceToRes(op->getOperand(0), func, cache, expectedTensor);
-    if (!res)
-      res = traceToRes(op->getOperand(1), func, cache, expectedTensor);
-  } else if (isTosaReduction(tensor.getDefiningOp()) &&
-             expectedTensor == tensor)
-    res = tensor;
+  if (tensor.getDefiningOp()) {
+    if (isTosaReduction(tensor.getDefiningOp()) && expectedTensor == tensor) {
+      res = tensor;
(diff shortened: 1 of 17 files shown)


dhernandez0:
@manupak I'm not sure what happens if there is a gemm/conv with splitK + fusion where the gemm output is not returned. Let's say we have
kernel(..., argN : write_access)
a= conv()
b=a+10
copy b to argN

a is not returned, so GemmToGridwise won't trace it back to an argument and won't "prefil=0" anything. But I think b should be init to 0s?

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -1120,7 +1216,8 @@ typename std::enable_if_t<
           func->getAttrOfType<SymbolRefAttr>("original_func");
       if (func::FuncOp originalFunc = dyn_cast<func::FuncOp>(
               symTable.lookupSymbolIn(rootMod, originalFuncAttr))) {
-        originalFunc.setResultAttr(0, "mhal.read_access", rw.getUnitAttr());
```
dhernandez0, line 1123: should we add "func.setResultAttr(resNumber, rock::PrefillAttr::getMnemonic(), outputInitVal)" for original_func as well?

manupak:
Lets sync up on this... 

In my mind (maybe in an incorrect way), you should be able to do this tracing w/o requiring to know about the type of the op.
In non-bufferized IR, tracing results to the end should work ? 
If I m mistaken, I d like to understand why...

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -215,6 +215,95 @@ makeRockConv(ConversionPatternRewriter &rw, Operation *op, Value input,
...
+      tosa::MulOp, tosa::PowOp, tosa::SubOp, tosa::TableOp, tosa::EqualOp,
+      tosa::GreaterOp, tosa::GreaterEqualOp>(op);
+}
+
+static bool isTosaReduction(Operation *op) {
+  return isa<tosa::ReduceMaxOp, tosa::ReduceSumOp, tosa::ReduceMinOp,
+             tosa::ReduceProdOp, tosa::ReduceAllOp, tosa::ReduceAnyOp>(op);
+}
+
+static Value traceToRes(Value tensor, func::FuncOp func,
```
manupak: Why not just iterate on getNumOperands()? 

I think we can remove all of the isXOp() functions.

manupak (commented):
Thanks for fixing this. 

One major comment i have is we can remove isTosaConv() (and the like) functions?
getOperands() should allow you to iterate into all inputs.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
dhernandez0: Done. I still need isTosaReduction but the rest are gone.

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
```diff
@@ -781,3 +782,66 @@ LogicalResult mlir::rock::checkLDSSize(StringAttr arch, int64_t ldsBytes) {
...
+    bool containsArg =
+        std::find(funcArgs.begin(), funcArgs.end(), arg) != funcArgs.end();
+    assert(containsArg &&
+           "Found BlockArgument does not belong to func.getArguments()");
+  }
+  if (!args.empty())
+    return args;
+
+  return failure();
+}
```
manupak: nit: newline

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
```diff
@@ -125,6 +128,52 @@ static Value getAccumulator(Value a, Value b, Value c, OpBuilder &builder,
...
+        return WalkResult::interrupt();
+      }
+
+      // initialize to zeros
+      auto elementType = cast<MemRefType>(matC.getType()).getElementType();
+      Attribute zero;
+      if (llvm::isa<FloatType>(elementType)) {
+        zero = builder.getFloatAttr(elementType, 0.0);
+      } else {
+        assert(llvm::isa<IntegerType>(elementType) &&
```
manupak: I d prefer a pass failure here... 
I know that we expect either int or float here but technically its provided by the user.
so it might be better to give feedback as a compilation failure rather than a crash.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -215,6 +215,58 @@ makeRockConv(ConversionPatternRewriter &rw, Operation *op, Value input,
...
+static bool isTosaReduction(Operation *op) {
+  return isa<tosa::ReduceMaxOp, tosa::ReduceSumOp, tosa::ReduceMinOp,
+             tosa::ReduceProdOp, tosa::ReduceAllOp, tosa::ReduceAnyOp>(op);
+}
+
+static Value traceToRes(Value tensor, DenseMap<Value, Value> &cache,
+                        Value expectedTensor) {
+  auto cached = cache.find(tensor);
+  if (cached != cache.end())
+    return cached->second;
```
manupak: use cache.contains(..)

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -215,6 +215,58 @@ makeRockConv(ConversionPatternRewriter &rw, Operation *op, Value input,
   return cop;
 }
 
+static bool isTosaReduction(Operation *op) {
+  return isa<tosa::ReduceMaxOp, tosa::ReduceSumOp, tosa::ReduceMinOp,
+             tosa::ReduceProdOp, tosa::ReduceAllOp, tosa::ReduceAnyOp>(op);
+}
+
+static Value traceToRes(Value tensor, DenseMap<Value, Value> &cache,
```
manupak, line 223: nit : what does Res means in the context ? 
I thought it was result but you are tracing to arguments right ?

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
```diff
@@ -781,3 +782,66 @@ LogicalResult mlir::rock::checkLDSSize(StringAttr arch, int64_t ldsBytes) {
...
+static void traceGemmAllocToArgs(memref::AllocOp buffer,
+                                 const BufferDependencyAnalysis &deps,
+                                 SmallVector<BlockArgument> &args) {
+  IRRewriter rewriter(buffer.getContext());
+  std::optional<llvm::SmallVector<OpOperand *>> readersOperands =
+      deps.getReaders(buffer);
+  if (!readersOperands.has_value())
+    return;
+  for (OpOperand *readerOperand : readersOperands.value()) {
+    if (memref::CopyOp copyOp =
```
manupak: why copy op special cased here ? 
I think we can still look for MemoryEffects where the operation's operand is being written && its a block argument. I can imagine its legit to provide kernels which directly writes to func args without a copy.

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
```diff
@@ -781,3 +782,66 @@ LogicalResult mlir::rock::checkLDSSize(StringAttr arch, int64_t ldsBytes) {
...
+
+FailureOr<SmallVector<BlockArgument>>
+mlir::rock::traceGemmOutputToArgs(Value matC, func::FuncOp func,
+                                  OpBuilder &builder,
+                                  const BufferDependencyAnalysis &deps) {
+  if (func.getNumArguments() == 0)
+    return failure();
+
+  // trace matC to its alloc
+  FailureOr<memref::AllocOp> allocOp = findMemrefAlloc(matC);
```
manupak, line 836: technically matC could be the func arg (whilst we might not see exercise that given how our highlevel pipelines work) but I could hand write a valid kernel as such.
So I dont think we should fail in the case?

manupak (commented):
Thanks for the fix. I have few minor comments and its looking good.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
dhernandez0: happy to use contains but I think it's less efficient as we query the cache twice. First with contains, then with find to actually get the value.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
dhernandez0: Res means results. In TosaToRock we trace to results (the returnop operands). In GemmToGridwise we trace to arguments.

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
dhernandez0: you're right, I can change that.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
manupak: i see. alright.

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
dhernandez0: I see, but then I can't use BufferDependencyAnalysis because there's no buffer. In that case we can check if matC == any kernel argument directly?

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
manupak: hmm.. its constant time access. just felt cleaner.
I ll let you decide.

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
manupak: yes -- I remember making the same complaint on BufferDependencyAnalysis work. :) 
So yea checking for argument directly might be worth it prior to failing.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -222,9 +222,8 @@ static bool isTosaReduction(Operation *op) {
 
 static Value traceToRes(Value tensor, DenseMap<Value, Value> &cache,
                         Value expectedTensor) {
-  auto cached = cache.find(tensor);
-  if (cached != cache.end())
-    return cached->second;
+  if (cache.contains(tensor))
+    return cache.find(tensor)->second;
```
manupak: nit : can use cache.at()

manupak (approved):
LGTM!

PR #1689, opened 2024-10-28
Add a script for generating the quick-tuning perfconfigs list

djramic:
This PR includes a script for finding quick-tuning perfconfigs in the following way:
 - For each tuning problem, it finds the top n perfcofnigs based on a given threshold
 - Using a linear programming model, it identifies the smallest subset of perfcofnigs that covers all problems.
 
 Additionally, there are option to automatically generated the QuickTuningPerfconfigs.inc file that contains selected perfconfigs, and its integration into the codebase.
 
 closes : ROCm/rocMLIR-internal#1641
 closes : ROCm/rocMLIR-internal#1258
 closes : ROCm/rocMLIR-internal#1518


mlir/utils/performance/analysis/quickTuningGen.py
```diff
@@ -0,0 +1,414 @@
...
+        Return a unique list of perfconfigs from the provided dictonary
+        """
+        perfconfigs_set = set()
+        for perfconfigs_lists in problems_to_perfconfigs.values():
+            perfconfigs_set.update(perfconfigs_lists)
+        return list(perfconfigs_set)
+
+    def get_top_n_perfconfigs_per_problems(self, df, targetColumns):
+        """
+        Identifies the top perfcofnigs for each problem based on a threshold
```
dhernandez0: perfcofnigs -> perfconfigs

mlir/utils/performance/analysis/quickTuningGen.py
```diff
@@ -0,0 +1,414 @@
...
+            threshold = max_value * self.th
+            problem_df[name] = grouped_df[grouped_df['TFlops']
+                                          >= threshold]['PerfConfig']
+        return problem_df
+
+    def find(self):
+        """
+        Finds the minimal set of perfconfigs that cover all
+        problems using set cover optimizaiton.
+        Returns : A dictionary containing data types as keys and thier
```
dhernandez0: thier -> their

mlir/test/Dialect/Rock/test_packed_arithmetic.mlir
dhernandez0, line 1: why does this PR change this? and other tests?

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
```diff
@@ -700,15 +544,28 @@ std::vector<InitParamsAccel>
...
-    params = {initParametersForward8Bit, nInitParametersForward8Bit};
-    break;
-  case 16:
-    params = {initParametersFp16, nInitParametersFp16};
-    break;
-  default:
-    params = {initParameters, nInitParameters};
+  if (opType == KernelType::Gemm) {
+    switch (dataTypeA.getIntOrFloatBitWidth()) {
+    case 8:
```
dhernandez0, line 549: this would apply to both f8 and i8 here, is that the goal?

mlir/include/mlir/Dialect/Rock/Tuning/QuickTuningPerfconfigs.inc
```diff
@@ -0,0 +1,353 @@
...
+// END_GEMM_Wmma_f16_DECS
+
+// BEGIN_CONV_Wmma_f16_DECS
+static constexpr size_t nInitParametersFp16Conv = 26;
+static const InitParamsAccel initParametersFp16Conv[nInitParametersFp16Conv];
+// END_CONV_Wmma_f16_DECS
+
+// BEGIN_GEMM_Wmma_i8_DECS
+static constexpr size_t nInitParametersForward8BitGemm = 15;
+static const InitParamsAccel initParametersForward8BitGemm[nInitParametersForward8BitGemm];
```
dhernandez0: is this only for forward? otherwise we can keep names consistent.

mlir/utils/performance/analysis/quickTuningGen.py
```diff
@@ -0,0 +1,414 @@
...
+            content = file.read()
+
+        pattern = re.compile(f'{begin_marker}.*?{end_marker}', re.DOTALL)
+        replacment = f'{begin_marker}\n{new_content}\n{end_marker}'
+        content = pattern.sub(replacment, content)
+
+        with open(file_path, 'w') as file:
+            file.write(content)
+
+    def isAccel(self, arch, datatype):
```
dhernandez0: is_accel? to maintain the same style as the rest of the file

mlir/utils/performance/analysis/quickTuningGen.py
```diff
@@ -0,0 +1,414 @@
...
+            "// END_GEMM",
+            "// BEGIN_CONV",
+            "// END_CONV",
+        ]
+
+        with open(file_path, 'w') as file:
+            file.write("// THIS IS AN AUTOGENERATED FILE.\n")
+            file.write("// DO NOT EDIT THIS FILE DIRECTLY!\n\n")
+            file.write("// clang-format off\n")
+            for instrction_type, datatypes in instruction_types_to_datatypes.items():
```
dhernandez0: instrction_type -> instruction_type

mlir/utils/performance/analysis/quickTuningGen.py
```diff
@@ -0,0 +1,414 @@
...
+
+    parser.add_argument('--op',
+                        required=True,
+                        type=str,
+                        choices=["gemm", "conv"])
+
+    parser.add_argument("--th",
+                        required=False,
+                        type=float,
+                        default=0.93)
```
dhernandez0, line 386: is 0.93 the number used here? from what I heard 0.95 is what we consider noise, by using 0.93 we could be getting worse performance?

mlir/utils/performance/analysis/quickTuningGen.py
```diff
@@ -0,0 +1,414 @@
...
+            problem_to_index = {
+                problem: idx for idx,
+                problem in enumerate(problems)}
+            perfconfig_to_index = {
+                perfconfig: idx for idx,
+                perfconfig in enumerate(perfconfigs)}
+
+            # Create coverage matrix
+            A = np.zeros((n, m), dtype=int)
+            for problem, perfconfig_list in problems_to_perfconfigs.items():
```
dhernandez0: nit: just enumerate here and don't use problem_to_index

mlir/utils/performance/analysis/quickTuningGen.py
```diff
@@ -0,0 +1,414 @@
...
+                i = problem_to_index[problem]
+                for perfconfig in perfconfig_list:
```
dhernandez0: same here

mlir/utils/performance/analysis/quickTuningGen.py
```diff
@@ -0,0 +1,414 @@
...
+                    j = perfconfig_to_index[perfconfig]
+                    A[i][j] = 1
+
+            # Linear programming model to minimize the number of perfconfigs
+            prob = pulp.LpProblem("SetCoverProblems", pulp.LpMinimize)
+            x = pulp.LpVariable.dicts("x", range(m), cat='Binary')
+            prob += pulp.lpSum([x[j]] for j in range(m))
```
dhernandez0, line 315: nice! could you add a comment here to explain, I think we are setting the objective function here, right?

mlir/utils/performance/analysis/quickTuningGen.py
```diff
@@ -0,0 +1,414 @@
...
+            for i in range(n):
+                prob += pulp.lpSum([A[i][j] * x[j]
```
dhernandez0, line 318: same here, I guess these are the constraints?

mlir/test/Dialect/Rock/test_packed_arithmetic.mlir
djramic: Since affix-params gathers perfconfigs from the list, it is needed to adjust the tests after updating the list

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
djramic: Here I 've just add select conv or gemm list based on opType. I think we're using non-accel configs for fp8

mlir/utils/performance/analysis/quickTuningGen.py
djramic: That's correct, but as I understood it, we needed faster quick-tuning, and it was acceptable to sacrifice a few percent of performance. So, I decided to extend on 0.93 since it noticeable reduced the number of configs.

mlir/include/mlir/Dialect/Rock/Tuning/QuickTuningPerfconfigs.inc
djramic: Forward is relevant for conv, but for gemm it doesn't seem quite right to keep it there. I'll correct that.

mlir/utils/performance/analysis/quickTuningGen.py
djramic: An indexing map is required here to provide a global index for the perfconfig, rather than using the index from the given problem

causten:
@dhernandez0  ready for another review

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
dhernandez0: There are fp8 mfmas and i8 mfmas, so here we are using the same params for both. Is that expected? or was this tuned for i8 only?

mlir/utils/performance/analysis/quickTuningGen.py
dhernandez0: you might have missed this one

mlir/utils/performance/analysis/quickTuningGen.py
dhernandez0: same here, still not corrected

mlir/utils/performance/analysis/quickTuningGen.py
dhernandez0: same here

mlir/utils/performance/analysis/quickTuningGen.py
dhernandez0: same here

mlir/include/mlir/Dialect/Rock/Tuning/QuickTuningPerfconfigs.inc
dhernandez0: thanks, let's change that.

dhernandez0 (changes requested):
there are some comments not addressed in the review

mlir/lib/Dialect/Rock/Tuning/GridwiseGemmParams.cpp
djramic: This is tuned only for the i8. When I run tuning for fp8, I get this output:
Tuning: -t fp8 -out_datatype f32 -transA false -transB false -g 3 -m 1024 -n 512 -k 769
v2:64,32,32,4,2,2,1     9.187230e+05
v2:64,32,32,4,2,4,1     7.604820e+05
v2:64,32,32,4,4,2,1     8.972820e+05
...
where it seems that the perfconfigs are in the format for non-xdlops

dhernandez0: Running this command (gfx942): 
```
./bin/rocmlir-gen -operation gemm -t fp8 --arch gfx942:sramecc+:xnack- --num_cu 304 -g 1 -m 1 -k 768 -n 768 -transA=False -transB=False --kernel-repeats 1 --perf_config= | ./bin/rocmlir-tuning-driver --tuning-space=exhaustive
```

I get this:
```
v2:16,16,4,16,16,8,1,1,1        1.952500e+04
v2:16,16,4,16,16,8,3,1,1        5.493000e+03
v2:16,16,4,16,16,8,4,1,1        3.728000e+03
v2:16,16,4,16,16,16,1,1,1       8.179000e+03
v2:16,16,4,16,16,16,3,1,1       3.448000e+03
v2:16,16,4,16,16,16,4,1,1       2.887000e+03
v2:16,16,8,16,16,8,1,1,1        7.617000e+03
v2:16,16,8,16,16,8,3,1,1        1.128990e+05
...
```

We get 9 numbers which corresponds to xdlops. 
What command are you running?

dhernandez0: So I think we either tune for f8 and bf8 or use the default params? So, instead of checking dataTypeA.getIntOrFloatBitWidth() we need to check for the actual type.

djramic: I ran **python3 tuningRunner.py --op gemm --data-type fp8 -c toy-gemm-configs -d**  on gfx90a, but I get the same result with the **rocmlir-tuning-driver** on gfx90a. It seems that f8 mfma is supported only on gfx942?
As our quick-tuning list on develop is outdated, I think we should merge this first, and then investigate f8 in a new ticket?

dhernandez0: I think it's supported in navi4 as well. Ok, that's fair, let's merge this and open a new ticket for fp8 and bf8.

dhernandez0 (approved):

PR #1694, opened 2024-11-14
Fix conversion of quantizelinear for unsigned types

dhernandez0:
In this PR we fix dequantizelinear for unsigned types. The bugs we fix are:
1. incorrect clamp (need to call getZExtValue() for unsigned) in QuantizeLinearConverter
2. incorrect cast to uint after scaling in QuantizeLinearConverter
3. incorrect handling of no bias case in QuantizeLinearConverter
4. missing conversion of uint to signless int for non-splat migraphx.literals

We also added end2end tests for these fixes copying the input from migraphx tests as migraphx.literals


mlir/test/fusion/pr-e2e/mixr-dequant-conv-quant-uint8-issue-1657.mlir
```diff
@@ -0,0 +1,24 @@
+// RUN: rocmlir-driver -kernel-pipeline=migraphx %s | rocmlir-gen -fut mlir_dequantizelinear_convolution_quantizelinear --arch %arch --clone-harness - | rocmlir-driver -host-pipeline=highlevel | rocmlir-gen -print-results -ph -fut mlir_dequantizelinear_convolution_quantizelinear_wrapper --verifier clone - | rocmlir-driver -host-pipeline mhal,runner -kernel-pipeline full | mlir-cpu-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_float16_utils%shlibext,%linalg_test_lib_dir/libmlir_c_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_async_runtime%shlibext --entry-point-result=void | FileCheck %s
```
umangyadav: change file name to not have issue number.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -990,20 +990,20 @@ LogicalResult QuantizeLinearConverter::matchAndRewrite(
...
-      biasType = isa<IntegerType>(biasType) ? cast<Type>(rewriter.getI32Type())
-                                            : cast<Type>(rewriter.getF32Type());
-      bias = createCastOp(rewriter, loc, biasType, bias,
-                          op.getBias().getType().getElementType());
-    }
   }
-  Value asShort =
-      createCastOp(rewriter, loc, biasType, scaled,
-                   op.getScale().getType().getElementType(), origOutputType);
+  if ((bias || origOutputType != outputType) &&
```
umangyadav, line 994: origOutputType != outputType shouldn't that be always true ?

From what i understand origOutputType is the type in migraphx dialect and outputType is the type in TOSA

dhernandez0: origOutputType is the original type (it could be signed/unsigned/signless int or float), outputType is the type after conversion (so, signless int or float). You can see how the type is converted in MIXRShapedToTensorConverter

dhernandez0: so they will be different only if it's a signed or unsigned int. Note that we are looking at the element type, not the type of the tensors. For example, !migraphx.shaped<1x4x5x5xf32, 100x1x20x4> will convert to tensor<100xf32>, but the element type is f32, it didn't change, so origOutputType != outputType is false. However, !migraphx.shaped<1x4x5x5xui8, 100x1x20x4> will get converted to tensor<100xi8>, so here origOutputType != outputType is true.

umangyadav (approved):

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
umangyadav: Thanks for the explanation.

PR #1697, opened 2024-11-25
Fix rocmlir-gen attention i8 verification bug

dhernandez0:
rocmlir-gen was verifying the wrong tensors for i8 attention.


mlir/test/rocmlir-gen/attention-verify-i8.mlir
```diff
@@ -0,0 +1,24 @@
+// RUN: rocmlir-gen -seq_len_q 384 -seq_len_k 384 -head_dim_qk 64 -head_dim_v 64 --with-attn-scale --with-attn-bias --operation attention --transK=true  --operation attention -t i8 --arch gfx90a:sramecc+:xnack- -pv   -RMS_threshold 0.003 | FileCheck %s --enable-var-scope --check-prefixes=CHECK_I8_FULL
```
umangyadav, line 1: `arch` should it be hardcoded to gfx90a ?

dhernandez0: it doesn't matter because we are just checking in the main() function if verification function takes the last tensor out of gpu and cpu calls, so we just don't look at the generated kernel mlir/code. You can see that in other tests such as attention-kernel.mlir it's also hardcoded.

umangyadav (approved):

PR #1699, opened 2024-11-29
Workaround for issue 1661

dhernandez0:
workaround for https://github.com/ROCm/rocMLIR-internal/issues/1661

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
@@ -1043,7 +1043,7 @@ struct GridwiseAttentionAccelRewritePattern
                             gemm0OutExpTrs, gemm0OutTrs},
         /*bounds=*/ArrayRef<int64_t>{g0Mpt, g0Npt},
         /*strides=*/ArrayRef<int64_t>{1, 1},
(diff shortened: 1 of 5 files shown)


mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2090,16 +2094,17 @@ struct GridwiseAttentionAccelRewritePattern
       postProcessFirstGemmSplat<ElementwiseMultOp>(
           rewriter, loc, gridCoordsGemm0, gemm0OutBuffer, gemm0OutSubTileViews,
           ln2Recip.getDefiningOp<arith::ConstantOp>().getValue());
-#endif
 
       // Handle padding
       bool hasPadding =
           op.getPrePadG0M().has_value() || op.getPrePadG0N().has_value();
       if (hasPadding) {
+        bool isGfx11 = arch.contains("gfx11");
```
umangyadav, line 2102: Are there any existing unit-tests for `gridwisetoblockwise` where we can add test case for gfx11 ?

dhernandez0: good point, we can add a test to check there's no forceunroll for gfx11

umangyadav (approved):

PR #1658, opened 2024-09-20
[CI] Add hipBLAS and hipBLASLt Dependencies

stefankoncarevic:
This update includes the addition of `hipblas-dev` and `hipblaslt-dev` packages in the Dockerfile. These dependencies are necessary for resolving CMake configuration errors and ensuring compatibility with the latest MIGraphX changes.

PR #1703, opened 2024-12-05
Upstream merge Nov 24

djramic:
Upstream merge November 2024

closes: [#1658](https://github.com/ROCm/rocMLIR-internal/issues/1658)


mlir/lib/Dialect/Rock/Transforms/CleanMath.cpp
```diff
@@ -93,6 +93,10 @@ static LogicalResult deleteTrivialRemainder(DataFlowSolver &solver,
 
 static LogicalResult replaceWithConstant(DataFlowSolver &solver, OpBuilder &b,
                                          OperationFolder &folder, Value value) {
+  IntegerType integerType = llvm::dyn_cast<IntegerType>(value.getType());
+  if (!integerType)
```
djramic: I add this check because maybeInferredRange->getValue().isUninitialized() now return 0 for a vector, which causes a crash on getIntegerAttr(). I'm not sure if this is entirely correct?

dhernandez0 (approved):

PR #1707, opened 2024-12-18
add fp8_fp8 in perfRunner

umangyadav:
Running `python3 tuningRunner.py --op conv --tuning-space=full --config="convfp8_fp8 -F 1 -f GNC01 -I NGC01 -O NGC01 -n 1 -c 128 -H 56 -W 56 -k 256 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -g 1"` doesn't work without this fix.

With this fix, it runs correctly. May need to do some more fixes for other FP8 types later on.


mlir/utils/performance/perfRunner.py
```diff
@@ -485,7 +485,7 @@ def __init__(self, dtype: str, direction: str, filterLayout: str, inputLayout:st
                     n: int, c: int, hi: int, wi: int, k: int, y: int, x: int,
                     convStrideH: int, convStrideW: int, paddingH: int, paddingW: int,
                     dilationH: int, dilationW: int, group: int, arch: str, numCU: int):
-        if dtype not in {"f16", "f32", "bf16", "i8", "fp8"}:
+        if dtype not in {"f16", "f32", "bf16", "i8", "fp8_fp8"}:
```
dhernandez0: let's have "fp8" as well? I think first_second is first -> input type, second -> weight type?

umangyadav: Yes first is activation and second one is weight

umangyadav: > let's have "fp8" as well?

Done

dhernandez0 (approved):

PR #1690, opened 2024-10-28
find BlockArgument from gemm output going through all view-like operations

dhernandez0:
nightly test failed (http://ml-ci-internal.amd.com:8080/blue/organizations/jenkins/MLIR%2Fmlir-nightly-all/detail/mlir-nightly-all/1424/pipeline/184/) because of a bug in the last commit: https://github.com/ROCm/rocMLIR/commit/5f51701ee5ff5216125eb3e62deb2dce402cdd9c

PR #1710, opened 2024-12-31
[DRAFT]Add support for bf16 attention

djramic:
This commit adds bf16 type for attention op, resolving the issue mentioned in ticket https://github.com/ROCm/rocMLIR-internal/issues/1666 . I tried adding the tests based on f16 attention, and after extending the diff thresholds, there are still tests that fail. The failing tests are within the comments. 
I am uncertain about the acceptable diff threshold for bf16, as the failing tests show a maxRelDiff reaching 2.8e+00

### DIFF
--- a/mlir/test/e2e/PrAttentionBF16.toml
+++ b/mlir/test/e2e/PrAttentionBF16.toml
@@ -1,0 +1,83 @@
+directory = "PrAttentionBF16"
+prefix = "rocmlir-gen"
+suffix = "--operation attention -t bf16 --arch %arch -pv %random_data %rocmlir_gen_flags -relDiff_threshold 0.3 -absDiff_threshold 0.3 -RMS_threshold 0.15 | rocmlir-driver -c | mlir-cpu-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_float16_utils%shlibext --entry-point-result=void | FileCheck %s --check-prefix="
+
+[[axis]]
+name = "operation"
+values = ["attention"]
+prefix = "--operation "
+
+[[axis]]
+name = "transK"
+values = ["true", "false"]
+prefix = "--transK="
+
+## attention variant
+[[suite]]
+name = "pr_attention_bf16"
+
+[[suite.test]]
+config = "-seq_len_q 384 -seq_len_k 384 -head_dim_qk 64 -head_dim_v 64"
+
+# reverse grid test
+[[suite.test]]
+config = "--reverse_grid -seq_len_q 384 -seq_len_k 384 -head_dim_qk 64 -head_dim_v 64"
+
+[[suite.test]]
+config = "-seq_len_q 64 -seq_len_k 64 -head_dim_qk 64 -head_dim_v 64 -perf_config attn:v1:32,32,64,32,32,32,4,1"
+
+## This one test kPerBlock (16 x 4) == head_dim case
+## Also it has drepeats > 1 to check correct fetching
+[[suite.test]]
+config = "-seq_len_q 256 -seq_len_k 256 -head_dim_qk 128 -head_dim_v 128 -perf_config attn:v1:128,128,128,32,64,64,4,1"
+
+[[suite.test]]
+config = "-seq_len_q 64 -seq_len_k 64 -head_dim_qk 64 -head_dim_v 64"
+
+# Check a padding config
+[[suite.test]]
+config = "-seq_len_q 64 -seq_len_k 64 -head_dim_qk 64 -head_dim_v 64 -perf_config attn:v1:64,64,256,4,64,64,8,1"
+
+# Check lds bypass config
+[[suite.test]]
+config = "-seq_len_q 256 -seq_len_k 256 -head_dim_qk 128 -head_dim_v 128 -perf_config attn:v1:128,128,128,32,128,32,4,1"
+
+# check a case where MPerBlock != gemm1M
+[[suite.test]]
+config = "-seq_len_q 384 -seq_len_k 384 -head_dim_qk 64 -head_dim_v 64 -perf_config attn:v1:32,32,64,8,32,32,8,1"
+
+# check a case where MPerBlock != gemm1M && MPerWave < 32
+[[suite.test]]
+config = "-seq_len_q 384 -seq_len_k 384 -head_dim_qk 64 -head_dim_v 64 -perf_config attn:v1:32,32,64,8,16,16,8,1"
+
+# check scale
+#[[suite.test]]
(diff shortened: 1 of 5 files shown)


giuseros:
> as the failing tests show a maxRelDiff reaching 2.8e+0

As a starter, I am surprised to see a RelDiff higher than 1. 
- The first thing you should do is to verify if the verifier works. You could try to get the output of the verifier and compare it against (say) pytorch, to see if they match
- If the verifier is doing the right thing, another thing you can do is to try with all 1s (there should be an option in rocmlir-gen to do that).
- Another experiment is to remove the part that does the softmax from FA, hence doing just a gemm+gemm. In this way you can isolate which part of the code is giving you issues
- Last experiment is to try integer random values (instead of floating points)
Depending by the results of these experiments we can try to move forward to understand what is wrong.

giuseros:
Also, which configurations (among the many you are using) is giving you this error?

mlir/test/e2e/PrAttentionBF16.toml
```diff
@@ -0,0 +1,83 @@
...
+
+# check scale and bias together
+#[[suite.test]]
+#config = "-seq_len_q 384 -seq_len_k 384 -head_dim_qk 64 -head_dim_v 64 --with-attn-scale --with-attn-bias"
+
+# cross attention
+[[suite.test]]
+config = "-seq_len_q 128 -seq_len_k 27 -head_dim_qk 64 -head_dim_v 32 --with-attn-scale --with-attn-bias"
+
+# issue 1661
```
umangyadav, line 69: Is this issue solved now with recent upstream merge by any chance  ?

mlir/test/e2e/PrAttentionBF16.toml
```diff
@@ -0,0 +1,83 @@
...
+[[suite.test]]
+config = "-seq_len_q 384 -seq_len_k 384 -head_dim_qk 64 -head_dim_v 64 -perf_config attn:v1:32,32,64,8,32,32,8,1"
+
+# check a case where MPerBlock != gemm1M && MPerWave < 32
+[[suite.test]]
+config = "-seq_len_q 384 -seq_len_k 384 -head_dim_qk 64 -head_dim_v 64 -perf_config attn:v1:32,32,64,8,16,16,8,1"
+
+# check scale
+#[[suite.test]]
+#config = "-seq_len_q 384 -seq_len_k 384 -head_dim_qk 64 -head_dim_v 64 --with-attn-scale"
```
dhernandez0, line 55: this doesn't seem related to kvcache. Does this fail too?

mlir/test/e2e/PrAttentionBF16.toml
```diff
@@ -0,0 +1,83 @@
...
+
+# check bias
+[[suite.test]]
+config = "-seq_len_q 384 -seq_len_k 384 -head_dim_qk 64 -head_dim_v 64 --with-attn-bias"
+
+# check scale and bias together
+#[[suite.test]]
+#config = "-seq_len_q 384 -seq_len_k 384 -head_dim_qk 64 -head_dim_v 64 --with-attn-scale --with-attn-bias"
```
dhernandez0, line 63: same here

mlir/test/e2e/PrAttentionBF16.toml
```diff
@@ -0,0 +1,83 @@
...
+
+# cross attention
+[[suite.test]]
+config = "-seq_len_q 128 -seq_len_k 27 -head_dim_qk 64 -head_dim_v 32 --with-attn-scale --with-attn-bias"
+
+# issue 1661
+#[[suite.test]]
+#config = "-seq_len_q 1 -seq_len_k 384 -head_dim_qk 64 -head_dim_v 64 --with-attn-scale --with-attn-bias"
```
dhernandez0, line 71: same here, not related to kvcache or GQA

dhernandez0 (commented):
some comments

djramic:
> Also, which configurations (among the many you are using) is giving you this error?

@giuseros, The error arises from this configuration:**-seq_len_q 384 -seq_len_k 384 -head_dim_qk 64 -head_dim_v 64 --with-attn-scale** , but when I use smaller values for the parameters, it pass. Validation also passes with all 1s and with rand n, the issue occurs with rand fixed.
When manually compare the results from the validator with other types (f32 and f16), I notice that valNum differs significantly for negative values, while GpuNum is quite similar. For example:
bf16: 
Validation Num[1]: -0.257812
Validation Num[2]: -0.062988
Validation Num[3]: 0.582031
Validation Num[4]: -0.257812
Validation Num[5]: -0.062988
f16:
Validation Num[1]: -0.144531
Validation Num[2]: -0.249146
Validation Num[3]: 0.643555
Validation Num[4]: -0.144531
Validation Num[5]: -0.249146
it gives me this error on bf16: maxRelDiff = 2.8e+00 (valNum = -0.0629882812, gpuNum = -0.2382812500)
Does this indicate that the issue is with the verifier and generating valNum?

djramic:
> > Does this indicate that the issue is with the verifier and generating valNum?

@umangyadav, @dhernandez0 Do you agree that the problem is probably in the verifier? Since it's a blocking issue, I think we could merge the changes and open a new ticket for the verifier?

dhernandez0:
something weird is going on, if I run `LIT_FILTER=PrAttentionBF16 ninja check-rocmlir`, the `issue 1661` test fails. However, if I copy-paste the test and run it, it works. Maybe there are some initialization issues?

```bash
./bin/rocmlir-gen -seq_len_q 1 -seq_len_k 384 -head_dim_qk 64 -head_dim_v 64 --with-attn-scale --with-attn-bias --operation attention --transK=false  --operation attention -t bf16 --arch gfx942:sramecc+:xnack- -pv   -relDiff_threshold 0.3 -absDiff_threshold 0.3 -RMS_threshold 0.15 | ./bin/rocmlir-driver -c | ./external/llvm-project/llvm/bin/mlir-cpu-runner -O2 --shared-libs=/home/danherna/mlir-dev/rocMLIR-bf16-attention/build/external/llvm-project/llvm/lib/libmlir_rocm_runtime.so,/home/danherna/mlir-dev/rocMLIR-bf16-attention/build/lib/libconv-validation-wrappers.so,/home/danherna/mlir-dev/rocMLIR-bf16-attention/build/external/llvm-project/llvm/lib/libmlir_runner_utils.so,/home/danherna/mlir-dev/rocMLIR-bf16-attention/build/external/llvm-project/llvm/lib/libmlir_float16_utils.so --entry-point-result=void
```

dhernandez0 (approved):

dhernandez0:
There are some special handling for f16 in rocmlir-gen. We should do the same for bf16 I guess: https://github.com/ROCm/rocMLIR/blob/6008ca2ed00241206239f3eaa49e7666d56eb93a/mlir/tools/rocmlir-gen/rocmlir-gen.cpp#L3134
and
https://github.com/ROCm/rocMLIR/blob/6008ca2ed00241206239f3eaa49e7666d56eb93a/mlir/tools/rocmlir-gen/rocmlir-gen.cpp#L3140

Tests still fail even with this change.

djramic:
Validation error moved to ticket: [#1690](https://github.com/ROCm/rocMLIR-internal/issues/1690)

PR #1716, opened 2025-01-14
Allow OCP FP8 emulation 

umangyadav:
Fixes https://github.com/ROCm/rocMLIR-internal/issues/1673

rocMLIR has `emulate-fp8-ext-trunc` pass which would convert Fp8 and OCP Fp8 to Fp32 if the hardware doesn't support FP8. 

ROCk pipeline was  set up to only convert Fp8 Nanoo to Fp32 but not the OCP fp8 even though `emulate-fp8-ext-trunc` pass supports it because MI300 only allowed Nanoo FP8. 

Until now we never had case where OCP fp8 and Nanoo FP8 both appeared in same kernel. https://github.com/ROCm/rocMLIR-internal/issues/1673 has the case where they both appear together.


mlir/include/mlir/Conversion/RocMLIRPasses.td
```diff
@@ -130,6 +130,12 @@ def EmulateFp8ExtTruncPass : Pass<"emulate-fp8-ext-trunc"> {
     "memref::MemRefDialect",
     "vector::VectorDialect",
   ];
+  let options = [
+    Option<"hasFp8ConversionInstrs", "f8-conversion-instrs", "bool", /*default=*/"false",
+           "if the chip has fp8 Nanoo to fp32 conversion instructions or not">,
+    Option<"hasOcpFp8ConversionInstrs", "ocpf8-conversion-instrs", "bool", /*default=*/"false",
```
dhernandez0, line 136: why do we need default values? also other arguments are not defined here (RocMLIRPasses.td), is this needed?

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
```diff
@@ -83,16 +94,31 @@ static bool isFp8(Type t) {
...
+}
+
+static bool isOcpF8(Type t) { return t.isFloat8E5M2() || t.isFloat8E4M3FN(); }
+
+static LogicalResult canBeConverted(Type t, bool hasF8ConversionInstrs,
+                                    bool hasOcpF8ConversionInstrs) {
+  Type elemType = getElementTypeOrSelf(t);
+  if (!isFp8(elemType))
+    return failure();
+  if (hasF8ConversionInstrs and isNanooF8(elemType)) {
```
dhernandez0: nit: I think we use && instead of and... But if the clang-format passes it might be fine?

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
```diff
@@ -83,16 +94,31 @@ static bool isFp8(Type t) {
...
     return failure();
+  }
+  if (hasOcpF8ConversionInstrs and isOcpF8(elemType)) {
```
dhernandez0: if we have ocp conversion and the type is ocp, should it fail or succeed?

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
```diff
@@ -739,14 +773,18 @@ void EmulateFp8ExtTruncPass::runOnOperation() {
...
     Type outElemType = getElementTypeOrSelf(op.getOut().getType());
-    if (outElemType.isFloat8E4M3FNUZ())
-      f8E4M3FNUZLocs.push_back(op->getLoc());
-    else if (outElemType.isFloat8E5M2FNUZ())
-      f8E5M2FNUZLocs.push_back(op->getLoc());
-    else if (outElemType.isFloat8E4M3FN())
-      f8E4M3FNLocs.push_back(op->getLoc());
-    else if (outElemType.isFloat8E5M2())
-      f8E5M2Locs.push_back(op->getLoc());
+    if (not hasFp8ConversionInstrs) {
```
dhernandez0: nit: same here for not and !.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -21,45 +21,53 @@ static constexpr AmdArchInfo
...
-              /*hasFp8ConversionInstrs=*/false, /*maxNumXCC=*/1),
+              /*hasFp8ConversionInstrs=*/false,
+              /*hasOcpFp8ConversionInstrs=*/false, /*maxNumXCC=*/1),
     cdna3Info(GemmFeatures::mfma | GemmFeatures::dot | GemmFeatures::atomic_add,
               /*waveSize=*/64, /*maxWavesPerEU*/ 10, /*totalSGPRPerEU*/ 512,
               /*totalVGPRPerEU*/ 512, /*totalSharedMemPerCU*/ 65536,
               /*maxSharedMemPerWG*/ 65536, /*numEUPerCU=*/4, /*minNumCU=*/228,
-              /*hasFp8ConversionInstrs=*/true, /*maxNumXCC=*/8),
+              /*hasFp8ConversionInstrs=*/true,
+              /*hasOcpFp8ConversionInstrs=*/false, /*maxNumXCC=*/8),
```
dhernandez0, line 49: no arch seems to have hasOcpFp8ConversionInstrs=true. Is that correct?

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -99,6 +107,7 @@ AmdArchInfo mlir::rock::lookupArchInfo(StringRef arch) {
     // after hardware release
     AmdArchInfo gfx12Info(gfx11Info);
     gfx12Info.hasFp8ConversionInstrs = true;
+    gfx12Info.hasOcpFp8ConversionInstrs = false;
```
dhernandez0: I thought gfx12 has ocp f8? sorry it's confusing for me. Please, could you double-check?

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
krzysz00: LLVM style wants `&&`, I'm pretty sure

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
umangyadav: It should fail because this function is checking for `canBeConverted()`. We wouldn't want to convert OCP F8 to F32 if the hardware has native instructions that can do the conversion.

mlir/include/mlir/Conversion/RocMLIRPasses.td
umangyadav: > also other arguments are not defined here (RocMLIRPasses.td) 

Can you elaborate which other argument you are referring to ? 

Default values are good, e.g. we run this pass as the very last pass in rock pipeline. having default argument allow us to be concise. 

https://github.com/ROCm/rocMLIR/blob/b02520516f3f2fc8603ff7a5b71da3114090faab/mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp#L271

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
umangyadav: Changed. Done

mlir/lib/Conversion/EmulateFp8ExtTrunc/EmulateFp8ExtTrunc.cpp
umangyadav: Done

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
umangyadav: I'll follow up on this over the chat.

umangyadav: Feature list for gfx12 and beyond is not accurate at the moment. I'll try to find information but i think until we do bring up we may not need to worry about his. 
I noticed that we don't list out MI350.

dhernandez0: I think gfx12 is navi4, according to our PR: https://github.com/llvm/llvm-project/pull/106160/files#diff-7a75bd4ce9136bf5fac1d8e3110cbdd56f635f3b91d9fb2d32f53ad443a0e0f2R55 it has ocp f8.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -119,8 +128,14 @@ GemmFeatures mlir::rock::AmdArchInfo::getDefaultFeatures(Type dataType) {
...
   bool isMfma = bitEnumContainsAll(theseFeatures, GemmFeatures::mfma);
+
   if (isMfma && !hasFp8ConversionInstrs) {
-    if (isa<FloatType>(dataType) && dataType.getIntOrFloatBitWidth() == 8)
+    if (isa<Float8E4M3FNUZType>(elementType) or
+        isa<Float8E5M2FNUZType>(elementType))
+      theseFeatures = bitEnumClear(theseFeatures, GemmFeatures::mfma);
+  }
+  if (isMfma && !hasOcpFp8ConversionInstrs) {
+    if (isa<Float8E4M3FNType>(elementType) or isa<Float8E5M2Type>(elementType))
```
dhernandez0: or -> ||

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -119,8 +128,14 @@ GemmFeatures mlir::rock::AmdArchInfo::getDefaultFeatures(Type dataType) {
     }
   }
   bool isMfma = bitEnumContainsAll(theseFeatures, GemmFeatures::mfma);
+
   if (isMfma && !hasFp8ConversionInstrs) {
-    if (isa<FloatType>(dataType) && dataType.getIntOrFloatBitWidth() == 8)
+    if (isa<Float8E4M3FNUZType>(elementType) or
```
dhernandez0: or -> ||

mlir/include/mlir/Conversion/RocMLIRPasses.td
dhernandez0: ah I didn't realize we call the pass twice (second time with flase for the arguments). Thanks

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
umangyadav: Done

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
umangyadav: Done.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
umangyadav: Updated `gfx12` to allow OCP FP8

dhernandez0 (approved):

mlir/test/rocmlir-gen/wmma-enablement.mlir
```diff
@@ -1,6 +1,6 @@
 // RUN: rocmlir-gen --arch gfx1201 --operation gemm --operation gemm -wmma infer -t f16 -p | grep '|wmma' | count 1
-// RUN: rocmlir-gen --arch gfx1201 --operation gemm -wmma infer -t fp8_fp8 -p | grep '|wmma' | count 1
-// RUN: rocmlir-gen --arch gfx1201 --operation gemm -wmma infer -t bf8_bf8 -p | grep '|wmma' | count 1
+// RUN: rocmlir-gen --arch gfx1201 --operation gemm -wmma infer -t f8E4M3FN -p | grep '|wmma' | count 1
```
dhernandez0: why is this change needed? rocmlir-gen should be able to figure out the right fp8 type from the architecture

umangyadav: here it's just testing whether it should use `wmma` or not.  e.g. for the gfx12 it has OCP FP8 therefore `rocmlir-gen` kernel should have `wmma` attribute. But if it let's say `rocmlir-gen fp8_fnuz` then it can't really use `wmma`.

dhernandez0: Yes, I understand. What I mean is that rocmlir-gen should be able to figure out that fp8 must be ocp fp8 for gfx12. So the original test should work

umangyadav: It does that only  if you don't supply `--arch`

umangyadav: https://github.com/ROCm/rocMLIR/blob/9bffae0f4929e14d4d269af4af19e7d0482dd67a/mlir/tools/rocmlir-gen/rocmlir-gen.cpp#L4018

dhernandez0: I think that line you provided checks if arch is not empty. In this test we provide arch so I think it should work.

umangyadav: Yes. i was not reading properly. reverted test changes.

Thanks for catching this.

dhernandez0: Thanks!

CharlieL7:
Fixes the issue I was seeing from https://github.com/ROCm/rocMLIR-internal/issues/1673 on MIGraphX!

PR #1718, opened 2025-01-15
Split-k fusions

dhernandez0:
In this PR we fuse some operations with split-k:
- AddFOp/SubFOp
- MulFOp/DivFOp
- NegFOp
- conversion operations (SIToFPOp, UIToFPOp, ExtUIOp, ExtSIOp, ExtFOp, TruncFOp, TruncIOp)

For add and sub, we have a pass that allows this fusion. Without the pass, adding `x` would be added `splitKFactor` times, so we divide by `splitKFactor`.

There are other minor fixes:
- do not check GEMM output type in GemmToGridwise (for split-k). The GEMM type could be int8, and then linalg.generic can convert to f32 or f16 (only valid types for atomic).

Note: reduce_sum fusion in the next PR: https://github.com/ROCm/rocMLIR/pull/1720

Depends on: https://github.com/ROCm/rocMLIR/pull/1721


mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -1198,8 +1198,9 @@ class ReduceSumConverter final : public OpConversionPattern<tosa::ReduceSumOp> {
                                 ConversionPatternRewriter &rw) const final {
     Type elementType =
         cast<ShapedType>(op.getInput().getType()).getElementType();
-    if (!elementType.isF32()) {
-      return rw.notifyMatchFailure(op, "We only support F32 reductions, yet.");
+    if (!elementType.isF32() && !elementType.isF16()) {
```
umangyadav: I think you can move these changes for the TOSA out of this PR to another one. 
Can you add relevant tests as well ? 
e.g. `tosa-to-rock.mlir` and also e2e tests if there are any ? 
i see one 
`mixr-dot-add-reduce-sum.e2e.mlir` 

also it seem migraphx is passing few more dtypes for the reduce ops 

https://github.com/ROCm/AMDMIGraphX/blob/889fabcda90cc075f9b7705aaf614a04c3c3f5f4/src/targets/gpu/fuse_mlir.cpp#L476

and it is only fusing `reduce_sum` and `reduce_mean` and not `reduce_max`.

mlir/lib/Dialect/Rock/Transforms/FoldBroadcast.cpp
```diff
@@ -202,6 +202,10 @@ struct FoldBroadcast : public OpRewritePattern<rock::GemmOp> {
         op.getNumCUAttr(), op.getFeatures(), op.getStoreMethod(),
         op.getDerivedBlockSizeAttr(), op.getGridSizeAttr(), op.getParamsAttr());
 
+    // Convert optional attributes
+    if (auto attr = (*op).template getAttrOfType<StringAttr>("perf_config"))
+      gemm->setAttr("perf_config", attr);
```
umangyadav, line 207: Can you attach perfConfig in one of the tests inside `fold_broadcast.mlir` and check that it exists after running the pass ?

mlir/lib/Dialect/Rock/utility/fusionUtils.cpp
```diff
@@ -20,9 +23,85 @@
...
+  otherTensor/splitKFactor)
+  - add/sub gemmResult, gemmResult
+  - mul/div gemmResult, otherTensor
+  - neg
+  - type conversion functions
+  Where gemmResult != otherTensor for all cases.
+  */
+  auto outputs = genericOp.getOutputs();
+  assert(outputs.size() == 1);
+  for (auto out : outputs) {
```
umangyadav: you don't really need a loop here if it's guaranteed to be size 1

mlir/lib/Dialect/Rock/utility/fusionUtils.cpp
```diff
@@ -20,9 +23,85 @@
...
+
+  // find tensor index
+  int tensorIndex = -1;
+  for (int i = 0; i < static_cast<int>(genericOp->getNumOperands()); ++i) {
+    auto genericOpInputAlloc = findMemrefAlloc(genericOp->getOperand(i));
+    if (llvm::succeeded(genericOpInputAlloc) &&
+        genericOpInputAlloc->getMemref() == gemmResult)
+      tensorIndex = i;
+  }
+  assert(tensorIndex >= 0);
```
umangyadav: It should be THROW or return `failure()` if it can't find tensorIndex

mlir/lib/Dialect/Rock/utility/fusionUtils.cpp
```diff
@@ -20,9 +23,85 @@
...
+    if (numGemmResults > 0) {
+      // check it's a valid operation
+      if (!validOperationGemmOut(nestedOp)) {
+        return failure();
+      }
+
+      if (isa<MulFOp>(nestedOp) && numGemmResults > 1) {
+        // gemmOut^2 is not allowed
+        return failure();
+      }
```
umangyadav: how about case like this ? 

```cpp
gemmOut = a . b 
addOut = gemmOut + c
mulOut = addOut * gemmOut
```

mlir/lib/Dialect/Rock/utility/fusionUtils.cpp
```diff
@@ -37,11 +116,13 @@ LogicalResult mlir::rock::testFusionLegality(func::FuncOp func) {
           return WalkResult::advance();
         }
 
-        // make sure that no `linalg::GenericOp` reads from a gemm output
+        // save all `linalg::GenericOp` that read from a gemm output
+        SmallVector<linalg::GenericOp> geneticOps;
```
umangyadav:
```suggestion
        SmallVector<linalg::GenericOp> genericOps;
```

??

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -799,20 +799,10 @@ static void reconfigureLAGeneric(LinalgAlignRewriter &b,
...
-          return truncOp.getOut().getType().isInteger(32);
-        })
-        .Default([](Operation &ignored) { return false; });
-  };
-  return success(
-      llvm::all_of(laGeneric.getRegion().getOps(), opCanSwapWithAtomic));
+  auto outElementType =
+      cast<ShapedType>(laGeneric.getOutputs()[0].getType()).getElementType();
+  return success(outElementType.isF32() || outElementType.isF16() ||
+                 outElementType.isInteger(32));
```
umangyadav, line 805: I wonder if this check if redundant provided we already check for valid types in `checkValidOutputFusion`

mlir/lib/Dialect/Rock/utility/fusionUtils.cpp
```diff
@@ -55,6 +136,14 @@ LogicalResult mlir::rock::testFusionLegality(func::FuncOp func) {
           }
         }
 
+        // check if generic ops are valid fusions
+        for (auto genericOp : geneticOps) {
+          SmallVector<std::tuple<Operation *, int>> adds;
```
umangyadav, line 141: What do we have `adds` variable here. it is not used anywhere otherwise.

umangyadav: I see some tests using `TestFunctionFusbility.cpp`  would it make sense to add such tests for split-k with fusion ?

mlir/lib/Dialect/Rock/Transforms/GemmLinalgSplitkNormalizationPass.cpp
```diff
@@ -0,0 +1,144 @@
...
+    assert(gemmOut.size() == splitKFactors.size());
+    if (gemmOut.size() == 1) {
+      LLVM_DEBUG(llvm::dbgs()
+                 << "Found linalg::GenericOp that reads GEMM output, let's "
+                    "modify it if it has addf and/or subf\n");
+      if (failed(divideAddBySplitkFactor(op, gemmOut[0], splitKFactors[0],
+                                         rewriter)))
+        return failure();
+    } else {
+      LLVM_DEBUG(llvm::dbgs() << "No valid linalg::GenericOp found\n");
```
umangyadav: return `failure()` in else branch ?

mlir/lib/Dialect/Rock/Transforms/GemmLinalgSplitkNormalizationPass.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+static LogicalResult divideAddBySplitkFactor(linalg::GenericOp genericOp,
+                                             Value gemmResult,
+                                             int64_t splitKFactor,
+                                             IRRewriter &b) {
+  SmallVector<std::tuple<Operation *, int>> adds;
+  if (failed(checkValidOutputFusion(genericOp, gemmResult, adds)))
+    return failure();
+
+  for (auto [op, index] : adds) {
```
umangyadav:
```suggestion
  for (auto [arith_op, gemm_index] : adds) {
```

mlir/test/fusion/pr-e2e/mixr-gemm-splitk/mixr-gemm-splitk-add.e2e.mlir
```diff
@@ -0,0 +1,15 @@
...
+// ALLOW_RETRIES: 2
+module {
+  // CHECK:  [5,     5,     5,  5,     5,     5,  5,     5,     5,  5,     5,     5,  5,     5,     5]
+
+  // CLONE: [1 1 1]
+  // CLONE-NEXT: Unranked Memref base
+
+  func.func @dot_splitk_add(%arg0: !migraphx.shaped<1x5x4xf32, 20x4x1>, %arg1: !migraphx.shaped<1x4x3xf32, 12x3x1>, %arg2: !migraphx.shaped<1x5x3xf32, 15x3x1>) -> !migraphx.shaped<1x5x3xf32, 15x3x1> attributes{arch = "##TOKEN_ARCH##", enable_splitk_for_tuning, kernel = "mixr"} {
+    %0 = migraphx.dot %arg0, %arg1 {perf_config="v2:16,32,4,16,16,4,4,1,1"} : <1x5x4xf32, 20x4x1>, <1x4x3xf32, 12x3x1> -> <1x5x3xf32, 15x3x1>
+    %2 = migraphx.add %0, %arg2 {} : <1x5x3xf32, 15x3x1>, <1x5x3xf32, 15x3x1> -> <1x5x3xf32, 15x3x1>
```
umangyadav: can you move gemm to second arg for one of such test cases ?

mlir/test/fusion/pr-e2e/mixr-gemm-splitk/mixr-gemm-splitk-sub.e2e.mlir
```diff
@@ -0,0 +1,15 @@
...
+// ALLOW_RETRIES: 2
+module {
+  // CHECK:  [3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3]
+
+  // CLONE: [1 1 1]
+  // CLONE-NEXT: Unranked Memref base
+
+  func.func @dot_splitk_sub(%arg0: !migraphx.shaped<1x5x4xf32, 20x4x1>, %arg1: !migraphx.shaped<1x4x3xf32, 12x3x1>, %arg2: !migraphx.shaped<1x5x3xf32, 15x3x1>) -> !migraphx.shaped<1x5x3xf32, 15x3x1> attributes{arch = "##TOKEN_ARCH##", enable_splitk_for_tuning, kernel = "mixr"} {
+    %0 = migraphx.dot %arg0, %arg1 {perf_config="v2:16,32,4,16,16,4,4,1,1"} : <1x5x4xf32, 20x4x1>, <1x4x3xf32, 12x3x1> -> <1x5x3xf32, 15x3x1>
+    %2 = migraphx.sub %0, %arg2 {} : <1x5x3xf32, 15x3x1>, <1x5x3xf32, 15x3x1> -> <1x5x3xf32, 15x3x1>
```
umangyadav, line 12: move gemm out to second arg for one of the sub tests

mlir/test/fusion/linalg-generic-with-atomic-store.mlir
```diff
@@ -4,16 +4,17 @@
...
     %alloc = memref.alloc() : memref<1x1024x512xf32>
     rock.gemm %alloc = %arg0 * %arg1 features =  mfma|dot|atomic_add storeMethod =  atomic_add {arch = "amdgcn-amd-amdhsa:gfx90a", derivedBlockSize = 256 : i32, params = #rock.xdlops_gemm_derived_params<kpackPerBlock = 2, mPerBlock = 256, nPerBlock = 256, kpack = 4, mPerWave = 128, nPerWave = 128, mnPerXdl = 32, splitKFactor = 1, forceUnroll = true>} : memref<1x1024x512xf32> = memref<1x1024x1024xf32> * memref<1x1024x512xf32>
     
     // expected-error @+1 {{'linalg.generic' op is infusible with non-`Set` store method}}
-    linalg.generic {indexing_maps = [#map, #map], iterator_types = ["parallel", "parallel", "parallel"]} ins(%alloc : memref<1x1024x512xf32>) outs(%arg2 : memref<1x1024x512xf32>) {
-    ^bb0(%in: f32, %out: f32):
+    linalg.generic {indexing_maps = [#map, #map], iterator_types = ["parallel", "parallel", "parallel"]} ins(%alloc : memref<1x1024x512xf32>) outs(%arg2 : memref<1x1024x512xi16>) {
+    ^bb0(%in: f32, %out: i16):
```
umangyadav, line 13: why is this changed to i16 ?

mlir/test/fusion/pr-e2e/mixr-gemm-splitk/mixr-gemm-splitk-mul.e2e.mlir
```diff
@@ -0,0 +1,17 @@
...
+  // CHECK:  [16, 16, 16, 16, 16, 16, 16, 16, 16, 16, 16, 16, 16, 16, 16]
+
+  // CLONE: [1 1 1]
+  // CLONE-NEXT: Unranked Memref base
+
+  func.func @dot_splitk_mul(%arg0: !migraphx.shaped<1x5x4xf32, 20x4x1>, %arg1: !migraphx.shaped<1x4x3xf32, 12x3x1>, %arg2: !migraphx.shaped<1x5x3xf32, 15x3x1>) -> !migraphx.shaped<1x5x3xf32, 15x3x1> attributes{arch = "##TOKEN_ARCH##", enable_splitk_for_tuning, kernel = "mixr"} {
+    %0 = migraphx.dot %arg0, %arg1 {perf_config="v2:16,32,16,16,16,16,16,1,1"} : <1x5x4xf32, 20x4x1>, <1x4x3xf32, 12x3x1> -> <1x5x3xf32, 15x3x1>
+    %cst = migraphx.literal(dense<3.000000e+00> : tensor<1x5x3xf32>) : <1x5x3xf32, 0x0x0>
+    %1 = migraphx.add %arg2, %cst {} : <1x5x3xf32, 15x3x1>, <1x5x3xf32, 0x0x0> -> <1x5x3xf32, 15x3x1>
+    %2 = migraphx.mul %0, %1 {} : <1x5x3xf32, 15x3x1>, <1x5x3xf32, 15x3x1> -> <1x5x3xf32, 15x3x1>
```
umangyadav, line 14: add test with variety of  `neg`, `ext` and `trunc` ops

mlir/lib/Dialect/Rock/utility/fusionUtils.cpp
dhernandez0: this is already fails because we keep track of "derivedGemmResult". I've added a fusability test.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
dhernandez0: It's not because checkValidOutputFusion only happens if split-k is enabled. We should check this is the case for reduce ops (they also enable atomics). A second reason is that fusionUtils is called for tuning (rocmlir-tuning-driver) but you could generate the kernel directly and it wouldn't fail if we don't have this check.

I think we might want to refactor fusionUtils.cpp to check for reductions as well or change its name to make it clear it only runs for split-k.

mlir/lib/Dialect/Rock/utility/fusionUtils.cpp
dhernandez0: > What do we have `adds` variable here. it is not used anywhere otherwise.

The adds variable is used in GemmLinalgSplitkNormalizationPass.

> I see some tests using `TestFunctionFusbility.cpp` would it make sense to add such tests for split-k with fusion ?

we are testing for fusability in mlir/test/fusion/fusability-*.mlir files

mlir/lib/Dialect/Rock/Transforms/GemmLinalgSplitkNormalizationPass.cpp
dhernandez0: this shouldn't be a failure, there might be a linalg::genericop but split-k factor=1, so we just don't have to do anything. I've updated the debug message to make it clear.

mlir/test/fusion/pr-e2e/mixr-gemm-splitk/mixr-gemm-splitk-add.e2e.mlir
dhernandez0: sorry I don't understand. What do you mean?

mlir/test/fusion/linalg-generic-with-atomic-store.mlir
dhernandez0: this test doesn't fail anymore due to the changes in the PR, notice that the set method of gemm is set to atomic and there's an add op inside the linalg.generic. So, to keep the test a failure case, I've changed the output type to int16.

mlir/test/fusion/pr-e2e/mixr-gemm-splitk/mixr-gemm-splitk-mul.e2e.mlir
dhernandez0: nice suggestion. I found a bug (related to rock.prefill) while adding those tests.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
dhernandez0: > I think you can move these changes for the TOSA out of this PR to another one.

sure

> also it seem migraphx is passing few more dtypes for the reduce ops

yes, I'm not sure why there's this difference. We can investigate that in a future ticket (test if running rocmlir with fp8 atomic add works for example).

> and it is only fusing `reduce_sum` and `reduce_mean` and not `reduce_max`.

I don't think there's migraphx.reduce_max (at least I couldn't find it in MIGraphX.td), the only way to get reduce_max from migraphx is softmax.

mlir/test/fusion/pr-e2e/mixr-gemm-splitk/mixr-gemm-splitk-add.e2e.mlir
umangyadav: I mean 

instead of doing `add gemm_out, other_input` in all the tests,

also have tests where it does `add other_input, gemm_out` and `sub other_input, gemm_out`

umangyadav (approved):

PR #1721, opened 2025-01-17
Set rock.prefill type to the blockargument type instead of gemm output type

dhernandez0:


mlir/lib/Dialect/Rock/utility/builderUtils.cpp
```diff
@@ -75,7 +75,7 @@ Value createConstantFloatOp(OpBuilder &b, Location loc, Type type,
     retValue = b.create<ConstantOp>(
         loc, SplatElementsAttr::get(shapedType, constValue));
   } else {
-    retValue = b.create<ConstantOp>(loc, type, b.getFloatAttr(elemType, value));
+    retValue = b.create<ConstantOp>(loc, type, b.getFloatAttr(elemType, apValue));
```
dhernandez0, line 78: I found this as well. Not sure how to add a test as it's an internal function.

mlir/lib/Dialect/Rock/utility/builderUtils.cpp
```diff
@@ -86,7 +86,7 @@ Value createZeroConstantOp(OpBuilder &b, Location loc, Type type) {
   if (elementType.isIntOrIndex()) {
     return createConstantIntOp(b, loc, type, elementType, 0);
   } else {
-    return createConstantFloatOp(b, loc, type, elementType, 0.0);
+    return createConstantFloatOp(b, loc, type, elementType, 0.0f);
```
umangyadav: shouldn't be necessary to add `f` at the end since it would bind with argument of type `float`.

mlir/lib/Dialect/Rock/utility/builderUtils.cpp
umangyadav: Not sure why it is necessary to pass `apValue`. `value` is float type and `apValue` is also float type

dhernandez0: apValue is APFloat, not float. It's converted to whatever fp type is needed (see lines 45 to 67)

umangyadav: yes apValue is APFloat, i meant it was instantiated with Float32 type.

umangyadav (approved):

PR #1722, opened 2025-01-17
Enable f16 sum reduction

dhernandez0:
In this PR we enable f16 sum reduction. Note that f16 max is not enabled here. After chatting with Manupa and reading SugarToLoop lowering of GlobalStore, it looks like no atomic f16 max implemented. We would need to implement it if it's needed in the future.

I've added a new GemmFeature called atomic_add_f16 to figure out whether the GPU has support for it.


umangyadav:
>  We would need to implement it if it's needed in the future. 

Can we create an issue to track this ?

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -1947,6 +1947,17 @@ LogicalResult ReduceOp::verify() {
...
+  
+  auto inElemType = getIn().getType().getElementType();
+  auto outElemType = getOut().getType().getElementType();
+  if(inElemType != outElemType)
+    return emitError("element type of input and output is different");
+
+  if (getReduceMethod() == ReduceMethod::Max &&
+      isa<FloatType>(outElemType))
+    if (!outElemType.isF32())
+      return emitError("reduce max only supports f32");
```
umangyadav:
```suggestion
  if (getReduceMethod() == ReduceMethod::Max &&
     !outElemType.isF32())
      return emitError("reduce max only supports f32");
```

mlir/test/fusion/pr-e2e/reductions/atomic_add/tosa-gemm-add-reduce-sum-f16.e2e.mlir
```diff
@@ -0,0 +1,26 @@
+// RUN: sed s/##TOKEN_ARCH##/%arch/g %s | rocmlir-driver -host-pipeline highlevel | rocmlir-gen -ph -print-results -rand 1 -rand_type float -fut dot_add --verifier clone - | rocmlir-driver -host-pipeline mhal -kernel-pipeline full | xmir-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_float16_utils%shlibext,%linalg_test_lib_dir/libmlir_c_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_async_runtime%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CLONE
+// ALLOW_RETRIES: 2
+// CLONE: [1 1 1]
+// CLONE-NEXT: Unranked Memref base
+
+module {
+  func.func private @dot_add__part_0(%arg0: tensor<1x128x64xf16> {mhal.read_access}, %arg1: tensor<1x64x256xf16> {mhal.read_access}, %arg2: tensor<1x128x256xf16> {mhal.read_access}) -> (tensor<1x128x1xf16> {mhal.write_access}) {
```
umangyadav:
```suggestion
  func.func private @dot_add_part_0(%arg0: tensor<1x128x64xf16> {mhal.read_access}, %arg1: tensor<1x64x256xf16> {mhal.read_access}, %arg2: tensor<1x128x256xf16> {mhal.read_access}) -> (tensor<1x128x1xf16> {mhal.write_access}) {
```

mlir/test/fusion/pr-e2e/reductions/atomic_add/tosa-gemm-add-reduce-sum-f16.e2e.mlir
```diff
@@ -0,0 +1,26 @@
...
+    %2 = "tosa.reduce_sum"(%1) {axis = 2 : i32} : (tensor<1x128x256xf16>) -> tensor<1x128x1xf16>
+    return %2 : tensor<1x128x1xf16>
+  }
+  func.func @dot_add(%arg0: tensor<1x128x64xf16>, %arg1: tensor<1x64x256xf16>, %arg2: tensor<1x128x256xf16>) -> tensor<1x128x1xf16> {
+    %token, %results = mhal.launch @dot_add__part_0 (%arg0, %arg1, %arg2) : (tensor<1x128x64xf16>, tensor<1x64x256xf16>, tensor<1x128x256xf16>) -> tensor<1x128x1xf16>
+    mhal.await %token : !mhal.token
+    return %results : tensor<1x128x1xf16>
+  }
+  module @__xmodule_ attributes {mhal.arch = "##TOKEN_ARCH##", mhal.module} {
+    func.func private @dot_add__part_0(%arg0: tensor<1x128x64xf16> {mhal.read_access}, %arg1: tensor<1x64x256xf16> {mhal.read_access}, %arg2: tensor<1x128x256xf16> {mhal.read_access}) -> (tensor<1x128x1xf16> {mhal.write_access}) attributes {kernel, original_func = @dot_add__part_0} {
```
umangyadav:
```suggestion
    func.func private @dot_add_part_0(%arg0: tensor<1x128x64xf16> {mhal.read_access}, %arg1: tensor<1x64x256xf16> {mhal.read_access}, %arg2: tensor<1x128x256xf16> {mhal.read_access}) -> (tensor<1x128x1xf16> {mhal.write_access}) attributes {kernel, original_func = @dot_add__part_0} {
```

dhernandez0:
> Can we create an issue to track this ?

Let's discuss this on the next meeting. This would be for kernels that end with reduce_max which I think they are not common. Also, migraphx dialect doesn't even have reduce_max anyway. So, unless there's a need from migraphx I don't think we would implement this I think.

umangyadav (approved):

dhernandez0:
f16 reduce would be slow on navi as it does not support atomic_add with packed_f16. Should we fail or return not fusable for this case? @umangyadav

mlir/test/fusion/fusability-dot-add-reduce-max.mlir
```diff
@@ -1,7 +1,7 @@
 // RUN: rocmlir-gen -emit-module-fusibility-for=v2:16,16,4,16,16,1,5,1,1 - < %s | FileCheck %s --check-prefixes=CHECK-SPLITK
 // CHECK-SPLITK: fusible:0
 // RUN: rocmlir-gen -emit-module-fusibility-for=v2:16,16,4,16,16,1,1,1,1 - < %s | FileCheck %s --check-prefixes=CHECK-NONSPLITK
-// CHECK-NONSPLITK: fusible:1
+// CHECK-NONSPLITK: fusible:0
```
umangyadav, line 4: why NONSPLITK is not fusible for this case ?

dhernandez0: only rdna arch has GemmFeatures::atomic_fmax_f32 enabled. There's atomic_max_f64, but no f32 on cdna. See AmdArchDb.cpp.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -1008,10 +1008,10 @@ RocmlirSplitKSelectionLikelihood isSplitKFaster(int64_t gDim, int64_t mDim,
...
 
 bool isModuleFusible(ModuleOp module, StringRef perfConfig) {
-  if (!rock::isSplitKRequested(module, perfConfig)) {
-    return true;
-  }
-  return succeeded(rock::testFusionLegality(module));
+  bool fusible = succeeded(rock::testFusionLegalityReduce(module));
+  if (!rock::isSplitKRequested(module, perfConfig))
+    return fusible;
+  return fusible && succeeded(rock::testFusionLegality(module));
```
umangyadav, line 1014: ideally speaking `testFusionLegalityReduce` should be subset of `testFusionLegality()` therefore we shouldn't really need to check both

mlir/lib/Dialect/Rock/utility/fusionUtils.cpp
```diff
@@ -167,3 +169,40 @@ LogicalResult mlir::rock::testFusionLegality(ModuleOp mod) {
...
+    auto outElemType = reduceOp.getOut().getType().getElementType();
+    if (reduceOp.getReduceMethod() == ReduceMethod::Max) {
+      if (!isa<Float32Type>(outElemType))
+        return WalkResult::interrupt();
+
+      if (!bitEnumContainsAll(reduceOp.getFeatures(),
+                              GemmFeatures::atomic_fmax_f32))
+        return WalkResult::interrupt();
+    } else {
+      if (!isa<Float32Type, Float16Type>(outElemType))
```
umangyadav, line 184: should assert that `else` branch is reduce_add case

dhernandez0: after second thought this might be redundant, because RockOps.td already has defined F32 and F16 types only. Other types would fail due to MLIR checks before we reach here.

PR #1724, opened 2025-01-21
Fix usage of llvm::reverse() and remove warnings

dhernandez0:


mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -543,8 +543,6 @@ traceToWriter(Value startVal,
         trOp = dyn_cast_if_present<TransformOp>(
             trOp.getViewSource().getDefiningOp());
       }
-      // This recursion got us our transform stack in the opposite order.
-      llvm::reverse(writerToStartValViews);
```
dhernandez0, line 547: when traceToWriter() is used, the views are passed to applyViewsOnDest() which calls llvm::reverse internally. So, this call is not needed.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -567,16 +567,6 @@ struct TransposeRewritePattern : public OpRewritePattern<tosa::TransposeOp> {
...
-            // Remove unit dims from smaller end of reassociation indices
-            // but we need at least one for the reassociation
-            // does so by reversing it.
-            llvm::reverse(newReassocIdx);
-            while (newReassocIdx.size() > 1 &&
-                   preTpUnitDims.contains(newReassocIdx.back())) {
-              newReassocIdx.pop_back();
-            }
-            // Needs to re-reverse it.
-            llvm::reverse(newReassocIdx);
```
dhernandez0, line 579: Test pass removing this. The collapse generated seems ok, if I try reversing it, the collapse is incorrect. So, it looks like the code is just not needed.

external/mlir-hal/lib/Dialect/MHAL/Transforms/BufferizableOpInterfaceImpl.cpp
```diff
@@ -30,24 +30,6 @@ namespace mlir {
 namespace mhal {
 namespace {
 
-/// Return the index of the bbArg in the given FuncOp that is equivalent to the
-/// specified return value (if any).
-static std::optional<int64_t>
-getEquivalentFuncArgIdx(FuncOp funcOp, const FuncAnalysisState &state,
```
dhernandez0, line 36: not called anywhere

umangyadav (approved):

PR #1725, opened 2025-01-24
Fix tuning for split-k fusions

dhernandez0:
RockTuningImpl.cpp was checking the GEMM output to decide if split-k is enabled when doing exhaustive tuning. Now we can allow for example: i8 gemm and then type conversion to f16, then split-k would still be valid. I changed the code to use fusionUtils to decide if split-k should be enabled.

Additionally, changed the code to use GemmFeatures (atomic_add and atomic_add_f16) to decide if split-k should be enabled.

### DIFF
--- a/mlir/include/mlir/Dialect/Rock/utility/fusionUtils.h
+++ b/mlir/include/mlir/Dialect/Rock/utility/fusionUtils.h
@@ -9,6 +9,7 @@
 #define ROCK_UTILITY_FUSION_H
 
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/Rock/IR/RockTypes.h"
 #include "mlir/IR/BuiltinOps.h"
 #include "mlir/Support/LogicalResult.h"
 
@@ -22,11 +23,8 @@ class FuncOp;
 } // namespace func
 
 namespace rock {
-// Checks whether a function contains any `linalg::GenericOp` which
-// reads or writes to the output of any `Operation` implementing
-// `RockGemmWrapperInterface`. The result of this test can be ignored
-// if the Data Parallel GEMM scheme is used.
-LogicalResult testFusionLegality(func::FuncOp func);
+// Checks whether a function is valid for split-k.
+LogicalResult testFusionLegalitySplitK(func::FuncOp func);
 
 // Checks whether a function contains any `rock::ReduceOp` and
 // the atomic operation is supported by the hardware.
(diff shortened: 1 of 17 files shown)


mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
```diff
@@ -185,11 +185,6 @@ GemmRewritePattern::matchAndRewrite(GemmOp op, GemmOpAdaptor adaptor,
 
   const int64_t splitKFactor = op.getParams()->getSplitKFactor();
   if (splitKFactor > 1) {
-    if (!bitEnumContainsAll(op.getFeatures(), GemmFeatures::atomic_add)) {
```
umangyadav, line 188: why  this code is removed ?

mlir/include/mlir/Dialect/Rock/utility/fusionUtils.h
```diff
@@ -37,15 +35,15 @@ LogicalResult testFusionLegalityReduce(func::FuncOp func);
 // `func:FuncOp` and calls the implementation `testFusionLegality` (see above).
 // Note, this overloaded function assumes that `ModuleOp` contains
 // a single `func:FuncOp`
-LogicalResult testFusionLegality(ModuleOp mod);
```
umangyadav, line 40: Description requires changes i think. 
Also i think we should keep overload `testFusionLegality()` which checks for everything e.g. `return testFusionLegalitySplitK() && testFusionLegalityReduce() && AnyOtherChecks()`

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
dhernandez0: This checks if there's atomic_add (f32) support in the hardware. Not sure if this happens currently, however we could have a hardware with f16 atomic_add support and no f32 atomic_add, this code will prevent us from using split-k.

I think we either do the test properly here (finding out the final output type) as well or just don't do it as it's been checked before.

mlir/include/mlir/Dialect/Rock/utility/fusionUtils.h
dhernandez0: That is what isModuleFusible() in RockTuningImpl.cpp does:
```C++
bool isModuleFusible(ModuleOp module, StringRef perfConfig) {
  bool fusible = succeeded(rock::testFusionLegalityReduce(module));
  if (!rock::isSplitKRequested(module, perfConfig))
    return fusible;
  return fusible && succeeded(rock::testFusionLegalitySplitK(module));
}
```

Should we just copy-paste this to fusionUtils as testFusionLegality() and have isModuleFusible() call testFusionLegality()?
The important detail is that to decide if we run the split-k checks, we need to have a perfConfig.

dhernandez0:
I have added another fix for non-fused kernels (split-k fusion checks in testFusionLegalitySplitK). We were checking that the output of linalg::genericop is the right type (f32 or f16 and supported by GemmFeatures). However, some kernels might output directly from the GEMM. So, I'm tracing from the GEMM output to kernel arguments (outputs) and checking the arguments types.

@umangyadav should we move this fix to another PR? or is it ok to have it here?

mlir/include/mlir/Dialect/Rock/utility/fusionUtils.h
umangyadav: > The important detail is that to decide if we run the split-k checks, we need to have a perfConfig.

Yeah. It should be okay for now to not add it. If we require in future, we can easily add it.

umangyadav (approved):

PR #1728, opened 2025-01-30
Upstream Merge [Jan]

stefankoncarevic:


dhernandez0:
I see Tosa Conv now takes acc_type. I think we shouldn't add acc_type to migraphx dialect. Unless we agree with migraphx team that they want control on how results will be accumulated. 

Not sure what is the best way to implement this. Currently, the accumulation type is the output type of mfma/wmma. So, I think it's f32 for all floating-point types and i32 for all integer types.

Therefore, I think it'd be easier to decide the acc_type based on the output_type, in MigraphxToTosa, you can map:
switch(output_type)
  case fp_type: return fp32
  case int_type: return i32

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -354,15 +354,14 @@ LogicalResult ConvConverter<ConvType>::matchAndRewrite(
   int64_t group = op.getGroup();
   auto accType = op.getAccTypeAttr().getValue();
   // Convert acc_type to a type supported by TOSA Conv2D operation.
-  // F8 types are converted to f16, bf16 is converted to f32, and 8-bit integers are converted to i32.
-  if (accType.isFloat8E4M3FNUZ() || accType.isFloat8E5M2FNUZ() || accType.isFloat8E5M2() ||
-      accType.isFloat8E4M3FN()) {
+  // F8 types are converted to f16, bf16 is converted to f32, and 8-bit integers
+  // are converted to i32.
+  if (accType.isFloat8E4M3FNUZ() || accType.isFloat8E5M2FNUZ() ||
```
dhernandez0: Let's do what mfma/wmma instructions do, so f32 for all fp types I think?

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -354,7 +352,19 @@ LogicalResult ConvConverter<ConvType>::matchAndRewrite(
   }
 
   int64_t group = op.getGroup();
-  auto accType = op.getAccTypeAttr();
+  auto accType = op.getAccTypeAttr().getValue();
+  // Convert acc_type to a type supported by TOSA Conv2D operation.
+  // F8 types are converted to f16, bf16 is converted to f32, and 8-bit integers are converted to i32.
```
dhernandez0: I think we shouldn't add acc_type to migraphx dialect, unless we agree migraphx team needs it. And if we decide to add it. I'd prefer to fail rather than silently changing the acc_type they request.

stefankoncarevic:
> I see Tosa Conv now takes acc_type. I think we shouldn't add acc_type to migraphx dialect. Unless we agree with migraphx team that they want control on how results will be accumulated.
> 
> Not sure what is the best way to implement this. Currently, the accumulation type is the output type of mfma/wmma. So, I think it's f32 for all floating-point types and i32 for all integer types.
> 
> Therefore, I think it'd be easier to decide the acc_type based on the output_type, in MigraphxToTosa, you can map: switch(output_type) case fp_type: return fp32 case int_type: return i32

Thank you for your suggestion. I agree with your suggestion that for fp_type, acc_type should be fp32, and for int_type, acc_type should be i32.  I was unsure about the best way to handle this.

@pfultz2 I would like to consult with you regarding the addition of the acc_type attribute in the migraphx dialect. Do you think this should be added?

dhernandez0:
> > I see Tosa Conv now takes acc_type. I think we shouldn't add acc_type to migraphx dialect. Unless we agree with migraphx team that they want control on how results will be accumulated.
> > Not sure what is the best way to implement this. Currently, the accumulation type is the output type of mfma/wmma. So, I think it's f32 for all floating-point types and i32 for all integer types.
> > Therefore, I think it'd be easier to decide the acc_type based on the output_type, in MigraphxToTosa, you can map: switch(output_type) case fp_type: return fp32 case int_type: return i32
> 
> Thank you for your suggestion. I agree with your suggestion that for fp_type, acc_type should be fp32, and for int_type, acc_type should be i32. I was unsure about the best way to handle this.
> 
> @pfultz2 I would like to consult with you regarding the addition of the acc_type attribute in the migraphx dialect. Do you think this should be added?

I think the issue is that even if they provide acc_type=f16, the mfma instructions is accumulating in f32. So, there's no way to use another type. We could get the acc_type of mfma, convert it to f16, and accumulate inside the loop with f16. But even then, acc_type is partially f32 and f16 (and some kernels will have no loop, so f32 only).

mlir/test/CAPI/mixr_cobj.cc
```diff
@@ -154,7 +149,7 @@ MlirModule makeAndDumpMIXR(MlirContext ctx, MlirLocation location) {
       mlirStringRefCreateFromCString("migraphx.convolution"), location);
   mlirOperationStateAddResults(&conv0OpState, 1, &conv0Type);
   mlirOperationStateAddOperands(&conv0OpState, 2, conv0Operands);
-  mlirOperationStateAddAttributes(&conv0OpState, 6, conv0Attrs);
+  mlirOperationStateAddAttributes(&conv0OpState, 5, conv0Attrs);
```
dhernandez0: why did this change?

mlir/test/fusion/e2e/generate-fusion-tests.py
```diff
@@ -62,13 +62,15 @@ def generate_op_variants_test(indir, outdir, type, file, opspec):
 def generate_type_only_test(indir, outdir, type, file):
     archNames = getArch()
     arch = ','.join(archNames)
+    # Use f32 instead of bf16 for certain tests because bf16 is not supported as an accumulation type for tosa.conv2d.
+    gen_type = 'f32' if type == 'bf16' and file in ['mixr-bcast-add', 'tosa-to-rock-bcast-add', 'tosa-to-rock-bias'] else type
```
dhernandez0: can't we modify the tests to have acc_type explicitly?

dhernandez0: apply same logic here: floating point -> f32, int type -> i32

mlir/test/fusion/e2e/generate-fusion-tests.py
```diff
@@ -62,13 +62,15 @@ def generate_op_variants_test(indir, outdir, type, file, opspec):
...
     if "bf16" in type and "gfx11" in arch:
```
dhernandez0: why is this disabled?

dhernandez0 (changes requested):
only some comments for mlir/test/fusion/e2e/generate-fusion-tests.py after that i can approve.

mlir/test/fusion/e2e/generate-fusion-tests.py
stefankoncarevic: The line in is not disabling anything.  It introduces a condition for the two tests that utilize tosa.conv2d and are executed for the types f32, f16, and bf16. The condition is specified for data types in the file located at rocMLIR/mlir/test/fusion/e2e/tests.toml.
Within their template, found at rocMLIR/mlir/test/fusion/e2e/tosa-to-rock-bias.e2e.template, tosa-to-rock-bcast-add.e2e.template. I had to differentiate the type used in multiple places by creating a acc_type variable. This variable is set to 'f32' when the type is 'bf16'. This is because 'bf16' is not supported as an accumulation type for tosa.conv2d in these specific tests. Should I also change it so that 'f16' is treated the same way and set to 'f32'? Please note that these tests are not run for 'int8'.

dhernandez0: Sorry, I think you refer to the other comment. This one is doing:

```python
if "bf16" in type and "gfx11" in arch:
  return
```

mlir/test/fusion/e2e/generate-fusion-tests.py
dhernandez0: I'm assuming your comment below refers to this. I think we can just replicate the logic we had for migraphxtotosa: if type is fp: gen_type=f32, if type is int: gen_type=i32. That is easier to read and understand.

mlir/test/fusion/e2e/generate-fusion-tests.py
stefankoncarevic: That was disabled inside this PR https://github.com/ROCm/rocMLIR/pull/1221 because of this problem https://github.com/ROCm/rocMLIR-internal/issues/1080
I think we can make a separate PR for that and test it on navi3x to see if this error still occurs.

dhernandez0: thanks! I wasn't aware of this. Please, can you create a ticket to investigate this?

dhernandez0 (approved):

PR #1729, opened 2025-01-30
KV-cache MIGraphX integration

dhernandez0:
In this PR we add kv-cache support for migraphx+tosa dialect. Specifically:

For migraphx dialect:
- add migraphx.greater_or_equal that lowers to tosa.greater_equal

For tosa dialect:
- add code to match attention with kv-cache in TosaToRock to create rock.attention with currentSeqLen.

TODO: @turneram to test this branch with his migraphx branch


mlir/include/mlir/Dialect/MIGraphX/IR/MIGraphX.td
```diff
@@ -77,6 +77,8 @@ def MIGraphX_DivOp :
     MIGraphX_ElementwiseBinaryOp<"div">;
 def MIGraphX_PowOp :
     MIGraphX_ElementwiseBinaryOp<"pow">;
+def MIGraphX_GreaterOrEqual :
```
umangyadav, line 80: Addition of operator  ideally can be another PR to make this one smaller.

mlir/include/mlir/Dialect/Rock/utility/builderUtils.h
```diff
@@ -9,6 +9,13 @@
 
 namespace mlir {
 namespace rock {
+// Utility to create an APInt of the requested type
+std::pair<APInt, bool> createAPInt(Type elemType, int32_t value);
```
umangyadav: If it is APint, it is probably better to use `int64_t`. `int32_t` should be fine for the inference. but i am not sure if we will see any `uint32_t` which can't be represented in `int32_t` range.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -61,6 +65,85 @@ static bool isConstantZero(Value v) {
...
+
+    return expectedValue == value;
+  }
+  if (auto splatValue = dyn_cast<SplatElementsAttr>(value))
+    return false;
+  if (auto elementsValue = dyn_cast<ElementsAttr>(value))
+    return llvm::all_of(llvm::enumerate(elementsValue.getValues<Attribute>()),
+                        [](const auto &indexedAttr) {
+                          return isConstRangeAttribute(indexedAttr.value(),
+                                                       indexedAttr.index());
```
umangyadav: Isn't it passing `indexedAttr.index()` as the expectedVal ? We want to compare against `expectedVal` passed to first recursive call

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -61,6 +65,85 @@ static bool isConstantZero(Value v) {
...
+    return llvm::all_of(llvm::enumerate(arrayValue.getValue()),
+                        [](const auto &indexedAttr) {
+                          return isConstRangeAttribute(indexedAttr.value(),
+                                                       indexedAttr.index());
+                        });
+
+  return false;
+}
+
+bool isConstRange(Value v) {
```
umangyadav, line 141:
```suggestion
bool isConstZero(Value v) {
```

??

mlir/include/mlir/Dialect/MIGraphX/IR/MIGraphX.td
dhernandez0: there are some bug fixes in this PR as well. Maybe I can split into 3 PRs:
- bug fixes
- migraphxtotosa: greaterorequal op (however, this one will be small)
- tosatorock

But first, let's make sure migraphx confirms this branch works.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
dhernandez0: we are checking if the constant is a range from 0 to N-1.

mlir/include/mlir/Dialect/MIGraphX/IR/MIGraphX.td
dhernandez0: as it seems it's urgent, is it ok if I don't split the PR?

umangyadav: that should be fine.

umangyadav (approved):

PR #1731, opened 2025-01-31
Support for gfx950 arch

dhernandez0:
Add support for gfx950.

TODO: pass tests on hardware

### DIFF
--- a/mlir/utils/performance/perfRunner.py
+++ b/mlir/utils/performance/perfRunner.py
@@ -145,6 +145,7 @@ def getMetricArgsForRocprof(arch):
     current_dir = os.path.dirname(os.path.abspath(__file__))
     metrics_path = os.path.join(current_dir, ROCMLIR_INPUT_METRICS_FILE_NAME)
     metrics = []
+    # TODO (gfx950): check if gfx950 supports this
     if (chip not in ["gfx942"]):
        metrics = ['-i', metrics_path]
     return metrics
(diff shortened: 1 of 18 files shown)


mlir/utils/performance/perfRunner.py
```diff
@@ -145,6 +145,7 @@ def getMetricArgsForRocprof(arch):
     current_dir = os.path.dirname(os.path.abspath(__file__))
     metrics_path = os.path.join(current_dir, ROCMLIR_INPUT_METRICS_FILE_NAME)
     metrics = []
+    # TODO (gfx950): check if gfx950 supports this
```
umangyadav, line 148: Can we verify this if we get access to machine ?

umangyadav (commented):
Looks good overall. Thanks.

mlir/utils/performance/perfRunner.py
dhernandez0: maybe, if rocprof works I can test it. Otherwise we'll have to wait.

umangyadav (approved):

stefankoncarevic:
Thank you, Daniel. I can see for the test cases that are being executed under the MFMA scope, which is important at the beginning to see how the tests will be executed. Additionally, it's good to see that has been included in parameterSweeps.py to test and identify any potential issues. I think this is sufficient for now. When drivers and machine been more stable, we will add gfx950 for tuning stage.

stefankoncarevic (approved):

dhernandez0:
all these tests fail:
image

dhernandez0:
This seems related to https://ontrack-internal.amd.com/browse/SWDEV-480399
If I set optimization level to O0, all tests pass (some of them after re-launching manually, I get some out of memory and hipErrorIllegalState)
I think we should merge upstream merge to see if some bugs have been fixed: https://github.com/ROCm/rocMLIR/pull/1728

PR #1734, opened 2025-02-04
Enable bf16 atomic add for gfx950

dhernandez0:
In this PR we enable bf16 atomic add for gfx950. That means bf16 split-k and bf16 reductions.

Note that https://github.com/ROCm/rocMLIR/pull/1732 needs to be merged first. Then, I'll point this PR to develop branch.


umangyadav (approved):
We should add some E2E tests to run on Gfx950 at some point. Or we can add them now and mark them disabled/xfail. 

Do we have any tracking issue ?

dhernandez0:
> We should add some E2E tests to run on Gfx950 at some point. Or we can add them now and mark them disabled/xfail.
> 
> Do we have any tracking issue ?

I think we don't need gfx950 specific tests, unless there's a good reason? all e2e tests we have now (or the ones that only run for mfma enabled) would run on gfx950. Regarding this PR, this depends on https://github.com/ROCm/rocMLIR/pull/1732 (needs to be merged first), and that adds e2e tests for bf16 atomic add.

PR #1735, opened 2025-02-05
Fix build failures

umangyadav:


dhernandez0 (approved):
sorry about that. Not sure how CI let that happen?

umangyadav:
> sorry about that. Not sure how CI let that happen?

perf tests against rocblas only happens in nightly runs. Therefore it didn't run this stage for PR.

PR #1738, opened 2025-02-07
Add fp8 convolution to the tuning runner

djramic:
In this PR, the DATA_TYPE in perf runner has been extended to include convolution configurations with FP8.
close: https://github.com/ROCm/rocMLIR-internal/issues/1708


mlir/utils/performance/perfRunner.py
```diff
@@ -29,7 +29,7 @@
 BENCHMARKING_METRICS_FILE_NAME = 'results.csv'
 ROCMLIR_INPUT_METRICS_FILE_NAME = 'rocmlir_metrics.txt'
 DIRECTIONS = ['-F 1', '-F 2', '-F 4']
-DATA_TYPES = ['conv', 'convfp16', 'convint8']
+DATA_TYPES = ['conv', 'convfp16', 'convfp8_fp8', 'convint8']
```
umangyadav: Do we have fp8 configs that needs to be tuned inside rocMLIR ?

djramic: It will run on ResNet50 and Unet configs from the conv-configs file on weekly CI.  Do you think we should expand that list?

umangyadav: I am asking if they are listed in any of the config files ? For example, i do not see `convfp8_fp8` here
https://github.com/ROCm/rocMLIR/blob/develop/mlir/utils/performance/sdxl-conv-configs

djramic: I think the sdxl-conv-config is an informational file, and I'm not sure why it has a predefined data type. Also, I haven't noticed it being used anywhere. 
In the configs files we use, there isn't a predefined data type, and we leave for the tuningRuuner script to assign it.
For example: 
-n 256 -c 1024 -H 14 -W 14 -k 2048 -y 1 -x 1 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -g 1 -m conv -t 1
so, this PR enables generating fp8 tuning keys for all problem from conv-configs.

dhernandez0: why fp8_fp8? convi8 is not i8_i32, maybe we can follow that style

dhernandez0 (approved):

umangyadav (approved):

PR #1739, opened 2025-02-10
Add bf16 to tuning runner

djramic:
This PR adds an option to select the bf16 type when running the tuning runner script.
close: https://github.com/ROCm/rocMLIR-internal/issues/1707

### DIFF
--- a/mlir/utils/performance/perfRunner.py
+++ b/mlir/utils/performance/perfRunner.py
@@ -32,9 +32,9 @@ DIRECTIONS = ['-F 1', '-F 2', '-F 4']
 DATA_TYPES = ['conv', 'convfp16', 'convint8']
 LAYOUTS = ['NHWC', 'NCHW']
 
-DATA_TYPES_GEMM = ['f32', 'f16', 'i8', 'fp8']
+DATA_TYPES_GEMM = ['f32', 'f16', 'bf16', 'i8', 'fp8']
 DATA_TYPES_ATTENTION = ['f32', 'f16']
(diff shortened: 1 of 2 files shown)


mlir/utils/performance/perfRunner.py
```diff
@@ -32,9 +32,9 @@
 DATA_TYPES = ['conv', 'convfp16', 'convint8']
 LAYOUTS = ['NHWC', 'NCHW']
 
-DATA_TYPES_GEMM = ['f32', 'f16', 'i8', 'fp8']
+DATA_TYPES_GEMM = ['f32', 'f16', 'bf16', 'i8', 'fp8']
 DATA_TYPES_ATTENTION = ['f32', 'f16']
```
dhernandez0: add bf16 to attention as well

mlir/utils/performance/perfRunner.py
```diff
@@ -32,9 +32,9 @@
 DATA_TYPES = ['conv', 'convfp16', 'convint8']
```
dhernandez0: convbf16 ?

dhernandez0: also, could you add convf8?

mlir/utils/performance/perfRunner.py
djramic: Done

mlir/utils/performance/perfRunner.py
djramic: Done

dhernandez0 (approved):

PR #1740, opened 2025-02-11
[DRAFT] Heuristic for Split-K (full tuning only)

dhernandez0:
In this PR we add a heuristic for Split-K full tuning. The idea is to filter out split-k values based on:
- occupancy (number of waves / numCU*numEUPerCU)
- padding added to K (due to split-k)
- update work imbalance to use waves instead of blocks

When occupancy is lower than 1.0, we directly return the "best" split-k factor according to imbalance. In practice, this filters out a lot of cases where we don't have to iterate over different split-k values and in some cases, we don't iterate at all.

I've checked the results and we keep the same accuracy for conv-configs, gemm-configs, sdxl-conv, sdxl-gemm (MI300X).

Moving back to draft because we have some regressions. We need to work a bit more on the heuristic.

### DIFF
--- a/mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
+++ b/mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
@@ -96,97 +96,167 @@ static void createAttnTuningRangeBF(TuningParamSet *newSpace,
   }
 }
 
-static double computeWorkImbalance(GemmSize origGemmSize, int32_t gemmMPerBlock,
-                                   int32_t gemmNPerBlock, int32_t gemmKPerBlock,
-                                   int32_t kPack, uint32_t numCUs,
-                                   int32_t splitKFactor = 1) {
-  const InitParams params{gemmMPerBlock, gemmNPerBlock, gemmKPerBlock};
-  const GemmSize gemmSize =
-      calculatePaddedGemmSize(params, origGemmSize, kPack);
+static double computeOccupancy(GemmSize gemmSize, int32_t gemmMPerBlock,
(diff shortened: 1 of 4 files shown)


mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -967,7 +1043,10 @@ bool isSplitKRequested(ModuleOp mod, StringRef perfConfig) {
 RocmlirSplitKSelectionLikelihood isSplitKFaster(int64_t gDim, int64_t mDim,
```
dhernandez0, line 1048: I think migraphx does not use this API call. So, for now let's leave this as future work. This needs to be investigated as well.

dhernandez0:
After running migxmlirperf.bsh with the following env variables:
```bash
MIGRAPHX_MLIR_ENABLE_SPLITK=1 MIGRAPHX_MLIR_USE_SPECIFIC_OPS=attention,fused,convolution,dot
```

We get performance regressions on:
- bert_base_cased_1.onnx (--fill1 input_ids --input-dim @input_ids 1 32 --exhaustive-tune), 0.7769ms vs 0.6785ms (develop)
- distilgpt2_1_fp16_gpu.onnx (--fill1 input_ids --input-dim @input_ids 1 32 --exhaustive-tune), 0.3615ms vs 0.3202ms (develop)
-  gpt2-10.onnx (--batch 1 --exhaustive-tune), 1.0527ms vs 0.9525ms (develop)

So, I think this heuristic needs to be refined a bit more to make sure we don't introduce performance regressions.

PR #1741, opened 2025-02-14
Fix performance for non-standard layouts

dhernandez0:
This PR fixes some issues for non-standard layouts, especially for convolutions. There are places in the codebase (convtogemm for example) where we assume dimensions are sorted [slowest, ..., fastest] where fastest is the fastest changing dimension in the memory layout.

I've implemented a pass that sorts the dimensions according to the underlying memory layout and adds a transform to make sure this assumption is true. This is done at the end of the highlevel pipeline.

There are a few TODOs added in this PR:
- failing tests due to https://github.com/ROCm/rocMLIR-internal/issues/1746 (fixed in this PR: https://github.com/ROCm/rocMLIR/pull/1744)
- getLowerSubDimensions partial support for TransformType::Broadcast. If there are more than one subdims, size will be incorrect. However, we don't use "size" for this pass, and the other place where it's used (reduction output), broadcast is not allowed as we learnt recently
- we skip the new pass (for convs) if C is slower chaning than spatial dimensions (H and W for 2D conv). This is because in convtogemm we reorder filter and output layouts to be the same as input (because we want gemmK to be the same for all). And for nchw layout, we end up forcing nhwc which might at least vectorize the loads for the weights. This is a weird way of doing this, we should make this logic more explicit in the future PR. I'm doing this to avoid regressions for now.

You can see improvement of this PR here: https://github.com/ROCm/rocMLIR-internal/issues/1709#issuecomment-2665367202

migraphx perf CI passed correctly: https://github.com/ROCm/AMDMIGraphX/pull/3831#issuecomment-2669152044


mlir/include/mlir/Conversion/RocMLIRPasses.td
```diff
@@ -138,4 +138,21 @@ def EmulateFp8ExtTruncPass : Pass<"emulate-fp8-ext-trunc"> {
   ];
 }
 
+//===----------------------------------------------------------------------===//
+// SortDimensionsMemoryLayout
```
umangyadav: This pass isn't doing conversion but it is rather a "transform". Looks like `mlir/lib/Dialect/Rock/transforms` is a better place

dhernandez0: you are right; however, it seems weird to add it there as this is a highlevel pass, not kernel pipeline. And I think all (or most) of the passes in mlir/lib/Dialect/Rock/transforms are passes that are called from buildKernelPipeline(). However, if there's no other idea I'll follow your suggestion and move it there.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -2483,26 +2491,46 @@ mlir::rock::getLowerSubDimensions(OpBuilder &b, ArrayAttr transformAttrs,
...
+                                      << upperDim << " to " << lowDim << "\n");
+            }
+          }
+        } break;
         case TransformType::ConstDim:
         case TransformType::AddDim: {
           // Nothing to do
-          break;
-        }
+        } break;
```
umangyadav: remove empty `{ }`

umangyadav (commented):
Looks good overall. Thanks

mlir/test/Conversion/SortDimensionsMemoryLayout/sort-dimensions-memory-layout.mlir
```diff
@@ -0,0 +1,94 @@
+// RUN: rocmlir-opt --rock-sort-dimensions-memory-layout %s -verify-diagnostics -o -| FileCheck %s
```
umangyadav: These tests should also be moved to `mlir/test/Dialect/Rock`

dhernandez0: done

umangyadav (approved):

PR #1745, opened 2025-02-19
[TOSA] Set `accType` to Float16 for the Fp8 types 

umangyadav:
https://github.com/llvm/llvm-project/pull/121466/files 

Tosa v1.0 spec introduced `accType` for a few TOSA ops. It requires accType for the float8 convolution to be fp16, otherwise verifier complains about. 

Fp16 accumulator may not be the right choice for fp8 dtypes. It can overflow. 

rocMLIR is not using accType attribute therefore, just set it and ignore it.

### DIFF
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -361,7 +361,10 @@ LogicalResult ConvConverter<ConvType>::matchAndRewrite(
     accType = rewriter.getF32Type();
     // accType is not used by rocMLIR when converting tosa to rock.
     // accType for Float8 type is required to be Float16 as per TOSA v1.0 spec
-    // therefore just set it as required, it is being ignored anyways.
+    // therefore just set it as required, it is being ignored anyways for GPU
+    // lowering using rocMLIR. [Risk]: CPU may generate different results
+    // compared to GPU if accType gets used on CPU lowering path. Currently it
+    // seems none of the TosaToXYZ converter uses this attribute.
   } else if (isa<FloatType>(outElementTy) &&
              outElementTy.getIntOrFloatBitWidth() <= 8) {
     accType = rewriter.getF16Type();
--- a/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
+++ b/mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
@@ -25,6 +25,7 @@
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/Dialect/Tosa/IR/TosaOps.h"
 #include "mlir/Dialect/Tosa/Utils/QuantUtils.h"
+#include "mlir/IR/BuiltinTypes.h"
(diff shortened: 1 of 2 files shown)


mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -355,12 +356,18 @@ LogicalResult ConvConverter<ConvType>::matchAndRewrite(
 
   // Determine the accumulation type based on the output type.
   Type accType;
-  if (isa<FloatType>(outElementTy)) {
+  if (isa<FloatType>(outElementTy) &&
+      outElementTy.getIntOrFloatBitWidth() >= 16) {
     accType = rewriter.getF32Type();
+    // accType is not used by rocMLIR when converting tosa to rock.
+    // accType for Float8 type is required to be Float16 as per TOSA v1.0 spec
+    // therefore just set it as required, it is being ignored anyways.
```
dhernandez0: when comparing our code vs tosa, this won't be ignored. However, I guess this was already the case? accumulation was done in f16 even before the change?

umangyadav: >when comparing our code vs tosa, this won't be ignored.  

That's true. For the numbers we generate in rocmlir-gen, fp16 accumulator should be sufficient. 

>However, I guess this was already the case? accumulation was done in f16 even before the change?

WMMA uses fp32 accumulator

dhernandez0: I see, thanks. I guess we have no alternative if tosa accumulates in f16. Approving the PR, just a nit:
nit: update the comment to mention this.

dhernandez0 (approved):

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
```diff
@@ -355,12 +356,18 @@ LogicalResult ConvConverter<ConvType>::matchAndRewrite(
 
   // Determine the accumulation type based on the output type.
   Type accType;
-  if (isa<FloatType>(outElementTy)) {
+  if (isa<FloatType>(outElementTy) &&
+      outElementTy.getIntOrFloatBitWidth() >= 16) {
```
dhernandez0, line 360: why is this needed?

umangyadav: Tosa spec allows Fp32 accumulator for 16bit and 32bit float types.

mlir/lib/Conversion/MIGraphXToTosa/MIGraphXToTosa.cpp
umangyadav: Done. Thanks. 
I've also asked here, IMO tosa should allow fp32 accumulator for fp8 types.
 https://discourse.llvm.org/t/rfc-tosa-dialect-increment-to-v1-0/83708/7?u=umangyadav

dhernandez0: thanks

PR #1747, opened 2025-02-19
Add E2E test for the OCP Fp8 fused kernel with QuantizeLinear and DeQuantizeLinear

umangyadav:
Related to https://github.com/ROCm/rocMLIR-internal/issues/1718

rocmlir-gen can not produce host code correctly for this kind of fused kernel at present. Therefore directly run on GPU with "fixed" data and compare against expected results.


dhernandez0 (approved):

dhernandez0 (commented):
Is there any arch we shouldn't this on? or will it run on all of them? This might not work if f8 is Nanoo?

umangyadav:
> Is there any arch we shouldn't this on? or will it run on all of them? This might not work if f8 is OCP?

It should be able to run on all of them. EmulateFp8ExtTrunc pass should take care of arches that do not support OCP FP8

dhernandez0:
> > Is there any arch we shouldn't this on? or will it run on all of them? This might not work if f8 is OCP?
> 
> It should be able to run on all of them. EmulateFp8ExtTrunc pass should take care of arches that do not support OCP FP8

I see, it would run the non-accel path I guess

PR #1748, opened 2025-02-19
Fixes for group conv emit-key

dhernandez0:
rocmlir-gen --emit-tuning-key is returning the wrong problem key for group convs.

See the result of: 
```bash
bin/rocmlir-gen --operation conv -t f16 --arch gfx90a:sramecc+:xnack- --num_cu 104 --fil_layout gkc01 --in_layout ngc01 --out_layout ngk01 --batchsize 64 --in_channels 256 --in_h 20 --in_w 20 --out_channels 256 --fil_h 7 --fil_w 7 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h 3 --padding_w 3 --groupsize 256 --kernel-repeats 1 --perf_config=v2:32,256,2,32,32,4,1,1,1 | bin/rocmlir-gen --emit-tuning-key -
```

is:
```
convfp16 -F 1 -f GNC01 -I NGC01 -O NGC01 -n 64 -c 1 -H 20 -W 20 -k 1 -y 7 -x 7 -p 3 -q 3 -u 1 -v 1 -l 1 -j 1 -g 256
```

And c and k must be 256. So, we are tuning for the wrong convs.

### DIFF
--- a/mlir/utils/performance/perfRunner.py
+++ b/mlir/utils/performance/perfRunner.py
@@ -314,7 +314,8 @@ class ConvConfiguration(PerfConfiguration):
     def computeTFlops(self, ns):
         # NaN will propagate as expected
         # Repeats are handled by the fact that we're using avarageNs
-        return (2.0 * self.n * self.c * self.k * self.ho * self.wo * self.y * self.x) / (float(ns) * 1e-9) / 1e12
+        assert(self.c % self.group == 0)
+        return (2.0 * self.n * (self.c//self.group) * self.k * self.ho * self.wo * self.y * self.x) / (float(ns) * 1e-9) / 1e12
 
     def tableEntry(self, nanoSeconds):
         # Future(kdrewnia): This can just be a dict literal on Python 3.7+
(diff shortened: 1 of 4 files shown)


umangyadav (approved):

mlir/utils/performance/perfRunner.py
```diff
@@ -314,7 +314,8 @@ class ConvConfiguration(PerfConfiguration):
     def computeTFlops(self, ns):
         # NaN will propagate as expected
         # Repeats are handled by the fact that we're using avarageNs
-        return (2.0 * self.n * self.c * self.k * self.ho * self.wo * self.y * self.x) / (float(ns) * 1e-9) / 1e12
+        assert(self.c % self.group == 0)
+        return (2.0 * self.n * (self.c//self.group) * self.k * self.ho * self.wo * self.y * self.x) / (float(ns) * 1e-9) / 1e12
```
umangyadav, line 319: why divide with `self.group` ?

dhernandez0: This is to compute number of fp operations. I think if you do group conv, you divide input channels/ group and out channels /group. Then do this “group” times, so multiply by group. So group/(2*group).

dhernandez0: Actually we could add an assert so that out  channels is also divisible by group. That should be a requirement as well.

PR #1750, opened 2025-02-20
Enable output swizzle for multiple outputs

dhernandez0:
This PR enables the output swizzle pass for multiple outputs. For example, this case below gets a 1.45x speed up on gfx90a because it can run the output swizzle pass:

```mlir
module {
  func.func @test(%arg0: !migraphx.shaped<2x1280x64x64xf16, 0x1x0x0>, %arg1: !migraphx.shaped<2x640x64x64xf16, 2621440x1x40960x640>, %arg2: !migraphx.shaped<1280x640x1x1xf16, 640x1x819200x819200>) -> (!migraphx.shaped<2x1280x64x64xf16, 5242880x1x81920x1280>, !migraphx.shaped<2x1280x64x64xf16, 5242880x1x81920x1280>) attributes {arch = "gfx90a:sramecc+:xnack-", kernel = "mixr"}
  {
    %0 = migraphx.convolution %arg1, %arg2 {perf_config="v2:64,64,8,32,32,8,1,1,1", dilation = [1, 1], group = 1 : i64, padding = [0, 0, 0, 0], padding_mode = 0 : i64, stride = [1, 1]} : <2x640x64x64xf16, 2621440x1x40960x640>, <1280x640x1x1xf16, 640x1x819200x819200> -> <2x1280x64x64xf16, 5242880x1x81920x1280>
    %1 = migraphx.add %0, %arg0 : <2x1280x64x64xf16, 5242880x1x81920x1280>, <2x1280x64x64xf16, 0x1x0x0> -> <2x1280x64x64xf16, 5242880x1x81920x1280>
    return %0, %1 : !migraphx.shaped<2x1280x64x64xf16, 5242880x1x81920x1280>, !migraphx.shaped<2x1280x64x64xf16, 5242880x1x81920x1280>
  }
}
```

I think there's no need to run perf comparisons as it's only enabling the pass for kernels with multiple outputs, what do you guys think? @umangyadav 
Also, we'll be adding output swizzle as a tuning param anyway.


mlir/lib/Dialect/Rock/Transforms/OutputSwizzle.cpp
```diff
@@ -211,18 +211,19 @@ struct ThreadwiseWriteAllRewritePattern
     VectorizationResult vectorRes =
         getMaxVectorization(destView, extraIdxCount);
     int64_t originalVectorLen = vectorRes.max;
+    int64_t elementsWrittenPerThread = math_util::gcd(dataPerThread, vectorLen);
 
-    if (vectorLen <= originalVectorLen) {
+    if (elementsWrittenPerThread <= originalVectorLen) {
```
dhernandez0, line 216: updated this because this is the actual value we use, not vectorLen.

mlir/lib/Dialect/Rock/Transforms/OutputSwizzle.cpp
```diff
@@ -205,24 +205,25 @@ struct ThreadwiseWriteAllRewritePattern
     if (auto elemVecType = dyn_cast<VectorType>(destElemType)) {
       LLVM_DEBUG(llvm::dbgs() << "ThreadwiseWriteAllOp saves a vector type"
                               << ", skipping swizzle\n");
-      return success();
+      return failure();
```
dhernandez0, line 208: I think we should return failure if we don't modify anything

umangyadav:
>I think there's no need to run perf comparisons as it's only enabling the pass for kernels with multiple outputs, what do you guys think? @umangyadav 

Yes it should fine. Multiple outputs it not a common case.

mlir/test/fusion/pr-e2e/mixr-multi-simple.mlir
```diff
@@ -0,0 +1,13 @@
+// RUN: rocmlir-gen -fut multi_simpl --arch %arch --clone-harness %s | rocmlir-driver -kernel-pipeline=migraphx | rocmlir-driver -host-pipeline=migraphx,highlevel | rocmlir-gen -ph -rand 1 -rand_type float -fut multi_simpl_wrapper --verifier clone -| rocmlir-driver -host-pipeline mhal -kernel-pipeline full | xmir-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_float16_utils%shlibext,%linalg_test_lib_dir/libmlir_c_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_async_runtime%shlibext --entry-point-result=void | FileCheck %s
+// ALLOW_RETRIES: 2
+
+// CHECK: [1 1 1]
+// CHECK: [1 1 1]
```
umangyadav, line 5:
```suggestion
// CHECK: [1 1 1]
```

umangyadav (approved):
Looks good, except minor comment

mlir/test/fusion/pr-e2e/mixr-multi-simple.mlir
dhernandez0: this test has two output tensors, that's why the check is duplicated

PR #1755, opened 2025-02-25
Use AddDim for unit input dimensions to help getMaxVectorization()

dhernandez0:
This PR fixes https://github.com/ROCm/rocMLIR-internal/issues/1747 
The last unmerge (lower), for example NxCxHxWxf16 to Pxf16, where P is the one-dimensional kernel input, is changed. So that, when some dimensions are 1, they are AddDimed instead of Unmerged. This helps getMaxVectorization() to get the right vector we can load from memory. 

For example, when a convolution filter has memory layout KxCxHxW and H=W=1, we now get the correct vectorization that depends on C. Because H and W will be AddDimed instead of Unmerged. Then, getMaxVectorization() doesn't take into account H and W to figure out the vectorization.

Due to this change we change getLowerSubDimensions() to not fail if no subdim info exists. Also, FoldBroadcast.cpp is affected, because if the batch is one, it will always be AddDim, so foldable.

I fixed a minor bug in SortDimensionsMemoryLayout to make sure we pass the perf_config in the new Conv/Gemm.


mlir/lib/Dialect/Rock/Transforms/FoldBroadcast.cpp
```diff
@@ -129,6 +131,10 @@ struct FoldBroadcast : public OpRewritePattern<rock::GemmOp> {
     if (trMap.getUpperBounds().size() != 3)
       return false;
 
+    // No need to fold if the batch size is 1
```
dhernandez0, line 134: adding this because now the batch is always "foldable" if it's 1, as it comes from AddDim. But there's no need to fold a unit dimension.

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
```diff
@@ -1460,7 +1460,7 @@ static LogicalResult insertBlockwiseReduction(
   FailureOr<llvm::SmallDenseMap<int64_t, SmallVector<SubDimInfo>>>
       lowerSubDims =
           getLowerSubDimensions(rewriter, toBeReducedViews, gridOnlyDimIdxs);
-  if (failed(lowerSubDims)) {
```
umangyadav, line 1463: `getLowerSubDimensions` can instead return `failure()` if `lowerSubDims.value().empty()`. 

We may forget to check for emptyness on all invokes for `getLowerSubDimensions`

umangyadav: nit : I see typo on line 1465. Spelling on unsuccessful is wrong.

mlir/lib/Dialect/Rock/Transforms/FoldBroadcast.cpp
umangyadav: Is there a way to tell if `1` is because of broadcast or `addDim ?

e.g. i assume for the broadcast, strides would be zero. Not sure about `AddDim`

mlir/lib/Dialect/Rock/Transforms/AlignTiling.cpp
dhernandez0: re returning failure(), that's the way it was. But in SortDimensionsMemoryLayout.cpp it's ok if it's a one but I want to know if it's failed, to report it in a message. That's why I changed it.

mlir/lib/Dialect/Rock/Transforms/FoldBroadcast.cpp
dhernandez0: yes, getLowerSubDimensions(b, transformAttrs, {batchDim}) where batchDim=0 should return an empty list if it's AddDimed, and non-empty if it's broadcasted. However, I think for this pass, to maintain the same behavior, it's easier to just check if batch==1, then assume it's not foldable. Even if both A is adddimed and B is broadcasted, they are batch=1, so there's no performance gain in using this pass, I think.

mlir/lib/Dialect/Rock/utility/transformMapUtils.cpp
```diff
@@ -1629,8 +1646,35 @@ void mlir::rock::expandFlatFunctionArguments(
       logicalVal = arg;
       continue;
     }
-    TopDownTMBuilder flattener(b, nameList, logicalShapedTy.getShape(), loc);
-    flattener.unmerge("raw", 0, nameList, logicalShapedTy.getShape());
+    SmallVector<uint32_t> upperDims(logicalShapedTy.getRank());
+    std::iota(upperDims.begin(), upperDims.end(), 0);
+    SmallVector<uint32_t> nonUnitUpperDim;
```
umangyadav, line 1651: Looks like we don't have unit-tests for `transformMapUtils`
IMO we should have one. We rely a lot on transformMapUtils.

umangyadav (approved):

PR #1758, opened 2025-02-27
Add lookup for more layouts in PerfRunner and Add an option for verifying each perfConfig with tuningRunner

umangyadav:
Example : 

```
python3 ../mlir/utils/performance/tuningRunner.py --op conv --data-type f16 -c $(pwd)/configs.txt --verify-mode=cpu --verify-perf-configs
```

Prints following log : 

```
Tuning: convfp16 -F 1 -f GNC01 -I N01GC -O NGC01 -n 1 -c 128 -H 40 -W 40 -k 80 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
Verifying with perfConfig = v2:16,16,4,16,16,4,1,1,1
Verifying with perfConfig = v2:16,16,4,16,16,8,1,1,1
Verifying with perfConfig = v2:16,16,8,16,16,4,1,1,1
Verifying with perfConfig = v2:16,16,8,16,16,8,1,1,1
Verifying with perfConfig = v2:16,32,4,16,16,4,1,1,1
Verifying with perfConfig = v2:16,32,4,16,16,8,1,1,1
Verifying with perfConfig = v2:16,32,8,16,16,4,1,1,1
Verifying with perfConfig = v2:16,32,8,16,16,8,1,1,1
Verifying with perfConfig = v2:16,64,4,16,16,4,1,1,1
Verifying with perfConfig = v2:16,64,4,16,16,4,3,1,1
Verifying with perfConfig = v2:16,64,4,16,16,4,4,1,1
Verifying with perfConfig = v2:16,64,4,16,16,8,1,1,1
Verifying with perfConfig = v2:16,64,4,16,16,8,3,1,1
Verifying with perfConfig = v2:16,64,4,16,16,8,4,1,1
.
.
.
.
.
.
.
Tested 400 configs, best perf 22.740750379527213 TFlops 4611.0 ns on perf_config v2:32,64,2,32,32,8,1,1,1,1

```

### DIFF
--- a/mlir/utils/performance/perfRunner.py
+++ b/mlir/utils/performance/perfRunner.py
@@ -376,10 +376,10 @@ class ConvConfiguration(PerfConfiguration):
 
     MLIR_FILTER_LAYOUTS = {"NCHW": "kcyx", "NCHWG": "kcyxg", "NHWC": "kyxc", "NHWCG": "kyxcg",
                            "NC01": "kc01", "NC01G": "kc01g", "N01C": "k01c", "N01CG": "k01cg",
-                           "GNC01":"gkc01"}
+                           "GNC01":"gkc01", "GN01C":"gk01c"}
     MLIR_OUTPUT_LAYOUTS = {"NCHW": "nkhw", "NCHWG": "nkhwg", "NHWC": "nhwk", "NHWCG": "nhwkg",
                            "NC01": "nk01", "NC01G": "nk01g", "N01C": "n01k", "N01CG": "n01kg",
-                           "NGC01":"ngk01"}
+                           "NGC01":"ngk01", "N01GC": "n01gk"}
     filterLayoutMap = {'N':'k', 'C':'c', 'H':'0', 'W':'1', 'G':'g'}
     inputLayoutMap = {'N':'n', 'C':'c', 'H':'0', 'W':'1', 'G':'g'}
     outputLayoutMap = {'N':'n', 'C':'k', 'H':'0', 'W':'1', 'G':'g'}
--- a/mlir/utils/performance/tuningRunner.py
+++ b/mlir/utils/performance/tuningRunner.py
@@ -119,7 +119,7 @@ def getWinningConfig(tuningOutput, testVector, config, allData, paths: Path, opt
         theseTFlops = entry['TFlops']
         # verify that each perfConfig passes accuracy verification
         if options.verifyPerfConfigs:
-            if options.verifyMode == "None":
+            if options.verifyMode == "none":
                 print(f"Use of `--verify-perf-configs` should happen in conjuction with `--verify-mode`. Please pass `--verify-mode=cpu` or `--verify-mode=gpu` flag")
                 sys.exit()
             else:
--- a/mlir/utils/performance/tuningRunner.py
+++ b/mlir/utils/performance/tuningRunner.py
@@ -116,7 +116,7 @@ def getWinningConfig(tuningOutput, testVector, config, allData, paths: Path, opt
         entry = config.tableEntry(nanoSeconds)
         allData.append(entry)
         theseTFlops = entry['TFlops']
-        ## verify that each perfConfig passes accuracy verification
+        # verify that each perfConfig passes accuracy verification
         if options.verifyPerfConfigs:
             if options.verifyMode == "None":
                 print(f"Use of `--verify-perf-configs` should happen in conjuction with `--verify-mode`. Please pass `--verify-mode=cpu` or `--verify-mode=gpu` flag")
--- a/mlir/utils/performance/tuningRunner.py
+++ b/mlir/utils/performance/tuningRunner.py
@@ -121,13 +121,13 @@ def getWinningConfig(tuningOutput, testVector, config, allData, paths: Path, opt
         if options.verifyPerfConfigs:
             if options.verifyMode == "none":
                 print("Use of `--verify-perf-configs` should happen in conjuction with `--verify-mode`. Please pass `--verify-mode=cpu` or `--verify-mode=gpu` flag")
-                return None, None 
+                sys.exit(1) 
             else:
                 verifyNs = verifyKernelWithPerfConfig(perfConfig, config, paths, options)
                 if np.isnan(verifyNs):
                     # Verification failed, abort the loop
                     print(f"verification failed on : {testVector} : {perfConfig}", file=sys.stderr)
-                    return None, None
+                    sys.exit(1)
 
         if not np.isnan(theseTFlops) and theseTFlops > maxTFlops:
             maxTFlops = theseTFlops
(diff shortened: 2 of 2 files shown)


mlir/utils/performance/perfRunner.py
```diff
@@ -376,10 +376,10 @@ def generateMlirDriverCommandLine(self, rocmlir_gen_flags):
 
     MLIR_FILTER_LAYOUTS = {"NCHW": "kcyx", "NCHWG": "kcyxg", "NHWC": "kyxc", "NHWCG": "kyxcg",
                            "NC01": "kc01", "NC01G": "kc01g", "N01C": "k01c", "N01CG": "k01cg",
-                           "GNC01":"gkc01"}
+                           "GNC01":"gkc01", "GN01C":"gk01c"}
```
dhernandez0, line 379: in another PR we can just generalize this, it's fine for now

mlir/utils/performance/tuningRunner.py
```diff
@@ -36,6 +36,7 @@ class Options:
     numCU: int
     rocmlir_gen_flags: str
     verifyMode: str
+    verifyPerfConfigs: bool
```
dhernandez0, line 39: nit: AFAIK, we don't use verifyMode anywhere, so you can also just change verifyMode behavior.

mlir/utils/performance/tuningRunner.py
```diff
@@ -115,6 +116,14 @@ def getWinningConfig(tuningOutput, config, allData, options: Options):
         entry = config.tableEntry(nanoSeconds)
         allData.append(entry)
         theseTFlops = entry['TFlops']
+        ## verify that each perfConfig passes accuracy verification
```
dhernandez0: nit: double "#"

mlir/utils/performance/tuningRunner.py
```diff
@@ -302,6 +311,11 @@ def main(args=None):
         choices=["none", "cpu", "gpu"],
         help="How to verify the winning tuned kernel")
 
+    parser.add_argument("--verify-perf-configs",
```
dhernandez0, line 319: do we want to provide cpu and gpu options as well?

mlir/utils/performance/tuningRunner.py
dhernandez0: well, we use it in CI I think

mlir/utils/performance/tuningRunner.py
umangyadav: it is controled by `verify-mode`

mlir/utils/performance/tuningRunner.py
umangyadav: It is used and i think it is useful to have `verifyMode` as a separate flag.

mlir/utils/performance/perfRunner.py
umangyadav: agreed

mlir/utils/performance/tuningRunner.py
dhernandez0: ah I see, that was not obvious to me, maybe we can add that in the description?

mlir/utils/performance/tuningRunner.py
umangyadav:
```suggestion
        # verify that each perfConfig passes accuracy verification
```

dhernandez0 (approved):

dhernandez0:
nit: we can also take into account --compact_print to avoid printing "Verifying with perfConfig...". Only print if it fails

mlir/utils/performance/tuningRunner.py
umangyadav: Added

mlir/utils/performance/tuningRunner.py
```diff
@@ -115,6 +116,18 @@ def getWinningConfig(tuningOutput, config, allData, options: Options):
         entry = config.tableEntry(nanoSeconds)
         allData.append(entry)
         theseTFlops = entry['TFlops']
+        # verify that each perfConfig passes accuracy verification
+        if options.verifyPerfConfigs:
+            if options.verifyMode == "None":
```
dhernandez0: I think it's "none"

umangyadav:
> nit: we can also take into account --compact_print to avoid printing "Verifying with perfConfig...". Only print if it fails

Done

mlir/utils/performance/tuningRunner.py
```diff
@@ -115,6 +117,18 @@ def getWinningConfig(tuningOutput, config, allData, options: Options):
...
+                print(f"Use of `--verify-perf-configs` should happen in conjuction with `--verify-mode`. Please pass `--verify-mode=cpu` or `--verify-mode=gpu` flag")
+                sys.exit()
+            else:
+                verifyNs = verifyKernelWithPerfConfig(perfConfig, config, paths, options)
+                if np.isnan(verifyNs):
+                    # Verification failed, abort the loop
+                    print(f"verification failed on : {testVector} : {perfConfig}", file=sys.stderr)
+                return None, None
```
dhernandez0: will it fail eventually if we return None, None here?

umangyadav: identation error. It should return immedidately on fail. Changed that

dhernandez0: thanks! what I mean is, will it fail if it returns None, None? otherwise we can just call sys.exit(1) as above

mlir/utils/performance/tuningRunner.py
```diff
@@ -115,6 +117,18 @@ def getWinningConfig(tuningOutput, config, allData, options: Options):
         entry = config.tableEntry(nanoSeconds)
         allData.append(entry)
         theseTFlops = entry['TFlops']
+        # verify that each perfConfig passes accuracy verification
+        if options.verifyPerfConfigs:
+            if options.verifyMode == "none":
+                print(f"Use of `--verify-perf-configs` should happen in conjuction with `--verify-mode`. Please pass `--verify-mode=cpu` or `--verify-mode=gpu` flag")
```
dhernandez0: nit: no need to use f""

dhernandez0 (approved):

mlir/utils/performance/tuningRunner.py
umangyadav:
```suggestion
                print("Use of `--verify-perf-configs` should happen in conjuction with `--verify-mode`. Please pass `--verify-mode=cpu` or `--verify-mode=gpu` flag")
```

mlir/utils/performance/tuningRunner.py
umangyadav: It will fail eventually but i think better to just do `sys.exit(1)`. 
Changed that one.

PR #1760, opened 2025-02-27
Fix dependency graph creation in RockPipeline and not generate loops with negative iterations

umangyadav:
This fixes two bugs in RockPipeline. 

1. It wasn't creating correct dependency graph while generating Schedule. It was only using "MultiBuffers" while creating dependency graph for generating schedule which was generating incorrect IR. 
2. For certain cases it was generating Loops with negative iterations. This PR adds some checks to avoid pipelining in such cases.

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
@@ -668,10 +668,11 @@ void RockPipeline::runOnOperation() {
         rock::computeConstDiff(forOp.getLowerBound(), forOp.getUpperBound());
     assert(isConstantIntValue(forOp.getStep(), 1) &&
            "Step size other one is not permitted in rock-pipeline");
-    if (!maybeNumIterations.has_value())
+    if (!maybeNumIterations.has_value()) {
       emitError(loc,
                 "Number of iterations are unknown while doing rock-pipeline\n");
-
+      return signalPassFailure();
+    }
     adjustInitiationInterval(maybeNumIterations.value(), numStages, ii);
 
     // Insert the barriers as new stages
--- a/mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
@@ -102,11 +102,11 @@ AddressSpace getAddressSpace(MemrefTypedValue val) {
 MemoryAccessType getOperandAccessType(Operation *op, Value operand) {
   if (hasEffect<MemoryEffects::Write>(op, operand)) {
     return MemoryAccessType::WRITE;
-  } else if (hasEffect<MemoryEffects::Read>(op, operand)) {
-    return MemoryAccessType::READ;
-  } else {
-    return MemoryAccessType::UNKNOWN;
   }
+  if (hasEffect<MemoryEffects::Read>(op, operand)) {
+    return MemoryAccessType::READ;
+  }
+  return MemoryAccessType::UNKNOWN;
 }
 
 // Simple rewrite pass to remove the stages and backward barriers in the
@@ -253,11 +253,13 @@ void createSchedule(SmallVector<rock::StageOp> &stages,
   // Create the dependency graph
   DagType dag = createDependencyGraph(stages, resources);
 
-  // Start building the schedules
-  //
-  // Since we accept the stages from the user, we don't need to do any
-  // analysis to determine what goes in each stage. We only have to group things
-  // in set of stages of length II.
+  // Definition of initiation interval (II)
+  // Initiation interval is defind by number of cycles in each iteration of a
+  // loop. Only one cycle is counted for parallel stages. Assume each stage
+  // executes in one cycle. Start building the schedules Since we accept the
+  // stages from the user, we don't need to do any analysis to determine what
+  // goes in each stage. We only have to group things in set of stages of length
+  // II.
   //  For instance, consider the following unpipelined schedule. The column `t`
   //  represents
   // the time slot, and the subsequent columns represents the iterations.
@@ -282,7 +284,7 @@ void createSchedule(SmallVector<rock::StageOp> &stages,
   //  +===+=========+
   // In this case, we reduced the time slots to 3, and we have 2 set of stages
   // runnning in parallel. Please note that conflicts can only happen between S0
-  // and S3. If we increase II, we generate the following pipeline:
+  // and S3. If we decrease II, we generate the following pipeline:
   //  +t\i+=== 0 ===++=== 1 ===+
   //  + 0 +== S0  ==++== S2  ==+
   //  +===+=========++=========+
@@ -385,7 +387,7 @@ void createSchedule(SmallVector<rock::StageOp> &stages,
 
     // Add the parallel stages
     for (auto stage : parallelStages) {
-      schedule.push_back({stage, stageIter[stage]});
+      schedule.emplace_back(stage, stageIter[stage]);
     }
   }
 }
@@ -393,15 +395,15 @@ void createSchedule(SmallVector<rock::StageOp> &stages,
 // Prune a dependency graph taking into account multi-buffers. Since
 // multi-buffers are logically different for each iteration, if the dependency
 // on a multi-buffer spans multiple iteration then it can be pruned
-DagType pruneGraph(DagType dag) {
+DagType pruneGraph(const DagType& dag) {
   DagType prunedGraph;
   // Multibuffers have the logical property of being unique for each iteration
   // of the loop Hence, if we know we are dealing with a multi-buffer and the
   // dependency concerns two different iteration. In other words, if stageA
   // accesses LDS in iteration i and stageB accesses LDS in iteration j stageA
   // and stageB have no dependencies as long as i!=j
-  for (auto [sink, edges] : dag) {
-    for (auto [source, deps] : edges) {
+  for (auto [source, edges] : dag) {
+    for (auto [sink, deps] : edges) {
       DenseSet<std::pair<rock::GpuAllocOp, DependencyType>> newDeps;
       for (auto [alloc, type] : deps) {
         if (getAddressSpace(alloc) != gpu::AddressSpace::Workgroup)
@@ -409,7 +411,7 @@ DagType pruneGraph(DagType dag) {
         newDeps.insert({alloc, type});
       }
       if (!newDeps.empty())
-        prunedGraph[sink][source] = newDeps;
+        prunedGraph[source][sink] = newDeps;
     }
   }
   return prunedGraph;
@@ -461,7 +463,7 @@ void placeBarriers(IRRewriter &rewriter, Location loc, scf::ForOp forOp,
     timeSlot++;
   }
 
-  // Algorithm for barrier placment:
+  // Algorithm for barrier placement:
   // a. Add forward barriers to address the dependency in the basic block
   // b. Add backward barriers to account for loop carried dependency
   // c. Add empty stages to make the pipeline balanced, so that we can double up
@@ -577,6 +579,9 @@ void RockPipeline::runOnOperation() {
 
   // Always (try to) multi-buffer by one and store the new
   // allocs in a set
+  // multiBuffering is only happening on LDS Buffers as "multiBuffer" checks for
+  // "i8" dtype store multibuffers in "multiAllocs" store all buffers including
+  // private and global in "resources"
   llvm::SetVector<rock::GpuAllocOp> multiAllocs;
   llvm::SetVector<rock::GpuAllocOp> resources;
   for (auto alloc : singleAllocs) {
@@ -643,6 +648,8 @@ void RockPipeline::runOnOperation() {
 
     // Insert the barriers as new stages
     SmallVector<rock::StageOp> extendedStages;
+    // use "multiAllocs" to place LDS barriers, no need to explicitly place
+    // barriers for registers or globals
     placeBarriers(rewriter, loc, forOp, stages, multiAllocs, extendedStages,
                   ii);
     auto maybeNumIterations =
@@ -657,6 +664,7 @@ void RockPipeline::runOnOperation() {
     }
 
     ScheduleType schedule;
+    // use all "resources" to generate dependency graph and generate schedule
     createSchedule(extendedStages, resources, ii, schedule, multiBufferFactors);
 
     RewritePatternSet patterns(&getContext());
--- a/mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
@@ -710,8 +710,8 @@ void RockPipeline::runOnOperation() {
   {
     if (removeStages) {
       RewritePatternSet patterns(&getContext());
-      patterns.add<RemoveStagesRewritePattern, PushBarrierDownRewritePattern>(
-          &getContext());
+      patterns.add<RemoveStagesRewritePattern, PushBarrierDownRewritePattern,
+                   RemoveBackToBackBarriersRewritePattern>(&getContext());
       (void)applyPatternsGreedily(getOperation(), std::move(patterns));
     }
   }
--- a/mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
@@ -469,8 +469,8 @@ void placeBarriers(IRRewriter &rewriter, Location loc, scf::ForOp forOp,
   // a. Add forward barriers to address the dependency in the basic block
   // b. Add backward barriers to account for loop carried dependency
   // c. Add empty stages to make the pipeline balanced, so that we can double up
-  //    the initiation interval and let the pipeline transformation automaticall
-  //    do the work for us
+  //    the initiation interval and let the pipeline transformation
+  //    automatically do the work for us
   DenseSet<rock::StageOp> forwardStages;
 
   // a. Place forward barriers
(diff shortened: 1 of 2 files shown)


mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -632,19 +642,42 @@ void RockPipeline::runOnOperation() {
...
+    if (!maybeNumIterations.has_value()) {
+      continue;
+    }
+    // if number of iterations are less than number of prologues that are going
+    // to be emitted, it will not result in correct output therefore do not do
+    // any loop pipelining this is achieved by setting II=numStages as we still
+    // want to correctly place forward and backward barriers
+    if (size_t(maybeNumIterations.value()) < numPrologues) {
+      ii = numStages;
+    }
```
umangyadav: TODO: instead of skipping loop pipelining, we can adjust `II` so that it would try best to do loop pipelining for smaller number of iterations.

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -253,11 +254,13 @@ void createSchedule(SmallVector<rock::StageOp> &stages,
...
 
-  // Start building the schedules
-  //
-  // Since we accept the stages from the user, we don't need to do any
-  // analysis to determine what goes in each stage. We only have to group things
-  // in set of stages of length II.
+  // Definition of initiation interval (II)
+  // Initiation interval is defind by number of cycles in each iteration of a
+  // loop. Only one cycle is counted for parallel stages. Assume each stage
+  // executes in one cycle. Start building the schedules Since we accept the
```
dhernandez0: New line here before "Start building the schedules", then new line again.

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -632,19 +642,42 @@ void RockPipeline::runOnOperation() {
     });
 
     if (stages.empty())
-      WalkResult::advance();
```
dhernandez0, line 635: nice catch! could be add a test for this?

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -577,19 +580,26 @@ void RockPipeline::runOnOperation() {
 
   // Always (try to) multi-buffer by one and store the new
   // allocs in a set
+  // multiBuffering is only happening on LDS Buffers as "multiBuffer" checks for
+  // "i8" dtype store multibuffers in "multiAllocs" store all buffers including
```
dhernandez0: dtype. Store multibuffers in "multiAllocs". Store all ...

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -632,19 +642,42 @@ void RockPipeline::runOnOperation() {
...
+    auto maybeNumIterations =
+        rock::computeConstDiff(forOp.getLowerBound(), forOp.getUpperBound());
+    assert(isConstantIntValue(forOp.getStep(), 1) &&
+           "Step size other one is not permitted in rock-pipeline");
+    if (!maybeNumIterations.has_value()) {
+      continue;
+    }
+    // if number of iterations are less than number of prologues that are going
+    // to be emitted, it will not result in correct output therefore do not do
+    // any loop pipelining this is achieved by setting II=numStages as we still
```
dhernandez0: any loop pipelining. This...

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -282,7 +285,7 @@ void createSchedule(SmallVector<rock::StageOp> &stages,
   //  +===+=========+
   // In this case, we reduced the time slots to 3, and we have 2 set of stages
   // runnning in parallel. Please note that conflicts can only happen between S0
-  // and S3. If we increase II, we generate the following pipeline:
+  // and S3. If we decrease II, we generate the following pipeline:
   //  +t\i+=== 0 ===++=== 1 ===+
```
dhernandez0, line 289: is the formatting of the example off? Not sure I understand it

mlir/test/Dialect/Rock/test_rock_pipeline.mlir
```diff
@@ -125,6 +125,61 @@ func.func @rock_pipeline_3_stages_ii_2(%input : memref<16xi8, #gpu.address_space
     return
 }
 
+// this test shouldn't pipeline loop but it would should add barriers and multibuffer by 1
```
dhernandez0: would should -> should

dhernandez0: multibuffer by 1?

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -632,19 +642,42 @@ void RockPipeline::runOnOperation() {
...
+    LLVM_DEBUG(DBGS() << "Number of parallel stages: " << numParallelStages
+                      << "\n");
+    LLVM_DEBUG(DBGS() << "Number of Prologues: " << numPrologues << "\n");
+    auto maybeNumIterations =
+        rock::computeConstDiff(forOp.getLowerBound(), forOp.getUpperBound());
+    assert(isConstantIntValue(forOp.getStep(), 1) &&
+           "Step size other one is not permitted in rock-pipeline");
+    if (!maybeNumIterations.has_value()) {
+      continue;
+    }
```
dhernandez0: nit: no need to use braces for this

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
dhernandez0, line 675: nit: no need to use braces for this

dhernandez0: please can you add the TODO to the code?

mlir/test/Dialect/Rock/test_rock_pipeline.mlir
umangyadav: rock pipeline pass always multibuffers by 1. 

it then calculates multiBuffers factors later and then updates the multibuffering.

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
umangyadav: I added logic to adjust II in this PR along with tests.

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
umangyadav: Done

mlir/test/Dialect/Rock/test_rock_pipeline.mlir
dhernandez0: what does it mean to multibuffer by 1?

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -253,11 +254,14 @@ void createSchedule(SmallVector<rock::StageOp> &stages,
   // Create the dependency graph
   DagType dag = createDependencyGraph(stages, resources);
 
-  // Start building the schedules
-  //
-  // Since we accept the stages from the user, we don't need to do any
-  // analysis to determine what goes in each stage. We only have to group things
-  // in set of stages of length II.
+  // Definition of initiation interval (II)
+  // Initiation interval is defind by number of cycles in each iteration of a
```
dhernandez0: nit: can we define cycle in this context?

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -253,11 +254,14 @@ void createSchedule(SmallVector<rock::StageOp> &stages,
...
+  // loop. Only one cycle is counted for parallel stages. Assume each stage
+  // executes in one cycle.
+  // Start building the schedules. Since we accept the
+  // stages from the user, we don't need to do any analysis to determine what
+  // goes in each stage. We only have to group things in set of stages of length
```
dhernandez0: seems confusing to use the word stage here, we are grouping stages in stages?

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -558,6 +562,26 @@ SmallVector<scf::ForOp> collectLoopLevels(mlir::func::FuncOp func) {
...
 }
 
+void adjustInitiationInterval(int64_t numIterations, size_t numStages,
+                              int64_t &ii) {
+  int64_t numParallelStages = llvm::divideCeil(numStages, ii);
+  // calculate number of prologue executions
+  int64_t numPrologues = numParallelStages - 1;
+  // if number of iterations are less than number of prologues that are going
+  // to be emitted, it will not result in correct output therefore increase II
+  // untill that condition becomes false. This can help achieve maximum loop
```
dhernandez0: nit: until

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -558,6 +562,26 @@ SmallVector<scf::ForOp> collectLoopLevels(mlir::func::FuncOp func) {
...
+  // pipelining
+  while (numIterations < numPrologues) {
+    ii++;
+    LLVM_DEBUG(DBGS() << "Adjusted II to  " << ii << "\n");
+    numParallelStages = llvm::divideCeil(numStages, ii);
+    numPrologues = numParallelStages - 1;
+  }
+  LLVM_DEBUG(DBGS() << "Number of parallel stages: " << numParallelStages
+                    << "\n");
+  LLVM_DEBUG(DBGS() << "Number of Prologues: " << numPrologues << "\n");
```
dhernandez0, line 577: can we print number of epilogues as well for completeness?

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -632,19 +662,30 @@ void RockPipeline::runOnOperation() {
...
 
     LLVM_DEBUG(DBGS() << "Number of stages: " << stages.size() << "\n");
     LLVM_DEBUG(DBGS() << "Initiation Interval: " << ii << "\n");
+    size_t numStages = stages.size();
+    auto maybeNumIterations =
+        rock::computeConstDiff(forOp.getLowerBound(), forOp.getUpperBound());
+    assert(isConstantIntValue(forOp.getStep(), 1) &&
+           "Step size other one is not permitted in rock-pipeline");
+    if (!maybeNumIterations.has_value())
+      continue;
```
dhernandez0: I think even if the number of iterations is variable, we should still add barriers that are needed. For example, if ii=2, barriers between LL and LS.

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -669,8 +710,8 @@ void RockPipeline::runOnOperation() {
   {
     if (removeStages) {
       RewritePatternSet patterns(&getContext());
-      patterns.add<RemoveStagesRewritePattern, PushBarrierDownRewritePattern>(
-          &getContext());
+      patterns.add<RemoveStagesRewritePattern, PushBarrierDownRewritePattern,
+                   RemoveBackToBackBarriersRewritePattern>(&getContext());
```
dhernandez0, line 713: why is this needed?

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -558,6 +562,26 @@ SmallVector<scf::ForOp> collectLoopLevels(mlir::func::FuncOp func) {
...
+void adjustInitiationInterval(int64_t numIterations, size_t numStages,
+                              int64_t &ii) {
+  int64_t numParallelStages = llvm::divideCeil(numStages, ii);
+  // calculate number of prologue executions
+  int64_t numPrologues = numParallelStages - 1;
+  // if number of iterations are less than number of prologues that are going
+  // to be emitted, it will not result in correct output therefore increase II
+  // untill that condition becomes false. This can help achieve maximum loop
+  // pipelining
+  while (numIterations < numPrologues) {
```
dhernandez0, line 569: I think we need to evaluate something like ii=numStages/numIterations and ii=numStages/(numIterations+1). It's not a big deal anyway, this loop should take few iterations.

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -558,6 +562,26 @@ SmallVector<scf::ForOp> collectLoopLevels(mlir::func::FuncOp func) {
...
+    ii++;
```
dhernandez0, line 570: so I guess because we increase ii, in practice we would never use more LDS buffers or register buffers? it's either the same or less? Then, I see no downside trying to do this

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -632,19 +662,30 @@ void RockPipeline::runOnOperation() {
     });
 
     if (stages.empty())
-      WalkResult::advance();
+      continue;
 
     LLVM_DEBUG(DBGS() << "Number of stages: " << stages.size() << "\n");
     LLVM_DEBUG(DBGS() << "Initiation Interval: " << ii << "\n");
+    size_t numStages = stages.size();
+    auto maybeNumIterations =
```
dhernandez0, line 667: nit: this is used inside placeBarriers, can we pass it instead of recompute it?

mlir/test/Dialect/Rock/test_rock_pipeline.mlir
```diff
@@ -92,15 +91,15 @@ func.func @rock_pipeline_3_stages_ii_2(%input : memref<16xi8, #gpu.address_space
     // CHECK: scf.for
       // CHECK: name = "__fwd_barrier__"
       // CHECK: rock.extract_multibuffer(%[[regAView]])
-      // CHECK name = "S0"
+      // CHECK: name = "S0"
```
dhernandez0, line 94: nice catch

mlir/test/Dialect/Rock/test_rock_pipeline.mlir
umangyadav: it means it adds
`rock.extract_multibuffer([view_1])` 

it only have one buffer to select from.  So it is equivalent of not doign multi-buffering.

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
umangyadav: >so I guess because we increase ii, in practice we would never use more LDS buffers or register buffers? it's either the same or less? 

Yes

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
umangyadav: Not sure i understand. Can you explain /

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
umangyadav: loop pipelining may generate back to back barriers. 
e.g. check `rock_pipeline_4_stages_ii_2` test in `test_rock_pipeline.mlir`

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
umangyadav: It will be same as number of prologues i think

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
umangyadav: to create schedule it need to put `II` number of stages in sequence and each group of `II` stages executes in parallel.

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
umangyadav: I dont' know how to best define it it is just symbolic cylce .  it later says assume it stage only takes one cycle.

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
umangyadav: Done.

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
umangyadav: It is a bit tricky to move `placeBarriers` before this check. 
placeBarriers need to know `II`. 
`II` is adjusted based on numIterations. 
Therefore `numIterations` needs to known before calling into `placeBarrier`.
It can probably be done, but for now i've placed "emitError" to make sure we are not hitting case of dynamic iterations.

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
```diff
@@ -632,19 +659,31 @@ void RockPipeline::runOnOperation() {
...
+        rock::computeConstDiff(forOp.getLowerBound(), forOp.getUpperBound());
+    assert(isConstantIntValue(forOp.getStep(), 1) &&
+           "Step size other one is not permitted in rock-pipeline");
+    if (!maybeNumIterations.has_value())
+      emitError(loc,
+                "Number of iterations are unknown while doing rock-pipeline\n");
```
dhernandez0, line 673: this won't make it crash I think, just emit an error an continue running.  We can do emitError then, return signalPassFailure();

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
dhernandez0: I mean that I think we can reduce the number of iterations to two. Because we can algebraically find that if we want numPrologues to be >= numIterations you have to try two "ii" values only. However, I'm happy if you leave it as it is now.

mlir/lib/Dialect/Rock/Transforms/RockPipeline.cpp
umangyadav: Added `signalPassFailure()`

dhernandez0 (approved):

PR #1762, opened 2025-03-04
Use hip-python API instead of rocm_agent_enumerator

dorde-antic:
Handle agent detection correctly by using hip-python API instead of rocm_agent_enumerator.
Add requirements.txt for external python libraries.

Resolves [#1760](https://github.com/ROCm/rocMLIR-internal/issues/1760)

### DIFF
--- a/mlir/test/common_utils/common.py
+++ b/mlir/test/common_utils/common.py
@@ -1,6 +1,6 @@
 import os
 import subprocess
-
+from hip import hip
 
 # Helper function to decode arch to its features
 # Keep this in sync with mlir/lib/Dialect/Rock/Generator/AmdArchDb.cpp:mlir::rock::lookupArchInfo
@@ -42,21 +42,27 @@ def get_arch_features(arch: str):
         pass
     return arch_features, support_mfma, support_wmma
 
+def hip_check(call_result):
+    err = call_result[0]
+    result = call_result[1:]
+    if len(result) == 1:
+        result = result[0]
+    if isinstance(err, hip.hipError_t) and err != hip.hipError_t.hipSuccess:
+        raise RuntimeError(str(err))
+    return result
 
-def get_agents(rocm_path):
-    if os.name != 'nt':
-        p = subprocess.run([rocm_path + "/bin/rocm_agent_enumerator", "-name"],
-                           check=True, stdout=subprocess.PIPE)
-        agents = set(x.decode("utf-8") for x in p.stdout.split())
-        if not agents or not all(agent.startswith("gfx") for agent in agents):
-            # TODO: Remove this workaround for a bug in rocm_agent_enumerator -name
-            # Once https://github.com/RadeonOpenCompute/rocminfo/pull/59 lands
-            q = subprocess.run([rocm_path + "/bin/rocm_agent_enumerator"],
-                               check=True, stdout=subprocess.PIPE)
-            agents = set(x.decode("utf-8") for x in q.stdout.split())
-        return set(a for a in agents if a != "gfx000")
-    else:
-        p = subprocess.run([rocm_path + "/bin/amdgpu_arch.exe"],
-                           check=True, stdout=subprocess.PIPE, shell=True)
-        return set(p.stdout.decode("utf-8").split())
+def get_agents():
+    agents = set()
+    device_count = hip_check(hip.hipGetDeviceCount())
+    for device in range(device_count):
+        props = hip.hipDeviceProp_t()
+        hip_check(hip.hipGetDeviceProperties(props,device))
+        agent = props.gcnArchName.decode('utf-8')
+        agents.add(agent)
 
+    return agents
+
+
+def is_xdlops_present() -> bool:
+    """This function checks whether a GPU with xdlops support is present"""
+    return any([agent.startswith("gfx9") for agent in get_agents()])
--- a/mlir/test/e2e/generateE2ETest.py
+++ b/mlir/test/e2e/generateE2ETest.py
@@ -35,18 +35,26 @@ import os
 import sys
 import getopt
 import glob
-import subprocess
+from hip import hip
+
+def hip_check(call_result):
+    err = call_result[0]
+    result = call_result[1:]
+    if len(result) == 1:
+        result = result[0]
+    if isinstance(err, hip.hipError_t) and err != hip.hipError_t.hipSuccess:
+        raise RuntimeError(str(err))
+    return result
 
 def getArch():
-    p = subprocess.run(["/opt/rocm/bin/rocm_agent_enumerator", "-name"], check=True,
-                       stdout=subprocess.PIPE)
-    agents = set(x.decode("utf-8") for x in p.stdout.split())
-    if not agents:
-        # TODO: Remove this workaround for a bug in rocm_agent_enumerator -name
-        # Once https://github.com/RadeonOpenCompute/rocminfo/pull/59 lands
-        q = subprocess.run(["/opt/rocm/bin/rocm_agent_enumerator"],
-                              check=True, stdout=subprocess.PIPE)
-        agents = set(x.decode("utf-8") for x in q.stdout.split() if x != b"gfx000")
+    agents = set()
+    device_count = hip_check(hip.hipGetDeviceCount())
+    for device in range(device_count):
+        props = hip.hipDeviceProp_t()
+        hip_check(hip.hipGetDeviceProperties(props,device))
+        agent = props.gcnArchName.decode('utf-8')
+        agents.add(agent)
+
     return agents
 
 def generate_option_list(prefixes: dict, table: list, key1: str, key2: str):
--- a/mlir/utils/performance/perfRunner.py
+++ b/mlir/utils/performance/perfRunner.py
@@ -18,6 +18,7 @@ from dataclasses import dataclass
 from typing import Optional, Dict, Tuple
 import numpy as np
 import pandas as pd
+from hip import hip
 
 import reportUtils
 from perfCommonUtils import Operation, GEMMLibrary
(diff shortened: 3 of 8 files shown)


mlir/test/common_utils/common.py
```diff
@@ -45,16 +47,15 @@ def get_arch_features(arch: str):
...
-        agents = set(x.decode("utf-8") for x in p.stdout.split())
-        if not agents or not all(agent.startswith("gfx") for agent in agents):
-            # TODO: Remove this workaround for a bug in rocm_agent_enumerator -name
-            # Once https://github.com/RadeonOpenCompute/rocminfo/pull/59 lands
-            q = subprocess.run([rocm_path + "/bin/rocm_agent_enumerator"],
-                               check=True, stdout=subprocess.PIPE)
-            agents = set(x.decode("utf-8") for x in q.stdout.split())
-        return set(a for a in agents if a != "gfx000")
+        agents = set()
+        _, device_count = hip.hipGetDeviceCount()
```
mirza-halilcevic: Consider doing some basic error checking for the HIP calls.

requirements.txt
```diff
@@ -0,0 +1,8 @@
+tomli==2.0.1
+nanobind>=2.4, <3.0
+numpy>=1.19.5, <=2.1.2
+pybind11>=2.10.0, <=2.13.6
+PyYAML>=5.4.0, <=6.0.1
+ml_dtypes>=0.1.0, <=0.5.0   # provides several NumPy dtype extensions, including the bf16
+hip-python
+scipy
```
dhernandez0: new line at the end

mlir/test/e2e/generateE2ETest.py
```diff
@@ -36,17 +36,18 @@
...
-    agents = set(x.decode("utf-8") for x in p.stdout.split())
-    if not agents:
-        # TODO: Remove this workaround for a bug in rocm_agent_enumerator -name
-        # Once https://github.com/RadeonOpenCompute/rocminfo/pull/59 lands
-        q = subprocess.run(["/opt/rocm/bin/rocm_agent_enumerator"],
-                              check=True, stdout=subprocess.PIPE)
-        agents = set(x.decode("utf-8") for x in q.stdout.split() if x != b"gfx000")
+    agents = set()
+    _, device_count = hip.hipGetDeviceCount()
+    for device in range(device_count):
```
dhernandez0, line 52: is it possible to call the function in common.py? same for other files

mlir/test/common_utils/common.py
```diff
@@ -45,16 +47,15 @@ def get_arch_features(arch: str):
...
+        for device in range(device_count):
+            props = hip.hipDeviceProp_t()
+            hip.hipGetDeviceProperties(props,device)
+            agent = props.gcnArchName.decode('utf-8')
+            agents.add(agent) if agent != "gfx000" else None
```
dhernandez0: does this still happen with the API? do we get gfx000?

mlir/utils/performance/perfRunner.py
```diff
@@ -1243,24 +1245,28 @@ def tuneMLIRKernels(configs, arch, numCU):
...
-    if r.returncode == 0:
-        return True
+    _, device_count = hip.hipGetDeviceCount()
+    for device in range(device_count):
+        props = hip.hipDeviceProp_t()
+        hip.hipGetDeviceProperties(props,device)
+        agent = props.gcnArchName.decode('utf-8')
+
+        if agent in xdlop_supported_gpus:
+            return True
```
dhernandez0: also add this function to wherever we put getArchs()

mlir/test/common_utils/common.py
```diff
@@ -1,5 +1,7 @@
 import os
 import subprocess
+from hip import hip
+from hip import hiprtc
```
umangyadav: I don't find hiprtc being used. Is it required to be imported ?

requirements.txt
```diff
@@ -0,0 +1,8 @@
+tomli==2.0.1
+nanobind>=2.4, <3.0
+numpy>=1.19.5, <=2.1.2
+pybind11>=2.10.0, <=2.13.6
+PyYAML>=5.4.0, <=6.0.1
+ml_dtypes>=0.1.0, <=0.5.0   # provides several NumPy dtype extensions, including the bf16
```
umangyadav: Why do you need all these packages for this PR ?

dhernandez0: I suggested to add requirements.txt in this PR to have a list of python dependencies. I think he listed all python libraries we use? @dorde-antic are these used by used or did you add all dependencies from external/llvm-project?

mlir/test/common_utils/common.py
dorde-antic: I added it when following the hip-python usage instructions, but seems that it is not required for our case. I will remove it and check locally if methods where we use it work without hiprtc.

requirements.txt
dorde-antic: I listed packages from both with versions same as in external/llvm-project. Should I remove those that are only in external, but keep others from the project?
Or I should just keep hip-python which is required for this PR?

mlir/test/common_utils/common.py
dorde-antic: It didn't notice that this happens when using API, but I have kept it since I saw that it was happening in the previous version of function. Should I remove this check `if agent != "gfx000" else None` ?

requirements.txt
dhernandez0: I think we can keep all libraries that are actually used for rocmlir (not what external uses). But keep the versions consistent with external.

mlir/test/common_utils/common.py
dhernandez0: if it's not needed, let's remove it.

mlir/test/e2e/generateE2ETest.py
dorde-antic: For now, in commit [Use hip-python API fixes](https://github.com/ROCm/rocMLIR/pull/1762/commits/f1dd5c45df0ed998565cdbff4fbbdbeb93ee06a6), as we talked we have duplicating where files are not in the same directory so that we don't use relative paths until we decide to handle it other way, and I've fixed where it could be done asap.

Maybe, for the future, we should consider having a package for a group of functions we need in different child-directories, so that we can use absolute import anywhere.

mlir/test/common_utils/common.py
```diff
@@ -43,20 +43,37 @@ def get_arch_features(arch: str):
...
+            raise Exception(f"Hip call hipGetDeviceProperties not successful: {err_code}")
+        agent = props.gcnArchName.decode('utf-8')
+        agents.add(agent)
+
+    return agents
+
+
+def is_xdlops_present() -> bool:
+    """This function checks whether a GPU with xdlops support is present"""
+    xdlop_supported_gpus = ['gfx908', 'gfx90a', 'gfx942', 'gfx950']
```
dhernandez0: change to .startsWith("gfx9")

mlir/test/common_utils/common.py
```diff
@@ -43,20 +43,37 @@ def get_arch_features(arch: str):
...
+    err_code, device_count = hip.hipGetDeviceCount()
+    if err_code != hip.hipSuccess:
```
dhernandez0: can we reuse get_agents() ? then iterate over the results

mlir/utils/performance/perfRunner.py
```diff
@@ -18,6 +18,8 @@
 from typing import Optional, Dict, Tuple
 import numpy as np
 import pandas as pd
+from hip import hip
+from hip import hiprtc
```
dhernandez0: no need to import hiprtc

mlir/utils/performance/perfRunner.py
```diff
@@ -1240,27 +1242,15 @@ def tuneMLIRKernels(configs, arch, numCU):
...
-        q = subprocess.run(["/opt/rocm/bin/rocm_agent_enumerator"],
-                              check=True, stdout=subprocess.PIPE)
-        agents = set(x.decode("utf-8") for x in q.stdout.split() if x != b"gfx000")
+    agents = set()
+    _, device_count = hip.hipGetDeviceCount()
+    for device in range(device_count):
+        props = hip.hipDeviceProp_t()
+        hip.hipGetDeviceProperties(props,device)
+        agent = props.gcnArchName.decode('utf-8')
+        agents.add(agent) if agent != "gfx000" else None
```
dhernandez0: gfx000 check not needed

mlir/test/e2e/generateE2ETest.py
dhernandez0: I understand we can't import mlir/test/common_utils/common.py in perfRunner.py because it's in utils/performance. But why can't we import common.py here? it's also in mlir/test

dorde-antic: We can import it, but in that case we would need to have a package, since those files are not all in same directory (others are in mlir/test/e2e and mlir/test/fusion/e2e). I thought that maybe using package wouldn't be appropriate solution for a file with 2 methods in it.

dhernandez0: it's fine then

dhernandez0 (approved):

umangyadav (approved):

mlir/utils/jenkins/Dockerfile
```diff
@@ -90,7 +90,9 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && apt upgrade -y \
     && python3 -m pip install --upgrade pip \
     && pip3 install pandas numpy scipy jinja2 tomli \
-    && pip3 install -r requirements.txt
+    && pip3 install -r llvm-requirements.txt \
+    && pip3 install -r ../../../requirements.txt \
+    && pip3 install -i https://test.pypi.org/simple hip-python
```
dhernandez0: I think it might work if you add --extra-index-url https://pypi.org/simple to the first line of requirements.txt ?

mlir/utils/jenkins/Dockerfile
```diff
@@ -90,7 +90,9 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && apt upgrade -y \
     && python3 -m pip install --upgrade pip \
     && pip3 install pandas numpy scipy jinja2 tomli \
-    && pip3 install -r requirements.txt
+    && pip3 install -r llvm-requirements.txt \
+    && pip3 install -r ../../../requirements.txt \
```
dhernandez0: let's use COPY instead

PR #1767, opened 2025-03-07
Introduce perfConfig V3 with param to select different schedule 

umangyadav:
This PR adds place holder for selecting different GEMM schedule as tuning param. 

Also adds param for output swizzle which may be used later. 

Keeping this as draft since i have to modify a few unit-tests, rest of the changes are good for review and early feedback.


mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
```diff
@@ -235,18 +238,23 @@ def Rock_GeneralGemmParamsAttr : Rock_Attr<"GeneralGemmParams", [RockTuningParam
     "int64_t":$mPerThread,
     "int64_t":$nPerThread,
     "int64_t":$kpack,
-    "int64_t":$splitKFactor
+    "int64_t":$splitKFactor,
+    "int64_t":$scheduleVersion,
+    "bool":$outputSwizzle
```
dhernandez0: I think for output swizzle we can have int64_t as well. Where:
0 - disabled
1 - enabled
2 - heuristic (current implementation)
By default it's 2.

If we enable it by default we might have some performance regressions.

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
```diff
@@ -286,6 +294,8 @@ def Rock_AttnPerfConfig : Rock_Attr<"AttnPerfConfig", [RockTuningParamAttrInterf
     }
 
     int64_t getSplitKFactor() { return 1; }
+    int64_t getScheduleVersion() { return 1;}
```
dhernandez0, line 297: what are the values for schedule version?
1 is the current one in develop? 0 is no schedule and 2 the new one?

mlir/include/mlir/Dialect/Rock/IR/RockTuningParamAttrInterface.td
```diff
@@ -69,7 +69,26 @@ def RockTuningParamAttrInterface : AttrInterface<"RockTuningParamAttrInterface">
...
+        }],
+        /*retType=*/"bool",
+        /*methodName=*/"getOutputSwizzle",
+        /*args=*/(ins),
+        /*methodBody=*/"",
+        /*defaultImplementation=*/""
+      >,
+    InterfaceMethod<
+        /*desc=*/[{
+          Returns version of the schedule used by the underlying GEMM algorithm
```
dhernandez0: we should provide info about what each number would do.

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
```diff
@@ -101,24 +101,30 @@ struct InitParamsNonAccel : InitParams, Serializable<InitParamsNonAccel> {
...
   uint32_t blockSize;
   int64_t splitKFactor;
+  int64_t gemmScheduleVersion;
+  bool outputSwizzle;
 
   constexpr InitParamsNonAccel(uint32_t bSize, int64_t mPerBlock,
                                int64_t nPerBlock, int64_t kPerBlock,
                                int64_t mPerThread, int64_t nPerThread,
-                               int64_t splitKFactor)
+                               int64_t splitKFactor, int64_t scheduleVersion,
```
dhernandez0, line 110: I guess RockPipeline runs for non-accel code as well, right? I haven't verified that's the case but I think so.

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
```diff
@@ -132,23 +138,30 @@ struct InitParamsNonAccel : InitParams, Serializable<InitParamsNonAccel> {
     f(self.gemmNPerThread);
     if (self.version != Version::V1)
       f(self.splitKFactor);
+    if (self.version == Version::V3) {
```
dhernandez0: nit: self.version >= Version::V3

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
```diff
@@ -132,23 +138,30 @@ struct InitParamsNonAccel : InitParams, Serializable<InitParamsNonAccel> {
...
   }
 };
 
 struct InitParamsAccel : InitParams, Serializable<InitParamsAccel> {
   constexpr InitParamsAccel(int64_t mPerBlock, int64_t nPerBlock,
                             int64_t kPerBlock, int64_t mPerWave,
                             int64_t nPerWaveOrMnPerXdl, int64_t kPack,
-                            int64_t splitKFactor, bool aThreadCopyMoreGemmK,
-                            bool bThreadCopyMoreGemmKPack)
+                            int64_t splitKFactor, int64_t scheduleVersion,
```
dhernandez0, line 152: nit: bool outputSwizzle after int64_t scheduleVersion? follow same order as they are defined

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
```diff
@@ -188,8 +203,14 @@ struct InitParamsAccel : InitParams, Serializable<InitParamsAccel> {
     if (self.version != Version::V1) {
       f(self.splitKFactor);
     }
+    if (self.version == Version::V3) {
```
dhernandez0: self.version >=

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
```diff
@@ -188,8 +203,14 @@ struct InitParamsAccel : InitParams, Serializable<InitParamsAccel> {
...
+      f(self.gemmScheduleVersion);
+    }
     f(self.gemmAThreadCopyMoreGemmK);
     f(self.gemmBThreadCopyMoreGemmKPack);
+    if (self.version == Version::V3) {
```
dhernandez0: self.version >=

mlir/include/mlir/Dialect/Rock/Tuning/Serializable.h
```diff
@@ -70,7 +70,8 @@ struct Serializable {
   }
 
   bool checkVersionFormat(const std::string &s) {
-    const int32_t maxNumTokens = version == Version::V1 ? 8 : 9;
+    const int32_t maxNumTokens =
+        version == Version::V1 ? 8 : (version == Version::V2 ? 9 : 11);
```
dhernandez0: nit: switch?

dhernandez0:
should we have EnableScheduleVersionForTuningAttr, EnableOutputSwizzleForTunningAttr? (same idea as EnableSplitKForTuningAttr) That would let you experiment (tune for one only), and it would keep exhaustive tuning as fast/slow as it is now.

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
umangyadav: I am calling current version `v1` . So it's 1 
new one `v2`. it would be 2. 

I am not sure `v0` makes sense. GEMM requries some sort of schedule. 

we can change those value later on to better align with CK's scheduleing versions though.

dhernandez0: sounds good. no schedule makes sense, it's just not defining rock.stage and not running rockpipeline. However, I'd expect worse performance in general.

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
umangyadav: It would run for any loop with `rock.pipeline` attribute.

Yes rockpipleline works for nonaccel path as well. I see `rock.pipeline` attribute being set for those. 
https://github.com/ROCm/rocMLIR/blob/a8ac1fb75c5d3472850f8c1cae249c325255ecb5/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp#L598

umangyadav:
> should we have EnableScheduleVersionForTuningAttr, EnableOutputSwizzleForTunningAttr? (same idea as EnableSplitKForTuningAttr) That would let you experiment (tune for one only), and it would keep exhaustive tuning as fast/slow as it is now.

We may need to discuss this with MIGraphX team about that. 

It would mean migraphx would need to attach those attributes conditionally whenever they want to tune for outputswizzle and/or scheduleVersions. They will have to introduce two more ENV/compiler flags for those. It's not the best for design.


Currently i am not enabling tuning on outputswizzle, tuning Param for output swizzle is just placeholder for future work.

dhernandez0:
> > should we have EnableScheduleVersionForTuningAttr, EnableOutputSwizzleForTunningAttr? (same idea as EnableSplitKForTuningAttr) That would let you experiment (tune for one only), and it would keep exhaustive tuning as fast/slow as it is now.
> 
> We may need to discuss this with MIGraphX team about that.
> 
> It would mean migraphx would need to attach those attributes conditionally whenever they want to tune for outputswizzle and/or scheduleVersions. They will have to introduce two more ENV/compiler flags for those. It's not the best for design.
> 
> Currently i am not enabling tuning on outputswizzle, tuning Param for output swizzle is just placeholder for future work.

Let's discuss first, I think adding more params would make exhaustive tuning at least 2x slower. My proposal is to keep new params like this until we figure out a working heuristic. So, users of exhaustive tuning (migraphx and ourselves internally) can keep using exhaustive tuning as it is now.

If we find a working heuristic (or not, and we give up and tune it), we can remove the TuningAttr.

dhernandez0 (approved):

mlir/include/mlir/Dialect/Rock/Tuning/QuickTuningPerfconfigs.inc
```diff
@@ -6,49 +6,49 @@
...
-    {128,32,32,16,2,4,1},
-    {64,64,64,8,2,2,1},
-    {256,32,128,8,2,2,1},
-    {64,32,64,8,4,2,1},
-    {64,32,128,16,4,2,1},
-    {128,64,128,4,4,2,1},
-    {64,64,32,16,4,2,1},
-    {64,64,32,8,2,2,1},
-    {64,32,32,4,2,4,1}
+    {128,128,128,16,4,4,1,1,2},
```
umangyadav, line 9: @djramic @dorde-antic 

I am changing quick tuning list to have two more parameters for `v3:` perfConfig. 

Do i need to update something for tuningRunner/perfRunner ?

mlir/include/mlir/Dialect/Rock/IR/RockTuningParamAttrInterface.td
```diff
@@ -69,7 +69,29 @@ def RockTuningParamAttrInterface : AttrInterface<"RockTuningParamAttrInterface">
         /*methodBody=*/"",
         /*defaultImplementation=*/""
       >,
-
+    InterfaceMethod<
+        /*desc=*/[{
+          Returns value  0 or 1 to decide if rocMLIR should enable output swizzle or not.
```
dhernandez0: nit: extra space

mlir/include/mlir/Dialect/Rock/IR/RockTuningParamAttrInterface.td
```diff
@@ -69,7 +69,29 @@ def RockTuningParamAttrInterface : AttrInterface<"RockTuningParamAttrInterface">
...
+        /*methodName=*/"getOutputSwizzle",
+        /*args=*/(ins),
+        /*methodBody=*/"",
+        /*defaultImplementation=*/""
+      >,
+    InterfaceMethod<
+        /*desc=*/[{
+          Returns version of the schedule used by the underlying GEMM algorithm. 
+          Return value of 1 means it should use 3 stage schedule with II=2. 
+          Return value of 2 means it should 2 stage schedule with II=1.
```
dhernandez0: 4 stage schedule?

umangyadav: Thanks. Updated.

mlir/include/mlir/Dialect/Rock/IR/RockTuningParamAttrInterface.td
umangyadav: You can use "suggest" feature when possible. it becomes much easier for me to make change.

mlir/test/CAPI/mixr_split_k.cpp
```diff
@@ -1,4 +1,4 @@
-//===- tosa_miir.cpp - Simple test of C and MIIR APIs ---------------------===//
+//===- tosa_miir.cpp - Simple test of C and MLIR APIs ---------------------===//
```
dhernandez0: change filename in the comment as well

mlir/test/fusion/pr-e2e/mixr-gemm-splitk/bf16/mixr-gemm-splitk-add.e2e.mlir
```diff
@@ -8,7 +8,7 @@ module {
   // CLONE-NEXT: Unranked Memref base
 
   func.func @dot_splitk_add_trunc(%arg0: !migraphx.shaped<1x5x4xbf16, 20x4x1>, %arg1: !migraphx.shaped<1x4x3xbf16, 12x3x1>, %arg2: !migraphx.shaped<1x5x3xbf16, 15x3x1>) -> !migraphx.shaped<1x5x3xbf16, 15x3x1> attributes{arch = "##TOKEN_ARCH##", enable_splitk_for_tuning, kernel = "mixr"} {
-    %0 = migraphx.dot %arg0, %arg1 {perf_config="v2:16,32,4,16,16,4,4,1,1"} : <1x5x4xbf16, 20x4x1>, <1x4x3xbf16, 12x3x1> -> <1x5x3xbf16, 15x3x1>
+    %0 = migraphx.dot %arg0, %arg1 {perf_config="v3:16,32,4,16,16,4,4,1,2,1,1"} : <1x5x4xbf16, 20x4x1>, <1x4x3xbf16, 12x3x1> -> <1x5x3xbf16, 15x3x1>
```
dhernandez0, line 11: can we leave v2 where possible? no need to change files for something that is experimental

mlir/test/CAPI/mixr_split_k.cpp
umangyadav: Done.

dhernandez0 (approved):

PR #1727, opened 2025-01-28
Implement performance checking script: i8 vs f16 non-fused conv kernels

dorde-antic:
Implement shell script that captures the performance difference between data types to validate expected kernel performance.
Resolves [ROCm/rocMLIR-internal#1674](https://github.com/ROCm/rocMLIR-internal/issues/1674)

PR #1768, opened 2025-03-07
Add Navi3x/Navi4x hardware to the weekly run

dorde-antic:
Do weekly tuning on Navi3x/Navi4x and use those perfDBs for nightly performance runs

Resolves [#1727](https://github.com/ROCm/rocMLIR-internal/issues/1727)


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -685,7 +688,7 @@ pipeline {
                 axes {
                     axis {
                         name 'ARCH'
-                        values 'gfx908', 'gfx90a', 'gfx1030'
+                        values 'gfx908', 'gfx90a', 'gfx1030', 'gfx1100', 'gfx1101', 'gfx1201'
```
umangyadav: gfx1100 and gfx1101 both are Navi3x based cards. I don't think we need to run on both of them.

umangyadav:
Can you also add Navi3x to parameterSweeps stage ? 

https://github.com/ROCm/rocMLIR/blob/a8ac1fb75c5d3472850f8c1cae249c325255ecb5/mlir/utils/jenkins/Jenkinsfile#L626 ?

mlir/utils/jenkins/Jenkinsfile
umangyadav: You would need to use same label later in this stage for Navi3x: 

https://github.com/ROCm/rocMLIR/blob/a8ac1fb75c5d3472850f8c1cae249c325255ecb5/mlir/utils/jenkins/Jenkinsfile#L815

dorde-antic: @umangyadav regarding gfx1100 and gfx1101, I talked with Stefan so we thought that it would be good to do it on both and see which is more stable.
But if you think it's better, we can run it only on one of them.

umangyadav: > it would be good to do it on both and see which is more stable. 

To me it seems we are not correctly making use of the Jenkins Tuning and Nightly Runs as of today. i.e. we are not really analyzing perf results on daily basis. 

Therefore it seems to me we are better off keeping just one of the Navi3x on the CI right now. otherwise we increase likelihood of failing weekly CI due to random errors and it will cause more problems.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -812,7 +814,7 @@ pipeline {
                 axes {
                     axis {
                         name 'CHIP'
-                        values 'gfx908', 'gfx90a', 'gfx1030', 'gfx1101'
+                        values 'gfx908', 'gfx90a', 'gfx1030', 'gfx1100', 'gfx1201'
```
umangyadav: can you trigger nightly and weekly tuning runs manually on this branch and see if it is tuning and generating perf reports  ?

PR #1719, opened 2025-01-16
[CI] Revert multi-step execution in Navi3x nightly E2E tests

stefankoncarevic:
Fixes: https://github.com/ROCm/rocMLIR-internal/issues/1148

PR #1769, opened 2025-03-09
Python script for testing metrics and plotting correlations

dorde-antic:
Implement python script for testing metrics and plotting correlation between metrics and parameters

Resolves [#1719](https://github.com/ROCm/rocMLIR-internal/issues/1719)

### DIFF
--- a/mlir/utils/performance/analysis/testing-metrics.py
+++ b/mlir/utils/performance/analysis/testing-metrics.py
@@ -23,7 +23,7 @@ from hip import hip
 
 numEUPerCU = 4 # may be changed in newer architectures
 
-def hip_check(call_result):
+def hipCheck(call_result):
     err = call_result[0]
     result = call_result[1:]
     if len(result) == 1:
@@ -38,13 +38,9 @@ def assignNumCu():
         return int(args.c)
     else:
         props = hip.hipDeviceProp_t()
-        hip_check(hip.hipGetDeviceProperties(props,0))
+        hipCheck(hip.hipGetDeviceProperties(props,0))
         print("Using info from GPU 0 in your system, the data should have be obtained from the same GPU.")
         return int(props.multiProcessorCount)
-
-
-numCUs = assignNumCu()
-minNumWaves = numCUs * numEUPerCU
 
 
 def analyzeGemmFile(file, n):
@@ -206,6 +202,9 @@ if __name__ == "__main__":
     parser.add_argument("--c", type=int, default=None) # numCUs (if data is not collected on the machine on which the script is executed)
 
     args = parser.parse_args()
+
+    numCUs = assignNumCu()
+    minNumWaves = numCUs * numEUPerCU
 
     rowList = []
 
--- a/mlir/utils/performance/analysis/testing-metrics.py
+++ b/mlir/utils/performance/analysis/testing-metrics.py
@@ -27,6 +27,7 @@ def hip_check(call_result):
         raise RuntimeError(str(err))
     return result
 
+
 props = hip.hipDeviceProp_t()
 hip_check(hip.hipGetDeviceProperties(props,0))
 numCUs = int(props.multiProcessorCount)
@@ -139,11 +140,11 @@ def calculateArithmeticIntensity(M, N, K):
     return (M*N*K)/(M*N + M*K + N*K) # opPerByte/bytesLoaded
 
 
-def calculateOccupancy(M, N, G, MPerBlock, NPerBlock, MNPerWave, minNumWaves):
+def calculateOccupancy(M, N, G, MPerBlock, NPerBlock, MNPerWave, minNumWaves, splitKFactor=1):
     MTiles = math.ceil(M/MPerBlock)
     NTiles = math.ceil(N/NPerBlock)
 
-    WorkGroups = G * MTiles * NTiles
+    WorkGroups = G * MTiles * NTiles * splitKFactor
     WavesPerBlock = MPerBlock * NPerBlock // MNPerWave
     Waves = WorkGroups * WavesPerBlock
 
--- a/mlir/utils/performance/analysis/testing-metrics.py
+++ b/mlir/utils/performance/analysis/testing-metrics.py
@@ -16,6 +16,8 @@ import argparse
 import math
 from hip import hip
 
+# TODO use AmdArchDb.py (when it's implemented)
+
 numEUPerCU = 4 # may be changed in newer architectures
 
 def hip_check(call_result):
@@ -32,6 +34,7 @@ props = hip.hipDeviceProp_t()
 hip_check(hip.hipGetDeviceProperties(props,0))
 numCUs = int(props.multiProcessorCount)
 minNumWaves = numCUs * numEUPerCU
+print("Using info from GPU 0 in your system, the data should have be obtained from the same GPU.")
 
 
 def analyzeGemmFile(file, n):
@@ -157,8 +160,9 @@ def calculateWorkImbalance(M, N, G, MPerBlock, NPerBlock, MNPerWave, minNumWaves
     WorkGroups = G * MTiles * NTiles * splitKFactor
     WavesPerBlock = MPerBlock * NPerBlock // MNPerWave
     Waves = WorkGroups * WavesPerBlock
+    WorkImbalanceIntermedResult = (Waves % minNumWaves) / minNumWaves
 
-    return ((1-((Waves % minNumWaves)/minNumWaves)) if ((Waves % minNumWaves)/minNumWaves) != 0 else 0)
+    return ((1-(WorkImbalanceIntermedResult)) if WorkImbalanceIntermedResult != 0 else 0)
 
 
 def determineFiletype(file):
--- a/mlir/utils/performance/analysis/testing-metrics.py
+++ b/mlir/utils/performance/analysis/testing-metrics.py
@@ -2,17 +2,19 @@
 # important metrics (Arithmetic Intensity, Occupancy, Work Imbalance) and
 # plots correlation between them with the selected parameters.
 #
-# Usage: python3 ./testing-metrics.py <debug file(s)> [--n <percent>] [--m <metrics>]
+# Usage: python3 ./testing-metrics.py <debug file(s)> [--n <percent>] [--m <metrics>] [--t <method for threshold>]
 # Arguments:
 #       <debug file(s)>               Input file(s) in .tsv.debug format
-#       --n <percent>                 Percent of the best perfconfigs to be considered (default=5)
+#       --n <percent>                 Percent of the best perfconfigs to be considered (default=5) - doesn't affect analysis when checking only the best perfConfigs
 #       --m <metrics>                 Metrics to be shown (ai, oc, wi, nmk)
+#       --t <method for threshold>    Method for calculating threshold (m - max, mn - maxN, qn - quantileN)
 
 import pandas as pd
 import seaborn as sns
 import matplotlib.pyplot as plt
 import argparse
 import math
+import hip
 
 numEUPerCU = 4 # may be changed in newer architectures
 numCUs = 304 # temporary hardcoded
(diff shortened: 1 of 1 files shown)


dorde-antic:
Function `def analyze_conv_file(file, n)` is left unimplemented, but is not deleted, in case we decide to calculate the metrics also for convs (to convert them to gemms and then calculate metrics).

mlir/utils/performance/analysis/testing-metrics.py
```diff
@@ -0,0 +1,173 @@
...
+#       --m <metrics>                 Metrics to be shown (ai, oc, wi, nmk)
+
+import pandas as pd
+import seaborn as sns
+import matplotlib.pyplot as plt
+import argparse
+import math
+
+numEUPerCU = 4 # may be changed in newer architectures
+numCUs = 304 # temporary hardcoded
```
umangyadav: If you are hardcoding these values then it is better if you check `get_gfx_arch` and assert that it is `Mi300`. 
Else i recommend using `import hip` and fetching these values from hipDeviceProperities dynamically.

mlir/utils/performance/analysis/testing-metrics.py
```diff
@@ -0,0 +1,173 @@
...
+minNumWaves = numCUs * numEUPerCU
+
+
+def analyze_gemm_file(file, n):
+    df = pd.read_csv(file, sep='\t')
+
+    gemm_keys = ['TransA', 'TransB', 'G', 'M', 'K', 'N']
+    perfConfig_params = ['MPerBlock', 'NPerBlock', 'KPerBlock', 'MPerWave', 'NPerWave', 'kPack', 'splitKFactor', 'forceUnroll', 'ThreadCopyMore']
+
+    df[perfConfig_params] = df["PerfConfig"].str.replace("v2:", "").str.split(",", expand=True)
```
umangyadav: I am adding perfConfig `v3: here : 
`https://github.com/ROCm/rocMLIR/pull/1767

can you make sure `.tsv` files are generated with `v2:` and assert for that or put comment somewhere ? 

Else you can check which perfConfig version it is first and based that set the columns differently for the `perfConfig_params`.

mlir/utils/performance/analysis/testing-metrics.py
```diff
@@ -0,0 +1,173 @@
...
+
+    top_list = []
+
+    for (key, group) in df.groupby(gemm_keys):
+        threshold = group['TFlops'].max() # change the grouping method according to the needs
+        top_list.append(group[group['TFlops'] == threshold])
+
+    list = pd.concat(top_list)
+
+    df[['Unnamed: 0', 'DataType', 'OutDataType', 'Chip', 'numCU', 'TransA',
```
umangyadav, line 75: why do you have `unnamed:0` field ?

mlir/utils/performance/analysis/testing-metrics.py
```diff
@@ -0,0 +1,173 @@
...
+    WorkGroups = G * MTiles * NTiles
+    WavesPerBlock = int(MPerBlock) * int(NPerBlock) / int(MNPerWave)
+    Waves = WorkGroups * WavesPerBlock
+
+    return Waves / minNumWaves
+
+
+def calculate_work_imbalance(M, N, G, MPerBlock, NPerBlock, MNPerWave, minNumWaves, splitKFactor=1):
+    MTiles = (int(M) + int(MPerBlock) - 1) / int(MPerBlock)
+    NTiles = (int(N) + int(NPerBlock) - 1) / int(NPerBlock)
```
umangyadav: I think you can use math.ceil for this one and similar places elsewhere

mlir/utils/performance/analysis/testing-metrics.py
dhernandez0: yes, let's check it's gfx942. This should be fixed with https://github.com/ROCm/rocMLIR-internal/issues/1745 when we get python AmdArchDb functionality.

mlir/utils/performance/analysis/testing-metrics.py
```diff
@@ -0,0 +1,173 @@
...
+    elif "TransA" in header:
+        return "gemm"
+    else:
+        raise Exception("Invalid file format or support for filetype not implemented yet: {file}")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Analyze .tsv.debug file")
+    parser.add_argument("files", nargs="+")
+    parser.add_argument("--n", type=float, default=5) # percent of configs close to winning
```
dhernandez0, line 198: default=0.05

mlir/utils/performance/analysis/testing-metrics.py
```diff
@@ -0,0 +1,173 @@
...
+    df[perfConfig_params] = df["PerfConfig"].str.replace("v2:", "").str.split(",", expand=True)
+
+    df["ArithmeticIntensity"] = df.apply(lambda row: calculate_arithmetic_intensity(row["M"], row["N"], row["K"]), axis=1)
+    df["MNPerWave"] = df.apply(lambda row: (int(row["MPerWave"]) * int(row["NPerWave"])), axis=1)
+    df["Occupancy"] = df.apply(lambda row: calculate_occupancy(row["M"], row["N"], row["G"], row["MPerBlock"], row["NPerBlock"], row["MNPerWave"], minNumWaves), axis=1)
+    df["WorkImbalance"] = df.apply(lambda row: calculate_work_imbalance(row["M"], row["N"], row["G"], row["MPerBlock"], row["NPerBlock"], row["MNPerWave"], minNumWaves, row["splitKFactor"]), axis=1)
+
+    top_list = []
+
+    for (key, group) in df.groupby(gemm_keys):
```
dhernandez0: not sure what these lines do. I'm not familiar with groupby. Could you add a comment here?

mlir/utils/performance/analysis/testing-metrics.py
```diff
@@ -0,0 +1,173 @@
...
+import argparse
+import math
+
+numEUPerCU = 4 # may be changed in newer architectures
+numCUs = 304 # temporary hardcoded
+
+minNumWaves = numCUs * numEUPerCU
+
+
+def analyze_gemm_file(file, n):
```
dhernandez0: where is "n" used?

mlir/utils/performance/analysis/testing-metrics.py
```diff
@@ -0,0 +1,173 @@
...
+
+def calculate_arithmetic_intensity(M, N, K):
+    return (M*N*K)/(M*N + M*K + N*K) # opPerByte/bytesLoaded
+
+
+def calculate_occupancy(M, N, G, MPerBlock, NPerBlock, MNPerWave, minNumWaves):
+    MTiles = (int(M) + int(MPerBlock) - 1) / int(MPerBlock)
+    NTiles = (int(N) + int(NPerBlock) - 1) / int(NPerBlock)
+
+    WorkGroups = G * MTiles * NTiles
```
dhernandez0: we should take into account splitKFactor here

mlir/utils/performance/analysis/testing-metrics.py
```diff
@@ -0,0 +1,173 @@
...
+    WavesPerBlock = int(MPerBlock) * int(NPerBlock) / int(MNPerWave)
```
dhernandez0: you can use "//" for integer division: WavesPerBlock = (MPerBlock* NPerBlock) // MNPerWave

mlir/utils/performance/analysis/testing-metrics.py
```diff
@@ -0,0 +1,173 @@
...
+
+    top_list = []    
+    return pd.concat(top_list)
+
+
+def calculate_arithmetic_intensity(M, N, K):
+    return (M*N*K)/(M*N + M*K + N*K) # opPerByte/bytesLoaded
+
+
+def calculate_occupancy(M, N, G, MPerBlock, NPerBlock, MNPerWave, minNumWaves):
```
dhernandez0: random idea: we could check if occupancy in terms of blocks has better correlation. I mean, it could be that having one or two waves per CU is enough? No need to explore this in this PR

mlir/utils/performance/analysis/testing-metrics.py
```diff
@@ -0,0 +1,173 @@
...
+                sns.scatterplot(x=list[param], y=list[nmk], alpha=0.7, ax=subplot)
+                subplot.set_xlabel(param)
+                subplot.set_ylabel(nmk)
+        plt.show()
+
+    return pd.concat(top_list)
+
+
+def analyze_conv_file(file, n):
+    # implementation goes here
```
dhernandez0, line 143: print an error message here

mlir/utils/performance/analysis/testing-metrics.py
```diff
@@ -0,0 +1,173 @@
...
+def calculate_work_imbalance(M, N, G, MPerBlock, NPerBlock, MNPerWave, minNumWaves, splitKFactor=1):
+    MTiles = (int(M) + int(MPerBlock) - 1) / int(MPerBlock)
+    NTiles = (int(N) + int(NPerBlock) - 1) / int(NPerBlock)
+    WorkGroups = G * MTiles * NTiles * int(splitKFactor)
+    WavesPerBlock = int(MPerBlock) * int(NPerBlock) / int(MNPerWave)
+    Waves = WorkGroups * WavesPerBlock
+
+    maxWavesPerCU = math.ceil(Waves / minNumWaves)
+
+    return (maxWavesPerCU * minNumWaves) / Waves
```
umangyadav: I don't fully understand how this work imbalance is computed. 

For example

let's consider workload with splitKFactor =1 only for ease.
 
In case A, it fills up 50% of the GPU. So 4 * (304/2) = 152 * 4 = 608 waves
 
In case B it fills up GPU 150% so it has  4 * (304) + 4 * 152 = 608 + 1216 = 1824 Waves. 
 
To me it seems both workloads are equally imbalanced. 
 
But based on following calculation, it will calculate 
Work imbalance for case A as 2
Work imbalance for case B as 1.33

dhernandez0: let's use (Waves % minNumWaves)/minNumWaves

mlir/utils/performance/analysis/testing-metrics.py
dorde-antic: It's used when filtering top N configs by
- group[ group['TFlops'] >= (group['TFlops'].max() * (1 - n / 100))]
- group['TFlops'].quantile(1 - n / 100.0)
This version of script was last version we used when we wanted to check plots considering only the best configs

We can also add all three options in code and then select by argument which kind of filtering we want.

mlir/utils/performance/analysis/testing-metrics.py
dorde-antic: I thought that user would put for example 5% so that n=5
which is later transformed to 0.05 using this `(1 - n / 100)`.
Anyway, it can be changed so that user passes 0.xx format for percentages as argument.

mlir/utils/performance/analysis/testing-metrics.py
dorde-antic: Groupby is used so that we can group the rows of the csv file in which we have the same `gemm_keys` (but different perfConfigs and TFlops) Gemm_keys param is previously defined.
Then for each subgroup generated by groupby, we calculate threshold using prefered function (max, best + n%, quantile...).

mlir/utils/performance/analysis/testing-metrics.py
dorde-antic: Since the first column are ordinal numbers and if we don't have "dummy" first row in line 44 I think that it would consider that DataType is: 0, 1, 2, 3 ... (which is in the first column)

mlir/utils/performance/analysis/testing-metrics.py
dorde-antic: In the files I was working with, all the perfConfigs were starting with v2: so that's why I used v2 (didn't know that there are also different cases).
It can be changed so that the script can recognize perfConfigs that doesn't start with v2. Should it be changed for this issue now?

mlir/utils/performance/analysis/testing-metrics.py
dhernandez0: I see, whatever you prefer then

mlir/utils/performance/analysis/testing-metrics.py
umangyadav: You can put an assert somewhere that it is `v2:`

mlir/utils/performance/analysis/testing-metrics.py
dhernandez0: sorry, I think it should be 1-((Waves % minNumWaves)/minNumWaves)

except if ((Waves % minNumWaves)/minNumWaves)=0, we should return 0.

mlir/utils/performance/analysis/testing-metrics.py
```diff
@@ -105,44 +115,41 @@ def analyze_gemm_file(file, n):
...
-    WavesPerBlock = int(MPerBlock) * int(NPerBlock) / int(MNPerWave)
+def calculateWorkImbalance(M, N, G, MPerBlock, NPerBlock, MNPerWave, minNumWaves, splitKFactor=1):
+    MTiles = math.ceil(M/MPerBlock)
+    NTiles = math.ceil(N/NPerBlock)
+    WorkGroups = G * MTiles * NTiles * splitKFactor
+    WavesPerBlock = MPerBlock * NPerBlock // MNPerWave
     Waves = WorkGroups * WavesPerBlock
 
-    maxWavesPerCU = math.ceil(Waves / minNumWaves)
+    return ((1-((Waves % minNumWaves)/minNumWaves)) if ((Waves % minNumWaves)/minNumWaves) != 0 else 0)
```
dhernandez0: nit: to make it easier to read, can you move Waves % minNumWaves)/minNumWaves to a variable as it's reused twice.

mlir/utils/performance/analysis/testing-metrics.py
```diff
@@ -0,0 +1,193 @@
...
+    if len(result) == 1:
+        result = result[0]
+    if isinstance(err, hip.hipError_t) and err != hip.hipError_t.hipSuccess:
+        raise RuntimeError(str(err))
+    return result
+
+
+props = hip.hipDeviceProp_t()
+hip_check(hip.hipGetDeviceProperties(props,0))
+numCUs = int(props.multiProcessorCount)
```
dhernandez0: we could use this script for data collected in another system assuming it's ok? Not sure what to do re that. Maybe just print a message saying that "we are using info from GPU 0 in your system, you should have got the data from the same GPU"?

mlir/utils/performance/analysis/testing-metrics.py
```diff
@@ -0,0 +1,193 @@
...
+#       --t <method for threshold>    Method for calculating threshold (m - max, mn - maxN, qn - quantileN)
+
+import pandas as pd
+import seaborn as sns
+import matplotlib.pyplot as plt
+import argparse
+import math
+from hip import hip
+
+numEUPerCU = 4 # may be changed in newer architectures
```
dhernandez0, line 24: add a TODO to use AmdArchDb.py (when it's implemented)

mlir/utils/performance/analysis/testing-metrics.py
dorde-antic: That's why it was initially was left hardcoded, so depending on the machine used hardcoded value was changed manually. But there was comment on previous review on changing that, so that's why I did it like this (it was one of recommended options if I understood well). 

I can add print message as you have mentioned for now and we can consider changes to be done later.
Anyway, maybe it makes more sense for the user to have option to select on which GPU was the data collected or to provide numCUs as argument.

dhernandez0 (approved):

PR #1771, opened 2025-03-10
Remove hasValidChip() from ConvGenerator

dorde-antic:
Remove hasValidChip() and checkChip. Modify isApplicable() function.

Resolves [#1753](https://github.com/ROCm/rocMLIR-internal/issues/1753)


umangyadav (approved):

mlir/lib/Dialect/Rock/Generator/ConvGenerator.cpp
```diff
@@ -126,9 +126,9 @@ LogicalResult ConvGenerator::isApplicable(bool checkChip) const {
     return failure();
   }
 
-  if (checkChip && failed(hasValidChip())) {
-    return failure();
-  }
+  // if (checkChip && failed(hasValidChip())) {
```
umangyadav: can we instead use `AmdArchDb.cpp` file to keep this validity check instead of removing it ?

dhernandez0: I think this is only used for miopen integration. Otherwise, checkChip is always false. Let's remove checkChip flag and other non used files

umangyadav: >  this is only used for miopen integration.

Okay. Looks like another maintenance overhead. We should check if MIOpen requires rocMLIR or not.

dhernandez0 (changes requested):
please remove checkChip and used files for miopen integration.

EDIT: Actually I think MLIRRockThin is needed for get_fat_library_deps_list.pl. So let's remove at least checkChip in isApplicable() and hasValidChip()

Ideally, we would get rid of MLIRRockThin and figure out dependencies using some other way? But let's do that in another ticket.

mlir/include/mlir/Dialect/Rock/Generator/ConvGenerator.h
dhernandez0, line 1: please remove "bool checkChip = true" in isApplicable

dhernandez0 (approved):

mlir/include/mlir/Dialect/Rock/Generator/ConvGenerator.h
```diff
@@ -138,10 +138,11 @@ class ConvGenerator {
     std::replace(targetLayout.begin(), targetLayout.end(), 'y', '0');
     std::replace(targetLayout.begin(), targetLayout.end(), 'x', '1');
     std::replace(targetLayout.begin(), targetLayout.end(), 'h', '0');
-    std::replace(targetLayout.begin(), targetLayout.end(), 'w', '1');
+    std::replace(targetLayout.begin(), targetLayout.end() P, 'w', '1');
```
dhernandez0: remove P

PR #1772, opened 2025-03-11
Add GEMM scheduleV2 

umangyadav:
This adds 4 stage gemm schedule with GlobalRead, LDSWrite, LDSRead, MFMA stages. 
Those 4 stages will get loop pipelined to run in parallel with initiation interval = 1. 
4 stage code is taken and modified from https://github.com/ROCm/rocMLIR/pull/1511 

By default this GemmSchedule is not enabled for tuning unless user has explictly asked for ExhaustiveTune. 

TODO: Add E2E and unit-tests


mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2551,6 +2551,117 @@ struct GridwiseGemmAccelRewritePattern
...
   using OpRewritePattern<GridwiseGemmAccelOp>::OpRewritePattern;
 
+  // Generate only the compute loop, i.e., we assume here that all
+  // the data that we need is already in LDS
+  void generateComputeLoop(
+      Location loc, PatternRewriter &b,
+      const std::unique_ptr<rock::accel::AccelEmitter> &accelEmitterPtr,
+      Value regsA, Value regsB, Value regsC, StringAttr arch,
+      GemmFeaturesAttr features,
+      const RockAccelTuningParamAttrInterface tuningParams) const {
```
mirza-halilcevic: Is there a reason for passing `tuningParams` by value and not by reference?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2551,6 +2551,117 @@ struct GridwiseGemmAccelRewritePattern
     : public OpRewritePattern<GridwiseGemmAccelOp> {
   using OpRewritePattern<GridwiseGemmAccelOp>::OpRewritePattern;
 
+  // Generate only the compute loop, i.e., we assume here that all
+  // the data that we need is already in LDS
```
dhernandez0: I think this comment should be "is already in registers"/"is already in private memory"?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2912,20 +3038,49 @@ struct GridwiseGemmAccelRewritePattern
...
+              loc, ldsViewForGemmA, ldsViewForGemmB,
+              b.getI32IntegerAttr(copyMPerThread),
+              b.getI32IntegerAttr(copyNPerThread),
+              (ldsLayoutConfigA.doRotateWithK ? b.getUnitAttr() : nullptr),
+              (ldsLayoutConfigB.doRotateWithK ? b.getUnitAttr() : nullptr),
+              arrayA, arrayB, regCAllocOp, op.getArchAttr(),
+              op.getFeaturesAttr(), op.getBlockSizeAttr(), op.getParamsAttr());
+          b.create<rock::YieldOp>(loc);
+        }
+      } else {
```
dhernandez0, line 3090: in the future we can refactor BlockwiseGemmAccelOp to take registers? maybe add a TODO here? or somewhere appropiate

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2825,11 +2938,23 @@ struct GridwiseGemmAccelRewritePattern
...
     // Logic to setup buffers for blockwise_gemm_accel.
-    auto arrayA =
-        gpuAlloc(b, loc, kBasePerThread, argTypeA, AddressSpace::Private);
-    auto arrayB =
-        gpuAlloc(b, loc, kBasePerThread, argTypeB, AddressSpace::Private);
+    int64_t arrayALen = kBasePerThread;
+    int64_t arrayBLen = kBasePerThread;
+    if (scheduleVersion == 2) {
+      arrayALen *= mRepeats;
+      arrayBLen *= nRepeats;
```
dhernandez0, line 2979: I understand then, that if scheduleVersion=1 we load from LDS in a loop only the values used in the iteration of mRepeats, right?
while scheduleVersion=2 pre-fetches everything to registers

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2912,20 +3038,49 @@ struct GridwiseGemmAccelRewritePattern
...
+              b.getI32IntegerAttr(copyMPerThread),
+              b.getI32IntegerAttr(copyNPerThread),
+              (ldsLayoutConfigA.doRotateWithK ? b.getUnitAttr() : nullptr),
+              (ldsLayoutConfigB.doRotateWithK ? b.getUnitAttr() : nullptr),
+              arrayA, arrayB, regCAllocOp, op.getArchAttr(),
+              op.getFeaturesAttr(), op.getBlockSizeAttr(), op.getParamsAttr());
+          b.create<rock::YieldOp>(loc);
+        }
+      } else {
+        // If we are running double-buffered pipeleines, it makes sense to also
```
dhernandez0: pipeleines -> pipelines

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -203,34 +203,48 @@ static void createGemmTuningRangeBF(TuningParamSet *newSpace,
   const std::vector<std::vector<uint32_t>> validRangeGeneralGemmParams = {
       {64, 128, 256}, {32, 64, 128}, {32, 64, 128}, {4, 8, 16}, {2, 4}, {2, 4}};
 
-  // M/block N/block K/block M/wave N/wave kPack aCopyMore/forceUnroll
+  // only enable tuning over gemm schedules when doing exhaustive tuning
+  auto getGemmSchedules = [&]() {
```
dhernandez0: nit: pass kind as argument to make it clear it depends on it. (feel free to ignore if you prefer it as it is)

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2825,11 +2938,23 @@ struct GridwiseGemmAccelRewritePattern
...
-        gpuAlloc(b, loc, kBasePerThread, argTypeA, AddressSpace::Private);
-    auto arrayB =
-        gpuAlloc(b, loc, kBasePerThread, argTypeB, AddressSpace::Private);
+    int64_t arrayALen = kBasePerThread;
+    int64_t arrayBLen = kBasePerThread;
+    if (scheduleVersion == 2) {
+      arrayALen *= mRepeats;
+      arrayBLen *= nRepeats;
+      initiationInterval = 1;
+    } else {
```
dhernandez0: else if == 1, then else with llvm_unreachable

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
umangyadav: Yes that's correct.

dhernandez0: thanks! I think we could improve scheduleVersion=1 if the pipeline pass supported nested loop pipelining? as we load LDS in a nested loop?

umangyadav: > if the pipeline pass supported nested loop pipelining? 

Yes It's possible.

dhernandez0:
please, make changes so that all CI tasks run full tuning instead of exhaustive tuning as well. Otherwise, we'll 2x our tuning time in CI.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
umangyadav: Done.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
umangyadav: Done

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
umangyadav: Done.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
umangyadav: Done

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
umangyadav: Changed that to const ref. THanks

mlir/test/e2e/GemmVariants.toml
dhernandez0, line 1: instead of hardcoding this we could use 

[[axis]]
name = "perfConfig"
values = ["", "v3:64,128,4,16,16,8,1,2,2,1,1"]
prefix = "--perf_config="

So, whenever we add another scheduling it's just adding an extra string.

Same for all other tests.

mlir/utils/performance/parameterSweeps.py
```diff
@@ -324,7 +325,9 @@ def to_conv_structure_type_test(params, options: Options) -> MLIROnlyConfig:
     # KPack (exponent)
     range(2, 5),
     # splitKFactor (exponent)
-    range(0, 1)
+    range(0, 1),
+    # GEMM Schedule Version
+    range(1, 3)
```
dhernandez0, line 331: parameterSweep is slow, 5hours I think, this would make it 10. Are we sure we want to do it?

mlir/utils/performance/tuningRunner.py
```diff
@@ -301,7 +301,7 @@ def main(args=None):
 
     parser.add_argument(
         "--tuning-space",
-        default="exhaustive",
+        default="full",
```
dhernandez0, line 299: seems reasonable to do full by default. But let's make sure everyone knows about this when we merge it. Let's send a message to the rocmlir team chat.

mlir/utils/performance/tuningRunner.py
dhernandez0, line 1: can you change the default in rocmlir-tuning-driver.cpp as well?

dhernandez0 (commented):
Some missing things:
- I still see some v2: in the codebase, for example in perfRunner.py, quickTuningGen.py and perfRegressionReport.py
- missing a test like gridwise_gemm_accel_lowering.mlir that checks the IR looks correct for the new soft pipeline strategy
- it'd be nice to have a table comparing the results vs develop when doing exhaustive tuning. But that can be done after it's merged. As we've seen some improvements already.

mlir/utils/performance/parameterSweeps.py
umangyadav: Yeah it is already 10 hours actually. It runs parameterSweeps twice. once for the convolution and other to test different perf_configs. 

I am not sure what would be best approach here. Ideally we would like to test scheduleV2 in some way

mlir/test/e2e/GemmVariants.toml
umangyadav: It's a good idea but 
I'll have to find out perfConfig that works for all the tests across all the hardware without running into any applicability or LDS errors.

dhernandez0: not sure I understand, if you have:

[[suite.test]]
config = "-g 3 -m 1024 -k 769 -n 1024 --perf_config v3:64,128,4,16,16,8,1,2,2,1,1"

it's already running with that perf_config on all different devices in CI? I think nothing changes, just instead of duplicating, we keep it in an axis?

mlir/utils/performance/parameterSweeps.py
dhernandez0: what if we set a number of iterations to run and we sample a number for each param? then, we should have pretty good coverage of the search space, although not fully iterating everything. If this is run nightly, we'll eventually see all possible combinations.

mlir/utils/performance/analysis/quickTuningGen.py
```diff
@@ -23,12 +23,12 @@ def parse_perfconfigs(self, perfconfig_str):
...
         Parses a perfconfigs in expected input format
         """
-        perfconfig_str = perfconfig_str.replace('v2:', '')
+        perfconfig_str = perfconfig_str.replace('v3:', '')
         config_values = perfconfig_str.split(',')
         total_values = len(config_values)
         converted_values = []
         for idx, value in enumerate(config_values):
-            if idx > total_values - 3 and total_values == 9:
+            if idx > total_values - 3 and total_values == 11:
```
dhernandez0, line 31: I think it's assuming the last 3 values are booleans, is this still true?

mlir/utils/performance/perfRegressionReport.py
```diff
@@ -25,7 +25,7 @@ def loadMlirData(filename: str):
     if 'numCU' not in df:
         df.insert(4, 'numCU', getNumCU(df['Chip'][0]))
     return df
-
+# TODO: Fix this for perfConfigV3
```
dhernandez0: do you want to fix this here or another ticket? if it's another ticket, let's open it and add it to the comment

mlir/utils/performance/analysis/quickTuningGen.py
umangyadav: It would check for last two values only. 

e.g. `idx > 8` --> `idx == 9, 10` 

last two values are booleans

mlir/utils/performance/perfRegressionReport.py
umangyadav: https://github.com/ROCm/rocMLIR-internal/issues/1809

mlir/utils/performance/parameterSweeps.py
dhernandez0: we can solve this in another PR, I'm not sure if we want to 2x the param sweep run-time. I think we can do some sampling to make it faster.

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
```diff
@@ -490,6 +490,15 @@ def Rock_PipelineAttr: Rock_Attr<"Pipeline">  {
   }];
 }
 
+def Rock_ScheduleVersionAttr : Rock_Attr<"ScheduleVersion"> {
```
dhernandez0, line 496: why do we need this if it's part of the perf config?

mlir/include/mlir/Dialect/Rock/Tuning/Serializable.h
```diff
@@ -142,7 +142,7 @@ struct Serializable {
   Version getVersion() { return version; }
 
 protected:
-  Version version{Version::V2};
+  Version version{Version::V3};
```
dhernandez0, line 145: how does this work with attention? where we still are in v1? sorry, I'm not familiar with this file

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2831,11 +2944,26 @@ struct GridwiseGemmAccelRewritePattern
...
+    int64_t initiationInterval;
+
     // Logic to setup buffers for blockwise_gemm_accel.
-    auto arrayA =
-        gpuAlloc(b, loc, kBasePerThread, argTypeA, AddressSpace::Private);
-    auto arrayB =
-        gpuAlloc(b, loc, kBasePerThread, argTypeB, AddressSpace::Private);
+    int64_t arrayALen = kBasePerThread;
+    int64_t arrayBLen = kBasePerThread;
+    if (scheduleVersion == 2) {
```
dhernandez0, line 2977: can we have constants defined somewhere like: v1, v2 or even more descriptive: v1OneLDSBuffer, v2DoubleBuffering

mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
```diff
@@ -118,20 +119,43 @@ void AffixTuningParameters::setUtilityKernelSizes(Value arg, T utilityOp) {
...
+  auto funcParent = op->getParentOfType<func::FuncOp>();
   std::string perfConfig;
+  if (funcParent->hasAttrOfType<rock::ScheduleVersionAttr>(
+          scheduleVersionAttrName) &&
+      op->hasAttrOfType<StringAttr>("perf_config")) {
+    op->emitError("kernel has both perf_config and schedule_version attribute "
+                  "set. Please modify schedule version directly inside "
+                  "perf_config and remove schedule_version\n");
+    signalPassFailure();
+    return;
```
dhernandez0, line 135: nit: you can do "return signalPassFailure();"

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
umangyadav: This  is mainly for testing purposes.  

Without having this attribute the problem i have is that,i have to find out perfConfig such that it is picking valid mfma/wmma instruction for all dtypes on all hardwares which is difficult and can also change in future. 

`rock-affix-params` will only accept either perfConfig or scheduleVersion not both. 

With this i do `rocmlir-gen --schedule_version 1/2` and it would pick default perfConfig that is valid for testing purposes.

mlir/include/mlir/Dialect/Rock/Tuning/Serializable.h
umangyadav: All the operations  continue accepting `v1` & `v2`. 

Serialization logic will make sure v1, v2 and v3 are correctly parsed.

mlir/include/mlir/Dialect/Rock/IR/RockAttrDefs.td
dhernandez0: I see, thanks

mlir/utils/performance/perfRegressionReport.py
umangyadav: No work is necessary for this one. Closed this ticket and removed "TODO" Item.

mlir/utils/performance/perfRegressionReport.py
```diff
@@ -25,7 +25,7 @@ def loadMlirData(filename: str):
     if 'numCU' not in df:
         df.insert(4, 'numCU', getNumCU(df['Chip'][0]))
     return df
-# TODO: Fix this for perfConfigV3
```
dhernandez0: why isn't this needed?

mlir/test/Dialect/Rock/lowering_to_threadwise_accel.mlir
```diff
@@ -0,0 +1,203 @@
+// RUN: rocmlir-opt -split-input-file -rock-gridwise-gemm-to-blockwise -rock-blockwise-gemm-to-threadwise %s | FileCheck %s
```
dhernandez0, line 1: thanks for adding this test. Would it be possible to add a test that also runs --rock-pipeline? That way we can be sure that we are using double buffering for v2 (and will be able to detect regressions in the future).

dhernandez0 (approved):

mlir/utils/performance/perfRegressionReport.py
umangyadav: Explanation here : https://github.com/ROCm/rocMLIR-internal/issues/1809#issuecomment-2806825065

PR #1773, opened 2025-03-11
Fix generateMlirDriverCommandLine for attention in perfRunner

dhernandez0:
Small fix for perfRunner

### DIFF
--- a/mlir/utils/performance/perfRunner.py
+++ b/mlir/utils/performance/perfRunner.py
@@ -37,6 +37,7 @@ DATA_TYPES_ATTENTION = ['f32', 'f16', 'bf16']
 OUTPUT_DATA_TYPES_MAP = {'f32': 'f32', 'f16': 'f16', 'bf16': 'bf16', 'i8': 'i32', 'fp8':'f32',
                          'fp8_fp8': 'f32', 'fp8_bf8': 'f32', 'bf8_fp8': 'f32',
                          'bf8_bf8': 'f32'}
+MLIR_N_REPEATS = 5
 
 # Compiled regexp object used for extracting elapsed time from MIOpenDriver's output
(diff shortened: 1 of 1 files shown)


mlir/utils/performance/perfRunner.py
```diff
@@ -37,6 +37,7 @@
 OUTPUT_DATA_TYPES_MAP = {'f32': 'f32', 'f16': 'f16', 'bf16': 'bf16', 'i8': 'i32', 'fp8':'f32',
                          'fp8_fp8': 'f32', 'fp8_bf8': 'f32', 'bf8_fp8': 'f32',
                          'bf8_bf8': 'f32'}
+MLIR_N_REPEATS = 5
```
dhernandez0, line 40: does somebody know if there's a reason we shouldn't increase this to 100? This is not run for tuning, only once per conv/gemm/attention when measuring performance.

dorde-antic (approved):

mlir/utils/performance/perfRunner.py
umangyadav: I don't understand how `kernel_repeats` are being used. from what i understand it is used for benchmarking. does it mean it is used when doing tuning ?

umangyadav (approved):
Changes look good

mlir/utils/performance/perfRunner.py
dhernandez0: kernel_repeats is used in rocmlir-gen to generate a loop that calls the kernel multiple times (--kernel-repeats N times). This is used for perfRunner.py to measure performance with rocprof. For tuning we use tuningRunner.py that uses rocmlir-tuning-driver, so there's no --kernel-repeats. 

The only effect I can see for increasing this is more run-time for CI? but more reliable measurements as well. Anyway, let's decide in a future PR and merge this for now.

PR #1774, opened 2025-03-12
Add Gemm+Elementwise+Gemm support

dhernandez0:
This PR introduces GEMM+GEMM fusion. There are some things left for future PRs:
- extend allowed types
- Jenkins GEMM+GEMM (tuna-script.sh, perfRegression report, etc)
- CK benchmark script
- conv+gemm

I've created an epic for the pending tasks: https://github.com/ROCm/rocMLIR-internal/issues/1791

I've done an initial perf comparison of these GEMM+GEMM problems:
```csv
-transO false -transC false -transB true -transA false -t f32 -g 32 -m 8192 -n 8192 -k 128 -gemmO 128
-transO false -transC false -transB true -transA false -t f32 -g 16 -m 16384 -n 16384 -k 128 -gemmO 128
```

Which correspond to these 4 GEMMs:
```csv
#-transO false -transC false -transB true -transA false -t f32 -g 32 -m 8192 -n 8192 -k 128 -gemmO 128
-transB true -transA false -g 32 -m 8192 -n 8192 -k 128 -t f32 -out_datatype f32
-transB false -transA false -g 32 -m 8192 -n 128 -k 8192 -t f32 -out_datatype f32
#-transO false -transC false -transB true -transA false -t f32 -g 16 -m 16384 -n 16384 -k 128 -gemmO 128
-transB true -transA false -g 16 -m 16384 -n 16384 -k 128 -t f32 -out_datatype f32
-transB false -transA false -g 16 -m 16384 -n 128 -k 16384 -t f32 -out_datatype f32
```

Using tuningRunner and perfRunner I get the following run-times:
| run-time (ms) | PR      | 2 GEMMs |
| ------------- | ------- | ------- |
| Problem 1     | 8.9470  | 10.9019 |
| Problem 2     | 17.5556 | 22.1153 |


mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -255,6 +255,54 @@ def Rock_AttentionOp :
...
+    Rock_GemmFeaturesAttr:$features,
+    OptionalAttr<I32Attr>:$numCU,
+    OptionalAttr<RockTuningParamAttrInterface>:$params0,
+    OptionalAttr<RockTuningParamAttrInterface>:$params1,
+    I32Attr:$firstGemmIdx
+  )>,
+  Results<(outs Optional<TensorOf<[F32, F16, BF16]>>:$result)> {
+  let summary = "GEMM-elementwise-GEMM operation";
+  let description = [{
+    Performs the operation out = (a * b) * c.
```
umangyadav: Is this missing elementwise in between ?

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -255,6 +255,54 @@ def Rock_AttentionOp :
...
+
+    `features` specifies what hardware features can be used in the generated code.
+  }];
+  let hasVerifier = 1;
+  let regions = (region AnyRegion:$preSecondGemmBody);
+  let assemblyFormat = [{
+    `{` `\n`
+        ` ` `ab` `=` (`tr` $aTransposed^)? $a `*` (`tr` $bTransposed^)? $b `:` type($a) `,` type($b) `\n`
+        (`ab` `=` `elementwise` (`otherIns` `(` $elemwiseInputs^ `:` type($elemwiseInputs) `)`)? $preSecondGemmBody^ `\n`)?
+        (`tr` $oTransposed^)? $out `=` `softmax` `(` `qk` `)` `*` (`tr` $cTransposed^)? $c `:` type($c) `->` type($out) `\n`
```
umangyadav: Why does it have a softmax ?

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -434,12 +482,12 @@ def Rock_GridwiseGemmAccelOp :
 // gridwise_attention_accel
 def Rock_GridwiseAttentionAccelOp :
     Rock_Op<"gridwise_attention_accel", [DeclareOpInterfaceMethods<MemoryEffectsOpInterface>, RockFusionRoot, AttrSizedOperandSegments]>,
-    Arguments<(ins MemRefRankOf<[F32, F16, BF16, I8], [3]>:$queries,
-                   MemRefRankOf<[F32, F16, BF16, I8], [3]>:$keys,
-                   MemRefRankOf<[F32, F16, BF16,], [3]>:$values,
+    Arguments<(ins MemRefRankOf<GemmInputTypes, [3]>:$queries,
+                   MemRefRankOf<GemmInputTypes, [3]>:$keys,
+                   MemRefRankOf<GemmInputTypes, [3]>:$values,
```
umangyadav: GemminputTypes also contains Fp8 types. 

Can we make this change to extend attention ops on FP8 as a separate changelist with tests ?

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -484,12 +486,8 @@ ConvOpType mlir::rock::convOpTypeFromKernelType(KernelType kernelType) {
     return ConvOpType::BwdData;
   case KernelType::ConvBwdWeight:
     return ConvOpType::BwdWeight;
-  case KernelType::Gemm:
-    llvm_unreachable(
-        "Gemm ops shouldn't be in convolution-specific lowering passes");
-  case KernelType::Attention:
-    llvm_unreachable(
-        "Attention ops shouldn't be in convolution-specific lowering passes");
+  default:
```
umangyadav: IMO, Previous way of checking for GEMM/Attention can produce better error messages.

mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
```diff
@@ -311,11 +315,32 @@ void AffixTuningParameters::affixTuningParametersImpl(AttentionOp op) {
...
+    GemmElementwiseGemmOp op) {
+  Type elemTypeQ = cast<MemRefType>(op.getA().getType()).getElementType();
+  Type elemTypeK = cast<MemRefType>(op.getB().getType()).getElementType();
+  Type elemTypeV = cast<MemRefType>(op.getC().getType()).getElementType();
+  func::FuncOp funcOp = getOperation();
+  // Let's use attention perf config for now
+  if (failed(affixTuningParametersImplAttentionPerf(op, funcOp, elemTypeQ,
+                                                    elemTypeK, elemTypeV)))
+    return signalPassFailure();
+}
```
umangyadav:
```suggestion
}

```

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -870,37 +1002,80 @@ struct CollapseExpandRewritePattern
...
-        tosa::GreaterOp,
-        tosa::GreaterEqualOp
-       >(op);
-  // clang-format on
-}
+struct GemmElementwiseGemmRewritePattern
+    : public OpRewritePattern<tosa::MatMulOp> {
+  using OpRewritePattern<tosa::MatMulOp>::OpRewritePattern;
+
+  LogicalResult match(tosa::MatMulOp op) const override {
```
umangyadav: FYI: 
https://github.com/llvm/llvm-project/pull/130031 

`match()` and `rewrite()` are being deprecated.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -870,37 +1002,80 @@ struct CollapseExpandRewritePattern
...
+    std::tie(std::ignore, maybeFirstMatMul) =
+        getElementwiseRegion(op.getA(), b, nullptr, vec);
+
+    if (succeeded(maybeFirstMatMul))
+      LLVM_DEBUG(llvm::dbgs()
+                 << "first matmul = " << maybeFirstMatMul.value() << "\n");
+    else
+      LLVM_DEBUG(llvm::dbgs() << "first matmul not found\n");
+
+    return maybeFirstMatMul;
```
umangyadav: I don't understand returning `maybeFirstMatmul`. would it automatically get casted into `logicalResult` for the case where it was found ?

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -351,6 +351,138 @@ static FailureOr<int64_t> traceToRes(Value expectedTensor, func::FuncOp func) {
...
+static Value addBlockArgument(OpBuilder &b, Value val, Block *block,
+                              Location loc) {
+  RankedTensorType valType = cast<RankedTensorType>(val.getType());
+  val = block->addArgument(
+      MemRefType::get(valType.getShape(), valType.getElementType()), loc);
+  val = rock::getAsTensor(b, loc, val);
+  return val;
+}
+
+// This function traverse an upward tree where the root is the softmax
```
umangyadav: It may not contain softmax for GEMM-EW-GEMM

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -870,37 +1002,80 @@ struct CollapseExpandRewritePattern
...
+struct GemmElementwiseGemmRewritePattern
+    : public OpRewritePattern<tosa::MatMulOp> {
+  using OpRewritePattern<tosa::MatMulOp>::OpRewritePattern;
+
+  LogicalResult match(tosa::MatMulOp op) const override {
+    OpBuilder b{op};
+    SmallVector<Value> vec;
+    FailureOr<tosa::MatMulOp> maybeFirstMatMul;
+    std::tie(std::ignore, maybeFirstMatMul) =
+        getElementwiseRegion(op.getA(), b, nullptr, vec);
```
umangyadav: Would this work if elementwise is attached on `op.getB()` instead ?

umangyadav: this is passing `nullptr` for the block. how is it working because at somepoint it would try to add blockArgument

mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
```diff
@@ -311,11 +315,32 @@ void AffixTuningParameters::affixTuningParametersImpl(AttentionOp op) {
...
+  if (failed(affixTuningParametersImplAttentionPerf(op, funcOp, elemTypeQ,
+                                                    elemTypeK, elemTypeV)))
+    return signalPassFailure();
 }
+
+void AffixTuningParameters::affixTuningParametersImpl(
+    GemmElementwiseGemmOp op) {
+  Type elemTypeQ = cast<MemRefType>(op.getA().getType()).getElementType();
+  Type elemTypeK = cast<MemRefType>(op.getB().getType()).getElementType();
+  Type elemTypeV = cast<MemRefType>(op.getC().getType()).getElementType();
```
umangyadav: Seems like if you name A == queries 
B = keys
C = values 
then you can possibly unify some of the code

mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
```diff
@@ -53,6 +55,7 @@ void AffixTuningParameters::runOnOperation() {
   func.walk(
       [&](RockGemmWrapperInterface op) { affixTuningParametersImpl(op); });
   func.walk([&](AttentionOp op) { affixTuningParametersImpl(op); });
+  func.walk([&](GemmElementwiseGemmOp op) { affixTuningParametersImpl(op); });
```
umangyadav: can you unit-tests for rock-affix-params for GEMM-EW-GEMM ?

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
```diff
@@ -93,6 +106,135 @@ struct AttentionRewritePattern : public OpConversionPattern<AttentionOp> {
...
+
+  SmallVector<int64_t, 3> cShape =
+      llvm::to_vector<3>(cast<MemRefType>(c.getType()).getShape());
+
+  GemmSize gemm0Size(/*g=*/aShape[0], /*m=*/bShape[2],
+                     /*k=*/aShape[1],
+                     /*n=*/aShape[2]);
+  GemmSize gemm1Size(/*g=*/aShape[0], /*m=*/cShape[2],
+                     /*k=*/cShape[1],
+                     /*n=*/aShape[2]);
```
umangyadav: Is it it `(a * b) * c` 
therefore `m` for the gemm1 should be coming from matrix A ?

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
```diff
@@ -93,6 +106,135 @@ struct AttentionRewritePattern : public OpConversionPattern<AttentionOp> {
...
+
+  int64_t gridSize =
+      ((gemm0Size.n) / accelParams0.getNPerBlock()) * gemm0Size.g;
```
umangyadav, line 131: I don't understand why gridsize only depends on `N`

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1830,31 +1832,28 @@ struct GridwiseAttentionAccelRewritePattern
     // Currently, there is a working assumption that this kernel is meant
     // support fp32/fp16 This should be guranteed by op verifiers.
```
umangyadav: This comment requires changes if we support BF16, Fp8 and Int8 now.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -498,4 +567,11 @@ void RockSortDimensionsMemoryLayoutPass::runOnOperation() {
   if (failed(applyOpPatternsGreedily(getOperations<rock::AttentionOp>(func),
                                      std::move(patternsAttention), config)))
     return signalPassFailure();
+
+  RewritePatternSet patternsGemmElementwiseGemm(&ctx);
+  patternsGemmElementwiseGemm.add<GemmElementwiseGemmRewritePattern>(&ctx);
+  if (failed(applyOpPatternsGreedily(
+          getOperations<rock::GemmElementwiseGemmOp>(func),
+          std::move(patternsGemmElementwiseGemm), config)))
```
umangyadav: I wonder instead of applying each patterns separately if you can provide ranking as did in TosaToRock.

or if there is some other way

mlir/utils/performance/perfRunner.py
```diff
@@ -737,6 +779,163 @@ def __init__(self, dtype: str, outDataType: str, g: int, m: int, k: int, n: int,
...
+        self.arch = arch
+        self.chip = GFX_CHIP_RE.search(arch).group(0)
+        self.numCU = numCU
+        self.perfConfig = perf_config
+
+    def computeTFlops(self, ns):
+        # NaN will propagate as expected
+        # Repeats are handled by the fact that we're using avarageNs
+        first_matmul_flops = 2.0 * self.g * self.m * self.k * self.n
+        # max, sub, exp, sum, div
```
umangyadav:
```suggestion
   
```

mlir/utils/performance/reportUtils.py
```diff
@@ -24,6 +24,7 @@
                         'PaddingH', 'PaddingW', 'PerfConfig']
 GEMM_TEST_PARAMETERS = ['DataType', 'OutDataType', 'Chip', 'numCU', 'TransA', 'TransB', 'G', 'M', 'K', 'N', 'PerfConfig']
 ATTN_TEST_PARAMETERS = ['DataType', 'Chip', 'numCU', 'TransQ', 'TransK', 'TransV', 'TransO', 'WithAttnScale', 'G', 'SeqLenQ', 'SeqLenK', 'HeadDimQK', 'HeadDimV', 'PerfConfig']
+GEMM_GEMM_TEST_PARAMETERS = ['ADataType', 'BDataType', 'CDataType', 'OutDataType', 'Chip', 'numCU', 'TransA', 'TransB', 'TransC', 'TransO', 'G', 'M', 'K', 'N', 'GEMM2N', 'PerfConfig']
```
umangyadav: Would it be better  to say `N1` and `N2` instead of `N` and `GEMM2N`

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1830,31 +1832,28 @@ struct GridwiseAttentionAccelRewritePattern
...
     Type gemmOutElemType = elemTypeQxK;
-    Type softmaxInElemType = elemTypeQxK;
+    Type fusionOutElemType = elemTypeQxK;
```
umangyadav, line 1846: Would it be same as `elemTypeQxK` if let's say there was a convert operation as elementwise operation after the first GEMM ?

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
dhernandez0: Just following the same style of AttentionOp or GemmOp, where elementwise fusions between first and second gemm (or input/output fusions) are not mentioned.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
dhernandez0: copy-paste mistake, thanks

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
dhernandez0: AttentionOp still have the same datatype limitations. So, even though GridwiseAttentionAccelOp accepts more types, that doesn't affection attention. We can extend attention to more types in another ticket. Re GridwiseAttentionAccelOp, I'll add tests.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
dhernandez0: thanks, good catch. We'll have to update this file because AttentionRewritePattern uses match()+rewrite().

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
dhernandez0: because maybeFirstMatmul is of type "FailureOr<<tosa::MatMulOp>>". FailureOr has a conversion operator to LogicalResult.

mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
dhernandez0: yes, there's already a function affixTuningParametersImplAttentionPerf that I'm using for both Attention and GEMM+GEMM

mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
dhernandez0: sure, I still need to add tests.

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
dhernandez0: this is because we are implementing flash attention here. It depends on n and g (g=batch*num_heads). This is why split-kv might be beneficial in some cases, we don't launch many workgroups if batch=1 and n=16 (really 1 padded to 16) when doing the decoding phase.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
dhernandez0: The issue here is that we are replacing (for example) GemmOp by another GemmOp. So, we need to use GreedyRewriteStrictness::ExistingOps, applyOpPatternsGreedily and pass exactly the operations that need to run the pass. Otherwise, the matcher will match the new GemmOp again and we'll go into an infinite loop.

It might work by passing all operations and all rewrite patterns, but I think that can be done in another clean up PR, I'm already doing too many changes here.

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
dhernandez0: > Would this work if elementwise is attached on op.getB() instead ?

This is A of the second GEMM. So, that comes from the output of the first GEMM.

> this is passing nullptr for the block. how is it working because at somepoint it would try to add blockArgument

We use block only if doRewrite is true. I think when block is nullptr we set doRewrite to false.

mlir/utils/performance/reportUtils.py
dhernandez0: it looks like CK calls this "O" (see the comment here: https://github.com/ROCm/composable_kernel/blob/52b1cd7780f412be64bb2b08aa10b91a6f2bf26a/example/31_batched_gemm_gemm/batched_gemm_gemm_xdl_fp32.cpp#L5). We can switch to O if you agree?

dhernandez0: https://github.com/ROCm/composable_kernel/blob/52b1cd7780f412be64bb2b08aa10b91a6f2bf26a/example/31_batched_gemm_gemm/run_batched_gemm_gemm_example.inc#L16

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
dhernandez0: It is like this currently in develop (for attention in AttentionRewritePattern::computeGridSize). Looks like somebody copy-pasted gemm0Size to gemm1Size and left it as is. I'll remove gemm1Size as it's not used anyway.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
dhernandez0: this is one of the things left to do here. I think currently, gridwisetoblockwise (for attention) adds conversions linalg generic ops automatically. I think that's very weird, we might not want that for gemm+gemm. Maybe we can simplify it here and restrict gemm+gemm to all types being equal for now? I feel this PR is too big right now and I'm not finished.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -255,6 +250,50 @@ def Rock_AttentionOp :
   }];
 }
 
+def Rock_GemmElementwiseGemmOp
+    : Rock_Op<"gemm_elementwise_gemm", [DeclareOpInterfaceMethods<
+                                            MemoryEffectsOpInterface>,
+                                        RockFusionRoot]>,
+      AllElementTypesMatch<["a", "b", "c"]>,
```
umangyadav, line 266: what if `elementwise` op converts the data type in between ? In that case 

`a * b` may generate output of different dtype than expected by `c`

we should check for allowed ElementWise Ops

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -255,6 +250,50 @@ def Rock_AttentionOp :
...
+      Arguments<(ins TensorOrMemRefOf<[F32]>:$a, TensorOrMemRefOf<[F32]>:$b,
+          TensorOrMemRefOf<[F32]>:$c,
+          Variadic<AnyTensorOrMemRef>:$elemwiseInputs,
```
umangyadav, line 269: I think dtypes that are allowed for the elementwise inputs will be tied to dtypes that GEMM expects later on. 

Therefore it can't be `AnyTensorOrMemRef`

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -255,6 +250,50 @@ def Rock_AttentionOp :
...
+    lowered into the `gridwise_attention_accel` stage of the code generation pipeline.
+
+    `features` specifies what hardware features can be used in the generated code.
+  }];
+  let hasVerifier = 1;
+  let regions = (region AnyRegion:$preSecondGemmBody);
+  let assemblyFormat = [{
+    `{` `\n`
+        ` ` `ab` `=` (`tr` $aTransposed^)? $a `*` (`tr` $bTransposed^)? $b `:` type($a) `,` type($b) `\n`
+        (`ab` `=` `elementwise` (`otherIns` `(` $elemwiseInputs^ `:` type($elemwiseInputs) `)`)? $preSecondGemmBody^ `\n`)?
```
umangyadav, line 306: I don't fully follow naming here.  both first gemm and elementwise op, are they named same  as `ab` ?

mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
```diff
@@ -240,16 +245,17 @@ deriveGemm1TuningParams(OpBuilder &builder, AttentionOp op,
...
   OpBuilder builder(op.getContext());
-  Type elemTypeQ = cast<MemRefType>(op.getQueries().getType()).getElementType();
-  Type elemTypeK = cast<MemRefType>(op.getKeys().getType()).getElementType();
-  Type elemTypeV = cast<MemRefType>(op.getValues().getType()).getElementType();
   bool isAccel = rock::isAccel(op.getFeatures());
   if (!isAccel) {
-    op.emitError("Currently, attention op is only supported on GPUs "
-                 "with matrix accelerator extentions");
-    return signalPassFailure();
+    return op.emitError("Currently, attention/GEMM+elementwise+GEMM op is only "
```
umangyadav:
```suggestion
    return op.emitError("Currently, " + op->name() "  + "....
```
You can use op->name() for better error message.

mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
umangyadav: I was wondering perhaps we can classify both Attention & GemmEwiseGemm into 
`GemmPlusGemmWrapperInterface` and use that instead whereever we want to apply same treatment for both Attention and GemmEwiseGemm. 

You can then use `dyn_cast` on GemmPlusGemmWrapperInterface to check if it is attention or GemmEwGemm and based on that you can enable/disable softmax and other stuffs.

mlir/lib/Dialect/Rock/Transforms/GemmToGridwise.cpp
```diff
@@ -93,6 +106,132 @@ struct AttentionRewritePattern : public OpConversionPattern<AttentionOp> {
...
+  // In the transposed layout, from a tuning params point of view
+  // the output dimensions are swapped. Though we will only be
+  // swapping them inside gridwise lowering to keep the surrounding
+  // fusions legit. So the extra pad needs to be swapped and applied.
+  out = padMatrix(out, rw, loc, "gemm1N", gemm1ExtraPad.n, "gemm1M",
+                  gemm1ExtraPad.m);
+
+  if (failed(computeGridSizeAttentionGemmElmtGemm(rw, op, a, b, c))) {
+    return op.emitError(
+        "failed to compute the grid size of `GemmElementwiseGemmOp`");
```
umangyadav:
```suggestion
        "failed to compute the grid size of `GemmElementwiseGemmOp`");
```

it could also be attention op

mlir/test/rocmlir-gen/attention-kernel.mlir
```diff
@@ -1,5 +1,4 @@
 // RUN: rocmlir-gen --arch gfx90a:sramecc+:xnack- --operation attention -seq_len_q 1024 -seq_len_k 1024 -head_dim_qk 32 -head_dim_v 32 --with-attn-scale -t f32 -pv --apply-bufferization-pipeline=false | rocmlir-opt | FileCheck %s --enable-var-scope --check-prefixes=CHECK_SCALE
-// RUN: rocmlir-gen --arch gfx90a:sramecc+:xnack- --operation attention -seq_len_q 1024 -seq_len_k 1024 -head_dim_qk 32 -head_dim_v 32 -t f32 -pv --apply-bufferization-pipeline=false | rocmlir-opt | FileCheck %s --enable-var-scope --check-prefixes=CHECK_NO_SCALE
```
umangyadav, line 2: why this was removed ?

mlir/lib/Dialect/Rock/Transforms/AffixTuningParameters.cpp
dhernandez0: yes, makes sense. Because we'll then have attention, gemm+gemm, conv+gemm, conv+conv, gemm+conv. And we treat them equally in some cases.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
dhernandez0: this PR is only fp32 to simplify it

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
dhernandez0: you can have a tensor that is used to indirectly change a*b (for example causal attention for attention kernels). So, the tensors don't have to be the same type.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
dhernandez0: yes, just following same naming as attention:
```mlir
    rock.attention{
     qk = tr %0 * tr %1 : memref<1x8x8xf16>, memref<1x8x8xf16>
     qk = elementwise {
    ^bb0(%arg4: memref<1x8x8xf16>, %arg5: memref<1x8x8xf16>):
      memref.copy %arg4, %arg5 : memref<1x8x8xf16> to memref<1x8x8xf16>
      rock.yield
    }
     %3 = softmax(qk) * tr %2 : memref<1x64x8xf16> -> memref<1x8x64xf16>
    }
```
happy to change it if you have a better name in mind, but let's keep it consistent with attention.

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
dhernandez0: see the comment below, we don't know if the element-wise tensors are used for indirect things (for example, a mask). So, forcing the type with AllElementTypesMatch to be the same would limit valid fusions. 

But you are right, I'm not sure how this is verified for attention either, I think we probably fail when lowering instead of at rock.attention verifiers.

umangyadav:
@dhernandez0  this PR has merge conflicts.

dhernandez0:
> @dhernandez0 this PR has merge conflicts.

solved

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
umangyadav: > But you are right, I'm not sure how this is verified for attention either, I think we probably fail when lowering instead of at rock.attention verifiers. 

It is better to fail during verification than during lowering IMO.

mlir/test/rocmlir-gen/attention-kernel.mlir
dhernandez0: it's repeated in line 38 (same file)

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
dhernandez0: update: this PR only handles f32, so this will be fixed in future PRs

mlir/tools/rocmlir-tuning-driver/rocmlir-tuning-driver.cpp
```diff
@@ -330,10 +327,7 @@ static LogicalResult runTuningLoop(ModuleOp source) {
     tuneCopy->walk([&perfConfigAttr](rock::RockGemmWrapperInterface op) {
       op->setAttr("perf_config", perfConfigAttr);
     });
-    tuneCopy->walk([&perfConfigAttr](rock::AttentionOp op) {
-      op->setAttr("perf_config", perfConfigAttr);
-    });
-    tuneCopy->walk([&perfConfigAttr](rock::GemmElementwiseGemmOp op) {
+    tuneCopy->walk([&perfConfigAttr](rock::RockGemmGemmWrapperInterface op) {
```
umangyadav, line 331: Nice. Looks great with WrapperInterface.

mlir/include/mlir/Dialect/Rock/IR/RockGemmGemmWrapperInterface.td
```diff
@@ -0,0 +1,256 @@
...
+        /*args=*/(ins),
+        /*methodBody=*/"",
+        /*defaultImplementation=*/ ""
+      >,
+
+    InterfaceMethod<
+        /*desc=*/[{
+          Set the tuning parameters attribute of the first GEMM
+
+          This is needed for --affix-tuning-params to work and can go away if it does
```
umangyadav: Can go away if it doesn't require it ?

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -204,39 +205,35 @@ def Rock_ReduceOp :
...
+          UnitAttr:$kTransposed, UnitAttr:$vTransposed, UnitAttr:$oTransposed,
+          StrAttr:$arch, Rock_GemmFeaturesAttr:$features,
+          OptionalAttr<I32Attr>:$numCU,
+          OptionalAttr<RockTuningParamAttrInterface>:$params0,
+          OptionalAttr<RockTuningParamAttrInterface>:$params1,
+          I32Attr:$firstGemmIdx)>,
+      Results<(outs Optional<TensorOf<[F32, F16, BF16]>>:$result)> {
   let summary = "Attention operation of transformer models";
   let description = [{
-    Performs the operation out = SOFTMAX((queries * keys) .* scale) * values.
```
umangyadav, line 231: I think it meant `scale` for preSoftmaxElemwise Ops . Do we want to keep it ?

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
```diff
@@ -250,8 +247,47 @@ def Rock_AttentionOp :
...
+  let description = [{
+    Performs the operation out = (a * b) * c.
+
+    This operation performs fused GEMM-elementwise-GEMM. 
+
+    Those creating a `rock.gemm_elementwise_gemm` must specify the GPU architecture being targetted
+    and the number of compute units (numCu) available. The parameters
+    `gridSize`, and `blockSize` are optional as they can be inferred by
+    a tuning process or a heuristic, but they must be set before the `gemm_elementwise_gemm` is
+    lowered into the `gridwise_attention_accel` stage of the code generation pipeline.
```
umangyadav, line 297: nit: we may want to change `attention_accel` to more generic name if we are going to support `conv + elem + gemm` and such fusions in future.

mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -2068,77 +2077,188 @@ LogicalResult BlockwiseFillOp::verify() {
...
+          offsetC = dimsC.size() == 2 ? 0 : 1;
+  int64_t g = offsetA ? dimsA[0] : 1,
+          m = dimsA[offsetA + (getATransposed() ? 1 : 0)],
+          k = dimsA[offsetA + (getATransposed() ? 0 : 1)],
+          n = dimsB[offsetB + (getBTransposed() ? 0 : 1)],
+          o = dimsC[offsetC + (getCTransposed() ? 1 : 0)];
+  return GemmGemmSize(g, m, k, n, o);
+}
+
+static LogicalResult verifyAttentionOp(RockGemmGemmWrapperInterface op,
```
umangyadav: nit:
```suggestion
static LogicalResult verifyGemmPlusGemmLikeOp(RockGemmGemmWrapperInterface op,
```

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
dhernandez0: you can see here: https://github.com/ROCm/rocMLIR/commit/b1e26dbf9f44f1bbd4d3de180cc38782a1a4c2f1
scale was an actual input for attentionop, I guess that was before pre-softmax fusion was implemented. So, it made sense to have scale there as it was an input. Now, scale is just one possible fusion. So, to me it doesn't make sense to have it in the description anymore. What do you think?

umangyadav (approved):
I don't have any pressing comments in particular. Looks good.

mlir/include/mlir/Dialect/Rock/IR/RockGemmGemmWrapperInterface.td
dhernandez0: sorry, this is just copy-paste from RockGemmWrapperInterface. Not sure what the context of this is, we use the method, so it can't go away currently. Should I remove the sentence?

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
umangyadav: Perhaps we can mention it also does elementwise op before doing softmax.

mlir/include/mlir/Dialect/Rock/IR/RockGemmGemmWrapperInterface.td
umangyadav: Yes please remove the sentence

mlir/include/mlir/Dialect/Rock/IR/RockOps.td
dhernandez0: I agree, I'd prefer to fix this in the future, because it affects attention as well. Also, in practice I'm not sure if this is an issue when lowering from migraphx or only in hand-crafted rock IR kernels.

PR #1776, opened 2025-03-13
Add requirements.txt file and modify Dockerfile

dorde-antic:
Required for https://github.com/ROCm/rocMLIR/pull/1762


mlir/requirements.txt
dhernandez0, line 1: I think typically requirements.txt is in the root folder of the project

dhernandez0 (approved):
I'd prefer requirements.txt to be in the root because most projects do that. Approving anyway

umangyadav (approved):

PR #1779, opened 2025-03-17
Add rocprofv3 to perfRunner

djramic:
This commit replace rocprof with rocprofv3
Close: https://github.com/ROCm/rocMLIR-internal/issues/1722


mlir/utils/performance/perfRunner.py
```diff
@@ -161,12 +162,12 @@ def getBankConflict(fileName):
     with open(fileName, 'r') as csv_file:
         reader = csv.DictReader(csv_file, delimiter = ',')
         header = reader.fieldnames
-        if 'LDSBankConflict' not in header:
+        if 'Counter_Value' not in header:
```
dhernandez0: what is the format now? something like this?
Counter_Name, Counter_Value
LDSBankConflicts, X

mlir/utils/performance/perfRunner.py
```diff
@@ -23,10 +23,11 @@
 from perfCommonUtils import Operation, GEMMLibrary
 
 # global variables.
-ROCPROF = '/opt/rocm/bin/rocprof'
```
dhernandez0, line 27: should we still have rocprof compatibility? are old archs supported by rocprofv3?

mlir/utils/performance/perfRunner.py
```diff
@@ -137,7 +138,7 @@ def getNanoSeconds(fileName):
 
         result = 0
         for row in reader:
-            result += int(row['AverageNs'])
+            result += int(float(row['AverageNs']))
```
dhernandez0, line 143: is it float or int?

mlir/utils/performance/perfRunner.py
```diff
@@ -897,7 +898,7 @@ def benchmarkExternal(cls, commandLine, paths: Paths, arch, numCU):
             raise ValueError("rocblas-benchmark-driver not built")
         benchmarkArgs = config.generateMlirDriverCommandLine("")
         # remove the result file generated by rocprof in previous benchmarking
-        os.system("rm -f "+BENCHMARKING_RESULT_FILE_NAME)
+        os.system("rm -f "+BENCHMARKING_STATS_FILE_NAME)
```
dhernandez0, line 1098: can we use os.remove(), that way we'll know if it fails.

mlir/utils/performance/perfRunner.py
```diff
@@ -929,14 +930,14 @@ def benchmarkExternal(cls, commandLine, paths: Paths, arch, numCU):
 
 def runConfigWithMLIR(config: PerfConfiguration, paths: Paths, arch, rocmlir_gen_flags, debug=True):
     # remove the result file generated by rocprof in previous benchmarking
-    os.system("rm -f "+BENCHMARKING_RESULT_FILE_NAME)
+    os.system("rm -f "+BENCHMARKING_STATS_FILE_NAME)
```
dhernandez0, line 1130: same here os.remove

mlir/utils/performance/perfRunner.py
```diff
@@ -1092,7 +1093,7 @@ def getFusionTestInfo(filename, paths: Paths):
     return testEntry
 
 def runFusionKernel(filename, rocmlirGenArgs, paths: Paths):
-    os.system("rm -f "+BENCHMARKING_RESULT_FILE_NAME)
+    os.system("rm -f "+BENCHMARKING_STATS_FILE_NAME)
```
dhernandez0, line 1293: same here

mlir/utils/performance/perfRunner.py
```diff
@@ -929,14 +930,14 @@ def benchmarkExternal(cls, commandLine, paths: Paths, arch, numCU):
...
     commandLineOptions = config.generateMlirDriverCommandLine(rocmlir_gen_flags)
     if debug:
         print("Running MLIR Benchmark: ", repr(config))
     rocmlirGenCommand = paths.mlir_paths.rocmlir_gen_path + ' -ph ' + commandLineOptions
     rocmlirDriverCommand = [paths.mlir_paths.rocmlir_driver_path, '-c']
     mlir_cpu_runner_args = [f'--shared-libs={paths.mlir_paths.libmlir_rocm_runtime_path},{paths.mlir_paths.libconv_validation_wrappers_path},{paths.mlir_paths.libmlir_runtime_utils_path},{paths.mlir_paths.libmlir_c_runner_utils_path}', '--entry-point-result=void']
-    profilerCommand = [ROCPROF] +  getMetricArgsForRocprof(arch) + ['--stats', '-o', BENCHMARKING_METRICS_FILE_NAME, paths.mlir_paths.cpu_runner_path] + mlir_cpu_runner_args
+    profilerCommand = [ROCPROF] + getMetricArgsForRocprof(arch) + ['--kernel-trace', '--stats', '-o', BENCHMARKING_RESULT_FILE_NAME, '--' ,paths.mlir_paths.cpu_runner_path] + mlir_cpu_runner_args
```
dhernandez0, line 1137: what is '--' just before cpu runner?

djramic: rocprofv3 expects '--' before invoking application

mlir/utils/performance/perfRunner.py
djramic: It's currently a float, with rocprofv3. I cast it to int to preserve the existing logic

mlir/utils/performance/perfRunner.py
djramic: I'm not sure whether the rocprof version affects the old architecture. We previously agreed that using only rocprofv3 is fine

mlir/utils/performance/perfRunner.py
djramic: Right, this is how the output looks now

mlir/utils/performance/perfRunner.py
djramic: I think it's fine to keep it this way since the result file doesn't exist on the first run

mlir/utils/performance/perfRunner.py
dhernandez0: then, we should check that the first data element is 'LDSBankConflicts' as well

mlir/utils/performance/perfRunner.py
dhernandez0: Not sure what we agreed, but we need to keep support for all architectures we run in CI. So, if v3 doesn't support them, that's an issue.

mlir/utils/performance/perfRunner.py
```diff
@@ -23,10 +23,11 @@
 from perfCommonUtils import Operation, GEMMLibrary
 
 # global variables.
-ROCPROF = '/opt/rocm/bin/rocprof'
+ROCPROF = '/opt/rocm/bin/rocprofv3'
```
umangyadav, line 27: Do you we need to change dockerfile to install rocprofv3  or does it come pre-installed with rocm ?

mlir/utils/performance/perfRunner.py
djramic: Lets run it trough CI and see. We can always rollback.

dhernandez0:
if we don't need rocprof v1, please remove the workaround in the Dockerfile:
https://github.com/ROCm/rocMLIR/blob/5bc802422471efa2d1aadb1df961f7575729d540/mlir/utils/jenkins/Dockerfile#L158

mlir/utils/performance/perfRunner.py
djramic: It comes preinstalled with rocm

umangyadav (approved):
LGTM, But make sure @dhernandez0  comments' are resolved before merging

PR #1784, opened 2025-03-19
Fix initialization for split-k

dhernandez0:
In this PR we fix some issues related to split-k tensor initialization:
- Make sure rocmlir-gen code to init by default to 100 actually works (MHALToGPU.cpp)
- Set prefill flags in TosaToRock similarly to what we do for reductions
- rocmlir-gen init output tensor to zero when not cloning and split-k is enabled

### DIFF
--- a/mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
+++ b/mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
@@ -17,6 +17,7 @@
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/Rock/IR/Rock.h"
 #include "mlir/Dialect/Rock/IR/TransformMapBuilder.h"
+#include "mlir/Dialect/Rock/Tuning/RockTuning.h"
 #include "mlir/Dialect/Rock/utility/AmdArchDb.h"
 #include "mlir/Dialect/Rock/utility/builderUtils.h"
 #include "mlir/Dialect/Rock/utility/loweringUtils.h"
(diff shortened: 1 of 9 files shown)


mlir/test/rocmlir-driver/populate_random_data_splitk.mlir
```diff
@@ -0,0 +1,66 @@
+// RUN: rocmlir-gen --arch gfx90a:sramecc+:xnack- -perf_config="v3:64,64,8,16,16,4,4,1,2,1,1" -ph -p -rand=none | rocmlir-opt -canonicalize | FileCheck %s --check-prefix=NONE
```
umangyadav, line 1: Q: what `-p` does ?

dhernandez0: Sets default values (for conv/gemm or attention). This is just copy-pasted from populate_random_data.mlir (with different CHECKS for split-k). I thought it would be easier to have another file rather than *2 the RUNs.

umangyadav (approved):
lgtm

mlir/lib/Conversion/TosaToRock/TosaToRock.cpp
```diff
@@ -335,21 +335,65 @@ static Value traceToRes(Value tensor, DenseMap<Value, Value> &cache,
...
+            op, "We only support F32, F16 and BF16 split-k, yet.");
+      }
+      Attribute outputInitVal = rw.getFloatAttr(elementType, 0.0);
+      func.setResultAttr(resNumber, rock::PrefillAttr::getMnemonic(),
+                         outputInitVal);
+      func.setResultAttr(resNumber, "mhal.read_access", rw.getUnitAttr());
+      // The original function also need the read access attr for the output.
+      if (func->hasAttr("original_func")) {
+        if (ModuleOp rootMod = func->getParentOfType<ModuleOp>()
+                                   ->getParentOfType<ModuleOp>()) {
```
umangyadav, line 383: Is there a test to check this ?

dhernandez0: no test, good catch. I'll add that.

dhernandez0: I'm not sure how to test this. The test would have to be:

```mlir
module {
  module {
    func.func @test_add_two_outputs(...) attributes {kernel, original_func = @test_add_two_outputs_original} {
    }
  }
  // CHECK: @test_add_two_outputs_original
  // CHECK-SAME: -> (tensor<2x128x256xf32> {mhal.read_access}
  // CHECK-SAME: tensor<2x128x256xf32> {mhal.read_access})
  func.func @test_add_two_outputs_original(...) attributes {} {
  }
}
```

Because test_add_two_outputs_original has to be in the parent module. However, I can't get rocmlir-opt to run the pass with this test. Not sure if there's something that has to be done to allow this kind of test?

PR #1788, opened 2025-03-21
Remove Machine Names Due to Security Team Advisory

stefankoncarevic:
In response to a warning from the security team regarding exposed machine names in our public Jenkinsfile.migraphxintegration, I have removed these references to better protect our infrastructure. This CI script has not been in use for some time, and there is an open ticket to update it in the future. https://github.com/ROCm/rocMLIR-internal/issues/1730 When it's updated, we will devise a secure method for handling machine configurations and validate models across various architectures.


dhernandez0 (approved):

umangyadav (commented):
Would it be better to move this script in internal repo ?

stefankoncarevic:
> Would it be better to move this script in internal repo ?

I think that also good suggestion. We can move script to rocmlir-internal repo.

dhernandez0:
> Would it be better to move this script in internal repo ?

we can add the machine list to the ticket: https://github.com/ROCm/rocMLIR-internal/issues/1730
and whoever takes care of that, can actually figure out how to use them without exposing them.

By the way, the list of machines is still visible in git history, and this PR for example. Should we find a way to remove them?

umangyadav:
> we can add the machine list to the ticket: https://github.com/ROCm/rocMLIR-internal/issues/1730 

This is okay too for now. 

>  By the way, the list of machines is still visible in git history, and this PR for example. Should we find a way to remove them? 

Yes but we will have to do it from all the forks/branches too

PR #1789, opened 2025-03-21
Remove fp8 check on nightly CI.

stefankoncarevic:
This causes a problem with nightly CI on mi300 inside Benchmark and Report stage. Remove fp8 check on nightly CI because it is not supported inside the MIOpen driver for now. I will open a ticket on their side to add support on driver for this type.

### DIFF
--- a/mlir/utils/performance/perfRunner.py
+++ b/mlir/utils/performance/perfRunner.py
@@ -277,7 +277,7 @@ def getConvConfigurations(fileName):
                 
                 # Skip unsupported datatypes
                 if datatype == 'convfp8':
-                    unsupported_chips = {'gfx908', 'gfx90a', 'gfx1030', 'gfx1101'}
+                    unsupported_chips = {'gfx908', 'gfx90a', 'gfx942', 'gfx1030', 'gfx1101'}
                     if getChip() in unsupported_chips:
                         continue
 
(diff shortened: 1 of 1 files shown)


umangyadav (approved):

mlir/utils/performance/perfRunner.py
```diff
@@ -277,7 +277,7 @@ def getConvConfigurations(fileName):
                 
                 # Skip unsupported datatypes
                 if datatype == 'convfp8':
-                    unsupported_chips = {'gfx908', 'gfx90a', 'gfx1030', 'gfx1101'}
+                    unsupported_chips = {'gfx908', 'gfx90a', 'gfx942', 'gfx1030', 'gfx1101'}
```
dhernandez0, line 280: it's ok for now. But I think we should disable MIOpen only. This disables perfRunner completely and we might want to run fp8 tuning.

dhernandez0: can we open a ticket for a proper fix?

stefankoncarevic: I'll create a ticket to ensure we can implement a proper fix.

dhernandez0 (approved):

mlir/utils/performance/perfRunner.py
stefankoncarevic: This is the link to ticket https://github.com/ROCm/rocMLIR-internal/issues/1790
I will merge now.

PR #1790, opened 2025-03-24
Introduce native arch support in arch database

mirza-halilcevic:
- Resolves https://github.com/ROCm/rocMLIR-internal/issues/1745


mlir/lib/Dialect/Rock/utility/CMakeLists.txt
```diff
@@ -19,4 +19,14 @@ add_rocmlir_dialect_library(MLIRRockUtility
...
+
+find_package(hip REQUIRED)
+# We do need both of these.  One is for linking, to get the HIP
+# functions.  The other is for compiling, to get -D__HIP_PLATFORM_AMD__
+target_link_libraries(MLIRRockUtility PUBLIC
+  hip::host hip::amdhip64
+)
+target_link_libraries(obj.MLIRRockUtility PUBLIC
+  hip::host hip::amdhip64
 )
```
dhernandez0, line 21: new line

mlir/unittests/Dialect/Rock/AmdArchDbTests.cpp
```diff
@@ -0,0 +1,20 @@
+//===- TransformMapBuilderTests.cpp - Tests for the Rock Transform
```
dhernandez0: update comment

mlir/unittests/Dialect/Rock/AmdArchDbTests.cpp
```diff
@@ -0,0 +1,20 @@
...
+#include "mlir/Dialect/Rock/utility/AmdArchDb.h"
+
+#include "gtest/gtest.h"
+#include "gtest/internal/gtest-internal.h"
+
+using namespace mlir::rock;
+
+TEST(AmdArchDbTest, NativeArch) {
+  auto info = lookupArchInfo("native");
+  EXPECT_EQ(info.totalVGPRPerEU, 256);
```
dhernandez0: the current arch in the system could have a different number, right?

dhernandez0:
> * Resolves [Improve gpu attribute collection technique rocMLIR-internal#1745](https://github.com/ROCm/rocMLIR-internal/issues/1745)

I think apart from the "native" feature. We need to have a python version of AmdArchDb, because right now we hardcode lots of hardware specific details in our python codebase. So, I guess let's create a ticket for that and reduce the scope of https://github.com/ROCm/rocMLIR-internal/issues/1745 to be only for C++.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -73,11 +75,36 @@ static constexpr AmdArchInfo
...
+                 << "; Falling back to defaults\n";
+    return gcnInfo;
+  }
+
+  auto ret = lookupArchInfo(prop.gcnArchName); // get baseline
+  ret.waveSize = prop.warpSize;
+  // are these the correct values to use?
+  ret.totalSharedMemPerCU = prop.sharedMemPerMultiprocessor;
+  ret.maxSharedMemPerWG = prop.sharedMemPerBlock;
+
```
dhernandez0: is there a way to get the number of CUs? that can change even with the same arch id (gfx90a could be MI210 and MI250, and numCU is different)

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -73,11 +75,36 @@ static constexpr AmdArchInfo
...
   // Keep this implementation in sync with
   // mlir/test/lit.site.cfg.py.in:set_arch_features()
   StringRef firstPart, remainingParts;
   std::tie(firstPart, remainingParts) = arch.split(':');
+  if (firstPart == "native") {
+    unsigned long long deviceId;
+    if (!llvm::getAsUnsignedInteger(remainingParts, 0, deviceId)) {
+      return fetchNativeArchInfo(deviceId);
+    }
+    return fetchNativeArchInfo();
```
dhernandez0: I'd prefer to fail if no gpu id is provided. What do you guys think?

mlir/unittests/Dialect/Rock/AmdArchDbTests.cpp
mirza-halilcevic: Yes, of course. I just used this test as a convenient way to experiment. This will be changed later.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
mirza-halilcevic: Yes, there is `prop.multiProcessorCount`. One thing to note though is that `prop.multiProcessorCount` on _gfx1200_ shows half the number `rocminfo` shows. For example, if `rocminfo` shows 32 CUs, `prop.multiProcessorCount` will be 16. So we might need to multiply that value by 2, depending on what the caller expects. I still need to confirm if the behavior is the same on other architectures.

dhernandez0: that only applies to navi because of WGP Mode I think? @umangyadav

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
mirza-halilcevic: From what I saw, CUDA and HIP generally default to device 0 if not specified otherwise. That's the convention I followed here.

But either way is fine with me if you want me to change this. I guess it depends on how often we expect to encounter non-homogenous systems, and how much of a hassle it is for the caller to attach a device ID to the argument.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
umangyadav: Yes by default we compile Navi machines for Workgroup processor mode which counts 2 CUs == 1CUs. 

That is why @mirza-halilcevic  you see half the number of CUs. We should use half the number of CUs for the Navi in archdb compared to what rocminfo shows.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -73,11 +75,36 @@ static constexpr AmdArchInfo
               /*hasFp8ConversionInstrs=*/false,
               /*hasOcpFp8ConversionInstrs=*/false, /*maxNumXCC=*/1);
 
+static AmdArchInfo fetchNativeArchInfo(unsigned long long deviceId = 0) {
```
umangyadav: nit : you don't need to use  very wide dtype for `deviceId`

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -73,11 +75,36 @@ static constexpr AmdArchInfo
...
+  if (auto err = hipGetDeviceProperties(&prop, deviceId); err != hipSuccess) {
+    llvm::errs() << "hipGetDeviceProperties error: " << hipGetErrorString(err)
+                 << "; Device ID: " << deviceId
+                 << "; Falling back to defaults\n";
+    return gcnInfo;
+  }
+
+  auto ret = lookupArchInfo(prop.gcnArchName); // get baseline
+  ret.waveSize = prop.warpSize;
+  // are these the correct values to use?
```
umangyadav: Yes they look correct to me.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -73,11 +75,36 @@ static constexpr AmdArchInfo
...
+  hipDeviceProp_t prop;
+  if (auto err = hipGetDeviceProperties(&prop, deviceId); err != hipSuccess) {
+    llvm::errs() << "hipGetDeviceProperties error: " << hipGetErrorString(err)
+                 << "; Device ID: " << deviceId
+                 << "; Falling back to defaults\n";
+    return gcnInfo;
+  }
+
+  auto ret = lookupArchInfo(prop.gcnArchName); // get baseline
+  ret.waveSize = prop.warpSize;
```
umangyadav: Can you please check what warpSize are you getting for the Navi ?

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
umangyadav: >From what I saw, CUDA and HIP generally default to device 0 if not specified otherwise. That's the convention I followed here. 

I agree with keeping this convention

mlir/lib/Dialect/Rock/utility/CMakeLists.txt
```diff
@@ -19,4 +19,14 @@ add_rocmlir_dialect_library(MLIRRockUtility
...
+)
+
+find_package(hip REQUIRED)
+# We do need both of these.  One is for linking, to get the HIP
+# functions.  The other is for compiling, to get -D__HIP_PLATFORM_AMD__
+target_link_libraries(MLIRRockUtility PUBLIC
+  hip::host hip::amdhip64
+)
+target_link_libraries(obj.MLIRRockUtility PUBLIC
+  hip::host hip::amdhip64
```
umangyadav: This is strange that you need to use `target_link_libraries` twice. I think even if you use target_link_libraries with `MLIRRockUtility` it should work

umangyadav:
> So, I guess let's create a ticket for that and reduce the scope of https://github.com/ROCm/rocMLIR-internal/issues/1745 to be only for C++. 

I wonder if we can use pybind or some other mechanism available in upstream llvm to use C++ API from Python

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
dhernandez0: Sounds reasonable although I prefer the "keep it simple" principle (we can require migraphx to provide the gpu id). If you guys prefer to default to 0, let's do it.

mlir/lib/Dialect/Rock/utility/CMakeLists.txt
mirza-halilcevic: I copied this from [here](https://github.com/ROCm/rocMLIR/blob/b6c726a324d2807494f770ed915af0e54cc6cb3f/external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt#L414). If I leave out the second one, I get compilation errors because `__HIP_PLATFORM_AMD__` isn't defined.

umangyadav: you can try doing 
`target_compile_definitions(MLIRRockUtility PRIVATE __HIP_PLATFORM_AMD__)`

umangyadav: https://github.com/ROCm/rocm-examples/blob/f9d4e5e78325c36b319d91ec37c6410b2b6e12fb/HIP-Basic/device_query/CMakeLists.txt#L83

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -71,18 +78,129 @@ static constexpr AmdArchInfo
               /*totalVGPRPerEU*/ 1536, /*totalSharedMemPerCU*/ 131072,
               /*maxSharedMemPerWG*/ 65536, /*numEUPerCU=*/4, /*minNumCU=*/12,
               /*hasFp8ConversionInstrs=*/false,
-              /*hasOcpFp8ConversionInstrs=*/false, /*maxNumXCC=*/1);
+              /*hasOcpFp8ConversionInstrs=*/false, /*maxNumXCC=*/1),
+    gfx12Info(GemmFeatures::dot | GemmFeatures::atomic_add |
```
dhernandez0: rdna4Info?

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -71,18 +78,129 @@ static constexpr AmdArchInfo
...
+                  GemmFeatures::atomic_fmax_f32 | GemmFeatures::wmma |
+                  GemmFeatures::atomic_add_f16 | GemmFeatures::atomic_add_bf16,
+              /*waveSize=*/32, /*maxWavesPerEU*/ 16, /*totalSGPRPerEU*/ 800,
+              /*totalVGPRPerEU*/ 1536, /*totalSharedMemPerCU*/ 131072,
```
dhernandez0, line 103: totalSharedMemPerCU was 65536 in the existing code. Is this a mistake or is 131072 the correct value?

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -71,18 +78,129 @@ static constexpr AmdArchInfo
...
+std::unique_ptr<const GCNTargetMachine>
+createTargetMachine(StringRef cpu, StringRef featureString = "") {
+  static std::once_flag flag;
+  std::call_once(flag, [] {
+    LLVMInitializeAMDGPUTargetInfo();
+    LLVMInitializeAMDGPUTarget();
+    LLVMInitializeAMDGPUTargetMC();
+  });
+
+  Triple triple("amdgcn-amd-");
```
dhernandez0: shouldn't this be amdgcn-amd-amdhsa?

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -71,18 +78,129 @@ static constexpr AmdArchInfo
...
     std::tie(firstPart, remainingParts) = remainingParts.split(':');
   }
-  StringRef chip = firstPart;
 
+  std::get<0>(ret) = firstPart;
+  return ret;
+}
+
+std::unique_ptr<const GCNTargetMachine>
+createTargetMachine(StringRef cpu, StringRef featureString = "") {
```
dhernandez0: cpu -> chip? as this won't be a cpu

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -71,18 +78,129 @@ static constexpr AmdArchInfo
...
+  if (!target) {
+    llvm::errs() << "WARNING: Target registry lookup failed with error: "
+                 << error << ".\n";
+    return nullptr;
+  }
+
+  TargetOptions options;
+  return std::unique_ptr<GCNTargetMachine>(
+      static_cast<GCNTargetMachine *>(target->createTargetMachine(
+          triple, cpu, featureString, options, std::nullopt, std::nullopt)));
```
dhernandez0: nit: "options, std::nullopt, std::nullopt" -> "{}, {}" (as in ModuleToObject.cpp)

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -71,18 +78,129 @@ static constexpr AmdArchInfo
...
+std::enable_if_t<std::is_assignable_v<LHS &, RHS &&>, void>
+checkAndSetInfo(StringRef name, LHS &lhs, RHS &&rhs) {
+  if (lhs != static_cast<LHS>(rhs)) {
+    llvm::outs() << "NOTE: Value discrepancy for " << name << ": " << lhs
+                 << " != " << rhs << ". Proceeding with " << rhs << ".\n";
+    lhs = std::forward<RHS>(rhs);
+  }
+}
+
+AmdArchInfo fetchNativeArchInfo(unsigned deviceId = 0) {
```
dhernandez0: I guess ROCR_VISIBLE_DEVICES / HIP_VISIBLE_DEVICES etc filter out gpus so that gpu 0 is already what we want, right?

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -71,18 +78,129 @@ static constexpr AmdArchInfo
...
+  llvm::outs() << "Fetching native arch info for device " << deviceId
+               << "...\n";
+
+  hipDeviceProp_t prop;
+  if (auto err = hipGetDeviceProperties(&prop, deviceId); err != hipSuccess) {
+    llvm::errs() << "WARNING: hipGetDeviceProperties failed with error: "
+                 << hipGetErrorString(err) << ". Falling back to defaults.\n";
+    return gcnInfo;
```
dhernandez0: The default for rdna should be rdnaNoDotInfo. Otherwise, the waveSize would be wrong. I guess we don't know if it's navi or MI at this point... Not sure what is the best to do. Maybe just fail?

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -71,18 +78,129 @@ static constexpr AmdArchInfo
...
+                  st.getMaxWavesPerEU());
+  checkAndSetInfo("(LLVM) totalSGPRPerEU", ret.totalSGPRPerEU,
+                  st.getTotalNumSGPRs());
+  checkAndSetInfo("(LLVM) totalVGPRPerEU", ret.totalVGPRPerEU,
+                  st.getTotalNumVGPRs());
+  checkAndSetInfo("(LLVM) waveSize", ret.waveSize, st.getWavefrontSize());
+  checkAndSetInfo("(LLVM) totalSharedMemPerCU", ret.totalSharedMemPerCU,
+                  st.getAddressableLocalMemorySize());
+  checkAndSetInfo("(LLVM) maxSharedMemPerWG", ret.maxSharedMemPerWG,
+                  st.getLocalMemorySize());
```
dhernandez0: I guess the only missing things are:
- fp8/ocp -> you can get that with amdgpu::hasOcpFp8(chipset), see rocmlir-gen.cpp
- maxNumXCC
- GemmFeatures: mfma/wmma, dot, atomic_add, atomic_add_f16, atomic_add_bf16, atomic_fmax_f32

Any chance we can get those as well from LLVM?

mlir/lib/Dialect/Rock/utility/Bindings/AmdArchDbBindings.cpp
```diff
@@ -0,0 +1,55 @@
...
+namespace py = pybind11;
+
+mlir::rock::AmdArchInfo lookupArchInfoWrapper(const std::string &arch) {
+  return mlir::rock::lookupArchInfo(arch);
+}
+
+PYBIND11_MODULE(amd_arch_db, m) {
+  m.doc() = "Database of AMD GPU features";
+
+  py::enum_<mlir::rock::GemmFeatures>(m, "GemmFeatures")
```
dhernandez0, line 19: add comment in GemmFeatures definition to keep it in sync with python bindings

mlir/lib/Dialect/Rock/utility/Bindings/AmdArchDbBindings.cpp
```diff
@@ -0,0 +1,55 @@
...
+      .value("NONE", mlir::rock::GemmFeatures::none)
+      .value("MFMA", mlir::rock::GemmFeatures::mfma)
+      .value("WMMA", mlir::rock::GemmFeatures::wmma)
+      .value("DOT", mlir::rock::GemmFeatures::dot)
+      .value("ATOMIC_ADD", mlir::rock::GemmFeatures::atomic_add)
+      .value("ATOMIC_ADD_BF16", mlir::rock::GemmFeatures::atomic_add_bf16)
+      .value("ATOMIC_ADD_F16", mlir::rock::GemmFeatures::atomic_add_f16)
+      .value("ATOMIC_FMAX_F32", mlir::rock::GemmFeatures::atomic_fmax_f32);
+
+  py::class_<mlir::rock::AmdArchInfo>(m, "AmdArchInfo")
```
dhernandez0, line 29: add comment in AmdArchInfo definition to keep it in sync with python bindings

mlir/lib/Dialect/Rock/utility/Bindings/AmdArchDbBindings.cpp
dhernandez0, line 1: nice! thanks, this will improve the python code. Will you change the python code to use this here or in a follow-up PR?

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
mirza-halilcevic: I believe that the previous value was incorrect. If we look at the others, we can see that for RDNA, the `totalSharedMemPerCU` is two times the `maxSharedMemPerWG`.

LLVM gives us the following values:
```
AddressableLocalMemorySize: 65536
LocalMemorySize: 131072
```
It seems like the `LocalMemorySize` should correspond to `totalSharedMemPerCU`.

HIP returns the following:
```
MaxSharedMemoryPerBlock: 65536
MaxSharedMemoryPerMultiprocessor: 65536
SharedMemPerMultiprocessor: 1048576 // NOTE: No idea what this value represents
```
That's why the initial mistake was made.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
mirza-halilcevic: Sure. In that case we should also change `gfx11Info` to `rdna3Info`. And regarding `gfx950Info` in https://github.com/ROCm/rocMLIR/pull/1802, maybe we can change that to `cdna3_5Info` or something similar?

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
mirza-halilcevic: CPU is the terminology that the LLVM Target related identifiers use, regardless of the actual architecture. That's why I kept it like this, but it's no trouble to change it if you think it's semantically more appropriate.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
mirza-halilcevic: I used [this](https://github.com/llvm/llvm-project/blob/209252f3d5bf66c8bff0863ba7617e37ead4fff8/llvm/unittests/Target/AMDGPU/AMDGPUUnitTests.cpp#L173) as a reference. It seems to be what LLVM is expecting.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
mirza-halilcevic: Yes. CUDA has the same behavior. The device IDs get remapped so that from the programmer's point of view, the device ID is always `[0, N-1]`.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
mirza-halilcevic: That's a good point. If you look at the `lookupArchInfo` implementation, it also returns `gcnInfo` as a fallback always. I would agree that it's probably best to just fail.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
mirza-halilcevic: I'll take a look if we can fetch those too.

mlir/lib/Dialect/Rock/utility/Bindings/AmdArchDbBindings.cpp
mirza-halilcevic: I'm leaning towards doing it in a separate PR. What would you suggest?

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
mirza-halilcevic: On Navi4x, I get a warp size of 32.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
dhernandez0: I think our code always generates amdgcn-amd-amdhsa (you can see that in our tests if you search "amdgcn-amd-". I'm not entirely sure of the differences between them: https://llvm.org/docs/AMDGPUUsage.html#amdgpu-target-triples

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
dhernandez0: let's fail then? @umangyadav

mlir/lib/Dialect/Rock/utility/Bindings/AmdArchDbBindings.cpp
dhernandez0: yes, separate PR is a good idea

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
dhernandez0: yes, that sounds right based on the comments here: https://github.com/llvm/llvm-project/blob/f04bfbc4162247c0074fdd87068053e4264dbefa/llvm/lib/Target/AMDGPU/AMDGPUSubtarget.h#L269

we use WGP mode for navi

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
dhernandez0: yes, that sounds right based on the comments here (we use WGP mode for navi): https://github.com/llvm/llvm-project/blob/f04bfbc4162247c0074fdd87068053e4264dbefa/llvm/lib/Target/AMDGPU/AMDGPUSubtarget.h#L269

Thanks!

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
dhernandez0: for atomic, there are:
- hasAtomicBufferPkAddBF16Inst: we use buffer atomics only for now for f16/bf16
- HasAtomicBufferGlobalPkAddF16NoRtnInsts and HasAtomicBufferGlobalPkAddF16Insts 
- HasAtomicFMinFMaxF32GlobalInsts (I guess it's this one for atomic_fmax_f32)
- hasAtomicFaddInsts (atomic_add)

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
mirza-halilcevic: Looks like both work fine. I set it to `amdgcn-amd-amdhsa`.

umangyadav:
@mirza-halilcevic  Can you resolve merge conflicts ?

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -73,24 +80,183 @@ static constexpr AmdArchInfo
...
+      features |= GemmFeatures::atomic_fmax_f32;
+    } else {
+      features &= ~GemmFeatures::atomic_fmax_f32;
+    }
+
+    checkAndSetInfo("(LLVM) defaultFeatures", ret.defaultFeatures, features);
+
+    checkAndSetInfo(
+        "(LLVM) hasFp8ConversionInstrs", ret.hasFp8ConversionInstrs,
+        st.hasFP8ConversionInsts()); // TODO double check the meaning of this
```
dhernandez0: It looks like this: st.hasFP8ConversionInsts() tells us if it has fp8+bf8 support or not. 
So, if st.hasFP8ConversionInsts() == false:
- our hasFp8ConversionInstrs=false and our hasOcpFp8ConversionInstrs=false

if st.hasFP8ConversionInsts() == true:
- set our hasFp8ConversionInstrs=not amdgpu::hasOcpFp8(maybeChipset.value())
- set our hasOcpFp8ConversionInstrs=amdgpu::hasOcpFp8(maybeChipset.value())

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -113,29 +279,14 @@ AmdArchInfo mlir::rock::lookupArchInfo(StringRef arch) {
...
-
-    return gfx12Info;
+    return rdna4Info;
   }
-  llvm::errs() << "Warning: unknown architecture, falling back to defaults: "
-               << arch << "\n";
-  return gcnInfo;
+  llvm::errs() << "WARNING: Unsupported architecture: " << arch
+               << ". Falling back to defaults.\n";
+  return gcnInfo; // TODO fail instead of default?
```
dhernandez0: I think we should fail here, what do you think @umangyadav ? (for context: https://github.com/ROCm/rocMLIR/pull/1790#discussion_r2034806314)

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -73,24 +80,183 @@ static constexpr AmdArchInfo
...
+                      "values.\n";
+    } else {
+      checkAndSetInfo("(LLVM) hasOcpFp8ConversionInstrs",
+                      ret.hasOcpFp8ConversionInstrs,
+                      amdgpu::hasOcpFp8(maybeChipset.value()));
+    }
   }
-  StringRef chip = firstPart;
 
+  // TODO check and set maxNumXCC
```
dhernandez0: this is needed for https://github.com/ROCm/rocMLIR-internal/issues/1801 but we can skip it in this PR

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -73,24 +80,183 @@ static constexpr AmdArchInfo
...
+                    st.getTotalNumVGPRs());
+    checkAndSetInfo("(LLVM) waveSize", ret.waveSize, st.getWavefrontSize());
+    checkAndSetInfo("(LLVM) totalSharedMemPerCU", ret.totalSharedMemPerCU,
+                    st.getLocalMemorySize());
+    checkAndSetInfo("(LLVM) maxSharedMemPerWG", ret.maxSharedMemPerWG,
+                    st.getAddressableLocalMemorySize());
+
+    auto features = ret.defaultFeatures;
+
+    if (st.hasAtomicFaddInsts()) {
```
dhernandez0: I guess you got these from https://github.com/ROCm/rocMLIR/pull/1790#discussion_r2034908692 please check if they are correct, I didn't check. You can verify if the result makes sense for MI300, Navi4, etc...

dhernandez0: actually, as we don't have tests, instead of manually checking you can create tests that verify the values for some archs are the expected ones (gfx942, gfx1200, ...)

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
umangyadav: yes better to fail IMO

mlir/unittests/Dialect/Rock/AmdArchDbTests.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+TEST_P(NativeArchTest, NativeArchInfoMatchesPresetInfo) {
+  auto presetInfo = lookupArchInfo(prop.gcnArchName);
+  auto nativeInfo = lookupArchInfo("native:" + std::to_string(GetParam()));
+
+  EXPECT_EQ(presetInfo.defaultFeatures, nativeInfo.defaultFeatures);
+  EXPECT_EQ(presetInfo.waveSize, nativeInfo.waveSize);
+  SOFT_CHECK_EQ(presetInfo.maxWavesPerEU, nativeInfo.maxWavesPerEU);
+  SOFT_CHECK_EQ(presetInfo.totalSGPRPerEU, nativeInfo.totalSGPRPerEU);
+  SOFT_CHECK_EQ(presetInfo.totalVGPRPerEU, nativeInfo.totalVGPRPerEU);
```
dhernandez0: just out of curiosity, why can't we use EXPECT_EQ?

dhernandez0 (approved):

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -260,6 +263,9 @@ AmdArchInfo fetchNativeArchInfo(unsigned deviceId = 0) {
 AmdArchInfo mlir::rock::lookupArchInfo(StringRef arch) {
   // Keep this implementation in sync with
   // mlir/test/lit.site.cfg.py.in:set_arch_features()
+  if (arch.empty()) {
```
dhernandez0: shouldn't we fail if it's empty?

mirza-halilcevic: A lot of tests were failing when we introduced the `llvm_unreachable`. Most of them were invoking the function with an empty arch string. A few of them had the target triple concatenated without a delimiter, like this `amdgcn-amd-amdhsagfx942:sramecc+:xnack-`. Previously, that would work because it would return the fallback.

This is a workaround for the empty arch case. To me, going through the codebase and fixing the individual call sites of `lookupArchInfo` doesn't seem like the right approach. But let's discuss this sometime. FYI @umangyadav 

One other approach I considered is just doing a native fetch if the arch string is empty. But that caused the tests to fail with `loc("<stdin>":1:1): error: custom op 'Fetching' is unknown (tried 'builtin.Fetching' as well)`. I have yet to figure out why that is happening.

mlir/unittests/Dialect/Rock/AmdArchDbTests.cpp
mirza-halilcevic: Initially, the preset and native values for these parameters didn't match. I made it a soft check so the tests wouldn't fail, in case it was expected variation.

But after discussing with the team, we concluded that we shouldn't expect any variation in these values for different GPU variants of the same base arch. So, in the end, I changed it back to `EXPECT_EQ` and fixed the incorrect preset values.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
dhernandez0: I think fixing all the cases is a good approach but out of the scope of the PR. We can create a ticket to do that and remove the workarounds.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -73,24 +80,196 @@ static constexpr AmdArchInfo
...
+               << "...\n";
+
+  hipDeviceProp_t prop;
+  if (auto err = hipGetDeviceProperties(&prop, deviceId); err != hipSuccess) {
+    auto reason = "hipGetDeviceProperties failed with error: " +
+                  std::string(hipGetErrorString(err));
+    llvm::report_fatal_error(reason.c_str());
+  }
+
+  llvm::outs() << "gcnArchName: " << prop.gcnArchName << "\n";
```
dhernandez0: llvm::outs()  -> LLVM_DEBUG(dbgs() << "mesage");

dhernandez0: you have to define: 
#define DEBUG_TYPE "rock-amd-arch-db" for example. Then, we can print out using --debug-only=rock-amd-arch-db flag

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
dhernandez0: fixed this in https://github.com/ROCm/rocMLIR/pull/1862, once that's merged we can fail if arch is empty or not valid.

mlir/lib/Dialect/Rock/utility/Bindings/AmdArchDbBindings.cpp
```diff
@@ -0,0 +1,55 @@
...
+                    &mlir::rock::AmdArchInfo::maxSharedMemPerWG)
+      .def_readonly("num_eu_per_cu", &mlir::rock::AmdArchInfo::numEUPerCU)
+      .def_readonly("min_num_cu", &mlir::rock::AmdArchInfo::minNumCU)
+      .def_readonly("has_fp8_conversion_instrs",
+                    &mlir::rock::AmdArchInfo::hasFp8ConversionInstrs)
+      .def_readonly("has_ocp_fp8_conversion_instrs",
+                    &mlir::rock::AmdArchInfo::hasOcpFp8ConversionInstrs)
+      .def_readonly("max_num_xcc", &mlir::rock::AmdArchInfo::maxNumXCC);
+
+  m.def("lookup_arch_info", &lookupArchInfoWrapper);
```
umangyadav: Why do you have to create a wrapper ?

mlir/lib/Dialect/Rock/utility/CMakeLists.txt
umangyadav: > you can try doing `target_compile_definitions(MLIRRockUtility PRIVATE __HIP_PLATFORM_AMD__)`


@mirza-halilcevic  did you try this one ?

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
umangyadav: @mirza-halilcevic  this one is still unresolved and ideally we should put all prints behind Debug symbol as daniel suggested.

mlir/lib/Dialect/Rock/utility/CMakeLists.txt
mirza-halilcevic: Yes, I tried. But unfortunately, I couldn't get it to work with that approach.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
dhernandez0: https://github.com/ROCm/rocMLIR/pull/1862 merged. Please, update the code to fail if arch is empty.

mlir/lib/Dialect/Rock/utility/Bindings/AmdArchDbBindings.cpp
mirza-halilcevic: Becuase `lookupArchInfo` takes an `llvm::StringRef` and that causes problems with pybind11. I'll try wrapping this in a lambda instead, it'll be cleaner like that.

umangyadav (approved):

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -74,24 +83,207 @@ static constexpr AmdArchInfo
...
+  hipDeviceProp_t prop;
+  if (auto err = hipGetDeviceProperties(&prop, deviceId); err != hipSuccess) {
+    auto reason = "hipGetDeviceProperties failed with error: " +
+                  std::string(hipGetErrorString(err));
+    llvm::report_fatal_error(reason.c_str());
+  }
+
+  LLVM_DEBUG(llvm::dbgs() << "gcnArchName: " << prop.gcnArchName << "\n");
+
+  std::lock_guard<std::mutex> lock(m);
```
umangyadav, line 309: nit : define mutex near it's use.

mlir/lib/Dialect/Rock/utility/Bindings/AmdArchDbBindings.cpp
```diff
@@ -0,0 +1,53 @@
...
+                    &mlir::rock::AmdArchInfo::totalSharedMemPerCU)
+      .def_readonly("max_shared_mem_per_wg",
+                    &mlir::rock::AmdArchInfo::maxSharedMemPerWG)
+      .def_readonly("num_eu_per_cu", &mlir::rock::AmdArchInfo::numEUPerCU)
+      .def_readonly("min_num_cu", &mlir::rock::AmdArchInfo::minNumCU)
+      .def_readonly("has_fp8_conversion_instrs",
+                    &mlir::rock::AmdArchInfo::hasFp8ConversionInstrs)
+      .def_readonly("has_ocp_fp8_conversion_instrs",
+                    &mlir::rock::AmdArchInfo::hasOcpFp8ConversionInstrs)
+      .def_readonly("max_num_xcc", &mlir::rock::AmdArchInfo::maxNumXCC);
```
umangyadav, line 48: HIP APIs would soon allow look up for maxNumXCC. it's been implemented now in HIP APIs. Do we have tracking issue for that when we move to rocm-7.0 ?

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -74,24 +83,207 @@ static constexpr AmdArchInfo
...
+
+GemmFeatures &operator|=(GemmFeatures &lhs, const GemmFeatures &rhs) {
+  lhs = lhs | rhs;
+  return lhs;
+}
+
+GemmFeatures &operator&=(GemmFeatures &lhs, const GemmFeatures &rhs) {
+  lhs = lhs & rhs;
+  return lhs;
+}
```
umangyadav: GemmFeatures already has `bitEnumClear` and `bitEnumSet` methods that you can use

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -74,24 +83,207 @@ static constexpr AmdArchInfo
...
+    if (st.hasAtomicBufferGlobalPkAddF16Insts() ||
+        st.hasAtomicBufferGlobalPkAddF16NoRtnInsts()) {
+      features |= GemmFeatures::atomic_add_f16;
+    } else {
+      features &= ~GemmFeatures::atomic_add_f16;
+    }
+    if (st.hasAtomicBufferPkAddBF16Inst()) {
+      features |= GemmFeatures::atomic_add_bf16;
+    } else {
+      features &= ~GemmFeatures::atomic_add_bf16;
```
umangyadav: ~I am not sure if you need to explictly remove those. It would generate incorrect encoding. 
You need to remove only if they existed initially~

mlir/unittests/Dialect/Rock/AmdArchDbTests.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+#include "gtest/gtest.h"
+
+#include "hip/hip_runtime_api.h"
+
+#include <numeric>
+
+using namespace mlir::rock;
+
+#define SOFT_CHECK_EQ(val1, val2)                                              \
```
umangyadav: Looks like this unused.

mlir/lib/Dialect/Rock/utility/Bindings/AmdArchDbBindings.cpp
mirza-halilcevic: I created one https://github.com/ROCm/rocMLIR-internal/issues/1886

umangyadav:
@mirza-halilcevic  CI is failing, Can you fix those failures and update the branch ?

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
dhernandez0, line 1: please can you change cdna35Info to cdna4Info? it seems it's cdna4 https://www.amd.com/content/dam/amd/en/documents/instinct-tech-docs/instruction-set-architectures/amd-instinct-cdna4-instruction-set-architecture.pdf @umangyadav

umangyadav: Yes let's change it to cdna4 then

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
```diff
@@ -136,29 +136,28 @@ std::tuple<StringRef, unsigned> parseArchString(StringRef arch) {
...
-AmdArchInfo fetchNativeArchInfo(const hipDeviceProp_t &prop) {
+struct agent_info_t {
+  // Input fields:
+  //   The ID of the GPU device that we are looking for.
+  unsigned deviceId;
+  //   The ID of the current GPU HSA agent (weird that HSA does not have a field
+  //   for this).
+  int num_cpus;
+  // Output fields:
+  uint32_t simds_per_cu;
```
umangyadav: nit: use camelcase

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
```diff
@@ -168,74 +167,61 @@ AmdArchInfo fetchNativeArchInfo(const hipDeviceProp_t &prop) {
...
-                      ret.hasFp8ConversionInstrs, false);
-      checkAndSetInfo("(LLVM) hasOcpFp8ConversionInstrs",
-                      ret.hasOcpFp8ConversionInstrs, false);
+  checkAndSetInfo("(HSA) numEUPerCU", ret.numEUPerCU, agent_info.simds_per_cu);
+  checkAndSetInfo("(HSA) maxWavesPerEU", ret.maxWavesPerEU,
+                  agent_info.max_waves_per_cu / agent_info.simds_per_cu);
+  // checkAndSetInfo("(HSA) totalSGPRPerEU", ret.totalSGPRPerEU, ???);
+  // checkAndSetInfo("(HSA) totalVGPRPerEU", ret.totalVGPRPerEU, ???);
+  checkAndSetInfo("(HSA) waveSize", ret.waveSize, agent_info.wavefront_size);
+  // checkAndSetInfo("(HSA) defaultFeatures", ret.defaultFeatures, ???);
```
umangyadav: I dont' expect HSA to have target features level details as it is runtimeAPI

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
```diff
@@ -136,29 +136,28 @@ std::tuple<StringRef, unsigned> parseArchString(StringRef arch) {
...
+  //   for this).
+  int num_cpus;
+  // Output fields:
+  uint32_t simds_per_cu;
+  uint32_t wavefront_size;
+  uint32_t max_waves_per_cu;
+};
+
+AmdArchInfo fetchNativeArchInfo(const hipDeviceProp_t &prop,
+                                agent_info_t &agent_info) {
```
dhernandez0: camelCase here as well

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
```diff
@@ -136,29 +136,28 @@ std::tuple<StringRef, unsigned> parseArchString(StringRef arch) {
...
+      return err;                                                              \
+    }                                                                          \
   }
 
-  return std::unique_ptr<GCNTargetMachine>(static_cast<GCNTargetMachine *>(
-      target->createTargetMachine(triple, chip, featureString, {}, {}, {})));
-}
-
-AmdArchInfo fetchNativeArchInfo(const hipDeviceProp_t &prop) {
+struct agent_info_t {
```
dhernandez0: AgentInfo (see other examples in the codebase such as struct ConvolutionDims)

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
```diff
@@ -136,29 +136,28 @@ std::tuple<StringRef, unsigned> parseArchString(StringRef arch) {
...
+  // Input fields:
+  //   The ID of the GPU device that we are looking for.
+  unsigned deviceId;
+  //   The ID of the current GPU HSA agent (weird that HSA does not have a field
+  //   for this).
+  int num_cpus;
```
dhernandez0: this is the ID of the gpu? why is it call num_cpus?

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
```diff
@@ -168,74 +167,61 @@ AmdArchInfo fetchNativeArchInfo(const hipDeviceProp_t &prop) {
...
-    } else {
-      checkAndSetInfo("(LLVM) hasFp8ConversionInstrs",
-                      ret.hasFp8ConversionInstrs, false);
-      checkAndSetInfo("(LLVM) hasOcpFp8ConversionInstrs",
-                      ret.hasOcpFp8ConversionInstrs, false);
+  checkAndSetInfo("(HSA) numEUPerCU", ret.numEUPerCU, agent_info.simds_per_cu);
+  checkAndSetInfo("(HSA) maxWavesPerEU", ret.maxWavesPerEU,
+                  agent_info.max_waves_per_cu / agent_info.simds_per_cu);
+  // checkAndSetInfo("(HSA) totalSGPRPerEU", ret.totalSGPRPerEU, ???);
+  // checkAndSetInfo("(HSA) totalVGPRPerEU", ret.totalVGPRPerEU, ???);
```
dhernandez0: The point of this PR was to get all the info from a reliable source. If HSA doesn't have all the relevant info, why not just expose the LLVM internal info and use it?

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
pabloantoniom: Wrong comment 😅

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
dhernandez0, line 1: I thought we were going to change cdna35Info -> cdna4Info, why did you revert that change?

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
```diff
@@ -77,24 +85,224 @@ static constexpr AmdArchInfo
...
+  // checkAndSetInfo("(???) totalSGPRPerEU", ret.totalSGPRPerEU, ???);
+  // checkAndSetInfo("(???) totalVGPRPerEU", ret.totalVGPRPerEU, ???);
+  // checkAndSetInfo("(???) defaultFeatures", ret.defaultFeatures, ???);
+  // checkAndSetInfo("(???) hasOcpFp8ConversionInstrs",
+  // ret.hasOcpFp8ConversionInstrs, ???); checkAndSetInfo("(???)
+  // hasFp8ConversionInstrs", ret.hasFp8ConversionInstrs, ???);
+  return ret;
+}
+
+static hsa_status_t AcquireAgentInfo(hsa_agent_t agent, void *data) {
```
dhernandez0: why not pass AgentInfo * instead of void*?

dhernandez0: also, names of functions don't start with capital letters

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
```diff
@@ -77,24 +85,224 @@ static constexpr AmdArchInfo
...
+          &agent_i->simdsPerCU);
+      RET_IF_HSA_ERR(err);
+
+      err = hsa_agent_get_info(
+          agent, (hsa_agent_info_t)HSA_AMD_AGENT_INFO_MAX_WAVES_PER_CU,
+          &agent_i->maxWavesPerCU);
+      RET_IF_HSA_ERR(err);
+    }
+  } else {
+    agent_i->numCpus++;
```
dhernandez0, line 237: looks like numCpus could be a local variable of this function, do we need it in AgentInfo?

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
```diff
@@ -77,24 +85,224 @@ static constexpr AmdArchInfo
...
+  //
+  // References:
+  // -
+  // https://rocm.docs.amd.com/projects/HIP/en/docs-6.0.2/user_guide/hip_rtc.html#cu-mode-vs-wgp-mode
+  // -
+  // https://www.amd.com/content/dam/amd/en/documents/radeon-tech-docs/instruction-set-architectures/rdna3-shader-instruction-set-architecture-feb-2023_0.pdf
+  // -
+  // https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/AMDGPU/Utils/AMDGPUBaseInfo.cpp
+
+  if (prop->warpSize == 32) {
```
dhernandez0, line 268: I guess there's no flag for WGP mode?

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
```diff
@@ -170,74 +169,61 @@ AmdArchInfo fetchNativeArchInfo(const hipDeviceProp_t &prop) {
...
-      }
-    } else {
-      checkAndSetInfo("(LLVM) hasFp8ConversionInstrs",
-                      ret.hasFp8ConversionInstrs, false);
-      checkAndSetInfo("(LLVM) hasOcpFp8ConversionInstrs",
-                      ret.hasOcpFp8ConversionInstrs, false);
+  checkAndSetInfo("(HSA) numEUPerCU", ret.numEUPerCU, agent_info.simds_per_cu);
+  checkAndSetInfo("(HSA) maxWavesPerEU", ret.maxWavesPerEU,
+                  agent_info.max_waves_per_cu / agent_info.simds_per_cu);
+  // checkAndSetInfo("(HSA) totalSGPRPerEU", ret.totalSGPRPerEU, ???);
```
dhernandez0: can we add a TODO explaining what params are missing instead of commeted code?

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
```diff
@@ -170,74 +169,61 @@ AmdArchInfo fetchNativeArchInfo(const hipDeviceProp_t &prop) {
...
-
-    checkAndSetInfo("(LLVM) numEUPerCU", ret.numEUPerCU, st.getEUsPerCU());
-    checkAndSetInfo("(LLVM) maxWavesPerEU", ret.maxWavesPerEU,
-                    st.getMaxWavesPerEU());
-    checkAndSetInfo("(LLVM) totalSGPRPerEU", ret.totalSGPRPerEU,
-                    st.getTotalNumSGPRs());
-    checkAndSetInfo("(LLVM) totalVGPRPerEU", ret.totalVGPRPerEU,
-                    st.getTotalNumVGPRs());
-    checkAndSetInfo("(LLVM) waveSize", ret.waveSize, st.getWavefrontSize());
-    checkAndSetInfo("(LLVM) totalSharedMemPerCU", ret.totalSharedMemPerCU,
```
dhernandez0: we are missing this one as well

pabloantoniom: In HIP it's called maxSharedMemoryPerMultiProcessor

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
pabloantoniom: I searched in HIP and HSA but I couldn't find it. Umang said it's actually a compile time thing so I'm not sure how to check it otherwise

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
pabloantoniom: Hope this is more clear with the comment that I added

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
pabloantoniom: Added a comment on top of the function to explain why it's void*

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
pabloantoniom: Ah, I didn't have that background, change reverted

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
```diff
@@ -77,32 +86,243 @@ static constexpr AmdArchInfo
              /*maxSharedMemPerWG*/ 65536, /*numEUPerCU=*/4, /*minNumCU=*/36,
              /*hasFp8ConversionInstrs=*/false,
              /*hasOcpFp8ConversionInstrs=*/false, /*maxNumXCC=*/1),
-    gfx11Info(GemmFeatures::dot | GemmFeatures::atomic_add |
+    rdna3Info(GemmFeatures::dot | GemmFeatures::atomic_add |
                   GemmFeatures::atomic_fmax_f32 | GemmFeatures::wmma,
-              /*waveSize=*/32, /*maxWavesPerEU*/ 20, /*totalSGPRPerEU*/ 512,
+              /*waveSize=*/32, /*maxWavesPerEU*/ 16, /*totalSGPRPerEU*/ 800,
```
dhernandez0, line 94: nit: remove space

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
```diff
@@ -33,27 +42,27 @@ static constexpr AmdArchInfo
...
+              /*waveSize=*/64, /*maxWavesPerEU*/ 8, /*totalSGPRPerEU*/ 800,
               /*totalVGPRPerEU*/ 512, /*totalSharedMemPerCU*/ 65536,
               /*maxSharedMemPerWG*/ 65536, /*numEUPerCU=*/4, /*minNumCU=*/104,
               /*hasFp8ConversionInstrs=*/false,
               /*hasOcpFp8ConversionInstrs=*/false, /*maxNumXCC=*/1),
     cdna3Info(GemmFeatures::mfma | GemmFeatures::dot |
                   GemmFeatures::atomic_add | GemmFeatures::atomic_add_f16 |
                   GemmFeatures::direct_to_lds_32b,
-              /*waveSize=*/64, /*maxWavesPerEU*/ 10, /*totalSGPRPerEU*/ 512,
+              /*waveSize=*/64, /*maxWavesPerEU*/ 8, /*totalSGPRPerEU*/ 800,
```
dhernandez0, line 63: nit: remove spaces

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
```diff
@@ -33,27 +42,27 @@ static constexpr AmdArchInfo
                /*hasOcpFp8ConversionInstrs=*/false, /*maxNumXCC=*/1),
     cdnaInfo(GemmFeatures::mfma | GemmFeatures::dot | GemmFeatures::atomic_add |
                  GemmFeatures::atomic_add_f16,
-             /*waveSize=*/64, /*maxWavesPerEU*/ 8, /*totalSGPRPerEU*/ 512,
-             /*totalVGPRPerEU*/ 512, /*totalSharedMemPerCU*/ 65536,
+             /*waveSize=*/64, /*maxWavesPerEU*/ 10, /*totalSGPRPerEU*/ 800,
```
dhernandez0, line 48: nit: same here, remove spaces

mlir/unittests/Dialect/Rock/AmdArchDbTests.cpp
dhernandez0, line 1: given that the following params are hardcoded (totalSGPRPerEU, totalVGPRPerEU, defaultFeatures, hasOcpFp8ConversionInstrs) we could have a test for gfx950 and/or navi4 at least? where we hardcode the expected values.

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
pabloantoniom: Why is this not handled by clang-format?

dhernandez0: I don't know, let's open a ticket

pabloantoniom: I have opened the ticket: https://github.com/ROCm/rocMLIR-internal/issues/2011

It turns out that this wrongly formatted issue happens in a lot of cases in the file, not only with `waveSize`, so I prefer to leave this as it is for now to avoid polluting the PR

mlir/unittests/Dialect/Rock/AmdArchDbTests.cpp
pabloantoniom: Not sure if I understand, you mean something like this?

```
checkAndSetInfo("totalSGPRPerEU", ret.totalSGPRPerEU, 800);
```

dhernandez0: yes, something like that

mlir/lib/Dialect/Rock/utility/CMakeLists.txt
mirza-halilcevic, line 1: Since AmdArchDb is no longer part of MLIRRockUtility, we can revert the changes to this file.

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
```diff
@@ -8,12 +8,25 @@
...
 #include "mlir/Dialect/Rock/IR/RockTypes.h"
 #include "mlir/IR/TypeUtilities.h"
 
 #include "llvm/ADT/StringSwitch.h"
+#include "llvm/MC/TargetRegistry.h"
+#include "llvm/Support/Debug.h"
 #include "llvm/Support/ErrorHandling.h"
+#include "llvm/Support/TargetSelect.h"
+
+#include "hip/hip_runtime_api.h"
```
apwojcik: This #include will not compile on Windows. We use MSVC (Visual Studio C++ compiler), which is unaware of HIP and device code. It should be limited to Linux for now.

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
```diff
@@ -77,32 +90,252 @@ static constexpr AmdArchInfo
...
+
+AmdArchInfo nativeArchInfo(unsigned deviceId = 0) {
+  static std::mutex m;
+  static std::unordered_map<std::string, AmdArchInfo> cache;
+
+  LLVM_DEBUG(llvm::dbgs() << "Retrieving native arch info for device "
+                          << deviceId << "...\n");
+
+  hipDeviceProp_t prop;
+  if (auto err = hipGetDeviceProperties(&prop, deviceId); err != hipSuccess) {
```
apwojcik, line 283: Same here, the MSVC will not compile that code.

mlir/lib/Dialect/Rock/IR/CMakeLists.txt
```diff
@@ -46,3 +46,21 @@ target_link_libraries(MLIRRockOps
   MLIRSupport
   MLIRLinalgTransformOps
   )
+
+# From: https://github.com/ROCm/rocMLIR/blob/b6c726a324d2807494f770ed915af0e54cc6cb3f/external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt#L414
+find_package(hip REQUIRED PATHS ${ROCM_PATH})
```
apwojcik: Please make it Linux only. Due to the compiler we use, the code will not compile.

mlir/lib/Dialect/Rock/IR/CMakeLists.txt
```diff
@@ -46,3 +46,21 @@ target_link_libraries(MLIRRockOps
...
+# We do need both of these. One is for linking, to get the HIP
+# functions.  The other is for compiling, to get -D__HIP_PLATFORM_AMD__
+target_link_libraries(MLIRRockOps PUBLIC
```
apwojcik: Same here

mlir/lib/Dialect/Rock/IR/CMakeLists.txt
```diff
@@ -46,3 +46,21 @@ target_link_libraries(MLIRRockOps
...
+  hip::host hip::amdhip64
+)
+target_link_libraries(obj.MLIRRockOps PUBLIC
+  hip::host hip::amdhip64
```
apwojcik: and here

mlir/lib/Dialect/Rock/utility/Bindings/AmdArchDbBindings.cpp
```diff
@@ -0,0 +1,53 @@
+//===- AmdArchDbBindings.cpp - Python bindings for the AMD arch database
+//------------------===//
+//
+// Part of the MLIR Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include <pybind11/pybind11.h>
```
apwojcik, line 10: We do not have pybind11 as a dependency on Windows. This will not compile either.

mlir/lib/Dialect/Rock/utility/Bindings/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+find_package(Python3 COMPONENTS Interpreter Development REQUIRED)
```
apwojcik: Same here; we do not require pybind11 as a dependency on Windows. This will not compile.

apwojcik (changes requested):

mlir/lib/Dialect/Rock/IR/AmdArchDb.cpp
```diff
@@ -77,32 +90,252 @@ static constexpr AmdArchInfo
...
+  //   The ID of the GPU device that we are looking for.
+  unsigned deviceId;
+  //   Used in acquireAgentInfo, to compute GPU internal IDs.
+  int numCpus;
+  // Output fields:
+  uint32_t simdsPerCU;
+  uint32_t maxWavesPerCU;
+};
+
+AmdArchInfo fetchNativeArchInfo(const hipDeviceProp_t &prop,
```
apwojcik, line 160: Can we move that to the migraphx_gpu target instead? To read the required properties there and pass them to the rocMLIR instead of makeing HIP a dependency of rocMLIR?

Currently the HIP dependency compilcates a lot on Windows for rocMLIR.

mlir/lib/Dialect/Rock/utility/Bindings/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+    find_package(Python3 COMPONENTS Interpreter Development REQUIRED)
+
+    include(${LLVM_EXTERNAL_SRC_DIR}/mlir/cmake/modules/MLIRDetectPythonEnv.cmake)
+    mlir_detect_pybind11_install()
+
+    find_package(pybind11 CONFIG REQUIRED)
+
+    pybind11_add_module(amd_arch_db AmdArchDbBindings.cpp)
+    target_link_libraries(amd_arch_db PUBLIC MLIRRockUtility)
+endif()
```
dhernandez0: new line

mlir/unittests/Dialect/Rock/CMakeLists.txt
```diff
@@ -1,9 +1,17 @@
...
+add_rocmlir_unittest(MLIRRockUnitTests
+  ${ROCK_UNITTEST_SOURCES}
+)
+
+target_link_libraries(MLIRRockUnitTests
   PRIVATE
   MLIRRockOps
   MLIRRockUtility
-)
+)
```
dhernandez0: new line

apwojcik (approved):

PR #1793, opened 2025-03-26
Sort Dimensions based on Layout in case of input fusion 

umangyadav:
Fixes slowdown reported in https://github.com/ROCm/rocMLIR-internal/issues/1784


mlir/test/fusion/pr-e2e/conv-layouts/nhwc/input-fusion-conv-nhwc-e2e.mlir
```diff
@@ -0,0 +1,20 @@
+// RUN: rocmlir-gen --clone-harness -arch %arch -fut test %s | rocmlir-driver -kernel-pipeline migraphx | rocmlir-driver -host-pipeline migraphx,highlevel -targets %arch | rocmlir-gen --emit-tuning-key - | FileCheck %s  --check-prefixes=EMITKEY
+// RUN: rocmlir-gen --clone-harness -arch %arch -fut test %s | rocmlir-driver -kernel-pipeline migraphx | rocmlir-driver -host-pipeline migraphx,highlevel -targets %arch | rocmlir-gen -ph -relDiff_threshold 0.09 -absDiff_threshold 1 -RMS_threshold 0.05 -verifier clone -fut test_wrapper - | rocmlir-driver -host-pipeline mhal -kernel-pipeline full | xmir-runner --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_float16_utils%shlibext,%linalg_test_lib_dir/libmlir_c_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_async_runtime%shlibext --entry-point-result=void | FileCheck %s --check-prefix=CLONE
```
umangyadav, line 2: I had to increase threshold for some reason for it to pass. Investigating if this is related to https://github.com/ROCm/rocMLIR-internal/issues/1788 or not.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -72,18 +73,35 @@ sortByMemoryLayout(Value tensor, const Container &layout, PatternRewriter &b) {
...
+         std::distance(source.getUsers().begin(), source.getUsers().end()) ==
+             2) {
+    for (auto *users : source.getUsers()) {
+      if (auto laGenericOp = dyn_cast<linalg::GenericOp>(users)) {
+        for (auto idxMap : laGenericOp.getIndexingMapsArray())
+          if (!idxMap.isIdentity())
+            return std::make_tuple(tensor, layout, SmallVector<uint32_t>{});
+        source = laGenericOp.getInputs().front();
+        std::tie(source, transforms, std::ignore) =
+            rock::untransform(b, source, transforms);
```
dhernandez0, line 89: I think the goal here is to get the "transforms" variable of the input tensors (from input fusion)? I think they could have different layouts than regular inputs, can't we iterate over input tensors from fusions and call sortByMemoryLayout() for each of them?

mlir/test/fusion/pr-e2e/conv-layouts/nhwc/input-fusion-conv-nhwc-e2e.mlir
```diff
@@ -0,0 +1,20 @@
...
+// ALLOW_RETRIES: 2
+// CLONE: [1 1 1]
+// EMITKEY: convfp16 -F 1 -f N01GC -I 01NGC -O N01GC -n 1 -c 128 -H 80 -W 80 -k 128 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -g 1
+// VECTORIZATION: aVectorDim: GemmDimension::K
+// VECTORIZATION-NEXT: aVectorLen: 8
+// VECTORIZATION: bVectorDim: GemmDimension::K
+// VECTORIZATION-NEXT: bVectorLen: 8
+
+module {
+func.func @test(%arg0: !migraphx.shaped<1x256x80x80xf16, 1638400x1x20480x256>, %arg1: !migraphx.shaped<128x128x3x3xf16, 1152x1x384x128>, %arg2: !migraphx.shaped<128xf16, 1>) -> !migraphx.shaped<1x128x80x80xf16, 819200x1x10240x128> {
```
dhernandez0: remove arg2 if it's not used

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -72,18 +73,35 @@ sortByMemoryLayout(Value tensor, const Container &layout, PatternRewriter &b) {
   ArrayAttr transforms;
   Value source;
   std::tie(source, transforms, std::ignore) = rock::untransform(b, tensor);
+  // handle input fusion, traverse past linalg generic op
+  while (source.getDefiningOp<memref::AllocOp>() &&
+         // allocOp will have two users, one rock.transform and other
+         // linalg.generic op
+         std::distance(source.getUsers().begin(), source.getUsers().end()) ==
```
dhernandez0: this seems very specific to the current ticket? ideally we can handle any number of users

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -72,18 +73,35 @@ sortByMemoryLayout(Value tensor, const Container &layout, PatternRewriter &b) {
...
+             2) {
+    for (auto *users : source.getUsers()) {
+      if (auto laGenericOp = dyn_cast<linalg::GenericOp>(users)) {
```
dhernandez0: can we do something like traceGemmOutputToArgs() (loweringUtils.cpp). traceGemmInputToArgs? that would be more flexible than this

dhernandez0: After thinking a bit more about this. If we trace the gemm input to args, one input could come from multiple args (not the case in the example in your tests). So, that would need another fix, to call sortByMemoryLayout() on each of them independently. But we can fix that in another ticket I think

mlir/test/fusion/pr-e2e/conv-layouts/nhwc/input-fusion-conv-nhwc-e2e.mlir
umangyadav: Removed

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: > can we do something like traceGemmOutputToArgs() (loweringUtils.cpp). traceGemmInputToArgs? that would be more flexible than this 

Done. 

>  one input could come from multiple args (not the case in the example in your tests). So, that would need another fix, 

Yes 

For now, i collect all the blockArgs but bail out if they have different sequences of rock.transforms.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: Changed to be more generic

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -66,24 +71,101 @@ FailureOr<Container> reorderArrayAttr(Container inputArray,
...
+  transformAttrsMap[source] = transformsOnSource;
+  if (isa<BlockArgument>(source)) {
+    blockArgs.push_back(source);
+    return success();
+  }
+  FailureOr<memref::AllocOp> allocOp = mlir::rock::findMemrefAlloc(source);
+  if (failed(allocOp)) {
+    return failure();
+  }
+  BufferDependencyAnalysis deps(allocOp.value());
```
dhernandez0: let's do this outside of the function. It's called recursively and we'll have to do all the dependency analysis multiple times.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -66,24 +71,101 @@ FailureOr<Container> reorderArrayAttr(Container inputArray,
...
+  bool hasSuccess = false;
+  for (OpOperand *allocWriteOperand : allocOpWriters.value()) {
+    auto writerOp =
+        dyn_cast<MemoryEffectOpInterface>(allocWriteOperand->getOwner());
+    if (!writerOp)
+      continue;
+    SmallVector<MemoryEffects::EffectInstance> effects;
+    writerOp.getEffects(effects);
+    for (const MemoryEffects::EffectInstance &effect : effects) {
+      OpOperand *writerOpOperand = effect.getEffectValue<OpOperand *>();
```
dhernandez0, line 130: I think this should be readerOpOperand, because we test "isa<MemoryEffects::Read>(effect.getEffect())"

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -66,24 +71,101 @@ FailureOr<Container> reorderArrayAttr(Container inputArray,
...
-
-  if (transforms.empty())
+  // trace input tensor to blockArgument first and do necessary error checking
+  llvm::DenseMap<Value, SmallVector<Attribute>> transformAttrsMap;
+  llvm::SmallVector<Value, 4> blockArgs;
+  if (failed(traceGemmArgToBlockArg(tensor, b, transformAttrsMap, blockArgs))) {
     return std::make_tuple(tensor, layout, SmallVector<uint32_t>{});
-
+  }
+  assert(!blockArgs.empty());
```
dhernandez0, line 160: is it possible that all of the gemm/attention/conv inputs are literals?

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -66,24 +71,101 @@ FailureOr<Container> reorderArrayAttr(Container inputArray,
...
+  SmallVector<Attribute> transformsList;
+  for (const auto blockArg : blockArgs) {
+    // make sure all the blockArgs have been mapped to some transform sequence
+    // or empty transform sequence
+    if (not transformAttrsMap.contains(blockArg)) {
```
dhernandez0: use ! instead of not (to follow llvm coding standard)

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -66,24 +71,101 @@ FailureOr<Container> reorderArrayAttr(Container inputArray,
   return reorderedElements;
 }
 
+//  traces input arguments of the GEMM operation back to blockArguments. It
+//  records sequence of rock.transforms between gemm argument to blockArgument
+//  if there is any. It is possible that single gemm arg is mapped to multiple
+//  blockArguments. BlockArguments are recorded in `blockArgs` and series of
+//  rock.TransformAttr sequences for each `blockArgs` is recorded in
+//  transformAttrsMap.
+static LogicalResult traceGemmArgToBlockArg(
```
dhernandez0: nit: GemmArg -> GemmInput?
nit: move to loweringUtils.cpp as this can be useful in the future for other parts of the code.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -66,24 +71,101 @@ FailureOr<Container> reorderArrayAttr(Container inputArray,
...
+    Value inputArg, PatternRewriter &b,
+    llvm::DenseMap<Value, SmallVector<Attribute>> &transformAttrsMap,
+    llvm::SmallVector<Value, 4> &blockArgs) {
```
dhernandez0: this should be BlockArgument, not Value

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -66,24 +71,101 @@ FailureOr<Container> reorderArrayAttr(Container inputArray,
...
+  ArrayAttr transforms;
+  std::tie(source, transforms, std::ignore) =
+      rock::untransform(b, inputArg, transformAttrsMap[inputArg]);
+  // avoid adding duplicates
+  if (llvm::is_contained(blockArgs, source)) {
+    return success();
+  }
+  SmallVector<Attribute> transformsOnSource{};
+  transformsOnSource.append(transforms.begin(), transforms.end());
+  transformAttrsMap[source] = transformsOnSource;
```
dhernandez0: looks like we copy the smallvector twice, can't we copy directly to transformAttrsMap[source]?

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -66,24 +71,101 @@ FailureOr<Container> reorderArrayAttr(Container inputArray,
...
+static LogicalResult traceGemmArgToBlockArg(
+    Value inputArg, PatternRewriter &b,
+    llvm::DenseMap<Value, SmallVector<Attribute>> &transformAttrsMap,
+    llvm::SmallVector<Value, 4> &blockArgs) {
+  Value source;
+  ArrayAttr transforms;
+  std::tie(source, transforms, std::ignore) =
+      rock::untransform(b, inputArg, transformAttrsMap[inputArg]);
+  // avoid adding duplicates
+  if (llvm::is_contained(blockArgs, source)) {
```
dhernandez0: nit: use SmallSetVector for blockArgs

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
dhernandez0: nit: why it's init to length of 4?

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -66,24 +71,101 @@ FailureOr<Container> reorderArrayAttr(Container inputArray,
...
+      continue;
+    SmallVector<MemoryEffects::EffectInstance> effects;
+    writerOp.getEffects(effects);
+    for (const MemoryEffects::EffectInstance &effect : effects) {
+      OpOperand *writerOpOperand = effect.getEffectValue<OpOperand *>();
+      // test that same buffer is not being read and written to
+      if (writerOpOperand && isa<MemoryEffects::Read>(effect.getEffect()) &&
+          writerOpOperand != allocWriteOperand) {
+        Value writerOpOperandValue = writerOpOperand->get();
+        transformAttrsMap[writerOpOperandValue] = transformsOnSource;
```
dhernandez0: I don't understand why we need this line

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
dhernandez0: use FailureOr<SmallVector<BlockArgument>> or FailureOr<SmallSetVector<BlockArgument>> as return type

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -66,24 +71,101 @@ FailureOr<Container> reorderArrayAttr(Container inputArray,
...
+    }
+  }
+  if (transformsList.empty()) {
+    return std::make_tuple(tensor, layout, SmallVector<uint32_t>{});
+  }
+  ArrayAttr transforms = b.getArrayAttr(transformsList);
   rock::TransformMapAttr firstCoordTransform =
-      cast<rock::TransformMapAttr>(transforms[0]);
+      dyn_cast<rock::TransformMapAttr>(transformsList[0]);
+  firstCoordTransform.dump();
```
dhernandez0: I guess this was for debugging? remove or make sure it's printed to llvm::dbgs()

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -66,24 +71,101 @@ FailureOr<Container> reorderArrayAttr(Container inputArray,
...
+      return std::make_tuple(tensor, layout, SmallVector<uint32_t>{});
+    }
+  }
+  if (transformsList.empty()) {
+    return std::make_tuple(tensor, layout, SmallVector<uint32_t>{});
+  }
+  ArrayAttr transforms = b.getArrayAttr(transformsList);
   rock::TransformMapAttr firstCoordTransform =
-      cast<rock::TransformMapAttr>(transforms[0]);
+      dyn_cast<rock::TransformMapAttr>(transformsList[0]);
```
dhernandez0: we know this has to be TransformMapAttr, right? if that's not the case either we fail here or we get a null pointer and fail later, I'd prefer to use cast to fail here

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -66,24 +71,101 @@ FailureOr<Container> reorderArrayAttr(Container inputArray,
...
+    if (not transformAttrsMap.contains(blockArg)) {
+      return std::make_tuple(tensor, layout, SmallVector<uint32_t>{});
+    }
+    if (transformsList.empty()) {
+      transformsList = transformAttrsMap[blockArg];
+    } else if (transformsList != transformAttrsMap[blockArg]) {
+      // Currently we do not handle case where some block arg goes through
+      // different sequence of transforms. All blockArgs must have same
+      // transforms for now.
+      return std::make_tuple(tensor, layout, SmallVector<uint32_t>{});
```
dhernandez0, line 174: nit: I think we'll generate different transforms for each blockargs always, even if they do the same. It might be easier to just check if transformAttrsMap contains more than one blockarg.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -66,24 +71,101 @@ FailureOr<Container> reorderArrayAttr(Container inputArray,
...
-  ArrayAttr transforms;
-  Value source;
-  std::tie(source, transforms, std::ignore) = rock::untransform(b, tensor);
-
-  if (transforms.empty())
+  // trace input tensor to blockArgument first and do necessary error checking
+  llvm::DenseMap<Value, SmallVector<Attribute>> transformAttrsMap;
+  llvm::SmallVector<Value, 4> blockArgs;
+  if (failed(traceGemmArgToBlockArg(tensor, b, transformAttrsMap, blockArgs))) {
     return std::make_tuple(tensor, layout, SmallVector<uint32_t>{});
```
dhernandez0, line 158: there's something I don't understand. We are storing all transforms of all sources as we go from the gemm input to its blockargs. However, below we just use the transforms of the blockargs.

I think getLowerSubDimensions() needs all the transforms that have happened from the gemm input to the blockarg to give us the correct subdims. So, for each blockarg we should have transforms(gemm input to alloc out of linalg) + transforms(linalg input to blockarg) (example where you only have one linalg). And that would be the transformAttrs input of getLowerSubDimensions(), I think currently we are only passing the last transforms? "transforms(linalg input to blockarg)", right?

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: I only want to perform that for `memref::allocOp`. BufferDependencyAnalysis takes `operation` as parameter. Therefore if i compute it outside this function then i would need to collect all `memref::allocOp` and then create a map.  Which may not be necessary.

I can instead maintain a  map of `<allocOp, BufferDeps>` on the fly

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: Reasoning behind naming was that we traversing backwards. 
I am searching for writers of the AllocOp in any invocation. And i name it `writerOp` hence. 

```
  for (OpOperand *allocWriteOperand : allocOpWriters.value()) {
    auto writerOp =
        dyn_cast<MemoryEffectOpInterface>(allocWriteOperand->getOwner());
```
Operands of the writerOp are named as `writerOpOperand`. Operands themselves may be read or written to.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
dhernandez0: There's no need to call dependency analysis for all allocops and create a map. You can just do this once outside the function: 
```c++
BufferDependencyAnalysis deps =
      BufferDependencyAnalysis(func.getOperation());
```

And then, call:
```
std::optional<llvm::SmallVector<OpOperand *>> writerOperands =
      deps.getWriters(buffer);
```

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: > use FailureOr<SmallVector> or FailureOr<SmallSetVector> as return type 

It is possible that input is being mapped to multiple block arguments and each blockArg is found at different times in recursions. I find it better to keep `blockArgs` Ref instead of return type.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: > nit: why it's init to length of 4? 

Randomly. 2 is better number probably.

umangyadav: > this should be BlockArgument, not Value 

That shouldn't make any difference, we are interested in sequence of rock.transforms that lead to blockArgs

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
dhernandez0: ok, I see your point, I just find it confusing to understand like this because here we filter out and only process operands of writerOp that are read from.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
dhernandez0: > It is possible that input is being mapped to multiple block arguments and each blockArg is found at different times in recursions

sorry, I don't understand why that matters to the return type?

> I find it better to keep blockArgs Ref instead of return type.

blockArgs and the transforms are what you want to return from the function. But you might fail in some cases, so FailureOr<pair<blockArgs, transforms>> would make sense. I think that's the coding style that we use in the codebase generally.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
dhernandez0: > > nit: why it's init to length of 4?
> 
> Randomly. 2 is better number probably.

yes, my point is that if there's no clear init lenght, you can just use llvm::SmallVector< Value >.

> That shouldn't make any difference, we are interested in sequence of rock.transforms that lead to blockArgs

In the blockArgs list you only store BlockArgument, this would use the typing checks to make sure you don't add a Value that is not BlockArgument by mistake. Also, the user of the function can be sure that there are only BlockArguments, otherwise, there might be potential non-BlockArgument in the future due to bugs.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
dhernandez0: btw, this is really a "nit", I'm happy with whatever you prefer.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: That's exactly what i was looking for thanks. 

SetVector actually has larger memory footprint it keeps both set and vector underneath to guarantee order of elements.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: It should maintains  and continue sequence of rock.transforms when doing recursion.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: >  It might be easier to just check if transformAttrsMap contains more than one blockarg.

TransformAttrsMap is map from `blockArg -> SequenceOfTransforms`

There could be multiple blockArgs. IMO, It is better to check if all of them have same transform sequence or not. instead of just checking if there is only one blockarg.

dhernandez0: what I'm saying is that they will never have the same sequence of transforms. Their transforms could be doing exactly the same, but they will be different.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -77,32 +78,32 @@ FailureOr<Container> reorderArrayAttr(Container inputArray,
...
-    blockArgs.push_back(source);
+    blockArgs.insert(source);
     return success();
   }
   FailureOr<memref::AllocOp> allocOp = mlir::rock::findMemrefAlloc(source);
   if (failed(allocOp)) {
     return failure();
   }
-  BufferDependencyAnalysis deps(allocOp.value());
+  if (not deps.contains(allocOp.value())) {
```
dhernandez0: not -> !

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -77,32 +78,32 @@ FailureOr<Container> reorderArrayAttr(Container inputArray,
 //  blockArguments. BlockArguments are recorded in `blockArgs` and series of
 //  rock.TransformAttr sequences for each `blockArgs` is recorded in
 //  transformAttrsMap.
-static LogicalResult traceGemmArgToBlockArg(
+static LogicalResult traceGemmInputToBlockArgs(
     Value inputArg, PatternRewriter &b,
     llvm::DenseMap<Value, SmallVector<Attribute>> &transformAttrsMap,
-    llvm::SmallVector<Value, 4> &blockArgs) {
+    llvm::SmallSetVector<Value, 2> &blockArgs,
+    llvm::DenseMap<memref::AllocOp, BufferDependencyAnalysis> &deps) {
```
dhernandez0: you can do BufferDependencyAnalysis& deps, then call deps.getWriters(allocOp) inside the function

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: `transforms` here is of type `ArrayAttr`
v/s
what is require is `SmallVector<Attribute>` 

I need `transformsOnSource` later as well. I need a copy.

`        transformAttrsMap[writerOpOperandValue] = transformsOnSource;`

umangyadav: Made some changes.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: Changed it to `cast`

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: Removed.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: I don't follow this, Can you explain ?

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: Done

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
dhernandez0: ins't that done in the recurrent call to traceGemmArgToBlockArg?

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: > "transforms(linalg input to blockarg)", right? 

No, 

that's exactly why i pass on transforms upto gemmArg to allocOp to operands of LinAlg here

 https://github.com/ROCm/rocMLIR/pull/1793#discussion_r2016008758

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
dhernandez0: you can see other exmplaes in the codebase where BufferDependencyAnalysis is used.
you call BufferDependencyAnalysis on the whole module/func (outside the function):
 auto analysis = BufferDependencyAnalysis(func.getOperation());

then, internally just get writers or readers for the specific alloc you want to trace.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: It is possible in theory. Perhaps we should check if its literal or blockArg. 

I don't think migraphx passes weights as literal though right now.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
dhernandez0: are the transforms concatenated somewhere? I missed that

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
dhernandez0: yes, I was thinking of manual unit tests. We can just return:
return std::make_tuple(tensor, layout, SmallVector<uint32_t>{});
if it's empty

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: Done. Thanks.

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: Done.Thanks

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: I pass existing transforms here

`  std::tie(source, transforms, std::ignore) =
      rock::untransform(b, inputArg, transformAttrsMap[inputArg]);`

which would do concatenation

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: `writerOpOperandValue` will be `inputArg` for next recurrent call. I need to pass existing sequence of transforms in next recurrent call. 

  std::tie(source, transforms, std::ignore) =
      rock::untransform(b, inputArg, transformAttrsMap[inputArg]);

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
dhernandez0: so, the way I understand it is:
1. call traceGemmInputToBlockArgs with gemm input, gets the transforms from gemm input to its alloc, save in transformAttrsMap[alloc]=transforms
2. findMemrefAlloc is redudant because we already traced back to the alloc (I think)
3. find all operations that write to the alloc, and for each of them:
4. trace to inputs of operation and call traceGemmInputToBlockArgs() on them
5. also save another copy of transforms to transformAttrsMap[input of op]

Then, the caller iterates over blockArgs and gets the transforms of one blockArg (transformAttrsMap[blockArg]). Which would be the transforms from the first linalg input to blockarg. I don't see where the transforms in transformAttrsMap are concatenated to generate a transform that goes: gemm input -> linalg 1 -> linalg 2 -> ... -> blockArg

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
umangyadav: Added test with `linalg.fill`

dhernandez0 (approved):
LGTM. Let's open a ticket for the case when two or more linalg inputs have different transforms.

There's also this one to do: https://github.com/ROCm/rocMLIR-internal/issues/1751

mlir/lib/Dialect/Rock/Transforms/SortDimensionsMemoryLayout.cpp
```diff
@@ -65,25 +71,117 @@ FailureOr<Container> reorderArrayAttr(Container inputArray,
...
+  // For example,
+  // SeqExisting -> inputArgs --> Seq --> source
+  // transforms == SeqExisting + Seq
+  // transformAttrsMap[inputArg] = SeqExisting
+  // transformAttrsMap[Source] = SeqExisting + Seq
+  std::tie(source, transforms, std::ignore) =
+      rock::untransform(b, inputArg, transformAttrsMap[inputArg]);
+  // insert transform sequence on source into the map if it doesn't already
+  // exists. if it does then we've found a loop or case where multiple operators
+  // are writing to same `memref.alloc`
```
dhernandez0, line 101: nit: code might be easier to understand if we use transformAttrsMap.contains(source)

PR #1796, opened 2025-03-31
[DRAFT] Implement DPP Reduction in wavefront

stefankoncarevic:
This pull request addresses the implementation of wavefront reduction using DPP (Data Parallel Primitives) instructions. Currently, the implementation covers case that there aren't sufficient threads for a parallel reduction, allowing each individual thread to perform its own reduction.

This is the link to ticket: https://github.com/ROCm/rocMLIR-internal/issues/1524

**Current Architecture Support**
The code currently supports the MI200 and MI300 architectures. For RDNA and MI100 architectures, Need to adjust the conditions based on the architecture and select the appropriate intrinsic instructions supported by these specific architectures.

**Considered Approaches**
_Kernel Pipeline Approach:_
Began by implementing an option for the chipset in the -rock-blockwise-gemm-to-threadwise pass. This option would be available for determining conditions to implement the desired functionality. It involves adding this option to the kernel pipeline, and modifying the our tools rocmlir-driver -c to pass the architecture information for generating the correct instructions for mlir passes. Feedback is needed to assess whether it's beneficial to add this option to the kernel pipeline pass.

_ROCDL Backend Approach:_
Since the chipset is already integrated within the ROCDL dialect backend pipeline, and instructions are lowered in this pass, another approach would be to add custom intrinsics within AMDGPU.td. These can be used to create conditions within the ROCDL that lower the required instructions based on the chipset being utilized.

_Generic Solution with Readlane:_
A third potential solution is to attempt a more generalized approach by using readlane and modifying the logic to try to utilize as many common instructions supported across various architectures as possible. This would help in achieving a correct and optimal solution.


krzysz00:
Happened to notice this PR and wanted to let y'all know that https://github.com/llvm/llvm-project/pull/133204 is a thing that's under development and might be useful.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -1098,31 +1098,132 @@ struct BlockwiseReduceRewritePattern
...
-                loc, arith::CmpIPredicate::eq, rIterArg, lastIterVal);
-            scf::IfOp ifb = rewriter.create<scf::IfOp>(
-                loc, isLastIter, /*withElseRegion=*/false);
-            {
-              OpBuilder thenb = ifb.getThenBodyBuilder();
-              thenb.create<InBoundsStoreOp>(
-                  loc, reduced, workspaceLDSBuffer,
-                  reductionLoop.getLowerCoords(/*domain=*/2));
+
+            Value BrodcastAll;
```
dhernandez0: I think we don't need to have rocdl and amdgpu dialects here. Can we have some new op rock::wavereduction or something like that? Then, we can lower it to rocdl later.

dhernandez0: also, that way it's easier to keep the current implementation if dpp is not supported.

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -1098,31 +1098,132 @@ struct BlockwiseReduceRewritePattern
...
+              SmallVector<Value, 4> scalarDppResults;
+
+              for (int64_t i = 0; i < vecLen; ++i) {
+
+                Value scalarVal = rewriter.create<vector::ExtractElementOp>(
+                    loc, loadVal,
+                    rewriter.create<arith::ConstantIndexOp>(loc, i));
+                Value scalarInactiveValue = rewriter.create<arith::ConstantOp>(
+                    loc, vecType.getElementType(),
+                    rewriter.getFloatAttr(vecType.getElementType(), 0.0));
```
dhernandez0: for max reduction it should be -inf?

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -1098,31 +1098,132 @@ struct BlockwiseReduceRewritePattern
...
+
+                Value setInactiveScalar = rewriter.create<ROCDL::SetInactiveOp>(
+                    loc, vecType.getElementType(), scalarVal,
+                    scalarInactiveValue);
+
+                Value dppResult1 = rewriter.create<amdgpu::DPPOp>(
```
dhernandez0: please add comments to understand what the constants are doing here: 0xF, ...

mlir/lib/Dialect/Rock/Transforms/BlockwiseGemmToThreadwise.cpp
```diff
@@ -1098,31 +1098,132 @@ struct BlockwiseReduceRewritePattern
...
+                                             amdgpu::DPPPerm::row_shr),
+                    rewriter.getI32IntegerAttr(1),
+                    rewriter.getI32IntegerAttr(0xF),
+                    rewriter.getI32IntegerAttr(0xF),
+                    rewriter.getBoolAttr(false));
+
+                Value dppResult = createReducingOp(op, setInactiveScalar,
+                                                   dppResult1, rewriter);
+
+                Value dppResult2 = rewriter.create<amdgpu::DPPOp>(
```
dhernandez0: please add comments to understand what the constants are doing here: 0xF, ...

stefankoncarevic:
This is deprecated, new PR open: https://github.com/ROCm/rocMLIR/pull/2301

PR #1797, opened 2025-03-31
Fix attention bugs (swap thread and iter when Q LDS is bypassed and bf16 tests)

dhernandez0:
- do not swap thread and iter subdims for Q if we are bypassing LDS
- use f32 attention in CPU code (as we discussed to improve the high threshold situation for f16/bf16 attention)
- fix bug in maskKVCacheTosa for bf16
- report correct dimensions in getAttentionDimNames

Note: I tried to add the test case for ticket https://github.com/ROCm/rocMLIR-internal/issues/1797 in PrAttention* but when adding axes for transposeQ, tranposeV, tranposeO I found [another bug](https://github.com/ROCm/rocMLIR-internal/issues/1802). So, I created a new test: rock-attention-issue-1797.mlir which we can integrate in PrAttention* when it's fixed and remove rock-attention-issue-1797.mlir

solves: https://github.com/ROCm/rocMLIR-internal/issues/1797
solves: https://github.com/ROCm/rocMLIR-internal/issues/1690


pfultz2:
> use f32 attention in CPU code (as we discussed to improve the high threshold situation for f16/bf16 attention)

Is it possible to use `double` for CPU code? Do you think that would get even better results?

umangyadav (approved):

dhernandez0:
> > use f32 attention in CPU code (as we discussed to improve the high threshold situation for f16/bf16 attention)
> 
> Is it possible to use `double` for CPU code? Do you think that would get even better results?

Currently, if attention is f16, we use mfmas that (I think) do the multiplication in f16 and accumulate in f32. Our cpu side for verification does everything in f16, both multiplication and accumulation. That's an issue because we are comparing (mul-add) f16-f32 vs f16-f16 and they produce very different results in some cases (max absolute diff of 100).

My change here is to do cpu attention in f32 but I think the goal should be to be as close as possible to what we do in the gpu code. So, multiply in f16 and accumulate in f32. That's something we'll be able to implement once Tosa has acc_type implemented correctly in upstream (part of Tosa 1.0).

PR #1799, opened 2025-04-02
Fix kernel generation when kernelRepeats are more than 1

umangyadav:
Fixes issue described in https://github.com/ROCm/rocMLIR-internal/issues/1803

### DIFF
--- a/mlir/tools/rocmlir-gen/rocmlir-gen.cpp
+++ b/mlir/tools/rocmlir-gen/rocmlir-gen.cpp
@@ -3843,7 +3843,8 @@ static LogicalResult populateHostHarnessLogic(
       wrappedFuncs[kernel.func] = kernel.func;
     }
   }
-
+  llvm::dbgs() << "root0: \n";
+  root0.func.dump();
   // Redirect calls to kernel functions to point at wrapped functions.
   module.walk([&](CallOpInterface callOp) -> WalkResult {
     // Don't substitute the call inside the wrapper.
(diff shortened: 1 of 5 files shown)


mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -1321,11 +1322,12 @@ static func::FuncOp createGPUWrapper(ModuleOp module, const KernelIF &kernel) {
...
-  auto emitWrappedCall = [&kernel, &gpuMem](OpBuilder &b, Location loc,
-                                            Value ignoredIv,
-                                            ValueRange noArgs) {
-    auto wrappedCall = b.create<func::CallOp>(loc, kernel.func, gpuMem);
-    wrappedCall->setAttr("wrapped_call", b.getUnitAttr());
+  auto emitWrappedCall = [&kernels, &gpuMem](OpBuilder &b, Location loc,
+                                             Value ignoredIv,
+                                             ValueRange noArgs) {
+    for (const auto &kernel : kernels) {
+      b.create<func::CallOp>(loc, kernel.func, gpuMem);
```
dhernandez0, line 1329: shouldn't we keep:
wrappedCall->setAttr("wrapped_call", b.getUnitAttr()); 
as above?

dhernandez0 (approved):

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
umangyadav: `wrapped_call` attribute is used to replace `call @funcName` inside `main()` with one that would do GPU allocs, deallocs, gpu memcpy etc and then do `call @funcName`. Therefore named "wrapped_call". 

`wrapped_call` attr is useful for `pv` or `pv_with_gpu` only and not for `--clone-harness`. 

With `pv` and `pv_with_gpu` flags, this PR would generate only one "wrapped_call" that would invoke all sub kenrels.   Therefore only one kernel inside `main()` need to be replaced and all other calls to subkernels can be erased.  I can know that one kernel that require replacement without having "wrapped_call" attr.
that is what i do later.

PR #1800, opened 2025-04-03
Workaround issue 1802

dhernandez0:
Workaround for issue 1802. Disable doSwapThreadIterSubDims for Q and add tests by increasing test coverage of PrAttention*.

Also: add a barrier between loading from memory and storing to LDS when gemm0K == gemm0KPerBlock and doBypassLDSForQ = false. This should not happen as this code is disabled, but it's good to keep it correct just in case.


umangyadav (approved):

dhernandez0:
I'm reducing the coverage of i8 PrAttention because of this: https://github.com/ROCm/rocMLIR-internal/issues/1811

PR #1802, opened 2025-04-07
Update AmdArchDb.cpp with gfx950 target info

mirza-halilcevic:
- Resolves https://github.com/ROCm/rocMLIR-internal/issues/1775

### DIFF
--- a/mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
+++ b/mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
@@ -51,7 +51,7 @@ static constexpr AmdArchInfo
               /*maxSharedMemPerWG*/ 65536, /*numEUPerCU=*/4, /*minNumCU=*/228,
               /*hasFp8ConversionInstrs=*/true,
               /*hasOcpFp8ConversionInstrs=*/false, /*maxNumXCC=*/8),
-    gfx950Info(GemmFeatures::mfma | GemmFeatures::dot |
+    cdna35Info(GemmFeatures::mfma | GemmFeatures::dot |
                    GemmFeatures::atomic_add | GemmFeatures::atomic_add_f16 |
                    GemmFeatures::atomic_add_bf16,
                /*waveSize=*/64, /*maxWavesPerEU*/ 8, /*totalSGPRPerEU*/ 800,
@@ -93,11 +93,12 @@ AmdArchInfo mlir::rock::lookupArchInfo(StringRef arch) {
 
   StringRef minor = chip.take_back(2);
   StringRef major = chip.slice(0, chip.size() - 2);
-  if (major == "gfx9" && minor != "50") {
+  if (major == "gfx9") {
     return llvm::StringSwitch<AmdArchInfo>(minor)
         .Case("08", cdnaInfo)
         .Case("0a", cdna2Info)
         .Case("42", cdna3Info)
+        .Case("50", cdna35Info)
         // gfx906 has the dot product instructions, uniquely
         .Case("06", cdna50Info)
         .Default(gcnInfo);
@@ -131,9 +132,6 @@ AmdArchInfo mlir::rock::lookupArchInfo(StringRef arch) {
         bitEnumSet(gfx12Info.defaultFeatures, GemmFeatures::atomic_add_bf16);
 
     return gfx12Info;
-  }
-  if (major == "gfx9" && minor == "50") {
-    return gfx950Info;
   }
   llvm::errs() << "Warning: unknown architecture, falling back to defaults: "
                << arch << "\n";
--- a/mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
+++ b/mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
@@ -125,11 +125,14 @@ AmdArchInfo mlir::rock::lookupArchInfo(StringRef arch) {
     return gfx12Info;
   }
   if (major == "gfx9" && minor == "50") {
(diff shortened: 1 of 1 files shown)


mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -125,11 +125,14 @@ AmdArchInfo mlir::rock::lookupArchInfo(StringRef arch) {
     return gfx12Info;
   }
   if (major == "gfx9" && minor == "50") {
```
dhernandez0: at the beginning of the file you'll see cdna2Info, cdna3Info, ... I think instead of having this if/else we can have cdna4Info which corresponds to gfx950.

I think we can do the same for Navi4 btw, but feel free to leave that for another ticket.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -125,11 +125,14 @@ AmdArchInfo mlir::rock::lookupArchInfo(StringRef arch) {
...
-    // TODO (gfx950): some of those information are not accurate and need to be
-    // adjusted after hardware release
     AmdArchInfo gfx950Info(cdna3Info);
     gfx950Info.hasFp8ConversionInstrs = false;
     gfx950Info.hasOcpFp8ConversionInstrs = true;
+    gfx950Info.maxWavesPerEU = 8;
+    gfx950Info.totalSGPRPerEU = 800;
+    gfx950Info.totalSharedMemPerCU = 163840;
+    gfx950Info.maxSharedMemPerWG = 163840;
```
dhernandez0: if you can find information, it'd be nice to have maxNumXCC as well.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
umangyadav: gfx950 is technically still cdna3.5 and not cdna 4.

dhernandez0: good point, but I still think it's better to have more variables (for example we have one for gfx11) instead of this approach of copying an existing one and modifying some params. With this approach, in the future we might modify "cdna3Info" and that would affect gfx950Info, so we'll have to keep in mind to check all copies and see if they still make sense.

mirza-halilcevic: I created a `gfx950Info` global constant. I'll do the same for `gfx12Info` as part of https://github.com/ROCm/rocMLIR/pull/1790, to not pollute this PR.

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
mirza-halilcevic: @umangyadav Do you by any chance know where we can find this information?

umangyadav: Not sure where to find that. I am not sure why `numXCC` would be useful for rocMLIR though.

umangyadav: numXCC is number of chiplets. Perhaps you can find that in internal ISA doc

dhernandez0: > Not sure where to find that. I am not sure why `numXCC` would be useful for rocMLIR though.

We shuffle consecutive workgroups to make sure they are closer together (in the same chiplet) in makeGroupedGridLayout. For that to work correctly, we need the number of chiplets. If that has changed for gfx950 we need to update it, or we'll see worse performance.

There's also this follow-up thing we'll need to do: https://github.com/ROCm/rocMLIR-internal/issues/1801

mlir/lib/Dialect/Rock/utility/AmdArchDb.cpp
```diff
@@ -51,6 +51,14 @@ static constexpr AmdArchInfo
               /*maxSharedMemPerWG*/ 65536, /*numEUPerCU=*/4, /*minNumCU=*/228,
               /*hasFp8ConversionInstrs=*/true,
               /*hasOcpFp8ConversionInstrs=*/false, /*maxNumXCC=*/8),
+    gfx950Info(GemmFeatures::mfma | GemmFeatures::dot |
+                   GemmFeatures::atomic_add | GemmFeatures::atomic_add_f16 |
+                   GemmFeatures::atomic_add_bf16,
+               /*waveSize=*/64, /*maxWavesPerEU*/ 8, /*totalSGPRPerEU*/ 800,
+               /*totalVGPRPerEU*/ 512, /*totalSharedMemPerCU*/ 163840,
+               /*maxSharedMemPerWG*/ 163840, /*numEUPerCU=*/4, /*minNumCU=*/256,
```
dhernandez0, line 59: there are some gfx950 machines with fewer num cus I think, see rocminfo output in https://github.com/ROCm/rocMLIR-internal/issues/1760

I'm not sure if this is temporary and the final hardware will have a minimum of 256?

mirza-halilcevic: I'll take a look at the ISA doc to see if there are any clues regarding that.

umangyadav (approved):

dhernandez0 (approved):

umangyadav:
@mirza-halilcevic  looks like formatting is incorrect and Jenkins is complaining. Can you rebase with develop and fix formatting issues ?

PR #1794, opened 2025-03-27
Add Tier1 model configs to rocMLIR

dorde-antic:
Add obtained Tier1 model configs to the rocMLIR repository.

Resolves https://github.com/ROCm/rocMLIR-internal/issues/1754

PR #1804, opened 2025-04-09
Remove unused files

dhernandez0:
Removed unused files, questions:
- should we remove mlir/examples as well?
- most of mlir/docs seems irrelevant (except f8 truncation code)

solves: https://github.com/ROCm/rocMLIR-internal/issues/1736


dorde-antic (approved):
LGTM. In https://github.com/ROCm/rocMLIR/pull/1794 files from the PR should be also moved then to performance/configs/ ?

dhernandez0:
> LGTM. In #1794 files from the PR should be also moved then to performance/configs/ ?

yes please

mlir/utils/jenkins/miopen-tests/resnext101-miopen-configs
```diff
@@ -1,24 +0,0 @@
--n 256 -c 1024 -H 14 -W 14 -k 1024 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
```
umangyadav, line 1: I find it better to not clean up miopen related files for now and do it after making sure MIOpen is not using rocMLIR and making sure we don't need to test and maintain miopen client.

mlir/utils/performance/convertRocBlasToPerfRunner.py
```diff
@@ -1,87 +0,0 @@
...
-            perfRunnerIns["-transB"] = "false"
-
-    return stringify(perfRunnerIns)
-
-
-def main():
-    parser = argparse.ArgumentParser(
-        prog="rocBLAS converter",
-        description="converts rocblas-bench parameter to perfRunner parameters",
-        allow_abbrev=False,
```
umangyadav: I see some value in keeping this file

mlir/utils/jenkins/miopen-tests/resnext101-miopen-configs
dhernandez0: These files are not used in Jenkins, so they are not tested. Do we still want to keep them?

mlir/utils/performance/convertRocBlasToPerfRunner.py
dhernandez0: I've reverted the deletion of this file

mlir/utils/jenkins/miopen-tests/resnext101-miopen-configs
dhernandez0: @umangyadav

umangyadav (approved):

PR #1829, opened 2025-05-09
Causal attention

dhernandez0:
This PR implements efficient causal attention. 
The idea of causal attention is, given the attention matrix (output of first gemm) of shape [B*NUM_HEADS, SEQ_LEN_Q, SEQ_LEN_KV]. Set -inf for elements where key position > query position. This prevents tokens from attending to future positions, maintaining the autoregressive property of LLMs.

The current implementation is just a fusion (see [this](https://github.com/ROCm/rocMLIR/blob/ab200855ab14872a08726f4cd1983d193dc6000e/mlir/test/fusion/pr-e2e/attention/mixr-attention-kvcache-causal.mlir)). So, all the attention matrix is computed, and then invalid values are set to -inf. This PR avoids extra computation by taking into account causal masking. 

Note that this causes blocks to have very different number of iterations, the first block will have 1 iteration, while block N-1 will have N-1 interations. In the future, we can try to improve this by having half the blocks and assigning the work of 0 and N-1 to block 0, 1 and N-2 to block 1, etc.

Migraphx integration will be a future PR.

See the results of this PR vs develop (TFlops). For develop we should use a fusion to set invalid values to -inf, instead we use unfused kernel, so actual results of develop will be worse than shown here.

|  | develop | PR     | Speed up |
| - | ------- | ------ | -------- |
| 0 | 369.47  | 511.15 | 1.3835 |
| 1 | 401.39  | 638.61 | 1.5910 |
| 2 | 418.53  | 651.27 | 1.5561 |
| 3 | 421.61  | 755.04 | 1.7908 |
| 4 | 412.05  | 658.31 | 1.5976 |

These are the attention configs:
```
0: -transK true -t f16 -g 256 -seq_len_q 1024 -seq_len_k 1024 -head_dim_qk 128 -head_dim_v 128 -causal
1: -transK true -t f16 -g 128 -seq_len_q 2048 -seq_len_k 2048 -head_dim_qk 128 -head_dim_v 128 -causal
2: -transK true -t f16 -g 64 -seq_len_q 4096 -seq_len_k 4096 -head_dim_qk 128 -head_dim_v 128 -causal
3: -transK true -t f16 -g 32 -seq_len_q 8192 -seq_len_k 8192 -head_dim_qk 128 -head_dim_v 128 -causal
4: -transK true -t f16 -g 16 -seq_len_q 16384 -seq_len_k 16384 -head_dim_qk 128 -head_dim_v 128 -causal
```

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
@@ -1383,12 +1383,14 @@ struct GridwiseAttentionAccelRewritePattern
     }
   }
 
-  void setGemm0OutputOutOfScopeKVCache(
-      PatternRewriter &rewriter, Location loc,
+  enum class OutOfScopeType { KVCache, Causal };
+
+  void setGemm0OutputOutOfScope(
+      PatternRewriter &rewriter, Location loc, OutOfScopeType outOfScopeType,
       layout::GridCoordinates gridCoords, Value gemm0OutBuffer,
-      RegsAsMatrixSubTiles gemm0OutSubTileViews, Value currentSeqLen,
-      Value mLoopIV, Value gemm0MBlocksLastIter) const {
-    if (currentSeqLen) {
+      RegsAsMatrixSubTiles gemm0OutSubTileViews, bool enabled, Value mLoopIV,
+      Value gemm0MBlocksLastIter, Value currentSeqLen = nullptr) const {
+    if (enabled) {
       auto isLastIteration = rewriter.create<arith::CmpIOp>(
           loc, arith::CmpIPredicate::eq, mLoopIV, gemm0MBlocksLastIter);
       scf::IfOp ifb = rewriter.create<scf::IfOp>(loc, isLastIteration,
@@ -1422,8 +1424,19 @@ struct GridwiseAttentionAccelRewritePattern
 
           Block::BlockArgListType lowerCoords = loop.getLowerCoords(0);
           Block::BlockArgListType upperCoords = loop.getLowerCoords(1);
-          auto isInvalid = thenb.create<arith::CmpIOp>(
-              loc, arith::CmpIPredicate::ugt, lowerCoords[2], currentSeqLen);
+          Value isInvalid;
+          switch (outOfScopeType) {
+          case OutOfScopeType::KVCache:
+            assert(currentSeqLen != nullptr);
+            isInvalid = thenb.create<arith::CmpIOp>(
+                loc, arith::CmpIPredicate::ugt, lowerCoords[2], currentSeqLen);
+            break;
+          case OutOfScopeType::Causal:
+            isInvalid = thenb.create<arith::CmpIOp>(
+                loc, arith::CmpIPredicate::ugt, lowerCoords[2], lowerCoords[1]);
+            break;
+          }
+
           scf::IfOp ifb = thenb.create<scf::IfOp>(loc, isInvalid,
                                                   /*withElseRegion=*/false);
           {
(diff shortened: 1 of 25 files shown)


mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
```diff
@@ -662,6 +662,15 @@ getTuningProblemStr(RockGemmGemmWrapperInterface gemmGemmOp,
   else
     problemOS << "false" << sep;
 
+  if (isAttention) {
+    auto attentionOp = cast<AttentionOp>(gemmGemmOp);
+    problemOS << "-causal ";
+    if (attentionOp.getCausal())
+      problemOS << "true" << sep;
+    else
+      problemOS << "false" << sep;
```
umangyadav, line 671: Does this mean, it is changing problem config string for the attention ops ? 
Would previous attention problem configs continue working in backward compatible way ?

mlir/test/fusion/pr-e2e/attention-layouts/sliced-attention-q1k1v1-e2e.mlir
```diff
@@ -5,7 +5,7 @@
 // ALLOW_RETRIES: 2
 // CLONE: [1 1 1]
 
-// EMITKEY: -t f16 -transQ true -transK true -transV true -transO false -g 1 -seq_len_q 32 -seq_len_k 64 -head_dim_qk 16 -head_dim_v 8
+// EMITKEY: -t f16 -transQ true -transK true -transV true -transO false -causal false -g 1 -seq_len_q 32 -seq_len_k 64 -head_dim_qk 16 -head_dim_v 8
```
umangyadav, line 8: Can you add fused E2E test with causal masking ?

mlir/utils/performance/reportUtils.py
```diff
@@ -23,7 +23,7 @@
                         'X', 'DilationH', 'DilationW', 'StrideH', 'StrideW',
                         'PaddingH', 'PaddingW', 'PerfConfig']
 GEMM_TEST_PARAMETERS = ['DataType', 'OutDataType', 'Chip', 'numCU', 'TransA', 'TransB', 'G', 'M', 'K', 'N', 'PerfConfig']
-ATTN_TEST_PARAMETERS = ['DataType', 'Chip', 'numCU', 'TransQ', 'TransK', 'TransV', 'TransO', 'WithAttnScale', 'G', 'SeqLenQ', 'SeqLenK', 'HeadDimQK', 'HeadDimV', 'PerfConfig']
+ATTN_TEST_PARAMETERS = ['DataType', 'Chip', 'numCU', 'TransQ', 'TransK', 'TransV', 'TransO', 'Causal', 'WithAttnScale', 'G', 'SeqLenQ', 'SeqLenK', 'HeadDimQK', 'HeadDimV', 'PerfConfig']
```
umangyadav: Is jenkins generating perf Reports for Attention ?

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
dhernandez0: Depends on how the problem config is parsed/used. In our internal tools, perfRunner/tuningRunner/rocmlir-gen, would still work if not given "-causal", because it defaults to "false". If the problem config with "-causal" is given to an older version it would fail.

I don't know how migraphx uses problem config, do they parse it or just treat it as a string?

mlir/utils/performance/reportUtils.py
dhernandez0: I don't know. According to the code, there's "MLIR_Performance_Changes_Attention.html". This change is needed for perfRunner.py to run correctly.

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
dhernandez0: Also, when we do --emit-tuning-key (so I guess also what migraphx gets), we always get --causal {true/false}.

mlir/test/fusion/pr-e2e/attention-layouts/sliced-attention-q1k1v1-e2e.mlir
dhernandez0: I'd prefer to do that with migraphx dialect in the follow-up PR (migraphx integration). Otherwise, it'd have to be very long kernels with rock dialect. Is it ok, or should we add fusion tests now?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1436,6 +1436,60 @@ struct GridwiseAttentionAccelRewritePattern
     }
   }
 
+  void setGemm0OutputOutOfScopeCausal(PatternRewriter &rewriter, Location loc,
```
umangyadav: This and OutOfScopeKVCache code looks very similar with only diff in 

```
     Value isInvalid = thenb.create<arith::CmpIOp>(
              loc, arith::CmpIPredicate::ugt, lowerCoords[2], lowerCoords[1]);

  auto isInvalid = thenb.create<arith::CmpIOp>(
              loc, arith::CmpIPredicate::ugt, lowerCoords[2], currentSeqLen);
```

Is there a way to combine both codes ?

mlir/lib/Dialect/Rock/Tuning/RockTuningImpl.cpp
umangyadav: > do they parse it or just treat it as a string? 

They just use it as string for compilation cache purposes. ProblemConfig is the Key and perfConfig is the value in compilation cache hashMap. 

It should be okay i think.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2452,6 +2457,84 @@ static Value addTensorArgToBlock(OpBuilder &builder, Location loc,
...
+static Value causalMaskingTosa(OpBuilder builder, Location loc,
+                               Value inputTensor, float initValue) {
+  // create a range for row and column
+  auto inpType = cast<RankedTensorType>(inputTensor.getType());
+  ArrayRef<int64_t> inpShape = inpType.getShape();
+
+  Value rowRange = createRange(builder, loc, 1, inpShape);
+  Value colRange = createRange(builder, loc, 2, inpShape);
+
+  // create mask (diagonal and lower triagunlar are zero)
```
umangyadav:
```suggestion
  // create mask (diagonal and lower triangular are zero)
```
nit

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2452,6 +2457,84 @@ static Value addTensorArgToBlock(OpBuilder &builder, Location loc,
...
+  Value initVal = builder.create<tosa::ConstOp>(loc, initValueAttr.getType(),
+                                                initValueAttr);
+
+  // mask is 1 for values we want to set to "initVal"
+  auto result = createOpAndInfer<tosa::SelectOp>(
+      builder, loc, inpType.getElementType(), mask, initVal, inputTensor);
+  return result;
+}
+
+static Value createRange(OpBuilder builder, Location loc, size_t index,
```
umangyadav, line 2485: I think it would be better to move attention  or GemmPlusGemm generation functions to some other CPP File. 

No need to do it in this PR though.

dhernandez0: I agree, rocmlir-gen needs a refactor, also conv and gemm code can be moved to their own files.

PR #1830, opened 2025-05-09
Parameter Sweep for Attention

dorde-antic:
Implement parameter sweep for attention (`attentionSweeps.py`) that will test combinations of input shapes and perfConfigs for attention and find potential bugs. Adjust `parameterSweeps.py` so that attentionSweeps.py can reuse its methods.

Resolves https://github.com/ROCm/rocMLIR-internal/issues/1800


mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,212 @@
...
+
+    arch = ','.join(getArch())
+    paths = createPaths(None, args.mlir_build_dir)
+    options = Options(debug=args.debug, quiet=args.quiet, arch=arch, flags=[], concurrent_tests=args.jobs)
+   
+    attentionShapes = list(itertools.product(
+        DATA_TYPES_ATTENTION, GROUPS, SEQ_LENGTHS, SEQ_LENGTHS, HEAD_DIMS, HEAD_DIMS, BOOLS, BOOLS, BOOLS, BOOLS, BOOLS
+    ))
+
+    perfConfigSpace = list(itertools.product(
```
dhernandez0: this is not the attention perf config. See Rock_AttnPerfConfig in RockAttrDefs.td or RockTuningImpl.cpp (createAttnTuningRangeBF)

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,212 @@
...
+
+from perfRunner import AttentionConfiguration
+from perfRunner import Paths
+from perfRunner import getArch
+from perfRunner import getNumCU
+from perfRunner import create_paths as createPaths
+
+# GLOBAL VARIABLES
+DATA_TYPES_ATTENTION = ['i8', 'f32', 'f16', 'bf16']
+SEQ_LENGTHS = [1, 27, 64, 77, 128, 256, 384, 1024, 2048, 4097, 8192, 16384]
```
dhernandez0: I'd prefer to sample from 1 to 16384, including all possible numbers. Same for all variables.

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,212 @@
...
+HEAD_DIMS = [32, 64, 96, 128, 512]
+GROUPS = [1, 4, 8]
```
dhernandez0: group can be bigger (see attention-configs where it's 256), of course, some combination might not fit into memory. We have to avoid failing if the issue is memory.

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,212 @@
...
+    parser.add_argument('--samples', type=int, default=20)
+    parser.add_argument('--log-failures', action='store_true')
+
+    args = parser.parse_args()
+
+    arch = ','.join(getArch())
+    paths = createPaths(None, args.mlir_build_dir)
+    options = Options(debug=args.debug, quiet=args.quiet, arch=arch, flags=[], concurrent_tests=args.jobs)
+   
+    attentionShapes = list(itertools.product(
```
dhernandez0: there are some things not represented here (because of AttentionConfiguration):
- causal: coming in https://github.com/ROCm/rocMLIR/pull/1829)
- with_attn_bias: implemented in rocmlir-gen but for some reason not in AttentionConfiguration
- GQA: this is group query attention, in rocmlir-gen -num_heads_q 4 -num_heads_kv 2, not implemented in AttentionConfiguration
- KV-Cache: -current_seq_len=17,1,32 where the size must match -g 3 (3 numbers for current_seq_len). These are dynamic seq_lens (known at run-time). Not implemented in AttentionConfiguration.

mlir/utils/performance/attentionSweeps.py
dhernandez0: maybe you can represent this as a range() instead of a list

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,212 @@
...
+        [16, 32, 64, 128, 256], # N/block
+        [1, 2, 4, 8, 16, 32], # K/block
+        [4, 8, 16, 32, 64, 128], # M/wave
+        [4, 8, 16, 32, 64, 128], # N/wave
+        [1, 4, 8, 16], # kPack
+        [1, 2], # scheduleVersion
+        [0, 1] # aCopyMore/forceUnroll
+    ))
+    
+    samples = random.sample(list(itertools.product(attentionShapes, perfConfigSpace)), args.samples)
```
dhernandez0: for kv-cache, we can probably have another call to sampling. Because seq_len_q = 1, seq_len_kv = maxLenSet (which could be something like 1024, 2048, ...), then we need current_seq_len... 

Maybe we can leave kv-cache for a follow up PR?

mlir/utils/performance/attentionSweeps.py
dorde-antic: - causal: waiting until PR is merged
- with_attn_bias: implemented
- GQA: https://github.com/ROCm/rocMLIR-internal/issues/1834
- KV-Cache: https://github.com/ROCm/rocMLIR-internal/issues/1835

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+def main():
+    parser = argparse.ArgumentParser(
+            description='Sweep parameter values for attention to detect bugs')
+    parser.add_argument('--debug', action='store_true')
+    parser.add_argument('--quiet', action='store_true')
+    parser.add_argument('--jobs', type=int, default=4)
+    parser.add_argument('--mlir-build-dir', type=str, required=True)
+    parser.add_argument('--samples', type=int, default=20)
```
dhernandez0: make default bigger

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,304 @@
...
+
+    passed, invalid, failing = asyncio.run(sweepParameters(samples, toAttentionConfig, options, paths))
+    print(f"Passed: {passed}, Invalid: {invalid}, Failed: {len(failing)}")
+    if failing:
+        print("\n*** Failing Configurations ***")
+        for fail in failing:
+            print(' '.join(fail.generateMlirDriverArgs([])))
+        if args.log_failures:
+            logFailingConfigs(failing, LOGFILE)
+
```
dhernandez0, line 215: return something because the return value is used below.

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,304 @@
...
+
+def sampleAttentionShape():
+    return (
+        random.choice(DATA_TYPES_ATTENTION),
+        random.randint(1, 16384), # GROUPS
+        random.randint(1, 16384), # SEQ_LEN_Q
+        random.randint(1, 16384), # SEQ_LEN_K
+        random.randint(1, 16384), # HEAD_DIM_QK
+        random.randint(1, 16384), # HEAD_DIM_V
```
dhernandez0: I don't think head dims are typically that large, maybe 512 or 1024 max?

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,304 @@
...
+GFX_CHIP_RE = re.compile(r"gfx[0-9a-z]+")
+
+
+class PerfConfigVersion(Enum):
+    '''Currently, attention uses only v1, so it's hardcoded in file at the moment,
+    Enumeration for PerfConfigVersion is not removed in case we reintroduce higher
+    perfConfig versions.'''
+    V1 = 'v1'
+    V2 = 'v2'
+    V3 = 'v3'
```
dhernandez0: there's only v1 for attention

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,304 @@
...
+
+from perfRunner import AttentionConfiguration
+from perfRunner import Paths
+from perfRunner import getArch
+from perfRunner import getNumCU
+from perfRunner import create_paths as createPaths
+from perfRunner import MLIR_N_REPEATS
+
+# GLOBAL VARIABLES
+DATA_TYPES_ATTENTION = ['f32', 'f16', 'bf16']
```
dhernandez0: i8 missing

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,311 @@
...
+        if options.debug:
+            print("TimeoutExpired: cpu-runner timed out")
+        return TestResult.FAIL
+
+    if proc3.returncode != 0:
+        if options.debug:
+            print("Runner failed:\nOutput = ", stdout3.decode().strip())
+            print("\nError = ", stderr3.decode().strip())
+        
+        if 'hipErrorOutOfMemory' in stderr3.decode().strip():
```
dhernandez0: to decide if something fits into memory and generally if it can be run, you can use:
rocmlir-gen ... | rocmlir-driver --kernel-pipeline=applicability

if it fails applicability it means, for some reason we can't run this problem with that particular perfconfig. 
You can check parametersSweep.py, they do the same for convs there.

dorde-antic:
https://github.com/ROCm/rocMLIR-internal/issues/1835#issuecomment-2894645293
_For the randomness, make sure it is re-creatable. So the seed should be capable of rerunning over and over. Possibly use the week of the year so the same random number is used for 7 days_

dorde-antic:
**when sampling num-heads (gqa):**
- sample x=[1,n] natural numbers, then do 2**x
- num_heads_q >= num_heads_kv and num_heads_q % num_heads_kv == 0 are requirements

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,311 @@
...
+        stdout=asyncio.subprocess.PIPE,
+        stderr=asyncio.subprocess.PIPE
+    )
+
+    rocmlirGeneratorOutput, rocmlirGeneratorError = await proc1.communicate()
+
+    if proc1.returncode !=0:
+        if options.debug:
+            print("rocmlir-gen failed:\nError = ", rocmlirGeneratorError.decode().strip())
+        return TestResult.INVALID
```
dhernandez0: I think if anything fails should be fail. The only INVALID should come from applicability failing.

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,311 @@
...
+        f"-transV={self.transV}",
+        f"-transO={self.transO}",
+        '--kernel-repeats', str(MLIR_N_REPEATS),
+        f"--perf_config={self.perfConfig}"
+    ]
+
+    result += rocmlir_gen_flags or []
+    return result
+
+AttentionConfiguration.generateMlirDriverArgs = generateMlirDriverArgs
```
dhernandez0: why do we need to do this?

dorde-antic: To avoid making whole new class that would inherit AttentionConfiguration (like it was done in _parameterSweeps.py_) and to avoid adding new class method in _perfRunner_ inside AttentionConfiguration class which would be only used in _attentionSweeps.py_ file (at least for now). I thought we could extend AttentionConfiguration class dynamically with generateMlirDriverArgs() method which is implemented in _attentionSweeps.py_ file using line 83. Without "attaching" generateMlirDriverArgs like this to AttentionConfiguration, we would get AttributeError, since objects of AttentionConfiguration class wouldn't have attribute generateMlirDriverArgs which we call in testAttentionConfig method.

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,359 @@
...
+# GLOBAL VARIABLES
+DATA_TYPES_ATTENTION = ['i8', 'f32', 'f16', 'bf16']
+BOOLS = [True, False]
+LOGFILE = 'failing_configs.csv'
+GFX_CHIP_RE = re.compile(r"gfx[0-9a-z]+")
+CURRENT_SEQ_LEN = None
+
+# Week number is used as seed
+seed = datetime.utcnow().isocalendar()[1]
+random.seed(seed)
```
dhernandez0, line 39: do we use any other library to get random numbers?

mlir/utils/performance/attentionSweeps.py
dhernandez0: can't we use generateMlirDriverCommandLine that already exists in AttentionConfiguration?

dhernandez0: that one has causal, which generateMlirDriverArgs is missing. And it seems it's pretty much the same thing?

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,359 @@
...
+
+
+def toAttentionConfig(params, options: Options) -> AttentionConfiguration:
+    """Converts a sampled parameter tuple into a AttentionConfiguration instance"""
+    shape, perf = params
+    dtype, g, slq, slk, hdqk, hdv, scale, bias, tq, tk, tv, to = shape
+    perfString = f"attn:v1:{','.join(str(x) for x in perf)}"
+    return AttentionConfiguration(
+        dtype, g, slq, slk, hdqk, hdv, scale,
+        bias, tq, tk, tv, to, options.arch,
```
dhernandez0: causal is missing here.

dhernandez0: I'd prefer to use dtype=dtype, ... so that if at some point we add new args to the constructor this will still work

mlir/utils/performance/attentionSweeps.py
dhernandez0: Ideally, we run applicability, if it fails, return TestResult.INVALID. Any other failure would be TestResult.FAIL

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,359 @@
...
+    numHeadsKV = 1
+    '''By default numHeadsQ and numHeadsKV are both 1. If numHeadsQ
+    and numHeadsKV are equal GQA is disabled. Both values are powers
+    of 2 typically. And numHeadsQ is divisible by numHeadsKV
+    Here we decide randomly if we will use numHeadsQ and numHeadsKV
+    different from the default values.
+    
+    Requirements:
+        - numHeadsQ >= numHeadsKV
+        - numHeadsQ % numHeadsKV == 0'''
```
dhernandez0, line 101: nice comment!

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,359 @@
...
+    genNumHeads = random.choice(BOOLS)
+    if genNumHeads:
+        while True:
+            numHeadsQ = 2**random.randint(1, 6)
+            numHeadsKV = 2**random.randint(1, 6)
+
+            if numHeadsQ >= numHeadsKV and numHeadsQ%numHeadsKV == 0: # found valid case
```
dhernandez0: if it's equal, we are actually not running GQA.

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,359 @@
...
+    '''By default numHeadsQ and numHeadsKV are both 1. If numHeadsQ
+    and numHeadsKV are equal GQA is disabled. Both values are powers
+    of 2 typically. And numHeadsQ is divisible by numHeadsKV
+    Here we decide randomly if we will use numHeadsQ and numHeadsKV
+    different from the default values.
+    
+    Requirements:
+        - numHeadsQ >= numHeadsKV
+        - numHeadsQ % numHeadsKV == 0'''
+    genNumHeads = random.choice(BOOLS)
```
dhernandez0, line 102: we might want non-50% chance here. But let's discuss that in the next meeting.

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,359 @@
...
+
+
+def genCurrentSeqLens(g: int, maxSeqLen: int) -> list[int]:
+    return [random.randint(1, maxSeqLen) for _ in range(g)]
+
+
+def sampleAttentionShape():
+    global CURRENT_SEQ_LEN
+
+    g = random.randint(1, 16384)
```
dhernandez0: let's do random.randint(1, 256) here

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,359 @@
...
+        random.choice(DATA_TYPES_ATTENTION),
+        g, # GROUPS
+        seqLenQ, # SEQ_LEN_Q
+        seqLenK, # SEQ_LEN_K
+        numHeadsQ, # NUM_HEADS_Q
+        numHeadsKV, # NUM_HEADS_KV
+        random.randint(1, 1024), # HEAD_DIM_QK
+        random.randint(1, 1024), # HEAD_DIM_V
+        random.choice(BOOLS),   # with_attn_scale
+        random.choice(BOOLS),   # with_attn_bias
```
dhernandez0, line 121: missing causal

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,359 @@
...
+
+perfConfigSpace = list(itertools.product(
+        [32, 64, 128, 256], # M/block G0
+        [32, 64, 128, 256], # M/block G1
+        [32, 64, 128, 256], # N/block G0
+        [8, 16, 32, 64], # Kpack/Block
+        [32, 64, 128, 256], # M/Wave
+        [4, 16, 32], # MN/Xdl
+        [4, 8, 16], # kPack
+        [0, 1] # forceUnroll
```
dhernandez0, line 140: due to an old bug (nobody got around to fix it, maybe it doesn't happen anymore?), forceUnroll is always 1. https://github.com/ROCm/rocMLIR-internal/issues/790.

No need to iterate over 0, 1

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,359 @@
...
+        random.choice(BOOLS),   # with_attn_scale
+        random.choice(BOOLS),   # with_attn_bias
+        random.choice(BOOLS),   # transQ
+        random.choice(BOOLS),   # transK
+        random.choice(BOOLS),   # transV
+        random.choice(BOOLS),   # transO
+    )
+
+
+perfConfigSpace = list(itertools.product(
```
dhernandez0: note that the perfConfig is different for wmma/mfma. The code here should take that into account.

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,359 @@
...
+
+def sampleAttentionShape():
+    global CURRENT_SEQ_LEN
+
+    g = random.randint(1, 16384)
+    seqLenQ = random.randint(1, 16384) # SEQ_LEN_Q
+    seqLenK = random.randint(1, 16384) # SEQ_LEN_K 
+
+    useKVCache = random.choice(BOOLS)
+    CURRENT_SEQ_LEN = genCurrentSeqLens(g, seqLenK) if useKVCache else None
```
dhernandez0: if useKVCache is true, we should force seqLenQ = 1. To emulate what kv-cache looks like. Also, maybe make genNumHeads more likely (higher probability).

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,359 @@
...
+from perfRunner import MLIR_N_REPEATS
+
+# GLOBAL VARIABLES
+DATA_TYPES_ATTENTION = ['i8', 'f32', 'f16', 'bf16']
+BOOLS = [True, False]
+LOGFILE = 'failing_configs.csv'
+GFX_CHIP_RE = re.compile(r"gfx[0-9a-z]+")
+CURRENT_SEQ_LEN = None
+
+# Week number is used as seed
```
dhernandez0: expand comment: to make sure weekly CI is reproducible... something like that

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,359 @@
...
+        '-num_heads_kv', str(self.num_heads_kv),
+        '-head_dim_qk', str(self.head_dim_qk),
+        '-head_dim_v', str(self.head_dim_v),
+        f"-with-attn-scale={self.with_attn_scale}",
+        f"-with-attn-bias={self.with_attn_bias}",
+        f"-transQ={self.transQ}",
+        f"-transK={self.transK}",
+        f"-transV={self.transV}",
+        f"-transO={self.transO}",
+        '--kernel-repeats', str(MLIR_N_REPEATS),
```
dhernandez0: in this case, as we don't want to measure run-time, 1 is ok. It'd be faster as well.

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,359 @@
...
+        for config in configs:
+            writer.writerow([' '.join(config.generateMlirDriverArgs([]))])
+
+
+def main():
+    parser = argparse.ArgumentParser(
+            description='Sweep parameter values for attention to detect bugs')
+    parser.add_argument('--debug', action='store_true')
+    parser.add_argument('--quiet', action='store_true')
+    parser.add_argument('--jobs', type=int, default=4)
```
dhernandez0: is it possible to default to num cpus/cores somehow? if it's not easy let's leave it like this.

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,359 @@
...
+            elif result == TestResult.INVALID:
+                invalid = invalid + 1
+            else:
+                failingConfigs.append(result)
+
+    return (passed, invalid, failingConfigs)
+
+
+def genCurrentSeqLens(g: int, maxSeqLen: int) -> list[int]:
+    return [random.randint(1, maxSeqLen) for _ in range(g)]
```
dhernandez0: due to a recent change: https://github.com/ROCm/rocMLIR/pull/1827. The valid range is 0 to maxSeqLen-1 (both included).

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,359 @@
...
+
+
+def sampleAttentionShape():
+    global CURRENT_SEQ_LEN
```
dhernandez0: can we reutrn this as the other params instead of using a global?

mlir/utils/performance/attentionSweeps.py
dorde-antic: Yes, but it's the way that we sample other valid values of numHeadsQ and numHeadsKV that are not default (not 1) but equal (for example both of them 2 or 8... ).

dhernandez0: I'm not sure I follow what you mean. What I mean is, we decided previously whether GQA is going to run (genNumHeads is true). Then, we want GQA to run, so we have to make sure numHeadsQ > numHeadsKV (not ==), otherwise, if they are equal GQA will not run.

mlir/utils/performance/attentionSweeps.py
dhernandez0: On second thought, let's leave it like this. So, when/if it's fixed, it's already iterating over it. Currently, 0 will fail on applicability really fast.

dorde-antic:
TODO:
- [x]  Consider the new flag from https://github.com/ROCm/rocMLIR/pull/1882

mlir/utils/performance/attentionSweeps.py
dorde-antic: The easiest way would be using default=os.cpu_count() if that is okay? Tried on Navi4x, returns 32

mlir/utils/performance/attentionSweeps.py
dorde-antic: Resolved on Teams (We use built-in module in python.)

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,417 @@
+#!/usr/bin/env python3
+"""Sweeps the parameters of the rocmlir driver for bugs for attention-based kernel configurations.
+
+Usage:
+    python3 attentionSweeps.py --mlir-build-dir <path-to-mlir-build-dir> [options]
+
+Options:
+    --mlir-build-dir    Path to the MLIR build directory (required)
+    --samples           Number of random configuration samples to the test (default: 1000)
+    --jobs              Number of concurrent tests to run in parallel (default: 4)
```
dhernandez0: default: os.cpu_count()

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,417 @@
...
+        for config in configs:
+            writer.writerow([' '.join(config.generateMlirDriverCommandLine(''))])
+            
+def main():
+    parser = argparse.ArgumentParser(
+            description='Sweep parameter values for attention to detect bugs')
+    parser.add_argument('--debug', action='store_true')
+    parser.add_argument('--quiet', action='store_true')
+    parser.add_argument('--jobs', type=int, default=os.cpu_count())
+    parser.add_argument('--mlir-build-dir', type=str, required=True)
```
dhernandez0: use perfRunner.find_mlir_build_dir() to have a default value, see parameterSweeps.py. Then, make it not required.

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,417 @@
...
+    parser.add_argument('--samples', type=int, default=1000)
```
dhernandez0, line 168: I wonder if we should use a time limit instead of number of samples? Because different machines will take diff time for 1k samples. @umangyadav

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,417 @@
...
+import re
+
+from perfRunner import AttentionConfiguration
+from perfRunner import Paths
+from perfRunner import getArch
+from perfRunner import getNumCU
+from perfRunner import create_paths as createPaths
+
+# GLOBAL VARIABLES
+DATA_TYPES_ATTENTION = ['i8', 'f32', 'f16', 'bf16']
```
dhernandez0: import from perfRunner

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,417 @@
...
+BOOLS = [True, False]
+LOGFILE = 'failing_configs.csv'
+GFX_CHIP_RE = re.compile(r"gfx[0-9a-z]+")
```
dhernandez0: import from perfRunner

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,417 @@
...
+
+# Week number is used as seed to make sure weekly CI is reproducible
+seed = datetime.utcnow().isocalendar()[1]
+random.seed(seed)
+
+class TestResult(Enum):
```
dhernandez0: Import from paramterSweep.py, it's already defined there.

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,417 @@
...
+    PASS = 1
+    INVALID = 2
+    FAIL = 3
+    
+@dataclass(frozen=True)
+class Options:
```
dhernandez0: import from perfRunner

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,417 @@
...
+        return_lse=rlse,
+        arch=options.arch,
+        numCU=options.numCu,
+        perf_config=perfString
+    )
+    attnConfig.currentSeqLen = currentSeqLen
+    return attnConfig
+
+
+def multilineRepr(obj, num_fields=4):
```
dhernandez0: can we use textwrap:
import textwrap

wrapped_text = textwrap.fill(text, width)

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,417 @@
...
+        if j + num_fields >= len(fields) and group and group[-1].endswith(','):
+            group[-1] = group[-1][:-1]
+        lines.append(f"{prefix}{' '.join(group)}")
+    if perf_config_str:
+        lines.append('\t' + perf_config_str.strip())
+        
+    return '\n'.join(lines)
+
+
+async def testAttentionConfig(config: AttentionConfiguration, options: Options, paths) -> TestResult:
```
dhernandez0: you are not using kernel-pipeline=applicability to decide if it's TestResult.INVALID. All the rest of failures must be TestResult.FAIL

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,417 @@
...
+            print("FAILED in output or NaN:\nOutput = ", output)
+        return TestResult.FAIL
+    if not options.quiet:
+        print(f"PASS: {multilineRepr(config)}")
+
+    return TestResult.PASS
+
+
+IterType = TypeVar('IterType')
+def grouper(iterable: Iterable[IterType], n: int):
```
dhernandez0, line 72: import from parameterSweep.py

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,417 @@
...
+            return
+        yield chunk
+
+
+async def dropGoodConfig(config: AttentionConfiguration,
+        options: Options, paths: Paths) -> Union[TestResult, AttentionConfiguration]:
+    """Test the given `params`, returning the corresponding `config` on failure
+    and `None` on success or inapplicability."""
+    result = await testAttentionConfig(config, options, paths)
+
```
dhernandez0: see paramterSweep dropGoodConfig function, it prints out value as well:
if not options.quiet:
        print(f"{result.name}: {config!r}")

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,417 @@
...
+def grouper(iterable: Iterable[IterType], n: int):
+    it = iter(iterable)
+    while True:
+        chunk = tuple(itertools.islice(it, n))
+        if not chunk:
+            return
+        yield chunk
+
+
+async def dropGoodConfig(config: AttentionConfiguration,
```
dhernandez0: can we use the function in parameterSweep.py? maybe pass the function instead of calling testAttentionConfig or testConfig

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,417 @@
...
+        options: Options, paths: Paths) -> Union[TestResult, AttentionConfiguration]:
+    """Test the given `params`, returning the corresponding `config` on failure
+    and `None` on success or inapplicability."""
+    result = await testAttentionConfig(config, options, paths)
+
+    return config if result == TestResult.FAIL else result
+
+
+async def sweepParameters(paramIter: Iterable[IterType],
```
dhernandez0: same here, try to use the function in paramterSweep, passing functions if needed.

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,417 @@
...
+        random.choice(BOOLS),   # transK
+        random.choice(BOOLS),   # transV
+        random.choice(BOOLS),   # transO
+        random.choice(BOOLS),   # causal
+        random.choice(BOOLS),   # return_lse
+        currentSeqLen
+    )
+
+
+perfConfigSpaceMFMA = list(itertools.product( # MFMA perfConfig space
```
dhernandez0, line 132: add a comment to keep this in sync with RockTuningImpl.cpp. Also, please add a comment in RockTuningImpl.cpp as well to keep in sync with this file.

mlir/utils/performance/attentionSweeps.py
```diff
@@ -0,0 +1,417 @@
...
+        flags=[],
+        concurrentTests=args.jobs,
+        numCu=getNumCU(chip)
+    )
+   
+
+    if not args.quiet:
+        print(f"Sampling {args.samples} configurations from attention space...")
+
+    if chip in ['gfx900', 'gfx906', 'gfx908', 'gfx90a', 'gfx940', 'gfx941', 'gfx942']:
```
dhernandez0: add gfx950 as well, gfx940 and gx941 are deprecated. Maybe, we can even do .startsWith('gfx9')?

dhernandez0: also, add a TODO to use AmdArchDb python version when available.

mlir/utils/performance/attentionSweeps.py
dhernandez0: also, wmma doesn't support f32. We need to have some kind of special case for that.

mlir/utils/performance/attentionSweeps.py
dorde-antic: > I wonder if we should use a time limit instead of number of samples? Because different machines will take diff time for 1k samples. @umangyadav @dhernandez0 

There is option to add timeout in the CI function which would call attentionSweeps.py, which can be used independently of the samples number.

dorde-antic: Maybe void parameterSweep(...) in CI, could be changed to something like:
```
void parameterSweep(String CONFIG, String codepath, String sweepType = "parameter") {
    timeout(time: 300, activity: true, unit: 'MINUTES') {
        dir('build') {
            if (sweepType == "attention") {
                sh """python3 ./bin/attentionSweeps.py"""
            } else {
                sh """python3 ./bin/parameterSweeps.py -j 5 ${CONFIG}"""
            }
        }
    }
}
```
it should be time limited then I think ...

and call in the stage would be something like:
```
stage("Parameter Sweep") {
                        steps {
                            script {
                         // ...
                                    ]) {
                                        parameterSweep("conv_structure", "${CODEPATH}")
                                        parameterSweep("perf_config", "${CODEPATH}")
                                        parameterSweep("", "${CODEPATH}", "attention")
                                    }
                                }
                            }
                        }       
                    }
```

dhernandez0: but I think if it times out, it will make CI fail?

dhernandez0 (approved):
LGTM, thanks for creating this script, great work!

dorde-antic:
TODO (in a follow-up tickets/PRs):
- Consider the new flag from https://github.com/ROCm/rocMLIR/pull/1895 when it's merged
- Add attentionSweeps.py to CI in https://github.com/ROCm/rocMLIR-internal/issues/1837

dorde-antic:
`perfRunner.py` and `tuningRunner.py` were tested after change in perfRunner from [67f0f4c](https://github.com/ROCm/rocMLIR/pull/1830/commits/67f0f4c84da1021041d5f5c2340ca0efece6d367)

PR #1832, opened 2025-05-12
[CI] Set 3600s test timeout and update LIT worker configuration

stefankoncarevic:
Temporarily increasing the CI timeout from 1 hour to 2 hours to gather more insights on nightly tests and improve reliability.

### DIFF
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -522,7 +522,9 @@ pipeline {
                             script {
                                 build_fixedE2ETests("${CODEPATH}")
                                 preMergeCheck("${CODEPATH}")
-                                sh 'cd build; ninja check-mlir check-rocmlir'
+                                timeout(time: 60, activity: true, unit: 'MINUTES') {
+                                    sh 'cd build; ninja check-mlir check-rocmlir'
+                                }
                             }
                         }
                     }
--- a/mlir/utils/jenkins/Jenkinsfile
+++ b/mlir/utils/jenkins/Jenkinsfile
@@ -245,7 +245,7 @@ void build_fixedE2ETests(String codepath) {
               -DROCMLIR_DRIVER_E2E_TEST_ENABLED=${params.nightly ? '1' : '0'}
               -DROCK_E2E_TEST_ENABLED=${params.nightly ? '1' : '0'}
               -DROCMLIR_DRIVER_TEST_GPU_VALIDATION=1
-              -DLLVM_LIT_ARGS='${llvm_lit_log_level} --time-tests --timeout=3600 --max-failures=1 ${ limit_lit_workers ? '-j 8' : ' ' }'
+              -DLLVM_LIT_ARGS='${llvm_lit_log_level} --time-tests --timeout=3600 --max-failures=1 ${ limit_lit_workers ? '-j 8' : '-j 40' }'
               -DCMAKE_EXPORT_COMPILE_COMMANDS=1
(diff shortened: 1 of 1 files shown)


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -522,7 +522,7 @@ pipeline {
                             script {
                                 build_fixedE2ETests("${CODEPATH}")
                                 preMergeCheck("${CODEPATH}")
-                                timeout(time: 60, activity: true, unit: 'MINUTES') {
+                                timeout(time: 120, activity: true, unit: 'MINUTES') {
```
umangyadav: This timeout is on the activity not on the how much time is spent on running the tests. 
So if it doesn't print anything on the logs for 60 minutes that is when it times out. 

Therefore i think we don't need  increase the timeout limit. 

You can look up `activity: true` flag meaning.

stefankoncarevic: `The activity: true` flag ensures that the timeout measures log activity during stage execution — as long as there's some output, the timer resets.
If we remove `activity: true,` the timeout will measure the total time from the start of the stage, and once the limit is reached, the stage will be stopped regardless of whether tests are still running or producing output.

stefankoncarevic: What I tested is that the `--timeout` flag for `lit` works as expected — if any test exceeds the specified timeout, it will eventually be marked as `FAILED.` This gives us a reliable way to detect which specific test is taking too long. So we can add the `--timeout` option in the CMake configuration (via `LLVM_LIT_ARGS`), and it will help us identify the problematic tests without needing to increase the overall stage timeout.

umangyadav (approved):

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -522,9 +522,7 @@ pipeline {
                             script {
                                 build_fixedE2ETests("${CODEPATH}")
                                 preMergeCheck("${CODEPATH}")
-                                timeout(time: 60, activity: true, unit: 'MINUTES') {
```
umangyadav: IMO We should keep timeout on the activity

stefankoncarevic: I agree, that it will be necessary to return it for cases when the machine is disconnected and the tests are not completed, because there will be nothing to interrupt the stage.

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -245,7 +245,7 @@ void build_fixedE2ETests(String codepath) {
               -DROCMLIR_DRIVER_E2E_TEST_ENABLED=${params.nightly ? '1' : '0'}
               -DROCK_E2E_TEST_ENABLED=${params.nightly ? '1' : '0'}
               -DROCMLIR_DRIVER_TEST_GPU_VALIDATION=1
-              -DLLVM_LIT_ARGS='${llvm_lit_log_level} --time-tests ${ limit_lit_workers ? '-j 8' : ' ' }'
+              -DLLVM_LIT_ARGS='${llvm_lit_log_level} --time-tests --timeout=3600 --max-failures=1 ${ limit_lit_workers ? '-j 8' : '-j 40' }'
```
umangyadav, line 248: Why do you need to  limit Lit works for MFMA path ?

stefankoncarevic:
I'll be merging these changes in for a specific period. The main goal is to try and make our nightly runs more stable for the time being with this update.

PR #1833, opened 2025-05-12
Change rounding mode for FP32 to Fp16 truncation

umangyadav:
By default ML applications expects Round to Nearest Even (RNE) rounding mode. 
To use packed RTZ (round towards zero) instruction, rocMLIR was trading off performance with accuracy previously. 
Based on benchmarks done by MIGraphX it was found that there is no performance impact by using RNE rounding mode and accuracy improves. 

Therefore, This PR changes rocMLIR pipeline to use RNE rounding mode instead of RTZ.


dhernandez0 (approved):

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
```diff
@@ -236,7 +236,7 @@ void rock::buildBackendPipeline(OpPassManager &pm,
   gpuPm.addPass(arith::createArithEmulateUnsupportedFloats(floatEmuOpts));
   ArithToAMDGPUConversionPassOptions arithOptions;
   arithOptions.chipset = options.chip;
-  arithOptions.allowPackedF16Rtz = true;
+  arithOptions.allowPackedF16Rtz = false;
```
dhernandez0, line 241: nit: add a comment with some explanation or link to this PR

mlir/test/Dialect/Rock/test_packed_arithmetic.mlir
dhernandez0, line 1: nit: name of the test doesn't make sense after this change

umangyadav: It still does it generates `v_pk_add_f16`

mlir/lib/Dialect/Rock/Pipelines/Pipelines.cpp
umangyadav: Done.

PR #1841, opened 2025-05-19
Correct rocprof invocation in fusion benchmarking path.

stefankoncarevic:


dorde-antic (approved):

mlir/utils/performance/perfRunner.py
```diff
@@ -1323,10 +1323,10 @@ def runFusionKernel(filename, rocmlirGenArgs, paths: Paths):
 
     rocmlirGenCommand = [paths.mlir_paths.rocmlir_gen_path] + rocmlirGenArgs
     commands.append(rocmlirGenCommand)
-    kernelPipelineCommand = [paths.mlir_paths.rocmlir_driver_path, '-host-pipeline', 'mhal,runner', '-kernel-pipeline','full']
+    kernelPipelineCommand = [paths.mlir_paths.rocmlir_driver_path, '-host-pipeline', 'mhal,runner', '-kernel-pipeline', 'full']
     commands.append(kernelPipelineCommand)
     mlir_cpu_runner_args = [f'--shared-libs={paths.mlir_paths.libmlir_rocm_runtime_path},{paths.mlir_paths.libconv_validation_wrappers_path},{paths.mlir_paths.libmlir_runtime_utils_path},{paths.mlir_paths.libmlir_c_runner_utils_path}', '--entry-point-result=void']
-    profilerCommand = [ROCPROF] + getMetricArgsForRocprof(getChip()) + ['--kernel-trace', '--stats', '-o', BENCHMARKING_RESULT_FILE_NAME] + '--' + [paths.mlir_paths.cpu_runner_path] + mlir_cpu_runner_args
+    profilerCommand = [ROCPROF] + getMetricArgsForRocprof(getChip()) + ['--kernel-trace', '--stats', '-o', BENCHMARKING_RESULT_FILE_NAME] + ['--', paths.mlir_paths.cpu_runner_path] + mlir_cpu_runner_args
```
dhernandez0, line 1337: isn't there a space added between every element in the list?

dhernandez0: why is the "--" needed?

stefankoncarevic: > isn't there a space added between every element in the list?
Yes, when you use subprocess.run(profilerCommand), all list elements are automatically treated as separate arguments.

dhernandez0: then, this would end up being "-- path/mlir_runner", is that what we want? I don't understand why we need "--" here

stefankoncarevic: > why is the "--" needed?

 I think that "--" is needed to separate rocprofv3 options from the actual runner command and its arguments, ensuring no conflict.  This is related to changes for rocprofv3 so I think @djramic  knows better why exactly it is needed.

djramic: We need " -- ", that's a change in rocprofv3, it now expects "--" before invoking the app

dhernandez0 (approved):

PR #1844, opened 2025-05-21
tuneMLIRKernels stdout

dorde-antic:
This PR is made to test (on CI) if adding potentially missing `kernelGen.stdout.close()` will solve the problem we are having with tuningRunner in Tune MLIR phase.

### DIFF
--- a/mlir/utils/performance/tuningRunner.py
+++ b/mlir/utils/performance/tuningRunner.py
@@ -167,6 +167,7 @@ def tuneMLIRKernels(configs, confClass, paths: Paths, options: Options):
             config = confClass.fromCommandLine(commandLine, options.arch, options.numCU)
             tuningLoop = subprocess.Popen([paths.mlir_paths.rocmlir_tuning_driver_path, f"--tuning-space={options.tuningSpaceKind}", testVector],
                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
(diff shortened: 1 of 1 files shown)


dorde-antic:
@umangyadav @dhernandez0 this change was initially made to see if there is problem because kernelGen standard output wasn't explicitly closed. It wasn't the cause of the problem, but I am thinking if anyway kernelGen.stdout.close() should be present in _else:_ part of tuneMLIRKernel as it is in _if:_ part ?

mlir/utils/performance/tuningRunner.py
```diff
@@ -167,6 +167,7 @@ def tuneMLIRKernels(configs, confClass, paths: Paths, options: Options):
             config = confClass.fromCommandLine(commandLine, options.arch, options.numCU)
             tuningLoop = subprocess.Popen([paths.mlir_paths.rocmlir_tuning_driver_path, f"--tuning-space={options.tuningSpaceKind}", testVector],
                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            kernelGen.stdout.close()
```
dhernandez0, line 170: kernelGen here is not defined? it's only defined in the if:

dorde-antic: I see now, my bad. Before there weren't if and else branch, so I compared initially to that.

PR #1847, opened 2025-05-21
Implement GQA in AttentionConfiguration

dorde-antic:
- Implements GQA in AttentionConfiguration class in perfRunner.
- attentionSweeps.py will be adjusted with changes from this PR in https://github.com/ROCm/rocMLIR-internal/issues/1800

Resolves https://github.com/ROCm/rocMLIR-internal/issues/1834

### DIFF
--- a/mlir/utils/performance/perfRunner.py
+++ b/mlir/utils/performance/perfRunner.py
@@ -932,7 +932,7 @@ class GemmGemmConfiguration(PerfConfiguration):
 
 class AttentionConfiguration(PerfConfiguration):
     TABLE_COLUMNS = reportUtils.ATTN_TEST_PARAMETERS + ['TFlops']
-    def __init__(self, dtype: str, g: int, seq_len_q: int, seq_len_k: int, head_dim_qk: int, head_dim_v: int, with_attn_scale: bool, with_attn_bias: bool,
+    def __init__(self, dtype: str, g: int, seq_len_q: int, seq_len_k: int, num_heads_q: int, num_heads_kv: int, head_dim_qk: int, head_dim_v: int, with_attn_scale: bool, with_attn_bias: bool,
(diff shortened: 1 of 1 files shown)


mlir/utils/performance/perfRunner.py
```diff
@@ -1294,6 +1301,8 @@ def fromCommandLine(cls, argv, arch, numCU):
         g = None
         seq_len_q = None
         seq_len_k = None
+        num_heads_q = None
+        num_heads_kv = None
```
dhernandez0: it's completely fine if num_heads_q and num_heads_kv  are not defined. Please, check manually that perfRunner and tuningRunner work if these two are not defined.

dhernandez0: I think we can set the default value to = 1 instead of None?

dorde-antic: @dhernandez0 works when we set value to 1, doesn't work when it's None (ValueError: Incomplete Attention configuration)

mlir/utils/performance/perfRunner.py
```diff
@@ -1210,10 +1212,11 @@ def __init__(self, dtype: str, g: int, seq_len_q: int, seq_len_k: int, head_dim_
     def computeTFlops(self, ns, only_matmul_flops=True):
         # NaN will propagate as expected
         # Repeats are handled by the fact that we're using avarageNs
-        first_matmul_flops = 2.0 * self.g * self.seq_len_q * self.head_dim_qk * self.seq_len_k
+        g = self.g * max(self.num_heads_q, self.num_heads_kv)
```
dhernandez0, line 1216: please add a comment here: # GQA broadcasts so that both num_heads_q == num_heads_kv

dhernandez0 (approved):

PR #1853, opened 2025-05-27
Add missing LDS barriers to attention

dhernandez0:
There is a race condition that was reproducible on navi and MI100. If there is more than one iteration, some threads might be loading from LDS while others are in the next iteration writing to the same buffer. We add a barrier to prevent that. If the number of iterations is one, there's no need for the barrier.

solves: https://github.com/ROCm/rocMLIR-internal/issues/1811 https://github.com/ROCm/rocMLIR-internal/issues/1844

### DIFF
--- a/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
+++ b/mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
@@ -18,6 +18,7 @@
 // This pass converts rock.gridwise_gemm[_v2] into block- and threadwise ops
 //
 //===-----------------------------------------------------===//
+#include "mlir/Dialect/Affine/Analysis/LoopAnalysis.h"
 #include "mlir/Dialect/Rock/IR/Rock.h"
 #include "mlir/Dialect/Rock/IR/TransformMapBuilder.h"
 #include "mlir/Dialect/Rock/Passes.h"
@@ -52,6 +53,7 @@
 #include "mlir/Dialect/Rock/IR/AccelEmitter.h"
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/FormatVariadic.h"
+#include <optional>
 
 namespace mlir {
 namespace rock {
@@ -749,7 +751,8 @@ struct GridwiseAttentionAccelRewritePattern
       RegsAsMatrixSubTiles toLDSViews, Value storeBuffer,
       Value ldsTileByteBuffer, int64_t kpacksPerBlock, StringRef nonKDimName,
       int64_t kPerBlock, int64_t dPerBlock, int64_t copyKPerThread,
-      int64_t copyDPerThread, bool forceUnroll, bool rotateDWithK) const {
+      int64_t copyDPerThread, bool forceUnroll, bool rotateDWithK,
+      bool barrierBeforeWrite) const {
     Type elemType = cast<MemRefType>(regBuffer.getType()).getElementType();
     ArrayAttr storeBufferViews =
         invertTransforms(rewriter, loc, toLDSViews.threadSubTile);
@@ -772,6 +775,11 @@ struct GridwiseAttentionAccelRewritePattern
     // This will produce a (tid, iter) --> flat LDS view
     wrappedLds = transform(rewriter, wrappedLds, toLDSViews.blockSubTile);
     auto tid = rewriter.create<WorkitemIdOp>(loc, rewriter.getIndexType());
+
+    // add LDS barrier to avoid write before load of previous iteration is done
+    if (barrierBeforeWrite)
+      rewriter.create<LDSBarrierOp>(loc);
+
     rewriter.create<ThreadwiseWriteAllOp>(
         loc, storeBuffer, wrappedLds, /*extraViews=*/rewriter.getArrayAttr({}),
         /*extraIndices=*/ValueRange{tid}, GemmFeatures::none, StoreMethod::Set,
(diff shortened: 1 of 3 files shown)


dorde-antic (approved):

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2164,6 +2173,18 @@ struct GridwiseAttentionAccelRewritePattern
           kLoopIV = rewriter.createOrFold<affine::AffineApplyOp>(
               loc, reverseMap, ValueRange{kLoopIV, kIterationsGemm0Val});
         }
+
+        // LDS Barrier (issue 1811): some threads might be loading from LDS
+        // while others are in the next iteration (here), writing to LDS. This
+        // barrier prevents that.
+        bool mIterOneIter =
+            isa<affine::AffineForOp>(mLoopOp) && gemm0MBlocks == 1;
```
umangyadav: Wouldn't mLoopOp always be AffineForOp ?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2164,6 +2173,18 @@ struct GridwiseAttentionAccelRewritePattern
...
+        // no need to have the barrier if there's just one iteration
+        bool addBarrierFirstGemm = kIterationsGemm0 > 1 || !mIterOneIter;
```
umangyadav: If `mLoopOp` is not affineForOp then it will be true.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
dhernandez0: mLoopOp is created with getMLoop(). The loop can be scf::ForOp if we are doing kv-cache or causal attention. In that case, for kv-cache we don't know at compile time if the number of iterations is one or not. So, we assume we need the barrier. For causal attention, we need to do more than one iteration anyway, so no need to avoid the barrier.

Otherwise, it's an AffineForOp, and we can check the number of iterations (gemm0MBlocks) at compile time.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
dhernandez0: yes, see my previous comment.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2176,22 +2197,25 @@ struct GridwiseAttentionAccelRewritePattern
...
           if (failed(statusLoadQ)) {
             return failure();
           }
           ldsTileBufferQ =
               viewBufferAs(rewriter, ldsByteBufferQ,
                            vectorTypeOrSelf(elemTypeQ, gemm0kpack));
         }
+        // if we added a barrier in the previous block (load Q), there's no need
+        // to add it again here.
+        bool fallbackBarrier = addBarrierFirstGemm && gemm0K == gemm0KPerBlock;
```
umangyadav: is this working as expected ?

I expect it to be more like `fallbackbarrier = !addBarrierFirstGemm`

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2445,13 +2470,23 @@ struct GridwiseAttentionAccelRewritePattern
 
           Value ldsByteBufferV = createLDSByteBuffer(
               rewriter, loc, gemm1KPerBlock * gemm1MPerBlock, elemTypeV);
+
+          // LDS Barrier (issue 1811): some threads might be loading from LDS
+          // while others are in the next iteration (here), writing to LDS. This
+          // barrier prevents that. No need to have the barrier if there's just
+          // one iteration
+          bool addBarrierSecondGemm = gemm1MBlocks > 1;
```
umangyadav: Do we have unit-tests for GridwiseGemmToBlockwise for attention where we can add this case ?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
umangyadav: > mLoopOp is created with getMLoop(). The loop can be scf::ForOp if we are doing kv-cache or causal attention. In that case, for kv-cache we don't know at compile time if the number of iterations is one or not. So, we assume we need the barrier. For causal attention, we need to do more than one iteration anyway, so no need to avoid the barrier.

 

Can we add comment with explanation ?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
dhernandez0: There are two calls to loadAndStoreGemmInputTile() here. The first one only happens if "gemm0K != gemm0KPerBlock". We need to add a barrier only if addBarrierFirstGemm == true.

We want to make sure we add a barrier (if needed) to the first call to loadAndStoreGemmInputTile(). The first call is:
```c++
if(gemm0K != gemm0KPerBlock) {
  ...
  loadAndStoreGemmInputTile(..., addBarrierFirstGemm)
}
```
The second call is (unconditional):
```c++
loadAndStoreGemmInputTile(..., addBarrierFirstGemm && gemm0K == gemm0KPerBlock)
```

So, the goal is to add a barrier in the first if possible (gemm0K != gemm0KPerBlock), otherwise, add the barrier to the second call (avoid adding two barriers).

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
dhernandez0: sure

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
dhernandez0: good idea, I'll add tests. Note that all this will go away when we implement pipelining for attention. As I guess pipelining will take care of barriers between iterations.

dhernandez0: done

umangyadav (approved):

PR #1856, opened 2025-05-27
Generate current_seq_lens randomly with seed based on week number

dorde-antic:
Add current_seq_lens, generate it and other params randomly with seed based on a week number. Pass current_seq_lens to rocmlir-gen.

Resolves https://github.com/ROCm/rocMLIR-internal/issues/1835

### DIFF
--- a/mlir/utils/performance/attentionSweeps.py
+++ b/mlir/utils/performance/attentionSweeps.py
@@ -201,7 +201,7 @@ async def dropGoodConfig(config: AttentionConfiguration,
     and `None` on success or inapplicability"""
     result = await testAttentionConfig(config, options, paths)
     if not options.quiet:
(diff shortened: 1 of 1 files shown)


mlir/utils/performance/attentionSweeps.py
```diff
@@ -37,14 +38,17 @@
 BOOLS = [True, False]
 LOGFILE = 'failing_configs.csv'
 GFX_CHIP_RE = re.compile(r"gfx[0-9a-z]+")
+CURRENT_SEQ_LEN = None
 
+# Week number is used as seed
+seed = datetime.utcnow().isocalendar()[1]
+random.seed(seed)
```
dhernandez0, line 45: this should be in the attention sweep PR, not here.

PR #1842, opened 2025-05-20
Unify the blockwise lowering path

dhernandez0:
In this PR we unify the accelerator and non-accelerator code paths. List of changes:
- remove rock.gridwise_gemm, rock.blockwise_gemm, rock.threadwise_gemm (only _accel_ exist now)
- remove non accel params: InitParamsNonAccel, PopulateParams...
- manually copied quick tuning from wmma
- updated backward conv: it used to work for accel code only
- new FmaEmitter in AccelEmitter for non-accel. I'm calling it Fma because we hope to at least use fused multiply and add instructions. Also, because it was the way it was called in the original draft PR: https://github.com/ROCm/rocMLIR/pull/1544)
- tuning for FmaEmitter
- lots of changes in tests
- this enables non-accel attention 

Some details about non-accel implementation in AccelEmitter:
- it's only able to process one m/n at a time. I think this should be ok as a first implementation as our target are skinny GEMMs (1xK x KxN)
- the threads in the block are distributed in the K dimension, we do a blockwise reduction

TODO:
- [x] https://github.com/ROCm/rocMLIR/pull/1862 is merged
- [x] run CI with mfma and wmma disabled
- [x] IR unit tests for non-accel code path
- [x] present this work on knowledge session to make it easier to review

TODO for future PR:
- quick tuning for fma
- investigate tuning space for fma (GEMM and attention).
- investigate if we can bypass LDS for fma

PR #1862, opened 2025-05-30
Make sure arch can't be empty in AmdArchDb

dhernandez0:
In this PR I make sure all tests have arch defined. This is an issue for PR https://github.com/ROCm/rocMLIR/pull/1842 and has been discussed previously (see context below). The change makes sure that if no valid arch is defined, AmdArchDb fails.

The reason why this is an issue for https://github.com/ROCm/rocMLIR/pull/1842 is that if arch="", then waveSize is defined as 64 even if it's a Navi arch.

I had to fix some tests:
- some were converted from empty or hardcoded arch to %arch
- some e2e were changed to %arch (switching from non-accel to accel run of the test)
- some had to be left to a particular arch (non-accel) to avoid breaking assumptions of the tests
- some e2e test are hard to convert to accel without rocmlir-gen changes (I think), so I've left them as non-accel run and left a comment

context:
https://github.com/ROCm/rocMLIR/pull/1790#discussion_r2075009804
https://github.com/ROCm/rocMLIR/pull/1842/files#r2114210305


mlir/test/Conversion/TosaToRock/no-mixed-layout.mlir
```diff
@@ -3,7 +3,7 @@
 // CHECK: rock.conv({{.*}}) {{.*}} {{{.*}}, filter_layout = ["g", "k", "c", "y", "x"], input_layout = ["ni", "gi", "ci", "hi", "wi"], output_layout = ["no", "go", "ko", "ho", "wo"]{{.*}}}
 
 module {
-  func.func @test(%arg0: tensor<1x512x1x1xf32>, %arg1: tensor<1x384x28x28xf32>, %arg2: tensor<512x384x1x1xf32>) -> tensor<1x512x28x28xf32> attributes {kernel, mhal.arch = "amdgcn-amd-amdhsagfx90a:sramecc+:xnack-"} {
```
umangyadav, line 6: 😮

mlir/test/Dialect/Rock/lowering_reuse_lds.mlir
```diff
@@ -5,7 +5,7 @@
 
 
 // CHECK-LABEL: func.func @rock_reuse_two
-func.func @rock_reuse_two() attributes{arch = "", block_size = 256 : i32, grid_size = 320 : i32, kernel} {
+func.func @rock_reuse_two() attributes{arch = "gfx90a:sramecc+:xnack-", block_size = 256 : i32, grid_size = 320 : i32, kernel} {
```
umangyadav: Why do you need to hardcode arch for this test ?

dhernandez0: good catch, this one doesn't need to be arch specific. Changed.

mlir/test/fusion/e2e/rock-bcast-idxmap.e2e.template
```diff
@@ -1,14 +1,16 @@
...
+// RUN: rocmlir-gen -ph -print-results -rand none %s | sed s/##TOKEN_ARCH##/%arch/g | rocmlir-driver -arch %arch -c  | mlir-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void --mattr=-avx512bf16 | FileCheck %s
 // ALLOW_RETRIES: 2
 
 // CHECK: 65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65,      65
 
+// NOTE: this tests non-accel path
```
umangyadav, line 5: Is this note correct ? It seems to be using `%arch`.

Is it using non-accel because it is using `features = None` ?

mlir/test/fusion/e2e/rock-bcast.e2e.template
```diff
@@ -1,14 +1,15 @@
-// RUN: rocmlir-gen -ph -print-results -rand none %s | rocmlir-driver -arch %arch -c  | mlir-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void --mattr=-avx512bf16 | FileCheck %s
+// RUN: rocmlir-gen -ph -print-results -rand none %s | sed s/##TOKEN_ARCH##/%arch/g | rocmlir-driver -arch %arch -c  | mlir-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext --entry-point-result=void --mattr=-avx512bf16 | FileCheck %s
 // ALLOW_RETRIES: 2
 
+// NOTE: this tests non-accel path
```
umangyadav, line 3: Is this note correct ?  It uses `features = Dot`

mlir/test/fusion/pr-e2e/rock-bcast-exp.e2e.mlir
```diff
@@ -1,14 +1,16 @@
-// RUN: rocmlir-gen -ph -print-results -rand none %s | rocmlir-driver -arch %arch -c  | mlir-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_float16_utils%shlibext --entry-point-result=void | FileCheck %s
+// RUN: rocmlir-gen -ph -print-results -rand none %s | sed s/##TOKEN_ARCH##/%arch/g | rocmlir-driver -arch %arch -c  | mlir-runner -O2 --shared-libs=%linalg_test_lib_dir/libmlir_rocm_runtime%shlibext,%conv_validation_wrapper_library_dir/libconv-validation-wrappers%shlibext,%linalg_test_lib_dir/libmlir_runner_utils%shlibext,%linalg_test_lib_dir/libmlir_float16_utils%shlibext --entry-point-result=void | FileCheck %s
 // ALLOW_RETRIES: 2
 
-
 // CHECK:  73,      73,      73,      73,      73,      73,      73,      73,      73,      73,      73,      73,      73,      73,      73,      73
+
+// NOTE: this tests non-accel path
```
umangyadav, line 5: Is this note correct ?

umangyadav (commented):
I find that we don't necessarily need to maintain `features` attribute on the ops if we keep correct `arch` correctly on all the tests.

umangyadav (approved):

dhernandez0:
> I find that we don't necessarily need to maintain `features` attribute on the ops if we keep correct `arch` correctly on all the tests.

I agree, there are multiple things that are duplicated. We can keep them as attributes of the main function (they already are). Such as arch, num_cu. Then, we can drop arch, num_cu and features in rock.conv, rock.gemm, ..., because you can call amdarchdb(mainfunc.arch) to get features.

That would make some test arch independent (just set arch="%arch"). We can open a ticket with low priority.

mlir/test/fusion/e2e/rock-bcast-idxmap.e2e.template
dhernandez0: yes, features = none means no accel. The issue is that making this test use accel code path is tricky. You'd have to duplicate it, one for features = wmma, other for features = mfma. For migraphx tests it's easier because you just set arch=%arch and it works.

mlir/test/fusion/e2e/rock-bcast.e2e.template
dhernandez0: AFAIK dot does nothing, I'm not sure why it's there, I guess it was introduced to represent dot instructions (there are some at least for navis I think). But currently it's not used anywhere. So, it's the same as features = none.

mlir/test/fusion/pr-e2e/rock-bcast-exp.e2e.mlir
dhernandez0: yes

PR #1865, opened 2025-05-30
Adjust lit parallelism for MFMA architectures based on GPU type

stefankoncarevic:
Set LLVM_LIT_ARGS -j value dynamically for MFMA codepath based on GPU architecture:
- gfx908 → -j 8
- gfx90a → -j 30

Solves: https://github.com/ROCm/rocMLIR-internal/issues/1845 and https://github.com/ROCm/rocMLIR-internal/issues/1841

PR #1866, opened 2025-05-30
Limit number of workers

dhernandez0:
Limit number of workers to 8. This doesn't seem to affect MI300 but there's no way to tell them apart currently, only "mfma". IMO let's fix this and we can iterate in the future to allow more workers.

Solves: https://github.com/ROCm/rocMLIR-internal/issues/1845 and https://github.com/ROCm/rocMLIR-internal/issues/1841


dhernandez0:
closing as https://github.com/ROCm/rocMLIR/pull/1865 is doing the same

PR #1870, opened 2025-06-01
Refactor splitConfigFile to use sed-based slicing

dorde-antic:
**Don't merge until manual nightly check passes.**

- Refactors splitConfigFile to use sed-based slicing instead of awk and may resolve https://github.com/ROCm/rocMLIR-internal/issues/1849 .


umangyadav (approved):

umangyadav:
I'll rekick CI once weekly has generated databases with convolution. Do not update the branch until then.

PR #1882, opened 2025-06-11
Attention: return LSE (log-sum-exp)

dhernandez0:
This PR introduces the ability to return LSE (log-sum-exp), note that migraphx requires log to be log2.

I've added return_lse to the problem config because it will change how the kernel behaves significantly (extra output tensor).

Please check the comment of computeLse() in GridwiseGemmToBlockwise.cpp to understand how it's computed.

migraphx integration will come in a follow-up PR.


mlir/lib/Dialect/Rock/IR/RockDialect.cpp
```diff
@@ -1872,6 +1872,10 @@ LogicalResult GridwiseAttentionAccelOp::verify() {
     return emitError("NPerBlock should be divisble by kpack.");
```
umangyadav:
```suggestion
    return emitError("NPerBlock should be divisible by kpack.");
```

mlir/lib/Dialect/Rock/utility/loweringUtils.cpp
```diff
@@ -409,6 +409,20 @@ Value mlir::rock::normalizeMatrix(Value matrix, OpBuilder &b, Location loc,
   return b.create<TransformOp>(loc, matrix, normalizeAttr);
 }
 
+Value mlir::rock::padVector(Value vector, OpBuilder &b, Location loc,
+                            StringRef firstDim, int64_t firstDimPad) {
+  if (firstDimPad == 0)
+    return vector;
+  ArrayRef<int64_t> shape = cast<MemRefType>(vector.getType()).getShape();
+  BottomUpTMBuilder padder(b, {"gemmG", firstDim}, shape, loc);
```
umangyadav, line 418: Would be good to assert it has it is two dimensional to begin with.

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -2359,6 +2370,11 @@ static void getAttentionTypes(SmallVectorImpl<Type> &result,
         MemRefType::get(currentSeqDims, elemTypes[currentSeqLenIndex]);
     result.push_back(currSeqLenType);
   }
+  if (returnLSE) {
+    SmallVector<int64_t> lseDims{groupSize * numHeadsQ, sequenceLengthQ};
+    MemRefType lseType = MemRefType::get(lseDims, elemTypes[lseIndex]);
+    result.emplace_back(lseType);
```
umangyadav: Rest of the places are using `result.push_back()`. For consistency change this to use `push_back` too or change all to use `emplace_back`

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -3699,16 +3727,55 @@ static func::FuncOp createCpuAttentionKernelWithMlir(ModuleOp module,
...
       builder, loc,
-      cast<ShapedType>(normilizedQkTensor.getType()).getElementType(),
-      normilizedQkTensor);
+      cast<ShapedType>(normalizedQkTensor.getType()).getElementType(),
+      normalizedQkTensor);
   auto expsSums = createOpAndInfer<tosa::ReduceSumOp>(
       builder, loc, cast<ShapedType>(expsTensor.getType()).getElementType(),
       expsTensor, reductionAxis);
+  Value lseTensor;
+  if (returnLSE) {
```
umangyadav, line 3744: Just put a comment up there it is doing, it would help
`(log(sum_exp) + m) / ln(2)`

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
```diff
@@ -3699,16 +3727,55 @@ static func::FuncOp createCpuAttentionKernelWithMlir(ModuleOp module,
...
+    lseTensor = createOpAndInfer<tosa::AddOp>(builder, loc, expsSumElemType,
+                                              lseTensor, qkMaxs);
+    // we want to return LSE where L is log2
+    constexpr float log2RecipConst = 1.44269504f;
+    assert(isa<FloatType>(expsSumElemType));
+    std::pair<APFloat, llvm::detail::opStatus> floatRes =
+        rock::createAPFloat(expsSumElemType, log2RecipConst);
+    APFloat log2Recip = floatRes.first;
+    auto status = floatRes.second;
+    if (expsSumElemType.getIntOrFloatBitWidth() >= 32)
```
umangyadav: Ideally for ref implementation we should use F32 accumulator for reductions.  

I notice that we do cast before doing `expSum` 


```
  Value qkTensor = builder.createOrFold<tosa::CastOp>(
      loc,
      cast<ShapedType>(qkTensorBeforeConversion.getType())
          .clone(firstGemmOutElemType),
      qkTensorBeforeConversion);
```

mlir/test/Dialect/Rock/gridwise_attention_accel_lowering.mlir
```diff
@@ -395,6 +395,50 @@ func.func @gridwise_attn_causal_kvcache(%arg0: memref<1x384x64xf32>, %arg1: memr
 
 // -----
 
+// CHECK: @gridwise_attn_lse_kvcache
```
umangyadav: Check-label ?

umangyadav (commented):
Looks good overall. 

Do we need to do something for parameter sweeps after this changes ?

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
dhernandez0: IMO we want the cpu and gpu versions to do the same. Currently, I think we compute softmax in f16/bf16 etc. So, I agree doing reduction in f32 makes more sense. But I think it's better to fix that at the kernel level first (let migraphx choose reduction type?), then change the cpu version as well.

dhernandez0:
> Looks good overall.
> 
> Do we need to do something for parameter sweeps after this changes ?

Yes: https://github.com/ROCm/rocMLIR/pull/1830#issuecomment-2964415251

mlir/tools/rocmlir-gen/rocmlir-gen.cpp
umangyadav: Thiking more it should be okay to not use higher accumulator for most cases. 

It is normalizing  by subtracting max. 

Value of `exp{x-max}` would always be between 0 and 1.  And Sum of those could only potentially overflow for very large reduction dimension e.g. more than 65k for fp16.

umangyadav: Do you know if we have cases of `seq_len_k` being more than 65k ?

dhernandez0: overflow is not the only issue, if you add 256 + 1 in bf16, you get 256. So, accumulating you can reach 256 and then you have to add 1.

umangyadav (approved):

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1161,6 +1161,63 @@ struct GridwiseAttentionAccelRewritePattern
     }
   }
 
+  // This computes LSE (log-sum-exp)
+  // Note that this happens at the end of the kernel, so m and l are not running
+  // sum/max anymore. They are the final values.
+  // input = gemm0 output
+  // x = input/log(2) -> we divide by log(2) to be able to use exp2()
```
umangyadav, line 1168: Q: 
Where is it dividing by `log(2)` ?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1161,6 +1161,63 @@ struct GridwiseAttentionAccelRewritePattern
...
+    ArrayAttr lseBufferTrs;
+    std::tie(lseBuffer, lseBufferTrs, std::ignore) =
+        untransform(rewriter, lseBufferView);
+    MemRefType lseBufferViewType = cast<MemRefType>(lseBufferView.getType());
+    Type outElemType = lseBufferViewType.getElementType();
+    int64_t g1Npt = lseBufferViewType.getShape()[0];
+    int64_t g1Mpt = lseBufferViewType.getShape()[1];
+    Value zero = rewriter.createOrFold<ConstantIndexOp>(loc, 0);
+    auto loop = rewriter.create<TransformingForOp>(
+        loc, ArrayRef<ValueRange>{{zero, zero}, {zero, zero}},
```
umangyadav, line 1194: It is using iteration space with 2 axis on.y. Is it guranteed that `sumRowbuffer` and `maxRowBuffer` are two dimensional ? Can we assert that ?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1161,6 +1161,63 @@ struct GridwiseAttentionAccelRewritePattern
...
+        ArrayRef<Attribute>{rewriter.getArrayAttr({}), lseBufferTrs},
+        /*bounds=*/ArrayRef<int64_t>{g1Npt, g1Mpt},
+        /*strides=*/ArrayRef<int64_t>{1, 1},
+        /*forceUnroll=*/true, /*useIndexDiffs=*/true);
+    {
+      OpBuilder::InsertionGuard guard(rewriter);
+      rewriter.setInsertionPointToStart(loop.getBody());
+      Block::BlockArgListType upperCoords = loop.getLowerCoords(0);
```
umangyadav, line 1203: `upperCoords = getLowerCoords()` sounds confusing. 

Can you change that name or perhaps put a comment that they both are same since transform map is empty ?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -2639,6 +2704,14 @@ struct GridwiseAttentionAccelRewritePattern
       createTypeConversionLaGeneric(rewriter, loc, outAccBuffer,
                                     outAccBufferOutTyped);
     }
+    if (lse) {
+      // it must be guaranteed by the verifier
+      assert(op.getEnableSoftmax());
+      assert(lseBuffer);
+      Value lseBufferView = transform(
+          rewriter, lseBuffer, attentionOutAccBufferThreadSubTileViewMaps);
```
umangyadav, line 2717: It is being transformed here and later inside `computeLSE` it is being untransformed. Can we possibly eliminate that ?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
```diff
@@ -1161,6 +1161,63 @@ struct GridwiseAttentionAccelRewritePattern
...
+    {
+      OpBuilder::InsertionGuard guard(rewriter);
+      rewriter.setInsertionPointToStart(loop.getBody());
+      Block::BlockArgListType upperCoords = loop.getLowerCoords(0);
+      Block::BlockArgListType lseBufferCoords = loop.getLowerCoords(1);
+
+      Value ldMaxRowBuffer = rewriter.create<InBoundsLoadOp>(
+          loc, inputElemType, maxRowBuffer, ValueRange{upperCoords[0]});
+      Value ldSumRowBuffer = rewriter.create<InBoundsLoadOp>(
+          loc, inputElemType, sumRowBuffer, ValueRange{upperCoords[0]});
```
umangyadav, line 1209: Q : 

maxRowBuffer and `sumRowBuffer` are reductions. But here they seem to be two dimensional. Are they broadcasted here ?

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
dhernandez0: we multiply by 1/log(2). This is already existing code, look for postProcessFirstGemmSplat<ElementwiseMultOp>() call in GridwiseGemmToBlockwise.cpp

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
dhernandez0: sumRowbuffer  and maxRowBuffer have one dimension only, not 2. See how we use the indices below to load from the buffers.

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
dhernandez0: I'm following the same style of scaleFinalOutput, createAttentionRowStateCorrections, updateRowSum, expSubstractMaxFromGemm0...

I think the advantage is that otherwise, we need to pass both the buffer and the transforms. This way we only pass one argument. IMO we can do this better with a struct or a class to keep both buffer and transforms. However, I'd prefer that we do that in another PR and change everything at the same time to keep the same style throughout the file

mlir/lib/Dialect/Rock/Transforms/GridwiseGemmToBlockwise.cpp
dhernandez0: we only use one dimension: ValueRange{upperCoords[0]}

umangyadav (approved):

PR #1883, opened 2025-06-11
Fix output file path for rocprofv3

djramic:
This PR fixes inconsistent output paths in rocprofv3.


umangyadav (approved):

mlir/utils/performance/perfRunner.py
```diff
@@ -1573,7 +1579,7 @@ def getFusionTestInfo(filename, paths: Paths):
     return testEntry
 
 def runFusionKernel(filename, rocmlirGenArgs, paths: Paths):
-    os.system("rm -f "+BENCHMARKING_STATS_FILE_NAME)
+    os.system("rm -f "+getProfilerOutputPath(arch, BENCHMARKING_STATS_FILE_NAME))
```
umangyadav: `arch` is undefined here

mlir/utils/performance/perfRunner.py
```diff
@@ -158,6 +158,12 @@ def getNanoSeconds(fileName):
         csv_file.close()
         return result
 
+def getProfilerOutputPath(arch, baseOutPath):
+    chip = GFX_CHIP_RE.search(arch).group(0)
+    if(chip not in ["gfx942"]):
```
dhernandez0, line 164: add TODO to figure out what we need to do with gfx950

mlir/utils/performance/perfRunner.py
```diff
@@ -1378,7 +1384,7 @@ def benchmarkExternal(cls, commandLine, paths: Paths, arch, numCU):
             raise ValueError("rocblas-benchmark-driver not built")
         benchmarkArgs = config.generateMlirDriverCommandLine("")
         # remove the result file generated by rocprof in previous benchmarking
-        os.system("rm -f "+BENCHMARKING_STATS_FILE_NAME)
+        os.system("rm -f "+getProfilerOutputPath(arch, BENCHMARKING_STATS_FILE_NAME))
```
dhernandez0: I think we shouldn't use os.system("rm ..."), this prevents us from catching errors.

I would prefer something like:
if os.path.exists(path):
  os.remove(path)

then, if we can't remove the file (permission or other issues) we'll actually get an error.

mlir/utils/performance/perfRunner.py
```diff
@@ -1433,7 +1439,7 @@ def benchmarkMLIR(commandLine, confClass, paths: Paths, arch, numCU, tuningDb: M
 
     runConfigWithMLIR(config, paths, arch, rocmlir_gen_flags)
     # get nanoseconds from rocprof output.
-    nanoSeconds = getNanoSeconds(BENCHMARKING_STATS_FILE_NAME)
+    nanoSeconds = getNanoSeconds(getProfilerOutputPath(arch,BENCHMARKING_STATS_FILE_NAME))
```
dhernandez0: lit: space between "arch," and BENCHMARKING_STATS_FILE_NAME

dhernandez0 (approved):

umangyadav (approved):

mlir/utils/performance/perfRunner.py
```diff
@@ -1573,7 +1583,8 @@ def getFusionTestInfo(filename, paths: Paths):
     return testEntry
 
 def runFusionKernel(filename, rocmlirGenArgs, paths: Paths):
-    os.system("rm -f "+BENCHMARKING_STATS_FILE_NAME)
+    if os.path.exists(getProfilerOutputPath(getChip(), BENCHMARKING_METRICS_FILE_NAME)):
+        os.remove(getProfilerOutputPath(getChip(), BENCHMARKING_METRICS_FILE_NAME))
```
umangyadav: I see `getChip()` call repeated manytimes in this function. You can just once assign to variable and use tha

PR #1884, opened 2025-06-11
Fix Azure builds

umangyadav:
Azure builds are failing with error that `assert` is undefined. 

Adding `#include <cassert>` to see if it fixes it.


umangyadav (commented):
Azure builds are passing now

dorde-antic (approved):

PR #1886, opened 2025-06-12
[DRAFT] CI: Fail on tuningRunner errors or invalid/empty tuning output files

dorde-antic:
This PR introduces:
- `set -e` for every tuna-script command in Tune rocMLIR phase which instructs bash to immediately exit if any command has a non-zero exit status. It would help detecting errors in tuningRunner execution based on it's return value
- Check for tuningRunner errors in logs
- Check for empty tuning files (or with only header or more (duplicate) headers in first lines)

Resolves https://github.com/ROCm/rocMLIR-internal/issues/1870


mlir/utils/jenkins/Jenkinsfile
```diff
@@ -969,28 +969,70 @@ pipeline {
...
-                                                    -t ${WORKSPACE}/MITuna -f mlir_quick_tuning_${CHIP}.tsv"""
+                                            sh """
+                                            set -e
+                                            ../mlir/utils/tuna/tuna-script.sh -o convolution \
+                                                -c ../mlir/utils/performance/configs/tier1-conv-configs -s quick \
+                                                -t ${WORKSPACE}/MITuna -f mlir_quick_tuning_${CHIP}.tsv
+                                            """
+                                            // Check for tuningRunner errors in logs
+                                            sh """
+                                            if grep -q "Error executing command: tuningRunner" build/*.log 2>/dev/null; then
```
umangyadav: Where these `.log` files are being generated ?

dorde-antic: I'll add that output from tuna-script is saved in .log file. Or is there better way to access logs from tuna-script after the whole group of tuna-script commands are executed?

> Where these `.log` files are being generated ?

mlir/utils/jenkins/Jenkinsfile
```diff
@@ -969,28 +969,70 @@ pipeline {
...
+                                                echo "ERROR: Detected tuningRunner execution error in logs!"
+                                                exit 1
+                                            fi
+                                            """
+                                            // Check for empty or duplicate header in tuning files
+                                            sh """
+                                            for f in build/mlir_tuning_${CHIP}.tsv build/mlir_quick_tuning_${CHIP}.tsv; do
```
umangyadav, line 994: I wonder if it would be better to instead do this error handling directly inside tuningRunner.py. 

i.e. 
Currently `main()` inside tuningRunner.py is not correctly returning return code 
https://github.com/ROCm/rocMLIR/blob/3a275a49c44198ad1c0154491b0b42abd5c13a29/mlir/utils/performance/tuningRunner.py#L431

`rocmlir_worker.py` is checking for `retcode` here 

https://github.com/ROCm/MITuna/blob/c3c2f87404c0e0e2f50bfd279950822ce56a4126/tuna/rocmlir/rocmlir_worker.py#L194

Therefore i think it would be much better to do error handling directly through those scripts instead of manually doing it inside this Jenkinsfile.

Same goes for `tuna-script.sh` too. If possible we should do return code checking directly inside it.  or you can put `set -e` inside that

mlir/utils/jenkins/Jenkinsfile
umangyadav: >  Or is there better way to access logs from tuna-script after the whole group of tuna-script commands are executed?

I am not sure, you can check for return code for each script (if they are propagating it properly in first place) possibly and see if that helps

mlir/utils/jenkins/Jenkinsfile
dhernandez0: I agree, also tuna-script.sh does a lot of things, lots of those commands can fail as well.

dorde-antic:
TODO: 
check if this causes -> empty file error because of more adjacent headers. If yes, fix that
image

dorde-antic:
https://github.com/ROCm/rocMLIR/pull/1913 - new PR because CI doesn't show option "Build with parameters" for this PR

PR #1893, opened 2025-06-18
Increase time out on resetGPUs to allow resetting more than 8 GPUs

umangyadav:
One of the server has 16 GPUs and it take more than 5 minutes to ResetGPUs on it, increase timeout limit to allow such cases.


stefankoncarevic (approved):

dorde-antic (approved):

umangyadav:
`gpu reset` passed on server with 16 GPUs merging now.

PR #1894, opened 2025-06-18
Define datatypes for attention dynamically based on chip type

dorde-antic:
- Set `DATA_TYPES_ATTENTION` dynamically based on chip type in `perfRunner`, since WMMA doesn't support f32.
- Adjust changes to support reused functions in `tuningRunner`:

   - `tuningRunner` calls `perfRunner.getAttentionConfigurations()` which uses `DATA_TYPES_ATTENTION` inside function implementation, therefore, change `DATA_TYPES_ATTENTION` for had to be done above that function, but also functions such as `getChip()`, `getArch()` and `hip_check()` since they are all executed when setting `DATA_TYPES_ATTENTION`.


mlir/utils/performance/perfRunner.py
```diff
@@ -117,6 +118,39 @@ def find_mlir_build_dir() -> str:
...
+def getChip():
+    arch = getArch()
+    chip = GFX_CHIP_RE.search(arch).group(0)
+    return chip
+
+if getChip().startswith('gfx94'):
+    DATA_TYPES_ATTENTION = DATA_TYPES_ATTENTION_WMMA
+else:
+    DATA_TYPES_ATTENTION = DATA_TYPES_ATTENTION_MFMA
+
```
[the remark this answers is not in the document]
umangyadav: Use this suggesiton from copilot

mlir/utils/performance/perfRunner.py
```diff
@@ -117,6 +118,39 @@ def find_mlir_build_dir() -> str:
...
+    # select first agent by default
+    return list(agents)[0]
+
+def getChip():
+    arch = getArch()
+    chip = GFX_CHIP_RE.search(arch).group(0)
+    return chip
+
+if getChip().startswith('gfx94'):
+    DATA_TYPES_ATTENTION = DATA_TYPES_ATTENTION_WMMA
```
umangyadav: You have it flipped here. `
gfx942 shouldbe using `MFMA`

mlir/utils/performance/perfRunner.py
```diff
@@ -117,6 +118,39 @@ def find_mlir_build_dir() -> str:
...
+        print("WARNING: Using the first agent by default. If you want to use a different agent, please set the HIP_VISIBLE_DEVICES environment variable.")
+    # select first agent by default
+    return list(agents)[0]
+
+def getChip():
+    arch = getArch()
+    chip = GFX_CHIP_RE.search(arch).group(0)
+    return chip
+
+if getChip().startswith('gfx94'):
```
umangyadav: We support attention with MFMA on gfx90a as well i think 
you should be adding those too. And MI35x
cc: @dhernandez0

dorde-antic: it was supposed to be startswith('gfx9'), typo

mlir/utils/performance/perfRunner.py
```diff
@@ -117,6 +118,43 @@ def find_mlir_build_dir() -> str:
...
+    if(len(agents) > 1):
+        print(f"WARNING: Found {len(agents)} different kinds of agents on the same machine :  {', '.join(agents)}")
+        print("WARNING: Using the first agent by default. If you want to use a different agent, please set the HIP_VISIBLE_DEVICES environment variable.")
+    # select first agent by default
+    return list(agents)[0]
+
+def getChip():
+    arch = getArch()
+    chip = GFX_CHIP_RE.search(arch).group(0)
+    return chip
```
umangyadav, line 147: I wonder if this code can be made to execute only once. 

e.g. in C++ equivalent : 

```
std::string getChip() {
static arch = getArch();
static chip = re.search(arch).group(0)
return chip;
}
```

mlir/utils/performance/perfRunner.py
```diff
@@ -117,6 +118,39 @@ def find_mlir_build_dir() -> str:
...
+    for device in range(device_count):
+        props = hip.hipDeviceProp_t()
+        hip_check(hip.hipGetDeviceProperties(props,device))
+        agent = props.gcnArchName.decode('utf-8')
+        agents.add(agent)
+    if(len(agents) > 1):
+        print(f"WARNING: Found {len(agents)} different kinds of agents on the same machine :  {', '.join(agents)}")
+        print("WARNING: Using the first agent by default. If you want to use a different agent, please set the HIP_VISIBLE_DEVICES environment variable.")
+    # select first agent by default
+    return list(agents)[0]
```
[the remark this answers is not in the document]
umangyadav: nit: 
Same for the `getArch()` if can be made to execute only once and cache the results, it would be nice.

umangyadav (approved):
Looks good except some nits

mlir/utils/performance/perfRunner.py
dorde-antic: - What first came to my mind was to use global chip and global arch, so then there wouldn't be the need to call getChip() or getArch() everytime we need them inside perfRunner
- Looked on internet, there is @lru_cache(maxsize=None) decorator that could be put above the function, which should ensure that function is run once (i haven't used this before, so not sure abt behaviour)
- Similar to @lru_cache, we could use functools.cache from functools and put @cache decorator above
- We could also have fixed class if we don't want to use global for chip and arch variables and then have there arch, chip getters and similar things if needed (it should be then executed only once too) or we could have something like Options class in parameterSweeps in which we store some things we need more times, and then initialize values once in main()
@umangyadav

PR #1887, opened 2025-06-13
Attention LSE migraphx integration

dhernandez0:
This PR introduces migraphx integration for LSE output (attention). List of relevant changes:
- Add migraphx.reduce_max (we will get decomposed softmax instead of migraphx.softmax)
- Add lse return value for rock.attention (needed for this PR as this is used before bufferization)
- TosaToRock: code to match LSE pattern
- RockDialect: add effects for LSE (and currSeqLen, it was missing). Fix AttentionOp::getOutArgument().
- Bufferization: special case for attention because it can have two outputs

Notes:
- at some point we want to remove migraphx.softmax if migraphx will not use it. I understand from now on, they will send softmax decomposed in its operations.
- reduce_max could potentially be used as output for migraphx. I think they do not plan to do it, but we can add some checks. The issue is that there's no atomic_max for MI architectures AFAIK.

TODO:
- [x] merge https://github.com/ROCm/rocMLIR/pull/1882 first
- [ ] @shivadbhavsar to try this branch and check if the IR matches what migraphx can generate

PR #1895, opened 2025-06-19
Attention: split-kv implementation

dhernandez0:
This PR implements flash decoding (split-kv for attention). The idea is that we use multiple blocks to process the the seqLenKV dimension (see https://pytorch.org/blog/flash-decoding/). By default, split-kv=1 means no flash decoding.

If split-kv>1, we have to return LSE as well. And both LSE and attention output will have batch*splitKV. 

Changes of the PR:
- add split-kv to operations and verifiers
- gridwise-to-gemm: use more threads (*splitkv) and pad if needed
- added split_kv as problem config param
- dynamically assign the start and end (iterations) in seqLenKV for each of the blocks (when KV-Cache or causal are enabled as well)
- early exit if the block has no work to do
- rocmlir-gen changes to enable split-kv. Note that the cpu version does not implement split-kv, it always does split-kv=1. Then, the gpu version returns the two outputs (LSE and attn out), and we combine those in cpu to compare against cpu. This is hacky but easier than implementing the second kernel on the GPU. Finally, both cpu and gpu outputs are broadcasted (splitKV times in the batch dimension) because the cpu and gpu buffers have the same size (batch = groupSize * headDimQ * splitKV) and modifying the code to allow different cpu and gpu buffer size was more complicated.

I've tested flash decoding with the following problem config which is similar to llama2 (same but with and without split-kv), with split_kv = 2,4,8,16,32,64,128:
```
-transO false -transV false -transK true -transQ false -t f16 -g 32 -seq_len_q 1 -seq_len_k 4096 -head_dim_qk 128 -head_dim_v 128 -with-attn-scale False -with-attn-bias False -causal False -return_lse False -split_kv 1
-transO false -transV false -transK true -transQ false -t f16 -g 32 -seq_len_q 1 -seq_len_k 4096 -head_dim_qk 128 -head_dim_v 128 -with-attn-scale False -with-attn-bias False -causal False -return_lse True -split_kv N
```

Here are the results (MI300X):
|   | DataType | Chip   | ReturnLSE | SplitKV | PerfConfig                      | TFlops   |  microseconds |
| - | -------- | ------ | --------- | ------- | ------------------------------- | -------- | ---- |
| 0 | f16      | gfx942 | FALSE     | 1       | attn:v1:128,128,32,8,32,32,16,1 | 0.55 | 122.02 |
| 1 | f16      | gfx942 | TRUE      | 2       | attn:v1:128,128,32,8,32,32,16,1 | 1.05 | 63.91 |
| 2 | f16      | gfx942 | TRUE      | 4       | attn:v1:128,128,32,8,32,32,16,1 | 1.92 | 34.95 |
| 3 | f16      | gfx942 | TRUE      | 8       | attn:v1:128,128,32,8,32,32,16,1 | 2.94 | 22.83 |
| 4 | f16      | gfx942 | TRUE      | 16      | attn:v1:64,128,32,8,32,16,16,1  | 3.49 | 19.23 |
| 5 | f16      | gfx942 | TRUE      | 32      | attn:v1:64,64,32,8,32,16,16,1    | 3.5   | 19.17 |
| 6 | f16      | gfx942 | TRUE      | 64      | attn:v1:64,64,32,8,32,16,8,1      | 3.34   | 20.09 |
| 7 | f16      | gfx942 | TRUE      | 128    | attn:v1:64,64,32,8,32,16,8,1      | 3.35   | 20.03 |
| 8 | f16      | gfx942 | TRUE      | 256    | attn:v1:64,64,32,8,32,16,8,1      | 3.35   | 20.03 |

So, we can get 6x speed up. Take into account I haven't added the second kernel for split-kv, so the speed up must be a bit worse.

As you guys asked in the previous meeting, I've run the kernels above (split-kv and perfconfig) with different values of currSeqLen (KV-Cache), run-time is in microseconds:
|                 | splitKV=1 | 2     | 4     | 8     | 16    | 32    | 64    | 128   | MIN   | Speed up |
| --------------- | --------- | ----- | ----- | ----- | ----- | ----- | ----- | ----- | ----- | -------- |
| currSeqLen=0    | 6.67      | 7.18  | 6.82  | 7.02  | 5.31  | 6.07  | 6.34  | 6.84  | 5.31  | 1.26     |
| currSeqLen=64   | 6.66      | 7.26  | 6.81  | 7.02  | 5.47  | 5.78  | 5.83  | 6.84  | 5.47  | 1.22     |
| currSeqLen=127  | 7.57      | 6.81  | 7.86  | 7.78  | 5.68  | 5.46  | 6.48  | 6.91  | 5.46  | 1.39     |
| currSeqLen=500  | 17.94     | 11.46 | 7.53  | 7.10  | 6.42  | 7.28  | 7.43  | 8.01  | 6.42  | 2.79     |
| currSeqLen=1000 | 31.69     | 17.52 | 11.06 | 7.28  | 6.87  | 8.66  | 8.32  | 9.31  | 6.87  | 4.61     |
| currSeqLen=2000 | 61.36     | 32.40 | 21.12 | 14.91 | 10.58 | 10.06 | 12.03 | 12.71 | 10.06 | 6.10     |
| currSeqLen=3000 | 95.57     | 49.75 | 27.89 | 19.77 | 16.29 | 17.98 | 17.05 | 14.83 | 14.83 | 6.44     |
| currSeqLen=4095 | 126.15    | 65.08 | 36.25 | 23.96 | 19.79 | 19.22 | 20.88 | 21.14 | 19.22 | 6.56     |

Experiments for seq_len_kv=100000:
|                   | splitKV=1 | 2       | 4      | 8      | 16     | 32     | 64     | MIN    | Speed up |
| ----------------- | --------- | ------- | ------ | ------ | ------ | ------ | ------ | ------ | -------- |
| currSeqLen=100    | 7.36      | 7.38    | 6.88   | 6.99   | 6.59   | 6.92   | 7.09   | 6.59   | 1.12     |
| currSeqLen=1000   | 31.60     | 17.84   | 10.57  | 7.23   | 8.67   | 11.16  | 11.37  | 7.23   | 4.37     |
| currSeqLen=10000  | 305.55    | 159.72  | 81.76  | 47.65  | 46.83  | 44.06  | 41.46  | 41.46  | 7.37     |
| currSeqLen=100000 | 3219.59   | 1657.56 | 823.59 | 465.80 | 484.37 | 406.57 | 402.07 | 402.07 | 8.01     |

TODO:
- [x] merge https://github.com/ROCm/rocMLIR/pull/1887 first
- [x] add more tests
- migraphx integration in a future PR


mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
```diff
@@ -58,7 +58,10 @@ GemmSize calculatePaddedGemmSize(const InitParams &params, GemmSize gemmSize,
 /// a given gemm size requires. Returns None if no padding is needed. The
 /// values in the returned gemm context represent the number of 0s that need to
 /// be added to the given dimension.
-std::optional<GemmSize> requiredPadding(Attribute params, GemmSize gemmSize);
+std::optional<GemmSize> requiredPadding(Attribute params, GemmSize gemmSize,
```
umangyadav, line 62: Can you describe in docstring what `mulBy.*` params are for ? It may not be obvious why there are here just by looking at this function.

mlir/lib/Dialect/Rock/Transforms/GridLayoutEmitter.h
```diff
@@ -31,13 +31,14 @@ namespace mlir {
...
 
-/// Struct containing the {g,m,n} block coordinates of a block
+/// Struct containing the {g,m,n,split} block coordinates of a block
 /// with a given bid. I.e., block bid will compute C[g_block, m_block, n_block]
 /// output
 struct GridCoordinates {
   Value g_block;
   Value m_block;
   Value n_block;
+  Value split_block;
```
umangyadav, line 46: `split_block` is specific to attention only. I wonder if we should make 

```cpp
struct AttnGridCoordinates : GridCoordinates {
}
```

mlir/include/mlir/Dialect/Rock/Tuning/GridwiseGemmParams.h
dhernandez0: done

mlir/lib/Dialect/Rock/Transforms/GridLayoutEmitter.h
dhernandez0: done

justinrosner (approved):
Changes look good to me

